PR #268, opened 2019-10-11
Add feature to create a session on behalf of other users by admin

adrysn:
This pull-request introduces ownership panel in launch session dialog for domain- and super-admins. Typing access key of a user and then click refresh icon. It fetches the groups and scaling groups allowed for that access key (user). Ensure this launch is for some user by checking the check-box. They admins can launch a session on behalf of the access key specified.

image


src/components/backend-ai-resource-monitor.ts
```diff
@@ -390,6 +399,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
     );
     this.shadowRoot.querySelector('#environment').addEventListener('selected-item-label-changed', this.updateLanguage.bind(this));
     this.shadowRoot.querySelector('#version').addEventListener('selected-item-label-changed', this.updateMetric.bind(this));
+    this.shadowRoot.querySelector('#owner-group').addEventListener('selected-item-label-changed', this._fetchSessionOwnerScalingGroups.bind(this));
```
inureyes: This event fires when the page / component is loaded / initialized at the first time. Therefore we need a solution to skip unnecessary API call.

src/lib/backend.ai-client-node.ts
```diff
@@ -2162,7 +2162,8 @@ class ScalingGroup {
   }
 
   list(group = 'default') {
-    const queryString = `/scaling-groups?group=${this.client.current_group}`;
+    if (!group) group = this.client.current_group;
```
inureyes: Make sure whether client.current_group exists or not. Because client.current_group is not a client member variable. It is set and used only on backend.ai-console code.

src/components/backend-ai-resource-monitor.ts
```diff
@@ -105,6 +106,10 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
   @property({type: Boolean}) metadata_updating;
   @property({type: Boolean}) aggregate_updating = false;
   @property({type: Object}) scaling_group_selection_box;
+  /* Parameters required to launch a session on behalf of other user */
+  @property({type: String}) ownerDomain;
+  @property({type: Array}) ownerGroups;
```
inureyes: set the default value even it is set on `init_resource()`. `init_resource()` is not intended as a constructor / initializations. It is intended to reset the current state of resource monitor.

Suggestion:
```
@property({type: String}) ownerDomain = '';
@property({type: Array}) ownerGroups = [];
@property({type: Array}) ownerScalingGroups = [];
```

src/lib/backend.ai-client-node.js
```diff
@@ -1991,7 +1991,9 @@ class ScalingGroup {
         }
     }
     list(group = 'default') {
-        const queryString = `/scaling-groups?group=${this.client.current_group}`;
+        if (!group)
```
inureyes: same as .ts review.

src/lib/backend.ai-client-node.ts
adrysn: Changed from `if (!group)` -> `if (!group && this.client.current_group)`.

src/components/backend-ai-resource-monitor.ts
adrysn: Moved `owner-group`'s `addEventListener` only when needed (just after press refresh button for the first time).

src/components/backend-ai-resource-monitor.ts
adrysn: I set the default values just right at the declaration.

src/lib/backend.ai-client-node.ts
inureyes: I just removed the lines.

PR #299, opened 2019-12-06
Feature/tus

gofeel:
@inureyes  // Update this commit with current master branch.


inureyes:
@gofeel The branch is too diverged since we cleaned up the repository for the public release. Therefore I cherry-picked f6e21ffa79 to  feature/#305-upload-component .

Issue #311, opened 2020-01-02
Enlarge session resource text input on session start dialog


inureyes:
* Enlarge session resource text input on session start dialog
 * 
<img width="463" alt="new_session_input_issue_200102" src="https://user-images.githubusercontent.com/949915/71672402-2cadaa80-2db9-11ea-82a1-06c763e09e5d.png">

inureyes:
* [ ] adopt `mwc-slider` to project and test it.
  * [ ] create new component: `lablup-slider` compatible with paper-slider
    1. [ ] replace `paper-slider` with `mwc-slider` without text input
    2. [ ] implement text input on the end of each `mwc-slider`
    3. [ ] test the behavior
 * [ ] replace the `paper-slider` with `lablup-slider`

inureyes:
@lizable

PR #322, opened 2020-01-10
refs #311 : modify - disable switching wl-expansions by enter key

lizable:
Now user can write down with enter key inside expansion slots.
User can switch expansion slots by mouse-click instead of pressing enter key.


adrysn:
Looks good!

PR #332, opened 2020-01-22
Feature/#277 replace policy input

lizable:
Now, user can update resource policy following these conditions.
- wherever contains checkbox, users have to input values or check unlimited box.
   if not, then alerts it will notify invalid input area immediately.
- if user inputs zero value, then it will be regarded as unlimited(∞).


inureyes:
@lizable I reviewed the PR but it its deleted and new PR is appeared. Any reason? My review was here: https://github.com/lablup/backend.ai-console/pull/329

p.s. There is a style convention in .editorconfig but is changed in your recent modifications. I will reformat the code after closing and merging so do not worry about it now.

lizable:
@inureyes I changed the branch name, and it made the previous branch with PR closed.
Since branch name should be relevant to the PR, therefore I renamed it.
It is not a general way to make a new PR and close the previous one.
However, keeping branch name with "slider", but no "slider" is in the component might be a bit problematic in the meaning.

inureyes:
@lizable Looks good to me. In addition to the UI, please add `unlimited` to folder capacity too. Thanks!

PR #338, opened 2020-01-24
Fix broken image registry/name/tag parsing for custom images

adrysn:
* Fix broken image parsing for images without namespace (eg: `registry.test.io/gpp:test-t114`).
* Handle registry with port number. For example, now it is possible to correctly parse a registry like `127.0.0.1:8888`.
* Add default image information for custom images which are not registered in `image_metadata.json`.
* Fix version number is displayed twice for NGC images.
* Hide registry text from session info since it may expose the IP address of private docker registry.
* Display image namespace, if exists, of custom a image.
* I set custom image's version string color with grey.

Before this PR:
image

After this PR:
image


inureyes:
LGTM. And working well with current setup.

PR #344, opened 2020-02-03
Get/update bootstrap script

adrysn:
Bootstrap script is a script which runs inside a session container upon launched. This pull-request add client API and UI to see and update the bootstrap script.

* `UserConfig` class is added in the JS client API.
* Add an edit button to see/update the bootstrap script in user settings page.
* Delete codemirror sources and replace them with `wc-codemirror`.


inureyes:
LGTM. UI needs to enhance so it will work as the new issue.

inureyes:
Checking trigger: OP#288

PR #370, opened 2020-02-21
Feature/#354 resource policy setting validation check

lizable:
User input is automatically validated. Only letters, numbers and -_ is valid.
Error message shows differently based on user input.


inureyes:
LGTM

PR #371, opened 2020-02-21
Hotfix/#362 session dialog slider fix

lizable:
Slider in session dialog is disabled, depending on resource limit.


inureyes:
LGTM

Issue #373, opened 2020-02-21
Version on session list displays wrong number


inureyes:
* Version on session list displays wrong number

<img width="167" alt="image" src="https://user-images.githubusercontent.com/949915/75081042-3057e500-5551-11ea-90c7-0dfdf4bd2680.png">

 * When I started TF 2.1, 2.0, 1.15, 1.14 kernel, list displays 1.6.

PR #375, opened 2020-02-22
Use baseversion for version tag, only when category exists

adrysn:
This pull request resoles #373.

The root cause of this problem is that the [`item` in `sessionInfoRenderer `](https://github.com/lablup/backend.ai-console/blob/master/src/components/backend-ai-session-list.ts#L1069) always contains the same `tag`, regardless of the real version (which is stored in `rowData.item.baseversion`). For example, when I launched two `python-tensorflow` sessions each with version 1.13 and 2.0, the `item.tag` is always 2.0 even for 1.13 session:

image

I don't know the logic behind displaying those items and tags (it's too complex in my opinion), so I just override the `tag` with real `rowData.item.baseversion`.


inureyes:
LGTM.

PR #378, opened 2020-02-24
Feature/#376 auto changing scaling group

### LINKED ISSUE #376: Scaling group not updated when Project is changed in session list page
In sessions list page, user can change the project (group). Upon changing the project, sessions only in the project are displayed.

If there are two projects which can access different scaling groups, changing the project will result in error since the scaling group is not updated upon project selection. Request to `check-preset` uses updated project with old scaling group, so the scaling group does not exists in the updated group.

This bug prevents admins to set different scaling groups to different projects.

Steps to produce:
* Create two projects: `gA`, `gB`
* Create two scaling groups: `sA`, `sB`
* Set `gA`'s allowed scaling groups to `sA`
* Set `gB`'s allowed scaling groups to `sB`
* Visit session list page in console. Probably, the `gA` project is automatically selected. In this case, check-preset request uses `gA` project with `sA` scaling group, which is okay.
* Change the project to `gB`
* Error occurs since check-preset request now uses `gB` project with `sA` scaling group, not `sB` scaling group.

lizable:
This PR resolves #376.

### DIFF
--- a/src/components/backend-ai-resource-monitor.ts
+++ b/src/components/backend-ai-resource-monitor.ts
@@ -121,7 +121,6 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
   @property({type: Array}) ownerGroups;
   @property({type: Array}) ownerScalingGroups;
   @property({type: Boolean}) project_resource_monitor = false;
-  @property({type: Boolean}) is_project_Changed = false;
 
   constructor() {
     super();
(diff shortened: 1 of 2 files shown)


adrysn:
There is a dropdown selection issue with scaling group. If there are two scaling groups in a group, there is a mismatch on the selected scaling group and visually displayed one. To reproduce this issue,

* Set two scaling groups are allowed to a group.
* In session launch dialog, select the second scaling group from the list.
* Close the session launch dialog.
* Open session launch dialog again.
* The scaling group's value is set to the first scaling group.
* However, when you click the scaling group's dropdown, the second one is selected.

src/components/backend-ai-resource-monitor.ts
```diff
@@ -684,11 +686,10 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
       if (window.backendaiclient.is_admin) {
         ownershipPanel.style.display = 'block';
       } else {
-
-
         ownershipPanel.style.display = 'none';
       }
-
+      // this value initialization is temporary due to non-dynamic value recongition of paper-dropdown
+      this.shadowRoot.querySelector('#scaling-groups').value = this.scaling_groups[0].name;
```
adrysn: Specifically, this does not update the selected label on the dropdown menu.

PR #381, opened 2020-02-25
Feature/#380 dropdown menu with export csv

### LINKED ISSUE #380: move export csv feature into menu in user/credentials and sessions
It would be better to make a dropdown menu and move export csv feature into menu.

- [x] create a dropdown menu inside in both user/credentials and sessions

- [x] move export csv feature into the dropdown menu and remove the button used as exporting csv file.

lizable:
This PR resolves #380

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@material/mwc-drawer": "^0.13.0",
     "@material/mwc-icon-button": "^0.13.0",
     "@material/mwc-list": "^0.13.0",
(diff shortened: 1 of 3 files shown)


adrysn:
I think there is an error when downloading user list:
image

adrysn:
In session list page, dropdown menu does not appear in the upper right cornor:
image

This is reproduced when you refresh the browser from the session list page. If you visit other page first, and then visit session list page, then it will appear.

PR #398, opened 2020-03-02
Feature/#396 remember selected project group

### LINKED ISSUE #396: Remember last selected group (project)
When a user belongs to multiple groups (projects), s/he can select the current group by the dropdown menu in the upper-left corner. However, page refresh resets the group information. Users often work in one specific group consecutively, so, I think it would provide better user experience if user's last-selected group  is remembered and applied after page refresh.

What this issue should achieve:
* Login with a user with multiple groups.
* One group (A) is selected automatically.
* Select another group (B) by the dropdown menu in the upper-left corner.
* Refresh the page.
* Group B should be selected automatically instead of A.

lizable:
This PR resolves #396

### DIFF
--- a/src/components/backend-ai-console.ts
+++ b/src/components/backend-ai-console.ts
@@ -441,6 +441,7 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
     this.user_id = window.backendaiclient.email;
     this.domain = window.backendaiclient._config.domainName;
     this.current_group = this._readRecentProjectGroup();
+    window.backendaiclient.current_group = this.current_group;
     this.groups = window.backendaiclient.groups;
     let groupSelectionBox = this.shadowRoot.getElementById('group-select-box');
     if (window.backendaiclient.isAPIVersionCompatibleWith('v4.20190601') === false) {
--- a/src/components/backend-ai-console.ts
+++ b/src/components/backend-ai-console.ts
@@ -440,7 +440,7 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
   _refreshUserInfoPanel() {
     this.user_id = window.backendaiclient.email;
     this.domain = window.backendaiclient._config.domainName;
-    this.current_group = window.backendaiclient.current_group;
+    this.current_group = this._readRecentProjectGroup();
     this.groups = window.backendaiclient.groups;
     let groupSelectionBox = this.shadowRoot.getElementById('group-select-box');
     if (window.backendaiclient.isAPIVersionCompatibleWith('v4.20190601') === false) {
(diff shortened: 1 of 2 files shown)


src/components/backend-ai-console.ts
```diff
@@ -644,6 +646,8 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
       for (let i = 0; i < keys.length; i++) {
         const key = keys[i];
         if (/^(backendaiconsole\.login\.)/.test(key)) localStorage.removeItem(key);
+        if (/^(backendaiconsole\.projectGroup)/.test(key))
+        localStorage.removeItem(key);
```
adrysn: Combine these lines into one, or use bracket `{` with indentation.

src/components/backend-ai-console.ts
```diff
@@ -716,6 +721,22 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
       document.dispatchEvent(event);
     }
   }
+  _readRecentProjectGroup() {
+    let value: string | null = localStorage.getItem('backendaiconsole.projectGroup');
+    if (value !== null && value !== '' && value !== '""') {
```
adrysn: I think the condition here can be simplified, unless you want explicit comparison, since `undefined`, `null`, `''`, and `""` are all evaluated as `false`:

```javascript
if (value) {
```

Be careful, however, `[]` and `{}` are evaluated as `true` in `if` statement.

If you want a compact expression here, it is also possible to use the ternary operator.

```javascript
return value ? value : this.backendaiclient.current_group;
```

src/components/backend-ai-resource-monitor.ts
```diff
@@ -492,7 +492,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
     this.shmem_request = 0.0625;
     this.gpu_request = 0;
     this.session_request = 1;
-    this.scaling_groups = [{name: ''}]; // if there isno scaling group, set the name as emptry string
+    this.scaling_groups = [{name: ''}]; // if there is no scaling group, set the name as empty string
```
adrysn, line 495: good catch

src/components/backend-ai-console.ts
```diff
@@ -716,6 +721,22 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
...
+      return value;
+    } else {
+      return window.backendaiclient.current_group;
+    }
+  }
+
+  _writeRecentProjectGroup(value) {
+    if (value !== null && value !== '' && value !== '""') {
```
adrysn: here as well.

src/components/backend-ai-console.ts
lizable: I guess there's a scarce chance of getting ```[]```  or ```{}``` type in this case, but I'll try to catch these values as well.

src/components/backend-ai-console.ts
lizable: I understand. It could be simplified, too. :)

adrysn:
LGTM.

PR #400, opened 2020-03-02
Feature/397 refactor input items

### LINKED ISSUE #397: Refactor signup routine to use native form validator
* Refactor signup routine to use native form validator

inureyes:
This PR resolves #397


adrysn:
I think I found two problems.

* Signup dialog is opened only when SIGNUP button is clicked twice. Perhaps a rendering timing issue?
* Resource policy cannot be created since policy name is always invalid.

image

inureyes:
> I think I found two problems.
> 
> * Signup dialog is opened only when SIGNUP button is clicked twice. Perhaps a rendering timing issue?
> * Resource policy cannot be created since policy name is always invalid.
> 
> image

Fixed both problems.

inureyes:
@adrysn Fixed and found that was long-lasting bug (not from this issue)

PR #433, opened 2020-03-18
feat: enable log button for terminated sessions

adrysn:
* Only for >= 20.03. Logging for terminated sessions is not supported <= 19.09.

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -1100,14 +1100,15 @@ export default class BackendAiSessionList extends BackendAIPage {
                                @click="${(e) => this._openTerminateSessionDialog(e)}"
                                icon="delete"><wl-icon>delete</wl-icon></wl-button>
           ` : html``}
-          ${this._isRunning ? html`
+          ${this._isRunning || globalThis.backendaiclient.APIMajorVersion > 4 ? html`
             <wl-button fab flat inverted class="fg blue controls-running" icon="assignment"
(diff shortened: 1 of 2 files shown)


inureyes:
@adrysn globalThis does not fire the changed event after component initialization so we need to create local variable to match with globalThis variable. e.g. this.switch1 = globalThis.switch1 in firstUpdated() / backend-ai-connected section.

src/components/backend-ai-session-list.ts
```diff
@@ -1100,14 +1100,15 @@ export default class BackendAiSessionList extends BackendAIPage {
                                @click="${(e) => this._openTerminateSessionDialog(e)}"
                                icon="delete"><wl-icon>delete</wl-icon></wl-button>
           ` : html``}
-          ${this._isRunning ? html`
+          ${this._isRunning || globalThis.backendaiclient.APIMajorVersion > 4 ? html`
```
inureyes: globalThis does not fire the changed event after component initialization so we need to create local variable to match with globalThis variable. 

e.g. 
```
firstUpdate() {
...
   document.addEventListener('backend-ai-connected', () => {
      this.switch1 = globalThis.switch1;
...
}
...
${this._isRunning || this.switch1 > 4 ? html`
```

adrysn: Would there be any case that APIMajorVersion can be changed after component initialization?

adrysn: I registered API major version as a property.

PR #441, opened 2020-03-20
fix: variable v should be a string to call includes method

adrysn:
Two consecutive creation of resource presets fail since some values of input controls are set to numbers after the first creation, which causes error in the following line:

```javascript
const wrapper = v => v !== undefined && v.includes('Unlimited') ? 'Infinity' : v;
```

If `v` is number, then there is no `includes` method. This pull-request just converts `v` to `string` before calling `includes` method:

```javascript
const wrapper = (v) => {
  v = v.toString();
  return v !== undefined && v.includes('Unlimited') ? 'Infinity' : v;
};
```

### DIFF
--- a/src/components/backend-ai-resource-preset-list.ts
+++ b/src/components/backend-ai-resource-preset-list.ts
@@ -148,20 +148,20 @@ class BackendAiResourcePresetList extends BackendAIPage {
         <div class="layout horizontal wrap center">
         ${rowData.item.resource_slots['cuda.device'] ?
         html`
-        <div class="layout horizontal configuration">
-          <wl-icon class="fg green">view_module</wl-icon>
-          <span>${this._markIfUnlimited(rowData.item.resource_slots['cuda.device'])}</span>
-          <span class="indicator">GPU</span>
-        </div>
-      ` : html``}
+          <div class="layout horizontal configuration">
+            <wl-icon class="fg green">view_module</wl-icon>
+            <span>${this._markIfUnlimited(rowData.item.resource_slots['cuda.device'])}</span>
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-resource-preset-list.ts
```diff
@@ -551,7 +551,10 @@ class BackendAiResourcePresetList extends BackendAIPage {
   }
 
   _createPreset() {
-    const wrapper = v => v !== undefined && v.includes('Unlimited') ? 'Infinity' : v;
+    const wrapper = (v) => {
+      v = v.toString();
+      return v !== undefined && v.includes('Unlimited') ? 'Infinity' : v;
```
inureyes: In JavaScript, comparison of undefined  `!== undefined` will generate strange behavior through browsers. The correct comparison is `typeof(v) !== 'undefined'` since JavaScript is prototype-based language.

adrysn: Updated!

src/components/backend-ai-resource-preset-list.ts
```diff
@@ -553,7 +553,7 @@ class BackendAiResourcePresetList extends BackendAIPage {
   _createPreset() {
     const wrapper = (v) => {
       v = v.toString();
-      return v !== undefined && v.includes('Unlimited') ? 'Infinity' : v;
+      return typeof(v) !== undefined && v.includes('Unlimited') ? 'Infinity' : v;
```
inureyes: typeof(v) !== 'undefined'

I'll fix and merge it.

PR #446, opened 2020-03-23
fix : asynchronous resource template refresh (#445)

### LINKED ISSUE #445: Hide GPU-enabled preset when kernel does not require GPU resource
* Hide GPU-enabled preset when kernel does not require GPU resource

inureyes:
This PR resolves #445


src/components/backend-ai-resource-monitor.ts
```diff
@@ -1253,6 +1255,9 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
           }
         });
         this.resource_templates = available_presets;
+        if (this.resource_templates_filtered.length === 0) {
```
inureyes, line 1258: This is intended when 
 * {4 CPU / 8 RAM}
 * {2 CPU / 4 RAM / 1 GPU}
resource presets are present and if only 3 CPU / 6 RAM is allocatable, use the second template even it has GPU option.

PR #450, opened 2020-03-25
Bugfix/web proxy 500 error

### LINKED ISSUE #449: Webpage via proxy sometimes produces 500 error while app is not ready yet
* Webpage via proxy sometimes produces 500 error while app is not ready yet
 * The websocket proxy issues an app execution command, opens the websocket proxy to the port used by the app, and attempts to connect. However, if the app cannot launch any service or app on the port until that point, the return value of the connection is 500.

inureyes:
This PR resolves #449


inureyes:
@adrysn It is timing bug so hard to simulate. We can delay the app running process via some tricks like adding sleep on Jupyter startup via terminal before running Jupyter.

PR #457, opened 2020-03-30
feat: httpproxy support for local wsproxy websocket connection

### LINKED ISSUE #456: Add option for local wsproxy to work behind external http proxy
In some cases, console app users are allowed to connect to the outside network only through a http proxy. Console app itself can be proxied by [`--proxy-server`](https://www.electronjs.org/docs/api/command-line-switches#--proxy-serveraddressport) setting, but this setting does not seem to be applied to the local wsproxy, so there needs to be a way to connect to the endpoint (manager or console-server) for web socket through external http proxy.

adrysn:
This pull-request resolves #456.

When console app is behind a corporate http proxy, let local proxy can pass throught it by sending `CONNECT` request first. If `EXT_HTTP_PROXY` environment variable is set, the URL will be used as an intermediate http proxy.

The value of `EXT_HTTP_PROXY` should include protocol, host, and port (if exists).
ex) http://10.20.30.40:3128


inureyes:
LGTM

inureyes:
@adrysn It should be documented in README.md

adrysn:
Updated `README.md`.

PR #461, opened 2020-04-06
feat: GUI option to change image pull behavior

### LINKED ISSUE #460: GUI option to change image pull behavior
* Add option to change image pull behavior for Backend.AI 20.03

inureyes:
*This PR resolves #460

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -466,6 +466,10 @@
     "RequireFGPUPlugin": "Requires Backend.AI Virtual CUDA API Layer Plugin.",
     "TPU": "TPU",
     "DescTPU": "Use TPU accelerator.",
-    "RequireTPUPlugin": "Requires resource nodes on Google Cloud with Cloud TPU enabled / binded."
+    "RequireTPUPlugin": "Requires resource nodes on Google Cloud with Cloud TPU enabled / binded.",
+    "Image": "Image",
+    "ImagePullBehavior": "Image Pull Behavior",
+    "DescImagePullBehavior": "Change image pulling behavior.",
+    "GUI": "GUI"
   }
 }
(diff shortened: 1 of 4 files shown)


resources/i18n/en.json
```diff
@@ -460,12 +460,25 @@
     "Scheduler": "Scheduler",
     "JobScheduler": "Choose default job scheduler",
     "Require1912orAbove": "Requires Backend.AI 19.12 or above.",
+    "Require2003orAbove": "Requires Backend.AI 19.12 or above.",
```
adrysn: Perhaps, `19.12` -> `20.03`.

src/components/backend-ai-settings-view.ts
```diff
@@ -286,16 +323,16 @@ export default class BackendAiSettingsView extends BackendAIPage {
...
-        console.log(response);
-      });
-    } else {
-      globalThis.backendaiclient.setting.set('docker/image/auto_pull', 'digest').then((response) => {
+  setImagePullingBehavior(e) {
+    if (e.target.selected === null) return false;
+    if (e.target.selected.value !== this.options['image_pulling_behavior'] && ['none', 'digest', 'tag'].includes(e.target.selected.value)) {
+      globalThis.backendaiclient.setting.set('docker/image/auto_pull', e.target.selected.value).then((response) => {
+        this.notification.text = _text("notification.SuccessfullyUpdated");
+        this.notification.show();
```
adrysn: I think `this.options['image_pulling_behavior']` should be updated with `e.target.selected.value` after successfully setting `docker/image/auto_pull`. Otherwise, the update will not work properly if you select a different option and then select the previous value again due to the checks in `if` statement.

resources/i18n/en.json
inureyes: Fixed.

src/components/backend-ai-settings-view.ts
inureyes: Fixed.

PR #464, opened 2020-04-06
feat: Queue support to upload files

### LINKED ISSUE #392: Queue support to upload files
The current implementation attempts to upload all files at once when uploading multiple files. In this case, too many connections are used at the same time. Also, the overall speed is the same, which slows down file-by-file uploads. To solve this problem, upload files sequentially.

inureyes:
* This PR resolves #392

### DIFF
--- a/src/components/backend-ai-storage-list.ts
+++ b/src/components/backend-ai-storage-list.ts
@@ -1274,14 +1274,12 @@ export default class BackendAiStorageList extends BackendAIPage {
     if (session !== null) {
       (this.fileUploadQueue as any).push(session);
     }
-    if (this.fileUploadCount < this.concurrentFileUploadLimit) {
-      let queuedSession;
-      for (let i = this.fileUploadCount; i < this.concurrentFileUploadLimit; i++) {
-        if (this.fileUploadQueue.length > 0) {
-          queuedSession = this.fileUploadQueue.shift();
-          this.fileUploadCount = this.fileUploadCount + 1;
-          queuedSession.start();
-        }
+    let queuedSession;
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-storage-list.ts
```diff
@@ -1266,6 +1270,22 @@ export default class BackendAiStorageList extends BackendAIPage {
...
+    if (this.fileUploadCount < this.concurrentFileUploadLimit) {
+      let queuedSession;
+      for (let i = this.fileUploadCount; i < this.concurrentFileUploadLimit + 1; i++) {
+        if (this.fileUploadQueue.length > 0) {
+          queuedSession = this.fileUploadQueue.shift();
+          this.fileUploadCount = this.fileUploadCount + 1;
+          queuedSession.start();
+        }
+      }
+    }
```
adrysn: I think this works well. Just a minor question that I was unable to understand.

* `if` statement in line 1277 needed? `for` loop in two lines below anyway will not run if file upload count is larger than concurrent upload limit.
* Reason of `+ 1` in line 1279? It seems like three files can concurrently be uploaded if `concurrentFileUploadLimit` is set to 2.

inureyes: 1. If 2 files are already uploading, code 1278-1285 does not run. If one or no file is uploading now, the codes will run to start file sessions. So it is intentional.
2. Bug. I will fix it.

inureyes: @adrysn Fixed the second one.

inureyes: @adrysn  Fixed the first one.

PR #466, opened 2020-04-07
feat: GUI option to allow/disallow images from specific repository

### LINKED ISSUE #421: GUI option to allow/disallow images from specific repository.
* GUI option to allow/disallow images from specific repository.
 * When admin add a repository, it is blocked by default.

inureyes:
This PR resolves #421


src/components/backend-ai-registry-list.ts
```diff
@@ -328,6 +332,37 @@ class BackendAIRegistryList extends BackendAIPage {
...
+      this.allowed_registries.push(hostname);
+      this.notification.text = _text("registry.RegistryTurnedOn");
+    } else {
+      let index = this.allowed_registries.indexOf(hostname);
+      if (index !== 1) {
+        this.allowed_registries.splice(index, 1);
+      }
+      this.notification.text = _text("registry.RegistryTurnedOff");
+    }
+    input.allowed_docker_registries = this.allowed_registries;
```
adrysn: Local variable `input` have no role here. It seems to be safe to delete it.

inureyes: It is the remnants of old implementation (did not see the light). Fixed.

PR #495, opened 2020-04-20
Feature/heterogeneous resource slot

### LINKED ISSUE #407: Handle heterogeneous resource slots for v1912+
From v19.12, individual agents may report different sets of resource slots.
For example, agents from the scaling group A is configured with CUDA fractional mode reporting `cuda.shares` and agents from the scaling group B is configured with CUDA discrete mode reporting `cuda.device`.
The console must be able to handle such heterogeneity.

* There is no longer "global" discrete/fractional mode discretion.
* When creating new sessions, the manager matches intrinsic resource slots and non-zero resource slots only and ignore zero-valued resource slots in the session creation request.
  - Setting a non-overlappable set of resource slots (e.g., `cuda.shares` and `cuda.devices`) at the same time to non-zero values will result "no available instances" error.

We could assume that each scaling group (aka resource group) has the same set of resource slots, but please note that the underlying manager has no such restrictions.

inureyes:
This PR resolves #407 

Now heterogeneous slots are handled by GUI.

TODO: Device can be selected via ResourcePreset but not manual. Need to implement UI for choosing device (cuda.device, cuda.shares, rocm.device, tpu.devices) in advanced resource allocation, after merge.


inureyes:
This PR is necessary to implement #496 .

PR #500, opened 2020-04-22
fix: blank Data & Storage page

### LINKED ISSUE #499: Blank Data & Storage page
Sometimes, Data & Storage page is displayed as blank. It seems like this is a timing issue whe `render` method is called before `globalThis.backendaiclient` is defined.

image

adrysn:
This PR fixes #499.


adrysn:
This fix create another problem: sometimes, created vfolder is not refreshed. We need to fix with other way.

adrysn:
> This fix create another problem: sometimes, created vfolder is not refreshed. We need to fix with other way.

I think this is fixed.

Issue #498, opened 2020-04-20
Unified event tasker


inureyes:
* Implement unified event tasker for 20.03+

## TODO 
 * [x] Global event tasker
    * [x] Task class format
    * [x] Task pool with handlers
    * [x] Garbage collector
 * [x] Global progress indicator pool
 * [x] Migrate old codes to user global objects

 * Task monitor UI component will be handled as separate issue.

PR #504, opened 2020-04-25
feat: unified event tasker

inureyes:
* This PR implements core components of unified tasker (#498)

## Implemented
 * Global event tasker
    * Task class format
    * Task pool with handlers
    * Garbage collector
 * Global progress indicator pool
 * Migrate old codes to user global objects

## Not implemented
  * Task monitor UI component (will be handled as separate issue)


src/components/backend-ai-indicator.ts
```diff
@@ -45,19 +46,25 @@ export default class BackendAIIndicator extends LitElement {
...
-    this.text = 'Initializing...';
     this.mode = mode;
+    await this.updateComplete;
     this.dialog.show();
   }
 
   set(value, text = '') {
     this.value = value / 100.0;
     this.text = text;
+    if (value >= 1) {
```
adrysn: Shouldn't it be `this.value`, not `value`?

src/components/backend-ai-tasker.ts
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+import {css, customElement, html, LitElement, property} from "lit-element";
+
+class Task {
+  taskid: string;
+  taskobj: Object;
+  status: string;
+  created_at: number;
+
```
adrysn, line 14: What about adding `finished_at` property as well, for possible future use before garbage collection.

src/components/backend-ai-indicator.ts
inureyes: Fixed at 413c09ce498

src/components/backend-ai-tasker.ts
inureyes: Fixed at 6aa5765

PR #535, opened 2020-05-17
feat: add support for signup user email verification

### LINKED ISSUE #536: Add support for signup user email verification
For Backend.AI public service, let's add a feature to support email verification.

adrysn:
This PR resolves #536.

* New component: `backend-ai-email-verification-view`.
* Add route `verify-email` for email verification.
  - Visiting this url will try to verify token, not login.
  - Token should be delivered as a query string by `verification_code`.
    - If verified, success dialog is popup and redirects to root page.
    - If not verified, fail dialog is popped-up with re-send verification email form.
  - If token is not supplied, fail dialog is popped-up with re-send verification email form.
* New `Cloud` class in the JavaScript SDK.
  - Add `verify_email` and `send_verification_email` in the `Cloud` API.

## How to test signup with email verification feature
* `cloud-beta-plugin` should be installed (by pip) in the manager.
* Signup with test account (with invitation token). You will see a dialog which states that an email is sent. Check your mailbox to discover delivered email.
  image
* Click VERIFY link. You should see email verified dialog. Or, you can see fail dialog if token is expired, or some error occurs.
  image
* In case of failure, user can resend a verification email again in the fail dialog.


src/components/backend-ai-email-verification-view.ts
```diff
@@ -0,0 +1,143 @@
...
+          width: 100%;
+        }
+      `
+    ];
+  }
+
+  _initClient(apiEndpoint: string) {
+    this.consoleShell = document.querySelector('#console-shell');
+    this.consoleShell.appBody.style.visibility = 'visible';
+    this.notification = this.consoleShell.notification;
```
inureyes: This should be `this.notification = globalThis.lablupNotification;`

src/components/backend-ai-email-verification-view.ts
```diff
@@ -0,0 +1,143 @@
...
+        mwc-textfield {
+          width: 100%;
+        }
+      `
+    ];
+  }
+
+  _initClient(apiEndpoint: string) {
+    this.consoleShell = document.querySelector('#console-shell');
+    this.consoleShell.appBody.style.visibility = 'visible';
```
inureyes, line 49: Why do we need this part? This part produces FOUC issue occasionally.

src/components/backend-ai-console.ts
```diff
@@ -493,6 +500,10 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
         this.menuTitle = _text("console.menu.Logs");
         this.updateTitleColor('var(--paper-deep-orange-800)', '#efefef');
         break;
+      case 'verify-email':
```
inureyes: Let's use summary and verify-email block together.

src/components/backend-ai-email-verification-view.ts
adrysn: Fixed.

src/components/backend-ai-email-verification-view.ts
adrysn: Without this, verification dialogs are not visible to users. Is there other workaround to make dialogs visible?

src/components/backend-ai-console.ts
adrysn: Merged `verify-email` block with `summary`.

src/components/backend-ai-email-verification-view.ts
inureyes: Ok. It is only called while email verification is processed. Let's see after merging.

PR #546, opened 2020-05-25
feat: UI for Background task UI (#542)

### LINKED ISSUE #542: UI for Background task
* Indicator pool and global tasker is ready but no UI is present.
 * Add UI for background tasks and indicators

 * This issue is developed in conjunction with the UI reorganization issue (#518)

inureyes:
This PR resolves #542


src/components/backend-ai-tasker.ts
```diff
@@ -97,11 +96,11 @@ export default class BackendAiTasker extends LitElement {
...
-  add(task, taskid: string = '') {
+  add(title, task, taskid: string = '', tasktype = 'general') {
     if (taskid === '') {
       taskid = this.generate_UUID();
     }
-    let item = new Task(task, taskid);
+    let item = new Task(title, task, taskid, tasktype);
     if (task != null && typeof task.then === 'function') { // For Promise type task
       task.then(() => {
           this.finished.push(taskid);
```
adrysn: Current implementation garbage collects finished tasks for 10 seconds period. So, a task that are finished in just 1 second stays in the list for 9 further seconds. What if garbage collect immediately here when a promise resolved, not waiting for 10 seconds for the GC cycle? Might it be too CPU consuming?

inureyes: When many tasks are finished together, it brings too many gc processes together. So I remove gc from original implementation (it was there after `finished.push`)

Reduce to 5 sec? It could be reasonable.

inureyes: Added responsive garbage collection.

PR #558, opened 2020-06-03
feat: add service worker creation routine on compile time (#554)

### LINKED ISSUE #554: Re-enable service worker to reduce traffic load
The service worker was used in the initial implementation of the console, but there is a problem that the webpage cache cannot be refreshed afterwards. 

Let's apply service workers again to reduce overall traffic. The current console implementation is using the rollupjs-based obfuscated file name, so there will be no problems as before.

inureyes:
This PR resolves #554


inureyes:
@adrysn It cannot be enabled on localhost OR non-SSL environment. And, sw.js is generated via rollup compilation process therefore there is no sw.js on dev environment.

inureyes:
@adrysn Fixed to turn off on development environment.

PR #612, opened 2020-07-19
feat: support editing .Renviron file

### LINKED ISSUE #608: Add edit feature of .Renviron file in home directory
`.Renviron` contains default R environment variables. Therefore it will be much easier to use R with custom `.Renviron`, like `.bashrc` and `.zshrc`.

adrysn:
This PR resolves #608.

* Fixed broken `wc-codemirror` UI.
  - `wc-codemirror` has an issue from >=1.6.0, which breaks codemirror UI inside a custom element.
  - From 1.6.0, base codemirror CSS styles are moved from `wc-codemirror`'s own template to `document.body`. `lablup-codemirror` is  a custom element, so it does not automatically inherit styles under `document.body`, and there is no base codemirror CSS style which breaks its UI.
  - [https://github.com/vanillawc/wc-codemirror/compare/v1.5.3...v1.6.0##diff-9de5e984ff98ce83bed512dd2a5f2d21R140](https://github.com/vanillawc/wc-codemirror/compare/v1.5.3...v1.6.0#diff-9de5e984ff98ce83bed512dd2a5f2d21R140)
  - We workaround this issue by inserting codemirror's base style directly into `lablup-codemirror`.
* Add `.Renviron` to the list of editable rc file list.
  - User can write their own `.Renviron` in user's settings page and save it like other dotfiles such as `.bashrc`, and/or `.zshrc`.
  - R is not for every users, so I'll not add the default `.Renviron` from Agent-side.

Related to OP#613.


inureyes:
Button locations are weird so it may be repositioned.

image

They should be right-aligned and

image

These buttons should be aligned too. 

Also, two TypeScript compilation errors exist.

adrysn:
> Button locations are weird so it may be repositioned.
> 
> image
> 
> They should be right-aligned and
> 
> image
> 
> These buttons should be aligned too.
> 
> Also, two TypeScript compilation errors exist.

Fixed TS compilation errors and footer button alignment.

PR #620, opened 2020-07-22
feat: match SDK naming between nodejs and python SDK

YoonYeoHwan:
First, I matched the SDK name that perform the same task. I want to know if my code is right.


src/components/backend-ai-storage-list.ts
```diff
@@ -1506,7 +1506,7 @@ export default class BackendAiStorageList extends BackendAIPage {
   _downloadFile(e, archive = false) {
     let fn = e.target.getAttribute("filename");
     let path = this.explorer.breadcrumb.concat(fn).join("/");
-    let job = globalThis.backendaiclient.vfolder.request_download_token(path, this.explorer.id, archive);
+    let job = globalThis.backendaiclient.vfolder.request_download(path, this.explorer.id, archive);
```
inureyes: request_download and request_download_token has different implementation. request_download_token requests JWT token to provide direct download using tus.io. Please review the implementation more.

inureyes (approved):
Now it works as expected. It could generate more bugs, so it will be covered after merge.

Issue #572, opened 2020-06-11
Queue file deletion


inureyes:
* When trying to delete many files, UI just shows let's check twice dialog and nothing happened. (Actually it is deleting the files)

<img width="565" alt="image" src="https://user-images.githubusercontent.com/949915/84391148-89987100-ac33-11ea-992f-a833aa8f6931.png">

inureyes:
It should be queued / also, timeout is returned due to the fetch timeout.

inureyes:
@YoonYeoHwan

inureyes:
Will open if problem persists.

PR #633, opened 2020-07-28
Queue file deletion (#572)

YoonYeoHwan:
Now I just added timeout and catch.
But I can't test my code, because error doesn't occur in my environment.


inureyes:
We will address this issue when it persists.

PR #636, opened 2020-07-30
fix: Add missing package dependencies

achimnol:


achimnol:
This was due to stale local build setup, and running `make clean` to rebuild the setup from scratch has worked.

PR #643, opened 2020-08-03
show a snackbar message when user starts session with invalid session name (#638)

### LINKED ISSUE #638: Although user inputs invalid session name, session creating dialog get closed.
Session name is optional by default, but user can input arbitrary name.

In order to avoid misunderstood about session creating with invalid input, such as less than 4 characters or including whitespaces, It seems more natural to make the dialog remained still rather than closed.

To sum up, In my point of view about user experiences, session creating dialog can be closed when user clicks the close button on right-top or there's nothing to be corrected starting a session.

lizable:
This PR resolves #638.

Related: OP#682.


src/components/backend-ai-session-launcher.ts
```diff
@@ -789,6 +790,12 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
       this.notification.show();
       return;
     }
+    if (!isSessionNameValid) {
+      this.notification.text = "Session name only allows at least 4 characters without any whitespaces."
```
inureyes: Please add i18n resource here. :)

src/components/backend-ai-session-launcher.ts
```diff
@@ -789,6 +790,12 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
       this.notification.show();
       return;
     }
+    if (!isSessionNameValid) {
+      this.notification.text = _text("Session name only allows at least 4 characters without any whitespaces.");
```
inureyes: It should be `_text('session.launcher.SessionNameAllowCondition');`

Issue #628, opened 2020-07-24
Show mounted folder information on session list


inureyes:
* Show mounted folder information on session list

PR #644, opened 2020-08-03
Feature/show mounted folder information on session list #628

Sujin-Kim1:
- show mounted folder information on session list using mwc-menu.


src/components/backend-ai-session-list.ts
```diff
@@ -977,14 +985,58 @@ export default class BackendAiSessionList extends BackendAIPage {
...
+
   /**
    * Render session information - category, color, description, etc.
    *
    * @param {Element} root - the row details content DOM element
    * @param {Element} column - the column element that controls the state of the host element
    * @param {Object} rowData - the object with the properties related with the rendered item
    * */
-  sessionInfoRenderer(root, column?, rowData?) {
+  sessionInfoRenderer(root, column ?, rowData ?) {
```
adrysn: This is a minor style issue but is there any reason to add a space between parameter name and question mark(`?`) to indicate optional variable?

src/components/backend-ai-session-list.ts
```diff
@@ -977,14 +985,58 @@ export default class BackendAiSessionList extends BackendAIPage {
...
 
+  /**
+   * Create dropdown menu that shows mounted folder names.
+   * Added menu to document.body to show at the top.
+   *
+   * @param e {Event} - mouseenter the mount-button
+   * @param mounts {Array} - array of the mounted folders
+   * */
+  _createDropdown(e, mounts) {
+    let menuButton: HTMLElement = e.target;
```
adrysn: Please use `const` rather than `let` if the variable does not need to be updated in the variable's scope.

src/components/backend-ai-session-list.ts
```diff
@@ -977,14 +985,58 @@ export default class BackendAiSessionList extends BackendAIPage {
...
   }
 
+  /**
+   * Create dropdown menu that shows mounted folder names.
+   * Added menu to document.body to show at the top.
+   *
+   * @param e {Event} - mouseenter the mount-button
+   * @param mounts {Array} - array of the mounted folders
+   * */
+  _createDropdown(e, mounts) {
```
adrysn: I think the method's name is too broad. If we need to create another dropdown area for some other element in the future, the name will be confusing.  What about to change to `_createMountedFolderDropdown` or something like that?

src/components/backend-ai-session-list.ts
```diff
@@ -977,14 +985,58 @@ export default class BackendAiSessionList extends BackendAIPage {
...
+    let menuButton: HTMLElement = e.target;
+    let menu = document.createElement('mwc-menu') as any;
```
adrysn: Same here. Also note that it is possible to update `const`'s properties after declaration.

src/components/backend-ai-session-list.ts
```diff
@@ -977,14 +985,58 @@ export default class BackendAiSessionList extends BackendAIPage {
...
+      })
+    }
+    document.body.appendChild(menu);
+  }
+
+
+  /**
+   * Remove the dropdown menu when mouseleave the mount-button.
+   * */
+  _removeDropdown() {
```
adrysn: Please use a more specific name here as well.

src/components/backend-ai-session-list.ts
```diff
@@ -977,14 +985,58 @@ export default class BackendAiSessionList extends BackendAIPage {
...
+   * Added menu to document.body to show at the top.
+   *
+   * @param e {Event} - mouseenter the mount-button
+   * @param mounts {Array} - array of the mounted folders
+   * */
+  _createDropdown(e, mounts) {
+    let menuButton: HTMLElement = e.target;
+    let menu = document.createElement('mwc-menu') as any;
+    menu.anchor = menuButton;
+    menu.className = 'dropdown-menu'
```
adrysn: Missing semi-colon.

src/components/backend-ai-session-list.ts
agatha197: Webstorm's refactor added a space. I'll fix them.

src/components/backend-ai-session-list.ts
```diff
@@ -1006,14 +1006,14 @@ export default class BackendAiSessionList extends BackendAIPage {
     if (mounts.length > 1) {
       mounts.map((key, index) => {
         if (index > 0) {
-          let item = document.createElement('mwc-list-item');
-          item.innerHTML = key[0];
```
adrysn: Use of simpler name `item` here would be okay since `item`'s scope is limited to the `if` block at line 1008. But, I think it's okay.

PR #645, opened 2020-08-03
fix: change scaling group lookup routine to be robust (#640)

### LINKED ISSUE #640: In a session log dialog, error message is raised when user clicks refresh log button
Related: OP#675.

inureyes:
This PR resolves #640


inureyes:
It is resolved with another PR.

Issue #663, opened 2020-08-14
Wrong password condition error message on signup


inureyes:
A warning message appears even if you enter the password according to the conditions guided by the password. But in that state, click signup to sign up again.

<img width="494" alt="image" src="https://user-images.githubusercontent.com/949915/90199970-2a143880-de11-11ea-8a26-6c5596122c89.png">

lizable:
Related: OP#718

PR #664, opened 2020-08-14
Fix/signup field validation #663

lizable:
Now password checking is fix and also signup dialog shows specific warning message for each user input.
Detailed messages make users more easily to find out their input error.
Input error details will be covered by this PR.
(see the details below.)

- less than minimum password inputs
- password mismatch
- change input field status to be "required"
- email pattern check with regular expression.
- reset all input field after re-open signup dialog.


inureyes:
@adrysn please double-check password procedure during login.

src/components/backend-ai-signup.ts
```diff
@@ -199,47 +218,72 @@ export default class BackendAiSignup extends BackendAIPage {
...
-    if (password1 !== password2) {
-      this.notification.text = 'Password mismatch. Please check your password.';
-      this.notification.show();
+
+    // if any input is invalid, signup fails.
+    if (inputFieldsValidity.map((el: boolean) => {
+      if (!el) {
+        return !el;
+      }
+    })) {
```
adrysn: Isn't this `if` statement always `true`? The return of `map` is an `Array`, and an `Array` is always `true` even if it is empty. For example, the following code will log `foo`.

```javascript
if ([false]) console.log('foo');
```

src/components/backend-ai-signup.ts
```diff
@@ -199,47 +218,72 @@ export default class BackendAiSignup extends BackendAIPage {
...
       return;
     }
+
     const token = (this.shadowRoot.querySelector('#id_token') as HTMLInputElement).value;
     const user_email = (this.shadowRoot.querySelector('#id_user_email') as HTMLInputElement).value;
     const user_name = (this.shadowRoot.querySelector('#id_user_name') as HTMLInputElement).value;
+    const password = (this.shadowRoot.querySelector('#id_password1') as HTMLInputElement).value;
     this.notification.text = 'Processing...';
     this.notification.show();
     let body = {
```
adrysn: Use `const` since `body` is not re-assigned in this block.

src/components/backend-ai-signup.ts
```diff
@@ -199,47 +218,72 @@ export default class BackendAiSignup extends BackendAIPage {
     //this.shadowRoot.querySelector('#signup-button').setAttribute('disabled', 'true');
   }
 
+  _clearUserInput() {
+    this._toggleInputField(true);
+    const inputFields: Array<string> = ['#id_user_email', '#id_user_name', '#id_token', '#id_password1', '#id_password2'];
+    inputFields.map((el: string) => {
+      this.shadowRoot.querySelector(el).value = "";
```
adrysn, line 225: Minor issue, but please be consistent with the quotes. Since all other strings in the method uses a single quote(`'`), replace double quote(`"`) here.

src/components/backend-ai-signup.ts
```diff
@@ -199,47 +218,72 @@ export default class BackendAiSignup extends BackendAIPage {
...
+      this.shadowRoot.querySelector('#signup-button').setAttribute('disabled', 'true');
+    }
+  }
+
   _signup() {
+    const inputFields: Array<string> = ['#id_user_email', '#id_user_name', '#id_token', '#id_password1', '#id_password2'];
+    let inputFieldsValidity: Array<boolean> = inputFields.map((el: string) => {
+      this.shadowRoot.querySelector(el).reportValidity();
+      return this.shadowRoot.querySelector(el).checkValidity();
+    })
```
adrysn: Semi-colon missing.

src/components/backend-ai-signup.ts
```diff
@@ -264,31 +308,145 @@ export default class BackendAiSignup extends BackendAIPage {
     return err;
   }
 
+  _validateEmail() {
+    const emailInput = this.shadowRoot.querySelector('#id_user_email');
+    emailInput.validityTransform = (newValue, nativeValidity) => {
+      // console.dir(emailInput._validity);
+      if (!nativeValidity) {
+        let isValid: boolean = nativeValidity;
```
adrysn: Use `const`.

src/components/backend-ai-signup.ts
```diff
@@ -264,31 +308,145 @@ export default class BackendAiSignup extends BackendAIPage {
...
+    }
+  }
+
+  _validatePassword1() {
+    const passwordInput = this.shadowRoot.querySelector('#id_password1');
+    const password2Input = this.shadowRoot.querySelector('#id_password2');
+    password2Input.reportValidity();
+    passwordInput.validityTransform = (newValue, nativeValidity) => {
+      if (!nativeValidity) {
+        let isValid: boolean = nativeValidity;
```
adrysn: `const` as well.

src/components/backend-ai-signup.ts
```diff
@@ -264,31 +308,145 @@ export default class BackendAiSignup extends BackendAIPage {
...
+        }
+      }
+    }
+  }
+
+  _validatePassword2() {
+    const password2Input = this.shadowRoot.querySelector('#id_password2');
+    password2Input.validityTransform = (newValue, nativeValidity) => {
+      if (!nativeValidity) {
+        let isValid: boolean = nativeValidity;
```
adrysn: `const` as well.

src/components/backend-ai-signup.ts
```diff
@@ -264,31 +308,145 @@ export default class BackendAiSignup extends BackendAIPage {
...
-                       label="${_t("signup.E-mail")}" value="${this.user_email}"
-                       @change="${() => this._clear_info()}"></mwc-textfield>
+                       label="${_t("signup.E-mail")}" validateOnInitialRender
+                       pattern="^[^\s@]+@[^\s@]+\.[^\s@]+$" @change="${this._validateEmail}" 
+                       validationMessage="${_t("signup.EmailInputRequired")}"
+                       value="${this.user_email}" required></mwc-textfield>
           <mwc-textfield type="text" name="user_name" id="id_user_name" maxlength="30"
-                       label="${_t("signup.UserName")}" value="${this.user_name}"></mwc-textfield>
+                       label="${_t("signup.UserName")}" value="${this.user_name}"
+                       validationMessage="${_t("signup.UserNameInputRequired")}" required></mwc-textfield>
```
adrysn: In the server-side, `username` is actually not a required field. If `username` is missing, manager will use `email` as a `username`.  But, if you think username should be required from the client-side, I have no further objection.

adrysn:
I also have noticed that there are some TypeScript compilation errors. There must be some cases where those lines do not return concrete values.
image

src/components/backend-ai-signup.ts
lizable: I'm all for that plan. I'll change `username` field to option from required.

Issue #651, opened 2020-08-04
Migrate compute_session from v4 to v5


inureyes:
Current session list implementation is based on v4 legacy support. Let's change it to use native v5 query.

PR #667, opened 2020-08-18
feat: Migrate compute session from v4 to v5 (#651)

Sujin-Kim1:
- implement session list based on v5 support.
- update `_exportToCSV()` in the session list to export only the desired fields.


adrysn (approved):
I fixed/refactored some codes.

PR #673, opened 2020-08-24
fix: resource preset validation

### LINKED ISSUE #672: Fix resource preset dialog validation
Let's fix those minor input validation checks.
- prevent resource-preset creation from invalid input. (e.g. name, resource_slot values, etc.)
- make input available in according to minimum steps in each resources (e.g. shared memory can be assigned in increments of 0.01)

lizable:
This PR Resolves #672.

### DIFF
--- a/src/components/backend-ai-resource-preset-list.ts
+++ b/src/components/backend-ai-resource-preset-list.ts
@@ -331,7 +331,7 @@ class BackendAiResourcePresetList extends BackendAIPage {
           </wl-button>
         </div>
       </backend-ai-dialog>
-      <backend-ai-dialog id="delete-resource-preset-dialog" fixed backdrop blqockscrolling>
+      <backend-ai-dialog id="delete-resource-preset-dialog" fixed backdrop blockscrolling>
          <span slot="title">${_t("dialog.title.LetsDouble-Check")}</span>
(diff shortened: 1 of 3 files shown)


src/components/backend-ai-resource-preset-list.ts
```diff
@@ -325,7 +331,7 @@ class BackendAiResourcePresetList extends BackendAIPage {
           </wl-button>
         </div>
       </backend-ai-dialog>
-      <backend-ai-dialog id="delete-resource-preset-dialog" fixed backdrop blockscrolling>
+      <backend-ai-dialog id="delete-resource-preset-dialog" fixed backdrop blqockscrolling>
```
adrysn: Is attribute `blqockscrolling` exists, or just typo?

adrysn (changes requested):
I cannot create/update a resource preset when the value of fGPU is a decimal. For example, when you input `0.2` to fGPU, you cannot update it.

image

src/components/backend-ai-resource-preset-list.ts
inureyes: @adrysn  It is definitely typo. haha

inureyes (approved):
LGTM

PR #674, opened 2020-08-25
feat: options for launching app (open app to public / preferred port)

adrysn:
This PR adds two options in launching a service app.
* Open app to the public
  - Exposes services like Terminal and Jupyter to the public.
  - Anyone who knows the app address can access the service by a browser.
  - This button is not visible in an Electron app.
* Try preferred port
  - Try preferred port when launching an app.
  - If the port is already occupied and in use, assign a random port silently.

image

The following issues will be dealt with in separate issues.
* Specify authorized users and/or hosts who can access a laucned app. -> what criteria to let clients access?
* Reuse port if (user, app) is the same. Currently, when a user launches the same app twice, they occupy different ports, respectively. It seems like a waste of port resources.

For this option to properly work, [approxy>=20.08.0](https://github.com/lablup/backend.ai-proxy-coordinator/releases/tag/20.08.0) should be used together.


adrysn:
Port number can already be handled in the local wsproxy. For example, SSH/sFTP can pass a preferred port number to the local proxy.

For the "Open app to public" option, I don't want to expose it in the App mode, since the service is launched in the localhost. So, I hide the option in the App mode.

src/components/backend-ai-app-launcher.ts
```diff
@@ -372,13 +377,19 @@ export default class BackendAiAppLauncher extends BackendAIPage {
...
+      let port;
       if (globalThis.isElectron && appName === 'sshd') {
         port = globalThis.backendaioptions.get('custom_ssh_port', 0);
         if (port === '0' || port === 0) { // setting store does not accept null.
           port = null;
         }
       }
+      const usePreferredPort = this.shadowRoot.querySelector('#chk-preferred-port').checked;
+      const userPort = parseInt(this.shadowRoot.querySelector('#app-port').value);
+      console.log(usePreferredPort, userPort)
```
inureyes: @adrysn please remove console log here 👍

inureyes (changes requested):
please remove console.log. Other things are LGTM.

src/components/backend-ai-app-launcher.ts
adrysn: Removed it!

inureyes (approved):
LGTM. I fixed some conflicts.

PR #683, opened 2020-09-01
fix: Incorrect image downloaded after filtered by keywords

### LINKED ISSUE #679: Incorrect image downloaded after filtered by keywords
**Describe the bug**
- After filtering table with user input (e.g. Tensorflow keyword at 'Language' column), selected record does not match with downloading image.

**How to reproduce:**
1. Login with Admin account in backend-ai-console
2. Go to Environments tab in left-drawer.
3. input any existing value that is in each column. (e.g: 'Language' - 'Tensorflow' or 'pytorch', etc.)
4. click left-side of table, and check out the Language of selected record and the Language of popup dialog.

**Expected behavior**
filtering table record and selected record have to be the **same record**, otherwise user can download wrong one.

**ScreenShots**
image

Jivvon:
This PR resolves #679 .

### DIFF
--- a/src/components/backend-ai-environment-list.ts
+++ b/src/components/backend-ai-environment-list.ts
@@ -248,10 +248,10 @@ export default class BackendAIEnvironmentList extends BackendAIPage {
   /**
    * Open the selected image.
    *
-   * @param {object} index - Selected image's index object.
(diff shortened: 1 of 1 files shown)


adrysn (approved):
LGTM.

PR #688, opened 2020-09-02
fix: replace session name with session id (uuid)

adrysn:
This PR resolves https://github.com/lablup/backend.ai-console/issues/687.

I think we don't need to change WSProxy for this to work.

Internal ticket: OP#773.


adrysn:
It seems like 502 error page is more frequently appears. I'll look into this PR more.

adrysn:
> It seems like 502 error page is more frequently appears. I'll look into this PR more.

Hmm. It is not reproduced any more...

lizable:
This happens to me quite often, too. 😭
I guess something goes wrong when executing many apps at the same time.

adrysn:
Fixed initial 502 error.

inureyes (approved):
LGTM.

PR #692, opened 2020-09-02
fix: display error message when kernel creation failed

adrysn:
This PR resolves https://github.com/lablup/backend.ai-console/issues/691 and https://github.com/lablup/backend.ai-console/issues/689.

### DIFF
--- a/src/components/backend-ai-error-log-list.ts
+++ b/src/components/backend-ai-error-log-list.ts
@@ -55,7 +55,7 @@ export default class BackendAiErrorLogList extends BackendAIPage {
   @property({type: Number}) _totalLogCount = 0;
 
   constructor() {
-      super();
+    super();
   }
 
   static get styles() {
@@ -154,8 +154,8 @@ export default class BackendAiErrorLogList extends BackendAIPage {
 
   /**
    * Update items from page target.
-   * 
-   * @param page 
+   *
+   * @param page
    */
   _updateItemsFromPage(page) {
     if (typeof page !== 'number') {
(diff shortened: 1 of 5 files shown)


lizable:
How about showing more detailed message when session creation failed?
Currently, It only shows 400 Bad Request - Missing or invalid API parameters.
In my opinion, It would be great to show what error occurred by the message.

adrysn:
After the latest commit, whenever I created a session, the following error message is displayed.
image

Also, would it be possible to limit max-width of the error message panel?
image

adrysn:
I still get error message when creating a session T_T.
image

Also, the session creation tasks normally were registered in the task panel (left side) right after the task is queued. You can check this by switching to the main branch. But recent commits in this branch removes the asynchronous queuing, and no session creation tasks were displayed in the side task queue panel. I think we should not destroy the existing feature.
image

adrysn:
Now, the error message is not displayed when kernel creation failed, which was the main purpose of this PR.

lizable:
I managed to fix no message appears when session creating failed.
However, in my point of view, this quick -fix in tasker component doesn't seem to be a proper error handling since it doesn't "handle" anything for now.

I tried throwing an error to caller and handle it in there using try-catch, but didn't work well.

It would be appreciated that If I could get any hint to fix this minor issue.

adrysn:
I am not sure the tasker element should handle exceptions of individual tasks. Because each task has different requests and different errors, I think the exceptions would be better handled in the callers themselves. So I think there's no problem the tasker does not handle anything. Just `finally` clause will do the job.

PR #704, opened 2020-09-10
feat: provide guide for text copy in terminal

### LINKED ISSUE #697: Provide explanation about maximum idle time
When I woke up, my open session was interrupted. If there is a maximum idle time, it will need an explanation.

## Possible implementation

* Show automatic turn-off time on session list item.
* Add maximum idle time on the last page of session launcher.

### LINKED ISSUE #698: Provide help about 'copying text from terminal'
* Provide help about 'copying text from terminal'
 * We support text copy, but it is not guided to user.

lizable:
This PR Resolves #698.
After This PR, new features will be available. Features are described below.

- when user clicks terminal app, web terminal guide with picture and some description will be provided through a dialog.
- If user don't want to see guide dialog after all, simply checking at the bottom of dialog, "Don't show this again", will hide the dialog forever.

Still minor issues left to be solved. Due to the limitation of **translation** in backend-ai-dialog, 
It was inevitable to add HTMLElement dynamically, which caused mismatching with current selected language until user refreshes the browser.
In my opinion, Although the minor issue mentioned above couldn't be solved, the main purpose of this issue is solved.


adrysn (changes requested):
LGTM in overall.

I think it would be better to abstract and create a dedicated custom element for this helper carousel to be reused in other pages and features, but let's postpone it later when there is another use case.

Some minor points to consider:
* It seems to be possible to drag images by mouse on the carousel, but its behavior is strange.
  - I cannot access the intermediate images (the second and the third) when dragging the images (on the `article` element) to the left/right. For example, if I drag the first image to the left (to see the second image), it jumps suddenly to the last image.
  - If it is a problem of `macro-carousel` element, is there an option to turn off mouse dragging at least?
  - Clicking the left/right button is working well.
* On the last `article`, what about adding a link to the online documentation section (as a new tab or window) for the further detailed explanation.

adrysn (approved):
LGTM.

Issue #686, opened 2020-09-01
Disable turn off button on finished session list


inureyes:
![imgpsh_mobile_save](https://user-images.githubusercontent.com/949915/91859890-eee58680-eca5-11ea-86c0-5b06988abe77.jpg)
 * Turn off button is shown on finished session list
 * It does not work so it does not need to be there.

PR #705, opened 2020-09-10
Remove API condition of turn off button #686

seeun-320:


src/components/backend-ai-session-list.ts
```diff
@@ -1118,7 +1118,7 @@ export default class BackendAiSessionList extends BackendAIPage {
             <wl-button fab flat inverted class="fg controls-running"
                                @click="${(e) => this._runTerminal(e)}"><wl-icon>keyboard_arrow_right</wl-icon></wl-button>
           ` : html``}
-          ${(this._isRunning && !this._isPreparing(rowData.item.status)) || this._APIMajorVersion > 4 ? html`
+          ${(this._isRunning && !this._isPreparing(rowData.item.status)) ? html`
```
adrysn: Good. We now do not need to support the legacy API (version < 5) in the main branch. But, the destroy button should be displayed for `ERROR` sessions to force-terminate them. So, please update the line like following:

```javascript
${(this._isRunning && !this._isPreparing(rowData.item.status)) || this._isError(rowData.item.status) ? html`
```

`_isError` function might look like:

```javascript
_isError(status) {
  return status === 'ERROR';
}
```

Good location for this function would be after `_isPreparing` function.

adrysn (changes requested):
Good to remove support for legacy API (<5). I want to display the destroy button for the `ERROR` session, so please update the PR. The destroy feature can also be triggered by the checkboxes in front of each session, so update `checkboxRenderer` as well in the same way.

inureyes (approved):
LGTM

PR #714, opened 2020-09-17
feat: add username updating feature ui

### LINKED ISSUE #693: Add user name registration / modification UI
* There is no UI for user name registration / modification
 * Add it by changing 'Change Password' to 'Change User Information' and adding name field on it.

image

lizable:
This PR resolves #693.
※ NOTICE: This feature is **only** available in **feature/add-username-update-api** branch of **backend.ai-manager repo**, until it merged to the master branch of backend.ai-manager.


resources/i18n/en.json
```diff
@@ -24,8 +26,12 @@
       "OriginalPassword": "Original password",
       "NewPassword": "New password",
       "NewPasswordAgain": "New password (again)",
+      "InputOriginalPassword": "Please Input Original Password",
       "InvalidPasswordMessage": "At least 1 alphabet, 1 number and 1 special character is required with at least 8 chars.",
+      "NewPasswordMismatch": "Two new passwords do not match.",
       "PasswordMismatch": "Password mismatch",
+      "PasswordUpdated": "Password is successfully updated.",
+      "UsernameUpdated": "User full name is successfully updated.",
```
adrysn: `username` and `full_name` are completely different and [distinguished in the DB-level](https://github.com/lablup/backend.ai-manager/blob/master/src/ai/backend/manager/models/user.py#L93-L98). So, I prefer to distinguish them in the Console as well.

src/lib/backend.ai-client-node.ts
```diff
@@ -617,6 +617,21 @@ class Client {
...
+   */
+  async update_username(oldFullname, newFullname) {
+    let body = {
+      'old_full_name': oldFullname,
+      'new_full_name': newFullname
+    };
+    let rqst = this.newSignedRequest('POST', `/auth/update-fullname`, body);
+    return this._wrapWithPromise(rqst);
+  }
+
```
adrysn: I think it would be better to use already-existing graphQL mutation to update user's information unless there is a good reason.

adrysn (changes requested):

src/components/backend-ai-console.ts
```diff
@@ -1116,25 +1260,43 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
...
+            <mwc-icon-button-toggle off onIcon="visibility" offIcon="visibility_off"
+                                      @click="${(e) => this._togglePasswordVisibility(e.target)}">
+              </mwc-icon-button-toggle>   
+          </div>
         </div>
         <div slot="footer" class="horizontal end-justified flex layout">
           <div class="flex"></div>
           <wl-button class="cancel" inverted flat @click="${this._hideUserPrefDialog}">${_t("console.menu.Cancel")}</wl-button>
-          <wl-button class="ok" @click="${this._updateUserPassword}">${_t("console.menu.Update")}</wl-button>
+          <wl-button class="ok" @click="${this._updateUserInformation}">${_t("console.menu.Update")}</wl-button>
```
inureyes: It should be changed to `$(()->this._updateUserInformation()}` while it is automatically run when page is loaded in some circumstances.

resources/i18n/ko.json
```diff
@@ -14,17 +14,23 @@
...
-      "NewPasswordAgain": "새 패스워드 (재입력)",
+      "OriginalPassword": "기존 비밀번호",
+      "NewPassword": "새 비밀번호",
+      "NewPasswordAgain": "새 비밀번호 (재입력)",
+      "InputOriginalPassword":"기존 비밀번호를 입력해주세요.",
       "InvalidPasswordMessage": "영문자, 숫자, 특수 문자를 하나 이상 포함해서 8자 이상 입력하세요",
+      "NewPasswordMismatch": "새로 입력한 두 비밀번호가 일치하지 않습니다.",
       "PasswordMismatch": "비밀번호가 일치하지 않습니다",
+      "PasswordUpdated": "비밀번호가 성공적으로 갱신되었습니다.",
+      "FullnameUpdated": "성명이 성공적으로 갱신되었습니다.",
```
inureyes: 성명 and 사용자 이름 are used. How about using 사용자 이름?

inureyes (changes requested):
Left some comments/

src/components/backend-ai-console.ts
```diff
@@ -329,6 +340,13 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
       });
   }
 
+  _initClient() {
```
inureyes: How about using `globalThis.backendaiclient` instead of spawning another instance?

inureyes (changes requested):
Let's use `globalThis.backendaiclient` instead of creating another one.

src/components/backend-ai-console.ts
lizable: I found out that 'client' property isn't used in this component. Therefore It should be removed.

inureyes:
@lizable I modified some codes (especially for API check) and I think the user preference dialog need to be independent component. I'll look up more.

inureyes:
Resolved conflicts.

inureyes (approved):
LGTM

PR #718, opened 2020-09-21
fix: Multiple clicks cause duplicated directory path

lizable:
I tried to find out the best way to ignore when the user clicks rapidly in the directory access icon(button) except using setTimeout, since setTimeout triggers unexpected results sometimes, and gives extra loads. Unfortunately, I couldn't get a decent solution to this. At least the main purpose of this PR is to avoid accessing non-existing sub-directory from the current directory path, and about that purpose, this PR works OK.

Still, there are some minor issues left.
Some click events are ignored when the user clicks the button, not often but it occurs.
If there is any idea to just fit in this issue, I would gladly replace from current commits.


lizable:
By @inureyes advice, I updated to using a promise instead of setTimeout so that the button can be clicked again only after the move current working directory command is completed.

inureyes (approved):
It works. 👍

PR #723, opened 2020-09-22
fix: extend current request timeout in session creation

### LINKED ISSUE #722: Extend current request timeout in session creation
The current timeout in session creation makes request timeout exception, although it's processing properly.
Let's extend the time to prevent timeout error.

lizable:
This PR resolves #722.


src/lib/backend.ai-client-node.ts
```diff
@@ -223,7 +223,7 @@ class Client {
     this._features = {}; // feature support list
     this.abortController = new AbortController();
     this.abortSignal = this.abortController.signal;
-    this.requestTimeout = 5000;
+    this.requestTimeout = 10000;
```
inureyes: You should never touch this request timeout. Instead, please change timeout for session creation routine.

inureyes (approved):
Now it will work.

PR #724, opened 2020-09-23
fix: new image cannot be downloaded

lizable:
After this PR, image downloading cannot be blocked by the undefined function call.


inureyes (approved):
LGTM

PR #727, opened 2020-09-24
fix: checkbox feature downsize in image downloading

lizable:
After merging this PR, there are several things will be changed in Images tab of 'Environments & Presets' page.
※ NOTICE: This feature is available in admin or super-admin account only.
You can see the detail below. (unchecked features are WIP.)

- [x] One or more selected image can be downloaded by clicking checkbox and clicking install button at right top.
- [ ] One or more selected image can be deleted by clicking checkbox and clicking delete button at right top.
- [x] If the user accidentally tries to install that is already checked, it will be filtered automatically, and at deleting part as vice versa.
- [x] The user can see image status by Installing / Installed tag next to the checkbox column.
- [x] All of the selected checkboxes will be deselected after the installation and deletion whether it fails or not.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -403,14 +403,19 @@
     "Base": "Base",
     "Constraint": "Constraint",
     "ResourceLimit": "Resource Limit",
-    "DescDownloadImage": "You are about to install the image: ",
+    "DescDeleteImage": "You are about to delete the image(s): ",
+    "DescDownloadImage": "You are about to install the image(s): ",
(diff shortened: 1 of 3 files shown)


adrysn (changes requested):
Delete image feature can only be implemented after the server-side API is developed. For now, just disable/hide the non-functional DELETE button. The full image installation/deletion feature will be dealt with in a separate issue.

For others, LGTM and works well as intended.

adrysn:
Also, I think we need to adapt linting. I'll create a separate issue for that.

adrysn (approved):

PR #733, opened 2020-10-16
feat: Add eslint to check typescript code

### LINKED ISSUE #728: Add eslint to check typescript code
We already have eslint in our package, but it seems it does not support TypeScript code. Let's add it by referencing, for example, https://khalilstemmler.com/blogs/typescript/eslint-for-typescript/ or other material.

lizable:
This PR resolves #728 .


adrysn (approved):
LGTM

PR #737, opened 2020-10-19
fix: make icons visible.

### LINKED ISSUE #736: icons in drawer and some panel don't show up in the new UI.
Current UI in main branch  doesn't show icon properly, It only shows squares instead of the icons.
Let's make it visible so that user can understand each menu more easily.
**Screenshots**
icon_error_in_drawer
icon_error_in_panel
icon_error_in_panel_2

lizable:
this PR resolves #736.


inureyes (approved):
LGTM

PR #746, opened 2020-10-21
Fix/minor errors in data storage page

### LINKED ISSUE #740: Minor errors occur in the creating folder dialog from Data & Storage page.
There are two minor bugs found in creating folder dialog from Data & Storage page, in the recent version of the main branch.

1. When user inputs folder name in the dialog, pattern attribute error occurs.
2. When user clicks 'create folder' button in the dialog, 500 internal server error occurs.
3. Drag and Drop operation doesn't work in the virtual folder.

Let's fix this in order to use virtual folder properly.

Internal ticket: OP#833.

lizable:
This PR resolves #740.

### DIFF
--- a/src/components/backend-ai-data-view.ts
+++ b/src/components/backend-ai-data-view.ts
@@ -273,7 +273,7 @@ export default class BackendAIData extends BackendAIPage {
       <backend-ai-dialog id="add-folder-dialog" fixed backdrop>
         <span slot="title">${_t("data.CreateANewStorageFolder")}</span>
         <div slot="content">
-          <mwc-textfield id="add-folder-name" label="${_t("data.Foldername")}" pattern="^[\.]?[a-zA-Z0-9_-]+$"
+          <mwc-textfield id="add-folder-name" label="${_t("data.Foldername")}" pattern="^[\.a-zA-Z0-9_-]+$"
               auto-validate required validationMessage="${_t("data.Allowslettersnumbersand-_dot")}"></mwc-textfield>
           <div class="horizontal layout">
             <mwc-multi-select id="add-folder-host" label="${_t("data.Host")}">
--- a/src/components/backend-ai-data-view.ts
+++ b/src/components/backend-ai-data-view.ts
@@ -273,7 +273,7 @@ export default class BackendAIData extends BackendAIPage {
       <backend-ai-dialog id="add-folder-dialog" fixed backdrop>
         <span slot="title">${_t("data.CreateANewStorageFolder")}</span>
         <div slot="content">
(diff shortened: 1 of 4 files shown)


src/components/backend-ai-data-view.ts
```diff
@@ -273,7 +273,7 @@ export default class BackendAIData extends BackendAIPage {
       <backend-ai-dialog id="add-folder-dialog" fixed backdrop>
         <span slot="title">${_t("data.CreateANewStorageFolder")}</span>
         <div slot="content">
-          <mwc-textfield id="add-folder-name" label="${_t("data.Foldername")}" pattern="[a-zA-Z0-9\_\-.]+"
+          <mwc-textfield id="add-folder-name" label="${_t("data.Foldername")}" pattern="^[\.]?[a-zA-Z0-9_-]+$"
```
adrysn: Do we need to prevent storage folder whose name containing dot(`.`) in the middle?

src/components/backend-ai-storage-list.ts
```diff
@@ -188,10 +188,10 @@ export default class BackendAiStorageList extends BackendAIPage {
         }
 
         #folder-explorer-dialog {
-          --component-height: calc(100vh - 170px);
+          --component-height: calc(100vh - 200px); /* calc(100vh - 170px); */
           right: 0;
           top: 0;
-          position: fixed;
+          position: fixed∫;
```
adrysn: Typo!

src/components/backend-ai-storage-list.ts
```diff
@@ -1339,8 +1339,12 @@ export default class BackendAiStorageList extends BackendAIPage {
       let temp: any = [];
       for (let i = 0; i < e.dataTransfer.files.length; i++) {
         const file = e.dataTransfer.files[i];
+        /* Drag & Drop file upload size limits to 1MiB */
         if (file.size > 2 ** 20) {
-          console.log('File size limit (< 1 MiB)');
+          this.notification.text = _text('data.explorer.DragDropFileUploadSizeLimit');
+          this.notification.show();
+          return;
+          // console.log('File size limit (< 1 MiB)');
```
adrysn: In my opinion, 1 MB is too small. How about increasing it to 10 MB or 50 MB, perhaps?

adrysn (commented):
A typo and some other opinions.

src/components/backend-ai-storage-list.ts
lizable: That's a good suggestion. I'll check with several files with different size, then choose maximum size.

src/components/backend-ai-data-view.ts
lizable: I guess we don't need it. I'll fix regex to make dot(`.`) can be put in the middle.

PR #747, opened 2020-10-21
fix: enable create user operation

### LINKED ISSUE #741: Cannot create user in the users page.
**Describe the bug**
When the user clicks create user button in the dialog, it throws an error.

**To Reproduce**
※ NOTICE:Only admin and super-admin account have an access to 'Users' page.
1. Go to backend.ai meta repository and download install-dev.sh in scripts directory
2. Execute install-dev.sh in the terminal and follow the instruction to execute both agent and manager in local after when installation is done.
3. Execute backend.ai console. (see the README.md of backend.ai-console repository)
4. Go to users page.
5. Click create user.
6. input values in the dialog.
7. click 'create user' button
8. Error occurs.(Error message: globalThis.backendaiclient.user.add is not a function)

**Expected behavior**
If the user clicks 'create user' button, it should be updated to the users list.

Internal ticket: OP#834.

lizable:
This PR resolves #741.


adrysn (changes requested):
In my environment, I faced the following problems:

* I was able to create a user account with a password that does not meet the validator's criteria (for example, 0000).
* In add credential panel, "User ID as E-mail (optional)" field is also present in the Advanced panel. Perhaps, either of the two should be removed?
* Cannot create keypair (credential) with the existing user email. When I tried to create a new keypair for a user, debug console prints error messages like below:
    image

lizable:
@adrysn Currently input access key and secret key for creating a keypair in the credential panel is deprecated, since those keys are generated automatically by the manager. (reference link: https://github.com/lablup/backend.ai-manager/blob/8ec58bed382c57a112cab224d840d049b20c6fd3/src/ai/backend/manager/models/keypair.py#L479-L485)
Therefore, Advanced panel is no longer necessary.

adrysn (approved):

PR #748, opened 2020-10-21
fix: minor errors in environments page

### LINKED ISSUE #742: Minor errors occur in Environments page.
There are several minor bugs found in every tabs of the Environments page in the recent version of the main branch.
※ NOTICE: Only admin and super-admin account have an access to the Environments page.

- Images tab
   - 400 Bad request error occurs when resources of the agent doesn't contain some resources that image requires (e.g. CUDA GPU)

- resource presets tab
   - In both modifying resource preset and creating resource preset dialog, they don't show any notification about invalid value of some resources.

- registries tab
  - If user creates the registry with the hostname that already exists in the list, it is not added, but updated.
  - TypeError occurs when registry is activated by the toggle button in the row.
  - notification in the drawer still remains after registries created/updated or deleted.

Let's fix this to make each feature of Environment page works.

Internal ticket: OP#835.

lizable:
This PR resolves #742.


adrysn (changes requested):
Some operations are not working in my environment.

No messages were displayed when I tried to create a registry with existing hostname. The dialog just stays open without any message. Also, Registry URL validation failed even if the value starts with `http://` correctly.
image

I cannot delete a registry with valid hostname:
image

inureyes (approved):
Found several flaws and fixed them.

inureyes:
@adrysn  please double-check even I found and modified some typo / code bugs.

lizable:
@inureyes I also fixed regular expression for check valid range of ip address(0 to 255) and port number(0 to 65536).

inureyes:
@lizable when I try to turn on repository, it generates error.

image

inureyes:
> @lizable when I try to turn on repository, it generates error.
> 
> image

@lizable nevermind. It is my setup's problem.

lizable:
@inureyes I guess it's already updated, It looks like the modify function is replaced to update function.
In this branch I couldn't find the error that you posted. You can see codes had been changed below.
(https://github.com/lablup/backend.ai-console/blob/33605605a28e2de1b025cb408082855c0922983d/src/components/backend-ai-registry-list.ts#L428)

PR #749, opened 2020-10-21
fix: resource group deletion reflected to table

### LINKED ISSUE #743: Resource group deletion doesn't get reflected to the table of resource groups tab in the Resources page.
※ NOTICE:Only admin and super-admin account have an access to the Resources page.

**Describe the bug**
When the user deletes the resource group, notification tells that it's succesfully deleted. But it is remained still in the table.

**To Reproduce**
1. login with admin or superadmin role.
2. go to Resources page in Administration.
3. go to the Resource Group page.
4. Add a resource group.
5. delete the resource group created in step 4.
6. see the table and error occurs.

**Expected behavior**
If resource group is successfully deleted, it should be deleted in the table, too.

Internal ticket: OP#836.

lizable:
This PR resolves #743.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -560,7 +560,12 @@
     "TypeResourceGroupNameToDelete": "Type resource group name to delete",
     "ResourceGroupName": "ResourceGroupName",
     "Active": "Active",
-    "SelectScheduler": "Select Scheduler"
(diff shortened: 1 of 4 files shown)


inureyes (approved):
LGTM.

adrysn (approved):
LGTM

Issue #772, opened 2020-11-03
Apply notification badge for each menu or panel when any data update occurs


lizable:
**Is your feature request related to a problem? Please describe.**
There has been a lot of progress since the release of Backend.AI.
As a result, the items updated in real-time have also increased.
When It comes to have multiple, simultaneous updates in GUI console,
It becomes hard to find out what's been changed, or how many things are changed.

Imagine the user tries to do two things at once.
- waiting for data sharing(vfolder invitations) by multiple vfolders owned by other users.
- creating multiple sessions with big amount of resources that have high chances to stuck in pending status instead of running.

In my opinion, simple notification badges could be a great answer to notify the user when any updates occurred.

**Describe the solution you'd like**
When an update occurs, mark small dot so called **"notification badge"** where the update occurred.
When user focuses or directs the place, hide the notification badge.

**Describe alternatives you've considered**
If the notification badge UI is against the design guideline of the Backend.AI,
then It can be replaced by notification in the menu-bar at the top area of the drawer.

PR #773, opened 2020-11-03
fix: disable app selection when session cancelled

### LINKED ISSUE #758: When session failed to start, App selection dialog still appears.
If session failed to start for some invalid vfolders mounting or old environment images(e.g. fortran) session status changes to prepared -> cancelled(failed to start), but It still shows app dialog with terminal. Let's show app dialog only if session successfully created.

lizable:
This PR resolves #758.


src/components/backend-ai-session-launcher.ts
```diff
@@ -917,7 +917,12 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
             appOptions['runtime'] = 'jupyter';
             appOptions['filename'] = this.importFilename;
           }
-          globalThis.appLauncher.showLauncher(appOptions);
+          // only launch app when it has valid service ports
+          if (service_info.length > 0) {
+            globalThis.appLauncher.showLauncher(appOptions);
+          } else {
+            throw { "message": _text('session.CreationFailed')}
+          }
```
adrysn: In creating a kernel, session-launcher poses a [timeout (10 s)](https://github.com/lablup/backend.ai-console/blob/main/src/components/backend-ai-session-launcher.ts#L961) to wait for the return data from the server. However, there are cases where a session is running after 10 s. For example, there may be a bootstrap scripts which runs longer than 10 s. In those cases, it is inappropriate to display session creation is failed. In my opinion, it would rather better not to display anything.

adrysn (changes requested):
I have comments on a minor message issue.

src/components/backend-ai-session-launcher.ts
lizable: Hmm, I thought that without notification, if the session was immediately removed from the running tab or moved to the exit tab, the user could be confused as to whether the session was successfully created or deleted.
However, I think this part could be solved in the other issue (#772), so I'll fix it as you suggested.

lizable (commented):
I removed throwing error messages to prevent from showing error even if session was created successfully with delay as @adrysn suggested.

src/components/backend-ai-storage-list.ts
```diff
@@ -1781,6 +1781,17 @@ export default class BackendAiStorageList extends BackendAIPage {
...
+   * Hide a wl dialog.
+   *
+   * @param {Event} e - Dispatches from the native input event each time the input changes.
+   */
+  _hideDialog(e) {
+    let hideButton = e.target;
+    let dialog = hideButton.closest('wl-dialog');
+    dialog.hide();
+  }
+
```
adrysn: Introduction `_hideDialog` prevents closing delete file dialog with Cancel button.

adrysn (changes requested):
Cannot close file delete dialog with Cancel button.

src/components/backend-ai-storage-list.ts
lizable: I just found out that _hideDialog is already implemented in the base class of BackendAiStorageList, which is BackendAIPage.
https://github.com/lablup/backend.ai-console/blob/0db0d2174f051193116e5fd3cc04c61f82f8c8ad/src/components/backend-ai-page.ts#L77-L85

Therefore, I'll remove `_hideDialog` function rightaway.

adrysn (approved):
LGTM.

PR #774, opened 2020-11-03
feat: apply loading indicator in login process

### LINKED ISSUE #760: Apply loading indicator in the login process.
Currently, the login process only shows the login panel after the user tries to access backend.AI.
When it comes to having high network latency due to various reasons,
It may confuse the user that the login process is on-going or not.
Let's show some loading indicator or at least disable user inputs in the login panel such as id and password
so that the user can find out the login process is proceeding.

lizable:
This PR resolves #760.


adrysn (approved):
LGTM.

PR #775, opened 2020-11-03
fix: show vfolder invitation message

### LINKED ISSUE #761: Vfolder invitation in the summary page shows up only after redirected to the summary page more than once.
**Describe the bug**
When the user added another user to access or edit to vfolder, it doesn't show up in the summary page invitation panel of another user at first, but appears in after accessing to summary page. 

**To Reproduce**
※ NOTICE: you need more than two account to test this. For example, you need account A and B.

1. Login with account A.
2. Go to Data & Storage page.
3. Add a folder.
4. click the third icon in the control panel of the folder created in 3.
5. dialog appears.
6. input account B email address in People textfield in the dialog.
7. click share button in the dialog.
8. Log out account A and Log in with account B.
9. go to summary page and see if invitation panel exists.
10. refresh the page or go to another page (such as Sessions or Import & Run page) and come back to the summary page.e status column.
11. invitation panel appears. 

**Expected behavior**
if the user got an invitation for vfolder, the invitation panel have to be revealed when the user logins.

lizable:
this PR resolves #761.
※NOTICE: the dashboard layout changed in this PR.


src/components/backend-ai-summary-view.ts
```diff
@@ -358,8 +357,8 @@ export default class BackendAISummary extends BackendAIPage {
       this.invitations = res.invitations;
       if (this.active && !refreshOnly) {
         setTimeout(() => {
-          this._refreshInvitations()
-        }, 15000);
+          this._refreshInvitations();
+        }, 1000);
```
adrysn: Folder invitation is not a frequent event. Do we need to refresh the invitation for 1 s?

adrysn (commented):
Minor comment on vfolder invitation refresh rate. Others are LGTM.

src/components/backend-ai-summary-view.ts
lizable: Well, as I mentioned in the other PR, I guess that It could be handled by #772.
Anyway, 1 sec is way too short, as you commented, so I'll change it.
Btw I presume that 15 seconds it quite long, would it be Okay to reduce the refreshing time such as 10 seconds or less?

adrysn: Okay, then let's go with 10 s for now.

inureyes (approved):
LGTM

PR #787, opened 2020-11-20
feat: max cores, max CUDA devices, and max shared mem can be configurable now

adrysn:
This PR resolves https://github.com/lablup/backend.ai-console/issues/788.

### DIFF
--- a/src/components/backend-ai-session-launcher.ts
+++ b/src/components/backend-ai-session-launcher.ts
@@ -1286,11 +1286,9 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
           if (cpu_metric.min >= cpu_metric.max) {
             if (cpu_metric.min > cpu_metric.max) {
               cpu_metric.min = cpu_metric.max;
-              cpu_metric.max = cpu_metric.max + 1;
               disableLaunch = true;
               this.shadowRoot.querySelector('#cpu-resource').disabled = true;
             } else { // min == max
-              cpu_metric.max = cpu_metric.max + 1;
               this.shadowRoot.querySelector('#cpu-resource').disabled = true;
             }
           }
--- a/src/components/backend-ai-session-launcher.ts
+++ b/src/components/backend-ai-session-launcher.ts
@@ -144,9 +144,9 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
   @property({type: String}) _helpDescription = '';
   @property({type: String}) _helpDescriptionTitle = '';
   @property({type: String}) _helpDescriptionIcon = '';
-  @property({type: Number}) max_cpu_core_per_session = 64;
-  @property({type: Number}) max_cuda_device_per_session = 16;
-  @property({type: Number}) max_shm_per_session = 2;
+  @property({type: Number}) max_cpu_core_per_session = globalThis.backendaiclient._config.maxCPUCoresPerSession;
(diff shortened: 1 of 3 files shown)


src/components/backend-ai-session-launcher.ts
```diff
@@ -144,9 +144,9 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
   @property({type: String}) _helpDescription = '';
   @property({type: String}) _helpDescriptionTitle = '';
   @property({type: String}) _helpDescriptionIcon = '';
-  @property({type: Number}) max_cpu_core_per_session = 64;
-  @property({type: Number}) max_cuda_device_per_session = 16;
-  @property({type: Number}) max_shm_per_session = 2;
+  @property({type: Number}) max_cpu_core_per_session = globalThis.backendaiclient._config.maxCPUCoresPerSession;
```
inureyes: Sometimes it is referenced before initialized from backend.ai-login. Need to be lazy-load. Tested on Firefox.

lizable (commented):
It seems working in Shared memory, but when it comes to CPU cores, maximum value doesn't work properly.
For example,  when I limit the maxCPUCoresPerSession to 1 by config value, It should be limited to one. But even if the slider turns out grayish which implies disabled, I can change the value.

adrysn:
That's because session launcher does not allow cpu metric to have the same value.
https://github.com/lablup/backend.ai-console/blob/main/src/components/backend-ai-session-launcher.ts#L1287

There might be a reason for this line, and will it be safe to delete this line? @inureyes 

Anyway, this will be very rare edge case since there will be no reason to limit maximum cpu cores per session to 2.

inureyes (commented):
> That's because session launcher does not allow cpu metric to have the same value.
> [https://github.com/lablup/backend.ai-console/blob/main/src/components/backend-ai-session-launcher.ts#L1287](https://github.com/lablup/backend.ai-console/blob/main/src/components/backend-ai-session-launcher.ts?rgh-link-date=2020-11-21T04%3A25%3A53Z#L1287)
> 
> There might be a reason for this line, and will it be safe to delete this line? @inureyes
> 
> Anyway, this will be very rare edge case since there will be no reason to limit maximum cpu cores per session to 1.



> That's because session launcher does not allow cpu metric to have the same value.
> [https://github.com/lablup/backend.ai-console/blob/main/src/components/backend-ai-session-launcher.ts#L1287](https://github.com/lablup/backend.ai-console/blob/main/src/components/backend-ai-session-launcher.ts?rgh-link-date=2020-11-21T04%3A25%3A53Z#L1287)
> 
> There might be a reason for this line, and will it be safe to delete this line? @inureyes
> 
> Anyway, this will be very rare edge case since there will be no reason to limit maximum cpu cores per session to 1.

With that code, slider will not work. (It requires different min / max value)

src/components/backend-ai-session-launcher.ts
adrysn: Yeap. Then, I will ignore the case where `maxCPUCoresPerSession` is `1`.

Anyway, for the property initialization, it is already updated by https://github.com/lablup/backend.ai-console/pull/787/commits/ea27541c25cef1982ff36079e49f693e3bca11b6.

inureyes:
@adrysn I fixed the slider issue. So it is ok to merge.

inureyes (approved):
LGTM

inureyes (approved):
LGTM

inureyes (approved):

PR #801, opened 2020-11-25
fix: allow keypair creation for users only

### LINKED ISSUE #798: Keypair can be created even though the user not exists.
> ※ NOTICE: Only admin and super-admin users have an access to the Users page.

**Describe the bug**
Credential key is for existing user only, but it can be created by arbitrary users, which doesn't exists in user list.

**To Reproduce**
문제를 재현하기 위한 순서를 자세히 적어주세요:
1. Go to the 'Users' page.
2. Click 'CREDENTIALS' tab in the page.
3. Click 'ADD CREDENTIAL' button in the right side of the panel.
4. 'Add Credential' dialog appears.
5. input any email address in the top of the input field in the dialog.
    (※ CAUTION: To see the error, please input user id which is not in the user list of 'USERS' tab.)
6. Click 'ADD' button in the dialog opened in step 4.
7. See the error.

**Expected behavior**
If the user id is not in the user list, credential creating operation should be treated as invalid with showing an error message.

**Client:**
- OS: macOS 10.15.6
- App version: 20.11.0
- NodeJS version: 13.8.0

**Server:**
- OS: macOS 10.15.6
- Python version: 3.8.3

lizable:
This PR resolves #798.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -318,6 +318,7 @@
     "UserAccountCreated": "User account is successfully created",
     "UserAccountCreatedError": "This account has already been registered.",
     "KeypairCreated": "Keypair successfully created.",
+    "UserNotFound": "Credential creation failed, since the following user ID could't be found : ",
     "UserIDRequired": "User ID is required."
   },
   "data": {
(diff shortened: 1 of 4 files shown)


inureyes (approved):
LGTM

PR #802, opened 2020-11-25
fix: session launcher reflects the current resource policy just after policy modification

### LINKED ISSUE #799: Updated resource policy assigned to user doesn't applied immediately in the session launcher.
> ※ NOTICE:
>    - Only admin and super-admin users have an access to the Users page.
>    - To Reproduce the error, you will need at least two user account.
         (One for admin or super-admin account and the other one for an user account.
          For convenience sake, an user will be called as user A.)

**Describe the bug**
Resource allocation panel in the session launcher dialog doesn't changed after admin or super-admin user changed the resource policy designated to the user.

**To Reproduce**
1. Login with an admin(or super-admin) account.
2. Go to 'Users' page and click 'RESOURCE POLICIES' tab.
3. Click 'CREATE POLICY' button at the right. 
4. 'Create Resource Policy' dialog appears.
5. Fill all of the mandatory input fields, uncheck every option in the resource policy and input 1 instead.
6. Click 'CREATE' button in the bottom of the dialog.
7. Click 'CREDENTIALS' tab.
8. click the gear button in the control column of the user A.
9. 'Modify keypair resource policy' dialog appears.
10. Select policy that created in step 4 to 6.
11. Click 'SAVE CHANGES' button in the right bottom of the dialog.
12. Logout.
13. Login with user A account. (※ CAUTION: the account must be different from the account used in step 1.)
14. Go to 'Summary' page and click 'START' button in the Start menu panel.
15. Scroll down to see the resource allocation slider and input field.
16. See the error.

**Expected behavior**
If resource policy assigned to user has been changed, then It should be applied not only the resource monitor, or resource policy column in the 'CREDENTIALS' tab in the 'Users' page, but also to the session launcher, too.

**Client:**
- OS: macOS 10.15.6
- App version: 20.11.0
- NodeJS version: 13.8.0

**Server:**
- OS: macOS 10.15.6
- Python version: 3.8.3

lizable:
This PR resolves #799.


lizable:
Resource monitor doesn't get updated just after the value in resource policy is updated or resource policy is changed itself to another one.
The only way to see the resource limits updated is redirect page from the sessions page to the summary page, or vice versa, for now.

In any cases, I guess the main purpose of bug-fix is solved.
Therefore if this PR is urgent, then maybe we could merge it first and fix the minor issue at a new thread.

lizable:
As a temporary measure, I updated the logic to refresh the resource stats panel every 10 seconds.

> Resource monitor doesn't get updated just after the value in resource policy is updated or resource policy is changed itself to another one.
> The only way to see the resource limits updated is redirect page from the sessions page to the summary page, or vice versa, for now.
> 
> In any cases, I guess the main purpose of bug-fix is solved.
> Therefore if this PR is urgent, then maybe we could merge it first and fix the minor issue at a new thread.

src/components/backend-ai-resource-monitor.ts
```diff
@@ -507,8 +506,18 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
...
 
-  async _refreshResourcePolicy() {
-    return this.resourceBroker._refreshResourcePolicy().then(() => {
+  /**
+   *  If bot refreshOnly and active are true, refresh resource monitor indicator
+   * 
+   * @param {boolean} refreshOnly 
+   * 
+   */
+  _refreshResourcePolicy(refreshOnly = false) {
```
inureyes: If you remove async here, `await` will not work on other references. See below.

src/components/backend-ai-session-launcher.ts
```diff
@@ -739,6 +739,8 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
         ownershipPanel.style.display = 'none';
       }
       this._updateSelectedScalingGroup();
+      /* To reflect current resource policy */
+      await this._refreshResourcePolicy();
```
inureyes, line 743: `this._refreshResourcePolicy()` should be async.

inureyes (changes requested):
_refreshResourcePolicy should be async because it usually calls with await.

src/components/backend-ai-resource-monitor.ts
lizable: Oops, My mistakes. I'll fix it right away.

src/components/backend-ai-resource-monitor.ts
```diff
@@ -704,6 +704,10 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
     }
   }
 
+  _calcConcurrencyMax() {
+    return this.concurrency_max === 1000000 ? `∞` : this.concurrency_max;
```
inureyes: If it is added to remove type miscast, it should be 

```
return this.concurrency_max === 1000000 ? `∞` : parseInt(this.concurrency_max);
```

inureyes (approved):
Fixed some issues now.

src/components/backend-ai-resource-monitor.ts
lizable: Got it! I'll keep this in mind. Thank you!

inureyes (approved):
LGTM

PR #818, opened 2020-12-01
feat: hide per week option when user account created less than a week.

### LINKED ISSUE #812: Let's hide per week option in the statistics for the users that assigned less than a week.
For those users who created account less than a week, there's no difference between per day and week graph in Statistics panel. 

Let's hide or disable per week selection if the user account has been created less than a week.

**Additional context**
Keypair is created just after the account is created, so may be keypair age will be helpful to decide whether hide or show the per week option.

lizable:
This PR resolves #812.


inureyes (approved):
LGTM

PR #823, opened 2020-12-01
fix: hide rename operation if the user is not the owner of Vfolder

### LINKED ISSUE #822: Let's allow vfolder renaming option only for the owner of the folder.
**Describe the bug**
After the owner of the Vfolder invited the other user with write-permission,
if the other user tries to change the Vfolder name itself, then it throws an error.

**To Reproduce**
> NOTICE: you will need two user accounts. one is for invitation, one is for receiving a vfolder with write permission.
                   For convenience sake, I will define the inviter as user A, and the invitee as user B.
1. Log in with User A account.
2. Go to 'Data & Storage' page.
3. Click 'NEW FOLDER' in the right top of the 'Folders' tab.
4. 'Create a new storage folder' dialog shows.
5. Input 'new-folder' and click 'CREATE' button in the bottom of the dialog.
6. Click  share_button button of the control panel.
7. 'Share Folder' pops up.
8. Input User B address in the e-mail address input field and check all of the option in the Permission part. (View and Edit)
9. Click 'SHARE' button in the bottom.
10. Logout.
11. Log in with User B account.
12. Go to 'Summary' page.
13. See the invitation panel and click 'ACCEPT' button.
14. Logout.
15. Log in with User A account.
16. Go to 'Data & Storage' page.
17. Click button image in the control panel of the folder created in step 5.
18. Select Edit + delete option in the permission column of the user B row and click 'SAVE CHANGES' button in the bottom of the dialog.
19. Logout.
20. Log in with User B account.
21. Go to 'Data & Storage' page.
22. Click rename_button button in the control panel of the folder invited in step 13.
23. 'Rename a folder' dialog appears.
24. Input with 'renamed-folder' in the dialog and click 'RENAME' button in the bottom of the dialog.
25. See the error.

**Expected behavior**
If renaming is available for those who acquired write-permission to the Vfolder owner, then it should be available.
But, if it's not allowed, then the button should rather be disabled or hidden than shown as if the operation is allowed.

**Client:**
- OS: macOS 10.15.6
- App version: 20.11.0
- NodeJS version: 13.8.0

**Server:**
- OS: macOS 10.15.6
- Python version: 3.8.3

**Additional context**
To provide more information, how about add an information panel for the maximum number of Vfolder that the user can create in the 'Data & Storage' page?
Currently, there's no visible notification for the maximum Vfolders that user can create,
unless the user gets error when he/she have reached the maximum count of created vfolder, and tries to create a new one.
Plus, I guess it would be great to have a single mark in each row if the user owns the Vfolder.

lizable:
This PR resolves #822.

### DIFF
--- a/src/components/backend-ai-storage-list.ts
+++ b/src/components/backend-ai-storage-list.ts
@@ -918,14 +918,16 @@ export default class BackendAiStorageList extends BackendAIPage {
             `
             : html``
           }
-
-          ${rowData.item.is_owner || this._hasPermission(rowData.item, 'd') || (rowData.item.type === 'group' && this.is_admin)
-            ? html`
+          ${rowData.item.is_owner ? 
+            html`
               <mwc-icon-button
                 class="fg blue controls-running"
                 icon="edit"
                 @click="${(e) => this._renameFolderDialog(e)}"
               ></mwc-icon-button>
+          ` : html``}
+          ${rowData.item.is_owner || this._hasPermission(rowData.item, 'd') || (rowData.item.type === 'group' && this.is_admin)
+            ? html`
               <mwc-icon-button
                 class="fg red controls-running"
                 icon="delete"


inureyes (approved):
LGTM

PR #831, opened 2020-12-04
feat: add option to change upload file size. (#830)

### LINKED ISSUE #830: Option to change upload file size
Currently, it is not configurable to limit the size of the uploaded file. Let's add a config item to set the policy on the uploadable size in a time.

Also, upload by drag & drop is limited to 1 GiB in size, but there is no limit when uploading by clicking Upload button. It also should be fixed.

inureyes:
This PR resolves #830


adrysn (approved):
LGTM.

PR #837, opened 2020-12-07
fix: show inner tab of users page properly

lizable:
> NOTICE: only admin or super-admin have an access to 'Users' page.

This is the hotfix for force activating inner-tabs(active and inactive) of 'Users' page.
After this PR, when the user clicks the tab with inner-tab in the 'Users' page will automatically activates inner 'active' tab of each tab.


inureyes (approved):
LGTM

PR #838, opened 2020-12-08
fix: remain SSH keypair after generation

lizable:
This is a hotfix for public SSH key initialized issue after refreshing the page.
After merging this PR, the public keypair will be remained whether there's page refreshing operation occurred or not, except for generating new SSH keypair by user.


inureyes (approved):
LGTM

adrysn (approved):
LGTM

PR #844, opened 2020-12-08
feat: show warning when reupload same file in vfolder

### LINKED ISSUE #836: Let's show a warning and confirmation dialog if the user tries to re-upload same file at same directory in Vfolder
When it comes to uploading a file that already exists, there should be an warning dialog to prevent un-expected file updates by user operation error.

If the user tries to upload same file and same name in a same directory, It should be applied after the user confirmation, since there's no turning back operation.

If user confirmation dialog is not available, adding an number could be another way to prevent from unexpected file update.

lizable:
This PR resolves #836.


adrysn (changes requested):
Please show warning when uploading by drag & drop for fairness.

inureyes (approved):
LGTM

PR #845, opened 2020-12-08
feat: limit max characters of input field

### LINKED ISSUE #832: Let's limit all the input field to prevent from malicious input
Almost every user input field should be given the maximum number of characters to avoid errors to the user.

If the user tries to enter more than the maximum character, an error message is output, but In my opinion,
this can be prevented without outputting an error message since it can be easily restricted within the input field.

If the maximum number of character cannot be limited, then at least It should show up error message near the input field
so that the user can recognize there's an invalid input in a certain field.

**Additional context**
For example, Session name, group name, domain name, user name, fullname of the user, vfolder name itself can have 64 characters in maximum.

lizable:
This PR resolves #832.


adrysn (approved):
LGTM.

PR #847, opened 2020-12-08
feat: apply unlimited when user input exceeds in setting resource policy

### LINKED ISSUE #834: When the value of input exceeds the maximum, let's automatically apply each check box to be checked (unlimited) in the Create/Modify Resource Policy dialog.
> NOTICE: only admin or super-admin have an access to 'Users' page and this feature.

User input in the resource policy dialog should be changed into unlimited at certain amount of value to prevent user from giving an belief that those inputs could be applied without any error.

If the user input is greater than the maximum value, it must be changed into "unlimited".

lizable:
This PR resolves #834.

### DIFF
--- a/src/components/backend-ai-credential-view.ts
+++ b/src/components/backend-ai-credential-view.ts
@@ -75,6 +75,17 @@ export default class BackendAICredentialView extends BackendAIPage {
   @property({type: Object}) exportToCsvDialog = Object();
   @property({type: String}) _defaultFileName = '';
   @property({type: Number}) selectAreaHeight;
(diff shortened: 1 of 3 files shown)


inureyes:
unlimited is literally 'unlimited', not bound to system limit.

lizable:
This PR is no longer necessary since the limitation for each input field is bounded to the max value by the commit below.
(refactor: add max values for number controls in resource policy - https://github.com/lablup/backend.ai-console/commit/8f97df89bcd5f9ca77bfbd31be1a731ee834b469)
Therefore This PR will be closed without merging in to main branch.

PR #848, opened 2020-12-08
feat: initialize vfolder invitation dialog to original layout 

### LINKED ISSUE #835: Let's initialize those adding / deleting input fields after applying or closing the dialog.
There's an ADD(+) / DELETE(-) button for email address in vfolder invitation dialog.
If the user continuously click add button and closes the dialog the input is dynamically added vertically.
The problem is that created input remains and it ruins the original layout of the dialog.

If the user closes or updates the dialog of vfolder invitation, it should go back to the original layout.

**Additional context**
vfolder_invitation

lizable:
This PR resolves #835

### DIFF
--- a/src/components/backend-ai-storage-list.ts
+++ b/src/components/backend-ai-storage-list.ts
@@ -640,7 +640,7 @@ export default class BackendAiStorageList extends BackendAIPage {
           </mwc-button>
         </div>
       </backend-ai-dialog>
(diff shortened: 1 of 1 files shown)


adrysn (commented):
I think it works. But, to prevent mis-close the dialog, for example by clicking the outside of the dialog, what about to add `persistent` attribute to the dialog?

inureyes (approved):
LGTM

PR #851, opened 2020-12-09
feat: logout by closing page

lizable:
~>NOTICE: In some browser, immediately closing tab and reopening action would not affect auto-logout.~

After merging this PR, user can enable/disable auto-logout feature by switch UI in usersettings page.
~Auto-logout feature will be enabled when the last page of Backend.AI GUI Console is closed and the user didn't opened the page in 5 seconds. If there are several pages are opened and one of the page is closed, the other pages will be affected, too. (Every request will be regarded as invalid, therefore Unauthorized access error will be pops up.)~

**※ Update for PR**
 - When auto-logout is enabled, there will be no maintaining login status after closing the last page of the Console
    (except session originated pages such as jupyter-notebook or web terminal).


inureyes:
We have to solve this:

`If there are several pages are opened and one of the page is closed, the other pages will be affected, too. `

src/components/backend-ai-console.ts
```diff
@@ -200,6 +202,51 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
         } else {
           this.loginPanel.login(false);  // Set showError flag to false for initial login
         }
+        
+        document.addEventListener('backend-ai-connected', (e) => {
+          if (globalThis.backendaioptions.get('auto_logout')) {
+              // prevent log-out from accidentally closed page (e.g. < 5sec.)
+              let currentTime = new Date().getTime();
```
inureyes: @lizable Please make redundant codes as method. (here and below.)

lizable:
@inureyes I fixed it, but unfortunately there was an side-effect occurred while solving this issue.
Maintaining login status was not possible when the user tries to restore the pages by tab restoring action(e.g. shortcut with keyboard).

Of course, I agree with that restoring session in a certain seconds is useful method to prevent unexpected logout from user's mistake.
However, if users enabled auto-logout by themselves, I presume that protecting their privacy would be more important rather than let login status being maintained.

Therefore, this side effect that I've mentioned could be acceptable.

> We have to solve this:
> 
> `If there are several pages are opened and one of the page is closed, the other pages will be affected, too. `

inureyes (approved):
LGTM

src/components/backend-ai-console.ts
lizable: Got it! I'll keep this in mind. :)

PR #852, opened 2020-12-09
fix: remove logic where is_amdin is not defined

adrysn:
Refreshing from admin pages always result in redirection to the summary page, even for admins.


lizable (approved):
LGTM.

PR #853, opened 2020-12-09
feat: apply update immediately in header text when current user information has been changed

### LINKED ISSUE #839: Let's make user information update operation applied right after completion
**Describe the bug**
Currently, There are two related issues in user information updating operation.
- [x] a.  In Backend.AI 20.03.11 cluster, user information updating operation doesn't seems to be working. (no responses.)
- [x] b. (Admin only) In Backend.AI 20.03.11 cluster and the latest version of the cluster, user info update operation on current logined user is applied only if the user refreshes the page.

**To Reproduce**
scenario of issue 'a'
1. Click on image button of the right top of the header.
2. dropdown menu appears
3. Click Change User Info menu in the dropdown.
4. 'Change User Information' dialog appears.
5. Input valid name in the Full Name input field. (NOTICE: input field value should be different value of current value.)
6. click 'UPDATE' button in the right-bottom of the dialog.
7. See the error.
scenario of issue 'b'
> NOTICE: Only super-admin have an access to 'Users' page.
1. Go to 'Users' page and click 'USERS' tab.
2. click gear button of the control panel of current user(logined user).
3. 'User Detail' dialog appears.
4. Input valid name in the Full Name input field. (NOTICE: input field value should be different value of current value.)
5. click 'SAVE CHANGES' button in the right-bottom of the dialog.
6. See the error.

**Expected behavior**
User information update should be applied immediately after update operation successfully completed.

lizable:
This PR resolves #839.

### DIFF
--- a/src/components/backend-ai-console.ts
+++ b/src/components/backend-ai-console.ts
@@ -214,6 +214,14 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
     globalThis.addEventListener("resize", (event) => {
       this._changeDrawerLayout(document.body.clientWidth, document.body.clientHeight);
     });
+
+    // apply update name when user info changed via users page
+    document.addEventListener('current-user-info-changed',  (e: any) => {
+      if (globalThis.backendaiclient.supports('change-user-name')) {
+        let input = e.detail;
+        this._updateFullname(input.full_name);
+      }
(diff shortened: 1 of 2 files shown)


inureyes (approved):
LGTM

PR #854, opened 2020-12-10
fix: update inputfield pattern and support translation in every notification

lizable:
After merging this PR, every notification will be translated to current language settings.
Also, there was a minor bug exporting csv menu in the users page (Admin Only), and It will be fixed via this PR.

### DIFF
--- a/src/components/backend-ai-console.ts
+++ b/src/components/backend-ai-console.ts
@@ -612,8 +612,6 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
    * Open the user preference dialog.
    */
   async _openUserPrefDialog() {
-    let fields = ['email', 'username', 'password', 'full_name'];
-    await globalThis.backendaiclient.user.get(this.user_id, fields);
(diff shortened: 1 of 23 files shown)


src/components/backend-ai-console.ts
```diff
@@ -612,6 +612,8 @@ export default class BackendAIConsole extends connect(store)(LitElement) {
    * Open the user preference dialog.
    */
   async _openUserPrefDialog() {
+    let fields = ['email', 'username', 'password', 'full_name'];
+    await globalThis.backendaiclient.user.get(this.user_id, fields);
```
inureyes: @lizable what is the purpose of these two lines?

inureyes (changes requested):
Remove mentioned lines if not needed. :)

src/components/backend-ai-console.ts
lizable: My mistakes. Probably didn't erased in the middle of conflict handling.
I removed it right away.

adrysn (approved):
LGTM.

PR #855, opened 2020-12-10
fix: put auto-logout panel into the group

lizable:
After merging this PR, auto-logout feature panel in usersettings will be rearranged automatically by screen width.


inureyes (approved):
LGTM

PR #859, opened 2020-12-14
fix: automatic logout when API keypair is not accessible

adrysn:
This PR resolves https://github.com/lablup/backend.ai-console/issues/860.

When API keypair is not accessible in `_connectViaGQL`, just automatically logout to invalidate existing session information.


lizable (approved):
LGTM

inureyes (commented):
LGTM

PR #861, opened 2020-12-14
feat: show detailed vfolder info

### LINKED ISSUE #858: Let's show current invited vfolders and created vfolders.
Currently, the maximum number of vfolders that can be created (not invited) follows the "resource policy".
Unfortunately, this limit doesn't show up until the user tries to create more vfolders than the maximum count.
And invited vfolder doesn't affects to vfolder count that can be created, but In current status,
the only way to find which one is invited is click every information icon in each row and see the owner in the dialog.
This might bring difficulty to the user when the user tries to find out which one is invited.

어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.
We should have UI for current status of the accessible vfolder of the user.

If It's not possible to create the UI (such as panel) for that, at least there should be a column to inform to the user that whether each folder is invited or not.

lizable:
This PR resolves #858.

**In detail:**
- add new panel with doughnut chart and numeric information for current storage status(e.g. created, invited, capacity)
- add new column for whether for each vfolder has owned by the user or not.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -387,6 +387,7 @@
       "Location": "Location",
       "Type": "Type",
       "Permission": "Permission",
+      "Owner": "Owner",
       "Control": "Control",
       "Name": "Name",
       "DeleteAFolder": "Delete a folder",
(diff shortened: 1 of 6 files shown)


inureyes (approved):
LGTM

PR #863, opened 2020-12-16
fix: adjust slider min max and support KO/EN translation in session launcher button

lizable:
After merging this PR,
There will be no mwc based console error occurred, and also session launcher button will be translated automatically with current language.


adrysn (approved):
LGTM.

PR #864, opened 2020-12-18
fix: notification error in safari and limit max-width of sidebar

lizable:
This PR contains two minor bug fix.
- notification error exposed in browser console in Safari (version: 14.0).
screenshot: 
promise_error

- limit max-width of sidebar and preserve logo ratio and size in the header of sidebar. (reported by @adrysn)


inureyes (approved):
LGTM

PR #867, opened 2020-12-18
fix: monkeypatch for min, max value force adjustment in lablup-slider

lizable:
After merging this PR, the value error will be handled, which triggered by the status that min value greater than max value
in lablup-slider component.


inureyes (approved):
LGTM

PR #871, opened 2020-12-23
fix: vfolder name validation pattern update

lizable:
This PR will allow '-' character in Vfolder creation and renaming.


adrysn (approved):
LGTM.

PR #875, opened 2020-12-28
fix: use type property to determine whether an entry is a directory or not

### LINKED ISSUE #874: Directory is displays as a normal file in vfolder explorer on vfolder host of purestorage type
When vfolder host of purestorage type is mounted, vfolder explorer displays every directories and symlinks as normal files. This prevents any directory traversing which is a critical bug.

adrysn:
This PR resolves #874.


inureyes (approved):
LGTM

PR #881, opened 2020-12-30
fix: max value of shared memory changes depending on current value of memory.

### LINKED ISSUE #880: Value and the maximum value of shared memory in session launcher does not updated by selected resource preset
**Describe the bug**
Two bugs are found shared memory slider and input-field in session launcher.
- shared memory doesn't get updated by selected resource preset.
- during the session creation, shared memory with 2 GB and memory to 10 GB makes shared memory regarded as larger value than memory as if each type of value is string, not number, eventually makes an error.

**To Reproduce**
1. Go to 'Sessions' menu.
2. Click on 'Start' button
3. Scroll down to 'resource-allocation' part
4. select any option in resource preset.
5. drag the slider of shared memory to each side.
6. select another option which was not selected in step 4.
6. Do the same action at step 5.
7. See error.

**Expected behavior**
Shared memory in session-launcher dialog should follow two things:
- The maximum value of shared memory must be equal or less than current memory
- Value comparison should be made as numbers, not strings.

**Screenshots**
MicrosoftTeams-image (7)

lizable:
This PR resolves #880.

### DIFF
--- a/src/components/backend-ai-session-launcher.ts
+++ b/src/components/backend-ai-session-launcher.ts
@@ -1888,10 +1888,11 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
     let shmemEl = this.shadowRoot.querySelector('#shmem-resource');
     let shmem_value = shmemEl.value;
     this.shmem_metric.max = parseFloat(this.shadowRoot.querySelector('#mem-resource').value);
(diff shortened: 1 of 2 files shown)


adrysn (approved):
LGTM.

PR #892, opened 2021-01-15
feat: show resource allocation explicitly

### LINKED ISSUE #891: Provide total resource allocation explicitly in session launcher
When supporting single-node multi-container or multi-node mode, the current session launcher confuses users since there is no information about whether the selected resource is allocated for each or separately.

어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.
Display resource allocation that makes the user understand how resources will be allocated in session creation.

**Additional context**
For example, when the user chooses the single-node mode with cluster-size as 3, resources will be allocated in each of 3 containers in 1 node. Unlikely single-node mode, multi-node mode with the same cluster-size can have multiple cases, such as:
 - each node will have one container.
 - one node will have more than one container.
 
Therefore, we cannot guarantee every node will have the same number of containers.

lizable:
This PR resolves #891.


inureyes:
@lizable 
 * Total allocation alone does not match the font size.
 * In addition, in the case of multi-container, the additional request should be fixed as 1 in the user setting resource allocation.
     * Suggesting scenario: cluster size exceeds 1 → The number of sessions seems to be 1 and disabled.
     * Then sess in total allication. The part can be omitted.

inureyes (changes requested):
@lizable 
 * Total allocation alone does not match the font size.
 * In addition, in the case of multi-container, the additional request should be fixed as 1 in the user setting resource allocation.
     * Suggesting scenario: cluster size exceeds 1 → The number of sessions seems to be 1 and disabled.
     * Then sess in total allication. The part can be omitted.

inureyes:
* Even when the number of options is more than 2, it seems that launcher show the container as x2.
   * Instead, if user upload a session, launcher panel need to increase the container number.
   * Node information panel should also be multi-node instead of x1 since cluster does not know where it will be spawned.

inureyes (changes requested):
* Even when the number of options is more than 2, it seems that launcher show the container as x2.
    * Instead, if user upload a session, launcher panel need to increase the container number.
    * Node information panel should also be multi-node instead of x1 since cluster does not know where it will be spawned.

inureyes (commented):
All areas that need to be modified have been modified. It would be nice to have a panel showing the total usage.

lizable:
> All areas that need to be modified have been modified. It would be nice to have a panel showing the total usage.

I guess this PR is related to issue #915, Applying carousel UI into session launcher.
Are those two issues going to be applied independently, @inureyes ?

inureyes:
> > All areas that need to be modified have been modified. It would be nice to have a panel showing the total usage.
> 
> I guess this PR is related to issue #915, Applying carousel UI into session launcher.
> Are those two issues going to be applied independently, @inureyes ?

Yes. I'm merging every PRs before merging https://github.com/lablup/backend.ai-webui/tree/feature/feature/eslint-adoption-940

inureyes:
@lizable  Since the ESlint branch is so huge, (https://github.com/lablup/backend.ai-webui/compare/feature/feature/eslint-adoption-940 ), I need to merge every branches / PRs before applying the refactoring branch.

inureyes (approved):
LGTM

PR #893, opened 2021-01-18
feat: add progress bar for flashblade volume

lizable:
This PR provides determinate progress bar UI for percentage of [FlashBlade](https://www.purestorage.com/products/file-and-object/flashblade.html) volume.


inureyes (approved):
LGTM

PR #900, opened 2021-02-01
feat: add progress bar support for image rescan (#898)

mickowale:
(#898)


src/lib/backend.ai-client-node.js
```diff
@@ -2428,6 +2428,11 @@ class Maintenance {
         this.client = client;
         this.urlPrefix = '/resource';
     }
+    attach_background_task(task_id) {
```
inureyes, line 2435: @mickowale We need a comment section above this method. (It is referenced by TypeScript to check the parameter type)

src/lib/backend.ai-client-node.ts
```diff
@@ -2633,6 +2633,12 @@ class Maintenance {
     this.urlPrefix = '/resource';
   }
 
+  attach_background_task(task_id: string) {
```
inureyes, line 2640: @mickowale We need a comment section above this method. (It is referenced by TypeScript to check the parameter type)

inureyes (changes requested):
@mickowale Please add comment to the methods. Other things looks fine (need more test / migrate into sidebar but it will be handled with another issue)

adrysn:
There is also a rescan images in backend-ai-registry-list.ts to update individual registries. I think it needs to be updated as well.

mickowale:
> There is also a rescan images in backend-ai-registry-list.ts to update individual registries. I think it needs to be updated as well.

I have updated this now and also merged with the main branch. I will squash the commits if the current state is okay.

inureyes (approved):
LGTM.

inureyes:
@mickowale Good! I'll squash and merge into main branch. :D

PR #905, opened 2021-02-09
fix: exception handling in misreferencing live stat field

lizable:
This PR resolves unexpected errors from missing fields from the server when receiving agent status by requests and fixes some typo error.


adrysn (approved):
LGTM.

PR #906, opened 2021-02-09
fix: close SSE(server-side event) after task finished

lizable:
This is a hotfix for #900.
It updates the type of events according to the common module and closes SSE tasks when they finish.
This will resolve recurring pushing events from the manager even after tasks are done.

There are all types of the background tasks changed in this fix.
task_updated -> [bgtask_updated](https://github.com/lablup/backend.ai-common/blob/master/src/ai/backend/common/events.py#L431-L432)
task_done -> [bgtask_done](https://github.com/lablup/backend.ai-common/blob/master/src/ai/backend/common/events.py#L476-L477)
task_cancelled -> [bgtask_cancelled](https://github.com/lablup/backend.ai-common/blob/master/src/ai/backend/common/events.py#L480-L481)
task_failed -> [bgtask_failed](https://github.com/lablup/backend.ai-common/blob/master/src/ai/backend/common/events.py#L484-L485)


mickowale:
@lizable  The fix should also be applied to rescan_images() in backend-ai-maintenance-view.ts.  It does stop the infinite background task events.

mickowale (approved):
LGTM

PR #911, opened 2021-02-16
Feature/support runtime env casting

### LINKED ISSUE #910: Adopt runtime environment parameter / argument feature

When starting session, environment parameters can be set to the instance for supporting images. Let's adopt runtime environment parameter / argument feature to GUI too.

Runtime environment parameter / argument feature. Wsproxy need to be patched too.

inureyes:
This PR resolves #910


inureyes:
TODO: environment parsing example. I'll add it to file browser on vfolder code.

lizable (approved):
LGTM

src/wsproxy/lib/backend.ai-ws-appproxy.js
```diff
@@ -54,21 +54,27 @@ module.exports = (proxy = class Proxy extends ai.backend.Client {
...
     this.port = port;
-    this._conn(sessionName, app);
+    this._conn(sessionName, app, envs, args);
     this.tcpServer.listen(this.port, this.ip);
   }
 
-  _conn(sessionName, app) {
+  _conn(sessionName, app, envs, args) {
+    if (typeof envs === 'object' && Object.keys(envs).length > 0) {
+      app = app + '&envs='+ JSON.stringify(envs);
```
adrysn: I think this should be url-encoded for safety. If url is encoded here, the receiver (`manager.js`) should also url-decode before parsing JSON string. For example, I'm not sure the following `envs` will be delivered properly as a querystring.

```javascript
'&envs='+JSON.stringify({myenv: 'ab&cd=e'})
> "&envs={"myenv":"ab&cd=e"}"
```

adrysn (changes requested):
I think simple `JSON.stringify(envs)` is not safe to be sin query string. Perhaps, we can wrap it with `encodeURI`.

adrysn (approved):

PR #914, opened 2021-02-16
feat: Add filter and sorting by each name of Vfolder

### LINKED ISSUE #909: Filter for storage folders
Need a function to filter storage folders as the number of the folders increases.
Filter could list up the folders according to name, ownership, date of production, occupied storage size, etc.

lizable:
This PR resolves #909.


inureyes (changes requested):
image

inureyes (changes requested):
image

'이름' layout need to be fixed.

inureyes (approved):
LGTM

PR #927, opened 2021-03-11
fix: invalid calculation logic for GPU utilization

adrysn:
Per-container GPU utilization is 0 or 100 due to multiplication by 100 to the percentage value.


inureyes (approved):
LGTM

PR #929, opened 2021-03-14
fix: unable to login when localStorage is full

### LINKED ISSUE #928: Unable to login when localStorage items exceeds size limit
When there are many items in `backendaiwebui.logs` and the total data size of `localStorage` exceedes browser's limit, user cannot login with the following error:

image

Also, deprecated `backendaiconsole.logs` still takes up quite a large amount of `localStorage` space since items in `localStorage` will not usually be deleted automatically. So, I think we need to insert a line of code to purge the data in `backendaiconsole.logs`.

adrysn:
This PR resolve #928.


inureyes:
We need to trim logs, not deleting whole logs. Maybe trimming logic already present, so we need to check the default log length to keep and reduce it too.

inureyes (approved):
LGTM

PR #931, opened 2021-03-15
fix: Add close button in env config

lizable:
After merging this PR, two things will be provided in modifying the environment dialog.
You can see the details below.
- closing button
- confirmation dialog for unsaved environment variables and values in modifying the environment dialog

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -191,11 +191,12 @@
       "NotEnoughResource": "Not enough resource",
       "EnvDeleted": "Envionment variable deleted",
       "SetEnv": "Set environment variable",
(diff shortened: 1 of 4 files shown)


adrysn (changes requested):
When adding multiple environment variables, the order of each field is somewhat strange. For example, when I click add (`+`) button after adding two items (`A` and `B`), the new fields are created in the middle of the first and the second item. This has nothing to do with the functionality, but needs to be fixed for natural UI.

image

lizable:
@adrysn
I truly agree with your comment and I applied the input field append to the bottom every time user clicks ``+`` button on the right side of the input field.
Also, I add a simple function for removing empty env input fields to provide a better UI for modifying environment variables and values. you can see the codes [here](https://github.com/lablup/backend.ai-webui/pull/931/commits/40f0c29b3cf93cf54d325558e3b93f20915ccb82#diff-ba6f5c1817ab368d2abef6c7543a522c8ae23002c3f0389fd20d897f25d31b3eR2119-R2130).

PR #934, opened 2021-03-19
feat: add .tmux.conf and .vimrc to editable dotfiles

### LINKED ISSUE #935: Let users to change .vimrc and .tmux.conf for better customizability
Some users request the feature to compose their own `.vimrc` and `.tmux.conf`.

adrysn:
This resolves #935.


lizable:
What do you think of exposing deleting ``rcfile`` feature @adrysn  and @inureyes ?
Yet, It's pretty useful when re-writing``rcfile``which is too long, but It affects the original ``rcfile``.
It would be great to have an additional file to allow users to overwrite them.

inureyes:
> What do you think of exposing deleting `rcfile` feature @adrysn and @inureyes ?
> Yet, It's pretty useful when re-writing`rcfile`which is too long, but It affects the original `rcfile`.
> It would be great to have an additional file to allow users to overwrite them.

It does not affect to original file since we use separate .local files.

inureyes (approved):
LGTM

lizable (commented):
I changed the logic of notifying the user when the user selects another rcfile without saving the current script.
It had three choices past.
- Cancel
- Discard
- Save and Close

But It seems to have too many operations in one confirmation dialog, so I decided to remove Discard and give users to have two choices, Cancel and Save.

adrysn (commented):
LGTM.

PR #936, opened 2021-03-21
fix: prevent "Infinity" max value to be converted as NaN in discrete GPU mode

### LINKED ISSUE #937: GPU slider in session launch dialog is always fixed as zero when discrete mode is enabled
How to reproduce:

* Activate `discrete` GPU mode from Etcd config server.
* Do not set any GPU resource limits in domain, group, and keypair-level.
* Connect to Web-UI and try to launch a session with GPU.
* GPU slider's maximum value is set to `NaN` and user cannot increase the GPU value to any integer values.

adrysn:
This PR resolves #937.


inureyes (commented):
LGTM

PR #943, opened 2021-03-24
fix: update previously selected rcfile to currently selected rcfile

lizable:
This PR is hotfix for #934 due to unexpected config file saving request on clicking "SAVE" button after clicking the close button (``x``)on the right top of the user config dialog.


adrysn (approved):
Good. I missed this!

Issue #680, opened 2020-09-01
Enable Tensorboard path


lizable:
Let's implement UI when user executes Tensorboard app.

inureyes:
@lizable I'll invest on this today. :)

inureyes:
OP#772

PR #952, opened 2021-04-08
fix: better port garbage collection for websocket proxy

inureyes:
This PR resolves bugs of #680, by adding app proxy garbage collector / port collector.


lizable (approved):
Why don't we provide information about tensorboard app will be restarted when the user set the path?
This suggestion can be achieved as related PR, so let's just merge this PR for now.

inureyes:
OP#772

PR #955, opened 2021-04-16
feat: share a group folder directly with overriding permission

### LINKED ISSUE #956: Share a group folder directly with overriding permission
A feature is requested to share or unshare a group (project) folder to users directly with overriding permission. This is to allow specified users (usually teachers with user account) can upload data/materials to a group folder while it is shared as read-only for other group users (usually students).

adrysn:
This PR resolves #956.

Requirement: lablup/backend.ai-manager#419 (This PR in manager must be merged first)


src/components/backend-ai-storage-list.ts
```diff
@@ -74,6 +74,7 @@ export default class BackendAiStorageList extends BackendAIPage {
   @property({type: String}) existingFile = '';
   @property({type: Array}) invitees = [];
   @property({type: String}) selectedFolder = '';
+  @property({type: String}) selectedType= '';
```
lizable: ```selectedType``` should be changed to ```selectedFolderType```.
I'll update it.

lizable (commented):
There are a few things that need to be clarified.

First, could we allow sharing group-folder to **the user in another group**?
Second, if the first assumption is true, then which one of the permission should be prioritized, the group or for individual permission? For now, If the group folder is created as READ ONLY permission, and we gave EDIT permission to the user from another group, the permission sticks to READ ONLY, not EDIT. This means that the permission update doesn't affect the users in another group, it only affects the users in its group. In my opinion, permission for each user should be prior to the permission of the group folder.

adrysn:
> There are a few things that need to be clarified.
> 
> First, could we allow sharing group-folder to **the user in another group**?
> 
> Second, if the first assumption is true, then how about adding the user in the group before sharing? For now, invitation seems valid but, the problem is that the invited user can't see the group-folder since the user is not in that group.

Updated Manager to raise an error when the shared user does not belong to a group where the virtual folder is included.

lizable (approved):
LGTM.

PR #959, opened 2021-04-26
fix: dropdown UI permission dialog

hydroxyde:
fixes permission dialog list when user close dialog without saving changes


lizable (changes requested):
@hydroxyde, I would like you to check If there are any linting issues.
You can simply check it by type down ``npm run lint`` at the root directory of WebUI repo.
Also you can fix minor linting issues automatically by this command, ``npm run lint-fix``.

src/components/backend-ai-storage-list.ts
```diff
@@ -962,27 +963,50 @@ export default class BackendAiStorageList extends BackendAIPage {
...
-            <mwc-list-item ?selected=${rowData.item.perm === 'ro'} value="ro">
-            <mwc-list-item ?selected=${rowData.item.perm === 'rw'} value="rw">${_t('data.folders.Edit')}</mwc-list-item>
-            <mwc-list-item ?selected=${rowData.item.perm === 'wd'} value="wd">${_t('data.folders.EditDelete')}</mwc-list-item>
-            <mwc-list-item value="kickout">${_t('data.folders.KickOut')}</mwc-list-item>
-          </mwc-select>-->
-        </div>
-      `, root
-    );
+    // Check if wl-select(permission list) exist
+    if (this.shadowRoot.querySelector('wl-select') != null) {
```
lizable: In my opinion, this conditional statement is not necessary.

lizable (changes requested):
Also, It seems working, but maybe we could remove some conditional statements.

src/components/backend-ai-storage-list.ts
hydroxyde: I can use render() at the start of this function and then have if statements checking user's modification. In this case render() is run twice. I think it is fine to leave wl-select existence check. Either way would work.
However, just removing this if statement would lead to an error because the function is started before wl-select has been generated.

hydroxyde:
lint fixes

lizable (approved):
LGTM.

src/components/backend-ai-storage-list.ts
```diff
@@ -962,27 +963,45 @@ export default class BackendAiStorageList extends BackendAIPage {
...
-        </div>
-      `, root
-    );
+    // Check if wl-select(permission list) exist
+    if (this.shadowRoot.querySelector('wl-select')) {
+      // user didnt select a new permission
+      if (this.shadowRoot.querySelector('wl-select').hasAttribute('pristine')) {
+        render(
+          html`
+            <div class="vertical layout">
```
inureyes: Is it duplicate of L980-L988?

inureyes (approved):
LGTM.

PR #973, opened 2021-04-28
feat: implement edu applauncher client API with token-based authorization

adrysn:
New component `edu-applauncher` is introduced. This component will provide a new endpoint:

```
/edu-applauncher?app=jupyter
```

When this URL is requested, the component will try to launch the requested app (delivered by `app` parameter) by the following logic:
* Try to login with cookie-based token (`/server/token-login`).
  - Token name can be configurable from Webserver and/or Manager. Anyway, they are just delivered as a cookie, so no explicit auth parameters need not be passed.
  - The default token name would be `sToken`.
* (Manager) Authenticate the login request and prepare the user.
  - Extract the token from delivered cookies.
  - Using the token, request to remote SSO server to check the validity of the token.
  - If the token is valid, the user's ID will be returned.
  - Create a new user with the ID (`<ID>@<some-email.com>`) as well as keypair and user virtual folder if not already exist.
  - Return success response.
* (Webserver) Using the response from Manager, save browser session in Redis.
* Now, Web-UI user is logged-in.
* Check if there are existing compute sessions by the user which can serve the requested app.
  - If not exists, create a new compute session by a pre-built session template from Manager DB (whose name should match with the requested app name).
* Using the compute session, open wsproxy connection and redirect user immediately when app is prepared.

### Related PRs
* https://github.com/lablup/backend.ai-webserver/pull/34
* https://github.com/lablup/backend.ai-integration-bok/pull/1
* https://github.com/lablup/backend.ai-manager/pull/420


lizable (approved):
LGTM.

PR #974, opened 2021-04-29
fix: temporary disable clone feature

lizable:
**⚠️  FYI: Since there's been a high chance of request blocking issues during cloning large-size virtual folder, cloning feature will be disabled for the time being.**

This PR disables the feature about cloning the virtual folder.


adrysn (approved):
LGTM.

PR #976, opened 2021-05-02
feat: display SCHEDULED session which is recently introduced

### LINKED ISSUE #977: Display SCHEDULED session status
Manager recently introduced `SCHEDULED` session status (https://github.com/lablup/backend.ai-manager/pull/415). We need to support this status as well.

adrysn:
This PR resolves #977.


inureyes (approved):
LGTM

PR #980, opened 2021-05-04
feat: add configuration button to set the num_retries_to_skip option.

Sujin-Kim1:
Add to change scheduler.
Clicking the config button on the settings page displays the scheduler change dialog, allowing you to change options such as num_retries_to_skip option.
Currently, only fifo is supported, so if a scheduler other than fifo is selected, it can only be set when num_retries_to_skip is 0.


src/components/backend-ai-settings-view.ts
```diff
@@ -50,16 +50,25 @@ export default class BackendAiSettingsView extends BackendAIPage {
   ];
   @property({type: Array}) jobschedulerType = [
     'fifo', 'lifo', 'drf'];
+  @property({type: String}) scheduler = 'fifo';
 
   constructor() {
     super();
+    try {
+      this.scheduler = localStorage.getItem('backendaiclient.settings.scheduler') || 'fifo';
```
lizable: How about using [backend-ai-settings-store](https://github.com/lablup/backend.ai-webui/blob/main/src/components/backend-ai-settings-store.ts) component instead of using localStorage directly?

src/components/backend-ai-settings-view.ts
```diff
@@ -320,6 +329,26 @@ export default class BackendAiSettingsView extends BackendAIPage {
                       </mwc-select>
                     </div>
                   </div>
+                  <div class="horizontal layout setting-item">
```
lizable: How about add an icon and show dialog instead of exposing two inputs with descriptions?
(e.g. SSH configuration in usersettings menu)

src/components/backend-ai-settings-view.ts
```diff
@@ -446,6 +484,8 @@ export default class BackendAiSettingsView extends BackendAIPage {
       globalThis.backendaiclient.setting.set(`plugins/scheduler/${scheduler}`, detail).then((response) => {
         this.notification.text = _text('settings.JobSchedulerUpdated');
         this.notification.show();
+        this.scheduler = scheduler;
+        localStorage.setItem('backendaiclient.settings.scheduler', this.scheduler);
```
lizable: I suggest you replace localStorage with backend-ai-settings-store components.

lizable (changes requested):
Since **default scheduler** doesn't affect the job scheduling algorithm itself, therefore we should change the title and description, too.

src/components/backend-ai-settings-view.ts
```diff
@@ -351,6 +358,55 @@ export default class BackendAiSettingsView extends BackendAIPage {
             </div>
           </div>
         </lablup-activity-panel>
+        <backend-ai-dialog id="env-dialog" fixed backdrop persistent>
+          <span slot="title" class="horizontal layout center">${_t('settings.SchedulerDefault')}</span>
+          <span slot="action">
+            <mwc-icon-button icon="info" @click="${(e) => this._showConfigDescription(e, 'default')}" style="pointer-events:auto;"></mwc-icon-button>
+          </span>
+          <div slot="content" class="vertical layout" style="width: 100%;">
+            <mwc-select id="scheduler-switch" required label="${_t('settings.SchedulerType')}"
```
lizable: If the user didn't select scheduler and click update button, "Successfully Updated" snack bar message pops out.
We need to show each input field that is required and If any input is empty, then we should show an error message right below the input and disable the update request.

src/components/backend-ai-settings-view.ts
```diff
@@ -301,23 +314,17 @@ export default class BackendAiSettingsView extends BackendAIPage {
...
-                                  label=""
-                                  style="width:130px;">
-                        ${this.jobschedulerType.map((item) => html`
-                          <mwc-list-item value="${item}"
-                                        ?selected=${this.options['scheduler'] === item}>
-                            ${item}
-                          </mwc-list-item>`)}
-                      </mwc-select>
+                    <div class="vertical center-justified layout">
+                      <mwc-button
```
lizable, line 336: How about adding some padding in the button? For now, It looks too densed.
스크린샷 2021-05-10 오후 1 28 08

src/components/backend-ai-settings-view.ts
```diff
@@ -351,6 +358,55 @@ export default class BackendAiSettingsView extends BackendAIPage {
...
+              style="margin-bottom: 10px;" @selected="${(e) => this.changeSelectedScheduleType(e)}">
+              ${this.jobschedulerType.map((item) => html`
+                <mwc-list-item value="${item}">
+                  ${item}
+                </mwc-list-item>`)}
+            </mwc-select>
+            <h4>${_t('settings.SchedulerOptions')}</h4>
+            <div class="horizontal center layout flex">
+              <span slot="title">${_t('settings.SessionCreationRetries')}</span>
+              <mwc-icon-button icon="info" @click="${(e) => this._showConfigDescription(e, 'retries')}" style="pointer-events:auto;"></mwc-icon-button>
```
lizable, line 396: Also, please center align info icon button.
스크린샷 2021-05-10 오후 1 28 24 복사본

src/components/backend-ai-settings-view.ts
```diff
@@ -432,28 +480,86 @@ export default class BackendAiSettingsView extends BackendAIPage {
...
+   * Update etcd scheduler options(num_retries_to_skip) and close the dialog.
+   */
+  saveAndCloseDialog() {
+    let tempNumRetries = this.shadowRoot.querySelector('#num-retries').value;
+    tempNumRetries = tempNumRetries === '' ? '0' : tempNumRetries;
+    if (['fifo', 'lifo', 'drf'].includes(this.selectedSchedulerType) && tempNumRetries.match(/^[0-9]*$/)) {
+      // currently, only support when scheduler type is fifo.
+      if (this.selectedSchedulerType === 'fifo' || (this.selectedSchedulerType !== 'fifo' && tempNumRetries === '0')) {
+        const numRetries = parseInt(tempNumRetries).toString();
+        globalThis.backendaiclient.setting.set(`plugins/scheduler/${this.selectedSchedulerType}/num_retries_to_skip`, numRetries)
```
lizable: How about generalizing features per each scheduler type?
In this code, the user can update `num_retries_to_skip` key and value only.
As we have a high chance of applying multiple keys and values, Let's make it more general,
such as sending as JSON below.
`{ num_retries_to_skip :  ... , new_feature: ...,}`

lizable (changes requested):

src/components/backend-ai-settings-view.ts
agatha197: Oops, my language configuration is Korean so I didn't know that. I'll add some padding.

src/components/backend-ai-settings-view.ts
```diff
@@ -433,27 +551,202 @@ export default class BackendAiSettingsView extends BackendAIPage {
...
+      textfield.value = clamp(textfield.value, textfield.min, textfield.max);
+    }
+  }
+
+  /**
+   * customize validation message.
+   * 
+   * @param {Event}  e - Dispatches from the native input event when input event occurs.
+   */
+  _customizeValidationMessage(e) {
```
lizable, line 717: @Sujin-Kim1
According to[ mwc-textfield documentation](https://github.com/material-components/material-components-web-components/tree/master/packages/textfield) and [ValidityState in native input field](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState), we can customize validation message per each invalid input.

Currently, there are two cases of invalid input.
The first case is empty input, so we can simply show the message such as 'This input is required'.
And the other case is that the input value exceeds[ the minimum or max range in the input field](https://github.com/lablup/backend.ai-webui/pull/980/files#diff-fc606a4cc01af6722e8fc5a686d3db2e7113862c172b2e7d80c5e27659e6503fR405-R406).
In this situation, we can show the message like this, 'please enter value between min value to max value.'

src/components/backend-ai-settings-view.ts
```diff
@@ -433,27 +551,202 @@ export default class BackendAiSettingsView extends BackendAIPage {
...
+  /**
+   * Update etcd scheduler options and close the dialog.
+   * 
+   */
+  saveAndCloseDialog() {
+    const scheduler = this.shadowRoot.querySelector('#scheduler-switch');
+    const numRetriesPerSchedulerEl = this.shadowRoot.querySelector('#num-retries');
+    const tempNumRetries = numRetriesPerSchedulerEl.value;
+    const inputValidationArr = [scheduler, numRetriesPerSchedulerEl];
+    if (inputValidationArr.filter(elem => elem.reportValidity()).length < inputValidationArr.length) {
```
lizable: @Sujin-Kim1 
Although we can use snackbar UI on the right bottom of the screen when showing input is invalid,
but If the component supports validation message right below the input field,
then we could rather use it instead of showing an error message with snackbar.
Since both mwc-select and mwc-textfield component supports validation message area,
I updated validity checking in both input fields(including select UI) before requesting configuration setting to the manager server.

src/components/backend-ai-settings-view.ts
```diff
@@ -351,6 +373,85 @@ export default class BackendAiSettingsView extends BackendAIPage {
...
+            <div class="horizontal center layout flex row">
+              <span slot="title">${_t('settings.SessionCreationRetries')}</span>
+              <mwc-icon-button icon="info" @click="${(e) => this._showConfigDescription(e, 'retries')}" style="pointer-events:auto;"></mwc-icon-button>
+              <mwc-textfield  id="num-retries"
+                              outlined
+                              required
+                              autoValidate
+                              validationMessage="${_t('settings.InputRequired')}"
+                              type="number"
+                              pattern="[0-9]+"
```
lizable, line 403: @Sujin-Kim1
As you've already seen there's[ a property called `type` in mwc-textfield component](https://github.com/material-components/material-components-web-components/tree/master/packages/textfield).
This also automatically ignores invalid input such as alphabetic character.

resources/i18n/en.json
```diff
@@ -855,10 +855,23 @@
...
     "DescROCMGPUsupport": "AMD ROCm GPU support. <br/>Requires Backend.AI ROCm Plugin.",
     "Scheduler": "Scheduler",
-    "JobScheduler": "Choose default job scheduler",
-    "JobSchedulerUpdated": "Default job scheduler has been updated.",
+    "SchedulerConfiguration": "Configuration per job scheduler.",
     "Require1912orAbove": "Requires Backend.AI 19.12 or above.",
     "Require2003orAbove": "Requires Backend.AI 20.03 or above.",
+    "Config": "Config",
+    "JobSchedulerUpdated": "Default job scheduler has been updated.",
+    "ConfigPerJobSchduler": "Configuration<br />per job scheduler",
```
lizable: Also, the title needs to imply what will this dialog exactly do.
In my opinion `default scheduler` is kind of a vague explanation, therefore I changed the title and description to `Configuration per job scheduler` for better explanation.

lizable (commented):
I just updated three things from the last commit.
- detailed title and description for the dialog
- customized input validation according to user input
- snackbar UI, which is displayed when input is invalid, has been replaced with validation message below in the input field

inureyes (approved):
LGTM

PR #1005, opened 2021-05-18
change the password icon

### LINKED ISSUE #993: the password masking on "Add registry" dialog is triangle. change to the circle one.
On the REGISTRIES tab of the environment page, when the ADD REGISTRY button is pressed, password masking appears as a triangle. I think it would be better to change to filed-circle for unity.

image

Sujin-Kim1:
the password masking on the "Add Registry" dialog is a triangle because the Roboto font does not support the circle icon. 
So, I added a font style to add-registry-password.

this PR closes #993


inureyes:
LGTM. However, general font family (e.g. San-serif) should be at the end of font-family setting. So please fix it too,

PR #1006, opened 2021-05-21
bugfix: custom resource allocation title overlaps

### LINKED ISSUE #982: custom resource allocation title overlaps
**Describe the bug**
When no resources are available at this time, users click anywhere on the slider to customize, resulting in overlapping titles and descriptions.

**To Reproduce**
1. Go to '/summary'
2. Click on the 'start' button to start a new sesssion.
3. Click on 'Custom allocation' and click somewhere inside any graph.

**Expected behavior**
display only when custom allocation is changed, and the title of the form must not overlap.

**Screenshots**
image

Sujin-Kim1:
To prevent the custom allocation text from being overwrapped, the custom assignment text was changed only when the resource is not disabled.

- The reason label and selectedText overlap is because selectedText is added to `mwc-list-item` without value and focusing.
  - So I just added dummy value (custom) and focus event.
- If the field is disabled, we do not need to change the resourceValue. So, I added the condition statement.
- Widen `resource-templates`.
- fix lint errors

this PR closes #982, #1017


inureyes:
There is no routine to set disableLanuch = false again. Since its scope becomes the element-wide, not instant variable, the flag should have reset code.

inureyes:
@Sujin-Kim1 firstUpdated() only called once. So it does not solve the reset code problem.

inureyes (changes requested):
@Sujin-Kim1 firstUpdated() only called once. So it does not solve the reset code problem.

Sujin-Kim1:
this PR is resolved by #1277

PR #1008, opened 2021-05-24
Feature/modify top bar style

### LINKED ISSUE #1004: Modify top bar style
1. It would be better to see the project name when the window size is narrow. Or put the project in the middle and margin on the right.
2. And align texts of list-item to be center.

**Screenshots**
image
image

Sujin-Kim1:
- Modify top bar styles. (make elements to be centered.)

- before
image
image

- after
image
  

this closes #1004


src/components/backend-ai-webui-styles.ts
```diff
@@ -551,12 +563,13 @@ export const BackendAIWebUIStyles = [
...
-
+      #project {
+        padding-top: 10px;
+      }
     }
 
     @media screen and (max-width: 1030px) {
-      span.email, #group-select-box > div > p
-       {
+      span.email, p#project {
```
lizable: In my opinion, we should better make project selecting UI visible unless the width is not too narrow.

lizable (commented):
@Sujin-Kim1 please check about the update I just committed.

lizable (approved):
LGTM.

PR #1061, opened 2021-07-13
feat: disable description icon in control panel of Storages

YejeeHa:
#1055 

Add condition to check if there is necessary data to show the information about the file system.
If not, the description icon in control panel of Storages tab in Resources page is disabled.


src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -24,6 +24,7 @@ import {BackendAiStyles} from './backend-ai-general-styles';
 import {IronFlex, IronFlexAlignment} from '../plastics/layout/iron-flex-layout-classes';
 import './backend-ai-dialog';
 import './lablup-progress-bar';
+import { booleanLiteral } from '@babel/types';
```
lizable: @YejeeHa 
Coud you explain the purpose of this code?

src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -426,12 +427,16 @@ export default class BackendAIStorageProxyList extends BackendAIPage {
    * @param {object} rowData
    */
   controlRenderer(root, column?, rowData?) {
+    let unmatched: boolean;
+    if (this.agentDetail.cpu_util_live == null || !('current_mem' in this.agentDetail)|| !('live_stat' in this.agentDetail && 'node' in this.agentDetail.live_stat)){
```
lizable: Also, please use a single `render()` function call, instead of using it in every control statement(`if`, `else`)

lizable (changes requested):
@YejeeHa 
Please see the comments and update the code if it's needed.

src/components/backend-ai-storage-proxy-list.ts
YejeeHa: I don't know why it was added cause I really didn't add that code line. I think it was added automatically when I clicked the automatic completing word 'booleanLiteral' by mistake. I will delete it. Thank you for letting me know.

src/components/backend-ai-storage-proxy-list.ts
YejeeHa: I changed the code with adding a separate method `unmatchedFileSystem()` to determine if it is unmatched or not so that I can erase the control statement in `controlRenderer(root, column?, rowData?)`.

src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -418,6 +418,18 @@ export default class BackendAIStorageProxyList extends BackendAIPage {
     return;
   }
 
+  /**
+     * Determine if the storage has unmatched file system.
+     *
+     * @return {boolean}
+     */
+  unmatchedFileSystem() {
+    if (this.agentDetail.cpu_util_live == null ) return true;
```
lizable: If there's VFS only, we cannot initialize property `agentDetail` since It initialized in a certain plugin. How about using some agent data received from the manager server?

lizable (changes requested):

lizable (approved):
LGTM.

src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -430,8 +430,9 @@ export default class BackendAIStorageProxyList extends BackendAIPage {
       // language=HTML
       html`
         <div id="controls" class="layout horizontal flex center" agent-id="${rowData.item.id}">
-          <mwc-icon-button class="fg blue controls-running" icon="assignment"
-                           @click="${(e) => this.showStorageProxyDetailDialog(rowData.item.id)}"></mwc-icon-button>
+          <mwc-icon-button class="fg blue controls-running" icon="assignment" 
+                          ?disabled=${Object.keys(JSON.parse(rowData.item.performance_metric)).length == 0 && JSON.parse(rowData.item.performance_metric).constructor == Object}
```
adrysn: If `JSON.parse(<string>)` is successful, its constructor must be an `Object`, otherwise, there will be `SyntaxError`. So, the second condition check after `&&` does not seem to be necessary. If `rowData.item.performance_metric` is not guaranteed to be a valid JSON string, you may write something like below:

```diff
    controlRenderer(root, column?, rowData?) {
+     let perfMetricDisabled;
+     try {
+       const perfMetric = JSON.parse(rowData.item.performance_metric);
+       perfMetricDisabled = Object.keys(perfMetric).length > 0 ? false : true;
+     } catch { // rowData.item.performance_metric is not a valid JSON string
+       perfMetricDisabled = true;
+     }
      render(
```

adrysn (changes requested):
There seems to be an unnecessary condition check.

src/components/backend-ai-storage-proxy-list.ts
YejeeHa: I haven't thought about the case that `JSON.parse(rowData.item.performance_metric)` causes an error. Thank you for letting me know what I missed. I change the code like you advised.

inureyes (approved):
LGTM

inureyes:
@YejeeHa LGTM. Please fix the conflict.

PR #1009, opened 2021-05-27
feat: divide backend-ai-session-launcher into several pages.

Sujin-Kim1:
- Relocate the dialog content of backend-ai-session-launcher.ts.
- And add previous, next buttons to move progress.

this closes #915

PR #1062, opened 2021-07-14
feature: show access and secret keys

### LINKED ISSUE #1058: Let's expand the range of Change User Info menu to show AK/SK of the logined account
> FYI: This feature was suggested by @achimnol. 

Backend.AI Client-py, which provides almost the same features as WebUI, shows AK/SK when the user requests.
Unlikely in WebUI, only accounts with Admin or Superadmin role can see their own AK/SK.
For better user experiences in both WebUI and Client SDK, we need to show AK/SK to not only for admin roles, also user roles, too.

어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.

- [x]  Update `Change User Info` to `My Account`.
- [x]  Show access key and secret key as Read-only field in User Info dialog.
        (⚠️ As access key and secret key is in pair and the relationship between account and those keypairs are 1:N,
              these keypairs should be displayed in a table as each row.)
- [x] Optional: Show the role information if necessary. (In Backend.AI we have four roles: user, monitor, admin, superadmin)

**Additional context**

User Info dialog shows up when the user clicks `Change User Info` menu in the right-top area at the header.
Screen Shot 2021-07-09 at 1 05 40 PM

Damdinbazar:
PR resolve #1058


Damdinbazar:
- [ ] Update `Change User Info` to `My Account`.
- [ ] Show access key and secret key as Read-only field in User Info dialog.
(⚠️ As access key and secret key is in pair and the relationship between account and those keypairs are 1:N,
these keypairs should be displayed in a table as each row.)
- [ ] Optional: Show the role information if necessary. (In Backend.AI we have four roles: user, monitor, admin, superadmin)

Damdinbazar:
resources/i18n/ko.json

Damdinbazar:
Values with the keys `MyAccountInformation` and `MyAccount` in json files (translations) in resources/i18n/ may need to be updated with new translations, as `ChangeUserInformation` and `ChangeUserInfo` are changed to `MyAccountInformation` and `MyAccount`, respectively.

src/components/backend-ai-webui.ts
```diff
@@ -1430,13 +1458,16 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
                           <mwc-list-item class="horizontal layout start center" disabled style="border-bottom:1px solid #ccc;">
                               ${this.user_id}
                           </mwc-list-item>
+                          <mwc-list-item class="horizontal layout start center" disabled style="border-bottom:1px solid #ccc;">
+                              Role: ${this.roleInfo.role}
```
agatha197: How about marking the user's role with `assignment_ind` mwc-icon instead of Role text?

ghost: Which one does look more suited?

With icon:
image
With text:
MicrosoftTeams-image (1)

ghost: In my opinion, this looks more consistent:
image

Damdinbazar:
Made the Access key and Secret key look uneditable:
image

Made the dropdown menu list look more consistent:
image

src/components/backend-ai-webui.ts
```diff
@@ -1237,6 +1244,27 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
     return name;
   }
 
+  _getAccessSecretKeys(accessKey) {
+    const fields = ['access_key', 'secret_key'];
+    return globalThis.backendaiclient.keypair.info(accessKey, fields);
+  }
+
+  async _showAccessSecretKeys() {
```
lizable: As I mentioned at the description of this issue, `user id`(email address) can have more than one keypair(credential). That is, the relation between `user id` and keypair(access key / secret key) is 1 : N. 
This is a screenshot for a better understanding.
multiple-keypair-in-a-single-user-id




Although default settings and mostly we provides `user id` and keypairs as 1:1, but we'd better show all of access key and secret key in the information panel.
> Hint: To see how user id has an relationship between multiple keypairs, please follow the guide below.
> 1. Check out the Users menu in the Administration section in the sidebar. 
> 2. Click `CREDENTIALS` tab in the page.
> 3. Click `ADD CREDENTIAL` button in the right side of the table.
> 4. Input the account of superadmin in the  User ID field and Click ADD button at the bottom of the dialog.
> 5. Check the node SDK function that requests and receives keypair information. (👉 [LINK](https://github.com/lablup/backend.ai-webui/blob/9df0af1cd719aba3275d61f5f493f8911ce8ac17/src/lib/backend.ai-client-node.ts#L2000-L2110))

lizable (changes requested):
@Damdinbazar about dropdown UI you implemented looks good to me.
For the information dialog, It would be great to use grid(table) UI instead of a read-only textfield. I commented on a detailed explanation about this.

src/components/backend-ai-webui.ts
ghost: As suggested, access keys and secret keys are listed in a grid table considering 1:N relationship of user id and keys.

image

lizable (changes requested):
I just found out that the width of the dialog is limited to support responsive view in both PC and mobile, therefore It's better to use Select UI for Access keys and read-only text field for Secret Key instead of grid UI. In grid UI, we cannot see keypairs even if the font size of each cell goes down to 8pt.
The main purpose of this feature is to show the keypairs that the account having.

Damdinbazar:
Access keys are show in select:
image
Secret key is show in MWC textfield with a different styling (font-size decreased from the default 14px to 10.5px) to fit the secret key so that it can be seen fully.
image

src/components/backend-ai-general-styles.ts
```diff
@@ -506,6 +506,21 @@ export const BackendAiStyles = [
       --mdc-typography-subtitle1-line-height: 16px;
     }
 
+    mwc-textfield#secretkey {
+      --mdc-text-field-fill-color: transparent;
```
lizable: `backend-ai-general-styles` component has been applied to almost every component in backendai-webui project. If you want to apply those styles setting id secretKey to `backend-ai-webui` component only, please add id-specific code into webui component instead of here.
For now, It can be applied into every component that imports and applies `backend-ai-general-styles`.

lizable: And also, please remind that **css** means "cascading styles sheet", so you don't need to apply the same style definition which already defined in `backend-ai-general-styles`.
It might be against programming rules of thumb, [DRY rule](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). 😉

src/components/backend-ai-webui-styles.ts
```diff
@@ -336,6 +336,10 @@ export const BackendAIWebUIStyles = [
       --mdc-select-disabled-dropdown-icon-color: #747474;
     }
 
+    mwc-select.fixed-position > mwc-list-item {
+      width: 255px;
```
lizable, line 340: If you enable `fixedMenuPosition` property in `mwc-select`, then you need to customize the width of menu item(`mwc-list-item`). I mentioned a similar case in [here](https://github.com/lablup/backend.ai-webui/pull/1009#discussion_r641374580).

src/components/backend-ai-webui-styles.ts
```diff
@@ -390,6 +394,10 @@ export const BackendAIWebUIStyles = [
       width: 100%;
     }
 
+    #user-preference-dialog mwc-textfield#secretkey {
```
lizable, line 397: In my opinion, we can increase the size of text based on the Root element's font size.
Therefore I updated it to `0.7rem`.

src/components/backend-ai-webui.ts
```diff
@@ -138,6 +138,9 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
   @property({type: Array}) superAdminOnlyPages = ['agent', 'settings', 'maintenance', 'information'];
   @property({type: Number}) timeoutSec = 5;
   @property({type: Boolean}) use_experiment = false;
+  @property({type: Object}) loggedAccount = Object();
```
lizable, line 141: I also changed the declaration of the property `loggedAccount` to Object instead of Object literal so that access key can be pre-selected.

src/components/backend-ai-webui.ts
```diff
@@ -251,6 +254,8 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
     this._changeDrawerLayout(document.body.clientWidth, document.body.clientHeight);
     globalThis.addEventListener('resize', (event) => {
       this._changeDrawerLayout(document.body.clientWidth, document.body.clientHeight);
+      // Tricks to close expansion if window size changes
+      document.body.dispatchEvent(new Event('click'));
```
lizable, line 258: This trick is also necessary when you enable `FixedMenuPosition` in a dialog, since the style of display is `fixed` therefore it might not able to follow when. width of the screen changes.

lizable (commented):
@Damdinbazar
I changed a few things, not many differences compared to the last commit.
And please resolve the conflicts between the main branch and the PR by merging it.

lizable (approved):
LGTM

inureyes (approved):
LGTM

PR #1078, opened 2021-07-20
fix: apply conditions about max memory per container in defining max value of memory resource slider

lizable:


Sujin-Kim1 (approved):
LGTM

inureyes (approved):
LGTM

PR #1080, opened 2021-07-21
feat: rename property and method

YejeeHa:
Rename properties and method to make sure that it is about storage-proxy, not about agents. 
I did not change the names related to notification which is translated to various languages (e.g. line 141 `this.storageProxyDetailDialog = this.shadowRoot.querySelector('#agent-detail');` ).

### DIFF
--- a/src/components/backend-ai-storage-proxy-list.ts
+++ b/src/components/backend-ai-storage-proxy-list.ts
@@ -24,7 +24,6 @@ import {BackendAiStyles} from './backend-ai-general-styles';
 import {IronFlex, IronFlexAlignment} from '../plastics/layout/iron-flex-layout-classes';
 import './backend-ai-dialog';
 import './lablup-progress-bar';
-import { booleanLiteral } from '@babel/types';
 
 /**
  Backend.AI Agent List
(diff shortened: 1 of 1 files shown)


lizable (commented):
I guess name-changes are fine, but we need to check one more since It's related to the enterprise plugin component.

src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -170,32 +170,32 @@ export default class BackendAIStorageProxyList extends BackendAIPage {
    *
    * @param {string} status - The agent's backend.ai client status.
```
lizable: Please update the description about param `status`.

src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -407,14 +407,14 @@ export default class BackendAIStorageProxyList extends BackendAIPage {
   /**
    * Show storage proxy detailed dialog.
    *
-   * @param {string} agentId - agent ID
+   * @param {string} storageProxyId - agent ID
```
lizable: Same here. `agent ID` -> `storage proxy ID`

src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -41,11 +41,11 @@ import './lablup-progress-bar';
 @customElement('backend-ai-storage-proxy-list')
 export default class BackendAIStorageProxyList extends BackendAIPage {
   @property({type: String}) condition = 'running';
```
lizable, line 43: Also, Backend.AI Agent List should be changed into `Backend.AI Storage List`.

lizable (changes requested):

lizable (approved):
LGTM.

Issue #1081, opened 2021-07-21
Attach virtual folder with alias name


inureyes:
**Is your feature request related to a problem? Please describe.**

Virtual folder aliasing is not currently implemented in the GUI. Let's implement aliasing function in node.js SDK and UI so that user can mount a virtual folder by renaming it when mounting.

**Describe the solution you'd like**

Attach virtual folder with alias name

<img width="397" alt="image" src="https://user-images.githubusercontent.com/949915/126481104-f263887c-68be-457a-bdaa-924e9b53c027.png">

 * [x] Implement node.js SDK
 * [x] Add UI for folder name aliases

**Describe alternatives you've considered**

**Additional context**

lizable:
@inureyes I guess the basic feature of aliasing virtual folder has been resolved via #1082.
Btw, @SonMinWoo is working on implelementing feature for aliasing not just for directory name but also on arbitrary path that written by user. (Issue: [backend.ai#329](https://github.com/lablup/backend.ai/issues/329) PR: #1194)
If this issue is related, I'll add a related comment, but if not, I think we should close this issue.

PR #1082, opened 2021-07-21
Attach virtual folder with alias name (#1081)

inureyes:
This PR resolves (#1081)


inureyes:
image

inureyes:
image

Taewan-P (changes requested):
There are two issues to solve:
  * If the alias name and the folder name are the same, it will not save the alias name and have the previous value.
  * After setting the alias name, deleting the alias name would not delete its value.

Taewan-P (changes requested):
Also, after creating a duplicate alias, deleting with the X mark keeps the folder mapping.

Taewan-P:
This is a suggestion related to UI.
drawing

How about putting the search box above the table? (Above image + some margins)
I also want to hear @Sujin-Kim1 's opinion.

Taewan-P (approved):
LGTM

PR #791, opened 2020-11-23
feat: support multi-node multi-gpu on gui

lizable:
This PR resolves #710.

PR #1086, opened 2021-07-27
fix: load vfolders according to current project

lizable:
related issue: OP#1431.


src/components/backend-ai-storage-list.ts
```diff
@@ -1646,9 +1646,9 @@ export default class BackendAiStorageList extends BackendAIPage {
   _deleteFolder(folderName) {
     const job = globalThis.backendaiclient.vfolder.delete(folderName);
     job.then((resp) => {
-      console.log(resp);
+      // console.log(resp);
       if (resp.msg) {
-        console.log(resp);
+        console.log(resp.msg);
```
inureyes: @lizable  please delete / comment this.

src/components/backend-ai-webui.ts
```diff
@@ -388,7 +388,7 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
       this.is_superadmin = false;
     }
     this._refreshUserInfoPanel();
-    this._writeRecentProjectGroup(this.current_group);
+    globalThis.backendaiclient.utils._writeRecentProjectGroup(this.current_group);
```
inureyes: @lizable you SHOULD NOT CHANGE THIS PART since it is general SDK, not WebUI-only. global.backendaioptions only exists in WebUI.

inureyes (changes requested):
Please read review and rollback some codes.

src/components/backend-ai-storage-list.ts
lizable: I'll delete it.

adrysn (changes requested):
I feature is working well. However, I found the following compilation error. I don't think this is raised by this PR, but it needs to be fixed.

image

lizable:
> I feature is working well. However, I found the following compilation error. I don't think this is raised by this PR, but it needs to be fixed.
> 
> image

It was found in `backend-ai-session-list` component, and It seems that It hasn't been used for a while, probably I just forgot to remove it when implementing #791.

adrysn (approved):
LGTM.

Issue #1077, opened 2021-07-19
Let's provide filter on status and user name.


lizable:
**Is your feature request related to a problem? Please describe.**
기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.

When it comes to having alot of resources and sessions created and terminated simultaneously, It's hard to find out whether the session I've been created is running properly. Therefore, It would be great to create filters in certain columns or maybe add a searching section to show specific sessions that match the query.

**Describe the solution you'd like**
어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.

For all user roles, they would need a filter for `status` of sessions. for admin roles, they would need one more filter: `user id`, since they can see sessions created from other users in each project.

**Describe alternatives you've considered**
혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.

**Additional context**
기능 요청에 대해 보다 잘 이해할 수 있는 다른 맥락을 기술해주세요.
There's been a code block to refer to filtering by column in [`backend-ai-storage-list`](https://github.com/lablup/backend.ai-webui/blob/52fa0eb2c22bcca2129ba230bd0b0f07bdfba16d/src/components/backend-ai-storage-list.ts#L464-L469) component.

lizable:
Resolved via #1089.

PR #1089, opened 2021-07-29
feat: Add status and user filter on session list

romanceformoon:
This PR started from #1077

Add status and user filter on session list.

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -1593,12 +1593,34 @@ export default class BackendAiSessionList extends BackendAIPage {
         ` : html``}
         <vaadin-grid-column width="40px" flex-grow="0" header="#" .renderer="${this._indexRenderer}"></vaadin-grid-column>
         ${this.is_admin ? html`
-          <vaadin-grid-sort-column resizable width="130px" header="${this._connectionMode === 'API' ? 'API Key' : 'User ID'}" flex-grow="0" path="access_key" .renderer="${this._boundUserInfoRenderer}">
-          </vaadin-grid-sort-column>
+          <vaadin-grid-column resizable width="270px" flex-grow="0" path="access_key" .renderer="${this._boundUserInfoRenderer}">
+            <template class="header">
+              <div class="horizontal layout center justified flex" style="margin-right:15px;">
+                <span class="title">${this._connectionMode === 'API' ? 'API Key' : 'User ID'}</span>
+                <vaadin-grid-sorter path="${this._connectionMode === 'API' ? 'access_key' : 'user_email'}" direction="asc" style="padding:0 10px;">
+                  <vaadin-grid-filter path="${this._connectionMode === 'API' ? 'access_key' : 'user_email'}" value="[[_filterInfo]]">
+                    <vaadin-text-field slot="filter" focus-target theme="small" value="{{_filterInfo::input}}">
+                    </vaadin-text-field>
+                  </vaadin-grid-filter>
+                </vaadin-grid-sorter>
+              </div>
+            </template>
+          </vaadin-grid-column>
         ` : html``}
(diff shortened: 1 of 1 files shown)


Sujin-Kim1 (approved):
LGTM

lizable (changes requested):
I have a few suggestions and requests.
1. How about adding filters on `session info` column in session list?
It would be great to support filtering session rows by session name, too.

2. Please merge the latest version of main branch into this branch.
For now, I get GraphQL generated error (error code: 400) when I switch pages from other menus to `Users` page.

src/components/backend-ai-session-list.ts
```diff
@@ -1593,12 +1593,45 @@ export default class BackendAiSessionList extends BackendAIPage {
         ` : html``}
         <vaadin-grid-column width="40px" flex-grow="0" header="#" .renderer="${this._indexRenderer}"></vaadin-grid-column>
         ${this.is_admin ? html`
-          <vaadin-grid-sort-column resizable width="130px" header="${this._connectionMode === 'API' ? 'API Key' : 'User ID'}" flex-grow="0" path="access_key" .renderer="${this._boundUserInfoRenderer}">
-          </vaadin-grid-sort-column>
+          <vaadin-grid-column resizable width="270px" flex-grow="0" path="access_key" .renderer="${this._boundUserInfoRenderer}">
```
lizable: Could we shorten the width of this column?
IMO, It seems too large. The average length of the text that values contain barely reaches the current length.
Screen Shot 2021-08-06 at 12 08 12 AM

Please check the similar column header on the Data & Storage page.

lizable (changes requested):

src/components/backend-ai-session-list.ts
lizable: Or, we could set the alignment of the column name and input field from horizontal to vertical.

romanceformoon: image
I've made column name and text field vertically because width of text-field increases automatically by length of contents on columns and i couldn't know how to adjust it.

lizable: @romanceformoon please check the API documentation of `vaadin-grid-filter-column`. Here's the [link](https://vaadin.com/components/vaadin-grid/html-api/elements/Vaadin.GridFilterColumnElement).

romanceformoon: My prior comment was about `vaadin-text-field` with using `vaadin-grid-column`, 
I could not find to set width of `vaadin-text-field`. 😢

lizable: @romanceformoon  As far as I know is that `vaadin-grid` component creates the textfield element internally when it renders.
And when it renders, the width of textfield is probably set by the width attribute of ` vaadin-grid-filter-column`.
So you should check whether setting width attribute (probably under 50px for significant changes) works or not.

lizable: Also, It would be appreciated that you could tell the approaches that you tried solving these kinds of issues, next time. 😉

lizable (approved):
LGTM.

src/components/backend-ai-session-list.ts
```diff
@@ -12,6 +12,7 @@ import '@vaadin/vaadin-grid/vaadin-grid-tree-toggle';
 import '@vaadin/vaadin-grid/vaadin-grid-selection-column';
 import '@vaadin/vaadin-grid/vaadin-grid-sorter';
 import '@vaadin/vaadin-grid/vaadin-grid-sort-column';
+import '@vaadin/vaadin-grid/vaadin-grid-filter-column'
```
inureyes: missing semicolon here.

inureyes (changes requested):
Please fix the typo. :)

inureyes (approved):
LGTM

PR #1090, opened 2021-07-29
feat: refresh image list after rescanning

YejeeHa:
#1065 
Add custom event `image-list-refreshed` where the rescanning image is done.

There is a problem when rescanning image in Maintenance page. 
- Click the 'rescan images' button and move to Environment page before rescanning is done ->  it seems refresh works well
- Wait until rescanning is done and go to Environment page -> only one item shows (screenshot below) and if I do something (e.g. check a checkbox, resize a column, etc.) the full list shows up.

 _**It only happens with Safari**. (Chrome and Firefox works fine.)_

스크린샷 2021-07-28 오후 6 49 45


src/components/backend-ai-environment-list.ts
```diff
@@ -1201,6 +1201,10 @@ export default class BackendAIEnvironmentList extends BackendAIPage {
       this._refreshSorter(e);
     });
 
+    document.addEventListener('image-list-refreshed', () => {
```
lizable: The event name `image-list-refreshed` seems that the image list has been changed, but inside of the event listener it's calling `_getImages()`, which requests the images from manager server.
In my opinion, we could change the name of custom event to have closer meaning of the event , such as `registry-rescanned`.

src/components/backend-ai-maintenance-view.ts
```diff
@@ -163,7 +163,7 @@ export default class BackendAiMaintenanceView extends BackendAIPage {
                   outlined
                   disabled
                   label="${_t('maintenance.CleanupImages')}"
-                  icon="delete"
+                  icon="delete">
```
lizable: Great! 👍

src/components/backend-ai-environment-list.ts
```diff
@@ -151,7 +151,7 @@ export default class BackendAIEnvironmentList extends BackendAIPage {
         span.resource-limit-title {
           font-size: 14px;
           font-family: var(--general-font-family);
-          font-align: left;
+          text-align: left;
           width: 70px;
```
lizable: Typo error catching, good! 👍

lizable (changes requested):
The issue mentioned in the description of this PR, seems not happening in my environment, so far.
We need to inspect more about how it happens, under certain circumstances.
Except that, renaming the custom event is the only request change, for now.

inureyes (approved):
LGTM

PR #1101, opened 2021-08-19
refactor: comment out number of file item in folder info dialog

lizable:
Related Ticket: OP#1453.


adrysn:
I changed my mind and let's take an official route, not a temporary one.

PR #1105, opened 2021-08-20
Feat/modifying registry configurations

### LINKED ISSUE #1102: Support for modifying registry configurations
We're only supporting registry addition and deletion.
If I want to modify the registry configurations, I have to delete the original one and create a new modified registry.
I think this is so uncomfortable for users.

I also want to support the ability to modify registry configuration.

**Additional context**
On webui environment page and click the registry button. 
Click the Modify icon in the Control column to bring up a dialog to modify the registry.

Sujin-Kim1:
this resolves #1102


inureyes:
@Sujin-Kim1  I tested via 

1. Change projects (removed one of 3 projects)
2. Rescan images
3. Nothing changed in image list (I also refresh the browser but no luck)

inureyes (changes requested):
I tested via

1. Change projects (removed one of 3 projects)
2. Rescan images
3. Nothing changed in image list
 
I also refresh the browser but no luck.

Sujin-Kim1:
The same result came from the existing modification method, creating a new one after deletion. (Clear the existing registry that had three projects and create a new registry with only two projects)
I also restarted the manager, but the image scan results were still the same. When I modified the registry on backend.ai-control-panel, the same result came out, wasn't it what you intended?
If I modify the project, do I have to change it so that only the project is scanned?

inureyes (approved):
LGTM

PR #1106, opened 2021-08-24
feat: remove empty mounted-folder-dropdown

Sujin-Kim1:
Show mounted folder dropdown menu only if the length of mounted folder is longer than 1.

Before:
image
After:
image
image


inureyes:
@Sujin-Kim1 It may better with number of folders on the folder icon overlay.

src/components/backend-ai-session-list.ts
```diff
@@ -1120,8 +1120,8 @@ export default class BackendAiSessionList extends BackendAIPage {
           menu.appendChild(mountedFolderItem);
         }
       });
+      document.body.appendChild(menu);
```
inureyes, line 1124: In this case, created menu object will be orphan object if it is not attached to body.

src/components/backend-ai-session-list.ts
```diff
@@ -1120,8 +1120,8 @@ export default class BackendAiSessionList extends BackendAIPage {
...
     }
```
inureyes, line 1125: How about add 'else' part and add some description about 'only one folder is attached' status?

inureyes (changes requested):

inureyes (approved):
LGTM

PR #1112, opened 2021-08-24
fix: Refactor utilizable functions

lizable:
This PR refactors webui components by assembling frequently used and duplicated functions into `backendai-common-utils`.
After merging this PR, I would like to suggest both contributors and maintainers regard `backendai-common-utils` as the "common" component of webui project, and using it.


src/components/backend-ai-storage-list.ts
```diff
@@ -2572,7 +2552,7 @@ export default class BackendAiStorageList extends BackendAIPage {
    * @return {string} UTC time string
    * */
   _humanReadableTime(d) {
-    const date = new Date(d * 1000);
```
Sujin-Kim1, line 2575: parameter `d` is unused.

src/components/backend-ai-storage-list.ts
```diff
@@ -2572,7 +2552,7 @@ export default class BackendAiStorageList extends BackendAIPage {
...
+    const date = new Date();
```
agatha197: parameter `d` is unused.

inureyes:
We will drop this branch since it makes components bound each other too strong. Before close, we need to revert flash-blade integration source code.

inureyes:
If some functions / methods are used frequently, please add it to backend-ai-page.

PR #1133, opened 2021-10-01
fix: safe fallback of API endpoint for education app launcher handler

Sujin-Kim1:
- Fix error caused by empty endpoint when connecting manager.
- Fix timeout error when creating session using a template by adding timeout parameter.

Related: [control-panel#120](https://github.com/lablup/backend.ai-control-panel/pull/120), [manager#480](https://github.com/lablup/backend.ai-manager/pull/480)

### DIFF
--- a/src/components/backend-ai-edu-applauncher.ts
+++ b/src/components/backend-ai-edu-applauncher.ts
@@ -99,6 +99,13 @@ export default class BackendAiEduApplauncher extends BackendAIPage {
     this.notification = globalThis.lablupNotification;
     const webUIShell: any = document.querySelector('#webui-shell');
     // webUIShell.appBody.style.visibility = 'visible';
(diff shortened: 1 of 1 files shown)


src/lib/backend.ai-client-node.js
```diff
@@ -848,15 +848,15 @@ class Client {
             if (resources['mounts']) {
                 params['config'].mounts = resources['mounts'];
             }
-            if (resources['scaling_group']) {
-                params['config'].scaling_group = resources['scaling_group'];
+            if (resources['spec']['scaling_group']) {
```
inureyes: resource['scaling_group'] depends on many codes. Is this modification working?

inureyes (changes requested):
It will be better to match parameter format between createSession and createSessionFromTemplate.

PR #1135, opened 2021-10-05
refactor: combine modify and add registry dialog

Sujin-Kim1:
- There were many overlaps between add-registry-dialog and modify-registry-dialog, so it was combined into one.
- Modify the registry password value to a valid value.


inureyes:
@Sujin-Kim1  Let me know when it is ready. :)

inureyes (approved):
LGTM

PR #1137, opened 2021-10-05
fix/Hide specific info from log file

### LINKED ISSUE #1136: Hide specific information from log file
Hide specific information from log file. e.g. connection details, API information, etc.

Hide specific information from log file.

lizable:
This PR resolves #1136.


src/lib/backend.ai-client-node.ts
```diff
@@ -261,8 +261,9 @@ class Client {
    * @param {AbortController.signal} signal - Request signal to abort fetch
    * @param {number} timeout - Custom timeout (sec.) If no timeout is given, default timeout is used.
    * @param {number} retry - an integer to retry this request
+   * @param {String} logText - the number of login attempts if not empty
    */
-  async _wrapWithPromise(rqst, rawFile = false, signal = null, timeout: number = 0, retry: number = 0) {
+  async _wrapWithPromise(rqst, rawFile = false, signal = null, timeout: number = 0, retry: number = 0, logText = '') {
```
adrysn: `_wrapWithPromise` should be generalized as much as possible since it is a kind of an abstract method that deals with every request from a client to a server. I'm not sure it is good to add a very specific parameter, `logText`, used only for hiding login passwords in login requests. Anyway, what is the reason to fill in the login attempt in the password field? Perhaps, we can just fill in with some stars?

lizable: I respect your opinion. I just update the code according to your comment!

lizable:
> I'm not sure about adding a very specific argument to `_wrapWithPromise`.
> 
> And, I found that the spinner is constantly running in the log page even though the logs are fully loaded. image

I guess this can be resolved in #1138 .

src/components/backend-ai-login.ts
```diff
@@ -811,9 +811,10 @@ export default class BackendAILogin extends BackendAIPage {
     );
     return this.client.get_manager_version().then(async ()=>{
       const isLogon = await this.client.check_login();
+      const loginAttempt = globalThis.backendaioptions.get('login_attempt', 0, 'general').toString();
       if (isLogon === false) { // Not authenticated yet.
         this.block(_text('login.PleaseWait'), _text('login.ConnectingToCluster'));
-        this.client.login().then((response) => {
+        this.client.login(loginAttempt).then((response) => {
```
adrysn, line 817: [`login`](https://github.com/lablup/backend.ai-webui/pull/1137/files#diff-c40dc377c78400ce60f662021339f44a4493199b7012d70166ee2d34ef058b69R620) does not take any parameter. Please remove `loginAttempt` from here if not used.

src/lib/backend.ai-client-node.ts
```diff
@@ -269,6 +277,12 @@ class Client {
     let errorDesc = '';
     let resp, body, requestTimer;
 
+    let encryptLoginInfo = (jsonEncodedStr :string) => {
+      let obj = JSON.parse(jsonEncodedStr);
```
adrysn, line 281: Better to use `const` instead of `let` since the `entryptLoginInfo` and `obj` themselves does not change. In JavaScript, a `const` object allows the changes in its property.

src/lib/backend.ai-client-node.ts
```diff
@@ -602,8 +615,7 @@ class Client {
   }
 
   /**
-   * Login into webserver with given ID/Password. This requires additional webserver package.
-   *
+   * Login into webserver with given ID/Password. This requires additional webserver package.   *
```
adrysn, line 618: Reason of this change?

PR #1088, opened 2021-07-28
feat: add webui grid status

sdeun24:
Task 1: Add massage for empty data. (Only for English - en.json)
Task 3: Modify lablup-loading-spinner.ts  then insert it into the loading grid.

Related: lablup/backend.ai-webui#1059

PR #1138, opened 2021-10-06
Feature: webui grid status

### LINKED ISSUE #1059: Let's show a certain message when there's no row to display in panel that using grids.

There are some panels that use grids to display data received from the manager server.
For example, running sessions list at running tab in Sessions page, and vfolders list at vfolders tab in Data & Storage pages. Sometimes, these panels show an empty list for several reasons.
- Case 1: there's literally **no data to display**.
- Case 2: manager server didn't brought any data (which means an error occurred during receiving results from requests)
- Case 3: due to network problem (poor speed, etc), data is still getting from the manager server. 

For now, the user can't distinguish each case, unless an error message shows up.
Therefore, It would be great to show proper information for each case written above.

어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.

> FYI: these are one of the suggestions to implement the feature. It could be changed if there's a better solution.
- Solution for case 1: Display message (e.g. `There's no data to display`) in the middle of the grid.
- Solution for case 2: This feature has been already implemented by checking `ok` msg.
- Solution for case 3: Show loading animation to make user feel data is on loading.

**Additional context**
Screen Shot 2021-07-10 at 1 47 25 AM
Screen Shot 2021-07-10 at 1 47 17 AM
Screen Shot 2021-07-10 at 1 47 12 AM

Sujin-Kim1:
- [x] User list is not working (GQL error)
- [x] Session table height is too large.
- [x] Image list table height is too small.
- [x] Resource group table height is too small.

this PR resolves #1059 
Related: #1088


Sujin-Kim1:
already resolved.

src/components/lablup-loading-dots.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.dots.style.display = 'none';
+    this.active = false;
+  }
+
+  /**
+   * Change whether dots are visible or not.
+   * */
+  async toggle() {
+    await this.updateComplete;
+    if (this.dots.active === true) {
```
Jaewoook: I think `this.dots` does not have `active` property because `this.dots` is instance of `HTMLDivElement`.

Jaewoook: I updated `lablup-loading-dots` component's toggle condition to use `this.dots` element display value.

inureyes (approved):
LGTM

PR #1141, opened 2021-10-09
In Session Creation, add Selected Image Details in last Page

TeusEE:
issue : https://github.com/lablup/backend.ai-webui/issues/1074

add plain text(Selected Image Information) in last stage of Creating New Session

image

### DIFF
--- a/resources/i18n/de.json
+++ b/resources/i18n/de.json
@@ -142,7 +142,7 @@
       "ClusterMode": "Cluster-Modus",
       "ClusterSize": "Clustergröße",
       "Node": "Knoten",
-      "Container": "Container",      
+      "Container": "Container",
       "TotalAllocation": "Gesamtzuweisung",
       "Ownership": "Eigentum",
       "SetSessionOwner": "Sitzungsinhaber",


src/components/backend-ai-session-launcher.ts
```diff
@@ -2700,6 +2700,14 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
 
     // monkeypatch for grid items in accessible vfolder list in Safari or Firefox
     this._grid?.clearCache();
+
```
lizable: Is there any specific reason to add  innerHTML instead of simply adding proper HTML tags (e.g. `p` or `span`) in [here](https://github.com/lablup/backend.ai-webui/blob/5b8cfb36dcf047033c099251959ef06ba625364c/src/components/backend-ai-session-launcher.ts#L3173-L3174)?

And also, It would be better to have some heading right above the environment and versions, like other allocation info does. Please refer the image below.
image

Lastly, It's my personal suggestion, is that may be we could display environment info as multi-selected tags(without icons).
e.g.)
image
Or we can simply add unordered list containing title such as "project", "language with version", "OS version".

src/components/backend-ai-session-launcher.ts
```diff
@@ -3172,6 +3180,7 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
           </div>
           <div id="progress-04" class="progress center layout fade">
             <p class="title">${_t('session.launcher.TotalAllocation')}</p>
+            <p class="vertical layout center flex blank-box" id = "selected_img"></p>
             <div class="vertical layout center center-justified cluster-total-allocation-container">
```
lizable: And also, please remove whitespaces between `id`, `=` and "selected_img".

PR #1142, opened 2021-10-11
Fix/remove one of the two scrolls

### LINKED ISSUE #1139: Remove one of the two scrolls in the TOS/PP dialog
**Describe the bug**
There are two scrolls in the Tos/PP dialog.
The dialog only needs one scroll.
The black scroll disappears at the bottom.

**To Reproduce**
1. Go to 'login Page'
2. Click on 'SIGN UP' button
3. Signup dialog appears.
4. Click Terms of service or Privacy Policy text at the bottom of the dialog, which is green colored.
5. ToS or Privacy Policy dialog shows up.
6. Scroll down

**Expected behavior**
There is only one scroll.

**Screenshots**
Screen Shot 2021-10-09 at 1 18 14 PM

https://user-images.githubusercontent.com/68562176/136643825-4228912c-118a-42e3-944a-d17ed4724b88.mov

**Client:**

- OS: macOS Big Sur 11.2
- App version: 21.03.5
- Browser version: Chrome 76
- Python version: 3.9.5
- NodeJS version: 16.10.0

**Server:**

- OS: macOS Big Sur 11.2
- [Backend.AI](http://backend.ai/) version: 21.09.0
- Python version: 3.9.5

**Additional context**

yujung7768903:
resolve #1139 
* Changed the value of the property max-height 

### The cause of the problem
* The height of the element with the class ".content.content-area" is longer than dialog.
Screen Shot 2021-10-12 at 2 24 27 AM

### Conclusion
* The max-height = dialog height(80vh) - {Header height(77px) - padding of the content(30px(top+bottom)) -footer height(20px)}
* The max-height should be `80vh - 127px`
Screen Shot 2021-10-12 at 2 39 18 AM

* The scroll bar has become very small because the value of max-width attribute of -webkit-scrollbar is 2px.
Screen Shot 2021-10-12 at 2 40 53 AM

### DIFF
--- a/src/components/lablup-terms-of-service.ts
+++ b/src/components/lablup-terms-of-service.ts
@@ -120,7 +120,7 @@ export default class LablupTermsOfService extends LitElement {
     return html`
       <backend-ai-dialog id="terms-of-service-dialog" class="terms-of-service-dialog" fixed blockscrolling persistent scrollable>
         <span slot="title">${this.title}</span>
-        <div slot="action" class="horizontal end-justified center flex layout>
+        <div slot="action" class="horizontal end-justified center flex layout">
           ${this.tosLanguages ? html`
             <mwc-select id="select-language" label="${_text('language.Language')}"
               @change=${() => this.changeLanguage()}>


src/components/lablup-terms-of-service.ts
```diff
@@ -116,7 +120,7 @@ export default class LablupTermsOfService extends LitElement {
     return html`
       <backend-ai-dialog id="terms-of-service-dialog" class="terms-of-service-dialog" fixed blockscrolling persistent scrollable>
         <span slot="title">${this.title}</span>
```
lizable: I guess we can apply width option in `mwc-select` style definition in [here](https://github.com/lablup/backend.ai-webui/pull/1142/files#diff-5918069c1ccd760be227141353f141e7cd7917e9108190e29435b2e6ac0ce62aR86) instead of using inline style.

src/components/lablup-terms-of-service.ts
```diff
@@ -102,6 +102,10 @@ export default class LablupTermsOfService extends LitElement {
             height: 20px;
           };
         }
+
```
lizable, line 105: Resolving overlapping in title with `line-height` is a good approach.

src/components/lablup-terms-of-service.ts
yujung7768903: Initially, I apply width option in mwc-select style definition. But the width of `action` is fixed 200px, causing a problem that the title went over two lines. Should I add style in css for `action` instead of using inline style?       

Screen Shot 2021-10-16 at 8 32 20 PM

yujung7768903:
If I remove the width attribute of `mwc-button.full`, the button becomes too small, so I changed it to 70%.
Screen Shot 2021-10-16 at 9 37 19 PM

Screen Shot 2021-10-16 at 9 37 45 PM

src/components/lablup-terms-of-service.ts
lizable: Hmm, for me It looks OK(no fixed width or nothing) when I removed inline style definition in `div` element and add width value to `mwc-select` definition.
Please remove the inline style unless there's no way to use it (only when point-scoring system of css about the element has been twisted). Instead, you can put `width:135px` into `mwc-select` definition.

This is what I see after changing the style definition just like I requested you above.
Screen Shot 2021-10-16 at 10 54 23 PM

yujung7768903:
I understood! I changed that as lizable requested.

src/components/lablup-terms-of-service.ts
```diff
@@ -116,7 +120,7 @@ export default class LablupTermsOfService extends LitElement {
     return html`
       <backend-ai-dialog id="terms-of-service-dialog" class="terms-of-service-dialog" fixed blockscrolling persistent scrollable>
         <span slot="title">${this.title}</span>
-        <div slot="action" class="horizontal end-justified center flex layout">
+        <div slot="action" class="horizontal end-justified center flex layout>
           ${this.tosLanguages ? html`
```
lizable: Please fix the typo.
You need to close the double quotes mark. (`class="horizontal end-justified center flex layout`<- right here, no `"` !!) If not, it will bring tremendous error.

Except that, everything is fine.

yujung7768903: I fixed it now!

lizable: I couldn't address your update but I just fixed it, anyway.

PR #1162, opened 2021-11-25
Feature/option to enqueue compute session

### LINKED ISSUE #1161: Add an option to enqueue a compute session even if there are no available resources
A user cannot create a compute session when there is no available resource in the Agent hosts. In the absence of the resources, the session launcher dialog disables the CREATE button.

Some customers find this inconvenient since they have to wait until other users destroy their compute sessions to create their own. They rather wanted to create a PENDING session immediately and then get the RUNNING session automatically just after the resources are available.

So, it would be better to provide a global option (in `config.toml`) to enqueue a compute session regardless of the availability of the resources. Perhaps, the name goes like `always_enqueue_compute_session` (just an example).

For example, let's assume that one Agent has 10 cores and 32 GB memory (omit fGPU for simplicity). In the current implementation, User2 cannot create a PENDING session in the following scenario:
* User1 creates a session with 8 cores and 30 GB memory.
* User2 wants to create (enqueue) a session with 4 cores and 10 GB memory, but (s)he can't because the LAUNCH SESSION button is disabled.

Now, let's assume that the global option `always_enqueue_compute_session` is set to `true`. Then,
* User1 creates a session with 8 cores and 30 GB memory.
* User2 creates (enqueues) a session with 4 cores and 10 GB memory.
* User2's compute session is enqueued and marked as PENDING.
* It will automatically converted to a RUNNING session when User1 destroys the first compute session.

Of course, when the `always_enqueue_compute_session` is set to `false`, the session creation behavior should be the same as before.

inureyes:
This PR resolves #1161

### DIFF
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -83,7 +83,7 @@ export default class BackendAILogin extends BackendAIPage {
   @property({type: Boolean}) allow_signout = false;
   @property({type: Boolean}) allow_project_resource_monitor = false;
   @property({type: Boolean}) allow_manual_image_name_for_session = false;
-  @property({type: Boolean}) always_enuqneue_compute_session = false;
+  @property({type: Boolean}) always_enqueue_compute_session = false;
   @property({type: Boolean}) openPortToPublic = false;
   @property({type: Boolean}) maxCPUCoresPerContainer = 64;
   @property({type: Boolean}) maxMemoryPerContainer = 16;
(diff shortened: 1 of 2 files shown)


inureyes:
* Need to follow : precondition-failed bug.

inureyes:
So manager-side fix is needed, I guess. @adrysn

adrysn (approved):
I tested and it works.

PR #1172, opened 2021-12-09
fix: mwc-switch not working

Sujin-Kim1:
- `mwc-switch` doesn't use checked property and overrides click method (not change).


inureyes (approved):
LGTM

PR #1181, opened 2021-12-30
Add a feature to update a session's name in Web-UI

Sujin-Kim1:
this resolves https://github.com/lablup/backend.ai/issues/296

- Edit icon only shows on the running session.
- How to update session name
  1. Click the edit icon.
  2. Write the new session name.
  3. Click the edit icon again.

**screenshot**


adrysn (commented):
I think the feature is working good.

But, I have some opinions on the visual aspect:
* The edit icon is too big for me.
* In edit mode, the size of the session name column changes too much and I think it's not good looking.
* The edit icon should be changed to a "done" icon in editing mode.

Sujin-Kim1:
> I think the feature is working good.
> 
> But, I have some opinions on the visual aspect:
> 
> * The edit icon is too big for me.
> * In edit mode, the size of the session name column changes too much and I think it's not good looking.
> * The edit icon should be changed to a "done" icon in editing mode.

Thanks for the feedback! 
I reduced the size of the session rename column until helper text doesn't get cut off (min-content) 
image

src/components/backend-ai-session-list.ts
```diff
@@ -1330,8 +1415,37 @@ export default class BackendAiSessionList extends BackendAIPage {
...
+            width: min-content;
+          }
+          #session-rename-icon {
+            --mdc-icon-size: 20px;
+          }
+        </style>
         <div class="layout vertical start">
-          <div>${rowData.item[this.sessionNameField]}</div>
+          <div class="horizontal center center-justified layout">
+            <div id="session-name-field">${rowData.item[this.sessionNameField]}</div>
```
lizable: Everything seems fine, but one thing bothers me.
If the session name is almost reached the limit (64 characters), it's bit cumbersome since the user always needs to resize the sessioninfo column first.

How about using `pre` tag instead of `div` tag with style definition like this:
```
#session-name-field {
            display: block;
            margin-left: 16px;
            white-space: pre-wrap;
            word-break: break-all;
          }
...
<pre id="session-name-field">${rowData.item[this.sessionNameField]}</pre>
```

**screenshot**:
wrapped-session-name mov

agatha197: you're right it'll be better! I'll apply that right away.

lizable (changes requested):
How about adding `auto-width` in session info column element?
For now, it seems too narrow to contain session name, and environment info.
(please refer [here](https://github.com/lablup/backend.ai-webui/pull/1181/files#diff-4d843242713270c38dd580c57115bfd4160a27e2b9922086d2bbb83efe5d1895R1921).)

**Before adding `auto-width`**:
before-adding-autowidth

**After adding `auto-width`**:
Screen Shot 2022-01-12 at 3 26 44 PM

inureyes (approved):
LGTM

PR #1182, opened 2022-01-03
add: session mode selection area and batch session config in session launcher

### LINKED ISSUE #1180: Let's support batch session UI.
Currently, Backend.AI CLI supports executing sessions in batch. In order to match the user experience in both CLI and WebUI, batch execution should be supported in WebUI as well.

어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.
Executing sessions in batch need three key features.
- Launch session after a certain time (e.g. 10min, 15min, or specific date in future, etc.)
- _Notify after the batch session is terminated._
- Split session list in running status into three: `running`, `interactive`, `batch` (suggested by @inureyes )
   - `running`: contains every session not in "TERMINATED", "CANCELLED" status
   - `interactive`: contains every session that satisfies `running` status and type of session is "interactive"
   - `batch`: contains every session that satisfies `running` status and type of session is "batch"

**Additional context**

lizable:
### Description
---------
This PR resolves #1180.

**Key Feature**:
- [x] Launch session after a certain time (e.g. 10min, 15min, or specific date in future, etc.)
- [ ] _Notify when the status of batch session has been changed._
- [x] Split session list in running status into three: `running`, `interactive`, and `batch` (suggested by @inureyes )

### Before
---------
- session launcher:
session launcher(before)

- session list:
session list(before)

- session status detail dialog:
session status detail dialog(before)

### After
---------
- session launcher:
   - selectable UI for session type: 
   session launcher(after)-01

   - configuration section for session in batch mode:
   session launcher(after)-02

- session list:
session list(after)-01
session list(after)-02

- session status detail dialog:

> NOTE: Show reserved time so that user could know when the session in batch mode starts

session status detail dialog(after)


lizable:
this feature; "_Notify when the status of batch session has been changed._" will be handled as a separate issue.

src/components/backend-ai-session-launcher.ts
```diff
@@ -1247,6 +1268,22 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
+    if (this.sessionType === 'batch') {
+      const editor = this.shadowRoot.querySelector('lablup-codemirror#command-editor');
+      config['startupCommand'] = editor.getValue();
+
+      const scheduledTime = this.shadowRoot.querySelector('vaadin-date-time-picker').value;
+      if (scheduledTime) {
+        // modify client timezone offset
+        const getClientTimezoneOffset = () => {
+          let offset = new Date().getTimezoneOffset();
+          const sign = offset <0? '+' : '-';
```
agatha197: Can you add a space after `<`?

tsconfig.json
```diff
@@ -22,7 +22,7 @@
     "esModuleInterop": false,
     "experimentalDecorators": true,
     "emitDecoratorMetadata": false,
-    "lib": ["es6", "dom", "es2016", "es2017"]
+    "lib": ["es6", "dom", "es2016", "es2017", "ES2020"]
```
agatha197: I got this message when running `npm run build:d` and can't bring every module.
image
And after changing this to `es2020`, the error is removed and works properly.

src/components/backend-ai-session-launcher.ts
```diff
@@ -2824,6 +2962,25 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
                            helper="${_t('maxLength.64chars')}"
                            validationMessage="${_t('session.launcher.SessionNameAllowCondition')}">
             </mwc-textfield>
+            <wl-expansion class="editor" name="editor" ?disabled="${this.sessionType === 'interactive'}">
+              <span slot="title">${_t('session.launcher.BatchModeConfig')}</span>
```
agatha197: How about this expansion to be open when batch mode?
I got to know the startup command is required after making a session. Therefore, it would be nice to inform the user more explicitly that startup commands are required in batch mode.

Sujin-Kim1 (changes requested):

src/components/backend-ai-session-launcher.ts
lizable: Thanks for the suggestion! I just applied it.

src/components/backend-ai-session-launcher.ts
lizable: Oops, you're right. Better to split numbers and symbols with whitespace.

src/components/backend-ai-session-launcher.ts
```diff
@@ -2746,6 +2794,98 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
 
+  /**
+   * Toggle startup code input section according to session type
+   *
+   * @param {Event} e
+   */
+  _toggleStartUpCommandEditor(e) {
+    this.sessionType = e.target.value;
+    const startUpCommandEditor = this.shadowRoot.querySelector('#batch-mode-config-section');
+    startUpCommandEditor.style.display = (this.sessionType === 'batch') ? 'inline-flex' : 'none';
```
agatha197: When `startUpCommandEditor` starts to appear, `command-editor` just looks like a black block.
I think it's better to show up with an index.
Here's the way to show an index.
```javascript
const editor = this.shadowRoot.querySelector('#command-editor');
editor.refresh();
```
[Result Screenshot]
image

Sujin-Kim1 (changes requested):

src/components/backend-ai-session-launcher.ts
lizable: @Sujin-Kim1 I agree. And I also added autofocus when the session type is batch mode.
Could you check it again?

agatha197: It works properly👍🏻

Sujin-Kim1 (approved):
LGTM!

lizable:
I changed Korean translation for both `batch` mode and `interactive` mode so that users can literally understand what it means.
batch-interactive-korean-translation

src/components/backend-ai-session-launcher.ts
```diff
@@ -2746,6 +2794,104 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
+  }
+
+  /**
+   * Toggle scheduling time when session type is in batch
+   *
+   */
+  _toggleScheduleTime() {
+    this.useScheduledTime = this.shadowRoot.querySelector('#use-scheduled-time').selected;
+    this.shadowRoot.querySelector('vaadin-date-time-picker').style.display = this.useScheduledTime ? 'block': 'none';
+    this._getSchedulableTime();
```
adrysn: Do we need to call `_getSchedulableTime` when `useScheduledTime` is off?

src/components/backend-ai-session-launcher.ts
```diff
@@ -2746,6 +2794,104 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
+      }
+    } else {
+      schedulerEl.value = getFormattedTime(futureTime);
+    }
+    this._setRelativeTimeStamp();
+
+    // interval every 1 sec.
+    setTimeout(() => {
+      this._getSchedulableTime();
+    }, 1000);
```
adrysn: This might lead to a potential memory leak and an unresponsive UI.

`_getSchedulableTime` is called whenever the value of the date-time picker AND the `#use-scheduled-time` switch change. This leads to the duplicated pending calls to this `setTimeout` which would not be the desired behavior. You may check this by inserting a debug message at the start of the `_getSchedulableTime` method and repeat turning on/off the `#use-scheduled-time`.

The `setTimeout` should be explicitly released whenever it is not used. Also, we need to ensure there is only one pending call to the `setTimeout` in this case. We don't need to call it when user did not enable the `#use-scheduled-time` switch.

src/components/backend-ai-session-list.ts
```diff
@@ -326,11 +326,11 @@ export default class BackendAiSessionList extends BackendAIPage {
   }
 
   get _isRunning() {
-    return this.condition === 'running';
+    return ['batch', 'interactive', 'running'].includes(this.condition);
   }
 
   _isPreparing(status) {
-    const preparingStatuses = ['RESTARTING', 'PREPARING', 'PULLING'];
+    const preparingStatuses = ['RESTARTING', 'PREPARING', 'PULLING', 'PENDING'];
```
adrysn: Is there a reason to add the `PENDING` status to the `preparingStatuses`?

src/components/backend-ai-session-list.ts
```diff
@@ -1434,7 +1451,7 @@ export default class BackendAiSessionList extends BackendAIPage {
             <mwc-icon-button class="fg red controls-running"
                                icon="power_settings_new" @click="${(e) => this._openTerminateSessionDialog(e)}"></mwc-icon-button>
           ` : html``}
-          ${(this._isRunning && !this._isPreparing(rowData.item.status)) || this._APIMajorVersion > 4 ? html`
+          ${(this._isRunning && !this._isPreparing(rowData.item.status)) && this._APIMajorVersion > 4 ? html`
```
adrysn: It is intended to display the log button always for running and preparing, but not pending, sessions when the api version is greater than 4. Is there a reason to change this?

adrysn (changes requested):
I have some opinions on this PR.

adrysn:
I also think it is better to display the type of session in the session list grid even though this PR introduces the "Interactive" and "Batch" tabs. Users may want to distinguish the type of the sessions in the "Running" and "Finished" tabs.

src/components/backend-ai-session-list.ts
lizable: Nope, I just added it to make sure the log icon to be disabled in `PENDING` status, but I found that it doesn't need to be added.

src/components/backend-ai-session-list.ts
lizable: @adrysn
Until now, if API version is upper than 4, then log icon has been enabled whether the session is `PENDING` status or not. But the session in `PENDING` status, which is corresponding to the container about to be created, will bring an error when it calls log from the manager server. I think the log icon should be `disabled` or at least we should provide more detailed error msg, such as "there is no corresponding container".

src/components/backend-ai-session-launcher.ts
lizable: > I also think it is better to display the type of session in the session list grid even though this PR introduces the "Interactive" and "Batch" tabs. Users may want to distinguish the type of sessions in the "Running" and "Finished" tabs.

I agree. I'll add a column for session type.

src/components/backend-ai-session-list.ts
lizable: Screen Shot 2022-01-10 at 9 10 17 PM

src/components/backend-ai-session-launcher.ts
lizable: I replaced the recursive call of `_getSchedulableTime` with pausing/resuming `setInterval`. This will remarkably reduce calling `_getSchedulableTime` function.

src/components/backend-ai-session-launcher.ts
lizable: Nope, we don't. I just fixed this issue using `setInterval` and `clearInterval` function.

lizable:
I updated the session type to be more simple display compared to last commit.
Screen Shot 2022-01-13 at 1 45 36 AM

adrysn (approved):
LGTM.

inureyes (approved):
LGTM

PR #1188, opened 2022-01-10
feat: toggle token email confirmation option

### LINKED ISSUE #1185: Add an option to hide the [invitation token field] and [email sent dialog] during user signup
The current signup dialog always displays an invitation token field.
image

This field is only meaningful when the cloud plugin is installed in the Manager server. Without the plugin, a user can still signup on their own, but they have to fill a dummy value in the token field. Also, after the successful signup, an email sent dialog always appears. This is absurd, so we need to add an option to hide this field.

lizable:
### Description
---------
This PR resolves #1185.
### Before
---------
No options about hiding unnecessary input fields and messages in signup dialog, such as "token" and "email sent msg".

### After
---------
Now `allowSignupWithoutConfirmation` stands for the visibility "token" input field and "emails sent msg".
if the value of `allowSignupWithoutConfirmation` in configuration file (`config.toml`) is `true`, then it will be regarded as "sign up without confirmation" as is, so signup dialog will not display "token" input field or "email sent msg" after registration.

### DIFF
--- a/config.toml.sample
+++ b/config.toml.sample
@@ -10,7 +10,7 @@ signupSupport = false
 allowSignout = false
 allowAnonymousChangePassword = false
 allowProjectResourceMonitor = true
-useManualSignUp = false
+allowSignupWithoutConfirmation = false
 autoLogout = false
 debug = false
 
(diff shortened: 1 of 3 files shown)


adrysn (approved):
LGTM!

PR #1193, opened 2022-01-14
style: change width and height of log dialog

Sujin-Kim1:
- Adjust the height to remove the wide blanks in the log dialog.
- Adjust the width to increase the consistency

[Screenshot]
- wide screen
image
-narrow screen
image


inureyes (approved):
LGTM

PR #1194, opened 2022-01-14
feat: arbitrary path vfolder mounts

SonMinWoo:
Resolve: lablup/backend.ai#329

This is only about vfolder mount. Original functions of /home/work directory is not changed(logs, entrypoint, any other dotfiles).
Now only English user can see notice text. If this is merged, the other languages have to see same meaning.

If more good idea about UX(notice to user about any pre-existing container paths overridden by the manually specified mountpoints become not visible and not accessible inside the container) like alert/modal, please make issue.

Related PR:
lablup/backend.ai-agent#334 
lablup/backend.ai-manager#516
lablup/backend.ai-client-py#204

**Test**
test1
test1_1
test1_2
test2
test3
test4


src/components/backend-ai-session-launcher.ts
```diff
@@ -1271,7 +1271,7 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
         config['mount_map'] = {};
         for (const f in this.folderMapping) {
           if ({}.hasOwnProperty.call(this.folderMapping, f)) {
-            config['mount_map'][f] = '/home/work/' + this.folderMapping[f];
+            config['mount_map'][f] = '/' + this.folderMapping[f];
           }
```
lizable, line 1304: For now, we've been using aliasing as renaming mounted folder in `/home/work/` subdirectory of container corresponding to session. When it comes to change this policy to mount in root(`/`) not in `/home/work/` It will bring chaos, unless we guide with an extra explanation or another option.

IMHO, we need to explicitly add the absolute path as a default input in every alias column, so that users could know that this alias would change the mount path completely.

resources/i18n/en.json
```diff
@@ -218,7 +218,7 @@
       "EnvConfigWillDisappear": "Any unsaved environment variable and value will be disappeared.",
       "SharedMemoryAbbr": "SHMEM",
       "GPUSlot": "Slot",
-      "FolderAlias": "Alias",
+      "FolderAlias": "Alias(※ be careful about overridden of pre-existing folder)",
       "FolderAliasOverlapping": "Aliases overlap with other names that already exist",
```
lizable, line 223: More than half of the words in the sentence are ellipsis. In light of the importance of the change, we need to fix this. (※ Please refer to the red box in the screenshot below:)

**screenshot**:
alias-review

lizable (changes requested):
I guess some parts need to be fixed. please refer to the comments below :

resources/i18n/en.json
inureyes: It will be better to add question icon after "alias" to introduce how-to mount folder at any location.

inureyes: Reference: (?) icons on resource sliders. @SonMinWoo @lizable

src/components/backend-ai-session-launcher.ts
```diff
@@ -2001,14 +2006,33 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
         </template>
       `,
       root
     );
   }
+
+  infoHeaderRenderer(root, column?) {
+    render(
+      html`
+      <div style='display: flex; align-items: center'>
```
lizable: We already have style definitions in class. Please refer to [the component link](https://github.com/lablup/backend.ai-webui/blob/e0ec8345640bdab6ce122bfce238199a09fab298/src/plastics/layout/iron-flex-layout-classes.js) and update it.

src/components/backend-ai-session-launcher.ts
```diff
@@ -2001,14 +2006,33 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
+        <span style='text-align:center; font-size:16px; font-family: var(--general-font-family); font-weight: 500'>${_t('session.launcher.FolderAlias')}</span>
+        <mwc-icon-button icon="info" class="fg green info" @click="${(e) => {
+          e.stopPropagation();
```
lizable: IMO, it's better to be converted to class function.

resources/i18n/en.json
```diff
@@ -218,8 +219,9 @@
       "EnvConfigWillDisappear": "Any unsaved environment variable and value will be disappeared.",
       "SharedMemoryAbbr": "SHMEM",
       "GPUSlot": "Slot",
-      "FolderAlias": "Alias",
+      "FolderAlias": "Path & Alias",
       "FolderAliasOverlapping": "Aliases overlap with other names that already exist",
+      "DescFolderAlias":"<p>1. If you fill out this input, folder name will change to alias when it mount<br/>2. Start with /: absolute path<br/>3. Not start with / or blank: sub path of /home/work<br/>4. Don't use alias same as default system folder of linux.<br/>5. Please check final path in 'Mounted folders' below.</p>",
```
lizable: I think this is important information, so It should be exposed automatically when the active carousel page is on the "mount page"(which is, page 2). For now, user can see the guide only if they click the info icon.
Also, Let's use a shorter but more powerful statement. Let's say, the statement "If you fill out this input, folder name will change to alias when it mount", seems too descriptive. I'd like to suggest "Folder name will be changed to the alias when it mounts." the other statements seem to be Okay, but It would be appreciated to polish them up.
Lastly, a number list would be appropriate when order matters for each instruction. I don't think these numbers are meaningful in this instruction. How about using a bullet list instead of a number list?

src/components/backend-ai-session-launcher.ts
```diff
@@ -3058,9 +3082,9 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
                                               flex-grow="0"
                                               text-align="center"
                                               auto-select></vaadin-grid-selection-column>
-                <vaadin-grid-filter-column header="${_t('session.launcher.FolderToMount')}"
+                <vaadin-grid-filter-column header="${_t('session.launcher.FolderToMountList')}"
                                           path="name" resizable></vaadin-grid-filter-column>
-                <vaadin-grid-column .renderer="${this._boundFolderMapRenderer}" header="${_t('session.launcher.FolderAlias')}">
+                                          <vaadin-grid-column .renderer="${this._boundFolderMapRenderer}" path=" ${_t('session.launcher.FolderAlias')}" .headerRenderer="${this._boundPathRenderer}">
```
lizable: Please remove the space in front of Line 3087. :)

lizable (changes requested):
@SonMinWoo 
I have a few requests for change. Please refer to the comments below.
Getting really close to merge! 💪

resources/i18n/en.json
inureyes: Please remove HTML tag from language file. (it should be on the .ts file). Some parts have inline HTML tags in the language resource, however they were inevitable.

Break the line into 5 resources.

src/components/backend-ai-session-launcher.ts
```diff
@@ -2001,14 +2006,37 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
       `,
       root
     );
   }
+
+  infoHeaderRenderer(root, column?) {
+    render(
+      html`
+      <div class="horizontal layout center">
+        <span style='text-align:center; font-size:16px; font-family: var(--general-font-family); font-weight: 500'>${_t('session.launcher.FolderAlias')}</span>
```
lizable: Let's use id in `<span>` element and style definition on id above.
Also please add a function definition.

src/components/backend-ai-session-launcher.ts
```diff
@@ -3058,9 +3090,9 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
                                               flex-grow="0"
                                               text-align="center"
                                               auto-select></vaadin-grid-selection-column>
-                <vaadin-grid-filter-column header="${_t('session.launcher.FolderToMount')}"
+                <vaadin-grid-filter-column header="${_t('session.launcher.FolderToMountList')}"
                                           path="name" resizable></vaadin-grid-filter-column>
```
lizable: Please remove spaces in line 3095.

resources/i18n/ko.json
```diff
@@ -218,8 +219,9 @@
       "EnvConfigWillDisappear": "저장되지 않은 환경 변수 설정값은 사라집니다.",
       "SharedMemoryAbbr": "공유메모리",
       "GPUSlot": "슬롯",
-      "FolderAlias": "대체이름",
+      "FolderAlias": "경로 & 대체이름",
       "FolderAliasOverlapping": "대체 이름이 이미 존재하는 다른 이름과 겹칩니다.",
```
lizable, line 223: "Path & Alias" should follow the translation of the current language.
For example, "Path & Alias" is "경로 & 대체이름" in Korean.

resources/i18n/ko.json
```diff
@@ -137,7 +137,8 @@
       "SessionNameOptional": "세션 이름 (옵션)",
       "Environments": "실행 환경",
       "Start": "시작",
-      "FolderToMount": "마운트할 폴더",
+      "FolderToMount": "마운트할 폴더 (중복 폴더명 사용 주의)",
```
lizable: @SonMinWoo Is there any reason to add "(중복 폴더명 사용 주의)" in Korean translation only?

lizable (changes requested):
There are some minor linting to fix and suggestions.

resources/i18n/ko.json
SonMinWoo: This is the content I wrote before creating the information window. Correction done.

resources/i18n/en.json
lizable: @inureyes Hmm, we've been using HTML tags in here. Is there any specific reason to get rid of tags? If we do split the resource into five keys and values, it's a bit over-descriptive.

SonMinWoo: For the reasons described above and for the unity with other '_helpDescription' in the code, only the p tag was used.

resources/i18n/ko.json
lizable: Any changes? Still I can see "Path & Alias" in here.

lizable (changes requested):
Please check the minor request in the translation part.

Sujin-Kim1:
For now, the alias field is slightly cropped.
image

As you see this [link](https://cdn.vaadin.com/vaadin-web-components/23.0.0-alpha2/#/elements/vaadin-grid-column#property-width), you can solve this one by adding a `width` property. When I tested, 135px is good for me.
image

src/components/backend-ai-session-launcher.ts
```diff
@@ -2750,6 +2785,10 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
     nextButton.style.visibility = this.currentIndex == this.progressLength ? 'hidden' : 'visible';
     this.shadowRoot.querySelector('#launch-button-msg').textContent = this.progressLength == this.currentIndex ? _text('session.launcher.Launch') : _text('session.launcher.ConfirmAndLaunch');
 
+    if (this.currentIndex == 2) {
+      this._showPathDescription();
```
agatha197: It's good to show this information right away when `currentIndex` is 2. However, if the user is already aware of this, showing it every time can cause irritation. Therefore, it is better to add a function not to see each other today. It would be nice to use localStorage(cookie) or implement it the way you want.
> You can refer localStorage by `bacakend-ai-login.ts` file

Sujin-Kim1 (changes requested):

Sujin-Kim1 (approved):
LGTM!

src/components/backend-ai-session-launcher.ts
```diff
@@ -2001,14 +2017,67 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
+    this._helpDescriptionIcon = '';
+    const pathDialog = this.shadowRoot.querySelector('#help-description');
+    setTimeout(() => this.setPathContent(pathDialog));
+    pathDialog.show();
+  }
+
+  setPathContent(pathDialog) {
+    const pathLastChild = pathDialog.children[pathDialog.children.length - 1];
+    const pathContentLastChild = pathLastChild.children[pathLastChild.children.length - 1];
+    if (pathContentLastChild.children.length < 6) {
```
lizable: Let's not use the hard-coded numbers only if it doesn't need any explanation. How about assigning value to a new variable so that others could understand your code in future.

lizable (changes requested):
@SonMinWoo  I have one minor thing to request!

src/components/backend-ai-session-launcher.ts
SonMinWoo: Edit done!

lizable (approved):
LGTM.

inureyes (approved):
LGTM

PR #1200, opened 2022-01-20
fix: change auto_logout value set logic

ggstone0523:
### Resolve: lablup/backend.ai-webui #1173 

**1. bugfix for auto logout feature**

- This bug is caused by changing the 'auto logout' setting on the 'UserSettings' page does not change value of 'this.auto_logout' accordingly.
- The reason why the value of 'this.auto_logout' does not change is that it continues to receive the value only from element 'autoLogout' of config.toml.

### DIFF
--- a/src/components/backend-ai-webui.ts
+++ b/src/components/backend-ai-webui.ts
@@ -306,9 +306,11 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
       this.connection_server = config.general.connectionServer;
       // console.log(this.connection_server);
     }
-    this.auto_logout = (typeof config.general !== 'undefined' && 'autoLogout' in config.general && globalThis.backendaioptions.get('auto_logout') === null) ?
-      config.general.autoLogout :
-      globalThis.backendaioptions.get('auto_logout', false);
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-webui.ts
```diff
@@ -306,7 +306,9 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
       this.connection_server = config.general.connectionServer;
       // console.log(this.connection_server);
     }
```
lizable, line 308: It works okay, but It seems too long to use ternary operation. How about using plain if-else statement for better readability?

lizable (changes requested):
I'm not sure it's okay to use ternary operation in such a long statement. @ggstone0523
But All seems to work fine, so I want your opinion on this code.

lizable:
LGTM!

ggstone0523:
I also think so writing an if-else statement is a better way in terms of readability than using a ternary operation.

Sujin-Kim1 (approved):
LGTM

lizable (approved):

inureyes:
LGTM

Issue #1199, opened 2022-01-19
Session resource allocation doesn't match


Sujin-Kim1:
**Describe the bug**
If you try to create a session with presets that don't have a fixed size of shared memory, it appears differently in all fields representing the size of the shared memory.

**To Reproduce**
1. Set a preset resource that doesn't have a size of shared memory.
2. Create a session with that resource preset.
4. See an error

**Expected behavior**
All fields representing the size of the shared memory are the same.

**Screenshots**
<img src="https://user-images.githubusercontent.com/28584164/150087809-01a9863c-d4a9-4dec-9dcd-c71b31734baa.png" width=400/>
<img src="https://user-images.githubusercontent.com/28584164/150087979-65c56fe5-d42f-403a-bfc8-2c314afeae8d.png" 
 width=400/>

**Client:**
- OS: macOS Monterey 12.4
- Browser version: Chrome
- Python version: 3.9.9
- NodeJS version: 16.13.1

PR #1205, opened 2022-01-24
fix: session resource allocation doesn't match

SonMinWoo:
Resolve: lablup/backend.ai-webui #1199

Original default shared memory setting code
img1
img2

Shared memory type value of this case is object.
img5

This is fixing code, but it is not activated mostly because of no shared memory limit in most cases. 
img3
img4

All session below don't have any shared memory limit
img6

So I changed preferred value to 0.0625GB(64MB) which is enough for single core CPU.


Sujin-Kim1 (approved):
LGTM.

inureyes (approved):

PR #1206, opened 2022-01-24
fix: monkeypatch style manipulation according to the number of list-items in environment

lizable:
## Description
This PR will resolve overflowing list-items in the environment selection UI in session launcher.

## Before
When items in the supported environment list exceed 7 or more, the scrollbar triggers extra width for each item(environment) and causes overflow.
environment-selection-overflow

## After
Width of each item in the supported environment list stays still regardless of scrollbar.
- When the number of item in supported environment list exceeds more than 10
environment-selection-overflow-resolved

> ⚠️ &nbsp; NOTE: `mwc-list-item` and `mwc-select` element internally use their own hierarchy, so it doesn't allow flexible custom layouts. Therefore I need to use arbitrary code to circumvent the overflow problem.

### DIFF
--- a/src/components/backend-ai-session-launcher.ts
+++ b/src/components/backend-ai-session-launcher.ts
@@ -1126,7 +1126,7 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
       this._toggleScheduleTime(!this.useScheduledTime);
 
       // monkeypatch for mwc-list-item width adjustment
-      const scrollEnabledNum = 7;
+      const scrollEnabledNum = 11;
       const environmentListItems = this.shadowRoot.querySelectorAll('mwc-select#environment > mwc-list-item');
       if (this.languages.length > scrollEnabledNum) {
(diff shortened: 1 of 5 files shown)


src/components/backend-ai-session-launcher.ts
```diff
@@ -568,10 +571,6 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
           --mdc-select-min-width: 190px;
         }
 
-        mwc-select > mwc-list-item {
-          width: 370px; // default width
```
lizable, line 572: This style definition enabled horizontal scrollbars, which was not a desirable result. It could be solved by two custom style variables in `mwc-select` element.

When the value of the attribute "fixedMenuPosition" in `mwc-select` set true,  the width of corresponding list-item elements can be set in both `--mdc-menu-min-width` and `--mdc-menu-max-width`. And the value of each custom style variable need to be same as the width of the dialog component. We shall not defining them in `mwc-list-item` directly.We shall not defining them in `mwc-list-item` directly.

Issue #1122, opened 2021-09-04
Text overlapping of the resources in the "Total Allocation" panel


adrysn:
When we allocate a large amount of resources to launch a compute session, the texts in the "Total Allocation" panel overlap with the neighbors, just like the images below.

<img width="424" alt="image" src="https://user-images.githubusercontent.com/7539358/132097063-308137ed-2008-481a-af68-3dade12211bf.png">
<img width="435" alt="image" src="https://user-images.githubusercontent.com/7539358/132097068-122e2cc7-ff72-496f-a056-7d94f6d44c56.png">

PR #1209, opened 2022-01-25
fix: round mem and shmem allocation size in panel

ggstone0523:
### Resolve: lablup/backend.ai-webui #1122

- This bug is caused by overflow the property(cpu_request, mem_request, shmem_request, gpu_request) value in "total allocation" panel when allocate large amount of resource.
- Add scroll to "total allocation" panel to prevent overflow.

**test image**
PR4

PR4(2)


lizable (changes requested):
I'm afraid that we should solve this problem in another way instead of using a horizontal scrollbar. It would degrade user experience, I guess.
How about using a flexible layout? When width is long enough, it shows all in one row, but when content seems to be too much to carry in one row, it splits into two rows.

ggstone0523:

- remove scroll in "total allocation" panel
- apply flexible layout to total allocation panel

**test image**
PR4_review1_2

PR4_review1_1

ggstone0523:

- Remove flexible layout from "total allocation" panel
- Round mem allocation size to the first decimal place.
- Round shmem allocation size to the second decimal place.
- "_roundResourceAllocation" function is round the value according to the specified number of digits. 

**test image**
PR4_review2_1

lizable (approved):
LGTM

adrysn (changes requested):
The code looks okay. Just please resolve the conflict before merge.

ggstone0523:
**The revision was completed by reflecting the comments.**

- resolve the conflict

adrysn (approved):
LGTM.

PR #1211, opened 2022-01-27
feature: send target app's protocol (http or tcp) when starting wsproxy

kyujin-cho:
This PR adds new protocol variable which defines app's underlying transportation protocol. Sending protocol to wsproxy will help wsproxy determine equivalent proxying mechanism.


src/components/backend-ai-app-launcher.ts
```diff
@@ -521,14 +521,23 @@ export default class BackendAiAppLauncher extends BackendAIPage {
...
-    const kInfo = await globalThis.backendaiclient.computeSession.get(['scaling_group'], sessionUuid);
+    const kInfo = await globalThis.backendaiclient.computeSession.get(['scaling_group', 'service_ports'], sessionUuid);
     if (kInfo === undefined) {
       this.indicator.end();
       this.notification.text = _text('session.CreationFailed'); // TODO: Change text
 
       this.notification.show();
       return Promise.resolve(false);
     }
+    console.log(kInfo.compute_session.service_ports)
```
inureyes: Please remove the debug lines. :D

inureyes (commented):
Do we need to update proxy in the web UI (`src/lib/wsproxy`) ? I think it does not need to accept the protocol since it just uses localhost HTTP, however for the equivalent API interface between web version and local one?

kyujin-cho:
> Do we need to update proxy in the web UI (`src/lib/wsproxy`) ? I think it does not need to accept the protocol since it just uses localhost HTTP, however for the equivalent API interface between web version and local one?

as you mentioned, we can just skip handling `protocol` query parameter on the webUI's wsproxy since we can't handle TCP based proxying on web version. But perhaps we can normalize `manager.js`' TCP app blocking code block with this `protocol` parameter. I'll add a new commit about this feature.

inureyes:
@kyujin-cho  It does not work since application call returns 404.

image

inureyes (changes requested):
It is not working as intended. Needs some polishings.

inureyes (approved):
LGTM

Issue #1202, opened 2022-01-21
A wide blank is created when reducing the size of window.


Sujin-Kim1:
**Describe the bug**
When the window size is reduced and converted to a mini menu, a wide blank is created.

**To Reproduce**
You can see this error on every page if you reduce the size of the window.

**Expected behavior**
When converted to mini-ui, only margins of an appropriate size should exist on both sides of the content. Get rid of the big blanks!

**Screenshots**
<img src="https://user-images.githubusercontent.com/28584164/150536411-009404de-34b4-48a0-b928-9371e2f6ad78.png" width=500/>

**Client:**
- OS: macOS Monterey 12.4
- Browser version: Chrome
- Python version: 3.9.9
- NodeJS version: 16.13.1

studioego:
In my opinion, It needs to check the PR #1214 . 
If the issue confliction is changed, I'll request to review the PR #1214 
<img width="1946" alt="image" src="https://user-images.githubusercontent.com/20002/166182310-71df60c1-035d-402b-b00f-fa58a5dfec87.png">

studioego:
Following the PR #1214 , I checked to remoe wide blank.

However, the PR #1214 , There are unnecessary code changes.

I'll refect the code and re-submit the PR and close PR #1214 

<img width="1946" alt="image" src="https://user-images.githubusercontent.com/20002/166182701-11734507-0946-48b5-a5aa-dabafc46876f.png">

studioego:
As the PR https://github.com/lablup/backend.ai-webui/pull/1214 has already merged to the main branch. So, I close the issue.

PR #1214, opened 2022-01-27
fix: content-body width calculation

daminu:
Resolve Issue #1202 

#content-body's width was not calculated when screen size is smaller than 700px.
Added the width calculation when it's the case.


inureyes (approved):
LGTM

Issue #1168, opened 2021-12-03
Prevent to disappear the login dialog when the esc button is pressed.


Sujin-Kim1:
**Describe the bug**
On the login page, the login dialog disappears if the ESC button is pressed.

**To Reproduce**
1. Go to 'login page'
2. Press the ESC button
3. Then you can see the cluster dialog (in cloud)
![image](https://user-images.githubusercontent.com/28584164/144560048-47892bf2-f3a2-48e8-90df-64e639fa6714.png)
    - In the local development environment, the login dialog disappears immediately.
4. Press the ESC button once again.
5. Then the cluster dialog will disappear and you won't be able to see the login dialog until you refresh it.

**Expected behavior**
If press the ESC button, the cluster dialog disappears and shows the login dialog again.
Or, set the cluster dialog does not disappear when the ESC button is pressed.

**Screenshots**
![image](https://user-images.githubusercontent.com/28584164/144559159-cb557e82-028e-4ee8-b0bc-9153ed167650.png)

**Client:**
 - OS: MacOS
 - Browser version: Chrome, Safari, etc.
 - NodeJS version: 16.13.1

**Server:**
 - OS: [e.g. Ubuntu 18.04, CentOS 7.6, macOS 10.14]
 - Installation method: using cloud.backend.ai service, used scripts/install-dev.sh

PR #1219, opened 2022-02-03
fix: prevent esc key close login dialog

ggstone0523:
Resolve: lablup/backend.ai-webui #1168 

- This bug caused by auto closing backend-ai-dialog when press esc button
- add 'escapeKeyAction' attribute to mwc-dialog in backend-ai-dialog
- change 'escapeKeyAction' attribute value from 'close'  to '' only when open cluster dialog

**Currently, If you press the esc key right after the 'login page' opens, you can see the 'cluster dialog'.**
PR5_review0_1

### DIFF
--- a/src/components/backend-ai-dialog.ts
+++ b/src/components/backend-ai-dialog.ts
@@ -166,9 +166,11 @@ export default class BackendAiDialog extends LitElement {
 
   /**
    * Hide a dialog.
+   *
+   * @param {Boolean} isRecur - check if this function is called by eventhandler.
    */
-  hide() {
-    if (this.closeWithConfirmation) {
+  hide(isRecur = false) {
+    if (this.closeWithConfirmation && !isRecur) {
       const closeEvent = new CustomEvent('dialog-closing-confirm', {detail: ''});
       this.dispatchEvent(closeEvent);
     } else {
--- a/src/components/backend-ai-dialog.ts
+++ b/src/components/backend-ai-dialog.ts
@@ -43,6 +43,7 @@ export default class BackendAiDialog extends LitElement {
   @property({type: Boolean}) open = false;
   @property({type: String}) type = 'normal';
   @property({type: Boolean}) closeWithConfirmation = false;
+  @property({type: String}) escapeKeyAction = 'close';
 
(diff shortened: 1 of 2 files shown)


ggstone0523:

1.  show login dialog when press X button

Sujin-Kim1:
I think `escapeAction` is not necessary because it works well without that. 
It's trivial, but from the second time, if you press esc to make the login dialog disappear, the blank page will appear again.

ggstone0523:
In my development environment(safari, chrome),  cluster dialog disappear when press esc key if remove `escapeKeyAction`.
and, i avoid closing the login dialog with the esc key when the login dialog appears second time or press esc key late (because if press esc key late, cluster dialog did not appear)

ggstone0523:

1. prevent login dialog closed by esc key pressing
2. add 'preventEscKeyClose' attribute to backend-ai-dialog

Sujin-Kim1 (approved):
LGTM.

src/components/backend-ai-dialog.ts
```diff
@@ -160,14 +162,28 @@ export default class BackendAiDialog extends LitElement {
    * Open a dialog.
    */
   show() {
+    this.checkPreventEscapeKeyClose();
     this.dialog.show();
   }
```
lizable: Why don't we just use string value (`close`) as default value which is the same way the `mwc-dialog` supports?

lizable (changes requested):
I'm not sure about whether we should need boolean property instead of using the internal value that mwc-dialog supports.
IMHO, using escapeKeyAction is better since we don't need to assign an extra function for value checking.
What do you think of this approach @ggstone0523 @Sujin-Kim1 ?

ggstone0523:
I think is good!

ggstone0523:

- remove preventEscapeKeyClose attribute and append escapeKeyAction attribute in backend-ai-dialog

ggstone0523:
The revision was completed

- remove cluster dialog protection from disappear

ggstone0523:

- add 'escapeKeyAction' attribute to block-panel

lizable (approved):
LGTM.

PR #1225, opened 2022-02-13
feat: Add create/update ui for scheduler_opts

ggstone0523:
**Resolve: lablup/backend.ai-webui #1190**

- display scheduler_opts's each key and value to table.
PR6_review0_1

- scheduler_opts UI in scaling group create operation dialog.
PR6_review0_2

- scheduler_opts UI in scaling group update operation dialog.
PR6_review0_3

- check scheduler_opts's key-value valid
PR6_review0_4


ggstone0523:

**feature**

- Add dialog for scheduler_opts and driver_opts
PR6_review1_1

- Change set scheduler options dialog
PR6_review1_2

- validation check
PR6_review1_3

**bugfix**

- fix allowed_session_value didn't change

src/lib/backend.ai-client-node.ts
```diff
@@ -3398,14 +3398,14 @@ class ScalingGroup {
...
-  async create(name, description = "", wsproxyAddress = null) {
+  async create(name, description = "", seheduler_opts = "{}", wsproxyAddress = null) {
     const input = {
       description: description,
       is_active: true,
       driver: "static",
       scheduler: "fifo",
       driver_opts: "{}",
-      scheduler_opts: "{}"
+      scheduler_opts: seheduler_opts
```
lizable: This might be a typo error, perhaps?

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -576,6 +772,56 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+              slot="footer"
+              icon="check"
+              label="${_t('button.Save')}"
+              @click="${()=>this.modifySchedulerOpts()}"
+              style="width:100%"></mwc-button>
+        </div>
+      </backend-ai-dialog>
+      <backend-ai-dialog id="show-detail-dialog" fixed backdrop persistent>
+        <span slot="title">${_t('agent.DetailedInformation')}</span>
+        <div slot="content" id="modify-scheduler-opts-container">
```
lizable: How about using UI layout based on "keypair detail dialog"?
And please refer to key, value based UI Layout [here](https://design.mindsphere.io/patterns/key-value.html).

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -576,6 +772,56 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
        </div>
       </backend-ai-dialog>
+      <backend-ai-dialog id="modify-scheduler-opts-dialog" fixed backdrop persistent>
+        <span slot="title">${_t('resourceGroup.SetSchedulerOptions')}</span>
+        <div slot="content" id="modify-scheduler-opts-container">
+          <mwc-select
+          id="allowed-session-types"
+          label="allowed_session_types"
+          required>
+            <mwc-list-item style="height:auto;" value="interactive">interactive</mwc-list-item>
```
lizable: Please check a CSS variable that manages the height of `mwc-list-item`, instead of using inline height in every HTML element.
You can see in [Styled section in README file of mwc-select](https://github.com/material-components/material-web/tree/master/packages/select).

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -500,6 +678,15 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
             label="${_t('resourceGroup.WsproxyAddress')}"
             placeholder="http://localhost:10200"
           ></mwc-textfield>
```
lizable: I'm not sure about using "nested dialog" here. Could you explain the reason for this layout?

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -438,7 +617,10 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
               id="add-scaling-group"
               icon="add"
               label="${_t('button.Add')}"
-              @click=${() => this._launchDialogById('#create-scaling-group-dialog')}>
+              @click=${() => {
+    this._clearValues();
```
lizable: Any reasons for using `_clearValues()`?
If you want to restore the values to the previous value, then we should change the name of functions to more proper terms.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -576,6 +772,56 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+        <div slot="content" id="modify-scheduler-opts-container">
+          <mwc-select
+          id="allowed-session-types"
+          label="allowed_session_types"
+          required>
+            <mwc-list-item style="height:auto;" value="interactive">interactive</mwc-list-item>
+            <mwc-list-item style="height:auto;" value="batch">batch</mwc-list-item>
+            <mwc-list-item style="height:auto;" value="both">both</mwc-list-item>
+          </mwc-select>
+          <mwc-textfield
```
lizable: Three things to ask in this section:
1. I'm not sure `pending timeout` is the required value.
2. And Let's replace `_(underbar)` in the label to whitespace since is a bit unnatural.
3. Let's use change event instead of input event, since `input event` is triggered much more frequently compared to "change event". Please refer to [this link](https://stackoverflow.com/questions/17047497/difference-between-change-and-input-event-for-an-input-element).

lizable (changes requested):
I have several things to request. Please check out the comments below.
Some of them need your explanations.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: Yes, 
 As the scheduler option is added, a new input form for the option should be added, and if these input forms are not separated through dialog, the length of the create/modify dialog will become longer and I think less readable each time the option is added.

lizable: Hmm, but the number of keys may not exceed 5 or 6, so I would like to suggest accordion UI instead of nested dialog.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: because to reset scheduler options's value to default when user open 'create dialog'.
and i changed name '_clearValues' to '_resetAllValues'

src/components/backend-ai-scaling-group-list.ts
ggstone0523: 1. The reason why I put pending_timeout was because pending_timeout came into thisPR(https://github.com/lablup/backend.ai-manager/pull/527) as a new scheduler_opts.
2. I revise it following comment.
3. I revise it following comment.

src/lib/backend.ai-client-node.ts
ggstone0523: yes it was a typo error, so i revise it to 'scheduler_opts'

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following comment.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following 'Key above value' key value pair types.

**image**
PR6_review2_1

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following comment.

**image**
PR6_review2_2

src/components/backend-ai-scaling-group-list.ts
lizable: Please check again about "keypair detail dialog" and update the UI according to that.

ggstone0523:

1. change resource group detail dialog title name from 'Detailed Information' to 'Resource Group Detail'
and change resource group detail dialog UI according to keypair detail dialog UI.

**image**
PR6_review3_1

2. prevent create/modify dialog scroll 
and check that if input is maximum length the description component display all input.

**image**
PR6_review3_2
PR6_review3_3

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -543,6 +783,29 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+            <span slot="title">${_t('resourceGroup.SchedulerOptions')}</span>
+            <mwc-select
+            id="modify-allowed-session-types"
+            label="allowed session types"
+            required>
+              <mwc-list-item value="interactive">interactive</mwc-list-item>
+              <mwc-list-item value="batch">batch</mwc-list-item>
+              <mwc-list-item value="both">both (interactive, batch)</mwc-list-item>
+            </mwc-select>
+            <mwc-textfield
```
lizable: Why would we need to use "text" type in the input field that only allows "number"?
It would bring redundant validation which can be handled by just assigning the type of input field to "number".
Please check the README file in `mwc-textfield` and set it with `min` value.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -427,6 +553,99 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+    }
+    return isValid;
+  }
+
+  /**
+   * save schedulerOptsDialog value to schedulerOpts
+   *
+   * @param {Element} allowedSessionType - 'allowed session type' input form element in create or modify dialog
+   * @param {Element} pendingTimeout - 'pending timeout' input form element in create or modify dialog
+   * */
```
lizable: Could you change this function without any args?
I guess we can assign the value inside the function.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -427,6 +553,99 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+  }
+
+  /**
+   * verify schedulerOptions key and value
+   *
+   * @param {Element} allowedSessionType - 'allowed session type' input form element in create or modify dialog
+   * @param {Element} pendingTimeout - 'pending timeout' input form element in create or modify dialog
+   *
+   * @return {Boolean} key-value is valid => true, key-value is invalid => false
+   * */
```
lizable: Please remove args, It doesn't seem to be "essential" in here.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -427,6 +553,99 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+  }
+
+  /**
+   * verify and save schedulerOptions
+   *
+   * @param {String} allowedSessionTypeName - 'allowed session type' input form element class name string for querySelector
+   * @param {String} pendingTimeoutName - 'pending timeout' input form element class name string for querySelector
+   *
+   * @return {Boolean} key-value is valid => true, key-value is invalid => false
+   * */
```
lizable: Please remove the args in this function, too.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -427,6 +553,99 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+   * @param {String} name - scheduler option key in selected scalingGroup
+   * @param {Any} value - scheduler option value in selected scalingGroup
+   */
+  _initializeSchedulerOpts(name = '', value = '') {
+    const allowedSessionType = this.shadowRoot.querySelector('#modify-allowed-session-types');
+    const pendingTimeout = this.shadowRoot.querySelector('#modify-pending-timeout');
+
+    switch (name) {
+    case 'allowed_session_types':
+      if (value.length === 2) {
```
lizable: Let's not use magic number in here ("2"), you need to change that number to more descriptive variable, instead.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -254,12 +297,23 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+  }}
+      ><wl-icon>assignment</wl-icon></wl-button>
           <wl-button fab flat inverted
             icon="settings"
             class="fg blue"
             @click=${() => {
     this.selectedIndex = rowData.index;
     this.shadowRoot.querySelector('#modify-scaling-group-active').selected = this.scalingGroups[rowData.index].is_active;
+    Object.entries(JSON.parse(this.scalingGroups[this.selectedIndex].scheduler_opts)).map((item: any, index) => {
+      this._initializeSchedulerOpts(item[0], item[1]);
```
lizable: Any reason for adding args in this way?

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -275,6 +329,72 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+      if (item[0] === 'allowed_session_types') {
+        return _schedulerKeyValueTemplate(item[0].replaceAll('_', ' '), item[1].join(', '));
+      } else if (item[0] === 'pending_timeout') {
+        return _schedulerKeyValueTemplate(item[0].replaceAll('_', ' '), item[1] + ' ' + _text('resourceGroup.TimeoutSeconds'));
+      } else {
+        return '';
+      }
+    });
+  }
+
```
lizable: This function also needs to be reconsidered when you change the type of corresponding input field from "text" to "number".

lizable: For now, I don't think it's necessary.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -275,6 +329,72 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
     );
   }
```
lizable: Let's remove this function since we don't provide modification of "driver option" yet.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -438,7 +657,10 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
               id="add-scaling-group"
               icon="add"
               label="${_t('button.Add')}"
-              @click=${() => this._launchDialogById('#create-scaling-group-dialog')}>
+              @click=${() => {
+    this._resetAllValues();
```
lizable: Let's create a wrapping function for better readability.

src/components/backend-ai-scaling-group-list.ts
lizable: And also I found there's a typo error in the function name. Please change it to `SchedulerOpts` instead of using `SchulerOpts`.

src/components/backend-ai-scaling-group-list.ts
lizable: Also, I'm not sure about doing extra operation in "save" function. 
Could we just call `verifySchedulerOpts` outside in the save function so that save function only called when all value is valid?
It seems this function does too many things inside compared to it's name.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -543,6 +783,29 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
             label="${_t('resourceGroup.WsproxyAddress')}"
             value=${this.scalingGroups.length === 0 ? '' : this.scalingGroups[this.selectedIndex].wsproxy_addr ?? ''}
           ></mwc-textfield>
+          <wl-expansion id="modify-scheduler-options-input-form">
+            <span slot="title">${_t('resourceGroup.SchedulerOptions')}</span>
+            <mwc-select
+            id="modify-allowed-session-types"
+            label="allowed session types"
+            required>
+              <mwc-list-item value="interactive">interactive</mwc-list-item>
```
lizable: Let's use the option as a component property instead of using raw text.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -576,6 +839,25 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
             </mwc-button>
        </div>
       </backend-ai-dialog>
+      <backend-ai-dialog id="keypair-detail-dialog" fixed backdrop blockscrolling>
```
lizable: Is this element id supposed to have "keypair-detail-dialog"?

lizable: And most of detail dialog contains not only for information that didn't cover in table UI, but it also contains other essential information. Please check out user detail dialog or vfolder detail dialog. You may get a clue from there.

lizable (changes requested):
@ggstone0523 
Before you ask for a PR review, I have a few things to request.
- Please check the "typo" and fix it.
- Please avoid using extra args in function unless it's necessary.
- Please create the function and element ID with a certain pattern for better readability.
- If you are struggling with the layouts, referring to other dialog would be welcomed.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following comment.

ggstone0523: I revise it following comment.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following comment.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following comment.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following comment.

src/components/backend-ai-scaling-group-list.ts
ggstone0523: I revise it following comment.

ggstone0523:

- Using min value in mwc-textfield for validate scheduler opts input forms value
- Remove unnecessary argument in function
- Divide save function to verify function and save function
- Change `schulerOpts` to `schedulerOpts`
- Using Array.prototype.includes() to remove magic number
- Change way that adding args to _setSchedulerOptsInputForm bacause improve readability
- Remove pending timeout validator
- Remove drive option renderer
- Add wrapping function for add button click event
- replace raw text to property in 'allowed session types' scheduler opts input forms
- change resource-group-detail-dialog UI
**image**
PR6_review4_1

src/components/backend-ai-scaling-group-list.ts
ggstone0523: because i change 'object in object' type to 'list in list' type to use map.
so, i approach key and value using list index. 
but, this approach is reduce readablity. so i change this logic

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -254,12 +311,23 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
           id="controls"
           class="layout horizontal flex center"
         >
+        <wl-button fab flat inverted
+          icon="device_hub"
+          class="fg green"
+          @click=${() => {
```
lizable: Let's wrap it with a function instead of stacking it all in the event handler.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -254,12 +311,23 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+    this.selectedIndex = rowData.index;
+    this._launchDialogById('#resource-group-detail-dialog');
+  }}
+      ><wl-icon>assignment</wl-icon></wl-button>
           <wl-button fab flat inverted
             icon="settings"
             class="fg blue"
             @click=${() => {
```
lizable: Same in here.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -275,6 +343,27 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+  _schedulerOptionsRenderer() {
+    const scheduler_opts = Object.entries(JSON.parse(this.scalingGroups[this.selectedIndex].scheduler_opts));
+    const _schedulerKeyValueTemplate = (key, value) => {
+      return html`
+      <vaadin-item>
+        <div><strong>${key}</strong></div>
+        <div class="scheduler-option-value">${value}</div>
+      </vaadin-item>`;
+    };
+
```
lizable: Could we just remove the extra renderer(_schedulerOptionsRenderer) and put all of HTML into render function? I think it's too much to display a single dialog information.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -427,6 +524,112 @@ export default class BackendAIScalingGroupList extends BackendAIPage {
...
+  /**
+   * set selected scaling group's values to scheduler option input form in modify dialog.
+   *
+   * @param {String} name - scheduler option key in selected scalingGroup
+   * @param {Any} value - scheduler option value in selected scalingGroup
+   * */
+  _initializeModifySchedulerOpts(name = '', value: any) {
+    const allowedSessionType = this.shadowRoot.querySelector('#modify-allowed-session-types');
+    const pendingTimeout = this.shadowRoot.querySelector('#modify-pending-timeout');
+
```
lizable: Let's not use another condition statement in switch statement.

lizable (changes requested):

ggstone0523:

- apply wrapping function for control panel buttons
- remove switch statement in '_initializeModifySchedulerOpts' function and change it to if statement
- remove '_schedulerOptionsRenderer' function and put all to render function
- remove required in pending_timeout input form

lizable (commented):
I fixed a few minor linting errors and got rid of some legacy codes.
We need a review and approval for this PR from one of the maintainers, @inureyes or @adrysn.

src/lib/backend.ai-client-node.ts
```diff
@@ -3398,14 +3398,14 @@ class ScalingGroup {
    * @param {string} description - Scaling group description
    * @param {string} wsproxyAddress - wsproxy url (NEW in manager 21.09)
    */
-  async create(name, description = "", wsproxyAddress = null) {
+  async create(name, description = "", scheduler_opts = "{}", wsproxyAddress = null) {
```
inureyes: The "order" of parameter should be kept as original (for API compatibility / legacy support). scheduler_opts will be the end of parameter.

src/lib/backend.ai-client-node.js
```diff
@@ -3168,14 +3168,14 @@ class ScalingGroup {
      * @param {string} description - Scaling group description
      * @param {string} wsproxyAddress - wsproxy url (NEW in manager 21.09)
      */
-    async create(name, description = "", wsproxyAddress = null) {
+    async create(name, description = "", scheduler_opts = "{}", wsproxyAddress = null) {
```
inureyes: The "order" of parameter should be kept as original (for API compatibility / legacy support). scheduler_opts will be the end of parameter.

inureyes (changes requested):
Check the comments and related calls on backend-ai-scaling-group-list.ts

ggstone0523:

- change order of create function's parameter (scheduler_opts go to end of parameter)

lizable:
I changed several things in order to keep the version compatibility and also the identity of terms used, especially in "scaling group".

- any `scaling group` terms used in WebUI component will be changed to `resource group`.
- since the milestone of the related issue([backend.ai#343](https://github.com/lablup/backend.ai/issues/343)) in Backend.AI component is `22.03`, configuring scheduler option will be supported from `22.03`.

inureyes (approved):
Goot to go

PR #1228, opened 2022-02-15
style: make endpoint resizable and unify heights of vaadin-grid

Sujin-Kim1:
- all endpoints in the agent page can be checked by adding the resizable attribute.
- Match `vaadin-grid` height in resource group tab to `vaadin-grid` in same page, different tab.

[Screenshot]
**Before**
image
image

**After**
image
👆 resizable endpoint

image


inureyes (approved):
LGTM

PR #1235, opened 2022-02-19
feature: add support for multi-architecture image

kyujin-cho:


src/lib/backend.ai-client-node.ts
```diff
@@ -1060,7 +1062,7 @@ class Client {
 
   // legacy aliases (DO NOT USE for new codes)
   createKernel(kernelType, sessionId = undefined, resources = {}, timeout = 0) {
-    return this.createIfNotExists(kernelType, sessionId, resources, timeout);
+    return this.createIfNotExists(kernelType, sessionId, 'x86_64', resources, timeout);
```
lizable: I'm not sure whether it's a good idea to use hard-coded `x86_64` value here.
How about adding key-value in config file instead?

src/lib/backend.ai-client-node.js
```diff
@@ -969,7 +971,7 @@ class Client {
     }
     // legacy aliases (DO NOT USE for new codes)
     createKernel(kernelType, sessionId = undefined, resources = {}, timeout = 0) {
-        return this.createIfNotExists(kernelType, sessionId, resources, timeout);
+        return this.createIfNotExists(kernelType, sessionId, 'x86_64', resources, timeout);
     }
```
lizable, line 975: Same here.

lizable (changes requested):
It works okay, but I have one thing to suggest.
Please check the comment below.

src/lib/backend.ai-client-node.ts
inureyes: Also, the parameter order should be kept for compatibility. @kyujin-cho

inureyes: Fixed.

inureyes: @lizable It is for legacy code. So only x86_64 will come into the role.

inureyes (approved):
LGTM

PR #1237, opened 2022-02-20
fix: activate superadmin's all vfolder listing feature

SonMinWoo:
resolve: https://github.com/lablup/backend.ai/issues/321
related: https://github.com/lablup/backend.ai-manager/pull/544

**When I tested, the function is working.
Trying with the admin accesskey, I get an error. Trying with the folder owner accesskey it succeeds.**

issue321_1

**But in webui, superadmin can't know all vfolders. This is because superadmin's all folder view function did not work.**
It was already implemented in the backend. One parameter required by the backend was missing, so add it.

image

It works well. (The two folders in the picture below are folders created by different accounts as shown in the picture above.)
image

To reduce the load, it request only one-time when superadmin check delegated session start checkbox.
issue321-2


src/components/backend-ai-resource-broker.ts
```diff
@@ -363,11 +363,11 @@ export default class BackendAiResourceBroker extends BackendAIPage {
    * Update virtual folder list. Also divide automount folders from general ones.
    *
    */
-  async updateVirtualFolderList() {
+   async updateVirtualFolderList(userEmail = false) {
```
lizable: The type of default value of parameter `userEmail` doesn't seem to have a match with SDK function in [here](https://github.com/lablup/backend.ai-webui/pull/1237/files#diff-6ab6a3b84b5af19510d01f3a5c155018416e21b94b7090c2a4a22e0a1d68e33bR1421-R1429).
It's string, not boolean.

src/components/backend-ai-session-launcher.ts
```diff
@@ -2284,6 +2284,20 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
     }
   }
 
+  async _fetchDelegatedSessionVfolder() {
+    const ownerEnabled = this.shadowRoot.querySelector('#owner-enabled');
+    const userEmail = this.shadowRoot.querySelector('#owner-email').value;
+    if (ownerEnabled && ownerEnabled.checked) {
+      await this.resourceBroker.updateVirtualFolderList(userEmail).then(() => {
```
lizable: Any reason for using `await` keyword and `.then()` keyword at the same time?

src/components/backend-ai-session-launcher.ts
```diff
@@ -3064,6 +3078,55 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
...
                 `}
             </div>
+              <wl-expansion name="ownership" style="--expansion-content-padding:15px 0;">
+              <span slot="title">${_t('session.launcher.SetSessionOwner')}</span>
+              <div class="vertical layout">
+                <div class="horizontal center layout">
+                  <mwc-textfield id="owner-email" type="email" class="flex" value=""
+                                pattern="^.+@.+\..+$" icon="mail"
+                                label="${_text('session.launcher.OwnerEmail')}" size="40"></mwc-textfield>
+                  <mwc-icon-button icon="refresh" class="blue"
```
lizable, line 3188: If I input any user account(email) and rewrite it, it doesn't update according to the new input.

src/lib/backend.ai-client-node.ts
```diff
@@ -1540,10 +1540,15 @@ class VFolder {
   /**
    * List Virtual folders that requested accessKey has permission to.
    */
-  async list(groupId = null) {
+  async list(groupId = null, userEmail = null) {
     let reqUrl = this.urlPrefix;
-    if (groupId) {
-      const params = {group_id: groupId};
+    if (groupId || userEmail) {
```
lizable: I think it's better to use condition statements separately.

lizable (changes requested):
I have a few things to request. Please refer to the comments below.

lizable (approved):
LGTM

lizable:
I just updated two things during delegate session creation:
- [x] Enable the session delegation checkbox only if the owner's keypair is valid.
- [x] Fetch the owner's accessible vfolder list only if the owner's keypair is valid.

inureyes:
@lizable Great!

inureyes (approved):

PR #1250, opened 2022-03-03
Feature: masking user info

### LINKED ISSUE #1251: An option to mask user's information
In a security-critical environment, user information such as email and user's full name should be hidden. To deal with such environments, let's provide an option to mask the information.

Option name might be `maskUserInformation`.
When the option is `true`, hide the user's full name and email in every location.

lizable:
This PR provides a new feature, which is string masking in order to fulfill privacy policy.
It masks not only the current login user but also the information of other users, which was only visible to those users who have `admin` or `super-admin` roles as well.

This resolves #1251.

### DIFF
--- a/config.toml.sample
+++ b/config.toml.sample
@@ -13,6 +13,7 @@ allowProjectResourceMonitor = true
 allowSignupWithoutConfirmation = false
 autoLogout = false
 debug = false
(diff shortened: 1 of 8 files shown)


adrysn (approved):
LGTM.

PR #1252, opened 2022-03-04
Fix: replace `is_active` column in user list to `status`

lizable:
This PR updates `is_active` column to `status`, which has been already changed in manager component.
Ref: [backend.ai-manager](https://github.com/lablup/backend.ai-manager/blob/main/src/ai/backend/manager/models/alembic/versions/0d553d59f369_users_replace_is_active_to_status_and_its_info.py)


src/components/backend-ai-user-list.ts
```diff
@@ -489,7 +482,7 @@ export default class BackendAIUserList extends BackendAIPage {
     const confirmEl = this.shadowRoot.querySelector('#confirm');
     const confirm = confirmEl.value;
     const description = this.shadowRoot.querySelector('#description').value;
-    const is_active = this.shadowRoot.querySelector('#is_active').selected;
+    const status = this.shadowRoot.querySelector('#status').selected ? 'active': 'inactive';
```
adrysn: Settings the status as `inactive` will retrieve `inactive` users only, but not users with the statuses of `deleted` or `before-verification`. Please refer to the manager's implementation:
https://github.com/lablup/backend.ai-manager/blob/2e3824e0a8eebf9165c44e0fa616fdc4c2ed7c26/src/ai/backend/manager/models/user.py#L359-L363

Is this intended behavior? If not, keeping `is_active` instead of `status` might be proper.

adrysn (commented):
This PR will not fetch users with the statuses of `deleted` or `before-verifacation`. If this is not the intended behavior, it should be fixed.

src/components/backend-ai-user-list.ts
lizable: I agree with your opinion. I changed the field to `is_active` instead of using `status`.

lizable:
In order to maintain user management in control-panel instead of WebUI, I allowed the status of the user to be only two statuses, `active` and `inactive`.

PR #1254, opened 2022-03-08
fix: get wsproxy version

Sujin-Kim1:
The invalid parameter is used to get wsproxy version. So I got the below error when opening any app from the session.
MicrosoftTeams-image

So I changed the parameter from `version` to `wsproxy_version`.


inureyes (approved):
LGTM

adrysn (approved):
LGTM.

PR #1259, opened 2022-03-29
Refactor: reformat legacy mounts information on each session list

lizable:
Related: [backend.ai-manager#537](https://github.com/lablup/backend.ai-manager/pull/537)

Before significant changes occurred in mounts column information in session list when receiving requests from manager server, mounts information always contain information that exactly corresponded to mounts column in kernels.
Since there have been updates in manager server, the column name remains but the format of the result has been changed.

## Before (legacy)
If user mounts "**test-folder**" and "**test-data**" in a session
then, response data in "mount" key contains data like this:
`[ 
     "['test-folder', 'storagehost:name', 'actual-vfolder-path', 'permission', 'kernel-path']",
     "['test-data', 'storagehost:name', 'actual-vfolder-path', 'permission', 'kernel-path']",
 ]`
 
 ## After applying updates in [manager server](https://github.com/lablup/backend.ai-manager/pull/537)
 If user mounts "**test-folder**" and "**test-data**" in a session
 then, response data in "mounts" key contains data like this:
 `['test-folder', 'test-data']`

To reflect those two different data format when receiving session list, I added single conditional statement to filter legacy mounts data format and reformat into current format. This PR will allow seamless session list receiving regardless of DB schema on kernels.

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -1184,8 +1184,6 @@ export default class BackendAiSessionList extends BackendAIPage {
   _createMountedFolderDropdown(e, mounts) {
     const menuButton: HTMLElement = e.target;
     const menu = document.createElement('mwc-menu') as any;
-    const regExp = /[[\],'"]/g;
-
     menu.anchor = menuButton;
     menu.className = 'dropdown-menu';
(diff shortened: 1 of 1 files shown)


adrysn (approved):
LGTM.

PR #1262, opened 2022-04-05
refactor: show path dialog only when clicking the icon

Sujin-Kim1:
For better user experience, show Path & Alias only when the user clicks the info icon.
I also hide the checkbox by commenting out the `setPathContent` function.
[Before]
image
[After]
image


lizable (approved):
LGTM.

PR #1264, opened 2022-04-05
Failing to update a compute session's name will set the value of edit element to `undefined`

### LINKED ISSUE #1260: Failing to update a compute session's name will set the value of edit element to `undefined`
Updating a compute session's name with a invalid string, such as `invalid-trailing-hyphen-` will raise a 400 Bad Request error.

image
image

After that, Clicking the edit button again will set the value of the input field to `undefined`. Refreshing the browser will resolve the issue, but this needs to be fixed for smoother user experience.

image

Also, when the length of the input element overflows the column width, the update button element is hidden which confuses user.

Tasks to be done:
- [ ] Better error message when user entered an invalid session name.
- [x] Correctly set the input element's value to its original name, not filling it with `undefined`.
- [x] Make the update element visible even in narrow-width browser. Perhaps, moving the element before the input element would be one solution.
- [x] Hide the session name edit button for other user's sessions. Or, do we need to make an admin can edit other user's session name?

Sujin-Kim1:
this resolves #1260

### DIFF
--- a/resources/i18n/ko.json
+++ b/resources/i18n/ko.json
@@ -292,9 +292,12 @@
     "OnlyOneFolderAttached": "하나의 폴더만 연결되어 있습니다.",
     "NoLogMsgAvailable": "현재 표시할 수 있는 로그 메시지가 없습니다.",
     "SessionRenamed": "세션 이름이 변경되었습니다",
-    "SessionNameRequired": "세션 이름을 입력해주세요",
-    "EnterValidSessionName": "올바른 세션 이름을 입력해주세요",
-    "SessionNameAlreadyExist": "같은 이름의 세션이 이미 있습니다"
+    "Validation": {
+      "SessionNameRequired": "세션 이름을 입력해주세요",
+      "LetterNumber-_dot": "문자, 숫자 및 -_. 만 허용됩니다",
+      "EnterValidSessionName": "올바른 세션 이름을 입력해주세요",
+      "SessionNameAlreadyExist": "같은 이름의 세션이 이미 있습니다"
+    }
   },
   "button": {
     "Cancel": "취소",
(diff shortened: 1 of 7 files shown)


adrysn (changes requested):
Overall, this PR improves the interaction with the rename field. However, I find that the validation checker sometimes displays inappropriate messages. Example images below:

image
image
image

If it is hard to display in every situation, I think just displaying `slugged strings (>3 chars)` might be an alternative option.

resources/i18n/ko.json
```diff
@@ -292,9 +292,12 @@
     "OnlyOneFolderAttached": "하나의 폴더만 연결되어 있습니다.",
     "NoLogMsgAvailable": "현재 표시할 수 있는 로그 메시지가 없습니다.",
     "SessionRenamed": "세션 이름이 변경되었습니다",
-    "SessionNameRequired": "세션 이름을 입력해주세요",
-    "EnterValidSessionName": "올바른 세션 이름을 입력해주세요",
-    "SessionNameAlreadyExist": "같은 이름의 세션이 이미 있습니다"
+    "Validation": {
+      "SessionNameRequired": "세션 이름을 입력해주세요",
+      "SluggedStrings": "고유문자열로 입력해주세요 (3 > 글자 수)",
```
lizable: How about just translating the number of minimum characters to 최소 4자 이상 instead of (3 > 글자수)?

lizable (changes requested):
It works fine. 
But IMHO, It would be nicer if there's a bit of touch-up on translation.
Please check out the review below.

resources/i18n/ko.json
agatha197: That's great. I fixed it right away.

lizable (approved):
LGTM.

inureyes (commented):
LGTM

PR #1277, opened 2022-04-19
Fix: allocate resource according to selected resource preset

lizable:
## Bug Description
There have been several annoying issues lingering on the resource allocation panel in session launcher, which degrades user experience. Please refer to the content below.
Issue 1. User cannot Input any value on textfield area of each resource section.
Screen Recording 2022-04-20 at 12 53 44 AM mov

Issue 2. Minimum value of each resource changes (e.g. the min value of cpu-resource was `1` but it changed to `0`)
Screen Recording 2022-04-20 at 12 45 31 AM mov

Issue 3. Selected resource preset doesn't match actual allocation and knob in every slider doesn't move even resource preset is selected.
Screen Shot 2022-04-20 at 12 46 02 AM

Issue 4. The number of containers in the total allocation section doesn't match the actual allocation.

Issue 5. If maximum number of container count in each session is zero, then it breaks the condition.

## Solution Description
- Only trigger value changes when the previous value is not null and it exceeds the minimum value.
- Set the minimum value of the resource to non-zero so that the user cannot accidentally allocate zero value on any resource. If there's no resource to allocate, then set the current value to min value, and disable the resource allocation section.
- Refactor the number of containers in the total allocation section to count not only cluster size but also the number of sessions to allocate.
- Set "0" value as an "infinite" value (Big number) in the resource policy setting dialog.

This resolves https://github.com/lablup/backend.ai-webui/issues/1292.

### DIFF
--- a/src/components/backend-ai-resource-policy-list.ts
+++ b/src/components/backend-ai-resource-policy-list.ts
@@ -31,6 +31,11 @@ import {default as PainKiller} from './backend-ai-painkiller';
 import {BackendAiStyles} from './backend-ai-general-styles';
 import {IronFlex, IronFlexAlignment} from '../plastics/layout/iron-flex-layout-classes';
 
+class BigNumber {
+  static getValue() {
+    return 1000000;
+  }
+}
 @customElement('backend-ai-resource-policy-list')
 export default class BackendAIResourcePolicyList extends BackendAIPage {
   @property({type: Boolean}) visible = false;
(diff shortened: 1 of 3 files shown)


Sujin-Kim1:
Thanks to you, many things have improved! But I found some errors like below.
1. Sometimes the ram and shared memory sliders move together. 

https://user-images.githubusercontent.com/28584164/164359518-86cd5285-3618-45b0-afbc-35529b192ef9.mov

2. After moving the slider or pressing the resize arrow button, I can't type as I want. So when users click on a textfield, it would be better to change it to focus or resolve it in a different way.

https://user-images.githubusercontent.com/28584164/164363791-27e98e32-2fac-4a53-9809-6489ec7e2a94.mov

lizable:
At the suggestion of @Sujin-Kim1  and @inureyes, I updated the shared memory allocation section to only change the value itself, not manipulating the maximum value when the current RAM allocation value changes.
Screen Recording 2022-04-22 at 3 43 30 PM mov

adrysn (approved):
I updated some parts of this PR, and looks good for now. I need double checks from others.

PR #1281, opened 2022-04-25
add: import weightless textfield to summary view

Sujin-Kim1:
Unable to set environment because weightless/textfield was not imported.
image


inureyes (approved):
LGTM

Issue #1111, opened 2021-08-24
Add margins to policy dialog to distinguish fields


Sujin-Kim1:
**Is your feature request related to a problem? Please describe.**
It is quiet hard to distinguish fields of create/modify policy dialog. (Resource policy, Folder)
<img src="https://user-images.githubusercontent.com/28584164/130578283-c959bb1d-0752-4924-a711-9d684c69e81a.png" height=500>  
<img src="https://user-images.githubusercontent.com/28584164/130577896-a32593dd-2816-439c-8268-d94d6f4e5cb7.png" height=500>

**Describe the solution you'd like**
Please add margins to fields of `new-policy-dialog` and `modify-policy-dialog` to distinguish each field.

**Describe alternatives you've considered**
혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.

**Additional context**
On the `Users` page, click the `RESOURCE POLICIES` tab. And click the `CREATE POLICY` button or setting icon. Then you can see the dialog what you want.

studioego:
Do these come from "User> Management" ( `http://***/credential` )menu's "Resource Policy" tab"?

<img width="748" alt="image" src="https://user-images.githubusercontent.com/20002/165072737-d9e29ea8-271d-401d-8fa7-0d4d605cb2c1.png">
<img width="909" alt="스크린샷 2022-04-25 오후 7 35 04" src="https://user-images.githubusercontent.com/20002/165072826-723196e6-9010-4cdc-a46c-fc0c88265dbf.png">

PR #1282, opened 2022-04-25
Add margins to policy dialog to distinguish fields

studioego:
Add margins to policy dialog to distinguish fields
Adjusted some element CSS margins.

close https://github.com/lablup/backend.ai-webui/issues/1111


Sujin-Kim1 (changes requested):
These inner style codes are duplicated. It would be better to use `class`.

Sujin-Kim1 (approved):
LGTM.

adrysn (approved):
LGTM.

Issue #1109, opened 2021-08-24
Enhance design of folder-explorer-dialog


Sujin-Kim1:
**Describe the solution you'd like**
It would be better to center the buttons on the top right, and remove the arrow on the left, or add padding.
![image](https://user-images.githubusercontent.com/28584164/130574636-99ee37a4-0ab6-4848-a473-b46dec1aa5c1.png)

**Describe alternatives you've considered**
If you have any ideas to make this dialog look better other than what I mentioned, feel free to try!

**Additional context**
Click the folder icon on the `Data & Storage` page, then you can see the `folder-explorer-dialog`. (You need to create the folder in advance to see it.)

studioego:
First, I add one div css  position & left elements.
<img width="1427" alt="image" src="https://user-images.githubusercontent.com/20002/165251876-cbb99dbb-3977-4902-85bc-47d7ea13af5d.png">

studioego:
Second, add margin-left attribute
<img width="1427" alt="image" src="https://user-images.githubusercontent.com/20002/165254764-2b6a7fbf-4840-4cb2-ac00-35e1aab937c4.png">

PR #1283, opened 2022-04-26
Enhance design of folder-explorer-dialog

studioego:
Enhance design of folder-explorer-dialog
1. move the buttons(file browser, file upload, new directory) on the top right to the top center.
2. add margin-left for arrow image.

https://github.com/lablup/backend.ai-webui/issues/1109

### DIFF
--- a/src/components/backend-ai-storage-list.ts
+++ b/src/components/backend-ai-storage-list.ts
@@ -452,10 +452,6 @@ export default class BackendAiStorageList extends BackendAIPage {
           margin:15px 10px;
         }
 
-        div.align-top-center {
-          position: relative;
-        }
-
         @media screen and (max-width: 700px) {
           #folder-explorer-dialog,
           #folder-explorer-dialog.mini_ui {
@@ -485,11 +481,6 @@ export default class BackendAiStorageList extends BackendAIPage {
           #modify-permission-dialog {
             --component-min-width: 100%;
           }
-
-          div.align-top-center {
-            position: relative;
-            right: unset;
-          }
         }
 
         @media screen and (min-width: 900px) {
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-storage-list.ts
```diff
@@ -451,6 +452,11 @@ export default class BackendAiStorageList extends BackendAIPage {
           margin:15px 10px;
         }
 
+        div.align-top-center {
+          position: absolute;
```
adrysn: Setting the absolute position on the action buttons container will overlap it with the close button in a narrow screen. It might be better to use the flex layout to center the buttons.
image

I'm not quite sure this layout is better than before, even on a wider screen. But, I don't have a good feel for the layout and design. I'll follow @Sujin-Kim1 and @lizable's opinions.
image

src/components/backend-ai-storage-list.ts
```diff
@@ -285,6 +285,7 @@ export default class BackendAiStorageList extends BackendAIPage {
           color: #637282;
           font-size: 1em;
           margin-bottom: 10px;
+          margin-left: 20px;
```
adrysn, line 288: This part is good.

src/components/backend-ai-storage-list.ts
agatha197: I'm sorry to reverse my initial offer, but it looks good to leave the button in the top right corner. Instead, how about aligning each button's icon and text to be vertical-align center and placing it in the center of the x button?

studioego: I'll try to work on your guide after Lunchtime.

studioego: 1. width more 900px 
 
image

2. width less 900px

image


However, according to some languages, broken layouts... oops.
So I add calculate class size and manupulate dynamic class size;

studioego: Hmm, Unlike English, Indonesian(Bahasa Indonesia), Korean, For French, some menu button word length is too long, So It can only be centered at least 1400px or higher
* French 
image

* Russian 

image

* Mongolian
image

As a result, I modified to center alignment only at 1400px and above.

adrysn (approved):
LGTM.

Issue #1110, opened 2021-08-24
Make the text of the config button horizontally.


Sujin-Kim1:
**Is your feature request related to a problem? Please describe.**
On the `configurations` page, when the system language setting is Korean, the text of the config button looks vertical. Make the text of the config button horizontally whatever the language environment it is.
- English (Nice 👍 )
  ![image](https://user-images.githubusercontent.com/28584164/130576970-fccaaf2c-33c4-4ef4-b3ed-516df6242f30.png)
- Korean (Bad 👎 )
  ![image](https://user-images.githubusercontent.com/28584164/130577086-1a9c3825-4f68-43a9-9e6d-75d5af4b35f4.png)

**Describe the solution you'd like**
어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.

**Describe alternatives you've considered**
혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.

**Additional context**
기능 요청에 대해 보다 잘 이해할 수 있는 다른 맥락을 기술해주세요.

studioego:
In my opinion, needs to add CSS's word-break feature.
Folloing link is CSS's word-break rule
https://developer.mozilla.org/ko/docs/Web/CSS/word-break

studioego:
Before set `word-break: keep-all;`
<img width="837" alt="image" src="https://user-images.githubusercontent.com/20002/165696698-13394f00-5fc9-4b6a-ab71-38c60c6a59d9.png">

After set `word-break: keep-all;`
<img width="842" alt="image" src="https://user-images.githubusercontent.com/20002/165697068-e5209fe7-f777-49d9-b2e6-f5dcb3cef681.png">

PR #1285, opened 2022-04-28
fix setting page's config button horizontally

studioego:
fix setting page's config button horizontally

In Korean case, If you resize the width of the setting page,
sometimes the config button is displayed vertically.

I solved the issue using the CSS's `word-break` feature.
close https://github.com/lablup/backend.ai-webui/issues/1110


Sujin-Kim1 (approved):
LGTM.

PR #1286, opened 2022-04-28
feat: show vfolder host when creating a session

### LINKED ISSUE #1271: Show vfolder host when creating a session
Show the vfolder host in the folder mount page when creating a new session.

**Additional context**
If you use multiple nas, it would be better to display the vfolder host so that you can know which nas you use.

Sujin-Kim1:
this fixes #1271 

[Screenshot]
image


adrysn (approved):
LGTM.

Issue #1287, opened 2022-04-28
Max memory per container is limited to 16GB in desktop app


adrysn:
In trying to launch a compute session from desktop apps, the maximum amount of main memory is limited to 16GB even if there are enough available resources in the node and the resource policy allows more memory.

<img width="465" alt="image" src="https://user-images.githubusercontent.com/7539358/165784139-82180d7f-029c-4b42-b4c1-b7c97c973a3c.png">

Nowadays, 16GB of memory is too small, so many desktop app users complain about this limitation. Currently, they need to create a compute session from their browser and switch to the desktop app to use the SSH/sFTP app.

### How to reproduce
* Launch a desktop app.
* Launch a session creation dialog and proceed to the resource panel.
* Open the custom resource panel.
* Try to increase the memory slider.
* You will find that you cannot increase the value of the memory over 16GB.

Issue #1288, opened 2022-04-28
Unable to import GitHub repository in a desktop app


adrysn:
In a desktop app, users are unable to import a GitHub repository. This is because the desktop apps always try to use `index.docker.io/lablup/python:3.8-ubuntu18.04` image, which is not available on most sites, to launch a temporary compute session to get the repository contents.

### Steps to reproduce
* Start a desktop app.
* Go to "Import & Run".
* Fill in GitHub URL in the "Import GitHub Repository".
* Click the "Import" button on the right of the URL input field.
* You will be faced with the error like the image below.

<img width="1008" alt="image" src="https://user-images.githubusercontent.com/7539358/165785915-504be2f1-0f62-42ef-960d-985c62e18229.png">

adrysn:
This is resolved by https://github.com/lablup/backend.ai-webui/pull/1289.

PR #1289, opened 2022-04-28
fix: override desktop app's configs with those of webserver

adrysn:
This PR resolves https://github.com/lablup/backend.ai-webui/issues/1287 and https://github.com/lablup/backend.ai-webui/issues/1288.


src/components/backend-ai-login.ts
```diff
@@ -606,6 +606,16 @@ export default class BackendAILogin extends BackendAIPage {
     this.api_endpoint = this.api_endpoint.trim();
     if (this.connection_mode === 'SESSION') {
       // this.block(_text('login.PleaseWait'), _text('login.ConnectingToCluster'));
+      if (!window.location.href.startsWith(this.api_endpoint)) {
```
inureyes: @adrysn Do we really need this part? Since every login logic uses check_login prior to login, this code do the same behavior along with the logic in check_login.

inureyes (changes requested):
@adrysn Please check the review. :)

src/components/backend-ai-login.ts
inureyes: Ah, we can switch between API and session mode. Nevermind;

inureyes (approved):
LGTM

src/components/backend-ai-login.ts
inureyes: Let's make it as function, not a duplicated code blocks.

inureyes (changes requested):
Code cleanup needed.

inureyes (approved):
LGTM.

inureyes:
@adrysn  IT blocks users to change endpoint after it successfully login to a service.

PR #1290, opened 2022-04-28
feat: show only folders that included allowed hosts

### LINKED ISSUE #1270: Show only folders that included in the allowed folder hosts
Even though users set the allowed folder hosts, not allowed folder hosts are exposed. If users select a folder not specified in the allowed folder hosts and create a session, it fails.

Get vfolder list with setting the project.

**Additional context**
It seems necessary if there are several folder hosts and set different folder hosts for each project.

Sujin-Kim1:
this resolves #1270


adrysn (approved):
LGTM.

PR #1293, opened 2022-05-02
fix layout calculation when reducing the size of window

studioego:
fix layout calculation when reducing the size of window

Content body's width was not calculated when screen size is smaller than 700px.
Added the width calculation when it's the case.
close https://github.com/lablup/backend.ai-webui/issues/1202


studioego:
As the PR https://github.com/lablup/backend.ai-webui/pull/1214 has already merged to the main branch. So, I close the issue.

studioego:
It already resolve the issue, PR https://github.com/lablup/backend.ai-webui/pull/1214

PR #1295, opened 2022-05-03
feat: notify update password

### LINKED ISSUE #1183: Alert users with notification dialog when the user needs to update their password
In the user's DB, there is a field [`need_password_change`](https://github.com/lablup/backend.ai-manager/blob/bd14b086ac4b48bac72bc9d83116416e45733e6f/src/ai/backend/manager/models/user.py#L110) which indicates the user needs to update their password. This field may be set to `True` when the password is created by an admin or the last updated date of the password is too old.

However, there is no way for a user to know the password update is required. I think it might be a good idea to display a notification dialog to change the password if the `need_password_change` is set to `True`. The dialog may be a persistent one (unable to close unless the password is updated) which is located at the header bar or some other good location.

Sujin-Kim1:
this resolves #1183 

- User will see the password update request message in every page of the header bar when `need_password_change` is true.
- User can close the message, but if refresh the page, then the message will show up again.
- If the user changes the password and refresh, this message disappear.

[Screenshot]
image

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -53,7 +53,8 @@
       "Experiments": "Experiments",
       "Menu": "Menu",
       "Run": "Run",
-      "UserName": "User Name"
+      "UserName": "User Name",
+      "PleaseChangeYourPassword": "Please change your password."
     },
     "LOGINREQUIRED": "LOGIN REQUIRED",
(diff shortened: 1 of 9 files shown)


adrysn (approved):
LGTM.

PR #1296, opened 2022-05-03
Reduce requests to fetch the session list

### LINKED ISSUE #1274: Clicking "RUNNING" tab will issue many requests to fetch the session list
The polling period of the RUNNING compute session is 5 seconds. However, when I keep clicking the "RUNNING" tab multiple times, the polling request is sent many times per second. It seems like a bug related to the `setTimeout` or `setInterval` which is used to enqueue the polling request.

To reproduce the problem:
* Go to the Sessions page.
* Open the inspector panel and click the "Network" tab to monitor requests.
* Keep clicking the "RUNNING" tab multiple times.
* In the inspector panel, you will see that the `graphql` requests sent many times per second.

I upload a video on how to reproduce the issue:
https://user-images.githubusercontent.com/7539358/163318231-055e33df-7bc2-49f0-a1bb-8e9bb3119352.mov

Sujin-Kim1:
this resolves #1274


inureyes (changes requested):
@Sujin-Kim1  It will be better if you use `this.active` as a trigger.

inureyes (approved):
LGTM

PR #1299, opened 2022-05-03
style: architecture column in session list

Sujin-Kim1:
- Reduce the width of the architecture column in session list
- Render architecture items with `lablup-shields`.

image


inureyes (approved):
LGTM

PR #1301, opened 2022-05-06
style: relocate columns of session list

Sujin-Kim1:
- Relocate columns of the session list like below.
   - User ID / SessionInfo / Status / Control / Configuration / Usage / Reservation / **SessionType / Architecture** / Agent
- Remove `auto-width` property in `Status` column.

[Screenshots]
image
image


adrysn (approved):
LGTM.

PR #1303, opened 2022-05-06
fix: override webserver configs only in the Electron app

adrysn:
This is a follow-up of https://github.com/lablup/backend.ai-webui/pull/1300.


inureyes (approved):
LGTM

PR #1310, opened 2022-05-17
fix: extend resources control detail popup width

studioego:
https://github.com/lablup/backend.ai-webui/issues/1309

fix: extend resources page control detail popup width.
cause: resources page's Control detail information popup layout is broken.

### DIFF
--- a/src/components/backend-ai-agent-list.ts
+++ b/src/components/backend-ai-agent-list.ts
@@ -109,7 +109,7 @@ export default class BackendAIAgentList extends BackendAIPage {
 
         backend-ai-dialog#agent-detail {
           --component-max-width: 90%;
-          --component-min-width: 350px;
+          --component-min-width: 400px;
         }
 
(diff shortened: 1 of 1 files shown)


lizable (commented):
I just add a comment for minor fixes, such as linting and touch-up for the layout.
IMHO, it looks much better when we split the section evenly.
What do you think about this update @studioego ?

## Screenshot
<table>
<tr>
</tr>
<tr>
<td>agent-detail-dialog-before</td>
<td>agent-detail-dialog</td>
</tr>
</table>

studioego:
> I just add a comment for minor fixes, such as linting and touch-up for the layout. IMHO, it looks much better when we split the section evenly. What do you think about this update @studioego ?

Today's lunch time, @achimnol  & I checked the screenshot. I think that It looks good.

lizable (approved):
LGTM.

PR #1313, opened 2022-05-18
Feature: enable filebrowser in readonly folder

lizable:
## Description
This PR enables filebrowser execution not only in writable folders but also in "read-only" folders.

> ⚠️ NOTE:
> Filebrowser button is enabled ONLY under conditions:
> - [x] : Filebrowser environment is already installed at least one or more agents.
> - [x] : Enough resources to execute session (**minimum resource to create the session: 1 CPU Core, 0.5 RAM**)

## Before
filebrowser-disabled-readonly

## After
filebrowser-enabled-readonly

Since users can't write any files or folders in read-only folders, uploading is not allowed likewise.
Therefore I added a notification dialog when executing filebrowser in folder-explorer dialog of read-only folder so that the user could understand the situation without facing errors.
notice-filebrowser


src/components/backend-ai-storage-list.ts
```diff
@@ -277,6 +281,10 @@ export default class BackendAiStorageList extends BackendAIPage {
           --mdc-icon-button-size: 28px;
         }
 
+        #filebrowser-notificatin-dialog {
```
adrysn: typo: `notification` -> `notification`.

src/components/backend-ai-storage-list.ts
```diff
@@ -965,6 +972,19 @@ export default class BackendAiStorageList extends BackendAIPage {
           </mwc-button>
         </div>
       </backend-ai-dialog>
+      <backend-ai-dialog id="filebrowser-notificatin-dialog" fixed backdrop narrowLayout>
```
adrysn: same here.

src/components/backend-ai-storage-list.ts
```diff
@@ -2508,6 +2528,10 @@ export default class BackendAiStorageList extends BackendAIPage {
   _executeFileBrowser() {
     if (this._isResourceEnough()) {
       if (this.filebrowserSupportedImages.length > 0) {
+        const isNotificationVisible = localStorage.getItem('backendaiwebui.filebrowserNotification');
+        if ((isNotificationVisible == null || isNotificationVisible === 'true') && !this.isWritable) {
+          this.shadowRoot.querySelector('#filebrowser-notificatin-dialog').show();
```
adrysn: same here.

adrysn (approved):
LGTM except minor typo in the element id. After fixing them, I think this PR can be merged.

src/components/backend-ai-storage-list.ts
lizable: Oops, I'll fix it right away.

achimnol:
Resolved lablup/backend.ai#414.

PR #1314, opened 2022-05-18
fix: logic error on validating scheduler options in creating resource group

lizable:
## Description
This PR fixes the logic error on validating scheduler options in creating resource group.


adrysn (approved):
Good catch!

PR #1315, opened 2022-05-18
Hotfix: translation for missing content

lizable:
## Description
This PR adds translations to certain notifications that remained only in "English", which could help users from non-English speaking countries understand the situations more easily.


adrysn (approved):
Good catch! Thanks.

PR #1174, opened 2021-12-14
feat: provide a dialog used for detailed status data

lizable:
### Description
---------
This PR resolves #882.

### Before
---------
Only `status_info`, which contained a simple title was displayed when hovering to each session status column.

### After
---------
`status_data` has three types:

1. Terminated
   - Corresponding `status_info`:
      - `user-requested`
      -  `self-terminated`
      -  `force-terminated`
  
Screen Shot 2021-12-14 at 11 43 22 PM

2.  Cancelled
    - Corresponding `status_info`:
      - `failed-to-start`
      -  `scheduler-error`
      -  `user-requested` 

> ⚠️&nbsp;  NOTE:  `Agent ID` is only visible when current user role is `superadmin` and `agent_id` key is not null.
Screen Shot 2021-12-15 at 1 36 01 AM

3. PENDING
   - Corresponding `status_info`:
      - `no-available-instances`
      -  `predicate-checks-failed`     

scheduler-status_data mov

PR #1316, opened 2022-05-18
Fix: display rich and more detailed msg in status info of session

lizable:
## Description
From 21.09, we are providing detailed session status for better user experiences. (#1174).
But recently I found that some of the descriptions are hidden because the contents to display exceed the default height of the section. This PR provides more flexible display information without ellipsis.

## Before
overlapped

## After
fix-overlap

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -295,6 +295,15 @@ export default class BackendAiSessionList extends BackendAIPage {
           font-weight: bold;
         }
 
+        mwc-list-item.predicate-check {
+          height: 100%;
+          margin-bottom: 5px;
+        }
+
+        .predicate-check-comment {
+          white-space: pre-wrap;
+        }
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-session-list.ts
```diff
@@ -1283,13 +1297,13 @@ export default class BackendAiSessionList extends BackendAIPage {
             <h3 style="width:100%;padding-left:15px;border-bottom:1px solid #ccc;">${_text('session.StatusDetail')}</h3>
             <div class="vertical layout flex" style="width:100%;">
               <mwc-list>
-                <mwc-list-item twoline noninteractiv>
+                <mwc-list-item twoline noninteractiv class="predicate-check">
```
agatha197: Isn't it `noninteractive`? (not `noninterativ`)

adrysn (approved):
LGTM.

PR #1317, opened 2022-05-19
fix: add the value of `wsproxy_addr` so that reflecting the setting value

Sujin-Kim1:
Even if the user has set the `wsproxy_addr`, its value disappears in modify dialog.
image
image

[After]
image


adrysn (approved):
LGTM.

Issue #1321, opened 2022-05-27
Found a querySelect error bug on the environment page when trying to download an image from a custom repo.


studioego:
**Describe the bug**
문제가 발생하는 상황에 대한 요약 설명을 적어주세요.

Found a querySelect error bug on the environment page when trying to download an image from a custom repo.

**To Reproduce**
문제를 재현하기 위한 순서를 자세히 적어주세요:
1. Go to '/environment' pages registry tab
2. regist custom gitlab repo(docker) url link
3. go to '/environment' page's image tab
4. select custom odcker image & click 'install' button
5. See error (show querySelect 

**Expected behavior**
원래 기대했던 동작은 어떤 것인지 명확하게 적어주세요.
install image from a custom docker repo.

**Screenshots**
Just take a screenshot in your clipboard or copy a picture file and paste it here (Ctrl+V/Cmd+V) so that GitHub automatically attach and insert it.
가능하다면, 문제 상황을 보여주는 스크린샷을 첨부해주세요.
클립보드에 스크린샷을 찍거나 이미지 파일을 복사해서 여기에 붙여넣으면(Ctrl+V/Cmd+V) GitHub이 자동으로 이곳에 첨부 및 삽입해줍니다.
<img width="1100" alt="스크린샷 2022-05-27 오후 2 28 13" src="https://user-images.githubusercontent.com/20002/170639979-e33da114-dc5d-48b5-a9f5-3e649f3d0fc9.png">
<img width="1975" alt="스크린샷 2022-05-27 오후 2 28 05" src="https://user-images.githubusercontent.com/20002/170639989-1b23296d-8d5d-4541-a87d-9a42a6b9151d.png">

**Client:**
 - OS: both mac OS 12.3.1 and Fedora Core 35 
 - Browser version: [e.g. Chrome 76, Safari, Edge] (if you are using a GUI web console)
 - Python version: 3.10.3
 - NodeJS version: 18.2.0

**Server:**
 - OS: both mac OS 12.3.1 and Fedora Core 35 
   - Output of `uname -a`: 
     - mac OS: Darwin seongdacBookPro 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64
[...]

PR #1322, opened 2022-05-27
Refactor: querySelector with leading Number id string(#1321)

studioego:
Refactor: querySelector with leading Number id string(#1321)

Changed the querySelector's value to explicitly stating id values `querySelector("[id='1']")`

example) id value: `10-20-30-10...blahblah`
If the first character of an identifier is numeric, you’ll need to escape it based on its Unicode code point.
However, If you don't change string to Unicode code point,
you would have to do this: `querySelector("[id='1']")`

Ref: Using querySelector with IDs that are numbers
https://stackoverflow.com/questions/20306204/using-queryselector-with-ids-that-are-numbers


adrysn (approved):
Good catch! LGTM.

PR #1323, opened 2022-05-30
fix: some session description dialog does not appear.

Sujin-Kim1:
Due to the unnecessary close brace `}`, some description dialog does not appear on the session launcher.


adrysn (approved):
LGTM.

PR #1327, opened 2022-06-10
Feature: refactor storage host selection

### LINKED ISSUE #1326: Can't select multiple storage hosts in resource policy setting
For now, the superadmin can select only one allowed storage host when creating/updating resource policies in WebUI, whereas Backend.AI itself supports adding or excluding more than one storage host. And also, the list only shows "allowed storage hosts", which doesn't show all of storage host that are connected to [`storage-proxy`](https://github.com/lablup/backend.ai-storage-proxy), which doesn't make sense.
Therefore we need to update user interface to provide multiple selections or none for the allowed storage host.

어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.
> ⚠️ &nbsp; **NOTE**: This feature is only for `superadmin`.
- [x] Show all storage hosts connected to storage-proxy in selection list
- [x] Provide multiple or none selection for allowed storage hosts

**Additional context**
image

lizable:
## Description
This PR resolves #1326.

## Before
image

## After
backendai-multiselect


lizable:
> It works well without any problem. But could you reduce the height of allowed hosts to 56px like other fields?

Hmm, I can't get what you mean. Reducing the total height or sth else?

Sujin-Kim1:
> > It works well without any problem. But could you reduce the height of allowed hosts to 56px like other fields?
> 
> Hmm, I can't get what you mean. Reducing the total height or sth else? IMO, total height could be extended when there's more than one accessible storage host to the keypair resource policy. So I think increasing the overall height of the section is inevitable.

I think there was a misunderstanding because I didn't tell you in detail. I meant that the height seemed a little high when nothing or only one item was selected. And when I tested it, and the simplest way to do it was to reduce the ` --selected-area-padding` value to 4. But if there is only one item, it seems strange because the padding is not constant. :( So I will follow your judge.
image

lizable:
By suggestion of @Sujin-Kim1, the final UI for storage host selection would look like this:
Screen Shot 2022-07-07 at 5 23 14 PM

src/components/backend-ai-credential-view.ts
```diff
@@ -368,15 +402,12 @@ export default class BackendAICredentialView extends BackendAIPage {
...
-    globalThis.backendaiclient.vfolder.list_hosts().then((response) => {
-      this.allowed_vfolder_hosts = response.allowed;
-      this.default_vfolder_host = response.default;
-      this.shadowRoot.querySelector('#allowed_vfolder-hosts').layout(true).then(()=>{
-        this.shadowRoot.querySelector('#allowed_vfolder-hosts').select(0);
-      });
+  * Get All Storage host information (superadmin-only)
+  */
+  _getAllStorageHostsInfo() {
+    globalThis.backendaiclient.vfolder.list_all_hosts().then((res) => {
```
adrysn: `list_all_hosts` is a superadmin-only API. Then, what happens when the user is a domain admin? (User page is exposed to domain admins, FYI.)

src/components/backend-ai-multi-select.ts
```diff
@@ -0,0 +1,262 @@
...
+ <backend-ai-multi-select></backend-ai-multi-select>
+
+ @group Backend.AI Web UI
+ @element backend-ai-multi-select
+ */
+
+@customElement('backend-ai-multi-select')
+export default class BackendAIMultiSelect extends LitElement {
+  public shadowRoot: any; // ShadowRoot
+  @property({type: Object}) comboBox;
```
adrysn: Do we need to expose the `comboBox` as an element's property? It seems like it is only used internally, and then, I think it is better to declare it as a private class variable.

src/components/backend-ai-credential-view.ts
```diff
@@ -256,6 +260,34 @@ export default class BackendAICredentialView extends BackendAIPage {
...
+          --selected-item-font-family: var(--general-font-family);
+          --selected-item-theme-color: #C8CED7;
+          --selected-item-theme-font-color: #182739;
+          --selected-item-unelevated-theme-color: #C8CED7;
+          --selected-item-unelevated-theme-color: #C8CED7;
+          --selected-item-outlined-theme-font-color: black;
+          --selected-item-unelevated-theme-font-color: black;
+          --selected-item-font-size: 14px;
+          --selected-item-text-transform: none;
+        }
```
adrysn: The style block for the `backend-ai-multi-select` is the same as the one defined in the `backend-ai-resource-policy-list.ts`: https://github.com/lablup/backend.ai-webui/pull/1327/files#diff-d61f09d5fee23b684f5d92bbb9987a8c9023fb98e1f639c9f4cfde7ff9d0d124R210-R236

How about take this out as the common/default style?

src/components/backend-ai-multi-select.ts
```diff
@@ -0,0 +1,262 @@
...
+  @property({type: Array}) selectedItemList;
+  @property({type: String, attribute: 'label'}) label = '';
+  @property({type: Array}) items;
+  // TODO: Clear button to remove all selected
+  // TODO: AutoComplete(filtering)
```
adrysn, line 54: One more todo (in the future):
Currently, checking one item closes the selection popup. This is inconvenient if a user wants to select/deselect multiple items at once. I think it would be better to keep open the selection popup until a user explicitly closes or the focus is moved to the other elements.

src/components/backend-ai-multi-select.ts
```diff
@@ -0,0 +1,262 @@
...
+    dropdownIcon.classList.toggle('expand');
+    if (e.detail.isOn) {
+      this._showMenu();
+    } else {
+      this._hideMenu();
+    }
+  }
+
+  /**
+   * Set the position of Dropdown menu
```
adrysn, line 168: Can you add an explanation/description of why we need to modify the position of the list item using this method for future reference?

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -202,14 +206,42 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
+        }
       `];
   }
 
   render() {
     // language=HTML
     return html`
       <vaadin-grid theme="row-stripes column-borders compact" aria-label="Resource Policy list"
-                   .items="${this.resourcePolicy}">
+                  .items="${this.resourcePolicy}">
```
adrysn, line 216: Purpose of the one-space removal?

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -440,15 +462,15 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
-                                 @click="${(e) => this._launchResourcePolicyDialog(e)}"><wl-icon>settings</wl-icon></wl-button>
-                                 ` : html``}
+                                @click="${(e) => this._launchResourcePolicyDialog(e)}"><wl-icon>settings</wl-icon></wl-button>
+                                ` : html``}
         ${this.is_admin ? html`
               <wl-button fab flat inverted class="fg red controls-running" icon="delete"
-                                 @click="${(e) => this._openDeleteResourcePolicyListDialog(e)}"><wl-icon>delete</wl-icon></wl-button>
-                                 ` : html``}
+                                @click="${(e) => this._openDeleteResourcePolicyListDialog(e)}"><wl-icon>delete</wl-icon></wl-button>
+                                ` : html``}
```
adrysn, line 445: Purpose of the one-space removal? I don't understand the criteria for this alignment change. (Although, I don't mean the original ones are correct.)

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -741,10 +762,10 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
 
   /**
-   * Set a TextEl value according to toggle checkbox checked state.
-   *
-   * @param {Event} e - Dispatches from the native input event each time the input changes.
-   */
+  * Set a TextEl value according to toggle checkbox checked state.
+  *
+  * @param {Event} e - Dispatches from the native input event each time the input changes.
+  */
```
adrysn: Is there a reason or a style guide to account for these changes? I am usually accustomed to the convention with spaced comments for readability: https://developer.wordpress.org/coding-standards/inline-documentation-standards/javascript/#aligning-comments.

Also, this kind of style principle should be mechanically checked by activating the ES linter, which might be a good separate refactoring issue.

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -819,12 +840,12 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
-   * Returns human-readable value according to certain conditions
-   *
-   * @param value - raw value
-   * @param enableUnitConvert - if true it enable unit conversion
-   * @return if number then returns number, else if then string
-   */
+  * Returns human-readable value according to certain conditions
+  *
+  * @param {String} value - raw value
+  * @param {Boolean} enableUnitConvert - if true it enable unit conversion
```
adrysn: Shouldn't the first character be lower cases?
Example reference: https://www.typescriptlang.org/docs/handbook/2/functions.html#call-signatures

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -851,21 +872,19 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
-    if (itemCount > 0) {
-      this.shadowRoot.querySelector('#dropdown-area').style.height = (actualHeight + itemCount * 14) +'px';
-    }
+  * Get All Storage host information (superadmin-only)
+  */
+  _getAllStorageHostsInfo() {
+    globalThis.backendaiclient.vfolder.list_all_hosts().then((res) => {
+      this.all_vfolder_hosts = res.allowed;
+    }).catch((err) => {
+      console.log(err);
```
adrysn: Please remove the debug line.

src/lib/backend.ai-client-node.ts
```diff
@@ -1584,6 +1588,17 @@ class VFolder {
     return this.client._wrapWithPromise(rqst);
   }
 
+  /**
```
adrysn, line 1591: This is just a question. I heard that now we only use the `backend.ai-client-esm.ts` without manual compilation of `backend.ai-client-node.ts`. However, I might be wrong, so my question is, do we still have to update and keep the traditional `backend.ai-client-node.ts`?

adrysn (changes requested):

src/components/backend-ai-credential-view.ts
lizable: You're right, I overlooked that part. It's better to fetch right before the dialog opens, which is only allowed to `superadmin` role.
Btw, I have one question about manipulating for (keypair) resource policy. For now, `admin` role users can access, modify and delete resource policy, but they cannot create. From My point of view, only allowing "READ" permission for resource policies is more appropriate for `admin` role users rather than keeping "WRITE" and "DELETE" permission; unless there's a certain reason to allow manipulating and even deleting.

src/components/backend-ai-multi-select.ts
lizable: Yes, It degrades the user experience. I'll fix it.

src/components/backend-ai-multi-select.ts
lizable: Oh, it's for manipulating the height of expanded items. I'll add it in the function description.

src/components/backend-ai-resource-policy-list.ts
lizable: Maybe it's because of the artifacts my experimental linter tests. I'll restore it.

src/components/backend-ai-resource-policy-list.ts
lizable: Perhaps It's the same reason described above.

src/components/backend-ai-resource-policy-list.ts
lizable: Good point. I'm not sure the current linting rule will automatically fix this without changing other codes strange, for this time I'll fix it manually.

src/components/backend-ai-resource-policy-list.ts
lizable: Oops, I'll get rid of it.

src/lib/backend.ai-client-node.ts
lizable: I'm not sure but, If this PR needs backporting (such as 21.09 or former versions), then we may modify `backend.ai-client-node.ts` code.

src/components/backend-ai-multi-select.ts
lizable: I agree, there's no reason to expose those items since it's only for "superadmin". I'll change it.

src/components/backend-ai-credential-view.ts
adrysn: (Domain) admin should be able to create, modify, and delete a keypair resource policy which is bound to a user who belongs to the same domain with the admin. If the create resource policy is not working in the admin's domain, I think it is a permission bug.

lizable: @adrysn 
Sorry, I'm a bit confused about accessing resource policy according to your last comment. Since `admin` role users do have permission to "create", "modify", and "delete", they can see all storage hosts that are connected to manager so that they can add or remove them as "allowed storage hosts". Otherwise, we should only allow "Read" permission for `admin` role users. (For now, there's only one access point for listing up all storage host in WebUI, which is "creating/modifying (keypair) resource policy" dialog).

And also, I think It's better to handle this issue as a separate issue.

adrysn: The permission system for the domain user is a separate backend topic. In this PR, we will ignore this issue.

src/components/backend-ai-credential-view.ts
```diff
@@ -77,9 +76,16 @@ export default class BackendAICredentialView extends BackendAIPage {
   @property({type: String}) _defaultFileName = '';
   @property({type: Number}) selectAreaHeight;
   @property({type: Boolean}) enableSessionLifetime = false;
+  @state() private all_vfolder_hosts;
+  @state() private allowed_vfolder_hosts;
```
Jaewoook: This state variable doesn't seem to be used anywhere.

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -67,9 +67,14 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
   @property({type: Object}) _boundPolicyNameRenderer = this.policyNameRenderer.bind(this);
   @property({type: Object}) _boundClusterSizeRenderer = this.clusterSizeRenderer.bind(this);
   @property({type: Object}) _boundStorageNodesRenderer = this.storageNodesRenderer.bind(this);
+  @query('#allowed-vfolder-hosts') private allowedVfolderHostsSelect;
+  @state() private all_vfolder_hosts;
+  @state() private allowed_vfolder_hosts;
+  @state() private default_vfolder_host = '';
```
Jaewoook: This state variable doesn't seem to be used anywhere.

Jaewoook (changes requested):
Although it works correctly, but there's few unused variable declaration found.

src/components/backend-ai-credential-view.ts
lizable: It's used here. https://github.com/lablup/backend.ai-webui/pull/1327/files#diff-1ac42de48a37ca285d9aa6edb95c2cbe64c8474285dee56d6d036008893842d8R381

src/components/backend-ai-resource-policy-list.ts
lizable: https://github.com/lablup/backend.ai-webui/pull/1327/files#diff-d61f09d5fee23b684f5d92bbb9987a8c9023fb98e1f639c9f4cfde7ff9d0d124R566

lizable (commented):
@Jaewoook the comments about `@state` seems to have mistakes. The variable is assigned right after fetching data from manager server, which defines the initial status of the current resource policy or new resource policy (we tend to select the default storage host for convenience sake.)
I would appreciate it if you could check it again and finish your review :).

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -440,15 +435,15 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
     render(
       html`
         <div id="controls" class="layout horizontal flex center"
-             .policy-name="${rowData.item.name}">
-        ${this.is_admin ? html`
+            .policy-name="${rowData.item.name}">
```
lizable, line 437: Btw, I need to revert it since it will be resolved via #1375.

src/components/backend-ai-credential-view.ts
Jaewoook: I didn't specified what variable is. The `allowed_vfolder_hosts` variable does not seem to be used in `credential-view` component. You can see the following TS error:

image

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -42,7 +44,7 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
   @property({type: Object}) keypairs = {};
   @property({type: Array}) resourcePolicy = [];
   @property({type: Object}) keypairInfo = {};
```
lizable: Another revert operation needed here.

lizable: `is_superadmin` -> `is_admin`

src/components/backend-ai-resource-policy-list.ts
Jaewoook: The `default_vfolder_host` variable does not seem to be used in anywhere in this file.

image

Jaewoook (approved):

Jaewoook:
LGTM

PR #1328, opened 2022-06-13
fix: quota and max size of virtual folder displays NaN, and magic number remains in resource policy setting dialog

lizable:
## Description
This PR fixes two minor bugs about setting size of virtual folder(vfolder) that located in storage host that supports directory-based quota and resource policy.
Please refer to the following checkbox descriptions below:
- [x] the max values of corresponding keys in resource policy to match with the max value of DB.
- [x] Show quota and max size of virtual folder correctly.

## Additional Context
As minimizing server-side updates, Let's just clamp the max value to about 2 PiB for now; the maximum value that the corresponding attribute on `Vfolders` table in manager DB can afford.

_According to the DB schema of "resource policy(keypair_resource_policy)", which is described in [Backend.AI manager](https://github.com/lablup/backend.ai-manager/blob/85d16f0f96ef088d5cc0e41e488a468645acbbff/src/ai/backend/manager/models/resource_policy.py#L66-L77) repo, It has both numeric type, "Integer" and "BigInt". For now, we've been dealing with "0(zero)" value as Infinity, but It's not a good approach when it comes to "disabling" certain values by setting them to zero. IMO, It's better to fix not to treat "0(zero)" value as "Infinity"._


Sujin-Kim1 (changes requested):
If user inputs invalid value, the bottom border turn into red but it can be updated.
image
image

lizable:
> 

I think this comment is irrelevant to this PR. Could you make a separate issue for this?

Sujin-Kim1 (approved):
LGTM.

Sujin-Kim1:
> I think this comment is irrelevant to this PR. Could you make a separate issue for this @Sujin-Kim1 ?
> 
> > If the user inputs an invalid value, the bottom border turns red but it can be updated. image image

https://github.com/lablup/backend.ai-webui/issues/1336 here it is.

src/components/backend-ai-credential-view.ts
```diff
@@ -848,7 +856,10 @@ export default class BackendAICredentialView extends BackendAIPage {
     if (textfield.value === '' || textfield.value === '0' ) {
       textfield.disabled = true;
       checkbox.checked = true;
-    } else {
+    } else if (['concurrency-limit', 'container-per-session-limit'].includes(textfield.id) && textfield.value === BackendAICredentialView.MAX_INT32) {
+      textfield.disabled = true;
+      checkbox.checked = true;
+    }else {
```
agatha197: a blank needed before the `else`

Sujin-Kim1 (changes requested):
IMO, it looks very similar between the `new-policy-dialog` dialog and the `modify-policy-dialog`. I know it's beyond the scope of this issue, but what do you think about creating an issue that merges the two dialogs into one? If you agree with me, I'll create it.

src/components/backend-ai-credential-view.ts
lizable: Nice catch! Also, I got rid of debug log and updated comments on JSDoc, too.

Sujin-Kim1 (approved):
LGTM!

lizable:
@inureyes  @adrysn Could you review this PR?

src/components/backend-ai-credential-view.ts
```diff
@@ -78,6 +78,8 @@ export default class BackendAICredentialView extends BackendAIPage {
   @property({type: Number}) selectAreaHeight;
   @property({type: Boolean}) enableSessionLifetime = false;
 
+  static readonly MAX_INT32 = 0x7FFFFFFF;
```
adrysn: This same constant is defined in multiple places. How about taking this out to a common module?

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -641,21 +633,27 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
-    this.idle_timeout['value'] = this.idle_timeout['value'] === '' ? 0 : parseInt(this.idle_timeout['value']);
-    this.container_per_session_limit['value'] = ['', 0].includes(this.container_per_session_limit['value']) ? BigNumber.getValue() : parseInt(this.container_per_session_limit['value']);
-    this.vfolder_capacity['value'] = this.vfolder_capacity['value'] === '' ? 0 : parseFloat(this.vfolder_capacity['value']);
-    this.vfolder_max_limit['value'] = this.vfolder_max_limit['value'] === '' ? 0 : parseInt(this.vfolder_max_limit['value']);
+
+    this.concurrency_limit['value'] = ['', 0].includes(this.concurrency_limit['value']) ? BackendAIResourcePolicyList.MAX_INT32 : parseInt(this.concurrency_limit['value']);
+    this.container_per_session_limit['value'] = ['', 0].includes(this.container_per_session_limit['value']) ? BackendAIResourcePolicyList.MAX_INT32 : parseInt(this.container_per_session_limit['value']);
+    this.vfolder_max_limit['value'] = (this.vfolder_max_limit['value'] === '') ? 0 : parseInt(this.vfolder_max_limit['value']);
+
+    this.idle_timeout['value'] = (this.idle_timeout['value'] === '') ? Number.MAX_SAFE_INTEGER : parseInt(this.idle_timeout['value']);
```
adrysn: Is there a reason to change the default from `0` to `MAX_SAFE_INTEGER` when the value is empty (`''`)? Practically, they may be the same, but I am curious.

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -670,13 +668,13 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
       'max_vfolder_count': this.vfolder_max_limit['value'],
-      'max_vfolder_size': this._gBToByte(this.vfolder_capacity['value']),
+      'max_vfolder_size': this.vfolder_capacity['value'],
       'allowed_vfolder_hosts': vfolder_hosts
     };
 
     if (this.enableSessionLifetime) {
       this._validateUserInput(this.session_lifetime);
-      this.session_lifetime['value'] = this.session_lifetime['value'] === '' ? 0 : parseInt(this.session_lifetime['value']);
+      this.session_lifetime['value'] = this.session_lifetime['value'] === '' ? BackendAIResourcePolicyList.MAX_INT32 : parseInt(this.session_lifetime['value']);
```
adrysn: Same question here.

src/components/backend-ai-credential-view.ts
```diff
@@ -753,27 +760,27 @@ export default class BackendAICredentialView extends BackendAIPage {
...
+    if (textfield.value === '') {
+      textfield.value = textfield.min ?? 0;
+    }
+
     if (textfield.classList.contains('discrete')) {
       textfield.value = Math.round(textfield.value);
     }
 
     if (textfield.value <= 0) {
       // concurrency job and container-per-session limit must be upper than 0.
```
adrysn: I'm not sure this statement is true, but it is the policy set before this PR, so let's keep it intact. And, "upper" should be changed to "greater" or "larger".

adrysn (changes requested):

adrysn:
Also, the capacity does not seem to be updated correctly.

https://user-images.githubusercontent.com/7539358/177576439-1549e9a1-6b79-4e60-b48c-25ead6c332e2.mov

adrysn (approved):

src/components/backend-ai-credential-view.ts
lizable: I agree. But I'm afraid we need to remove those statements since it doesn't do anything.

lizable: > Also, the capacity does not seem to be updated correctly.
> 
>  2022-07-06.23.36.52.mov

I confirmed the bug happened with the same approach. I just fixed it.

lizable:
I fixed miscellaneous errors, as @adrysn mentioned in the comments above. Also, I polished up codes related to displaying the maximum size of virtual folders in (keypair) resource policy.

### Before
Screen Shot 2022-07-21 at 2 53 37 AM


### After
Screen Shot 2022-07-21 at 2 46 53 AM

lizable:
### Additional Comments
Some of the resource configurations in resource policy should treat `0(zero)` as is, for certain reasons. For example, If the admin user wants to block session/virtual folder creation for certain users(probably violated rules?),  he/she would likely use `0(zero)` value as restricting the limit of maximum counts that users can create.
In this situation, `0(zero)` should be treated as 0, not infinity.
But It seems that it may need server-side updates, and more detailed strategies and scenarios to test, so I think it's better to leave it like this and continue this discussion in another issue.

src/components/backend-ai-resource-policy-list.ts
```diff
@@ -773,27 +780,24 @@ export default class BackendAIResourcePolicyList extends BackendAIPage {
...
       } else {
         textfield.value = Math.min(Math.round(textfield.value), (textfield.value < 0) ? textfield.min : textfield.max);
       }
     }
-    // automatically check when textfield is min
+
+    // automatically check when textfield is min or max
     if (checkbox) {
-      textfield.disabled = checkbox.checked = (textfield.value == parseFloat(textfield.min));
+      textfield.disabled = checkbox.checked = ((textfield.value == parseFloat(textfield.min)) || (textfield.value == parseFloat(textfield.max)));
```
agatha197: How about `(textfield.value <= parseFloat(textfield.min)) || (textfield.value >= parseFloat(textfield.max))`?

Sujin-Kim1 (changes requested):
I updated the capacity using the arrow button, and the step is 0.1 but on the screen, it is displayed as an integer, not a float. Did you intend to?

https://user-images.githubusercontent.com/28584164/180113111-51297aa7-0847-4a6b-a083-c172f56818ef.mov

lizable:
> I updated the capacity using the arrow button, and the step is 0.1 but on the screen, it is displayed as an integer, not a float. Did you intend to?
> 
>

Nope, but I think it's because the type of ["`max_vfolder_size`" in the keypair resource policy is BigInt](https://github.com/lablup/backend.ai/blob/57065b65ecc1aa87942880199965af7de3d1c92f/src/ai/backend/manager/models/resource_policy.py#L227).
I think It's better to make it discrete.

lizable:
Btw, I need to change the capacity limit of virtual folder (`vfolder-capacity-limit`) since the type of attribute corresponding in DB schema is BigInteger. 
But there's one thing that concerns me. I'm not sure about the exact size of [BigInteger](https://www.postgresql.org/docs/current/datatype-numeric.html)(In PostgreSQL says, 8bytes)  the description in "sqlalchemy". It should be changed, anyway.

lizable:
### Additional context 
For now, we don't have any criteria for "maximum" value for actual resources, such as CPU, main memory, GPU, etc.
I'll leave them as the max integer of 4 bytes as a workaround, but I think we need to set the guide.

src/components/backend-ai-resource-policy-list.ts
lizable: Your right. It's better to cover values under min and max.

src/components/backend-ai-resource-policy-list.ts
lizable: Right, it's weird. I'll just change it to zero since it means infinity. (https://github.com/lablup/backend.ai/blob/141844d383970b7274b9861b625d7f51a6601998/src/ai/backend/manager/models/resource_policy.py#L55)

src/components/backend-ai-resource-policy-list.ts
lizable: I'll change here too.

Sujin-Kim1 (approved):
LGTM

lizable:
@inureyes  @adrysn  @achimnol  @Sujin-Kim1

## Sticking Point of this issue.

I found that many attributes that should treat `0`(zero) value "as is" in keypair resource policy.
The reason why I issued this is because of avoiding the gap between what users expect on value configuration and the actual value works inside Backend.AI components.

Here are all attributes that contain in a single keypair resource policy. 
For better understanding, I put a "`Y`" mark in every "**Does it correspond to the zero value issue? (Y/N)**" column that fits into this condition.
**attribute name** | **description** | **Does it correspond to the zero value issue? (Y/N)**
---- | ---- | ----
`name` | the name of keypair resource policy | `N`
`created_at` | timestamp of when keypair resource policy created | `N`
`default_for_unspecified`| [enum value used for configuring unspecified value in `total_resource_slots`](https://github.com/lablup/backend.ai/tree/main/src/ai/backend/common/types.py#L612-L629) | `N`
`total_resource_slots` | physical resources used for computing (except storage) | `N`
`max_session_lifetime` | [the amount of time a session can exist in the 'RUNNING' state from the time it was created.](https://github.com/lablup/backend.ai/blob/141844d383970b7274b9861b625d7f51a6601998/src/ai/backend/manager/idle.py#L378-L392) | `N`
`max_concurrent_sessions` | the maximum number of sessions that user can create on person's own | `Y`
`max_containers_per_session` | the maximum number of containers that user can create in a single session | `Y`
`max_vfolder_count` | the maximum number of vfolder that user can create | `Y`
`max_vfolder_size` | the maximum size of vfolder(storage) that user can allocate in a single vfolder | `Y`
`idle_timeout` | [the idleness of a session by the elapsed time since last used.](https://github.com/lablup/backend.ai/blob/141844d383970b7274b9861b625d7f51a6601998/src/ai/backend/manager/idle.py#L235-L375) | `N`
`allowed_vfolder_hosts` | the type of storage hosts that allowed to mount for corresponding keypairs | `N`

IMO, attributes with `Y` mark should be handled in the same policy with unlimited and zero value. 
For example, `max_vfolder_count` attribute treat zero as not allowing any extra folder creation/clone operation, and treat unlimited with four options:
- Option 1. Use the maximum number of the type as `Unlimited`. (⚠️ NOTE: using maximum number !== magic numbers)
   - pros: Relatively Fast implementation. No need to do extra chores on the server side, for now. (May need "some conditions to check unlimited in certain situations, though").
   - cons: Bad for maintainability. 
      - If the type changes, we need to change the value on the client side manually.
      - Need to know every type of the corresponding attribute, like whether is Integer(4bytes), BigInteger(8bytes), etc.
- Option 2. Use specific values such as unlimited, like `-1`
   - pros: Also Fast implementation. No need to do extra chores as Option 1 does.
   - cons: Need to consider the side-effects of malicious or unintended value manipulation.
- Option 3. Use `null`(empty) value as `unlimited`.
   - pros: No need to think about the type and no concerns for affected by type changes, of course.
   - cons: Same reason for option 2.
- Option 4: Create a column for `unlimited` and the type would be Boolean.
   - pros: Code readers/contributors can easily understand it because of the explicit factor of "unlimited", including the benefit that Option 3 does.
   - cons: Extra server-side works.
       - Need DB migrations
       - Need to consider backporting since the consequences of the change may affect many major operations such as session creation, vfolder creation, etc. we might violate our latest backporting policy because of this.

I've been also thinking about which option would be the best way in both time efficiency and maintainability, but It's hard to decide. It would be great to give your opinion about this issue and if none of them meets your satisfaction, then adding your own opinion(s) would be appreciated.

yomybaby:
If this PR is no longer valid, please close. If there is remaining something  to do, please close this PR and create an issue. @lizable 👍

lizable:
Closing this issue and probably handled as a new issue.

PR #1332, opened 2022-06-20
add: storage host that supports directory based quota setting

lizable:
## Description
This PR adds a storage host that supports directory-based quota (e.g. [Weka.io](https://docs.weka.io/fs/quota-management#directory-quotas))

Also, for further implementation goal is to make quota setting to "optional", not "mandatory".

related PR: #1328.


Sujin-Kim1 (approved):
LGTM

inureyes (approved):
LGTM

PR #1337, opened 2022-06-29
fix: make gpu required image available to download

### LINKED ISSUE #1131: Let's make GPU-required image available to download if none of the agent doesn't have GPU resources.
**Describe the bug**
문제가 발생하는 상황에 대한 요약 설명을 적어주세요.
For now, If there are no GPU resources supporting in the agent nodes and the image to download needs GPU resources, WebUI forbids image downloading, since it causes errors during session creation right after image pulling progress is done.
There's a condition to evaluate whether the image needs GPU resources [`isGPURequired`](https://github.com/lablup/backend.ai-webui/blob/17f46a65b0371bc18097a67215ca337cd7e2bb7c/src/components/backend-ai-environment-list.ts#L387-L425), and this condition blocks session creation even if the user just wants to create a session without any GPU resources, which seems a bit conservative way.
Therefore, we need to remove this blocking condition for downloading a GPU-required image and leave the choices of resource allocation to the users.

**To Reproduce**
문제를 재현하기 위한 순서를 자세히 적어주세요:
> ※ NOTE: This feature is for Admin(or Superadmin) role ONLY.
1. Go to 'Environments' page.
2. Click on 'checkbox' of the record in Images tab, which has GPU resources in Resource Limit column.
3. Click 'install' button in the right-top of the panel.
4. Install image dialog pops up.
5. Click 'Okay' button in the dialog.
6. See the snack bar message. (This is the error)

**Expected behavior**
원래 기대했던 동작은 어떤 것인지 명확하게 적어주세요.
Even if the agent doesn't have a GPU resource, at least the user could download the image.

**Screenshots**
Just take a screenshot in your clipboard or copy a picture file and paste it here (Ctrl+V/Cmd+V) so that GitHub automatically attach and insert it.
가능하다면, 문제 상황을 보여주는 스크린샷을 첨부해주세요.
클립보드에 스크린샷을 찍거나 이미지 파일을 복사해서 여기에 붙여넣으면(Ctrl+V/Cmd+V) GitHub이 자동으로 이곳에 첨부 및 삽입해줍니다.
image

**Additional context**
추가적으로 문제 상황을 설명할 수 있는 다른 맥락을 기술해주세요.
If the image optionally adds GPU resources(0 ~ ∞), then It could be Okay to download.

Sujin-Kim1:
- fix: value disappears when moving the slider to modify the image resource limit
- fix: lint errors

this resolves #1131

### DIFF
--- a/src/components/backend-ai-app-launcher.ts
+++ b/src/components/backend-ai-app-launcher.ts
@@ -348,9 +348,9 @@ export default class BackendAiAppLauncher extends BackendAIPage {
     const kInfo = await globalThis.backendaiclient.computeSession.get(['scaling_group'], sessionUuid);
     const scalingGroupId = kInfo.compute_session.scaling_group;
     const groupId = globalThis.backendaiclient.current_group_id();
-    const wsproxyVersion = (globalThis.isElectron)
-      ? 'v1'
-      : (await globalThis.backendaiclient.scalingGroup.getWsproxyVersion(scalingGroupId, groupId)).wsproxy_version;
+    const wsproxyVersion = (globalThis.isElectron) ?
+      'v1' :
+      (await globalThis.backendaiclient.scalingGroup.getWsproxyVersion(scalingGroupId, groupId)).wsproxy_version;
     return wsproxyVersion;
   }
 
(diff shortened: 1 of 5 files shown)


adrysn (approved):
LGTM.

PR #1338, opened 2022-06-29
fix: prevents resource policy updates with incorrect values

### LINKED ISSUE #1336: Block invalid value updates in resource policy
**Describe the bug**
If the user inputs an invalid value to resource policy, the bottom border turns red but it can be updated.

**To Reproduce**
1. Go to 'Users' page (for admin)
2. Click on 'Resource Policy' tab
3. Update resource policy by clicking the setting icon button.
4. Input invalid value to RAM or FGPU fields (higher than the maximum value or lower than the minimum value)
5. See error

**Expected behavior**
Automatically change to a values within a given range.

**Screenshots**
image
image

Sujin-Kim1:
- refactor: rename some variables for matching naming convention
- this resolves #1336

### DIFF
--- a/src/components/backend-ai-credential-view.ts
+++ b/src/components/backend-ai-credential-view.ts
@@ -1184,7 +1184,7 @@ export default class BackendAICredentialView extends BackendAIPage {
             </div>
             <div class="vertical layout popup-both-margin">
               <wl-label>RAM(GB)</wl-label>
-              <mwc-textfield class="resource-input" id="ram-resource" type="number" min="0" max="1024" step="0.01"
+              <mwc-textfield class="resource-input" id="ram-resource" type="number" min="0" max="100000" step="0.01"
                             @change="${(e) => this._validateResourceInput(e)}"></mwc-textfield>
               <wl-label class="unlimited">
                 <wl-checkbox @change="${(e) => this._toggleCheckbox(e)}"></wl-checkbox>
--- a/src/components/backend-ai-credential-view.ts
+++ b/src/components/backend-ai-credential-view.ts
@@ -550,6 +550,25 @@ export default class BackendAICredentialView extends BackendAIPage {
     return input;
   }
 
+  _checkResourcePolicyInputValidity() {
+    let isValid = true;
+    const resourceIds = ['cpu-resource', 'ram-resource', 'gpu-resource', 'fgpu-resource', 'container-per-session-limit',
+      'idle-timeout', 'concurrency-limit', 'session-lifetime', 'vfolder-capacity-limit', 'vfolder-count-limit'];
+    for (let i = 0; i < resourceIds.length; i++) {
+      const textfield = this.shadowRoot.querySelector('#' + resourceIds[i]);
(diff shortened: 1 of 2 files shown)


adrysn (commented):
I was unable to see an error in the main branch when I updated a resource policy with a large value of cores (eg. 5120). It just automatically decreased to 512, which is the max of the input field. So, can you help me to understand the purpose of this PR?

Sujin-Kim1:
> I was unable to see an error in the main branch when I updated a resource policy with a large value of cores (eg. 5120). It just automatically decreased to 512, which is the max of the input field. So, can you help me to understand the purpose of this PR?

**FYI, instead of blocking create / update when users input a value that does not fit into the existing range, I modified it to automatically change to a value within the range.**

There was a problem with updating when values out of range were entered for RAM and fGPU. This problem is caused by not considering min and max values when the step (based on the value increased or decreased by pressing the arrow button) is decimal. As @adrysn mentioned, I think it's better to automatically convert when users input a value that doesn't fit the range, so I modified the validate function, so please check it again.

adrysn:
Sorry, but I still could not get it. In the **main** branch, I entered large numbers on CPU and GPU resources, but they rounded down to the maximum value of the inputs (see below move). Can you please give me an example in the main branch that causes the issue?

Uploading 화면 기록 2022-07-06 23.45.28.mov…

Sujin-Kim1:
> Sorry, but I still could not get it. In the **main** branch, I entered large numbers on CPU and GPU resources, but they rounded down to the maximum value of the inputs (see below move). Can you please give me an example in the main branch that causes the issue?
> 
>  2022-07-06.23.45.28.mov

I meant the **RAM and fGPU** cause an error like below. (I tested it in the main branch)

https://user-images.githubusercontent.com/28584164/177692622-83a46adf-a145-48da-810e-18e403b6ca74.mov

Don't we need to fix it?

adrysn:
> > Sorry, but I still could not get it. In the **main** branch, I entered large numbers on CPU and GPU resources, but they rounded down to the maximum value of the inputs (see below move). Can you please give me an example in the main branch that causes the issue?
> > 2022-07-06.23.45.28.mov
> 
> I meant the **RAM and fGPU** cause an error like below. (I tested it in the main branch)
> 
>  2022-07-07.1.38.45.mov 
> 
> Don't we need to fix it?

Okay. For the memory and fGPUs.

adrysn (changes requested):
LGTM. But, I think the max value of memory, 1024 GB, is too small. The newest DGX systems have 2 TB of memory nowadays. Let's increase the max to a greater value, such as 100 TB.

Sujin-Kim1:
> LGTM. But, I think the max value of memory, 1024 GB, is too small. The newest DGX systems have 2 TB of memory nowadays. Let's increase the max to a greater value, such as 100 TB.

Thanks for reviewing! I updated the max value to 100TB. But it might be updated after merging https://github.com/lablup/backend.ai-webui/pull/1328

adrysn (approved):

PR #1339, opened 2022-06-30
hotfix: import client-esm to fix backend-ai client ReferenceError

Sujin-Kim1:
The client esm file is not imported, so if you go to the change password link via email, a blank page appears.

### DIFF
--- a/src/components/backend-ai-change-forgot-password-view.ts
+++ b/src/components/backend-ai-change-forgot-password-view.ts
@@ -19,6 +19,8 @@ import {
   IronPositioning
 } from '../plastics/layout/iron-flex-layout-classes';
 
+import * as ai from '../lib/backend.ai-client-esm';
+
 /**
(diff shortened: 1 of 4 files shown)


adrysn (changes requested):
I found one more place using `ai.backend.*` in signup component:
https://github.com/lablup/backend.ai-webui/blob/a0973224ffd359404e9bc4adc912ec1095b02a56/src/components/backend-ai-signup.ts#L205

Sujin-Kim1:
> I found one more place using `ai.backend.*` in signup component:
> 
> https://github.com/lablup/backend.ai-webui/blob/a0973224ffd359404e9bc4adc912ec1095b02a56/src/components/backend-ai-signup.ts#L205

I left that file because it was imported as a whole esm file, but it seemed good to import only the necessary modules, so I modified it as well.

adrysn (approved):
LGTM.

PR #1340, opened 2022-06-30
fix: mount shared folders when creating a session

Sujin-Kim1:
allow mounting shared folders that are not allowed by the host when creating a new session.


adrysn (approved):
LGTM.

PR #1341, opened 2022-06-30
docs: Clarify dev-setup guide and fix code-block formats

achimnol:
- Clarify that we should run `build:d` and `server:d` simultaneously.
- Update the submodule checkout command with the new mono-repo.
- Make the number of spaces between shell commands and comments
  to 2, just like Python's PEP-8 style comments, to improve readability.
  And also align the beginning column of comments within the same subsection.
- Properly apply code-block syntax declarations (console, html, json, etc.)
- In code-blocks with the "console" syntax, the shell commands must start with "$"
  while the outputs of commmands should start without it.
- The JSON syntax only uses double quotes, so fixed wrong example codes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Backend.AI Web UI uses `config.toml` located in app root directory. You can prep
 
 These are options in `config.toml`.
 
-```
+```toml
 [general]
 apiEndpoint = "[Default API Endpoint. If blank, user input field will be shown.]"
 apiEndpointText = "[Placeholder text instead of API endpoint input field.]"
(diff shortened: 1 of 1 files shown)


Yaminyam (approved):
Comments are arranged so that it is much easier to see,
It is good to make it easier to distinguish between shell scripts

README.md
```diff
@@ -280,10 +291,10 @@ Note: This command will use Web UI source in `build/rollup` directory. No certif
...
 
-```
-$ docker-compose build webui-dev // build only
-$ docker-compose up webui-dev    // for testing
-$ docker-compose up -d webui-dev // as a daemon
+```console
+$ docker-compose build webui-dev  # build only
+$ docker-compose up webui-dev     # for testing
+$ docker-compose up -d webui-dev  # as a daemon
 ```
```
lizable, line 267: According to one-line installation script in Backend.AI mono-repo, the default port number of webserver has been changed to `8090`. IMO, the description in L300 should be changed likewise.

lizable (changes requested):
I found out that some of the configuration files are outdated, such as missing certain configuration keys. I already made a PR for that(#1343), so after merging it, then we could update README as well.

achimnol:
> I found out that some of the configuration files are outdated, such as missing certain configuration keys. I already made a PR for that(#1343), so after merging it, then we could update README as well.

Yep, please finish #1343 first and then let's make updates here.

README.md
lizable: According to @inureyes and the latest developing environment of backend.ai components, this should be treated as a new issue, since we need to rebuild dockerfile as well as change port number.
Let's leave it as it is, and update README in there.

lizable (approved):

PR #1343, opened 2022-07-01
Fix: Add missing configurations

lizable:
## Description
This PR updates multiple configuration files to the latest, which contains every configuration used for a certain purpose corresponding to its name.

Related PR: #1341.


adrysn (approved):
LGTM.

Sujin-Kim1 (approved):
LGTM

achimnol (changes requested):
Shouldn't we update README as well?

config.toml.sample
```diff
@@ -1,54 +1,55 @@
...
-maxCUDASharesPerContainer = 8  # Maximum CUDA shares per container.
-maxShmPerContainer = 4 # Maximum shared memory per container.
-maxFileUploadSize = 4294967296 # Maximum size of single file upload. Set to -1 for unlimited upload.
+openPortToPublic = true          # Show option to open app proxy port to anyone.
+maxCPUCoresPerContainer = 256    # Maximum CPU per container.
+maxMemoryPerContainer = 64       # Maximum memory per container.
+maxCUDADevicesPerContainer = 16  # Maximum CUDA devices per container.
+maxCUDASharesPerContainer = 8    # Maximum CUDA shares per container.
+maxShmPerContainer = 1           # Maximum shared memory per container.
+maxFileUploadSize = 4294967296   # Maximum size of single file upload. Set to -1 for unlimited upload.
```
achimnol: Are the changes in the default sample values for `openPortToPublic`, `maxShmPerContainer` okay?

lizable: I just set the same values in the resources section on `debug.toml`, `default.toml` to sync with `config.toml.sample`. 
And I think it's okay, but better to ask @inureyes or @adrysn about this.

adrysn: Yes. `openPortToPublic` should be `false` by default. I also think the sane default for `maxCUDA*PerContainer` would be `8`.

lizable: Got it! I'll update those files right away.

lizable:
## IMPORTANT! ##
configuration files in `configs` directory are auto-created via Makefile. You don't need to change the value of each key manually.

config.toml.sample
```diff
@@ -1,46 +1,55 @@
...
 [plugin]
 # Reserved to load plugins
 #login = "signup-cloud.js"
 #sidebar = "report-cloud.js"
+#page = "test-plugin1,test-plugin2" # Show menus on the list which are separated by comma.
 
-
-
+#[license]
+#This field is generated by console-server.
```
inureyes: Let's modify `console-server `→ `backend.ai-webserver`

inureyes (changes requested):

inureyes (approved):
LGTM

inureyes (approved):
LGTM (with small modification)

Issue #1320, opened 2022-05-26
Add `Edit+Delete` permission in sharing a data folder to other users


adrysn:
In the "Modify Permissions" panel for a shared data folder, the folder owner can change the invitees' permission. Currently, the available permissions are `View`, `Edit`, and `Edit+Delete` except `Kick out user`.
<img width="642" alt="image" src="https://user-images.githubusercontent.com/7539358/170599684-26e3304e-23c6-46d4-90b9-f27a6af9e2dd.png">

In the "Share Foder" panel, however, there are only two permissions: `View` and `Edit`. It would be logical that the available permissions are the same in sharing and modifying. I suggest adding the `Edit+Delete` permission in the initial sharing process.
<img width="380" alt="image" src="https://user-images.githubusercontent.com/7539358/170600516-b38bc9c4-8317-4297-94cc-0a0f8add1aa8.png">

AshwinNema:
@adrysn I'm new to open-source contribution and I'm interested by your issue. Can I take up this issue?

adrysn:
> @adrysn I'm new to open-source contribution and I'm interested by your issue. Can I take up this issue?

Great! Please take this issue and let me know if there are any updates. Thanks.

AshwinNema:
@adrysn  I forked the project and tried doing npm i but I got this error, ![Screenshot (31)](https://user-images.githubusercontent.com/72392780/170851307-06a0bbdf-ce3c-4509-b90a-009144a14c59.png) 
 I am using windows 10 64-bit with latest version of npm. Any idea how to resolve this issue?

AshwinNema:
@adrysn  Can you please share default configuration for config.toml for setting it up in my local system

AshwinNema:
@adrysn Can you please share default configuration for config.toml for setting it up in my local system. I am using windows-10 64 bits

AshwinNema:
@adrysn  I am getting this error when I am doing npm run server:d. I am using windows-10 64 bit.
Can you please advice on how to resolve this issue
 ![Screenshot (35)](https://user-images.githubusercontent.com/72392780/170865724-962c3daf-b2b6-49f3-8953-c18a1c7ed8b0.png)

adrysn:
[...]

Issue #1044, opened 2021-06-29
Replace weightless component to mwc


lizable:
**Is your feature request related to a problem? Please describe.**
기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.

Since [mwc(material-compontents-web-components)](https://github.com/material-components/material-components-web-components) has become our new UI-library standard, we need to replace weightless components with mwc unless it's irreplaceable.

**Describe the solution you'd like**
어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.

According to the latest release of mwc(v0.21.0), almost every UI component except wl-expansion could be replaced to mwc.

<figure class="table op-uc-figure_align-center op-uc-figure">

weightless | mwc(material-components-web-components)
-- | --
wl-button | [mwc-button](https://github.com/material-components/material-components-web-components/tree/master/packages/button)
wl-card | [mwc-dialog](https://github.com/material-components/material-components-web-components/tree/master/packages/dialog) or backend-ai-dialog
wl-checkbox | [mwc-checkbox](https://github.com/material-components/material-components-web-components/tree/master/packages/checkbox)
wl-dialog | [mwc-dialog](https://github.com/material-components/material-components-web-components/tree/master/packages/dialog) or backend-ai-dialog
wl-icon | [mwc-icon](https://github.com/material-components/material-components-web-components/tree/master/packages/icon) or [mwc-icon-button](https://github.com/material-components/material-components-web-components/tree/master/packages/icon-button)
wl-label | [mwc-button](https://github.com/material-components/material-components-web-components/tree/master/packages/button)* ( *disabled must be enabled)
wl-list-item | [mwc-dialog](https://github.com/material-components/material-components-web-components/tree/master/packages/dialog) or backend-ai-dialog
wl-popover | [mwc-tooltip](https://github.com/material-components/material-components-web-components/tree/master/packages/tooltip) * (NOT IMPLEMENTED YET)
wl-progress-bar | [mwc-linear-progress](https://github.com/material-components/material-components-web-components/tree/master/packages/linear-progress)
wl-progress-spinner | [mwc-circular-progress](https://github.com/material-components/material-components-web-components/tree/master/packages/circular-progress)
wl-select | partially [mwc-select](https://github.com/material-components/material-components-web-components/tree/master/packages/select) (since, mwc-select doesn't store initial selection,some components need to us wl-select instead of mwc-select)
[...]

PR #1351, opened 2022-07-05
fix: support wd permission in initial sharing process

Jaewoook:
## Fixes

- replace checkbox to radio button to enforce single selection for selecting permission
- update logic for getting selected permission
- use i18n data.folders keys instead of button keys

---
related issue: #1320

### DIFF
--- a/src/components/backend-ai-storage-list.ts
+++ b/src/components/backend-ai-storage-list.ts
@@ -33,6 +33,7 @@ import 'weightless/dialog';
 import 'weightless/divider';
 import 'weightless/icon';
 import 'weightless/label';
+import 'weightless/radio';
 import 'weightless/select';
 import 'weightless/title';
 import 'weightless/textfield';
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-storage-list.ts
```diff
@@ -860,12 +861,16 @@ export default class BackendAiStorageList extends BackendAIPage {
           <div style="margin: 10px 0px">${_t('data.explorer.Permissions')}</div>
           <div style="display: flex; justify-content: space-evenly;">
             <wl-label>
```
lizable: We've been replacing [`weightless`](https://github.com/andreasbm/weightless) components with [`mwc`](https://github.com/material-components/material-web) (*there will be **ground-breaking updates** on mwc though🤷🏻‍♀️*), unless there's no counterpart element implemented in mwc. (e.g. `wl-expansion`, `wl-tooltip`).
The reason why we changing the major library is that `mwc` has been updating more frequently (more likely to say "alive") rather than `weightless`. For more information, please checkout the last commit date of main/master branch of each library.

IMO, It's better to use mwc rather than using weightless.
Also, I already created an issue about "replacing weightless components to mwc(#1044)".
It's not that urgent issue, but we shall make a move soon.

Jaewoook: I understand and agree with your opinion. I searched about both two libraries (weightless and mwc) and I checked the weightless latest release year is 2019. I'll replace weightless to mwc-radio and request review again. 🙏

Jaewoook: I applied `--general-textfield-selected-color` color to Share Folder panel's permission radio inputs because there's no color for radio input. Please check the blow image and give your opinion.

image

lizable (approved):
LGTM.

PR #1356, opened 2022-07-07
fix: cluster size slider to be active by cluster mode selection

Jaewoook:
## Fixes

- call `updateResourceAllocationPane()` on cluster mode selection handler

### additional comment:

I think `updateResourceAllocationPane()` method need code refactoring. The method cause logical confusion, but I didn't do this because I don't know all intent of condition statement and variables behavior.

---
related issue: #1347

### DIFF
--- a/src/components/backend-ai-session-launcher.ts
+++ b/src/components/backend-ai-session-launcher.ts
@@ -2225,6 +2225,7 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
    */
   _setClusterMode(e) {
     this.cluster_mode = e.target.value;
+    this.updateResourceAllocationPane();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Sujin-Kim1 (changes requested):
I found this below error.
1. Update resource allocation on the 3rd session page 
2. Set another image on 1st session page
3. Go to the 3rd page
4. Then all custom resource allocations are reset.

Jaewoook:
> I found this below error.
> 
> 1. Update resource allocation on the 3rd session page
> 2. Set another image on 1st session page
> 3. Go to the 3rd page
> 4. Then all custom resource allocations are reset.

I think it's bit more complex to fix. I'll refactor update resource allocation logic after #1360 done.

lizable:
> ## Fixes
> * call `updateResourceAllocationPane()` on cluster mode selection handler
> 
> ### additional comment:
> I think `updateResourceAllocationPane()` method need code refactoring. The method cause logical confusion, but I didn't do this because I don't know all intent of condition statement and variables behavior.
> 
> related issue: #1347

I agree with your opinion, especially in refactoring `updateResourceAllocationPane()`. It does too many things at one function which degrades understanding logic inside. Let's create an issue for that and solve it!

inureyes (approved):
Let's merge and close this issue. :D

PR #1358, opened 2022-07-08
feat: allow user to force termination of sessions

Sujin-Kim1:
Allows the user to force-terminate the session to prevent the session from remaining in the PREPARING, TERMINATING status.

Related PR: https://github.com/lablup/backend.ai/pull/525


Yaminyam (commented):
I think you activated the button so that can use the forced end function even if I am not an admin. So, is there a separate screen that notifies the user when I request a forced end for a non-session?

adrysn:
> I think you activated the button so that can use the forced end function even if I am not an admin. So, is there a separate screen that notifies the user when I request a forced end for a non-session?

When clicking the force terminate button, a confirmation dialog will appear. Also, a normal user cannot terminate another user's compute session. It is prevented from the server-side.

adrysn (approved):
LGTM.

PR #1361, opened 2022-07-11
fix: side menu icon button ripple to disappear after click

Jaewoook:
## Description
The **Material Web Component**'s **`<icon-button />`** has default ripple component its inside. The problem is the background focus ripple after clicked button is show by default but it does not disappear even user click button's outside after `display: block` -> `display: none`. I think this is mwc's wrong behavior. So I changed background focus not to disappear by default.

## Fixes

- remove remaining ripple from menu drawer's icon button

### Screenshots

**before**
image
> icon button's background focus ripple remained

**after**
image
> icon button's background focus ripple does not appear

---
related issue: #1352

### DIFF
--- a/src/components/backend-ai-webui-styles.ts
+++ b/src/components/backend-ai-webui-styles.ts
@@ -441,6 +441,7 @@ export const BackendAIWebUIStyles = [
     mwc-icon-button.side-menu {
       --mdc-icon-button-size: 44px;
       --mdc-theme-text-disabled-on-light: var(--paper-grey-800);
+      --mdc-ripple-focus-opacity: 0;
     }
 
     #sidebar-navbar-footer mwc-icon-button {
--- a/src/components/backend-ai-webui-styles.ts
+++ b/src/components/backend-ai-webui-styles.ts
@@ -444,6 +444,10 @@ export const BackendAIWebUIStyles = [
       --mdc-ripple-focus-opacity: 0;
     }
 
+    mwc-icon-button#mini-ui-toggle-button {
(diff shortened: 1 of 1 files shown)


Sujin-Kim1 (changes requested):
Can you do the same for the menu icon?
image

Jaewoook:
> Can you do the same for the menu icon? image

@Sujin-Kim1  I just changed toggle button to the same style

lizable (approved):
LGTM. But there's one thing to make clear before merging this PR to main branch. 
@inureyes Should it be better to backport to other versions, such as 21.09 ? IMO, it's a minor cosmetic issue, so we could just apply this to the latest release (22.06 perhaps?)

Sujin-Kim1 (approved):
LGTM!

Yaminyam (approved):
LGTM!

adrysn:
I think there is no need to backport this PR.

lizable:
@Yaminyam suggested applying this globally when the icon doesn't actually do any interactions, like opening dialog or something else. And I think It's a good suggestion, and probably we can raise his suggestion as a `good-first-issue`.

Jaewoook:
> @Yaminyam suggested applying this globally when the icon doesn't actually do any interactions, like opening dialog or something else. And I think It's a good suggestion, and probably we can raise his suggestion as a `good-first-issue`.

I agree with this opinion. 👍 However we should take cafe of that there's another library (weightless) icon button component. I think the behavior between two libraries are different.

Issue #1124, opened 2021-09-07
Let's show resource statistics according to current resource group in sessions page


lizable:
\*Is your feature request related to a problem? Please describe.\*A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.Currently, the resource statistics panel only reflects the first selected resource group on the summary page.This term degrades user experiences when the user can access more than one resource group.

\*Describe the solution you'd like\*A clear and concise description of what you want to happen.어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.the gauges in resource statistics panel at sessions page need to reflect the currently selected resource group in summary page.

\*Describe alternatives you've considered\*A clear and concise description of any alternative solutions or features you've considered.혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.If it's tricky to reflect (since Backend.AI WebUI aims SPA), just simply adding resource group selection could solve this problem, too.

\*Additional context\*Add any other context or screenshots about the feature request here.기능 요청에 대해 보다 잘 이해할 수 있는 다른 맥락을 기술해주세요.

JIRA Issue: FR-163

inureyes:
The panels are independent so they do not share the global resource group selection. If we really need this, we can store the selection to the global store.

PR #1362, opened 2022-07-11
fix: show resource statistics about current resource group

dan-2ee:
## Fix
I just added a resource group selection to the summary page. (The design is the same as the Resource Group checkbox on the Session page.)
Now, you can view resources by selecting a resource group on the Summary page.

## Change
<table>
<tr>
</tr>
<tr>
<td>before</td>
<td>after</td>
</tr>
</table>
issue: #1124

### DIFF
--- a/src/components/backend-ai-resource-monitor.ts
+++ b/src/components/backend-ai-resource-monitor.ts
@@ -899,7 +899,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
           <p style="font-size:12px;color:#242424;margin-right:10px;">
             ${_t('session.launcher.ResourceMonitorToggle')}
           </p>
-          <mwc-switch class="fg blue ${this.direction}" id="resource-gauge-switch-button"
+          <mwc-switch selected class="fg blue ${this.direction}" id="resource-gauge-switch-button"
             @click="${(e) => this._toggleResourceGauge(e)}">
           </mwc-switch>
         </div>
--- a/src/components/backend-ai-resource-monitor.ts
+++ b/src/components/backend-ai-resource-monitor.ts
@@ -716,7 +716,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
   _toggleResourceGauge(e) {
     const legend = this.shadowRoot.querySelector('#resource-legend');
     if (e.target.selected) {
-      this.resourceGauge.style.display = 'flex';
+      this.resourceGauge.style.visibility = 'visible';
       if (legend) {
         legend.style.display = 'flex';
       }
@@ -725,7 +725,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
         this.resourceGauge.style.right = '20px';
       }
     } else {
-      this.resourceGauge.style.display = 'none';
+      this.resourceGauge.style.visibility = 'collapse';
       if (legend) {
         legend.style.display = 'none';
       }
--- a/src/components/backend-ai-resource-monitor.ts
+++ b/src/components/backend-ai-resource-monitor.ts
@@ -366,7 +366,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
         .horizontal-card > #resource-gauges {
           display: grid !important;
           grid-auto-flow: row;
-          grid-template-columns: repeat(auto-fill, 420px);
+          grid-template-columns: repeat(auto-fill, 320px);
           justify-content: center;
         }
 
--- a/src/components/backend-ai-resource-monitor.ts
+++ b/src/components/backend-ai-resource-monitor.ts
@@ -443,6 +443,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
 
   async updateScalingGroup(forceUpdate = false, e) {
     await this.resourceBroker.updateScalingGroup(forceUpdate, e.target.value);
+    
     if (this.active) {
       if (this.direction === 'vertical') {
         const scaling_group_selection_box = this.shadowRoot.querySelector('#scaling-group-select-box');
@@ -752,7 +753,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
       </div>
       ` : html``}
       <div class="layout ${this.direction}-card flex wrap">
-        <div id="resource-gauges" class="layout ${this.direction} ${this.direction}-panel resources flex wrap">
+        <div id="resource-gauges" class="layout horizontal ${this.direction}-panel resources flex wrap">
           <div class="layout horizontal center-justified monitor">
             <div class="layout vertical center center-justified resource-name">
               <div class="gauge-name">CPU</div>
--- a/src/components/backend-ai-resource-monitor.ts
+++ b/src/components/backend-ai-resource-monitor.ts
@@ -367,7 +367,6 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
           width: 250px;
         }
   
-
         @media screen and (min-width: 750px) {
           div#resource-gauges {
             display: flex !important;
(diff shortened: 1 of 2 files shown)


src/components/backend-ai-resource-monitor.ts
```diff
@@ -752,7 +753,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
       </div>
       ` : html``}
       <div class="layout ${this.direction}-card flex wrap">
-        <div id="resource-gauges" class="layout ${this.direction} ${this.direction}-panel resources flex wrap">
+        <div id="resource-gauges" class="layout horizontal ${this.direction}-panel resources flex wrap">
```
inureyes: It should be kept to original implementation. If you want to show the panel on another view, it is better to change the CSS part, not the class casting.

src/components/backend-ai-session-view.ts
```diff
@@ -444,7 +444,7 @@ export default class BackendAiSessionView extends BackendAIPage {
     return html`
       <lablup-activity-panel title="${_t('summary.ResourceStatistics')}" elevation="1" autowidth>
         <div slot="message">
-          <backend-ai-resource-monitor location="session" id="resource-monitor" ?active="${this.active === true}"></backend-ai-resource-monitor>
+          <backend-ai-resource-monitor location="session" id="resource-monitor" ?active="${this.active === true}" direction="vertical"></backend-ai-resource-monitor>
```
inureyes: Please explain why you cast the 'direction' on resource monitor.

inureyes (changes requested):

src/components/backend-ai-resource-monitor.ts
dan-2ee: I accepted the feedback and modified it to leave the original implementation as it is. Please review it again 😃
@inureyes

src/components/backend-ai-session-view.ts
dan-2ee: The initial code was written to show "scaling-group-select-box" in the resource monitor only if the direction is vertical.
So I also set direction to vertical to show the select box on the session page.
However, since the select box is shown everywhere using the resource monitor, I modified it to show the select box by default, not by direction.

dan-2ee: Also, unlike other pages (summary, import page), the session page displays the select box horizontally. So I changed the following items.
1. Added css code when direction is horizontal.
2. Change the updateScalingGroupSelector() to operate regardless of direction so that the scaling group can be updated even when the direction is horizontal (Previously, it only works when direction is vertical)

lizable (changes requested):
Let's remove spaces around resource monitor panel in session page to keep consistency of the layout.
Perhaps like this?
Screen Shot 2022-07-11 at 8 32 28 PM

src/components/backend-ai-resource-monitor.ts
```diff
@@ -443,6 +448,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
 
   async updateScalingGroup(forceUpdate = false, e) {
     await this.resourceBroker.updateScalingGroup(forceUpdate, e.target.value);
+
```
agatha197: Please remove this blank line.

src/components/backend-ai-resource-monitor.ts
```diff
@@ -880,6 +881,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
           <div class="layout horizontal center-justified monitor">
             <div class="layout vertical center center-justified resource-name">
               <span class="gauge-name">${_t('session.launcher.Sessions')}</span>
+
```
agatha197: Same here.

src/components/backend-ai-resource-monitor.ts
```diff
@@ -901,7 +903,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
           </mwc-switch>
         </div>
       </div>
-      ${this.direction === 'vertical' ? html`
+
```
agatha197: Same here.

src/components/backend-ai-resource-monitor.ts
```diff
@@ -912,7 +914,7 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
           <span class="resource-legend">${_t('session.launcher.UserResourceLimit')}</span>
         </div>
       </div>
-      ` : html``}
+
```
agatha197: Same here.

src/components/backend-ai-session-view.ts
```diff
@@ -444,7 +444,7 @@ export default class BackendAiSessionView extends BackendAIPage {
     return html`
       <lablup-activity-panel title="${_t('summary.ResourceStatistics')}" elevation="1" autowidth>
         <div slot="message">
-          <backend-ai-resource-monitor location="session" id="resource-monitor" ?active="${this.active === true}"></backend-ai-resource-monitor>
+          <backend-ai-resource-monitor location="session" id="resource-monitor" ?active="${this.active === true}" direction="horizontal"></backend-ai-resource-monitor>
```
agatha197: I think that it's not necessary to cast to horizontal. Because you modified it to show the scaling group select box regardless of direction.

Sujin-Kim1 (changes requested):

src/components/backend-ai-resource-monitor.ts
dan-2ee: I applied it

src/components/backend-ai-resource-monitor.ts
dan-2ee: I applied it

src/components/backend-ai-resource-monitor.ts
dan-2ee: I applied it

src/components/backend-ai-resource-monitor.ts
dan-2ee: I applied it

src/components/backend-ai-session-view.ts
dan-2ee: thank you. I removed the direction cast part. 😁

src/components/backend-ai-resource-monitor.ts
```diff
@@ -357,10 +362,11 @@ export default class BackendAiResourceMonitor extends BackendAIPage {
         .vertical-card > #resource-gauges > .monitor > .resource-name {
           width: 60px;
         }
-
-        .horizontal-card > #resource-gauges > .monitor {
+
```
agatha197: Please remove the blanks.

Sujin-Kim1 (changes requested):

lizable (changes requested):
There are two things to be fixed before merging this PR to main branch.

1. Please update for each component to be vertically aligned. For now `lablup-slider` element in resource statistics panel is slightly moved to the right compared to other elements.
Here are screenshots that refer the issue that I mentioned above:
Screen Shot 2022-07-14 at 4 04 43 PM

Screen Shot 2022-07-14 at 4 05 12 PM

2. Please check whether "see resource monitor" works properly in this branch.
I found that there's a regression in "see resource monitor" switch. Unlikely in main branch, It works OK.
### "see resource monitor" switch in main branch
see-resource-monitor mov

lizable (changes requested):
Let's auto-toggle when the width changes. For now, when a resource gauge showing toggle element appears, the initial status is "off", whereas resource gauges are visible.
Screen Recording 2022-07-15 at 2 04 37 PM mov

Please refer to the code below and apply it.
https://github.com/lablup/backend.ai-webui/blob/1ddb167b53d3e17c42f26d8f8711f9af0661cacf/src/components/backend-ai-webui.ts#L259-L263

lizable (approved):
I think we need slight cosmetic changes. I'll merge this PR after applying them.
- [x] Removing extraneous color style applied in resource gauge toggle button
- [x] Hide/Show not only for the element but also the space of resource gauges
- [x] Forced showing resource gauges when resource statistics panel is in `horizontal` mode.

## Screenshots
resource-gauge-toggle-button-visibility mov

PR #1383, opened 2022-07-19
fix: change the size of the popup to fit the text

dan-2ee:
## Fix
I just increased the length of the dialog to prevent scrolling in the text.

## Change
<table>
<tr>
</tr>

<tr>
<td>after</td>
<td>after</td>
</tr>
</table>
resolve: #1349

### DIFF
--- a/src/components/backend-ai-indicator.ts
+++ b/src/components/backend-ai-indicator.ts
@@ -39,7 +39,7 @@ export default class BackendAIIndicator extends LitElement {
           right: 20px;
           bottom: 20px;
           z-index: 9000;
-          --dialog-height: 80px;
+          --dialog-height: 100px;
           --dialog-width: 250px;
           --dialog-content-padding: 15px;
         }
--- a/src/components/backend-ai-indicator.ts
+++ b/src/components/backend-ai-indicator.ts
@@ -39,7 +39,7 @@ export default class BackendAIIndicator extends LitElement {
           right: 20px;
           bottom: 20px;
           z-index: 9000;
-          --dialog-height: 100px;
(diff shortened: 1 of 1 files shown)


Jaewoook:
If the message is longer than two lines, the same scrolling issue could appear again. I suggest that adaptive component height is better than this.

lizable (changes requested):
I confirmed that if the description in indicator exceeds two lines, then it also creates a scrollbar, [which has been issued](https://github.com/lablup/backend.ai-webui/pull/1383#issuecomment-1189212613) by @Jaewoook.
Instead of using a hardcoded value, let's use `auto`.

lizable:
@dan-2ee And please check whether it works properly in cross-browsers.

lizable:
And once more, it's not necessary, but I think it's better to "auto-resize" the font size to avoid line breaks if there are one or two characters exceeding the total width of snackbar.

dan-2ee:
I think it's better to set the height automatically rather than hard.

dan-2ee:
> And I have one more thing to suggest. it's not necessary, but I think it's better to "auto-resize" the font size to avoid line breaks if there are one or two characters exceeding the total width of snackbar. How does this sound to you? @dan-2ee

It's good to adjust the letter size automatically, but I think it's also a way to prevent words from being separated by letters. So I added the "word-break" property to prevent words from being separated from unacceptable positions.
Please review it.

lizable (approved):
LGTM.

dan-2ee:
<table>
<tr>
</tr>

<tr>
<td>after</td>
<td>after</td>
</tr>

</table>

PR #1385, opened 2022-07-22
fix: show the modified registry information

### LINKED ISSUE #1346: Don't show the modified Registries information after modify registry information
**Describe the bug**
문제가 발생하는 상황에 대한 요약 설명을 적어주세요.
Don't show the modified Registries information after modify registry information 

**To Reproduce**
문제를 재현하기 위한 순서를 자세히 적어주세요:
1. Go to 'Environments' menu (needs Administration level)
2. Click on 'Registry' tab
3. Click modify Registry button 
4. Show Modify Registries popup, Try to modify registries information & save it
5. retry to Click modify Registry button 
6. See an error

**Expected behavior**
원래 기대했던 동작은 어떤 것인지 명확하게 적어주세요.

Don't show the registry information on Modify Registry tab

**Screenshots**
Just take a screenshot in your clipboard or copy a picture file and paste it here (Ctrl+V/Cmd+V) so that GitHub automatically attach and insert it.
가능하다면, 문제 상황을 보여주는 스크린샷을 첨부해주세요.
클립보드에 스크린샷을 찍거나 이미지 파일을 복사해서 여기에 붙여넣으면(Ctrl+V/Cmd+V) GitHub이 자동으로 이곳에 첨부 및 삽입해줍니다.

For the first time, Modify Registry information
스크린샷 2022-07-05 오후 12 12 40

After modified Registry information, try to click and show Modify Registrty popup.
It shows that don't show the modified Registries information after modify registry information 
스크린샷 2022-07-05 오후 12 12 56

https://user-images.githubusercontent.com/20002/177250248-9c87df78-693d-4c0c-b684-8eb70f13f2e2.mov

**Client:**
 - OS: macOS 12.4 
 - App version: [e.g. 19.08.3] (if you are using a GUI client app)
 - Browser version: Chrome 103.0.5060.53
 - Python version: 3.10.5
 - NodeJS version: 18.4

**Server:**
 - OS: macOS 12.4 
   - Output of `uname -a`: `Darwin seongdacBookPro.office.lablup 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:37 PDT 2022; root:xnu-8020.121.3~4/RELEASE_ARM64_T6000 arm64`
 - Backend.AI version: main
 - Python version: 3.10.5
 - Installation method: [ used scripts/install-dev.sh ]

**Additional context**
추가적으로 문제 상황을 설명할 수 있는 다른 맥락을 기술해주세요.

dan-2ee:
## Fix
* avoid resetting registry fields when in edit mode.
* Shows the currently selected registry type when in edit mode (previously, docker was unconditionally selected)

## Change
https://user-images.githubusercontent.com/65989401/180361854-bfa0a15b-9ec9-421f-809d-e8eab34f2ab7.mov

resolve #1346


studioego:
I checked the source code. It works good!
In my opinion, It can merge it.
LGTM. 
@Sujin-Kim1 , @lizable  If you check the PR, Please, merge it.

studioego (commented):
LGTM

src/components/backend-ai-registry-list.ts
```diff
@@ -689,7 +690,7 @@ class BackendAIRegistryList extends BackendAIPage {
                       validationMessage="${_t('registry.PleaseSelectOption')}"
                       value="${this.registryList[this.selectedIndex]?.type || this.registryType}">
             ${this._registryTypes.map((item) => html`
-              <mwc-list-item value="${item}" ?selected="${item === 'docker'}">${item}</mwc-list-item>
+              <mwc-list-item value="${item}" ?selected="${this.editMode ? item === this.registryList[this.selectedIndex]?.type : item === 'docker'}">${item}</mwc-list-item>
```
lizable: And If the line is too long horizontally, then perhaps you could add line breaks for better readability.

lizable (changes requested):
I have one more request about indentations.

dan-2ee:
> I confirmed that the code works fine, but It would be better to change the id of the dialog since it does two operations in the registry, "add" and "modify". Currently, the id of the dialog is `add-registry-dialog`, but perhaps we can change it to `configure-registry-dialog`. If so, those who read the code don't have to spend time finding `modify-registry-dialog`.

I agree. I reflected it

src/components/backend-ai-registry-list.ts
```diff
@@ -227,22 +227,22 @@ class BackendAIRegistryList extends BackendAIPage {
    * */
   _addRegistry() {
     // somehow type casting is needed to prevent errors, unlike similar use cases in other files
```
lizable, line 245: Speaking of the type, How about declaring them as a member variable with `@query` decorator and access modifier(perhaps `private`)? I think we don't need to expose them to avoid unexpected value changes outside of the component. And If we convert them to member variables, we don't need to call `querySelector` or `getElementbyId` functions.

lizable (changes requested):
How about converting local variables to member variables with an access modifier?

src/components/backend-ai-registry-list.ts
dan-2ee: > How about converting local variables to member variables with an access modifier?

I think it's good! Are you saying that we should change the variables necessary for registry information such as hostname and url into member variables?

dan-2ee:
> How about converting local variables to member variables with an access modifier?

I think it's good! Are you saying that we should change the variables necessary for registry information such as hostname and url into member variables?

src/components/backend-ai-registry-list.ts
dan-2ee: > Speaking of the type, How about declaring them as a member variable with `@query` decorator and access modifier(perhaps `private`)? I think we don't need to expose them to avoid unexpected value changes outside of the component. And If we convert them to member variables, we don't need to call `querySelector` or `getElementbyId` functions.

I applied it. 
I replaced the variables for registry information with member variables with access modifier(private).

src/components/backend-ai-registry-list.ts
```diff
@@ -51,6 +51,13 @@ class BackendAIRegistryList extends BackendAIPage {
   @property({type: String}) registryType = 'docker';
```
lizable: We already defined `registryType` in L57 (`registerType` seems like typo error), but there's still `registryType` property declaration left in L51. Please remove the property unless you need to expose them in public for a certain reason.

src/components/backend-ai-registry-list.ts
```diff
@@ -51,6 +51,13 @@ class BackendAIRegistryList extends BackendAIPage {
...
   @property({type: Boolean}) editMode;
```
lizable: let's insert `_`(underscore) at the beginning of every `private` variable to follow the underlying de facto code convention.

src/components/backend-ai-registry-list.ts
lizable: Also, I think we don't need to declare `_registryTypes` as "property" since It doesn't do any array manipulation (such as insert, add, append, or pop) in overall components. Let's just declare it as a `private` and `static` member variable.

lizable: Also, Since accessing to attributes of the domain is only allowed to "superadmin" role(refer [here](https://github.com/lablup/backend.ai/blob/main/src/ai/backend/manager/models/domain.py)), which should be protected from any manipulation occurred from outside of the component.
IMO, It's better to declare member variable with `@state()` and `private`.

lizable: And I wonder whether we need to add `public` access modifier to `registryList`. If there's no specific reason to access from outside of the component, let's replace the access modifier with `private`.

src/components/backend-ai-registry-list.ts
```diff
@@ -227,22 +234,22 @@ class BackendAIRegistryList extends BackendAIPage {
...
-    if (!this.shadowRoot.querySelector('#add-registry-hostname').valid) {
+    const hostname = this.hostname.value;
+    const url = this.url.value;
+    const username = this.username.value;
+    const password = this.password.value;
+    const registerType = this.registerType.value;
+    const projectName = this.projectName.value.replace(/\s/g, '');
+
+    if (!this.shadowRoot.querySelector('#configure-registry-hostname').valid) {
       const validationMessage = this.shadowRoot.querySelector('#registry-hostname-validation');
```
lizable: I think we can use the member variable instead of declaring and assigning local variables here.
And in L244, we already declared member variable for [querying `#configure-registry-hostname` in L54](https://github.com/lablup/backend.ai-webui/pull/1385/files#diff-4d9956da4424c56f3eca5366ff7485b320c865897c146fb86cc88c9f51d1239eR54), so let's use them instead of calling `querySelector` explicitly.

src/components/backend-ai-registry-list.ts
```diff
@@ -227,22 +234,22 @@ class BackendAIRegistryList extends BackendAIPage {
...
       if (validationMessage) {
         validationMessage.style.display = 'block';
       }
       return;
     }
 
-    if (!this.shadowRoot.querySelector('#add-registry-url').valid) {
+    if (!this.shadowRoot.querySelector('#configure-registry-url').valid) {
       const validationMessage = this.shadowRoot.querySelector('#registry-url-validation');
```
lizable: Same here.

src/components/backend-ai-registry-list.ts
```diff
@@ -452,12 +460,12 @@ class BackendAIRegistryList extends BackendAIPage {
   }
 
   _resetRegistryField() {
```
lizable: Same suggestion here. I think it's not necessary to declare and assign member variables to local variables.

src/components/backend-ai-registry-list.ts
```diff
@@ -408,23 +415,24 @@ class BackendAIRegistryList extends BackendAIPage {
...
   }
 
   _validateUrl() {
-    const url = this.shadowRoot.querySelector('#add-registry-url');
+    const url = this.url;
     const validationMessage = this.shadowRoot.querySelector('#registry-url-validation');
     validationMessage.style.display = url.valid ? 'none' : 'block';
   }
 
   _validateHostname() {
```
lizable: Same here.

lizable (changes requested):
I have several change requests mostly related to duplicated (or originated from typo error) variables declaration and minor code conventions.
I know that we need many changes like this in other components as well(so It could be great to change them simultaneously), but let's change this component first as much as we can so that we can make a good example for refactoring other components.

src/components/backend-ai-registry-list.ts
dan-2ee: > We already defined `registryType` in L57 (`registerType` seems like typo error), but there's still `registryType` property declaration left in L51. Please remove the property unless you need to expose them in public for a certain reason.

The registryType declared as a property and the member variable registerType are used for different purposes. Therefore, I have clearly changed the name of the register type without removing either one.

dan-2ee: and I applied it. please check again 😀

src/components/backend-ai-registry-list.ts
dan-2ee: I reflected it

src/components/backend-ai-registry-list.ts
dan-2ee: I reflected it

src/components/backend-ai-registry-list.ts
dan-2ee: I reflected it

src/components/backend-ai-registry-list.ts
dan-2ee: I reflected it.

src/components/backend-ai-registry-list.ts
dan-2ee: I reflected it

src/components/backend-ai-registry-list.ts
lizable: > Also, I think we don't need to declare `_registryTypes` as "property" since It doesn't do any array manipulation (such as insert, add, append, or pop) in overall components. Let's just declare it as a `private` and `static` member variable.

I guess the request hasn't been applied yet. I wonder if there's any specific reason for leaving them as is.
From my point of view, we don't need to assign it twice, since we don't do any manipulation on this `_registryTypes` overall component. How about assigning value like this instead of the current code?
~~~
private static _registryTypes = ['docker', 'harbor', 'harbor2'];
~~~
If we change the code like this, then we need to do two changes accordingly:
- Replace `this` for accessing the _registryTypes to class `BackendAIRegistryList`. (in [L697](https://github.com/lablup/backend.ai-webui/pull/1385/files#diff-4d9956da4424c56f3eca5366ff7485b320c865897c146fb86cc88c9f51d1239eR697))
- Remove value assignment in `_viewStateChanged()` function

src/components/backend-ai-registry-list.ts
```diff
@@ -1,66 +1,80 @@
...
-import {default as PainKiller} from './backend-ai-painkiller';
-import {BackendAiStyles} from './backend-ai-general-styles';
-import {IronFlex, IronFlexAlignment} from '../plastics/layout/iron-flex-layout-classes';
+import '@vaadin/vaadin-grid/vaadin-grid';
+
+import 'weightless/button';
+import 'weightless/card';
+import 'weightless/icon';
+import {Label as WlLabel} from 'weightless/label';
+import {Textfield as WlTextfield} from 'weightless/textfield';
```
lizable, line 32: For avoiding confusion with other textfield components(such as `mwc-textfield`), I aliased every element of `weightless` starts with `Wl`.

src/components/backend-ai-registry-list.ts
```diff
@@ -1,66 +1,80 @@
...
 
-@group Backend.AI Web UI
- @element backend-ai-release-check
- */
+  @group Backend.AI Web UI
+  @element backend-ai-release-check
+*/
 
 @customElement('backend-ai-registry-list')
 class BackendAIRegistryList extends BackendAIPage {
```
lizable, line 43: As you can see in the member variable declaration, there's no need to expose any variable as public.
Therefore we could use at most one decorator, `@query`.
And of course, for better code readability, I add an access modifier with `_`(underscore) prefix on private variables (and functions).

src/components/backend-ai-registry-list.ts
```diff
@@ -1,66 +1,80 @@
...
+  @query('#project-name-validation') private _projectNameValidationMsg!: WlLabel;
 
   constructor() {
     super();
-    this.registryList = [];
-    this._registryTypes = [];
-    this.allowed_registries = [];
-    this.hostnames = [];
-    this.editMode = false;
+    this._allowed_registries = [];
```
lizable, line 70: When assigning initial values to member variables inside `constructor()`, I rearranged them in alphabetical order.

src/components/backend-ai-registry-list.ts
```diff
@@ -1,66 +1,80 @@
...
+    this._editMode = false;
+    this._hostnames = [];
+    this._registryList = [];
+    this._registryList = [];
   }
```
lizable, line 75: I add access modifier(`public`) and keyword(`override`) by referring `styles` declaration on superclass ([ReactiveElement in Lit](https://github.com/lit/lit/blob/main/packages/reactive-element/src/reactive-element.ts)).

src/components/backend-ai-registry-list.ts
```diff
@@ -150,9 +164,13 @@ class BackendAIRegistryList extends BackendAIPage {
       `];
   }
```
lizable, line 159: Please refer to more detailed information about the keyword and access modifier in [here](https://github.com/lit/lit/blob/1ac58058fbdf58633b2bb8d10649d0e11c409256/packages/reactive-element/src/reactive-element.ts#L1468-L1485).

src/components/backend-ai-registry-list.ts
```diff
@@ -175,15 +193,18 @@ class BackendAIRegistryList extends BackendAIPage {
...
+      this._allowed_registries = response.domain.allowed_docker_registries;
       return globalThis.backendaiclient.registry.list();
     }).then(({result}) => {
-      this.registryList = this._parseRegistryList(result);
-      this.hostnames = this.registryList.map((value) => {
+      this._registryList = this._parseRegistryList(result);
+      this._hostnames = this._registryList.map((value) => {
         return value.hostname;
-      });
+      }) as string[];
```
lizable, line 200: Type casting is essential since plain `map()` function could returns `unknown[]`.

src/components/backend-ai-registry-list.ts
```diff
@@ -193,7 +214,7 @@ class BackendAIRegistryList extends BackendAIPage {
    *
    * @param {Boolean} active - whether view state change or not
    * */
```
lizable, line 209: Please refer [`backend-ai-page`](https://github.com/lablup/backend.ai-webui/blob/1f29d3ead9212acea56db011e3d05453b38dacae/src/components/backend-ai-page.ts#L46) component.

src/components/backend-ai-registry-list.ts
```diff
@@ -202,50 +223,50 @@ class BackendAIRegistryList extends BackendAIPage {
...
+        this._refreshRegistryList();
       }, true);
     } else { // already connected
       this._refreshRegistryList();
-      this._registryTypes = ['docker', 'harbor', 'harbor2'];
     }
   }
 
   /**
    * Get host name from url
```
lizable, line 227: If there's a function that doesn't use anymore, then we can add `@deprecated` decorator explicitly.

src/components/backend-ai-registry-list.ts
```diff
@@ -277,36 +298,37 @@ class BackendAIRegistryList extends BackendAIPage {
...
           this._refreshRegistryList();
         } else {
           this.notification.text = _text('dialog.ErrorOccurred');
         }
-        this._hideDialogById('#add-registry-dialog');
-        this._resetRegistryField();
+        this._hideDialogById('#configure-registry-dialog');
         this.notification.show();
       });
   }
```
lizable, line 310: Updated function description with more detailed contents.

src/components/backend-ai-registry-list.ts
```diff
@@ -320,16 +342,16 @@ class BackendAIRegistryList extends BackendAIPage {
       this.notification.show();
     }
     // remove written hostname
```
lizable, line 337: Use local variable instead of calling multiple `querySelector()`.

src/components/backend-ai-registry-list.ts
```diff
@@ -320,16 +342,16 @@ class BackendAIRegistryList extends BackendAIPage {
...
+    deleteRegistryInputField.value = '';
   }
 
   /**
    * Rescan the images if registry has been updated.
    * */
-  async _rescanImage() {
-    const indicator = await this.indicator.start('indeterminate');
+  private async _rescanImage() {
+    const indicator: BackendAIIndicator = await this._indicator.start('indeterminate');
```
lizable, line 345: Type annotation in indicator element. Please notice that `this._indicator` is not [`BackendAIIndicator`](https://github.com/lablup/backend.ai-webui/blob/1f29d3ead9212acea56db011e3d05453b38dacae/src/components/backend-ai-indicator.ts#L21) itself, but [`BackendAIIndicatorPool`](https://github.com/lablup/backend.ai-webui/blob/1f29d3ead9212acea56db011e3d05453b38dacae/src/components/backend-ai-indicator-pool.ts#L30).

src/components/backend-ai-registry-list.ts
```diff
@@ -373,101 +395,115 @@ class BackendAIRegistryList extends BackendAIPage {
       });
   }
```
lizable, line 390: Add missing description of the function.

src/components/backend-ai-registry-list.ts
```diff
@@ -373,101 +395,115 @@ class BackendAIRegistryList extends BackendAIPage {
...
 
-  _hideValidationMessage() {
-    this.shadowRoot.querySelector('#registry-hostname-validation').style.display = 'none';
-    this.shadowRoot.querySelector('#registry-url-validation').style.display = 'none';
-    this.shadowRoot.querySelector('#project-name-validation').style.display = 'none';
+  /**
+   * Reset validation message in registry configuation dialog
+   */
+  private _resetValidationMessage() {
```
lizable, line 424: It's a bit getting my nerves that repetitive value manipulation in here, but let's leave it for now with `FIXME` comments.

src/components/backend-ai-registry-list.ts
```diff
@@ -373,101 +395,115 @@ class BackendAIRegistryList extends BackendAIPage {
...
-
-  _toggleProjectNameInput() {
-    this.registryType = this.shadowRoot.querySelector('#select-registry-type').value;
-    this._validateProjectName();
+    this._registryList[this._selectedIndex] = registryInfo;
+    this._registryType = this._registryList[this._selectedIndex]?.type as string;
+    this.requestUpdate();  // call for explicit update
+    this._resetValidationMessage();
+    this._launchDialogById('#configure-registry-dialog');
   }
```
lizable, line 457: The function doesn't do anything for validation but just toggles the visibility of the message of corresponding input field. So I changed the function name to a more precise name.

src/components/backend-ai-registry-list.ts
```diff
@@ -373,101 +395,115 @@ class BackendAIRegistryList extends BackendAIPage {
...
-  _validateUrl() {
-    const url = this.shadowRoot.querySelector('#add-registry-url');
-    const validationMessage = this.shadowRoot.querySelector('#registry-url-validation');
-    validationMessage.style.display = url.valid ? 'none' : 'block';
+  /**
+   * Hide/Show validation msg on url input field in registry configuration dialog
+   */
+  private _toggleValidationMsgOnUrlInput() {
+    this._registryUrlValidationMsg.style.display = this._urlInput.valid ? 'none' : 'block';
   }
```
lizable, line 464: Same here.

src/components/backend-ai-registry-list.ts
```diff
@@ -373,101 +395,115 @@ class BackendAIRegistryList extends BackendAIPage {
...
-    } else {
-      validationMessage.style.display = 'block';
-    }
+  /**
+   * Hide/Show validation msg on hostname input in registry configuration dialog
+   */
+  private _toggleValidationMsgOnHostnameInput() {
+    const hostname = this._hostnameInput.value;
+    this._registryHostnameValidationMsg.style.display = (hostname && hostname !== '') ? 'none' : 'block';
   }
```
lizable, line 472: Same here as well.

src/components/backend-ai-registry-list.ts
```diff
@@ -373,101 +395,115 @@ class BackendAIRegistryList extends BackendAIPage {
...
       }
-      projectTextEl.disabled = false;
+      this._projectNameInput.disabled = false;
     } else {
-      validationEl.textContent = _text('registry.ForHarborOnly');
-      projectTextEl.disabled = true;
+      this._projectNameValidationMsg.textContent = _text('registry.ForHarborOnly');
+      this._projectNameInput.disabled = true;
     }
   }
```
lizable, line 492: Same here.

src/components/backend-ai-registry-list.ts
```diff
@@ -476,30 +512,60 @@ class BackendAIRegistryList extends BackendAIPage {
...
 
   /**
```
lizable, line 507: Same here.

src/components/backend-ai-registry-list.ts
```diff
@@ -565,42 +655,38 @@ class BackendAIRegistryList extends BackendAIPage {
...
       html`
-        <div
-          icon="settings"
-          id="controls"
-          class="layout horizontal flex center"
-        >
+        <div icon="settings" id="controls" class="layout horizontal flex center">
           <wl-button fab flat inverted
               class="fg blue"
               @click=${() => {
```
lizable: @dan-2ee 
Former indentation degrades readability. It might be originated from current linting rules.
Let's double-check whether the indentation is correct after auto-fixing via the linting command.
If something went wrong, please revert to the original indentation.

src/components/backend-ai-registry-list.ts
```diff
@@ -609,7 +695,14 @@ class BackendAIRegistryList extends BackendAIPage {
     );
   }
```
lizable, line 690: Please refer [here(`render()` declared in Lit-Element)](https://github.com/lit/lit/blob/1ac58058fbdf58633b2bb8d10649d0e11c409256/packages/lit-element/src/lit-element.ts#L217-L227).

src/components/backend-ai-registry-list.ts
```diff
@@ -638,87 +731,82 @@ class BackendAIRegistryList extends BackendAIPage {
...
         </vaadin-grid-column>
       </vaadin-grid>
-      <backend-ai-dialog id="add-registry-dialog" fixed backdrop blockscrolling>
-      ${this.editMode ? html`
-        <span slot="title">${_t('registry.ModifyRegistry')}</span>
-      ` : html`
-        <span slot="title">${_t('registry.AddRegistry')}</span>
-      `}
+      <backend-ai-dialog id="configure-registry-dialog" fixed backdrop blockscrolling>
+        <span slot="title">${this._editMode ? _t('registry.ModifyRegistry') : _t('registry.AddRegistry')}</span>
```
lizable, line 727: Minor refactoring for reducing HTML codes.

src/components/backend-ai-registry-list.ts
```diff
@@ -638,87 +731,82 @@ class BackendAIRegistryList extends BackendAIPage {
...
         <div slot="footer" class="horizontal center-justified flex layout">
-          ${this.editMode ? html`
-            <mwc-button unelevated fullwidth icon="add" label="${_t('button.Save')}"
-            @click=${this._addRegistry}></mwc-button>
-          ` : html`
-            <mwc-button unelevated fullwidth icon="add" label="${_t('button.Add')}"
-            @click=${this._addRegistry}></mwc-button>
-          `}
+          <mwc-button unelevated fullwidth icon="add"
+            label=${this._editMode ? _t('button.Save') : _t('button.Add')}
```
lizable, line 801: Same here.

lizable (commented):
I just applied three major changes and two minor changes to this component which will be reflected in other components near future.

#### Major changes
- Type annotation and Type casting (if needed)
- Access modifier
- Keywords (`override`)

#### Minor changes
- Add description (including parameter and return value) in function for JSDoc
- Rename the function name to fit into the operation inside.

Although there are some parts that need more attention (such as repetitive value manipulations, etc.), I think those changes are acceptable for now.

Before merging this, I need approval from @inureyes  or @adrysn.
Please review these code changes if possible.

src/components/backend-ai-registry-list.ts
Jaewoook: @lizable It seems there are same assignment across two lines.

`this._registryList = [];`

src/components/backend-ai-registry-list.ts
```diff
@@ -1,66 +1,80 @@
...
-  @property({type: Array}) _registryTypes;
-  @property({type: Array}) allowed_registries;
-  @property({type: Array}) hostnames;
-  @property({type: String}) registryType = 'docker';
-  @property({type: Boolean}) editMode;
+  private _allowed_registries: Array<object>;
+  private _editMode: boolean = false;
+  private _hostnames: Array<string>;
+  private _indicator!: BackendAIIndicatorPool;
+  private _registryList: Array<Record<string, unknown>>;
```
Jaewoook, line 48: using unknown type is not recommended because the type is not clear. How about define the value type manually?

src/components/backend-ai-registry-list.ts
Jaewoook: The return type of `map` function follows the original value type from array. In this case, it seems the `this._registryList` variable's value typed as `unknown`.As the result, map function returns unknown[] type. You can type the `map`'s return type by doing like this: `map<OtherType>(...)`. I suggest using like this.

src/components/backend-ai-registry-list.ts
lizable: In that case, we could consider using [`Partial`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) type since [the type of value is optional on the server-side.](https://github.com/lablup/backend.ai/blob/141844d383970b7274b9861b625d7f51a6601998/src/ai/backend/common/etcd.py#L300-L371)

dan-2ee:
> I just applied three major changes and two minor changes to this component which will be reflected in other components near future.
> 
> #### Major changes
> * Type annotation and Type casting (if needed)
> * Access modifier
> * Keywords (`override`)
> 
> #### Minor changes
> * Add description (including parameter and return value) in function for JSDoc
> * Rename the function name to fit into the operation inside.
> 
> Although there are some parts that need more attention (such as repetitive value manipulations, etc.), I think those changes are acceptable for now.
> 
> Before merging this, I need approval from @inureyes or @adrysn. Please review these code changes if possible.

https://user-images.githubusercontent.com/65989401/182762214-d2cadf16-e611-41c9-a2ac-6fc28be3eca4.mov

After some changes, the modify function does not work properly. Please make sure it doesn't work only in my environment

inureyes (approved):
LGTM

PR #1386, opened 2022-07-22
fix: disable resource policy control button for domain admin

Jaewoook:
## Description

In the Resource Policy panel, only the admin can create / modify / delete the policies but originally, domain admin should also have same permission to execute action about resource policies. But current version of [Backend.AI](https://github.com/lablup/backend.ai), this feature does not implement yet. Until the feature is implemented, I updated the resource policy control panel's button to disable for domain admin user.

related issue: #1375 

## Changes

- add style for button disabled state
- improve control button render logic not to do conditional rednering
- add super admin check for button state

### Extra Information:

I deleted the icon prop of wl-button component because the button does not receive icon prop. This can find at [here](https://github.com/andreasbm/weightless/tree/master/src/lib/button#-properties).


lizable (approved):
LGTM.

PR #1395, opened 2022-07-27
fix: parse unicode strings for some items in the toml config

adrysn:
Recent server PR https://github.com/lablup/backend.ai/pull/599 introduces bare unicode for some text items in composing the toml config file. For example, if an admin sets a string `JPark's Test Node` for `api.text` item in `webserver.conf`, the the resulting text would be `JPark\u0027s Test Node`.

image

This PR manually handles this case by parsing the unicode text in loading the config.

Affected config items:
- `config.general.apiEndpointText`
- `config.general.siteDescription`


Sujin-Kim1 (approved):
LGTM

Issue #1387, opened 2022-07-25
Let's enable button only if user input is valid on every button inside panels in Import & Run page.


lizable:
**Is your feature request related to a problem? Please describe.**
기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.
For users, when any button has a ripple effect on hover or click, then they may easily think there would be an operation occur (such as open dialog or modal, etc.) on click event. Currently click event only reacts when input is not empty, which is likely to bring unintended confusion. And there's a "ripple event" around every button on Import & Run page regardless of inputs.

**Describe the solution you'd like**
어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.
Let's enable the button only if the user input is valid. if not(like empty or wrong), disabling the button would be enough.

**Describe alternatives you've considered**
혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.
If disabling the button seems to be too strict, displaying both error messages on snackbar and near input field would be the alternative way to inform user input is wrong.

**Additional context**
기능 요청에 대해 보다 잘 이해할 수 있는 다른 맥락을 기술해주세요.
This issue has been raised during code-reviewing and E2E tests on #1359.

### Screenshots
![import_run_page_buttons](https://user-images.githubusercontent.com/46954439/180721999-446874f1-7370-4693-a2c2-e23da7a8a4c6.gif)

lizable:
Closing this issue since it's been resolved in #1400.

PR #1400, opened 2022-07-30
fix: add validation check and enable button only valid status in import and run page

mihilt:
#1387

To solve this problem, I created a variable called `importButtonDisabled`.
Checking length of textfield's input value through `@change` property and it can handle `importButtonDisabled` value properly.
The button dynamically changes the property called `disabled` according to the variable.

### DIFF
--- a/src/components/backend-ai-import-view.ts
+++ b/src/components/backend-ai-import-view.ts
@@ -46,7 +46,6 @@ export default class BackendAIImport extends BackendAIPage {
   @property({type: Object}) indicator = Object();
   @property({type: Object}) notification = Object();
   @property({type: Object}) sessionLauncher = Object();
-  @property({type: Object}) resourcePolicy;
   @property({type: String}) requestURL = '';
   @property({type: String}) queryString = '';
   @property({type: String}) environment = 'python';
@@ -61,7 +60,6 @@ export default class BackendAIImport extends BackendAIPage {
   @property({type: String}) _helpDescription = '';
   @property({type: String}) _helpDescriptionTitle = '';
   @property({type: String}) _helpDescriptionIcon = '';
-  @property({type: Object}) storageProxyInfo = Object();
 
   constructor() {
     super();
--- a/src/components/backend-ai-import-view.ts
+++ b/src/components/backend-ai-import-view.ts
@@ -61,6 +61,7 @@ export default class BackendAIImport extends BackendAIPage {
   @property({type: String}) _helpDescriptionTitle = '';
   @property({type: String}) _helpDescriptionIcon = '';
   @property({type: Object}) storageProxyInfo = Object();
+  @property({type: Boolean}) importButtonDisabled = true;
 
   constructor() {
     super();
@@ -581,6 +582,10 @@ export default class BackendAIImport extends BackendAIPage {
     }
   }
 
+  notebookUrlChanged(e) {
+    this.importButtonDisabled = !(e.target.value.length > 0);
+  }
+
   render() {
     // language=HTML
     return html`
@@ -589,8 +594,8 @@ export default class BackendAIImport extends BackendAIPage {
         <div slot="message">
           <div class="horizontal wrap layout center">
             <mwc-textfield id="notebook-url" label="${_t('import.NotebookURL')}"
-                           maxLength="2048" placeholder="${_t('maxLength.2048chars')}"></mwc-textfield>
-            <mwc-button icon="cloud_download" @click="${() => this.getNotebookFromURL()}">
+                           maxLength="2048" placeholder="${_t('maxLength.2048chars')}" @change="${(e) => this.notebookUrlChanged(e)}"></mwc-textfield>
+            <mwc-button ?disabled="${this.importButtonDisabled}" icon="cloud_download" @click="${() => this.getNotebookFromURL()}">
               <span>${_t('import.GetAndRunNotebook')}</span>
             </mwc-button>
           </div>
--- a/src/components/backend-ai-import-view.ts
+++ b/src/components/backend-ai-import-view.ts
@@ -594,7 +594,7 @@ export default class BackendAIImport extends BackendAIPage {
         <div slot="message">
           <div class="horizontal wrap layout center">
             <mwc-textfield id="notebook-url" label="${_t('import.NotebookURL')}"
-                           maxLength="2048" placeholder="${_t('maxLength.2048chars')}" @change="${(e) => this.notebookUrlChanged(e)}"></mwc-textfield>
+                           maxLength="2048" placeholder="${_t('maxLength.2048chars')}" @keyup="${(e) => this.notebookUrlChanged(e)}"></mwc-textfield>
             <mwc-button ?disabled="${this.importButtonDisabled}" icon="cloud_download" @click="${() => this.getNotebookFromURL()}">
               <span>${_t('import.GetAndRunNotebook')}</span>
             </mwc-button>
--- a/src/components/backend-ai-import-view.ts
+++ b/src/components/backend-ai-import-view.ts
@@ -50,7 +50,8 @@ export default class BackendAIImport extends BackendAIPage {
   @property({type: String}) requestURL = '';
   @property({type: String}) queryString = '';
   @property({type: String}) environment = 'python';
-  @property({type: String}) importMessage = '';
+  @property({type: String}) importNotebookMessage = '';
+  @property({type: String}) importGithubMessage = '';
   @property({type: String}) importGitlabMessage = '';
   @property({type: Array}) allowedGroups = [];
   @property({type: Array}) allowed_folder_type = [];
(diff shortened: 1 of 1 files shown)


mihilt:
This comment is about the second commit.
I checked in the first commit that the function what i made works only when i move focus off the textfield.
So I changed `@change` to `@keyup`.

mihilt:
I am sorry, I just found that i need to fix three more buttons.

I will fix the remaining buttons in the same way as I committed before.
But I don't know that this way is right. Can someone give me an advice?

lizable:
> I am sorry, I just found that I need to fix three more buttons.
> 
> I will fix the remaining buttons in the same way as I committed before. But I don't know that this way is right. Can someone give me an advice?

Hi, @mihilt! 
First of all, Thanks for making your effort on this open-source project!
Speaking of your request, I think it's better to describe the issue with several tasks so that you can easily understand what to do.

In this issue, there are two tasks to achieve.
1. Add validation logic(including showing up message with snackbar UI) in two parts:
   - [ ] input field in Import Notebook panel
   - [ ] input field in Create Notebook button panel
2. Enable each button on "Import & Run" page only when the corresponding input value is valid.

FWIW, I'm not sure about adding a property for each button. Could we just set it as disabled at the first render and change it when the input field is valid?

mihilt:
Thank you for advicing @lizable.
I'm sorry. I should have read the issue in more detail. I thought the thing about snackbar was optional.

BTW, I have two questions.

> - [ ] input field in Create Notebook button panel

I think the input field in **Create Notebook Button** panel already has its validation logic.
Cuz I saw snackbar with message when I clicked button with empty input field in this panel.

And I didn't found any snackbar with message when I clicked button with empty input field in **Import Github Repository** panel.
So I think I should add validation logic in **Import Github Repository** panel instead of **Create Notebook Button** panel.
Is it right?

> FWIW, I'm not sure about adding a property for each button. Could we just set it as disabled at the first render and change it when the input field is valid?
> 

I'm not sure, but In this case, is there have a way to handle if the valid input value is changed back to an invalid input value?
(In more detail, if it's textfield's value goes from invalid to valid and invalid again. In this situation, I think the button can be clicked even if there has a problem with the input value.)
I think each button needs their status value. So it needs its own property with each separated variable.

I think I bothered you a little bit with a lot of questions even if this issue has **easy** and **good first issue** label.
It's because I'm a beginner contributor in this repo. I'm sorry.

lizable:
> 



> Thank you for advicing jihyunkang (@lizable). I'm sorry. I should have read the issue in more detail. I thought the thing about snackbar was optional.
> 
> BTW, I have two questions.
> 
> > * [ ]  input field in Create Notebook button panel
> 
> I think the input field in **Create Notebook Button** panel already has its validation logic. Cuz I saw snackbar with message when I clicked button with empty input field in this panel.
> 
> And I didn't found any snackbar with message when I clicked button with empty input field in **Import Github Repository** panel. So I think I should add validation logic in **Import Github Repository** panel instead of **Create Notebook Button** panel. Is it right?
> 
> > FWIW, I'm not sure about adding a property for each button. Could we just set it as disabled at the first render and change it when the input field is valid?
> 
> I'm not sure, but In this case, is there have a way to handle if the valid input value is changed back to an invalid input value? (In more detail, if it's textfield's value goes from invalid to valid and invalid again. In this situation, I think the button can be clicked even if there has a problem with the input value.) I think each button needs their status value. So it needs its own property with each separated variable.
> 
> I think I bothered you a little bit with a lot of questions even if this issue has **easy** and **good first issue** label. It's because I'm a beginner contributor in this repo. I'm sorry.

You don't need to feel sorry about it. If you have anything uncertain while handling the issue, It's okay to ask me anytime.
My answer to the first question is that we need to apply validation logic not only for the empty string but also the format.
As long as I know, only opened URL (accessible in public) for notebook file is available for making an instant session and a badge button. Therefore you need to apply validation logic that checks whether the user input is empty and also whether the user input is valid URL(e.g. starting from `https://`). This code block below may help you to figure out the regex for valid URL.
https://github.com/lablup/backend.ai-webui/blob/77f284acdb6e7b88d70eee772d608bbfef9ee39b/src/components/backend-ai-registry-list.ts#L659

And I think we don't need to care about frequent state changing of input field, (including as you mentioned above, invalid->valid->invalid) when we replace `keyup` event handler to `change`. Unlikely to `keyup` event, `change` event triggered when user input field is blurred, which refers that the user input has been finished.
Therefore I suggest using `change` event rather than `keyup` event, like most input field in backend.ai-webui do.

lizable (changes requested):
Even if the URL is valid, input field keep saying it's invalid.
Here's the URL address that I used:
- jupyter notebook file shared in public
https://github.com/hephaex/deeplearning-note/blob/master/08_dimensionality_reduction.ipynb
- GitHub address shared in public
https://github.com/rickiepark/handson-ml2.git



### Screenshots
not-working mov

mihilt:
Thank you for telling me. I changed regex.

https://user-images.githubusercontent.com/50368758/184861995-76c0e18b-f93f-4907-b322-4974f29ae939.mp4

inureyes:
@mihilt mwc-textfield has custom validation methods such as `checkValidity()`, `reportValidity()` and `setCustomValidity()` Please refer the link https://www.npmjs.com/package/@material/mwc-textfield and revise the implementation to use official validation logic. :)

You already wrote the validation code therefore I think small modification to bind the validation logic with proper way will be necessary.

@lizable I mentioned you to explain the proper validation logic too.

lizable:
> @mihilt mwc-textfield has custom validation methods such as `checkValidity()`, `reportValidity()` and `setCustomValidity()` Please refer the link https://www.npmjs.com/package/@material/mwc-textfield and revise the implementation to use official validation logic. :)
> 
> You already wrote the validation code therefore I think small modification to bind the validation logic with proper way will be necessary.
> 
> jihyunkang (@lizable) I mentioned you to explain the proper validation logic too.

I agree. We already have custom validation logic of `mwc-textfield`, so you may refer to the code below and apply it to this PR.
https://github.com/lablup/backend.ai-webui/blob/64048cefb6aa66a6b8c68ade49f5ba29d9fef54a/src/components/backend-ai-signup.ts#L334-L364

mihilt:
@inureyes @lizable
Thank you for kind explanation. I think I changed it as you wanted.
I deleted validation functions that existed before and used mwc-textfield's method and properties.

https://user-images.githubusercontent.com/50368758/185409481-c4a6557a-7612-46f9-ad83-7576ab52a7f8.mp4

src/components/backend-ai-import-view.ts
```diff
@@ -581,6 +581,18 @@ export default class BackendAIImport extends BackendAIPage {
...
   }
 
+  urlTextfieldChanged(e, buttonIdValue) {
+    const button = this.shadowRoot.querySelector(`#${buttonIdValue}`);
+    if (e.target.value !== '' && e.currentTarget.checkValidity()) {
+      button.removeAttribute('disabled');
+    } else {
+      this.notification.text = _text('import.WrongURLType');
+      button.setAttribute('disabled', 'true');
+      this.importMessage = this.notification.text;
```
lizable: I guess we should split the importMessage to import notebook and import GitHub repo, since when either the value of the input field of Import Notebook or Import From GitHub is invalid, the error message is shown in both areas.
Here are the screenshots that refer to the issue that I mentioned above.

### Screnshots
GitHub_Repo mov
ImportNotebookRepo

lizable (changes requested):
I confirmed that validation logic is okay, but there's one minor thing to be better fixed.
Please refer to the comment below. except that, all is good!

mihilt:
It is just my curious, but I wanted to know about never used lit's @property status value.

I found that some values 'condition', 'resourcePolicy', '_helpDescriptionTitle', '_helpDescriptionIcon', 'storageProxyInfo' are never used in backend-ai-import-view.ts file.

I think it is hard to notice because ESLint can't check it.
Is it just useless value? Or not used in the component, but necessary?
If it is a useless value, can I delete it all?

lizable:
> It is just my curious, but I wanted to know about never used lit's @Property status value.
> 
> I found that some values 'condition', 'resourcePolicy', '_helpDescriptionTitle', '_helpDescriptionIcon', 'storageProxyInfo' are never used in backend-ai-import-view.ts file.
> 
> I think it is hard to notice because ESLint can't check it. Is it just useless value? Or not used in the component, but necessary? If it is a useless value, can I delete it all?

I confirmed that those properties described above seem to be legacy, but some of them should be left as is because they were used for the backbone component. Therefore let's remove only two properties, `resourcePolicy` and `storageProxyInfo` for now.

lizable (approved):
LGTM.
Well done, @mihilt! 👍

Btw, I just renamed the PR title to be a more precise description of what we want to achieve through this issue and PR.
FYI, we already have [the guide(Writing change logs)](https://github.com/lablup/backend.ai/tree/main/changes) for writing a good commit message, descriptions, etc, written by @achimnol.
Please refer to the guide, I'm sure that It will help you and other contributors a lot!

Issue #1416, opened 2022-08-16
Don't close the modified Registries information popup after modify registry information


studioego:
**Describe the bug**
문제가 발생하는 상황에 대한 요약 설명을 적어주세요.

Don't close the modified Registries information popup after modifying registry information on the popup. 

**To Reproduce**
문제를 재현하기 위한 순서를 자세히 적어주세요:
1. Go to 'Environments' menu (needs Administration level)
2. Click on 'Registry' tab
3. Click modify Registry button
Show Modify Registries popup, Try to modify registries information & save it
4. See error (Save it. But It doesn't close.)

**Expected behavior**
원래 기대했던 동작은 어떤 것인지 명확하게 적어주세요.

Save it and then close the modified Registries information popup. 

**Screenshots**
Just take a screenshot in your clipboard or copy a picture file and paste it here (Ctrl+V/Cmd+V) so that GitHub automatically attach and insert it.
가능하다면, 문제 상황을 보여주는 스크린샷을 첨부해주세요.
클립보드에 스크린샷을 찍거나 이미지 파일을 복사해서 여기에 붙여넣으면(Ctrl+V/Cmd+V) GitHub이 자동으로 이곳에 첨부 및 삽입해줍니다.
Screenshot
<img width="1260" alt="image" src="https://user-images.githubusercontent.com/20002/184802498-fc7ae22e-c12e-495b-8e1e-390f0451753b.png">

Screencast

https://user-images.githubusercontent.com/20002/184802916-fa5ba6ad-7f95-405b-b4b7-96df4b396d6a.mov

**Client:**
 - OS: MacOS 12.5
 - App version: 22.03.5.220804
 - Browser version: Chrome 104.0.5112.79 (arm64)
 - Python version: 3.10.5
 - NodeJS version: v18.7.0

**Server:**
 - OS:  MacOS 12.5
   - Output of `uname -a`: Darwin seongdacBookPro.office.lablup 21.6.0 Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:22 PDT 2022; root:xnu-8020.140.41~1/RELEASE_ARM64_T6000 arm64
 - Backend.AI version: main 
 - Python version: 3.10.5
 - Installation method:  used scripts/install-dev.sh

**Additional context**
추가적으로 문제 상황을 설명할 수 있는 다른 맥락을 기술해주세요.

inureyes:
Also, data has gone after modifying the information.

[...]

PR #1417, opened 2022-08-16
fix:  close the modify registry information popup after modify registry information

dan-2ee:
## Fix
This is an issue that has already been fixed in the PR(#1385). 
In the process of refactoring the code, it seems that issues related to modifying registry information have appeared. 
So I just put back the code before the issue and added some refactoring.

## Change

https://user-images.githubusercontent.com/65989401/184825236-134988be-e925-440a-9297-1424e792ccc4.mov

resolve: #1416


dan-2ee:
resolve #1418

dan-2ee:
resolve #1418

PR #1413, opened 2022-08-15
feature: support SSO login (SAML 2.0)

adrysn:
- Expose SSO Login buttons by referencing `general.singleSignOnVendors` parameter.

### Related PRs:
- https://github.com/lablup/backend.ai/pull/652

PR #1419, opened 2022-08-16
fix: SAML login via form submit instead of `fetch`

adrysn:
`fetch` API has some troubles in dealing with the redirect response. So, I gave up the API and falls back to the traditional, stable `form` submit.

This is a follow-up of https://github.com/lablup/backend.ai-webui/pull/1413.


src/components/backend-ai-login.ts
```diff
@@ -804,6 +818,25 @@ export default class BackendAILogin extends BackendAIPage {
     });
   }
 
+  async _token_login(sToken) {
+    // If token is delivered as a querystring, just save it as cookie.
+    document.cookie = `sToken=${sToken}; expires=Session; path=/`;
+    try {
+      console.log('logging with (cookie) token...');
```
inureyes: Remove this line?

inureyes (approved):
Fixed several conflicts.

PR #1427, opened 2022-08-21
feature: add extra information in the last page of session launcher

### LINKED ISSUE #1367: Let's show "extra information" for image in the last page of session launcher.

For now, the last page of session launcher displays image information, which contains environment name, versions, and architecture if is explicitly set.
But we could display more detailed information such as `projects`; where certain registries have for distinguishing projects by each purpose of the image(environment).

Display the project name at the session info section in the last page of session launcher.

mihilt:
resolve #1367

image

I added project name.
I made it simply to match styles around it.
Please let me know if I need to change in style or to add additional information.

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -310,7 +310,8 @@
       "EnterValidSessionName": "Enter valid session name",
       "SessionNameAlreadyExist": "Session name already exists"
     },
-    "EnterValidSessionName": "EnterValidSessionName"
+    "EnterValidSessionName": "EnterValidSessionName",
+    "EnvironmentInfo": "Environment Info"
   },
   "button": {
     "Cancel": "Cancel",
--- a/resources/i18n/id.json
+++ b/resources/i18n/id.json
@@ -313,7 +313,8 @@
     "CommitFinished": "__NOT_TRANSLATED__",
     "CommitFailed": "__NOT_TRANSLATED__",
     "DescCommitSession": "__NOT_TRANSLATED__",
-    "Message": "Pesan"
+    "Message": "Pesan",
+    "EnvironmentInfo": "__NOT_TRANSLATED__"
   },
   "button": {
     "Cancel": "Batalkan",
--- a/resources/i18n/ko.json
+++ b/resources/i18n/ko.json
@@ -312,7 +312,8 @@
     "CommitOnGoing": "세션 커밋 작업 진행중",
     "CommitFinished": "세션 커밋 작업 완료",
     "CommitFailed": "세션 커밋 작업 실패",
-    "DescCommitSession": "다음 세션에 대응하는 컨테이너를 커밋하려고 합니다: "
+    "DescCommitSession": "다음 세션에 대응하는 컨테이너를 커밋하려고 합니다: ",
+    "EnvironmentInfo": "환경 정보"
   },
   "button": {
     "Cancel": "취소",
--- a/resources/i18n/mn.json
+++ b/resources/i18n/mn.json
@@ -313,7 +313,8 @@
     "CommitFinished": "__NOT_TRANSLATED__",
     "CommitFailed": "__NOT_TRANSLATED__",
     "DescCommitSession": "__NOT_TRANSLATED__",
-    "Message": "зурвас"
+    "Message": "зурвас",
+    "EnvironmentInfo": "__NOT_TRANSLATED__"
   },
   "button": {
     "Cancel": "Цуцлах",
--- a/resources/i18n/ru.json
+++ b/resources/i18n/ru.json
@@ -313,7 +313,8 @@
     "CommitFinished": "__NOT_TRANSLATED__",
     "CommitFailed": "__NOT_TRANSLATED__",
     "DescCommitSession": "__NOT_TRANSLATED__",
-    "Message": "сообщение"
+    "Message": "сообщение",
+    "EnvironmentInfo": "__NOT_TRANSLATED__"
   },
   "button": {
     "Cancel": "Отмена",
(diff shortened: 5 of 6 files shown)


lizable:
Hmm, Isn't it too big for locating the project name at the top of the panel?
I think what users really want is "the environment information", "resources to allocate" mostly, and virtual folders to mount, and then "the project name".
I would rather put them below the mounted folder section, what do you think @mihilt ?

lizable:
And I just updated the description of the issue, so It would be appreciated if you refer to it and apply your ideas to it.

mihilt:
I agree with you.
I moved `Project Name`'s location under `Mounted folders`.
And I checked suggestion and also added `Resource Group`.

results are as follows:
<kbd>image</kbd>

inureyes:
> Hmm, Isn't it too big for locating the project name at the top of the panel?
> I think what users really want is "the environment information", "resources to allocate" mostly, and virtual folders to mount, and then "the project name".
> I would rather put them below the mounted folder section, what do you think @mihilt ?

Also, it will be better to locate 프로젝트 and its name.

e.g.

프로젝트           default
----------------------

mihilt:
> > Hmm, Isn't it too big for locating the project name at the top of the panel?
> > I think what users really want is "the environment information", "resources to allocate" mostly, and virtual folders to mount, and then "the project name".
> > I would rather put them below the mounted folder section, what do you think @mihilt ?
> 
> Also, it will be better to locate 프로젝트 and its name.
> 
> e.g.
> 
> 프로젝트 default

Does this mean it need to express name and value in one line?

lizable:
> > > Hmm, Isn't it too big for locating the project name at the top of the panel?
> > > I think what users really want is "the environment information", "resources to allocate" mostly, and virtual folders to mount, and then "the project name".
> > > I would rather put them below the mounted folder section, what do you think @mihilt ?
> > 
> > 
> > Also, it will be better to locate 프로젝트 and its name.
> > e.g.
> > 프로젝트 default
> 
> Does this mean it need to express name and value in one line?

Yes. Also, there's one more thing to suggest. 
I think we can put the project name and the name of resource group into the session info section rather than splitting each of them into a separate section.

mihilt:
Making design without exact decided design is hard for me.
First I simply put strings in it. And It looked so insincere so I tried to copy the surrounding design. 

Can I get some advice on the design you want?

<kbd>image</kbd>

<kbd>image</kbd>

lizable:
> Making design without exact decided design is hard for me.
> 
> First I simply put strings in it. And It looked so insincere so I tried to copy the surrounding design. 
> 
> 
> 
> Can I get some advice on the design you want?
> 
> 
> 
> <kbd>image</kbd>
> 
> 
> 
> <kbd>image</kbd>
> 
> 

I think it seems okay, but I guess it's better to ask @yomybaby . What do you think of this layout above?

yomybaby:
Already @inureyes mentioned, "Project : default" instead of "Project Name : default" is LGTM.

About the design, Plain text looks better than tag box(badge). Because the name is not a spec(category) of the session such as python. It's a name which is uniq, right? The tag box design is good for categorizing rather than showing name.

lizable:
> Already @inureyes mentioned, "Project : default" instead of "Project Name : default" is LGTM.
> 
> About the design, Plain text looks better than tag box(badge). Because the name is not a spec(category) of the session such as python. It's a name which is uniq, right? The tag box design is good for categorizing rather than showing name.

How about Just adding every title inside the section like this?
I just found an appropriate layout idea from dribbble:
image

You see, adding the title on the left-side of the content doesn't break the whole layout.
In our case, you may put title/content like the screenshot above, like `Assignees` / `Date` / `Tags`
And each title and content will go like this...
- Environment Info: Python 3.9 / Ubuntu 20.04 / AARCH64 / Interactive
- Project Name: default
- Resource Group Name: default

mihilt:
I tried this, but I couldn't complete adding `Environment Info: Python 3.9 / Ubuntu 20.04 / AARCH64 / Interactive` in one line because there was not enough space. In order to proceed in this way, it seems that needs to be further separated title.

And it is also hard for me to choose to remove language icon image. Can I do that?

lizable:
> I tried this, but I couldn't complete adding `Environment Info: Python 3.9 / Ubuntu 20.04 / AARCH64 / Interactive` in one line because there was not enough space. In order to proceed in this way, it seems that needs to be further separated title.
> 
> And it is also hard for me to choose to remove language icon image. Can I do that?

Any examples? perhaps screenshots?

mihilt:
<kbd>image</kbd>

Like this way, even if I didn't put `Environment Info:` text yet, but it already don't have any space.
So, `Ubuntu 20.04` lablup-shields component turned two lines.

And the language icon image that I wrote before is the python image on the far left.

---

<kbd>image</kbd>



Anyway, I made it similar to image you put in. 
But I couldn't show all badge near `Environment Info:` in one line.
And, I deleted the language icon image.

Does this look all right?

inureyes:
> image
> 
> Like this way, even if I didn't put `Environment Info:` text yet, but it already don't have any space. So, `Ubuntu 20.04` lablup-shields component turned two lines.
> 
> And the language icon image that I wrote before is the python image on the far left.
> 
> image
> 
> Anyway, I made it similar to image you put in. But I couldn't show all badge near `Environment Info:` in one line. And, I deleted the language icon image.
> 
> Does this look all right?

It looks good! Just an additional comment:
* Let's remove colon between text and tags
* Align text and tag area. (how about environment info, Project name, resource group text have same width?)

mihilt:
image

I did it. But it doesn't seem not much pretty compared to the surroundings.
I'm sorry I think I don't have an aesthetic sense. If there is anything strange, tell me.

inureyes:
> image
> 
> I did it. But it doesn't seem not much pretty compared to the surroundings. I'm sorry I think I don't have an aesthetic sense. If there is anything strange, tell me.

LGTM

inureyes (approved):
LGTM

PR #1435, opened 2022-08-22
feature: display richer agent information

### LINKED ISSUE #1433: Let's display richer information of agent in the list
Nevertheless, we already provide resource information of each agent in the agent list, we shall show more detailed information, such as the actual amount of currently allocated CPU cores compared to total allocatable CPU cores, etc.

Also, there are many keys that show what actual agent resources are not only for used by Backend.AI session(s) or other management services related to Backend.AI Cluster, but also other services that have been executed inside the agent node. It's important when a session is not allocatable for some reason (such as CPU cores are fully used by other services except Backend.AI), this would be helpful since at least it gives a plausible reason by utilization field.

[DELETE THIS LINE] A clear and concise description of what you want to happen.
[DELETE THIS LINE] 어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.
- [x] Rename Resources column to Allocation column
- [x] Refactor Allocation column(formerly known as Resources) to show like this
       `current-usage-of-resource / total-allocatable-resources`
- [x] Add Utilization column right next to Allocation column so that it reminds the utilization of resources in each agent node.
- [x] (Optional) Show progress bar with percentage in utilization column if possible

If it's hard to add utilization column, let's just apply the first two tasks described above will be enough.

**Additional context**
None

lizable:
## Description
This PR resolves #1433.

## Screenshots
- Before
Screen Shot 2022-08-23 at 12 38 05 AM

- After
Screen Shot 2022-08-23 at 12 37 34 AM


resources/i18n/en.json
```diff
@@ -357,6 +359,7 @@
     "DetailedInformation": "Detailed Information",
     "AgentSetting": "Agent Settings",
     "AgentSettingUpdated": "Agent Setting Updated.",
+    "NoAvailableLiveStat": "",
```
agatha197: Can you fill this message?

src/components/backend-ai-agent-list.ts
```diff
@@ -672,46 +705,44 @@ export default class BackendAIAgentList extends BackendAIPage {
...
+              <lablup-progress-bar id="mem-usage-bar" progress="${rowData.item.mem_total_usage_ratio}"
+                                   description="${rowData.item.total_mem_percent}%"></lablup-progress-bar>
             </div>` : html``}
           ${rowData.item.cuda_gpu_slots ? html`
-            <div class="layout horizontal center flex">
+            <div class="layout horizontal center-justified flex progress-bar-section">
               <div class="layout horizontal start resource-indicator">
                 <img class="indicator-icon fg green" src="/resources/icons/file_type_cuda.svg"/>
-                <span style="padding-left:5px;">${rowData.item.cuda_gpu_slots}</span>
+                <span class="monospace" style="padding-left:5px;">${rowData.item.used_cuda_gpu_slots}/${rowData.item.cuda_gpu_slots}</span>
```
agatha197, line 733: I think GPU utilization is different from the control panel.
In control panel, cuda_util ratio is calculated like [this](https://github.com/lablup/backend.ai-control-panel/blob/main/frontend/src/pages/hub-resource-agents.js#L848-L854).

src/components/backend-ai-agent-list.ts
```diff
@@ -1044,14 +1115,55 @@ export default class BackendAIAgentList extends BackendAIPage {
...
+                              .renderer="${this._boundRegionRenderer}">
+          </vaadin-grid-column>
+          <vaadin-grid-sort-column auto-width flex-grow="0" resizable path="architecture" header="${_t('agent.Architecture')}">
+          </vaadin-grid-sort-column>
+          <vaadin-grid-column resizable auto-width flex-grow="0" header="${_t('agent.Starts')}" .renderer="${this._boundContactDateRenderer}">
+          </vaadin-grid-column>
+          <vaadin-grid-column resizable width="160px" header="${_t('agent.Allocation')}"
+                              .renderer="${this._boundResourceRenderer}">
+          </vaadin-grid-column>
+          <vaadin-grid-column resizable width="120px" header="${_t('agent.Utilization')}"
```
agatha197: In the 14-inch monitor, it looks like this. So it's better to widen the utilization column. 
image

Sujin-Kim1 (changes requested):
I've attached some reviews, so please check them out :)

src/components/backend-ai-agent-list.ts
lizable: Okay, let's make it 10px wider than now.

resources/i18n/en.json
lizable: Got it!

src/components/backend-ai-agent-list.ts
lizable: @Sujin-Kim1  It's not as "identically" same as agent list in control-panel.
In WebUI for now in Allocation column, the admin user will see two information with progress bar:
- current allocated resources
- total amount of resources

in Utilization column, It will show up the ratio of actual resources usage in agent (it will contain other usages such as background processes in the agent(host))

And also, I don't know why there's hard-coded number `100 `in control-panel you mentioned.
Could you guess and tell me the reason?

agatha197: According to this [commit](https://github.com/lablup/backend.ai-control-panel/commit/89614f822c3a8ccbc18e9830955d5500a22bf569), capacity is delivered from the server as 0. So I guess @adrysn handled it as 100.

lizable: Since getting an empty object as `live_stat` from Redis server has been solved, I think we shall leave this commit as is.

Sujin-Kim1 (approved):
LGTM!

PR #1437, opened 2022-08-25
fix: CUDA utilization was displayed as `Infinity`

adrysn:
Server reports the `cuda_util.capacity` to 0. In this case, the GPU
utilization per node is displayed as `Infinity`.


lizable (approved):
LGTM.

PR #1438, opened 2022-08-25
fix: disable container commit when session is batch type

lizable:
## Description
This PR fixes two things, disable container commit icon when the session is `BATCH` type and adjust the width of Controls column depending on whether the version and configuration support container commit.

#### Before

before-container-commit-icon-arrangement-fixed

#### After

after-container-commit-icon-arrangement-fixed


inureyes (approved):
LGTM

PR #1439, opened 2022-08-25
feat: add agent summary page

Sujin-Kim1:
Add the `Agent Summary` page for the general users. This page can be shown only when the `hide-agents` option that existed in the config file is false.

refs: https://github.com/lablup/backend.ai/issues/640
related PR: https://github.com/lablup/backend.ai/pull/645

[Screenshot]
image


config.toml.sample
```diff
@@ -20,7 +20,7 @@ debug = false                           # Please refer to `[Debug mode](https://
 maskUserInfo = false                    # If true, then account information in header, session list will be masked
 #singleSignOnVendors = "saml"           # Comma-separated list of vendors that support single sign-on.
 enableContainerCommit = false           # If true, then user can request container commit corresponding to the session in running status
-
+hide-agents = true                      # If false, show the `Agent Summary` menu in the sidebar.
```
lizable: I think it doesn't match with the configuration setting page in `backend.ai-webui` component, which checks whether `hideAgents` exists.

lizable: Perhaps there's key mismatching in `backend-ai-webui` component [here](https://github.com/lablup/backend.ai-webui/pull/1439/files#diff-c970ee927d9c8e6050388af9c74159ccfd53cdff999e3e89f836009010c01aabR334-R336).

resources/i18n/ko.json
```diff
@@ -7,6 +7,7 @@
       "Sessions": "세션",
       "Import&Run": "가져오기 & 실행",
       "Data&Storage": "데이터 & 폴더",
+      "AgentSummary": "자원 요약",
       "Pipeline": "파이프라인 (베타)",
```
lizable, line 11: I'm not sure to paraphrase "Agent Summary" to "자원 요약". Would It be okay @inureyes ?

lizable: If it's acceptable, then just ignore my comment, please.

src/backend-ai-app.ts
```diff
@@ -62,7 +62,10 @@ const loadPage = (page, params: Record<string, unknown> = {}) => (dispatch) => {
     break;
   case 'pipeline-job':
     import('./pipeline/components/pipeline-job-view.js');
-    break; */
+    break;
```
lizable, line 65: Merging main branch with fixing conflicts to this branch may have commented out `agent-summary` page by accident.
adding `*/` at the end of L68 will fix this.

src/components/backend-ai-agent-summary-list.ts
```diff
@@ -0,0 +1,501 @@
...
+@group Backend.AI Web UI
+@element backend-ai-agent-list
+*/
+
+@customElement('backend-ai-agent-summary-list')
+export default class BackendAIAgent extends BackendAIPage {
+  @property({type: String}) condition = 'running';
+  @property({type: Boolean}) useHardwareMetadata = false;
+  @property({type: Array}) agents = [];
+  @property({type: Object}) list_status = Object();
```
lizable: Let's apply `@query` decorator instead of `@property` when we use member variable as referencing the custom component declared in HTML template.

src/components/backend-ai-agent-summary-list.ts
```diff
@@ -0,0 +1,501 @@
...
+      break;
+    case 'archived':
+    default:
+      status = 'ALIVE';
+    }
+
+    if (globalThis.backendaiclient.supports('schedulable')) {
+      fields.push('schedulable');
+    }
+
```
lizable: Not a big deal, but argument with hard-coded number could be altered with local variables with corresponding value.
For example, `10 * 1000` means timeout.

src/components/backend-ai-agent-summary-list.ts
```diff
@@ -0,0 +1,501 @@
...
+  @property({type: Object}) agentDetail = Object();
+  @property({type: Object}) notification = Object();
+  @property({type: Object}) agentDetailDialog = Object();
+  @property({type: Object}) agentSettingDialog = Object();
+  @property({type: Boolean}) enableAgentSchedulable = false;
+  @property({type: Object}) _boundEndpointRenderer = this.endpointRenderer.bind(this);
+  @property({type: Object}) _boundResourceRenderer = this.resourceRenderer.bind(this);
+  @property({type: Object}) _boundSchedulableRenderer = this.schedulableRenderer.bind(this);
+  @property({type: String}) filter = '';
+  @property({type: String}) list_condition = 'loading';
```
lizable: Is there a certain reason to use snake case in "list condition", but use camelcase in [`backend-ai-agent-list`](https://github.com/lablup/backend.ai-webui/blob/a21eb36ad500bc0a10a914ce991d87173ecc682e/src/components/backend-ai-agent-list.ts#L87)?

src/components/backend-ai-agent-summary-list.ts
```diff
@@ -0,0 +1,501 @@
...
+  @property({type: String}) condition = 'running';
+  @property({type: Boolean}) useHardwareMetadata = false;
+  @property({type: Array}) agents = [];
+  @property({type: Object}) list_status = Object();
+  @property({type: Object}) agentsObject = Object();
+  @property({type: Object}) agentDetail = Object();
+  @property({type: Object}) notification = Object();
+  @property({type: Object}) agentDetailDialog = Object();
+  @property({type: Object}) agentSettingDialog = Object();
+  @property({type: Boolean}) enableAgentSchedulable = false;
```
lizable: And we don't need to expose `enableAgentSchedulable` as a property. let's just remove the property decorator and add access modifier.

src/components/backend-ai-agent-summary-list.ts
```diff
@@ -0,0 +1,501 @@
...
+  */
+  _loadAgent(status = 'running') {
+    if (this.active !== true) {
+      return;
+    }
+    this.list_condition = 'loading';
+    this.list_status.show();
+    const fields = ['id', 'status', 'available_slots', 'occupied_slots', 'architecture'];
+    switch (this.condition) {
+    case 'running':
```
lizable: I think there are only two kinds of status settings in here,
If the condition is `terminated` then status will be set `TERMINATED`, in the other cases, it goes to `ALIVE`.
In that case, we shall change the code in a neater way like this:
```
switch (this.condition) {
  case 'running':
  case 'archived':
  default:
    status = 'ALIVE';
  case 'terminated':
    status =  'ALIVE';
}

Comment (lizable) [reply src/components/backend-ai-agent-summary-list.ts]: And also I wonder why we treat "archived" condition as `ALIVE`.
Do you know any clue for that @Sujin-Kim1 ?

File: src/lib/backend.ai-client-esm.ts [hunk 9]
```diff
@@ -2128,6 +2130,51 @@ class Agent {
...
+  /**
+   * List of agent summary.
+   *
+   * @param {string} status - Status to query. Should be one of 'ALIVE', 'PREPARING', 'TERMINATING' and 'TERMINATED'.
+   * @param {array} fields - Fields to query. Queryable fields are:  id, status, scaling_group, schedulable, schedulable, available_slots, occupied_slots.
+   * @param {number} limit - limit number of query items.
+   * @param {number} offset - offset for item query. Useful for pagination.
+   * @param {number} timeout - timeout for the request. Default uses SDK default. (5 sec.) 
+   */
+  async list(status = 'ALIVE', fields = ["id", "status", "scaling_group", "schedulable", "available_slots", "occupied_slots", "architecture"],
```
lizable, line 2154: It would be great to treat status as a type since there are only 4 cases in status "ALIVE", "PREPARING", "TERMINATING", "TERMINATED".
It is a similar case that I mentioned in #1441 code review, so maybe we can add a single issue about that (since it's just about declaring type, right?).

lizable: Here's my comment related to the mention above.
https://github.com/lablup/backend.ai-webui/pull/1441#discussion_r957065637

lizable (changes requested):
I have several requests on components mostly about enhancing code quality.
Please refer to the comments below.

config.toml.sample
lizable: Oops! duplicated comments! just refer to the first one, please. 😅

src/components/backend-ai-agent-summary-list.ts
agatha197: Because it was written before merging https://github.com/lablup/backend.ai-webui/pull/1441 😅 I'll reflect it!

config.toml.sample
agatha197: ah I didn't reflect on this file. Thanks! (I didn't recognize it because I changed `config.toml` 😅)

src/components/backend-ai-agent-summary-list.ts
agatha197: Hmm.. I think it is a legacy. And there seems to be no place to use the "archived" condition. Without it, it wouldn't be a problem in my think.

ref:
https://github.com/lablup/backend.ai-webui/blob/main/src/components/backend-ai-agent-view.ts#L136-L139

agatha197: Then, the syntax of the switch can be changed as follows.
`status = this.condition === 'running' ? 'ALIVE' : 'TERMINATED';`

lizable: I think referencing from model schema of `Agents` would be the best way.
https://github.com/lablup/backend.ai/blob/14373e7b9ffd70bcc2e8ccbeb6855636a6bb2bca/src/ai/backend/manager/models/agent.py#L53

According to the AgentStatus schema definition in the code above, It has four statuses. 
`ALIVE`, `LOST`, `TERMINATED`, `RESTARTING`.
I wonder whether we should cover `LOST` or `RESTARTING` here.

And there's one more thing that seems a bit awkward in this function `_loadAgentList` (perhaps because of the legacy... I guess). It receives status as an argument, but It manipulates by the `condition`.
Therefore I don't know the exact purpose of property how condition and status... maybe we should treat it as one unless there's certain reason to add "condition". 
I know in `backend-ai-agent-view`, there are two tabs for `backend-ai-agent-list` split by conditions, but we could give a better approach without using two purpose-overlapping property in vague.

src/backend-ai-app.ts
agatha197: I resolved it. thank you :)

src/lib/backend.ai-client-esm.ts
agatha197: Yes, and there are several situations like this throughout the component, so it would be good to create a separate issue that does type alias.

agatha197: I'll handle it from https://github.com/lablup/backend.ai-webui/issues/1448!

src/components/backend-ai-agent-summary-list.ts
```diff
@@ -0,0 +1,494 @@
...
+  @property({type: Object}) agentsObject = Object();
+  @property({type: Object}) agentDetail = Object();
+  @property({type: Object}) notification = Object();
+  @property({type: Object}) agentDetailDialog = Object();
+  @property({type: Object}) agentSettingDialog = Object();
+  @property({type: Object}) _boundEndpointRenderer = this.endpointRenderer.bind(this);
+  @property({type: Object}) _boundResourceRenderer = this.resourceRenderer.bind(this);
+  @property({type: Object}) _boundSchedulableRenderer = this.schedulableRenderer.bind(this);
+  @property({type: String}) filter = '';
+  @property({type: String}) listCondition = 'loading';
```
agatha197: I'll use type alias after merging https://github.com/lablup/backend.ai-webui/pull/1447

src/components/backend-ai-agent-summary-list.ts
agatha197: I just remove the status parameter in the `_loadAgentList` in both agent and agent-summary.
And how about handling the `LOST` and `RESTARTING` agents list when it is needed?

adrysn (approved):
LGTM.

src/components/backend-ai-agent-summary-list.ts
agatha197: I applied it.

PR #1441, opened 2022-08-26
bugfix: list status unrecognized

Sujin-Kim1:
- fix: blocking cannot read properties of null (list_status)
- refactor: use camelcase to properties that are related to `backend-ai-list-status`

[Minor changes]
- fix: lint errors (`backend-ai-environment-list` indentations)
- fix: typo error for `cudaUtilCapacity`

### DIFF
--- a/src/components/backend-ai-error-log-list.ts
+++ b/src/components/backend-ai-error-log-list.ts
@@ -48,8 +48,8 @@ export default class BackendAiErrorLogList extends BackendAIPage {
   @property({type: String}) message = '';
   @property({type: Array}) logs = [];
   @property({type: Array}) _selected_items = [];
-  @property({type: Object}) list_status = Object();
-  @property({type: String}) list_condition = 'loading';
+  @property({type: Object}) listStatus = Object();
+  @property({type: String}) listCondition = 'loading';
   @property({type: Object}) _grid = Object();
   @property({type: Array}) logView = [];
   @property({type: Number}) _pageSize = 25;
@@ -124,7 +124,7 @@ export default class BackendAiErrorLogList extends BackendAIPage {
   }
 
   firstUpdated() {
-    this.list_status = this.shadowRoot.querySelector('#list-status');
+    this.listStatus = this.shadowRoot.querySelector('#list-status');
     this._updatePageItemSize();
     this._grid = this.shadowRoot.querySelector('#list-grid');
     if (!globalThis.backendaiclient || !globalThis.backendaiclient.is_admin) {
@@ -147,17 +147,17 @@ export default class BackendAiErrorLogList extends BackendAIPage {
    * Refresh log data.
    */
   _refreshLogData() {
-    this.list_condition = 'loading';
-    this.list_status.show();
+    this.listCondition = 'loading';
+    this.listStatus?.show();
     this._updatePageItemSize();
     this.logs = JSON.parse(localStorage.getItem('backendaiwebui.logs') || '{}');
     this._totalLogCount = this.logs.length > 0 ? this.logs.length : 1;
     this._updateItemsFromPage(1);
     this._grid.clearCache();
     if (this.logs.length == 0) {
-      this.list_condition = 'no-data';
+      this.listCondition = 'no-data';
     } else {
-      this.list_status.hide();
+      this.listStatus?.hide();
     }
   }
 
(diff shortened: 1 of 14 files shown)


src/components/backend-ai-resource-policy-list.ts
```diff
@@ -42,8 +42,8 @@ class BigNumber {
 @customElement('backend-ai-resource-policy-list')
 export default class BackendAIResourcePolicyList extends BackendAIPage {
   @property({type: Boolean}) visible = false;
-  @property({type: Object}) list_status = Object();
-  @property({type: String}) list_condition = 'loading';
+  @property({type: Object}) listStatus = Object();
```
lizable: How about using `@query` instead of a property declaration?

src/components/backend-ai-resource-preset-list.ts
```diff
@@ -36,8 +36,8 @@ class BackendAiResourcePresetList extends BackendAIPage {
   @property({type: String}) condition = '';
   @property({type: String}) presetName = '';
   @property({type: Object}) resourcePresets;
-  @property({type: Object}) list_status = Object();
-  @property({type: String}) list_condition = 'loading';
+  @property({type: Object}) listStatus = Object();
```
lizable: Same here.

src/components/backend-ai-scaling-group-list.ts
```diff
@@ -54,8 +54,8 @@ import {IronFlex, IronFlexAlignment} from '../plastics/layout/iron-flex-layout-c
 export default class BackendAIScalingGroupList extends BackendAIPage {
   @property({type: Object}) _boundControlRenderer = this._controlRenderer.bind(this);
   @property({type: Number}) selectedIndex = 0;
-  @property({type: Object}) list_status = Object();
-  @property({type: String}) list_condition = 'loading';
+  @property({type: Object}) listStatus = Object();
```
lizable: Same here.

src/components/backend-ai-session-list.ts
```diff
@@ -119,8 +119,8 @@ export default class BackendAiSessionList extends BackendAIPage {
   @property({type: Object}) terminateSelectedSessionsDialog = Object();
   @property({type: Object}) sessionStatusInfoDialog = Object();
   @property({type: Boolean}) enableScalingGroup = false;
-  @property({type: Object}) list_status = Object();
-  @property({type: String}) list_condition = 'loading';
+  @property({type: Object}) listStatus = Object();
```
lizable: Here's also same suggestion that I mentioned above.

src/components/backend-ai-storage-list.ts
```diff
@@ -93,8 +93,8 @@ export default class BackendAiStorageList extends BackendAIPage {
   @property({type: Object}) deleteFileDialog = Object();
   @property({type: Object}) downloadFileDialog = Object();
   @property({type: Object}) sessionLauncher = Object();
-  @property({type: Object}) list_status = Object();
-  @property({type: String}) list_condition = 'loading';
+  @property({type: Object}) listStatus = Object();
```
lizable: Same here as well.

src/components/backend-ai-storage-proxy-list.ts
```diff
@@ -42,8 +42,8 @@ import './lablup-progress-bar';
 export default class BackendAIStorageProxyList extends BackendAIPage {
   @property({type: String}) condition = 'running';
   @property({type: Array}) storages;
```
lizable, line 45: Here, too.

src/components/backend-ai-user-list.ts
```diff
@@ -65,7 +65,7 @@ export default class BackendAIUserList extends BackendAIPage {
   @property({type: Array}) userInfoGroups = [];
   @property({type: String}) condition = '';
   @property({type: Object}) _boundControlRenderer = this.controlRenderer.bind(this);
```
lizable, line 68: Here as well.

lizable (changes requested):
There are multiple same requests about replacing the `@property` decorator with `@query`.
Except that everything seems fine.

src/components/backend-ai-error-log-list.ts
```diff
@@ -147,17 +146,17 @@ export default class BackendAiErrorLogList extends BackendAIPage {
...
+    this.listCondition = 'loading';
+    this._listStatus?.show();
     this._updatePageItemSize();
     this.logs = JSON.parse(localStorage.getItem('backendaiwebui.logs') || '{}');
     this._totalLogCount = this.logs.length > 0 ? this.logs.length : 1;
     this._updateItemsFromPage(1);
     this._grid.clearCache();
     if (this.logs.length == 0) {
-      this.list_condition = 'no-data';
+      this.listCondition = 'no-data';
```
lizable, line 157: Just a quick suggestion, how about adding a type on `listCondition` on every component that uses that property, if there are only two types used in there? If does, It will check whether the assigning value is valid or not in compile time(inside IDE).
Here's a good example that uses type definition on string-based values in `backend-ai-session-list` component.
https://github.com/lablup/backend.ai-webui/blob/63a1b1d0891933e2cddcc2aa38972c0185a12d47/src/components/backend-ai-session-list.ts#L73-L85

agatha197: Good suggestion! I'll apply right away thanks :)

lizable (commented):
If there are only two strings (`loading`, `no-data`)used in `listCondition` property, maybe adding a type on this property would be helpful.

src/components/backend-ai-error-log-list.ts
lizable: Let's apply this to a separate PR.

lizable (approved):
LGTM.

src/components/backend-ai-error-log-list.ts
agatha197: I'll handle it another PR 😉

PR #1444, opened 2022-08-29
hotfix: remove wrong parsing number logic in initial configuration

lizable:
## Description
This PR fixes human error based on calling the number parsing function(e.g. `ParseInt`, `ParseFloat`, etc.) on wrong data.

Also, there was an issue with sanitizing the whitespace when reading from configuration values with commas(`,`), which are regarded as an array. when there are no values (undefined or null or empty string), It was treated as an empty string(`''`). But when appending `split(',')` function at the end, it changed to `['']`, which is different from an empty array(`[]`). Therefore I added extra conditional logic to set an empty array when the value is nullish.

### DIFF
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -465,6 +465,11 @@ export default class BackendAILogin extends BackendAIPage {
     }
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


Sujin-Kim1 (approved):
LGTM!

adrysn (approved):
LGTM.

PR #1445, opened 2022-08-30
update: lit and mwc-components to the latest version

inureyes:
update: lit and mwc-components to the latest version


inureyes (commented):
Works flawlessly.

Issue #1468, opened 2022-09-07
Let's add helper dialog and validation for endpoint


Sujin-Kim1:
### Main idea

When trying to log in, many people have a hard time figuring out what to set the endpoint to.
So let's add the following to make endpoint configuration easier. 
- [ ] an icon button to the endpoint field
- [ ] add a dialog explaining which value is appropriate to set as the endpoint.
- [ ] add a validation (e.g. URL Starts with http:// or https://")

### Anything else?

ref: https://docs.backend.ai/en/latest/install/development-setup.html, https://github.com/lablup/backend.ai/blob/main/scripts/install-dev.sh#L904-L926

jangjichang:
i will check this issue

jangjichang:
@Sujin-Kim1 https://github.com/lablup/backend.ai-webui/pull/1470

di-uni:
I will check this issue about icon and dialog!

PR #1470, opened 2022-09-11
fix: add endpoint validation

jangjichang:
related issue: https://github.com/lablup/backend.ai-webui/issues/1468

before
User has entered a invalid endpoint format
image

after
image

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -1066,6 +1066,7 @@
     "CancelLogin": "Cancel Login",
     "SignoutFinished": "Signout finished.",
     "APIEndpointEmpty": "API Endpoint is empty. Please specify API endpoint to login.",
+    "APIEndpointStartWith": "API Endpoint must start with http:// or https://.",
     "PleaseWait": "Please wait to login.",
     "ConnectingToCluster": "Connecting to Backend.AI Cluster...",
     "ForgotPassword": "Forgot password?",
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -1214,6 +1214,9 @@ export default class BackendAILogin extends BackendAIPage {
       this._enableUserInput();
       if (showError) {
         this.notification.text = PainKiller.relieve('Endpoint is unreachable. Please check the connection or endpoint');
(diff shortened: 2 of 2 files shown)


src/components/backend-ai-login.ts
```diff
@@ -1214,6 +1214,9 @@ export default class BackendAILogin extends BackendAIPage {
       this._enableUserInput();
       if (showError) {
         this.notification.text = PainKiller.relieve('Endpoint is unreachable. Please check the connection or endpoint');
+        if (!this.api_endpoint.toLowerCase().startsWith('http://') && !this.api_endpoint.toLowerCase().startsWith('https://')) {
```
agatha197, line 1217: How about using `validityTransform`?
ref: https://github.com/lablup/backend.ai-webui/blob/main/src/components/backend-ai-signup.ts#L340

resources/i18n/en.json
```diff
@@ -1066,6 +1066,7 @@
     "CancelLogin": "Cancel Login",
     "SignoutFinished": "Signout finished.",
     "APIEndpointEmpty": "API Endpoint is empty. Please specify API endpoint to login.",
+    "APIEndpointStartWith": "API Endpoint must start with http:// or https://.",
```
agatha197, line 1069: Please use i18n.
1. Run `make i18n`
2. Translate other words if you can.

More details: https://github.com/lablup/backend.ai-webui#extracting-i18n-resources

Sujin-Kim1 (changes requested):

src/components/backend-ai-login.ts
inureyes: I agree with Sujin :D

inureyes (changes requested):
Please review the comments and let's upgrade input form to use validityTransform :)

src/components/backend-ai-login.ts
inureyes: @Sujin-Kim1 Let's finish this PR. It is from contributor's repo, therefore there is no way to update the repo. It will be better to branch and make a new PR with reference to this PR.

agatha197: okay, I'll do that.

Sujin-Kim1:
this PR will be handled by https://github.com/lablup/backend.ai-webui/pull/1689.
close this issue.

PR #1476, opened 2022-09-14
style: change `modify-env-dialog` style

### LINKED ISSUE #1471: Redesign environment setting dialog in the session launcher.
### Main idea

The environment setting dialog currently in the session launcher is challenging to add new items if the user inputs many setting values. There is a possibility that saves and delete all are close to each other, so there is a possibility of incorrectly pressing. So let's change the style of the dialog to solve this problem.

### Alternative ideas

_No response_

### Anything else?

Here is the reference image.
Users can make a new row by clicking the bottom add button.
A row will disappear if click the minus icon.
And the entire row except the first row will disappear if click the `x` icon. But if the user inputs something, the confirm dialog appears.
image

Sujin-Kim1:
resolves #1471 
- Click the **`Reset` button** with input any value -> show confirm dialog -> delete all of the values and rows. but the first row will not be removed.
- Click the **close button** with input any value -> show confirm dialog -> delete all values and rows except for the pre-saved environ values. (the first row will not be removed too.)
- Click the reset button without input value -> delete all values and rows except the first row.

[Screenshots]
- Default 
image

- When clicking the `Reset` button or close button with input.
image

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -344,7 +344,8 @@
     "Continue": "Continue",
     "GoBackToSummaryPage": "Go back to Summary page",
     "ForceTerminate": "Force Terminate",
-    "Commit": "Commit"
+    "Commit": "Commit",
+    "Reset": "Reset"
   },
   "agent": {
     "Endpoint": "Endpoint",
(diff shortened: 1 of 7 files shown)


inureyes (approved):
LGTM

PR #1477, opened 2022-09-14
style: adjust the License Key button size

ky3vin:
### Changes
Adjust the License Key button size to be the same as the other buttons.

### Before
image
image

### After
image
image


Sujin-Kim1 (approved):
LGTM

PR #1481, opened 2022-09-20
feature: allow plugins work as hidden page plugin.

inureyes:
* This PR allows plugins to act as API endpoint
 * When parsing the sidebar items, menu handling routine ignores `page` type plugins if its' menu item name is empty.


Sujin-Kim1 (commented):
Hiding the plugin where the title is blank in the menu works well. However, it is possible to access the page by modifying url, so is it better to prevent it?

inureyes:
> Hiding the plugin where the title is blank in the menu works well. However, it is possible to access the page by modifying url, so is it better to prevent it?

It is intended so we do not consider the situation now. :D

Issue #1380, opened 2022-07-19
Use common functions and variables of `backend-ai-page`


lizable:
**Is your feature request related to a problem? Please describe.**
기능 요청이 특정 문제에 연관된 것이라면 여기에 그 문제를 설명해주세요.
There are many duplicated functions that are used in many components, which work exactly the same or slightly differently per each component. It's better to rearrange certain functions into superclass([`backend-ai-page`](https://github.com/lablup/backend.ai-webui/tree/main/src/components/backend-ai-page.ts)).

**Describe the solution you'd like**
어떤 기능이 있으면 좋겠는지 자세히 설명해주세요.
Here's a guide for deduping scattered functions:
1. Find duplicated functions and variables used in multiple(more than one) components.
2. Rearrange the functions and variables found in step 1 to superclass component.
3. If there are any component-specific statements in the functions or variables, then use `override` keyword and override them to the component.

**Describe alternatives you've considered**
혹시 다른 대안들을 생각해본 적이 있다면 함께 적어주세요.

**Additional context**
기능 요청에 대해 보다 잘 이해할 수 있는 다른 맥락을 기술해주세요.

yomybaby:
Closing this issue as it is no longer valid.

PR #1482, opened 2022-09-20
Show warning dialog and validation message when setting rate limit below 100

### LINKED ISSUE #1469: Show warning dialog when setting rate limit below 100
### Main idea

The rate limit function limits the number of API calls per 900 seconds to prevent unexpected DDoS attacks and API abuse. However, if this value is too small, API operations such as login using API calls will also be blocked. 

The rate limit value is also used to block spammers, so the setting can be as small as 1 as before, but when setting a value less than 100, let's improve the user interface to provide a warning and a dialog asking to continue.

### Alternative ideas

_No response_

### Anything else?

_No response_

Sujin-Kim1:
this resolves #1469 

[Screenshots]
- Input below 100
image

- Click the `SAVE CHANGES` button when the rate limit is below 100.
image


src/components/backend-ai-credential-list.ts
```diff
@@ -696,25 +697,91 @@ export default class BackendAICredentialList extends BackendAIPage {
...
+
+  closeDialog(id) {
+    (this.shadowRoot?.querySelector('#' + id) as BackendAIDialog).hide();
+  }
+
   /**
    * Save a keypair modification.
    *
-   * @param {Event} e - Dispatches from the native input event each time the input changes.
+   * @param {boolean} confirm - Save keypair info even if rateLimit is less the warningRateLimit if `confirm` is true.
```
agatha197, line 765: When setting `closeWithConfirm` to `keypair-modify-dialog`, the `keypair-modify-dialog` cannot be closed immediately by pressing the close button. So I added confirm param to this function.

lizable: I'll check it.

agatha197: [backend-ai-credential-list.ts.zip](https://github.com/lablup/backend.ai-webui/files/9650443/backend-ai-credential-list.ts.zip)
this is the version of adding confirm dialog

src/components/backend-ai-credential-list.ts
```diff
@@ -696,25 +697,91 @@ export default class BackendAICredentialList extends BackendAIPage {
       `, root);
   }
 
+  _validateRateLimit() {
```
lizable, line 704: I think we should use "[_addInputValidator](https://github.com/lablup/backend.ai-webui/blob/3ee743375ebe4fc26baa0a26d71a67394efbae7f/src/components/backend-ai-page.ts#L77-L126)" instead of adding custom validator in every components. Anyway let's handle them as a separate issue.

src/components/backend-ai-credential-list.ts
```diff
@@ -889,7 +960,8 @@ export default class BackendAICredentialList extends BackendAIPage {
                 label="${_t('credential.RateLimit')}"
                 validationMessage="${_t('credential.RateLimitValidation')}"
                 helper="${_t('credential.RateLimitValidation')}"
-                @change=${() => this._adjustRateLimit()}
+                @change="${() => this._validateRateLimit()}"
```
lizable, line 976: Any reasons to add two event handlers (change, input) in here?

lizable (changes requested):
I found one minor curiosity.

src/components/backend-ai-credential-list.ts
agatha197: Now that I see it, I'll just need one.

src/components/backend-ai-credential-list.ts
agatha197: I agree. It's better to be handled by https://github.com/lablup/backend.ai-webui/issues/1380.

PR #1495, opened 2022-10-22
Open dialog and add credential with URL option

snaag:
- Open dialog, add credential when path from action=add
- #1394

### DIFF
--- a/src/components/backend-ai-credential-view.ts
+++ b/src/components/backend-ai-credential-view.ts
@@ -327,7 +327,7 @@ export default class BackendAICredentialView extends BackendAIPage {
   /**
    * If admin comes, prepare the user list page.
    */
-  _preparePage() {
+  async _preparePage() {
     if (globalThis.backendaiclient.is_admin !== true) {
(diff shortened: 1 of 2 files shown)


src/components/backend-ai-credential-view.ts
```diff
@@ -995,6 +996,16 @@ export default class BackendAICredentialView extends BackendAIPage {
     return Math.round(gigabyte * value);
   }
 
+  async _runAction() {
+    const action = location.search.split('action=')[1];
```
inureyes: @snaag This line will generate error when ?action GET parameter is missing.

image

inureyes (changes requested):
@snaag `_runAction` generates an error when `action` GET parameter is missing.

src/components/backend-ai-credential-view.ts
snaag: Oh right. Modified by adding code for checking.

inureyes (approved):
LGTM

PR #1500, opened 2022-10-27
ci: Fix sync-labels option error

Yaminyam:
The sync-labels option of the labeler, which syncs the current label according to this change, is not working properly.

It was an issue raised 2 months ago, but it is an issue that is not being dealt with and left unattended.
https://github.com/actions/labeler/issues/404

A branch with a bug-fixed version exists and changes to that version
https://github.com/actions/labeler/pull/405


inureyes (approved):
LGTM

PR #1503, opened 2022-10-28
feat: add download link button for webui desktop

### LINKED ISSUE #1486: Let's embed direct download link for WebUI desktop in WebUI
### Main idea

Backend.AI GUI is not only limited to the web browser but also has a Desktop application implemented in electron. Sometimes users want to download Desktop App, and every time they ask about it we need to copy and paste the link, which is cumbersome.
Therefore, it's better to add a download button in somewhere in WebUI, so that we can guide it to the users easily. Also, we need to provide the version to download to be the same version of WebUI’s.

### Alternative ideas

_No response_

### Anything else?

_No response_

Sujin-Kim1:
- this resolves #1486 
- Add `appDownloadUrl` option to config.
- Add a button that links to `<appDownloadUrl>/v<webui version>/backend.ai-desktop-<webui-version>-<OS>-<architecture>.<extension>`

image
image
image
image

related PR: https://github.com/lablup/backend.ai/pull/861

### DIFF
--- a/src/components/backend-ai-webui.ts
+++ b/src/components/backend-ai-webui.ts
@@ -915,11 +915,6 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
     }
   }
 
-  showDownloadPage() {
-    const version = globalThis.packageVersion;
-    window.open(`https://github.com/lablup/backend.ai-webui/releases/tag/v${version}`, '_blank');
-  }
-
   /**
    * Add tool tips by create popovers.
    */
@@ -933,7 +928,6 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
     // this._createPopover('#pipeline-menu-icon', _text('webui.menu.Pipeline'));
     // this._createPopover('#pipeline-job-menu-icon', _text('webui.menu.PipelineJob'));
     this._createPopover('#statistics-menu-icon', _text('webui.menu.Statistics'));
-    this._createPopover('.download-link-menu-icon', _text('webui.menu.Download'));
     this._createPopover('#usersettings-menu-icon', _text('webui.menu.Settings'));
     this._createPopover('backend-ai-help-button', _text('webui.menu.Help'));
     if (this.is_admin) {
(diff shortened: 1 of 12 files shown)


src/components/backend-ai-webui.ts
```diff
@@ -1152,6 +1158,7 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
               </address>
               <div id="sidebar-navbar-footer" class="vertical start end-justified layout" style="margin-left:16px;">
                 <backend-ai-help-button active style="margin-left:4px;"></backend-ai-help-button>
+                <mwc-icon-button class="download-link-menu-icon fg white" icon="download" slot="graphic" style="margin-left:4px;" @click="${() => this.showDownloadPage()}"></mwc-icon-button>
```
inureyes: It should be only shown on Web mode, not on App mode.

src/components/backend-ai-webui.ts
```diff
@@ -1176,6 +1183,7 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
             </address>
             <div id="sidebar-navbar-footer" class="vertical start end-justified layout" style="margin-left:16px;">
               <backend-ai-help-button active style="margin-left:4px;"></backend-ai-help-button>
+              <mwc-icon-button class="download-link-menu-icon fg white" icon="download" slot="graphic" style="margin-left:4px;" @click="${() => this.showDownloadPage()}"></mwc-icon-button>
```
inureyes: Here, too.

inureyes (changes requested):
The download button should be only shown on Web mode, not on App mode.

And, how about moving it to another location? e.g. Panel on summary page?

src/components/backend-ai-webui.ts
agatha197: reflected!

src/components/backend-ai-webui.ts
agatha197: reflected.

inureyes:
@Sujin-Kim1 How about 'Download App' instead of 'Download' ?

inureyes (approved):
LGTM

PR #1505, opened 2022-10-31
feat: add log download button on session log dialog

### LINKED ISSUE #1504: Add log download button on session log dialog
### Main idea

When log is too long, it is better to download and read the log. It is also helps users debug current code.

Let's add log download button on session log dialog.

### Alternative ideas

_No response_

### Anything else?

_No response_

Sujin-Kim1:
resolves #1504 

image

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -313,7 +313,8 @@
       "SessionNameAlreadyExist": "Session name already exists"
     },
     "EnterValidSessionName": "EnterValidSessionName",
-    "EnvironmentInfo": "Environment Info"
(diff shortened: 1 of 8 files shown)


inureyes:
@Sujin-Kim1 Code seems ok. However, we need tests with Chrome / Safari / Firefox. (not sure about Blob type instantiation)

Sujin-Kim1:
> kimsujin (@Sujin-Kim1) Code seems ok. However, we need tests with Chrome / Safari / Firefox. (not sure about Blob type instantiation)

It worked when I tested it.

adrysn (approved):
LGTM.

PR #1516, opened 2022-11-14
fix: send preferred port when only checked

Sujin-Kim1:
It solves the problem of sending port information together without checking the preferred port when launching the session app in the WebUI.


adrysn (approved):
LGTM.

PR #1518, opened 2022-11-21
fix: TOS dialog doesn't show because of id mismatching

Sujin-Kim1:
In the signup dialog, all of the Terms Of Service dialog doesn't show because of the id mismatching.


lizable (approved):
LGTM.

PR #1530, opened 2022-12-13
feat: show all resource templates when `alwaysEnqueueComputeSession` is true

### LINKED ISSUE #1523: show the preset list when `always_enqueue_compute_session` is on.
### Main idea

If `always_enqueue_compute_session` is true, the user can create a new session even if the user doesn't have enough resources to create a session. But the preset list is blocked and this is wired. Let's show the preset list when `always_enqueue_compute_session` is on.

### Alternative ideas

_No response_

### Anything else?

_No response_

Sujin-Kim1:
resolve #1523 

- show all of the preset lists (resource templates) when `alwaysEnqueueComputeSession` is true.
- set allocatable resource limit according to `total_resource_group_slot` when `alwaysEnqueueComputeSession` is true to prevent automatically canceling the creation of a new session because of the lack of resources.
   - reference image
    image

### DIFF
--- a/src/components/backend-ai-resource-broker.ts
+++ b/src/components/backend-ai-resource-broker.ts
@@ -405,7 +405,7 @@ export default class BackendAiResourceBroker extends BackendAIPage {
     // console.log('aggregate from:', from);
     this.aggregate_updating = true;
     const total_slot = {};
-    const total_resource_group_slot = {};
+    const total_resource_group_slot: any = {};
     const total_project_slot = {};
 
     return globalThis.backendaiclient.keypair.info(globalThis.backendaiclient._config.accessKey, ['concurrency_used']).then(async (response) => {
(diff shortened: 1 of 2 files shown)


src/components/backend-ai-resource-broker.ts
```diff
@@ -671,6 +646,54 @@ export default class BackendAiResourceBroker extends BackendAIPage {
...
+          item.mem = globalThis.backendaiclient.utils.changeBinaryUnit(item.resource_slots.mem, 'g');
+          if (item.shared_memory) {
+            item.shmem = globalThis.backendaiclient.utils.changeBinaryUnit(item.shared_memory, 'g');
+          } else {
+            item.shmem = null;
+          }
+          if (item.allocatable) {
+            availablePresets.push(item);
+          }
+        });
```
yomybaby: You can divide this part into `map` and `filter`. This is more easy to read and easy to handle type.
```js
const availablePresets  = presets.map(item => {
  //modify item
  return item
}).filter(item => item.allocatable)

```

src/components/backend-ai-resource-broker.ts
```diff
@@ -671,6 +646,54 @@ export default class BackendAiResourceBroker extends BackendAIPage {
       this.available_slot = remaining_sg_slot;
       this.used_slot_percent = used_slot_percent;
       this.used_resource_group_slot_percent = used_resource_group_slot_percent;
+
+      let enqueueSession = false;
+      if (globalThis.backendaiclient._config.always_enqueue_compute_session === true) {
+        enqueueSession = true;
+      }
```
yomybaby:
```suggestion
      const enqueueSession = globalThis.backendaiclient._config.always_enqueue_compute_session === true
```

yomybaby (commented):
The logic is LGTM.
But I suggest refactoring some parts. 😀

Sujin-Kim1:
> The logic is LGTM. But I suggest refactoring some parts. 😀

Super! Thank you for the feedback. I reflected them all :)

src/components/backend-ai-resource-broker.ts
```diff
@@ -671,6 +646,43 @@ export default class BackendAiResourceBroker extends BackendAIPage {
...
+              item[slotName] = item.resource_slots[slotKey];
+            }
+          }
+        } else if (enqueueSession) {
+          // Even if allocatable is false, if enqueueSession is true,
+          // allocatable is determined based on when no resources are allocated.
+          item.allocatable = true;
+          for (const [slotKey, slotName] of Object.entries(slotList)) {
+            if (slotKey in item.resource_slots && slotName in total_resource_group_slot) {
+              if (item.resource_slots[slotKey] <= total_resource_group_slot[slotName]) {
```
adrysn: When the `slotKey` is `mem`, two comparison slots are in different units: one in bytes and the other in Gbytes. This makes some presets not displayed.

image

adrysn (changes requested):
Some presets are not displayed due to unit mismatch for memory resource slot.

e.g.) configured presets:
image

displayed presets when there are no resources:
image

src/components/backend-ai-resource-broker.ts
agatha197: Oops, thanks for catching it! I fixed.

adrysn (approved):
LGTM.

PR #1540, opened 2023-01-09
ci: automatically build and commit static files

kyujin-cho:


inureyes:
Since we have revamped build pipeline, this PR is resolved and not necessary.

PR #1545, opened 2023-01-12
feat/fix: initialize aliases after successfully creating a session and fix the error that mounts using another name

### LINKED ISSUE #1524: The alias for the mount folder doesn't work.

macOS (Apple Silicon)

### Backend.AI version

22.09.3

### Describe the bug

Even if the user sets the alias for the mount folder, the alias doesn't work. (original name used.)

### To Reproduce

1. create a session with mounting the aliased folder.
2. create a new session.
    2-1. mount the previously mounted folder exactly the same without the browser refreshing. 
    2-2. use the alias that already exists in the dialog. / remove using the backspace keyboard and type the same name.
3. see the error.

### Expected Behavior

It should be mounted as an alias applied folder.

### Anything else?

There would be two options to fix this issue. 
1. initialize all aliases to blank whenever opening the session dialog.
2. keep the existing alias, but the folder should be mounted using the alias without any problem.

Sujin-Kim1:
resolves #1524

### DIFF
--- a/src/components/backend-ai-session-launcher.ts
+++ b/src/components/backend-ai-session-launcher.ts
@@ -1053,6 +1053,14 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
     }
   }
 
+  _initializeFolderMapping() {
+    this.folderMapping = {};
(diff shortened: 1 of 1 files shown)


inureyes (approved):
LGTM

PR #1547, opened 2023-01-17
fix: call `toString` method in setting quota value

adrysn:
The PR resolves https://github.com/lablup/giftbox/issues/25.


yomybaby (approved):
LGTM, thank you for your effort.

PR #1549, opened 2023-01-18
fix: display full image tags in session launcher and environment list to better distinguish images with long similar tags

Sujin-Kim1:
resolves https://github.com/lablup/giftbox/issues/39


inureyes (approved):
LGTM

adrysn (commented):
I tested with the following two images and confirmed that all the tags are displayed in image constraint column.
* cr.backend.ai/stable/filebrowser:21.02-ubuntu20.04
* cr.backend.ai/stable/filebrowser:21.02-ubuntu20.04-jpark-test-v1
image

I think this logic can also be applied to the session launcher dialog as well?
image

adrysn (approved):
LGTM.

Issue #1541, opened 2023-01-10
fGPU progress bar is always zero in resource indicator


adrysn:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

22.9.8

### Describe the bug

I launched a session with 8 fGPUs, but the fGPU progress bar in the resource indicator seems to be always zero. The number `8` is displayed correctly. You can check this bug by looking at the upper-right corner of the image below:

<img width="836" alt="image" src="https://user-images.githubusercontent.com/7539358/211570305-c23ee673-0a86-40f1-afc9-25f930a497af.png">

### To Reproduce

Create any session with non-zer fGPU.

### Expected Behavior

The progress bar should have a non-zero value like other resource bars (CPU, RAM).

PR #1553, opened 2023-01-25
bugfix: gpu, fgpu progress bar are always zero

Sujin-Kim1:
resolves https://github.com/lablup/backend.ai-webui/issues/1541 

reason of this bug: the progress value is `NaN`


inureyes (approved):
LGTM

PR #1564, opened 2023-01-29
Fix: Support to import custom github & gitlab repos

studioego:
* Support to import custom github & gitlab repos (Note: url includes 'github' or 'gitlab')
* remove '.git' suffix about both github & gitlab repos Because Checking Both github & gitlab repo use https APIs


studioego:
In my opinion, The feature don't work Some github repo (ex. custom public github and custom private github). So I replace new PR https://github.com/lablup/backend.ai-webui/pull/1565

PR #1565, opened 2023-01-30
Fix: remove '.git' suffix about git repos

studioego:
remove '.git' suffix about both github & gitlab repos Because Checking Both github & gitlab repo use https APIs

I checked the git repo's URL issue. However, If importing feature get `***.git` URL, The feature removes `***.git`'s postfix `.git`. So I'll improve the importing repo both github and gitlab repo.


adrysn (approved):
LGTM.

PR #1566, opened 2023-01-30
fix: refactor vfolder quota setting dialog

### LINKED ISSUE #1559: Refactor vfolder quota update dialog

Linux (x86-64)

### Backend.AI version

23.03.0dev0

### Describe the bug

There are two problems reported in quota setting dialog.
- No exception handling when the quota value is empty. (default value is 1 MiB)
- NaN Error occurs when changing unit on value change because of restricted type conversion (BigInt is not allowed)
- Width of the dropdown inside the quota setting dialog breaks the layout

Screenshot(s)
Screenshot 2023-01-27 at 6 33 07 PM
image

### To Reproduce

_No response_

### Expected Behavior

- Even if when responses from the server didn't have quota information, It should expose the minimum size of quota, which is 1 MiB.
- During value changes, any unit conversion should be applied without error.

### Anything else?

_No response_

lizable:
This PR resolves #1559.


adrysn (approved):
LGTM.

PR #1567, opened 2023-01-31
feat: add disk column on agent tab and display the total capacity of the storage

Sujin-Kim1:
resolves https://github.com/lablup/giftbox/issues/6

agent tab (add disk column)
image
storage tab (display the capacity value)
image


adrysn (approved):
LGTM.

PR #1578, opened 2023-02-10
feat: Add modal to enter SSH keypair manually

yomybaby:
ref https://github.com/lablup/giftbox/issues/58


src/components/backend-ai-usersettings-general-list.ts
```diff
@@ -1138,6 +1188,13 @@ export default class BackendAiUsersettingsGeneralList extends BackendAIPage {
               unelevated
               label="${_t('button.Generate')}"
               @click="${this._refreshSSHKeypair}"></mwc-button>
+          <mwc-button
+              unelevated
+              label="${_t('button.EnterManually')}"
+              @click="${() => {
+                this._initManualSSHKeypairFormDialog();
```
lizable, line 1195: @yomybaby I just add a function for resetting the input field when opening the dialog since It should be deleted after registration.

src/components/backend-ai-usersettings-general-list.ts
```diff
@@ -843,6 +847,52 @@ export default class BackendAiUsersettingsGeneralList extends BackendAIPage {
...
+      pubkey: sshPublicKey,
+      privkey: sshPrivateKey
+    });
+    p.then((resp) => {
+      this.notification.text = _text('usersettings.SSHKeypairEnterManuallyFinished');
+      this.notification.show();
+      this._hideSSHKeypairFormDialog();
+      this._openSSHKeypairRefreshDialog();
+    }).catch((err) => {
+      if (err && err.message) {
```
lizable, line 888: Also, I think It's better to show the result of the ssh keypair registration operation request, regardless of the result (success or failure). So I just added a notification at the end of the request.

lizable (approved):
It works fine, but I slightly did a touch-up for a better user experience.
I think it's okay to merge.

src/components/backend-ai-usersettings-general-list.ts
yomybaby: 👏

src/components/backend-ai-usersettings-general-list.ts
yomybaby: 👏

PR #1585, opened 2023-02-14
fix: align gpu process bar

leejiwon1125:
image
- left: before   /   right: after

Until now, function **_numberWithPostfix()** returned empty string '' because the first parameter was NaN which is the value of **this.used_resource_group_slot_percent.cuda_device** and **this.used_slot_percent.cuda_device**.

The value NaN is set in those values because there is no key 'cuda_device' in **~_percent component**. See the picture below.

image

So I made the value which can replace percent by dividing using value with total value.

---
My implementation could be incorrect or correct only in 'one manager and one agent' condition.
The reason for this is...
1. I can't use **real** gpu because I used cuda-mock.
image
2. I can't make situation like below. So. I can't test sophisticated 
image

If you have solution on this two problem, please let me know.


yomybaby:
Thank you for finding that "_numberWithPostfix" returns an empty string `''` when it is NaN.

I looked into the relative code. There are two problems.
#### 1. When `slot` is not in `used _slot`, there is nothing happens. 
  https://github.com/lablup/backend.ai-webui/blob/313f7ac513ae2f8a95231627be922ac965a9b208/src/components/backend-ai-resource-broker.ts#L634-L635

  It should be set percentages like this.
  ```javascript
} else {
    used_slot_percent[slot] = 0;
    used_resource_group_slot_percent[slot] = 0;
    used_project_slot_percent[slot] = 0;
}
  ```

#### 2. Wrong keys are used.
https://github.com/lablup/backend.ai-webui/blob/313f7ac513ae2f8a95231627be922ac965a9b208/src/components/backend-ai-resource-monitor.ts#L747-L748
https://github.com/lablup/backend.ai-webui/blob/313f7ac513ae2f8a95231627be922ac965a9b208/src/components/backend-ai-resource-monitor.ts#L769-L770

use `*****slot_percent["cuda.device"]` instead of `*****slot_percent.cuda_device`

After fixing this, we don't need to re-calculate the percentage.

yomybaby (changes requested):
Thank you for finding that "_numberWithPostfix" returns an empty string `''` when it is NaN.

I looked into the relative code. There are two problems.
#### 1. When `slot` is not in `used _slot`, there is nothing happens. 
  https://github.com/lablup/backend.ai-webui/blob/313f7ac513ae2f8a95231627be922ac965a9b208/src/components/backend-ai-resource-broker.ts#L634-L635

  It should be set percentages like this.
  ```javascript
} else {
    used_slot_percent[slot] = 0;
    used_resource_group_slot_percent[slot] = 0;
    used_project_slot_percent[slot] = 0;
}
  ```

#### 2. Wrong keys are used.
https://github.com/lablup/backend.ai-webui/blob/313f7ac513ae2f8a95231627be922ac965a9b208/src/components/backend-ai-resource-monitor.ts#L747-L748
https://github.com/lablup/backend.ai-webui/blob/313f7ac513ae2f8a95231627be922ac965a9b208/src/components/backend-ai-resource-monitor.ts#L769-L770

use `*****slot_percent["cuda.device"]` instead of `*****slot_percent.cuda_device`

After fixing this, we don't need to re-calculate the percentage.

lizable (approved):
I think we need to do more touch-ups to make it a perfect alignment, but let's wrap it up here.
It seems acceptable.
### Screenshot(s)
Screenshot 2023-02-22 at 1 35 30 PM

PR #1593, opened 2023-02-16
feature: show total resource group amount in resource monitor gauge

### LINKED ISSUE #1592: Display total amount of resources in the current resource group in the upper part of the resource indicator in Web UI

Linux (x86-64)

### Backend.AI version

22.09.10

### Description
For now, the gauge in resource monitor reflected the usage of other users which confuses the user.
As you can see below, the value inside the gauge and the percentage right next to the gauge doesn't match.

### Screenshot(s)
MicrosoftTeams-image (51)

### To Reproduce

_No response_

### Expected Behavior

Set the value in the upper gauge of each resource to reflect the total amount of resources in resource group, regardless of usage by other users.

### Anything else?

_No response_

lizable:
This PR resolves #1592.

1. Login to account A.
2. Create a session with the maximum resources that you can allocate on behalf of account A.
3. Login to account B.
4. See each gauge in resource monitor.

### Screenshot(s)
|      | Before | After |
| ---| --- | --- |
|resource monitor of account A | Screenshot 2023-02-17 at 10 59 25 AM | Screenshot 2023-02-17 at 10 55 15 AM |
|resource monitor of account B | Screenshot 2023-02-17 at 10 59 33 AM | Screenshot 2023-02-17 at 10 55 01 AM |

### DIFF
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -110,6 +110,7 @@ export default class BackendAILogin extends BackendAIPage {
   @property({type: Boolean}) maxShmPerContainer = 2;
   @property({type: Boolean}) maxFileUploadSize = -1;
   @property({type: Boolean}) maskUserInfo = false;
+  @property({type: Boolean}) hideAgents = false;
   @property({type: Array}) singleSignOnVendors: string[] = [];
   @property({type: Array}) allow_image_list;
   @property({type: Array}) endpoints;
@@ -703,6 +704,14 @@ export default class BackendAILogin extends BackendAIPage {
         defaultValue: 'https://github.com/lablup/backend.ai-webui/releases/download',
         value: (generalConfig?.appDownloadUrl),
       } as ConfigValueObject) as string;
+
+    // Enable hide agent flag
+    this.hideAgents = this._getConfigValueByExists(generalConfig,
+      {
+        valueType: 'boolean',
+        defaultValue: true,
+        value: (generalConfig?.hideAgents),
+      } as ConfigValueObject) as boolean;
 
     // Enable pipeline flag
     // FIXME: temporally disable pipeline feature in manual
@@ -1417,6 +1426,7 @@ export default class BackendAILogin extends BackendAIPage {
       globalThis.backendaiclient._config.singleSignOnVendors = this.singleSignOnVendors;
       globalThis.backendaiclient._config.enableContainerCommit = this._enableContainerCommit;
       globalThis.backendaiclient._config.appDownloadUrl = this.appDownloadUrl;
+      globalThis.backendaiclient._config.hideAgents = this.hideAgents;
       globalThis.backendaiclient.ready = true;
       if (this.endpoints.indexOf(globalThis.backendaiclient._config.endpoint as any) === -1) {
         this.endpoints.push(globalThis.backendaiclient._config.endpoint as any);
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -110,7 +110,7 @@ export default class BackendAILogin extends BackendAIPage {
   @property({type: Boolean}) maxShmPerContainer = 2;
   @property({type: Boolean}) maxFileUploadSize = -1;
   @property({type: Boolean}) maskUserInfo = false;
-  @property({type: Boolean}) hideAgents = false;
+  @property({type: Boolean}) hideAgents = true;
   @property({type: Array}) singleSignOnVendors: string[] = [];
   @property({type: Array}) allow_image_list;
   @property({type: Array}) endpoints;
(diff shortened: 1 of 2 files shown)


src/components/backend-ai-resource-broker.ts
```diff
@@ -601,7 +603,51 @@ export default class BackendAiResourceBroker extends BackendAIPage {
...
+            } else {
+              availableSlots['cuda_shares'] = 0;
+            }
+            return availableSlots;
+          })
+          .reduce((acc, curr) => {
+            Object.keys(curr).forEach((key) => {
+              acc[key] += curr[key];
+            });
+          return acc;
```
agatha197: how about giving default value just in case?

Sujin-Kim1 (changes requested):
When I tested, the progress bar doesn't work properly.
- account A
  image

- account B
  image

If I understand correctly, the top gauge of the CPU should be 2/4.

src/components/backend-ai-resource-broker.ts
lizable: > When I tested, the progress bar doesn't work properly. And please fix the lint errors 😉
> 
> * account A
>   image
> * account B
>   image
> 
> If I understand correctly, the top gauge of the CPU should be 2/4.

You mean the account B is wrong? I don't think so. Because account B didn't create any session at all so it should display 0 / 4 core and etc. Please check the test condition again.

agatha197: okay. I misunderstood.

src/components/backend-ai-resource-broker.ts
```diff
@@ -601,7 +603,51 @@ export default class BackendAiResourceBroker extends BackendAIPage {
...
+        });
+        this.total_resource_group_slot.mem = this.total_resource_group_slot.mem.toFixed(2);
```
agatha197:
```suggestion
        this.total_resource_group_slot.mem = this.total_resource_group_slot.mem?.toFixed(2);
```

src/components/backend-ai-login.ts
```diff
@@ -110,6 +110,7 @@ export default class BackendAILogin extends BackendAIPage {
   @property({type: Boolean}) maxShmPerContainer = 2;
   @property({type: Boolean}) maxFileUploadSize = -1;
   @property({type: Boolean}) maskUserInfo = false;
+  @property({type: Boolean}) hideAgents = false;
```
agatha197: In https://github.com/lablup/backend.ai-webui/pull/1593/files#diff-efcecd6c9b6f3d9e866667b37775627d072d9cde0d40326edcba878e773fc109R712, the default value is true but this is false. Wouldn't it be better to unify?

Sujin-Kim1 (changes requested):

src/components/backend-ai-resource-broker.ts
```diff
@@ -601,7 +603,57 @@ export default class BackendAiResourceBroker extends BackendAIPage {
       }
 
       this.total_slot = total_slot;
-      this.total_resource_group_slot = total_resource_group_slot;
+      if (!globalThis.backendaiclient._config.hideAgents) {
+        let initialResourceSlots = {
```
yomybaby: This is small feedback. You can use `const` here.  😬 
Other option sis  don't assign the initial value to variable(`initialResourceSlots`) and just put that on reducer (649 line)

src/components/backend-ai-login.ts
lizable: Yup, you're right I'll change it.

src/components/backend-ai-resource-broker.ts
lizable: Yes, It doesn't seem necessary to assign it as a local variable.

Sujin-Kim1 (approved):
LGTM!

src/components/backend-ai-resource-broker.ts
```diff
@@ -601,7 +603,51 @@ export default class BackendAiResourceBroker extends BackendAIPage {
       }
 
       this.total_slot = total_slot;
-      this.total_resource_group_slot = total_resource_group_slot;
+      if (!globalThis.backendaiclient._config.hideAgents) {
+        const status = 'ALIVE';
+        const limit = 20;
```
adrysn, line 608: What happens when there are more than 20 Agents?

lizable: I'm afraid that didn't test it yet.
For now, I just referenced the basic parameter value from [here](https://github.com/lablup/backend.ai-webui/blob/e4949338a3cf561b10cc606e963d770d41992c15/src/components/backend-ai-agent-summary-list.ts#L203-L207).
If the value needs adjustment following code should be changed accordingly.

adrysn: Temporary follow-up PR: https://github.com/lablup/backend.ai-webui/pull/1599.
Please review and merge if it looks okay. @lizable

lizable: Merged!

PR #1594, opened 2023-02-17
add: more CSS injection point to custom.css

inureyes:
This PR increases the design injection points through custom.css, making the design override process easier.


Sujin-Kim1 (approved):
LGTM

PR #1599, opened 2023-02-18
fix: increase item count 20 -> 100 for agent summary list query

adrysn:
If there are over 20 Agents in a reosurce group, #1593 will not work properly. this PR increases the fetch item count by 100.

:warning: Let's assume that the number of agents is less than 100 for user-accessible resource group. This will meet our immediate need, but we need to fix this when refactoring the resource indicator.


lizable (approved):
LGTM.

PR #1604, opened 2023-02-21
hotfix: typo errors in getting 2FA

Sujin-Kim1:


lizable (changes requested):
@Sujin-Kim1
2FA means 'two factor authentication'. I think the key for enable/disable 2FA should be labeled "enable-2FA" only or "enable-two-factor-authentication". It sounds a bit awkward to use the word "authentication" repetitively.

Sujin-Kim1:
> kimsujin (@Sujin-Kim1) 2FA means 'two factor authentication'. I think the key for enable/disable 2FA should be labeled "enable-2FA" only or "enable-two-factor-authentication". It sounds a bit awkward to use the word "authentication" repetitively.

Thanks! I'll change it to just '2FA' because it is used to check whether user is using totp plugin or not.

PR #1607, opened 2023-02-23
fix: display all available presets when `always_enqueue_compute_session` options is configured

Sujin-Kim1:
resolves https://github.com/lablup/giftbox/issues/187


adrysn (approved):
LGTM.

PR #1608, opened 2023-02-24
Bugfix: set max values of session slider based on `keypair_limits` and `hideAgents` setting when `alwaysEnqueueComputeSession` is true

agatha197:
fixes https://github.com/lablup/giftbox/issues/195

`this.total_slot` is calculated by considering both `keypair_resource_limit`(keypair_limits) and `total_resource_group_slot` as shown below. Use this value to calculate the slider maximum when enqueue session is possible.

🚨 FYI,`this.total_slot` value will vary if other users within the same Resource Group create sessions.

```typescript
if (keypair_resource_limit['cpu'] === 'Infinity') { // When resource is infinity, use scaling group limit instead.
  total_slot['cpu'] = total_resource_group_slot['cpu'];
} else {
  total_slot['cpu'] = keypair_resource_limit['cpu'];
}
```
+) fix: avoid getting property from undefined.


lizable (changes requested):
@agatha197   It's more complicated than we expected. I'll handle this issue from now on.

adrysn (changes requested):
For me, this is not working. I confirmed that the `hideAgent` and `alwaysEnqueueComputeSession` are set to `true`.
image

I have three resource presets:
image

I set the user's resource policy to 4 cores, 8 GB memory, and 0.5 fGPU.
image

So, at least the resource preset `GPU-Suppor` should appear in the list. But, nothing is displayed in the user's session launcher. Also, GPU slider is disabled.

image

When I change the resource policy of the user with unlimited values:
image

I see two resource presets (I don't know why `GPU-Full` is not displayed):
image

lizable:
> For me, this is not working. I confirmed that the `hideAgent` and `alwaysEnqueueComputeSession` are set to `true`. image
> 
> I have three resource presets: image
> 
> I set the user's resource policy to 4 cores, 8 GB memory, and 0.5 fGPU. image
> 
> So, at least the resource preset `GPU-Suppor` should appear in the list. But, nothing is displayed in the user's session launcher. Also, GPU slider is disabled.
> 
> image
> 
> When I change the resource policy of the user with unlimited values: image
> 
> I see two resource presets (I don't know why `GPU-Full` is not displayed): image

@adrysn Did you set the image doesn't require GPU resource such as Python 3.9? If you did, then GPU resource slider would be disabled. If not, I need to dig down little more.

adrysn:
I used `ngc-pytorch:22.12`. Every resource presets are visible at first, but after other users allocate resources, it dissapeared.

lizable:
## Updates
- When both `always_enqueue_session` and `hideAgents` key in the configuration are enabled, then users' max resource limit follows `max_***_per_container keys` in the configuration file.
-  When `always_enqueue_session` is enabled but `hideAgents` is set to `false`, then the users' max resource limit follows the total amount of resources that the current(selected) resource group contains.

@Sujin-Kim1 @adrysn Could you review this PR?

src/components/backend-ai-session-launcher.ts
```diff
@@ -1946,26 +1929,25 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
             mem_metric.max = 0;
           }
           const image_mem_max = globalThis.backendaiclient.utils.changeBinaryUnit(mem_metric.max, 'g', 'g');
```
agatha197: LGTM. But just remove this line, please. I got this warning message.
image

lizable: Confirmed. I just removed Line 1931 since there's a statement that uses `mem_metric.max`.

lizable (approved):

lizable:
@adrysn Could you review this issue this week?

inureyes:
Please resolve the conflict. @Sujin-Kim1

adrysn:
Is this PR stilll mergable? If the NEO launcher will be introduced soon, it might be better to close this and deal this issue in the new launcher.

yomybaby (commented):
This PR is no longer valid, so we will close it. The `alwaysEnqueueComputeSession` has been deprecated, and the changes in this PR have already been implemented in the NEO Session Launcher.

PR #1610, opened 2023-02-26
refactor: display total remaining resources in resource group indicator

adrysn:
This PR displays the value of the resource group indicator as the total **remaining** slots in the current resource group, only when `hideAgents` is set to `false`.

image


lizable (approved):
@adrysn 
IMO It would be great if most of the calculation for the resource group had been done on the server-side, but let's just wrap it up now considering the urgency.

adrysn:
I don't think so. This metric is a per-user or per-browser, so it is more efficient to calculate in user's browser. All the server have to send is enough raw metrics. If not, the server have to calculate every per-user metrics which increases server load with unnecessary ephemeral calculations.

PR #1615, opened 2023-03-02
fix: get `totpSupported` flag asynchronously

Sujin-Kim1:
The logic of setting `totpSupported` option is incorrect, and I found a problem that the totp-related columns are not visible. Resolve the above error by getting the `totpSupported` option asynchronously.


yomybaby (approved):
LGTM!

PR #1626, opened 2023-03-13
update: change '강제종료' to '종료' for multiple session termination button text

### LINKED ISSUE #1595: When session is selected, mark '강제종료' button as '종료'
### Main idea

image

image

When you press the 'Forced Termination' button, a window appears that allows you to perform both the actual termination button and the normal termination button, so it seems better to mark it as the 'End' button.

This is also the reason why it should be marked as 'terminate' because the English version is marked as 'terminate' instead of 'force terminate'.

### Alternative ideas

_No response_

### Anything else?

_No response_

inureyes:
This PR resolves #1595

### DIFF
--- a/resources/i18n/ko.json
+++ b/resources/i18n/ko.json
@@ -246,7 +246,7 @@
     "Finished": "종료",
     "Others": "기타",
     "exportCSV": "CSV로 내보내기",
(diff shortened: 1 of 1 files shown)


Sujin-Kim1 (approved):
LGTM

PR #1628, opened 2023-03-13
feat: add force 2fa option

Sujin-Kim1:
Display this dialog on every page if `force2FA` is true and 2FA is not activated. 
스크린샷 2023-03-13 오후 9 57 43

resolves [Add an option to enforce users 2FA#261](https://github.com/lablup/giftbox/issues/261)
related PR: https://github.com/lablup/backend.ai/pull/1161

### DIFF
--- a/config.toml.sample
+++ b/config.toml.sample
@@ -22,6 +22,7 @@ maskUserInfo = false                    # If true, then account information in h
 enableContainerCommit = false           # If true, then user can request container commit corres ponding to the session in running status
 hideAgents = true                       # If false, show the `Agent Summary` menu in the sidebar.
 enable2FA = false                       # If true and backend.ai-totp-plugin is installed, users can activate 2FA.
+force2FA = false                        # If true, user should be register the 2FA to use Backend.AI WebUI.
 appDownloadUrl = ""                     # URL to download the electron app. If blank, https://github.com/lablup/backend.ai-webui/releases/download will be used.
 
 [wsproxy]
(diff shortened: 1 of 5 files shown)


config.toml.sample
```diff
@@ -22,6 +22,7 @@ maskUserInfo = false                    # If true, then account information in h
 enableContainerCommit = false           # If true, then user can request container commit corres ponding to the session in running status
 hideAgents = true                       # If false, show the `Agent Summary` menu in the sidebar.
 enable2FA = false                       # If true and backend.ai-totp-plugin is installed, users can activate 2FA.
+force2FA = false                        # If true, user should be register the 2FA to use Backend.AI WebUI.
```
lizable, line 25: We need to add this to webserver configuration, too.
Every new key added to the configuration has to be reflected [here](https://github.com/lablup/backend.ai/blob/main/configs/webserver/sample.conf), accordingly.

lizable: @Sujin-Kim1  please create a related PR for this in [Backend.AI Mono repo](https://github.com/lablup/backend.ai).

lizable (commented):
Considering the situation that It could be implemented only on the "client-side", It works okay.
But please add the corresponding PR that I mentioned below.

config.toml.sample
agatha197: @lizable I already wrote it to the PR description. https://github.com/lablup/backend.ai/pull/1161

PR #1645, opened 2023-03-17
feat: display idle checker remaining time

Sujin-Kim1:
related PR: https://github.com/lablup/backend.ai/pull/1160
resolves https://github.com/lablup/giftbox/issues/260

If idle_checks is like the below,
```
{
  "timeout": 10.0,
  "session_lifetime": 20.0,
  "utilization": {"cpu": true, "mem": false}
}
```
result is   `(Timeout: 00:00:10)`

If idle_checks is like the below,
```
{
  "timeout": 70.0,
  "session_lifetime": 65.0,
  "utilization": {"cpu": true, "mem": false}
}
```
result is   `(Session Lifetime: 00:01:05)`

if the idle checks have a negative timeout, the result is `(Timeout Exceeded)`.

[Screenshots]
image


src/lib/backend.ai-client-esm.ts
```diff
@@ -583,6 +583,7 @@ class Client {
     if (this.isManagerVersionCompatibleWith('22.09')) {
       this._features['image-commit'] = true;
       this._features['fine-grained-storage-permissions'] = true;
+      this._features['idle-checks'] = true;
```
inureyes: Should be '22.09.19' instead of '22.09'

inureyes (changes requested):
Need to modify compatible check code.

inureyes (approved):
LGTM

PR #1640, opened 2023-03-16
feat: add a host usage indicator to "create vFolder dialog"

yomybaby:
- resolve https://github.com/lablup/backend.ai-webui/issues/1630
- related PR: https://github.com/lablup/backend.ai/pull/1170
- How to test: set `allowed_volume_info` to etcd.
```
# example
# the default value is the percentage.
./backend.ai mgr etcd put volumes/allowed_volume_info percentage,used-bytes,capacity-bytes
```
- TODO: add setting UI for setting allowed_volume_info

## After
Indicator : Green/ Yellow / Red  

스크린샷 2023-03-16 오후 5 58 38

스크린샷 2023-03-16 오후 5 58 24

스크린샷 2023-03-16 오후 5 57 58

### Status of selected host
스크린샷 2023-03-17 오후 1 41 54

### Info modal
image

---

### Get storage usage info according to the changed server structure.
it is under the usage object. ([ref](https://github.com/lablup/backend.ai/pull/1170/files#diff-ac5fe69367ba4b06e41987788344c84b1b28efc8fbc0d0b176b96124da60259aR699-R703))
```
f"{proxy_name}:{volume_data['name']}": {
            "backend": volume_data["backend"],
            "capabilities": volume_data["capabilities"],
            "usage": await fetch_allowed_volume_usage(
                storage_manager=root_ctx.storage_manager,
                proxy_name=proxy_name,
                volume_name=volume_data["name"],
            ),
        }
```
### Indicator: reduce font size
image

### Info modal: add left-padding
image

PR #1648, opened 2023-03-22
hotfix: rename `volumeInfo` property to `storageProxyInfo`

Sujin-Kim1:
follows https://github.com/lablup/backend.ai-webui/pull/1640
`storageProxyInfo` is not defined as a property. So below errors occurs.
And `volumeInfo` actually acts `storageProxyInfo`. So I rename `volumeInfo` property to `storageProxyInfo` and renamed the related function.
image
image


adrysn (approved):
LGTM.

PR #1636, opened 2023-03-15
feat: make use of SI/binary prefixes consistent for disk, network, and memory

lizable:
This PR resolves #1631.

PR #1650, opened 2023-03-24
fix: use SI unit (e.g., GB not GiB) for quota-related fields

adrysn:
follow-up of https://github.com/lablup/backend.ai-webui/pull/1636


Sujin-Kim1 (changes requested):
How about adding `MiB: Math.pow(2, 20)` to `this.quotaUnit` and using it?

Sujin-Kim1 (approved):
LGTM

PR #1651, opened 2023-03-24
add: add utilization idle checks column

Sujin-Kim1:
- related PR: https://github.com/lablup/backend.ai/pull/1185
- feat: display utilization idle checks (utilization, threshold)
- feat: display Reservation's Elapsed Time and Idle timeout with `lablup-shields`.

image

**UPDATED**
If (utilization - threshold) / threshold < 1, change the font color to red.
image


inureyes:
@Sujin-Kim1 Some weird situation happens after checkout this PR. Rollback to main resolves the problem.

Here is a screenshot: 
image

PR #1654, opened 2023-03-25
fix: add margin to resource preset dialog

### LINKED ISSUE #1653: Give margin to resource preset creation dialog content
### Main idea

Unlike other creation / modification dialog, resource preset creation dialog uses tight margin on its content area.

Normal dialogs:
Screenshot 2023-03-25 at 2 07 14 PM

Preset creation dialog:

image

It could be better to unify the margin design. 

### Alternative ideas

_No response_

### Anything else?

_No response_

inureyes:
This PR resolves  #1653


src/components/backend-ai-storage-list.ts
```diff
@@ -147,7 +147,11 @@ export default class BackendAiStorageList extends BackendAIPage {
     MB: Math.pow(10, 6),
     GB: Math.pow(10, 9),
     TB: Math.pow(10, 12),
-    PB: Math.pow(10, 15)
+    PB: Math.pow(10, 15),
```
jopemachine: It seems this change is unrelated to this PR (maybe inserted into this accidentally)

inureyes: It will help bit-wise quota calculation, I guess.

Issue #1638, opened 2023-03-15
There remains a highlight that should disappear after activating the button.


Yaminyam:
### What Operating System(s) are you seeing this problem on?

macOS (Intel)

### Backend.AI version

22.09

### Describe the bug

This seems to be a similar issue to #1352.
Click the help button and the highlighting remains.
<img width="246" alt="image" src="https://user-images.githubusercontent.com/31057849/225262225-25c67e79-e2b7-4862-aa31-eb99523be8e1.png">

### To Reproduce

click the help button

PR #1655, opened 2023-03-25
fix : remain highlight after click on help button

kin4496:
follow-up of #1638 
remove highlight after click help button


inureyes (approved):
LGTM

PR #1668, opened 2023-03-30
feat: add support for setting `inference` as allowed session type

kyujin-cho:
This commit updates resource group list page to allow adding `inference` session type as allowed session type of scaling group.

### Screenshot(s)
| Before | After |
| --- | --- |
|Screenshot 2023-03-30 at 1 15 43 PM|Screenshot 2023-03-30 at 1 15 10 PM|

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -929,7 +929,8 @@
     "TimeoutSeconds": "Seconds",
     "ResourceGroupDetail": "Resource Group Detail",
     "RetriesToSkip": "Times",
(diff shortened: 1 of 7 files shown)


lizable (commented):
It works fine, but I did a few touch ups.
I just replaced `mwc-select` which has been used for representing "allowed session type" with `backend-ai-multi-select` since It's better to inform that those options are multi-selectable.

Here's the update that I made. If it seems fine, I'll merge it.

### Screenshot(s)
Screenshot 2023-03-30 at 1 46 02 PM

PR #1672, opened 2023-03-31
fix: add `enqueueOnly` option to image installation

Sujin-Kim1:
resolves https://github.com/lablup/giftbox/issues/196

Web UI's image download feature always returns 504 error since the session stays in PULLING status for more than request timeout.


lizable:
We really need SSE on receiving the progress (to measure how much time is left...) but let's set the timeout to infinity for now...

src/lib/backend.ai-client-esm.ts
```diff
@@ -2827,7 +2827,7 @@ class ContainerImage {
     if (Object.keys(resource).length === 0) {
       resource = {'cpu': '1', 'mem': '512m'};
     }
-    return this.client.createIfNotExists(registry + name, sessionId, resource, 600000, architecture).then((response) => {
+    return this.client.createIfNotExists(registry + name, sessionId, resource, Infinity, architecture).then((response) => {
       return this.client.destroy(sessionId);
```
lizable: @Sujin-Kim1 Does it really "treat" timeout as "Infinity" in `createIfNotExists`?
`createIfNotExists` calls [`wrapWithPromise`](https://github.com/lablup/backend.ai-webui/blob/07f27fc1974abc25fa2adfdb28644c5e9d651803/src/lib/backend.ai-client-esm.ts#L951), and It inside the `wrapWithPromise` eventually calls [`setTimeout`](https://github.com/lablup/backend.ai-webui/blob/07f27fc1974abc25fa2adfdb28644c5e9d651803/src/lib/backend.ai-client-esm.ts#L297-L308) which doesn't treat `Infinity` as is.
Could you check it again?

lizable (changes requested):
I left a comment related to `setTimeout()` function.

adrysn (changes requested):
This doesn't resolve this issue. Installing images still waits until timeout.

adrysn:
I updated the code. Please test and review. @Sujin-Kim1

PR #1673, opened 2023-03-31
fix: show correct msg when vfolder sharing fails

### LINKED ISSUE #1667: Let's show correct error msg when sharing vfolder fails
### Main idea

There are several cases where sharing a vfolder fails.

- Case 1. Inviter(the owner of vfolder) already invited Invitee.
- Case 2. Invitee is not in `ACTIVE` status.
- Case 3. Folder to share is `automount-folder`* (*Currently, We prevent sharing `automount` folder in terms of policy reason.)

We only display a single msg("Invitation failed. User may already be invited.") in every case until now, which needs to be improved. For example, Except for case 1, case 2 and 3 should display messages respectively:
- Case 2. Invitee is not available. Please check the account again or ask administrator.
- Case 3. You cannot share automount-folder.
And also, If we "prevent sharing `automount` folder for policy reason", then we shall hide related icons as well.

### Screenshot(s)
image

### Alternative ideas

_No response_

### Anything else?

_No response_

lizable:
This PR resolves #1667.
Related PR: [backend.ai#1204](https://github.com/lablup/backend.ai/pull/1204)

There are several cases of sharing vfolder failing since requesting sharing vfolder with multiple accounts is possible in the current UI. I categorized those cases below with screenshots:

> NOTE:
> To prevent specifying the existence of the user by malicious requests, I regarded a user that doesn't exist and a user not in `ACTIVE` status in the same category.
> Also, I created dummy accounts to test under following conditions:
> 
> | User email(account) | status*(based on keypair) | already shared(or requested sharing) |
> | ---| --- | --- |
> | `user@lablup.com`| `active` | X |
> | `inactive@lablup.com`| `inactive` | X |

#### Case 1. [Some positive] Some of the users are either inactive or don't exist and others exist and are not yet invited.
Requested sharing account: `user@lablup.com`, `unknown-user@lablup.com`
ERROR MSG: Folder sharing is not available for requested user(s): `unknown-user@lablup.com`
Screenshot 2023-03-31 at 5 05 52 PM

Requested sharing account: `user@lablup.com`, `unknown-user@lablup.com`
ERROR MSG: No one invited since invitation already exists
Screenshot 2023-03-31 at 5 06 06 PM

#### Case 3. [All negative] All of the users are either inactive or don't exist.
ERROR MSG: `unknown-user@lablup.com`, `anonymous@lablup.com`
MSG: Folder sharing is not available for requested user(s).

Screenshot 2023-03-31 at 5 06 46 PM


inureyes (approved):
LGTM

PR #1675, opened 2023-03-31
refactor: remove innerHTML-based content rendering on session list status dialog

inureyes:
This PR refactor the innerHTML-based content rendering on session list status dialog, to prevent potential page layout breaking.

`innerHTML` should be used for only 'content pouring', not for 'template rendering'. If not necessary, you should avoid using programming patterns that involve innerHTML as much as possible.

This PR also reverts https://github.com/lablup/backend.ai-webui/pull/1670, which is not needed anymore.


lizable (approved):
I just found another minor issue related to HTML based content and fixed it.
It seems working fine.

PR #1683, opened 2023-04-06
feature: save session config and load

### LINKED ISSUE #1681: Let's provide a feature to remember the last option of creating a new compute session
### Main idea

There are certain requests for re-using session configuration which is applied in former session creation; such as environment to execute, environment variables to apply, folder(s) to mount, amount of resource to allocate, etc.
Let's add an option to support saving and using session configuration according to the session created just before.

### Alternative ideas

If possible, we can store those configurations as a session template and add it to the list.

### Anything else?

_No response_

lizable:
This PR partially resolves #1681.
> ⚠️ NOTE: Task 2 will be implemented in continued PR.

- [ ] Phase 1. Implement simple storing feature using localstorage or sessionstorage.
   - [x] Task 1. Store/load essential session config
        - [x] Session Type
        - [x] Environment / Version (including manual image)
        - [x] Session Name
        - [x] Environment variables
        - [x] Folder to mount (including model storage if version supports)
        - [x] Resource Preset
        - [x] Resource amount to allocate (CPU Core, Memory (GiB), NPU, cluster mode, size of cluster, etc.)
   - [ ] Task 2. (TODO)
        - [ ] Session owner info (for delegation)
        - [ ] alias path information for mounted folder list
        - [ ] High-Performance Computing Optimization

### Screenshot(s)
Screenshot 2023-04-24 at 7 21 16 PM


yomybaby (changes requested):
Please check these cases:
- the number of RAM is saved correctly, But it isn't loaded properly.
- After set as default config with ENV variables, I can see the values are loaded.(hi=2)
image
But Env variable config modal is empty
image

lizable:
Closing this issue since it will be handled in #2025 with session template.

PR #1687, opened 2023-04-12
feat: add sorting button to project selector

Sujin-Kim1:
- resolves https://github.com/lablup/giftbox/issues/297

---

**🤔 Question**
~~Currently, if the sort is changed, `this.projectSelectionBox.firstChild` is deleted and the rearranged project items are newly added. **After sorting, `this.projectSelectionBox` is closed.**
To prevent `this.projectSelectionBox` from closing after sorting, select only the `mwc-list-item` type among the child nodes of `this.projectSelectionBox`, delete them, and add the rearranged project items again. However, this way seems time-consuming when the number of projects is large.
Please advise if you have any good ideas.~~

+) Added
We don't allow creating a project in webui. So I changed to render the project select in the `render()` instead of rendering dynamically.

+) Added
Now the selector is not closed after sorting.

---
**Screenshots**
Before
image

After
image
image


src/components/backend-ai-project-switcher.ts
```diff
@@ -66,55 +72,47 @@ export default class BackendAIProjectSwitcher extends LitElement {
...
-    this.groups = globalThis.backendaiclient.groups;
-    // Detached from template to support live-update after creating new group (will need it)
-    if (this.groupSelectionBox?.hasChildNodes()) {
-      this.groupSelectionBox?.removeChild(this.groupSelectionBox.firstChild as ChildNode);
+  _sortProjects(e) {
+    const isAscending = e.target.on;
+    if (isAscending) {
+      this.projects = [...this.projects.sort()];
+    } else {
+      this.projects = [...this.projects.reverse()];
```
yomybaby:
```suggestion
      this.projects = [...this.projects.sort().reverse()];
```
If there is no 'sort' here, 'this.projects' should already be sorted. It's better to implement it to avoid external dpendancy.

yomybaby (changes requested):

adrysn (commented):
As @Sujin-Kim1 already mentioned in the PR description, closing the project dropdown menu is counter-intuitive when a user clicks the sort button. I think the sort button is not necessary for the project selector.

The feature itself looks good.

inureyes (approved):
LGTM

PR #1688, opened 2023-04-13
ci: Add size label automation actions job

Yaminyam:
The size label was added while the label was recently reorganized.
Automatically manage the corresponding size label through actions.


inureyes (approved):
LGTM

PR #1689, opened 2023-04-17
feat: add endpoint validation

Sujin-Kim1:
follows #1470 
image

### DIFF
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -1040,7 +1040,7 @@ export default class BackendAILogin extends BackendAIPage {
    * Show signup dialog. And notify message if API Endpoint is empty.
    * */
   private _showSignupDialog() {
-    this.api_endpoint = this.api_endpoint.trim();
+    this.api_endpoint = this.apiEndpointInput.value || this.api_endpoint.trim();
     if (this.api_endpoint === '') {
       this.notification.text = _text('error.APIEndpointIsEmpty');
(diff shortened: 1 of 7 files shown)


src/components/backend-ai-login.ts
```diff
@@ -1040,7 +1040,7 @@ export default class BackendAILogin extends BackendAIPage {
    * Show signup dialog. And notify message if API Endpoint is empty.
    * */
   private _showSignupDialog() {
-    this.api_endpoint = this.api_endpoint.trim();
+    this.api_endpoint = this.apiEndpointInput.value || this.api_endpoint.trim();
```
inureyes: @Sujin-Kim1  Please check the logic whether it works.

inureyes (changes requested):
@Sujin-Kim1 double-check the endpoint logic whether || works as intended.

src/components/backend-ai-login.ts
agatha197: If we don't set the [apiEndpoint](https://github.com/lablup/backend.ai-webui/blob/main/configs/default.toml#L2) option, it is not available to signup because of the empty endpoint value.
So I modified it not only to allow the user to set the specific endpoint with input value but also to allow the signup dialog to float even when the `apiEndpoint` option is not set.
However, if the policy allows signup only with the `apiEndpoint` option, I will revert.

inureyes (approved):
LGTM

PR #1690, opened 2023-04-17
feature: folder upload 

Sujin-Kim1:
- resolves https://github.com/lablup/giftbox/issues/251
- related PR: https://github.com/lablup/backend.ai/pull/1233
- Added an upload folder button
> TODO: support uploading folders via drag and drop

image


src/components/backend-ai-storage-list.ts
```diff
@@ -2508,7 +2524,16 @@ export default class BackendAiStorageList extends BackendAIPage {
     for (let i = 0; i < this.uploadFiles.length; i++) {
       this.fileUpload(this.uploadFiles[i]);
     }
-    (this.shadowRoot?.querySelector('#fileInput') as HTMLInputElement).value = '';
+    if (isEmptyFileIncluded) {
+      this.notification.text = _text('data.explorer.EmptyFilesAreNotUploaded');
+      this.notification.show();
+    }
+    if (isFolderUpload) {
+      this.notification.text = _text('data.explorer.EmptyFoldersAreNotUploaded');
```
lizable: I found that It skips uploading not only for the empty folder(s) but also for the empty file(s). IMO, It's better to inform users about both conditions at once.

lizable (changes requested):
It works fine, but there's one thing to consider. Unlikely uploading file(s), it doesn't detect the directory name that already exists in the destination path. 
let's make sure users to confirm before upload operation begins as we already implemented in [here](https://github.com/lablup/backend.ai-webui/blob/4311b3848712bc4696504fe0bedc838cc21a01e5/src/components/backend-ai-storage-list.ts#L2405).

lizable (approved):
LGTM.

lizable:
Sorry, I double-checked it but, I think it doesn't work in certain conditions. Please refer to the video clip below:

https://user-images.githubusercontent.com/46954439/236799852-b236b4ef-a596-4311-a3c7-fabcee2e4927.mov

lizable (changes requested):
We need to resolve "failing folder upload" on certain conditions that I mentioned above before merging.

lizable:
> We need to resolve "failing folder upload" on certain conditions that I mentioned above before merging.

It resolved, dunno why but I suspect It's due to my local system crashing on Docker. (reinstalling docker smoothly fixed the issue.)
Anyway, It works fine with a neater interface so that user can see the progress of upload easier than before.

lizable (approved):

PR #1691, opened 2023-04-18
feat: show agent id to session list if `hideAgent` is false

Sujin-Kim1:
resolves https://github.com/lablup/giftbox/issues/249
display agent id even if the user is not admin if `hideAgent` is set to false.


inureyes (approved):
LGTM

yomybaby (approved):
LGTM

PR #1692, opened 2023-04-21
feature: rename usage tab and add info dialog in statistics page

Sujin-Kim1:
resolves https://github.com/lablup/giftbox/issues/337
image
image


resources/i18n/ko.json
```diff
@@ -1005,10 +1005,18 @@
...
     "1Day": "1일",
-    "1Week": "1주"
+    "1Week": "1주",
+    "UsageHistoryDesc": "사용 기록 탭에서는 연산 세션 사용과 관련된 간단한 통계를 그래프로 확인할 수 있습니다. 좌측 상단의 기간 선택 메뉴에서 이용 기간을 선택하여 하루 또는 일주일 동안의 사용 기록 통계를 확인할 수 있습니다. 표시되는 항목은 다음과 같습니다.",
+    "SessionsDesc": "생성된 연산 세션의 수입니다.",
+    "CPUDesc": "연산 세션이 생성되면서 할당한 CPU 코어의 수입니다.",
+    "MemoryDesc": "연산 세션이 생성되면서 할당한 메모리의 양입니다.",
+    "GPUDesc": "연산 세션이 생성되면서 할당한 GPU 단위의 수입니다. Fractional GPU 기능이 작동하고 있을 경우, 물리 GPU 와 일치하지 않을 수 있습니다.",
+    "IOReadDesc": "IO에서 받은 데이터의 양입니다.",
+    "IOWriteDesc": "IO로 전송된 데이터의 양입니다.",
```
adrysn: IO (Input/Output) is not a device, so this expression is strange. How about:

* 연산 세션에서 읽어들인(read) 데이터의 크기입니다. 단, 데이터 폴더 등 외부에서 마운트 된 장치에서 읽어들인 데이터 양은 포함되지 않습니다.
* 연산 세션에서 기록한(write) 데이터의 크기입니다. 단, 데이터 폴더 등 외부에서 마운트 된 장치에서 생성한 데이터의 양은 포함되지 않습니다.

adrysn (changes requested):
I have minor comments on the description of I/O stats.

Also, I think it is hard to know if I can open a dropdown menu since there is no background color or borderline in `vaadin-select` components (the top project select box as well). In my opinion, it might be better to clearly distinguish the components as clickable menus, not just text.

image

yomybaby (commented):
@Sujin-Kim1   How about moving the note from bottom of Info modal to the top? It might be better to first inform users that only terminated session logs are displayed.
image

resources/i18n/ko.json
agatha197: Thanks for the kind feedback! I reflected them.

Sujin-Kim1:
> kimsujin (@Sujin-Kim1) How about moving the note from bottom of Info modal to the top? It might be better to first inform users that only terminated session logs are displayed. image

I removed the note to the top! Please check again :)

adrysn (approved):
LGTM.

PR #1701, opened 2023-04-27
feat: set `allowed_session_types` default value and make required.

### LINKED ISSUE #1700: Can't create a new resource group

Linux (x86-64)

### Backend.AI version

22.09.19

### Describe the bug

When creating a new scaling group, 400 Bad Request occurs.
image

### To Reproduce

Create a new resource group in the agent page - resource group tab.

### Expected Behavior

Create resource group without error.

### Anything else?

_No response_

Sujin-Kim1:
- resolves #1700 
- Resolving to create resource-group error (not allowed to create with empty resource group)
- Set `allowed_session_types` default value as `interactive, batch`.
- If the allowed session type is not selected, the border color will be changed to red and disallowed to save.

[Screenshots]
- Add a red border and validation message if allowed session type is empty.
image


lizable (changes requested):
It works fine, but I guess we can handle this issue in more clever and maintainable way.
IMO it's better to add `required` property and `validation message` to backend-ai-multi-select component rather than just changing the border color to red. How about you @Sujin-Kim1 ?

Sujin-Kim1:
> It works fine, but I guess we can handle this issue in more clever and maintainable way. IMO it's better to add `required` property and `validation message` to backend-ai-multi-select component rather than just changing the border color to red. How about you kimsujin (@Sujin-Kim1) ?

I think it's better to add all of them(required, validation message, changing border color). Thank you for your opinion.

PR #1703, opened 2023-05-10
feat: show idle check information based on only applied configurations#361

Sujin-Kim1:
Since the policy of the session can be known, only the description of the set Idle checker is visible.
resolves https://github.com/lablup/giftbox/issues/361

+ Updated)
Hide idle checker column if manager doesn't support idle checks or no configuration

Example
image
image

### DIFF
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -364,7 +364,7 @@
     "UtilizationThreshold": "Utilization Threshold",
     "ExpiresAfter": "Expires After",
     "GracePeriod": "Grace Period",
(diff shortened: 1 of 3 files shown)


lizable (changes requested):
How about removing the `idle-checks` column if there's no configuration used for idle-checks?
When I delete all of `idle-checks` configuration from setting, and the information dialog looks like this. The following description seems a bit unnecessary to me.

## Screenshot(s)
Screenshot 2023-05-10 at 3 12 17 PM

Sujin-Kim1:
> 

That's a good idea! I reflected it :)

adrysn (approved):
LGTM.

PR #1705, opened 2023-05-12
fix: utilization values in idle checker pop-up are not updated after the initial rendering

Sujin-Kim1:
duplicate renderings in the idle checker column occur invalid utilization of idle checker values. When filtering the session list, there is a problem with taking advantage of the idle checker value of the session that was at the top of the list before filtering.

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -2505,7 +2505,7 @@ export default class BackendAISessionList extends BackendAIPage {
       let button;
       if (key === 'utilization') {
         button = html`
-            <button
+          <button
             class="idle-check-key"
             style="color:#42a5f5;"
             @mouseenter="${(e) => this._createUtilizationIdleCheckDropdown(e, rowData.item.idle_checks?.utilization?.extra?.resources)}"
@@ -2516,7 +2516,7 @@ export default class BackendAISessionList extends BackendAIPage {
         `;
       } else {
         button = html`
-            <button
+          <button
             class="idle-check-key"
             style="color:#222222;"
           >
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -2506,7 +2506,6 @@ export default class BackendAISessionList extends BackendAIPage {
       if (key === 'utilization') {
         button = html`
             <button
-            id="${key}"
             class="idle-check-key"
             style="color:#42a5f5;"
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-session-list.ts
```diff
@@ -2505,17 +2502,36 @@ export default class BackendAISessionList extends BackendAIPage {
...
+            @mouseenter="${(e) => this._createUtilizationIdleCheckDropdown(e, rowData.item.idle_checks?.utilization?.extra?.resources)}"
+            @mouseleave="${() => this._removeUtilizationIdleCheckDropdown()}"
+          >
+            ${_text('session.' + this.idleChecksTable[key])}
+          </button>
+        `;
+      } else {
+        button = html`
+            <button
+            id="${key}"
```
yomybaby: We no longer need to set the `id`, so let's remove it. If we do need an id, it should be changed to a unique one. Setting `id` to `key` (utiliazation..) can cause it to be used more than once in the same document.

yomybaby (changes requested):

src/components/backend-ai-session-list.ts
agatha197: Thanks! I removed it.

src/components/backend-ai-session-list.ts
```diff
@@ -2505,17 +2502,34 @@ export default class BackendAISessionList extends BackendAIPage {
         );
       }
 
+      let button;
+      if (key === 'utilization') {
+        button = html`
+            <button
```
adrysn: 4 spaces, but 2 spaces is preferable.

src/components/backend-ai-session-list.ts
```diff
@@ -2505,17 +2502,34 @@ export default class BackendAISessionList extends BackendAIPage {
...
+            style="color:#42a5f5;"
+            @mouseenter="${(e) => this._createUtilizationIdleCheckDropdown(e, rowData.item.idle_checks?.utilization?.extra?.resources)}"
+            @mouseleave="${() => this._removeUtilizationIdleCheckDropdown()}"
+          >
+            ${_text('session.' + this.idleChecksTable[key])}
+          </button>
+        `;
+      } else {
+        button = html`
+            <button
```
adrysn: Same here.

adrysn (approved):
LGTM except minor indentation issue.

PR #1710, opened 2023-05-18
feat: display allocated shared memory in session config column

lizable:
This PR adds a simple user interface for "shared memory" allocated in each session.
Please note that shared memory also follows the policy; which aggregates all allocated amounts of the same resources in each container on multi-container conditions; as the other resources do.
related PR: [backend.ai#1282](https://github.com/lablup/backend.ai/pull/1282)

### Screenshot(s)
* Before
Screenshot 2023-05-18 at 9 32 54 PM

* After
Screenshot 2023-05-18 at 9 31 46 PM

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -127,6 +127,7 @@ export default class BackendAISessionList extends BackendAIPage {
   @property({type: String}) _connectionMode = 'API';
   @property({type: Object}) notification = Object();
   @property({type: Boolean}) enableScalingGroup = false;
+  @property({type: Boolean}) isDisplayingAllocatedShmemEnabled = false;
   @property({type: String}) listCondition: StatusCondition = 'loading';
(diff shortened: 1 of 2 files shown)


adrysn (changes requested):
With 22.09 backend.ai core, the shared memory is always displayed as `0`.
image

lizable:
> With 22.09 backend.ai core, the shared memory is always displayed as `0`. image

@adrysn 
I think it's because of the backporting issue in Backend.AI Core. Should it be implemented in 22.09 as well?
If so, we need to backport the changes in 22.09 in Backend.AI Core. If not, then I'll just add a compatibility check condition statement in WebUI.

lizable:
I just confirmed that amount of allocated shared memory at session config column in session list is only displayed in Backend.AI Core 23.03.

### screenshot(s)
- 22.09
Screenshot 2023-05-23 at 11 11 21 AM

- 23.03
Screenshot 2023-05-23 at 11 20 56 AM

PR #1716, opened 2023-05-22
feat: Support max_password_age

yomybaby:
## Description
copilot:summary

## Explanation of Changes
copilot:walkthrough

### DIFF
--- a/react/src/components/ResetPasswordRequired.tsx
+++ b/react/src/components/ResetPasswordRequired.tsx
@@ -1,0 +1,121 @@
+import { Button, Form, Modal, Typography, Input, theme } from "antd";
+import { useWebComponentInfo } from "./DefaultProviders";
+import Flex from "./Flex";
+import { useTranslation } from "react-i18next";
+import { WarningTwoTone } from "@ant-design/icons";
+
+// src/components/backend-ai-common-utils.ts
+export const passwordPattern = /^(?=.*\d)(?=.*[a-zA-Z])(?=.*[_\W]).{8,}$/;
+
+const ResetPasswordRequired = () => {
+  const { value, dispatchEvent } = useWebComponentInfo();
+  const open = value === "true";
+  const { token } = theme.useToken();
+  const { t } = useTranslation();
+  return (
+    <Modal
+      open={open}
+      centered
+      mask={false}
+      onCancel={() => {
+        dispatchEvent("cancel", null);
+      }}
+      onOk={() => {
+        dispatchEvent("ok", null);
+      }}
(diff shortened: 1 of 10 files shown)


lizable:
⚡️ Quick question:
There are two cases in that `max-password-age` expires. Whether the user is already logged in or not.
What I want to know is about the first case, when the user is already logged in.
It doesn't expire when the user logged in already.
I confirmed that the session doesn't regard as invalid even if `max-password-age` already expired.
I wonder whether this is intended or not. Could you give me the answer @yomybaby?

yomybaby:
@lizable  Yes. It works properly. This is intended. 
You can check that [here](https://teams.microsoft.com/l/message/19:d121a283119b4d509b53f4f08db17f9d@thread.tacv2/1684831585562?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=b1f3bcf4-facf-40d3-94ab-169abb4f0f52&parentMessageId=1684382020774&teamName=customers%20%26%20clients&channelName=LotteDCC&createdTime=1684831585562)

react/src/components/ResetPasswordRequired.tsx
```diff
@@ -0,0 +1,177 @@
...
+import { parse } from "path";
+import { useAnonymousBackendaiClient } from "../hooks";
+import { baiSignedRequestWithPromise } from "../helper";
+import { useEffect } from "react";
+// import * as ai from "../lib/backend.ai-cleint-esm_on_react";
+// import { useSuspendedBackendaiClient } from "./BackendaiClientProvider";
+
+// src/components/backend-ai-common-utils.ts
+export const passwordPattern = /^(?=.*\d)(?=.*[a-zA-Z])(?=.*[_\W]).{8,}$/;
+
```
lizable, line 13: This is just a minor thought, but I wonder if we could store password patterns into a single component(just came to my mind), so that we could import and apply it and change it if the policy updates.

lizable (approved):
LGTM. Btw I just left a request that may be right for the next issue.

react/src/components/ResetPasswordRequired.tsx
lizable: We have several components that need password patterns such as creating users, or modifying user information.

yomybaby: Thank you for the suggestion. For now, I think this is okay to export the reg expression only, it's more flexibility in composition. But it would be better to move this pattern to another file.

lizable (approved):
LGTM!

Issue #1695, opened 2023-04-25
Setup GUI for is_public field to scaling_group


inureyes:
### Main idea

This is follow-up issue of lablup/backend.ai#1236

```[tasklist]
### Tasks
- [ ] https://github.com/lablup/backend.ai-webui/pull/1718
```

PR #1718, opened 2023-05-24
feat: add setup GUI for `is_public` to resource group

Sujin-Kim1:
resolves https://github.com/lablup/backend.ai-webui/issues/1695

Screenshots
---
- Add `Public` column
image
- Change `is_active`, `scheduler`, and `is_public` to be configurable when creating or modifying a resource group.
- The is_active and scheduler setup layout goes down.
- Display resource group name when modifying.

| Before | After |
|--------|------|
| image | image |
| image | image |


inureyes (approved):
LGTM

PR #1733, opened 2023-06-05
feat: automatically fill sign up token if query param exists

kyujin-cho:
This PR updates `backend-ai-login` and `backend-ai-signup` component to automatically open sign up popup and fill out token to value passed by query parameter `token` if exists.


lizable (approved):
LGTM.

PR #1740, opened 2023-06-14
fix: manually add roundup decimal places in used_slots and total_slots

### LINKED ISSUE #1587: Need to clamp decimal places in resource progress bar in resource monitor panel to at most 2

Linux (x86-64)

### Backend.AI version

22.09.10

### Describe the bug

Unlikely to control-panel, webUI resource gauge doesn't have any fixed decimal places, which could break the text layout.

### Screenshot(s)

image

### To Reproduce

_No response_

### Expected Behavior

even if the value received from server (by request) has long decimal point, it should be displayed human-friendly value. IMO, at most 2 decimal places would be enough.

### Anything else?

_No response_

lizable:
This PR resolves #1587.

### DIFF
--- a/src/components/backend-ai-resource-broker.ts
+++ b/src/components/backend-ai-resource-broker.ts
@@ -588,7 +588,8 @@ export default class BackendAiResourceBroker extends BackendAIPage {
         total_project_slot['cuda_shares'] = parseFloat(total_project_slot['cuda_shares']).toFixed(2);
       }
 
-      this.total_slot = total_slot;
+      // clamp into two decimal point as maximum
+      this.total_slot = this._roundResourceDecimalPlaces(total_slot);
       if (!globalThis.backendaiclient._config.hideAgents) {
         // When `hideAgents` is false, we display the total resources of the current resoure group.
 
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-resource-broker.ts
```diff
@@ -756,6 +758,20 @@ export default class BackendAiResourceBroker extends BackendAIPage {
     });
   }
 
+  _roundResourceDecimalPlaces(resourceSlots: Object, roundUpNumber = 2) {
+    Object.keys(resourceSlots).map((resource) => {
+      // convert undefined or NaN to 0
+      let resourceValue = Number(resourceSlots[resource] ?? 0).toString();
```
agatha197: If `resourceSlots[resource]` is `NaN`, it occurs `TypeError: Cannot read properties of undefined (reading 'length')`. 
```suggestion
      let resourceValue = Number(resourceSlots[resource] ? resourceSlots[resource] : 0).toString();
```

src/components/backend-ai-resource-broker.ts
```diff
@@ -756,6 +758,20 @@ export default class BackendAiResourceBroker extends BackendAIPage {
...
+
+      // clamp to roundUpNumber if the number of decimal place exceeds
+      if (!Number.isInteger(Number(resourceValue)) && resourceValue.split('.')[1].length > roundUpNumber) {
+        resourceValue = (Math.round(Number(resourceValue)* 100) / 100).toFixed(2);
```
agatha197, line 743:
```suggestion
        resourceValue = (Math.round(Number(resourceValue) * 100) / 100).toFixed(2);
```

Sujin-Kim1 (approved):
LGTM

PR #1748, opened 2023-06-20
docs: Modified to fit nvidia brand notation

Yaminyam:
When used as a logo, it is nVIDIA, and when written, NVIDIA seems to be the correct expression.
https://www.nvidia.com/en-us/about-nvidia/legal-info/logo-brand-usage/


inureyes (approved):
LGTM

Issue #1751, opened 2023-06-26
"Keep Login Session Information while Logout" option not working on electron app


chisacam:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

23.03.1(62cfb6f9)

### Describe the bug

Even if I turn on the "Keep Login Session Information while Logout" option, I still have to log in again when I exit and reopen the app.

OS: macOS 14, sonoma developer beta 2
HW: macbook pro 16 2021, m1 pro

### To Reproduce

1. login into app.
2. open option.
3. turn on `Keep Login Session Information while Logout` option.
4. close app.
5. reopen app.

### Expected Behavior

app will request login to you again.

### Anything else?

When I cloned the project and ran it in the Electron environment and looked at the developer tools, the first problem was that when I closed the app, it always cleared the sessionid in local storage regardless of the "Keep Login Session Information while Logout" option.

```typescript
  async close_app_window(performClose = false) {
    if (globalThis.backendaioptions.get('preserve_login') === false) { // Delete login information.
      ...
    }
    if (typeof globalThis.backendaiclient != 'undefined' && globalThis.backendaiclient !== null) {
      if (globalThis.backendaiclient._config.connectionMode === 'SESSION') {
        await globalThis.backendaiclient.logout();  // <- looks like execute every app close.
      }
      globalThis.backendaiclient = null;
    }
  }
```

```typescript
  logout() {
    let body = {};
    let rqst = this.newSignedRequest('POST', `/server/logout`, body, null);
    // clean up log msg for security reason
    const currentLogs = localStorage.getItem('backendaiwebui.logs');
    if (currentLogs) {
      localStorage.removeItem('backendaiwebui.logs');
    }
    localStorage.removeItem('backendaiwebui.sessionid'); // <- and not check option.
    return this._wrapWithPromise(rqst);
  }
```

Secondly, even if you don't clear the session ID in local storage when closing the app, the process of clearing the information in local storage when login (authentication) fails is executed due to the first problem or other complex problems.

```typescript
      if (result.authenticated === true) {
        if (result.data.role === 'monitor') {
          await this.logout();
          return Promise.resolve({fail_reason: 'Monitor user does not allow to login.'});
        }
        await this.get_manager_version();
        if (this._loginSessionId !== null && this._loginSessionId !== '') {
          localStorage.setItem('backendaiwebui.sessionid', this._loginSessionId);
        }
        return this.check_login();
[...]

PR #1752, opened 2023-06-26
fix: do not auto logout when preserve login true

chisacam:
- just move logout inside preserve_login check block...
- resolves https://github.com/lablup/backend.ai-webui/issues/1751


inureyes (commented):
LGTM

PR #1760, opened 2023-07-05
fix: annotate unused codes

Sujin-Kim1:
for fix the build errors. these codes will be used in the future, so I just annotate them.


inureyes (changes requested):
src/components/StorageHostSettingsPanel.tsx
  Line 29:11:  'token' is assigned a value but never used  @typescript-eslint/no-unused-vars
  
Also need to follow-up.

inureyes (approved):

PR #1763, opened 2023-07-05
cosmetic changes of quota scope setting page

yomybaby:
- Match button icon and styles
- Change the ID column name of the quota setting to "Quota Scope ID"
image


Sujin-Kim1:
LGTM!

PR #1765, opened 2023-07-05
compact ui version of storage quota setting

yomybaby:
| Before | After |
|--------|------|
| image | image |


Sujin-Kim1 (approved):
LGTM!

PR #1768, opened 2023-07-06
feat: merge running and others tab in the session list

### LINKED ISSUE #1767: Merge RUNNING and OTHERS tabs in Web UI session list
### Main idea

If the session status becomes `ERROR`, users have to click OTHERS to see it. Unless you click OTHERS, users can think that resources are lacking resources even though resources are not being used. This is because the session in the ERROR state also occupies resources. So let's merge the RUNNING and OTHERS tabs.

### Alternative ideas

_No response_

### Anything else?

_No response_

Sujin-Kim1:
resolves #1767 
image

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -479,7 +479,7 @@ export default class BackendAISessionList extends BackendAIPage {
   }
 
   get _isRunning() {
-    return ['batch', 'interactive', 'inference', 'system', 'running'].includes(this.condition);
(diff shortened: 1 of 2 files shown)


lizable (approved):
It seems okay to me. Btw, I think we could solve with this issue by providing UI (perhaps multi-select UI?) to make user filter by status. I'll create an issue for that if @Sujin-Kim1 and @yomybaby agree with this idea.

Sujin-Kim1:
> 

Cool! Let's wrap up this issue and handle it as another one! Thanks.

PR #1769, opened 2023-07-06
feat: Remove resource policy panel on Quota setting and bugfix

yomybaby:
- remove resource policy panel 
- use email value as a the value of UserSelector's options
- sort users by email on UserSelecotr
- increase the limit of UserSelecotor query to 150


Sujin-Kim1 (changes requested):
image
When selecting the user and clicking the selector again, it is not possible to delete the user's email. Can you provide the UI selected from the user list as shown below when I made a selection rather than a search?
image

Sujin-Kim1 (approved):
LGTM!

PR #1776, opened 2023-07-11
fix: `newPublicRequest` -> `newSignedRequest` when getting resource-slots

Sujin-Kim1:
Because of changes to throwing X-BackendAI-SessionID instead of cookies, it occurs a 401 error because the `newSignedRequest` or authentication header does not pass. To resolve this, change to use `newSignedRequest` when importing resource-slots.


inureyes:
LGTM

PR #1783, opened 2023-07-18
fix: update the text of launch button using @property

yomybaby:
## Problem
There is an error when setting the content of a lit element using textContent.

> Uncaught (in promise) Error: This `ChildPart` has no `parentNode` and therefore cannot accept a value. This likely means the element containing the part was manipulated in an unsupported way outside of Lit's control such that the part's marker nodes were ejected from DOM. For example, setting the element's `innerHTML` or `textContent` can do this.

## Changes
- update the text of launch button using @property
- set _text() value using `textNode` instead of `innnerHTML`


inureyes (approved):
LGTM

Issue #1799, opened 2023-07-20
Registry modification button not working


inureyes:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

23.09.0 alpha 2

### Describe the bug

<img width="826" alt="image" src="https://github.com/lablup/backend.ai-webui/assets/949915/caf08fcf-2226-4384-8a85-3b144583a2ba">

When pressing registry modification button is pressed, it shows error and dialog does not spawn. After the error, when I click it again, it only shows registry category option, without any other options.

<img width="487" alt="image" src="https://github.com/lablup/backend.ai-webui/assets/949915/6848bdf4-7090-47d5-bc89-8ce4ffd9d2fc">

### To Reproduce

1. Goto `/environment` and choose Registry tab
2. Press gear button.

yomybaby:
this is resolved by https://github.com/lablup/backend.ai-webui/pull/1786

PR #1803, opened 2023-07-24
feature: Proxy react server to web dev server by using `devServer` and `proxy`

### LINKED ISSUE #1779: Configurable bind host and port for craco server
### Main idea

On our current NPM script, when starting create-react-app server by executing `craco start`, the server always binds to `127.0.0.1:3081` as this value is hardcoded to `react/.env.development` file.
Screenshot 2023-07-13 at 7 41 59 PM
Let's find out a way for user to override this behaviour, by passing their own `HOST` and `PORT` variable when starting server.

### Alternative ideas

_No response_

### Anything else?

_No response_
```[tasklist]
### Tasks
- [x] change port number so that developer can connect to the same port
- [ ] test on a remote environment
- [x] update README.md to guide how to change the port and host
```

Choi-Jiwon-38:
This PR resolves #1779


react/.env.development
```diff
@@ -1,4 +1,2 @@
-# disable opening browser on start
-BROWSER=none
 HOST=127.0.0.1
 PORT=3081
```
yomybaby: Please change the port number so that the developer can connect to the same port as before.

yomybaby: Also, did you test on a remote environment such as vmfarm?

yomybaby (changes requested):

react/.env.development
yomybaby: Additionally, let's update README.md to guide how to change the port and host for the remote environment

yomybaby: When ready to review again, please move #1799 to the "In review" column on the Daily board

yomybaby (approved):
LGTM. 👊

yomybaby:
This PR is related to https://github.com/lablup/backend.ai-webui/issues/1805

PR #1819, opened 2023-07-31
fix: unable to terminate a compute session when wsproxy address is not properly

pderer:
There was a bug that the session was not deleted when the wsproxy address of the scaling group was invalid.

This PR is related to this PR in backend.ai repository. [#1423](https://github.com/lablup/backend.ai/pull/1423)

In order to prevent side effects caused by server-side code modification, it was modified to handle exceptions in the web UI-side code.

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -1462,12 +1462,8 @@ export default class BackendAISessionList extends BackendAIPage {
     });
   }
 
-  _requestDestroySession(sessionId, accessKey, forced, err = null) {
+  _requestDestroySession(sessionId, accessKey, forced) {
     globalThis.backendaiclient.destroy(sessionId, accessKey, forced).then((req) => {
-      if (err) {
-        this.notification.text = PainKiller.relieve(err["title"]);
-        this.notification.show(true, err);
-      }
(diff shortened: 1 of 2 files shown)


src/components/backend-ai-session-list.ts
```diff
@@ -1462,35 +1462,47 @@ export default class BackendAISessionList extends BackendAIPage {
...
-        if ('description' in err) {
-          this.notification.text = PainKiller.relieve(err.description);
-        } else {
-          this.notification.text = PainKiller.relieve('Problem occurred during termination.');
-        }
-        this.notification.show(true, err);
-      });
+      this._requestDestroySession(sessionId, accessKey, forced);
     }).catch((err) => {
       console.log(err);
```
fregataa: could you remove this console.log?
it is not written in this pr but it would be good!

fregataa (commented):
Thanks! it seems good for me.
Please wait for other's review!

src/components/backend-ai-session-list.ts
pderer: Thanks for reviewing!
I will fix it.

inureyes (approved):
LGTM

PR #1833, opened 2023-08-04
fix: http status code

pderer:
This PR is related to these PRs
- #1819 in webui repository.
- [#1423](https://github.com/lablup/backend.ai/pull/1423) in core repository

Using 404 http status code, it can be assumed that the API endpoint does not exist.
Since this error is a configuration error, 500 status code is appropriate.

For backward compatibility with 22.09, 404 status code should be included.


inureyes (approved):
LGTM

PR #1836, opened 2023-08-08
hotfix: custom.css and theme.json for react component

yomybaby:
- feat: Add `link` to custom.css for React-based component
- refactor: Remove custom `<ConfigProvider/>`
- fix: Remove hard-coded color values in the antd component.

With this pull request, react based components(and antd components) can be customized by updating `custom.css` and `theme.json`, even in production builds.
- The `Tabs` component can be customized using `resources/theme.json`, which is an antd theme configuration file.
- The `Tabs` component can also be customized using `custom.css`.

```css
/* custom.css */
.ant-tabs {
  background-color: #2A2C30;
}
  ```


Sujin-Kim1:
the button color and active tab font color are brighter than web component. Let's change [this line](https://github.com/lablup/backend.ai-webui/blob/main/resources/theme.json#L4) to `#27824F`, too.

PR #1841, opened 2023-08-08
add: Ubuntu mono font for monospace text

inureyes:
Provide local ubuntu mono font for consistent text glyphs.


Sujin-Kim1 (approved):
LGTM

PR #1847, opened 2023-08-10
fix:  Remove license errors in open-source project version

### LINKED ISSUE #1812: Remove license-related console errors from the console window in an open-source project version.

Linux (x86-64)

### Backend.AI version

23.03

### Describe the bug

Remove license-related errors from the console window in an open-source project version that does not have the activator plugin installed. Because in the absence of the plugin, the license-related URL itself does not exist.
image
This error should be removed.

### To Reproduce

1. go to the information page.
2. see an error.

### Expected Behavior

no errors if the user uses an open-source project version that does not have the activator plugin installed.
if not, the validation of the license should work properly (if the license is expired, occur the error.)

### Anything else?

_No response_
```[tasklist]
### Tasks
- [x] handle error when non-exist URL type
```

Choi-Jiwon-38:
This PR resolves #1812

|Result|
|------|
|스크린샷 2023-08-10 13 37 43|

### DIFF
--- a/react/src/components/Information.tsx
+++ b/react/src/components/Information.tsx
@@ -49,9 +49,7 @@ const Information: React.FC<InformationProps> = () => {
   }>(
     "licenseInfo",
     () => {
-      return baiClient.enterprise.getLicense().catch((e: any) => {
-        if (e.statusCode === 404) return null;
-      });
+      return baiClient.enterprise.getLicense();
     },
     {
       // for to render even this fail query failed
--- a/react/src/components/Information.tsx
+++ b/react/src/components/Information.tsx
@@ -49,7 +49,9 @@ const Information: React.FC<InformationProps> = () => {
   }>(
     "licenseInfo",
     () => {
-      return baiClient.enterprise.getLicense();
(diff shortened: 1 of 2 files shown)


Sujin-Kim1 (approved):
LGTM

yomybaby (changes requested):
This is not a good solution, because it complicates the code flow. 
1. there is no resp.ok, so it gets thrown
2. it falls into the catch and is thrown again inside the catch, but if it's a 404, it exits the catch block without throwing. 

For now, it's better to treat a 404 from Information as an error rather than changing the fetch handling logic in `backend.ai-client-esm.ts`. 
For example:
```
return baiClient.enterprise.getLicense().catch((e: any) => {
        if (e.statusCode === 404) {
          return null;
        }
      });
```

Choi-Jiwon-38:
At first, I thought changing the handling logic will be good in another case where 404 occurs.
But, I think it is better to treat from `information.tsx`, not `backend.ai-client-esm.ts`  now. 

Thanks for reviewing my PR, I reflected on your comments! 😊

yomybaby (approved):
LGTM!

PR #1842, opened 2023-08-09
fix: Add the React component to i18n scan target

Choi-Jiwon-38:
This PR resolves #1828

PR #1849, opened 2023-08-10
Refactor user pref modal

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui-ossca-2023/issues/12

스크린샷 2023-08-10 오후 2 35 22
Before PR

258709691-c5bf20f4-58d7-42fa-855b-6539ede311c2
After PR


Sujin-Kim1 (changes requested):
Please fix the lint errors and rename `UserPrefModal` to `UserProfileSettingModal` (also `backend-ai-react-user-pref-modal`)

react/src/components/UserPrefModal.tsx
```diff
@@ -0,0 +1,300 @@
...
+    loggedAccount: {
+      access_key: string;
+    };
+    keyPairInfo: {
+      keypairs: [{
+        access_key: string;
+        secret_key: string;
+      }];
+    }
+  };
```
agatha197: It's better to query user info in this component rather than getting it from the props. Please replace these lines with querying using graphQL or `useSuspendedBackendaiClient`.

react/src/components/UserPrefModal.tsx
```diff
@@ -0,0 +1,300 @@
...
+          access_key: selectOptions[i].label,
+          secret_key: selectOptions[i].value
+        })
+        break;
+      }
+    }
+    form.setFieldsValue({
+      full_name: full_name
+    });
+  });
```
agatha197: How about using `initialValues`? ([ref](https://ant.design/components/form))

react/src/components/UserPrefModal.tsx
```diff
@@ -0,0 +1,300 @@
...
+      email: string
+    }) => {
+      return baiSignedRequestWithPromise({
+        method: "POST",
+        url: `/auth/update-full-name`,
+        body,
+        client: baiClient,
+      });
+    }
+  });
```
agatha197: It can be written more simply by referring to the [following](https://github.com/lablup/backend.ai-webui/blob/1debfdce78b335728bb3d553f97c865b2c17b346/react/src/components/SessionListColums/SessionInfoCell.tsx#L56-L62). (e.g. `baiClient.update_full_name(<params>)`)

react/src/components/UserPrefModal.tsx
```diff
@@ -0,0 +1,300 @@
...
+  const mutationToUpdateUserPassword = useTanMutation({
```
agatha197: same as above.

Sujin-Kim1 (changes requested):
by the https://github.com/lablup/backend.ai-webui/pull/1842, we can trace translation-needed files automatically. Please merge the main branch and run `make i18n`.

react/src/components/UserPrefModal.tsx
gee05053: @Sujin-Kim1
Thanks for the review!
I have question this review.
I think input of full_name is correct to use initialValue according to your review.
But access_key and secret_key is correct to use setFieldsValue. Because access_key need to check if it same loggedAccount. And secret_key is pair of access_key.
is it correct to use setFieldsValue??🧐

agatha197: Ah, I didn't consider this uses a select component. Then, with reference this code, set `secret_key` automatically.   https://github.com/lablup/backend.ai-webui/blob/51852efbb696c20e94a1055fe6db85e234e85eaf/react/src/components/VFolderSelect.tsx#L86-L97

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,287 @@
...
+        if (loggedAcount === selectOptions[i].label) {
+          form.setFieldsValue({
+            access_key: selectOptions[i].label,
+            secret_key: selectOptions[i].value
+          })
+          break;
+        };
+      };
+    };
+  };
```
agatha197: these codes can be shortened if you use `_.find()`.
ref: https://github.com/lablup/backend.ai-webui/blob/51852efbb696c20e94a1055fe6db85e234e85eaf/react/src/components/ImageEnvironmentSelectFormItems.tsx#L97

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,287 @@
...
+      old_password: string;
+      new_password: string;
+      new_password2: string;
+    }) => {
+      return baiClient.update_password(values.old_password, values.new_password, values.new_password2);
+    }
+  });
+
+  const _onSelectAccessKey = (value: string) => {
+    for (let i=0; i< keyPairInfo.keypairs.length; i++) {
```
agatha197: same as above.

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,287 @@
...
+    };
+  };
+
+  const _showKeypairInfo = async(email: string) => {
+    keyPairInfo = await baiClient.keypair.list(email, ["access_key", "secret_key"], true);
+    if (keyPairInfo.keypairs) {
+      for (let i=0; i < keyPairInfo.keypairs.length; i++) {
+        selectOptions.push({value: keyPairInfo.keypairs[i].secret_key, label: keyPairInfo.keypairs[i].access_key})
+      };
+      for(let i=0; i<selectOptions.length; i++) {
```
agatha197:
```suggestion
      for (let i = 0; i < selectOptions.length; i++) {
```
If you really want to use for loop, please take care of the code convention.

react/src/components/UserPrefModal.tsx
gee05053: @Sujin-Kim1
Thank you😍

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,287 @@
...
+            secret_key: selectOptions[i].value
+          })
+          break;
+        };
+      };
+    };
+  };
+
+  _showTotpRule();
+  _showKeypairInfo(email);
```
agatha197: This function would be called many times. (if you want to test it, print console.log) Rather than calling this function here, just use a variable such as [totpSupported](https://github.com/lablup/backend.ai-webui/blob/51852efbb696c20e94a1055fe6db85e234e85eaf/react/src/components/UserInfoModal.tsx#L42).

agatha197: Naming: And please use `_show<something>()` when you want to display some component such as modal.

Sujin-Kim1 (changes requested):
- Center the modal so that the user can see the content directly without scrolling.
image
- Please fix the lint errors. You can automatically fix the errors with prettier, lint plugins.
  - suggest:
      https://marketplace.visualstudio.com/items?itemName=dsznajder.es7-react-js-snippets
      https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

Sujin-Kim1 (changes requested):
image
Let's add a list of `access_key` and automatically select the current `access_key`.

react/src/components/UserProfileSettingModal.tsx
gee05053: I think I should use this loop.😭
I modified code using prettier!

gee05053:
> image Let's add a list of `access_key` and automatically select the current `access_key`.

@Sujin-Kim1
I soleved this problem! if you have time, please review one more time🙏🏻

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,314 @@
...
+      <Modal
+        open={isOpen}
+        okText={t("webui.menu.Update")}
+        cancelText={t("webui.menu.Cancel")}
+        onCancel={() => dispatchEvent("cancel", null)}
+        onOk={() => _onSubmit()}
+        centered
+      >
+        <h2>{t("webui.menu.MyAccountInformation")}</h2>
+        <Divider />
```
agatha197:
```suggestion
        centered
        title={t("webui.menu.MyAccountInformation")}
      >
```

agatha197: The modal title design will be unified by https://github.com/lablup/backend.ai-webui/issues/1876. So let's just add a title using title property.

src/components/backend-ai-user-dropdown-menu.ts
```diff
@@ -586,76 +402,15 @@ export default class BackendAiUserDropdownMenu extends LitElement {
...
-            @click="${this._updateUserInformation}"></mwc-button>
-      </div>
-    </backend-ai-dialog>
+    <backend-ai-react-user-profile-dialog
+      value="${JSON.stringify({isOpen: this.isOpen})}"
+      @cancel="${()=> this._hideUserPrefDialog()}"
+      @refresh="${()=> this._refreshUserInfoPanel()}"
+      @updateFullName="${(e)=> this._updateUserFullName(e.detail.newFullName)}"
+      @confirmRemovingTotp="${(e)=> this._confirmRemovingTotp(e)}"
+      @startActivatingTotp="${(e)=> this._startActivatingTotp(e)}"
```
agatha197: Are these props necessary?

Sujin-Kim1 (changes requested):

react/src/components/UserProfileSettingModal.tsx
gee05053: @Sujin-Kim1 
oh.. I will fix right now

src/components/backend-ai-user-dropdown-menu.ts
gee05053: @Sujin-Kim1 
_updateFullName need props, but other function is not necessary props. So I removed props original function and comment line

src/components/backend-ai-user-dropdown-menu.ts
```diff
@@ -586,76 +402,15 @@ export default class BackendAiUserDropdownMenu extends LitElement {
...
-            @click="${this._updateUserInformation}"></mwc-button>
-      </div>
-    </backend-ai-dialog>
+    <backend-ai-react-user-profile-dialog
+      value="${JSON.stringify({isOpen: this.isOpen})}"
+      @cancel="${()=> this._hideUserPrefDialog()}"
+      @refresh="${()=> this._refreshUserInfoPanel()}"
+      @updateFullName="${(e)=> this._updateUserFullName(e.detail.newFullName)}"
+      @confirmRemovingTotp="${()=> this._confirmRemovingTotp()}"
+      @startActivatingTotp="${()=> this._startActivatingTotp()}"
```
agatha197: I think it's better to handle them in the `UserProfileSettingModal`.

Sujin-Kim1 (changes requested):

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,313 @@
+import React from "react";
+import {
+  Modal,
+  Input,
+  Form,
+  Select,
+  SelectProps,
+  Divider,
```
agatha197: Please remove unused import :)

Sujin-Kim1 (changes requested):

src/components/backend-ai-user-dropdown-menu.ts
gee05053: @Sujin-Kim1 
I removed @refresh props according to your review, because that was unnecessary props.
But other props necessary!
@updateFullName props is need to change user full name in header after mutation
@confirmRemovingTotp and @startActivatingTotp is need to open other modal and close user profile modal.

gee05053: @Sujin-Kim1 
I tried to use "current-user-info-changed", but the full name side by dropdown button was not changed after user changed info.
I think must to use @updateFullName or @refresh😔

src/components/backend-ai-user-dropdown-menu.ts
```diff
@@ -192,19 +192,26 @@ export default class BackendAiUserDropdownMenu extends LitElement {
     this.keyPairInfo.keypairs.reverse();
   }
 
+  async _setIsOpen(openState: boolean) {
+    this.isOpen = openState;
+  }
```
yomybaby: Let's change the name `isOpen` to more explicit. For example `isOpenUserPrefDialog`.
Also it's better to remove `_setIsOpen` and chage the property directly in `_openUserPrefDialog`  and `_hideUserPrefDialog`

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,304 @@
...
+            originalPassword: "",
+            newPassword: "",
+            newPasswordConfirm: "",
+          });
+        },
+      }
+    );
+  };
+
+  const _onSubmit = () => {
```
yomybaby: Please remove the `_`` prefix. This component is a function component. All functions defined inside of a function are not accessible outside of the function. Therefore, it is not necessary to prefix functions with an underscore in this case.

However, it may still be needed in some cases. For example, when the component has onSubmit props and you want to override onSubmit inside of the component.

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,304 @@
...
+  const [form] = Form.useForm();
+
+  const [messageApi, contextHolder] = message.useMessage();
+  const _showSuccessMessage = (successMessage: string) => {
+    messageApi.open({
+      type: "success",
+      content: successMessage,
+    });
+  };
+  const _showErrorMessage = (errorMessage: string) => {
```
yomybaby: Please use `messageAPI.open` directly where it is needed. You don't need to define it as a separate function when the code is used more than once. It's better when the code inside the function is simple and clear.

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,304 @@
...
+    messageApi.open({
+      type: "error",
+      content: errorMessage,
+    });
+  };
+
+  const baiClient = useSuspendedBackendaiClient();
+  let email = baiClient.email;
+  let full_name = baiClient.full_name;
+  let loggedAccount = baiClient._config.accessKey;
```
yomybaby: You don't need to defined every variable to use this component. You can use it directly where it neeeds. It's better to increae readability.

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,304 @@
...
+  let totpSupported = false;
```
yomybaby: Most of time, you don't need to `let`. 
```js
const totpSupported = baiClient.supports("2FA") && isManagerSupportingTOTP;
```

yomybaby (changes requested):

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -0,0 +1,315 @@
...
+        baiClient.email,
+        ["access_key", "secret_key"],
+        true
+      );
+    },
+    {
+      suspense: false,
+    }
+  );
+  let selectOptions: SelectProps["options"] = [];
```
yomybaby: `selectOptions` is initialized as an empty array using the `let`. However, since `selectOptions` is not reassigned a new value anywhere in the code, it can be safely declared as a constant using the `const` instead of `let`. `selectOptions.push` is not a reassigning.

yomybaby (changes requested):

react/src/components/UserProfileSettingModal.tsx
gee05053: @yomybaby 
oh.. I didn't know that..😭
I modified it

yomybaby:
@gee05053 
The UI that you implemented using React is good. However, the code to send a request should be located in the React modal. Let's separate the issues.
1. Please revert backend-ai-user-dropdown-menu and I'll merge it.
2. In your other PR that will replace "backend-ai-user-dropdown-menu", let's also replace the modal.

resources/i18n/ko.json
```diff
@@ -26,7 +26,7 @@
       "MyAccountInformation": "사용자 정보 변경",
       "MyAccount": "사용자 정보 변경",
       "FullName": "사용자 이름",
-      "FullNameInvalid": "사용자 이름은 64자 이하로 입력하십시오",
+      "FullNameInvalid": "사용자 이름은 0자 이상 64자 이하로 입력하십시오",
```
yomybaby:
```suggestion
      "FullNameInvalid": "사용자 이름은 64자 이하로 입력하십시오",
```
"0자 이상" is not necessary because this field is required.

yomybaby (changes requested):

PR #1850, opened 2023-08-10
Refactor maintenance view

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui-ossca-2023/issues/19.

스크린샷 2023-08-09 오후 3 15 04
Before PR

스크린샷 2023-08-09 오후 3 15 16
After PR


react/src/pages/MaintenancePage.tsx
```diff
@@ -0,0 +1,123 @@
...
+							borderColor: "#e0e0e0",
+							fontWeight: 400,
+							padding: "15px",
+							justifyContent: "center",
+							display: "flex"
+						}}
+						onClick={()=> dispatchEvent("recalculate", null)}
+					>
+						{recalculating ? t("maintenance.Recalculating") : t("maintenance.RecalculateUsage")}
+					</Button>
```
yomybaby: ou don't need to make the React-based component look exactly the same as the previous component. The React-based component on the WebUI adopts the Ant Design System, so most of the time, you don't need to explicitly set font weight, color, size, etc.

You can change this part like this:
```jsx
<Card title={t("maintenance.Fix")}>
	<Typography.Title level={5} style={{ margin: 0 }}>
		{t("maintenance.MatchDatabase")}
	</Typography.Title>
	<Trans>{t("maintenance.DescMatchDatabase")}</Trans>
	<Button
		block
		disabled={recalculating}
		icon={<RedoOutlined />}
		onClick={() => dispatchEvent("recalculate", null)}
	>
		{recalculating
			? t("maintenance.Recalculating")
			: t("maintenance.RecalculateUsage")}
	</Button>
</Card>
```

react/src/pages/MaintenancePage.tsx
```diff
@@ -0,0 +1,123 @@
...
+				</Card>
+			</Col>
+			<Col>
+				<Card
+					className="ImageEnvironment"
```
yomybaby: Why did you set the `className`?

yomybaby (changes requested):
Thank you for this PR. Please check my comments.

react/src/pages/MaintenancePage.tsx
gee05053: Thank you. I will request review again after modiying according your comment.😊

react/src/pages/MaintenancePage.tsx
gee05053: Uh.. I just set the classname for easy identification..
when I modify above comment, I'll modify this too

gee05053:
@yomybaby

I modified like picture!
How about this??

스크린샷 2023-08-17 오후 11 00 05

react/src/pages/MaintenancePage.tsx
gee05053: @yomybaby
I modified code accroding your review! If you have time, please review one more time

yomybaby:
Please resolve conflict. @gee05053

yomybaby (changes requested):
Could you improve these?
- page content container default margin or padding ([ref](https://github.com/lablup/backend.ai-webui/blob/main/react/src/components/Information.tsx#L83))
- Please check the layout when screen is small. You can set the `xs` and `sm` props of `Col`. Please check the screenshot below. It doensn't have vertical gutter.
image

gee05053:
> 

@yomybaby 
Sorry.. I will fix and request reivew again

react/src/pages/MaintenancePage.tsx
```diff
@@ -0,0 +1,90 @@
...
+  };
+  try {
+    parsedValue = JSON.parse(value || "");
+  } catch (error) {
+    parsedValue = {
+      recalculating: false,
+      scanning: false,
+    };
+  }
+  const { recalculating, scanning } = parsedValue;
```
KSoonYo, line 24: How about checking with `typeof`?
@yomybaby  @gee05053 
```ts
type MaintenanceValue = {
     recalculating: boolean;
    scanning: boolean;
}

const MaintenancePage: React.FC = () => {

  const { value, dispatchEvent } = useWebComponentInfo();
  const { recalculating, scanning } = typeof value === MaintenanceValue ? JSON.parse(value) : {  recalculating: false, scanning: false }
  
 // ...
 // ...
```

KSoonYo: oops... I didn't realize the JSON.parse() has to be checked for an error and the default value should be set.
Sorry!

KSoonYo: But I hope to suggest you to separate the parsing process from the component, like `function parsingValue(value)`.
This can be helpful when someone understands the meaning!

PR #1852, opened 2023-08-10
update: fix '자원를' to '자원을' for smaller resource text

mnxmnz:
Fix typo on "Smaller Resource Then Image Requires" text

img


yomybaby (approved):
Thank you for your contribution.

PR #1853, opened 2023-08-11
refactor: modify client config constructor type check in backend.ai-client-esm.ts

KSoonYo:
Creating a ClientConfig instance requires arguments as a string and createFromEnv() is currently using the "?? operator" to set a default empty string ('').

I believe it would be more appropriate to use the "?: operator" to allow for undefined arguments. This is because ClientConfig already includes an "if-else" statement to check if a required key is undefined or null and dynamically create configurations.

If there's a better alternative, I would appreciate your feedback.

Thank you.


yomybaby (approved):
LGTM 👍

PR #1857, opened 2023-08-11
test: Flex component rendering test & snapshot

KSoonYo:
I'd like to resolve the issue: test codes of react components

Before starting the tasks, I hope to check if I am doing properly.

---

I found two test methods of rendering components: using either "@testing-library/react" or "react-test-renderer"

I think it would be ok that using "react-test-renderer" just for rendering a specific component because my main goal is to leave its snapshot.("react-test-renderer" has good utilities for snapshots)

On the other hand, when I have to make some codes for testing user's interaction or relay-test(graphql reqeust and response) in the component, I will choose "@testing-library/react" with other mocking libraries.

So, can I mix these ways for testing? Aren't there any side effects?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "lint": "npx eslint src -c ./.eslintrc.json --quiet --ext .js,.ts; exit 0",
     "lint-fix": "npx eslint src -c ./.eslintrc.json  --quiet --ext .js,.ts --fix;exit 0",
     "test": "npx testcafe chrome ./tests/*.js",
+    "test:react" : "npm test --prefix ./react test",
     "wsproxy": "node ./src/wsproxy/local_proxy.js",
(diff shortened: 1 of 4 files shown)


yomybaby:
If possible, please remove react-test-renderer and use @testing-library/react instead. Thank you.

yomybaby (changes requested):
If possible, please remove `react-test-renderer` and use `@testing-library/react` instead. Thank you.

KSoonYo:
> If possible, please remove `react-test-renderer` and use `@testing-library/react` instead. Thank you.

@yomybaby I removed `react-test-renderer` and change the library into  `@testing-library/react` in [91f6044](https://github.com/lablup/backend.ai-webui/pull/1857/commits/91f60448496a40250f85761ec7f53ff340a51b3a)

yomybaby (approved):
'It's a good example for the first snapshot test code for a React component.

PR #1859, opened 2023-08-14
feat: react-based image environment select display tags with search highlight

yomybaby:
### environment search on Select
image

### version search on Select
image


lizable (approved):
It works fine. LGTM.

PR #1862, opened 2023-08-15
feat: edu-applauncher opens app for a session

adrysn:
- feat: Pass token as body parameter in token login
- feat: launch app when session_id is given


Sujin-Kim1 (approved):
LGTM

Issue #1809, opened 2023-07-25
bug: Version label overlaps with '/' symbol in 'Start Model Serving'


yomybaby:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

23.09

### Describe the bug

When there is no inference image in the 'Start Model Serving' screen, there is an issue where the version label overlaps with the '/'.

<img width="449" alt="image" src="https://github.com/lablup/backend.ai-webui/assets/621215/df1ca997-0305-4302-99ce-555c832249e2">

### To Reproduce

Connect to the backend.ai instance that has no inference image:
- Click the "New Folder" button on the Data & Storage page.
- Choose the usage mode as "Model".
- Enter the desired folder name.
- Press the "Create" button.

After creating the folder, you will see the "Model Serving" button in the "Control" column.
- Click that button.

### Expected Behavior

- no overlab
-(optional) provide a message for empty

PR #1866, opened 2023-08-16
fix: version label overlapping

hamo-o:
This PR resolves [#1809](https://github.com/lablup/backend.ai-webui/issues/1809#issue-1820069285)

**Before PR**
image
image

**After PR**
image
image

The element `'Not Selected'` of the default `version` array is pushed to the `info` array of the `_getVersionInfo()` function and displayed in the `version` list.

I've made the wrong tags invisible (they still exist in `info` array), would it be better to fix them so that the function `_getVersionInfo()` doesn't include them in the `info` array at all?


yomybaby (approved):
LGTM!  Thank you for your contribution 👍

PR #1871, opened 2023-08-18
feature: add tab for pipeline dedicated folder list

yomybaby:
## With this PR:
image


yomybaby:
@lizable please review this. :)

lizable (approved):
LGTM!
@yomybaby 
Btw, we need to block "renaming" pipeline dedicated folder (folder type: data) since when deleting piepline dedicated folder after renaming it to other names, the pipeline deletion process is partially done, not completely. I'll create it as a new issue.

PR #1875, opened 2023-08-19
feat: Do not allow batch session startup command to have an empty string.

Sujin-Kim1:
By adding validation to the `lablup-codemirror`, it was prevented from moving on to the next step without entering the startup command in the batch session.
image


yomybaby (approved):
LGTM

PR #1877, opened 2023-08-21
style: apply custom header to Modal 

### LINKED ISSUE #1876: Add a general modal
### Main idea

Let's add a new general modal that has a custom title named `BAIModal`.
`BAIModal` should extend props of ant.d Modal, so that the usage should be made to be same as ant.d Modal.
But the design of the title part is the same as the existing original modal. (title color, close button, underline, margin, etc.)
And replace every `Modal` component with `BAIModal`.

```[tasklist]
- [x] Add a new general modal that has a custom title named `BAIModal`.
- [x] Replace every `Modal` component with `BAIModal`.
- [ ] https://github.com/lablup/backend.ai-webui/pull/1877
```

- Original modal
image

- Antd modal
image

### Alternative ideas

_No response_

### Anything else?

_No response_

Sujin-Kim1:
- style: center modal component

resolves #1876 

## Screenshot (example)
|Before|After|
|------|-------|
|image|image|

### DIFF
--- a/react/src/components/BAIModal.css
+++ b/react/src/components/BAIModal.css
@@ -1,3 +1,6 @@
+.ant-modal.bai-modal .ant-modal-content {
+  padding: var(--general-modal-content-padding, 0);
+}
 .ant-modal.bai-modal .ant-modal-body {
   padding: var(--general-modal-body-padding, 0 24px);
 }
--- a/react/src/components/BAIModal.css
+++ b/react/src/components/BAIModal.css
@@ -1,0 +1,26 @@
+.bai-modal .ant-modal-body {
+  padding: var(--general-modal-body-padding, 0 24px);
+}
+
+.bai-modal .ant-modal-footer {
+  padding: var(--general-modal-footer-padding, 0 20px 24px 20px);
+}
+
+.bai-modal .ant-modal-header {
+  border-bottom: 1px solid rgb(221, 221, 221);
+  border-width: 100%;
+  justify-content: space-between;
+  display: flex;
+  align-items: center;
+}
+
+.bai-modal .ant-modal-content .ant-modal-header,
+.bai-modal .ant-modal-content > button.ant-modal-close {
+  padding: var(--general-modal-header-padding, 10px 20px);
+  height: var(--general-modal-header-height, 69px);
+}
+
+.bai-modal .ant-modal-content > button.ant-modal-close {
+  /* center */
+  top: 0;
+}
--- a/react/src/components/BAIModal.css
+++ b/react/src/components/BAIModal.css
@@ -1,12 +1,12 @@
-.bai-modal .ant-modal-body {
+.ant-modal.bai-modal .ant-modal-body {
   padding: var(--general-modal-body-padding, 0 24px);
 }
 
-.bai-modal .ant-modal-footer {
+.ant-modal.bai-modal .ant-modal-footer {
   padding: var(--general-modal-footer-padding, 0 20px 24px 20px);
 }
 
-.bai-modal .ant-modal-header {
+.ant-modal.bai-modal .ant-modal-header {
   border-bottom: 1px solid rgb(221, 221, 221);
   border-width: 100%;
   justify-content: space-between;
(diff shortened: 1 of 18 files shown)


react/src/components/ModelServiceSettingModal.tsx
```diff
@@ -93,6 +93,7 @@ const ModelServiceSettingModal: React.FC<Props> = ({
         loading: mutationToUpdateService.isLoading,
       }}
       title={t("modelService.EditModelService")}
+      centered
```
yomybaby: Let's create a type of BAIModal and use it instead of adding centered to the ant.d Modal. It has centered as a default and its own CSS for overriding, which does not apply to others.

react/src/fix_antd.css
```diff
@@ -39,6 +38,40 @@
   animation: loadingCircle 1s infinite linear;
 }
 
+.ant-tabs {
```
yomybaby: `fix_antd.css` is for fixing this [issue](https://github.com/ant-design/ant-design-icons/issues/568) temporary. After this issue is resolved, this file will be removed.

Please do not add CSS for customization.

yomybaby (changes requested):

react/src/fix_antd.css
agatha197: 😲 Oops, okay I'll fix it.

Sujin-Kim1:
Updated
Image

react/src/components/BAIModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+      }}
+    >
+      <style>{rawBAIModalCss}</style>
+      <Modal
+        centered={modalProps.centered ?? true}
+        className="bai-modal"
+        {...modalProps}
+      >
+        {modalProps.children}
+      </Modal>
```
yomybaby:
```suggestion
      />
```
`{...modalProps}` is passs the children already. You don't need to set again.

react/src/components/ServingRouteErrorModal.tsx
```diff
@@ -22,7 +17,7 @@ const ServingRouteErrorModal: React.FC<Props> = ({
   onRequestClose,
   onCancel,
   inferenceSessionErrorFrgmt,
-  ...modalProps
+  ...props
```
yomybaby: How about changing this to "baiModalProps"? It has only BAIModalProps. Let's make it clear. 

If you agree, please change the other files as well.

react/src/components/BAIModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+export interface BAIModalProps extends ModalProps {}
+const BAIModal: React.FC<BAIModalProps> = ({ ...modalProps }) => {
+  return (
+    <ConfigProvider
+      theme={{
+        token: {
+          paddingContentHorizontalLG: 0,
+          paddingLG: 0,
+          paddingMD: 0,
+        },
```
yomybaby: It's not a good idea to change these tokens for the modal header. These tokens can be used for all other components under the children of the modal.

react/src/components/BAIModal.css
```diff
@@ -0,0 +1,26 @@
+.bai-modal .ant-modal-body {
```
yomybaby:
```suggestion
.ant-modal.bai-modal .ant-modal-body {
```

Please add .ant-modal to make it affect only this component modal which has both .ant-modal and .bai-modal.

react/src/components/BAIModal.tsx
yomybaby: For this purpose, it's better not to customize the theme if possible.

react/src/components/ServingRouteErrorModal.tsx
```diff
@@ -71,12 +66,13 @@ const ServingRouteErrorModal: React.FC<Props> = ({
           {t("button.Close")}
         </Button>,
       ]}
-      {...modalProps}
+      {...props}
```
yomybaby: I found a potential problem related to overriding props for modals. I'll make another PR to fix this.

yomybaby (changes requested):

react/src/components/BAIModal.tsx
agatha197: I couldn't see the children at first, so I added them, but now I can see them well now. Thank you for checking. I reflected it!

react/src/components/ServingRouteErrorModal.tsx
agatha197: reflected!

react/src/components/BAIModal.tsx
agatha197: Oh, I see it again today. resolved!

yomybaby (approved):
LGTM!

PR #1888, opened 2023-08-25
feature: add sorting filter in new column `created_at` in serving list

### LINKED ISSUE #1887: (SERVERSIDE UPDATE NEEDED) Let's add sorting filter and corresponding attribute
Related: [backend.ai#1509](https://github.com/lablup/backend.ai/pull/1509)

```[tasklist]
### Tasks
- [x] Add new column `created_at` for sorting in the list of serving page
- [x] Sort by the latest time of `created_at` in the list.
```

lizable:
This PR resolves #1887.

> ⚠️ NOTE: ~~pagination filter goes to `created_at` as a default.~~
> There's a bit changed in `package.json` and `package-lock.json`. You need to update installed node packages in `src/react` by `npm i` before reviewing the code and UI whether it works properly or not.
> We need to implement pagination right after implementing monitoring by access-control(e.g. admin user can see all of the services that user(s) have been created) as a future feature.

## Before
Screenshot 2023-08-25 at 8 12 50 PM

## After
Screenshot 2023-08-28 at 4 20 59 PM

### DIFF
--- a/react/src/hooks/reactPaginationQueryOptions.tsx
+++ b/react/src/hooks/reactPaginationQueryOptions.tsx
@@ -1,0 +1,177 @@
+import { offset_to_cursor } from "../helper";
+import { LazyLoadQueryOptions } from "../helper/types";
+import { SorterResult } from "antd/lib/table/interface";
+import _ from "lodash";
+import { useState } from "react";
+import {
+  fetchQuery,
+  GraphQLTaggedNode,
+  useRelayEnvironment,
+} from "react-relay";
+import {
+  ArrayParam,
+  NumberParam,
+  ObjectParam,
+  useQueryParams,
+} from "use-query-params";
+
+export type SorterInterface = Pick<SorterResult<any>, "field" | "order">;
+
+export const antdSorterResultToOrder = (
+  sorter: SorterInterface | SorterInterface[]
+) => {
+  const sorterArray = _.castArray(sorter).filter((s) => s.field);
+
+  return _.filter(
+    _.map(sorterArray, (s) =>
+      _.isNull(s.order)
+        ? undefined
+        : `${_.snakeCase(s.field as string).toUpperCase()}_${
+            s.order === "ascend" ? "ASC" : "DESC"
+          }`
+    )
+  );
+};
+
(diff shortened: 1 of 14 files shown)


react/src/hooks/reactPaginationQueryOptions.tsx
```diff
@@ -0,0 +1,177 @@
+import { offset_to_cursor } from "../helper";
```
yomybaby: I think it's better to remove this file from this PR because it cannot be used until backend.ai GraphQL supports the Relay pagination specification using a cursor.

react/src/pages/ServingListPage.tsx
```diff
@@ -330,6 +334,15 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
                     <EndpointStatusTag endpointFrgmt={row} />
                   ),
                 },
+                {
+                  title: t('modelService.CreatedAt'),
+                  dataIndex: 'created_at',
+                  render: (created_at) => {
+                    return _humanReadableTime(created_at);
```
yomybaby: `/react/package.json` has `dayjs` already. You can use it.

react/src/pages/ServingListPage.tsx
```diff
@@ -330,6 +334,15 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
...
+                  },
+                  defaultSortOrder: 'descend',
```
yomybaby, line 344: You can set it as ['descend', 'ascend', 'descend'] to prevent the sorter from reverting to its default status if you want. I think the default status of the sorter can be confusing.
```suggestion
                  defaultSortOrder: 'descend',
                  sortDirections: ['descend', 'ascend', 'descend'],
```

yomybaby (changes requested):

react/src/hooks/reactPaginationQueryOptions.tsx
lizable: I agree. I'll remove this!

Sujin-Kim1 (approved):
LGTM. Let's just remove unused codes.

PR #1901, opened 2023-08-30
feature: add token generation UI in routing

### LINKED ISSUE #1900: Let's provide token generation / list-up UI on routing page as a part of model-serving feature
### Main idea

On model-serving, the end user; who uses(e.g. sends "sentences" or "images" to get results from trained models) the deployed model may need a "token" to access the endpoint URI. This is crucial because when the service is not open to the public, the user cannot access the endpoint URI.

For now, we decided to use JWT as a token, so we need to achieve certain tasks below:

- [x] Provide list for generated token list in routing page (param: duration)
- [x] Provide token generation UI

### Alternative ideas

_No response_

### Anything else?

_No response_

lizable:
This PR resolves #1900.

### Before:
Screenshot 2023-08-30 at 6 37 09 PM

### After:
* Serving (Service list) Page
Screenshot 2023-09-04 at 8 26 43 PM

* Service Info (Routing) Page
Screenshot 2023-09-04 at 7 59 31 PM

* Token Generation Modal
Screenshot 2023-09-04 at 8 02 19 PM
Screenshot 2023-09-04 at 8 02 29 PM

## To reviewers
Since this is the major feature of `23.09` version, I need to break down a little bit of the testing procedure so that reviewers can confirm whether it's Okay to merge. Please remind that this is an essential/crucial feature that proves model-serving in Backend.AI works properly, rather than the final UI for the model-serving feature.

#### Prerequisites
1. Set Backend.AI Core as the latest commit of `main` branch. (You may need to migrate DB to the latest version.)
2. Make sure you can create a session properly.
3. You may need to install node-packages by using `npm i` command in both root directory and in `src/react` directory.

#### List of model-serving feature needs to be provided in this PR
- [ ] Check whether creating model type vfolder works properly in `Data & Storage` page
- [ ] Check whether Serving page listing up services properly.
- [ ] Check whether you can create(start) service* (I mean, we need more detailed setting feature in slider min-max, but let's just leave it as is. It's not a "crucial" thing in phase 1.)
- [ ] Check whether you can change the `Desired Session Count` by clicking `Gear` icon in service list, or inside the routing page where you can access via clicking `Endpoint Name` column in service list.
- [ ] Check whether you can terminate the services.
- [ ] Check whether you can access service info page(routing page) by clicking `Endpoint Name`
- [ ] Check whether breadcrumb UI works properly in service info page (which is located on top-left side of the page)
- [ ] Check whether every copy icon right next to the string (Endpoint Name, Service Endpoint, Image, Token) works as you expected.
- [ ] Check whether refresh icon in service info page(routing page) works properly.
- [ ] Check whether the generate token modal pops up properly and is configurable.
- [ ] Check the rows in Generated Tokens updated when you requested generating token
- [ ] Check the routes info changes when desired session count changes (Please remind that IT TAKES SEVERAL TIME.)
- [ ] (IMPORTANT) Check whether service that is NOT OPENED TO PUBLIC works fine by using `curl command`
   ```
   $ export API_TOKEN="<i>" // input without "<" and ">"
   $ curl -H "Content-Type: application/json" -X GET \
      -H "Authorization: BackendAI $API_TOKEN" \
      <service-endpoint> // input without "<" and ">"
  ```


adrysn (changes requested):
I've tested the features and here are some observations:

* While setting up a service, there's no option to select a GPU. This means we can't use GPU for serving.
  image
* fixed ~~I wasn't able to close the dialog that popped up when I clicked the "Modify" button on the service detail page.~~
  image
* Even though the service is showing as healthy, the service endpoint isn't displayed.
  image
* fixed ([link](https://github.com/lablup/backend.ai-webui/pull/1901#issuecomment-1705883732)) ~~When generating a new token, the default expiration time is set to the current time. This triggers the 'valid_until is older than now' error when users simply hit the "Generate" button.~~
  image
* I was unable to generate a token even after specifying a valid expiration time. This could either be a server issue or something I might have misconfigured.
  image

yomybaby (changes requested):
resovled: ~~Please remove unused variables. @lizable ~~
image

yomybaby (changes requested):
~~Sometimes, I can't copy the token by clicking. I'll look into this.~~
Ignore this comments. It's happened with Raycast app. It can not paste text in the command input when it is long.

yomybaby (approved):

yomybaby (changes requested):
It's minor. The shadow of fixed table is displayed on the header bar.
image

yomybaby:
I changed the initial exp date to 24 hours later to resolve @adrysn 's feedback 
image

agatha197 (changes requested):
Sometimes, I cannot create a new service. This was the case in most cases when the length was 3 or less, and when tested with a length of 4 to 5, sometimes it is created or sometimes it is not created.

agatha197 (changes requested):
If I set the desired session count to 0, the desired session count will not be changed. Let's set the minimum value to 1.
image

yomybaby:
resolved by [7b72eef](https://github.com/lablup/backend.ai-webui/pull/1901/commits/7b72eefc3ecbd54e27b758f0bcb0fa0b51a7b203) 
image

yomybaby:
FYI, I forgot to change my review status to `approve` before merging.

PR #1906, opened 2023-09-05
feat: better msg when coordinator connection timeout

agatha197:
resolves https://github.com/lablup/giftbox/issues/462


inureyes:
@agatha197 This PR makes 'adding...' message to 'failed...' without exception.

src/components/backend-ai-app-launcher.ts
```diff
@@ -785,7 +785,10 @@ export default class BackendAiAppLauncher extends BackendAIPage {
       uri = uri + '&is_inference=true';
     }
     uri += '&protocol=' + (servicePortInfo.protocol || 'tcp');
-    this.indicator.set(50, _text('session.launcher.AddingKernelToSocketQueue'));
+    this.indicator.set(
```
inureyes: This change makes every progress message to 'FailedToConnectCoordinator'.

inureyes (changes requested):

src/components/backend-ai-app-launcher.ts
agatha197: oops, yes you're right. I modified it to show that's message when only the request failed.

inureyes (approved):
LGTM

PR #1908, opened 2023-09-06
fix: add optional chaining to show app launcher

agatha197:
Fixes a bug where the app launcher does not appear after the session is created.


inureyes (approved):
LGTM

PR #1914, opened 2023-09-07
feat: add support for OpenID login

kyujin-cho:
This PR adds support for ability to sign in the cluster with OpenID SSO (Single Sign-On). The name of the realm (defaults to "OpenID") inside the login button can be customized by `general.ssoRealmName` config.

### DIFF
--- a/index.html
+++ b/index.html
@@ -30,7 +30,7 @@
       }
     };
     globalThis.packageVersion = "23.09.0-beta.1";
-    globalThis.buildVersion = "230906.190949";
+    globalThis.buildVersion = "230907.120939";
 </script>
   <!-- DO NOT CHANGE BELOW LINE -->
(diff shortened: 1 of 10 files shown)


index.html
```diff
@@ -30,7 +30,7 @@
       }
     };
     globalThis.packageVersion = "23.09.0-beta.1";
-    globalThis.buildVersion = "230906.190949";
+    globalThis.buildVersion = "230907.120939";
```
yomybaby: Is there any reason to include a buildVersion update?

yomybaby (changes requested):
How can I test it? If a backend for testing this PR exists, please let me know. 😁

index.html
kyujin-cho: Urgh, must be automatically generated when running `make compile` and forgot to remove it.

yomybaby (approved):

PR #1915, opened 2023-09-07
feat: add an auto selected volume in creating and cloning a data folder

agatha197:
- resolves https://github.com/lablup/giftbox/issues/502
- Add 'auto' vhost when creating or cloning a data folder. (Auto volume is added only when there are multiple hosts.)
- The auto-selected folder is the lowest usage folder.
- If the usage percentage of the vhost is empty, set auto-selected vhost to the default folder.

## Screenshots
- create `autoSelectedVhostFolder` folder with auto host.
image
- The auto-selected folder is set to the ceph which is the lowest usage.
image
- But after creating a new folder, the original host name (ceph) will be displayed on the storage list.
image
- Info description. The usage info will follow the original host's (ceph).
image

### DIFF
--- a/resources/storage_metadata.json
+++ b/resources/storage_metadata.json
@@ -84,6 +84,11 @@
         "SpectrumScale",
         "GPFS"
       ]
+    },
+    "auto": {
+      "name": "Auto",
+      "description": "Automatically select the host with the least usage among the accessible folders for users. This prevents the folders from being concentrated on a single host.",
+      "icon": "local.png"
     }
   }
 }
--- a/resources/storage_metadata.json
+++ b/resources/storage_metadata.json
@@ -87,7 +87,7 @@
     },
     "auto": {
       "name": "Auto",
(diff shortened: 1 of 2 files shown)


resources/storage_metadata.json
```diff
@@ -84,6 +84,11 @@
         "SpectrumScale",
         "GPFS"
       ]
+    },
+    "auto": {
+      "name": "Auto",
+      "description": "Automatically select the host with the least usage among the accessible folders for users. This prevents the folders from being concentrated on a single host.",
```
adrysn: Let's rewrite:
> Choose the host with the lowest usage from the available host volumes for users. This ensures that usage is evenly distributed across multiple storage volumes, avoiding over-concentration on any single volume.

adrysn (changes requested):
Let's specify which volume is automatically selected like the image below:
image

agatha197:
> Let's specify which volume is automatically selected like the image below: image

Thanks for the feedback. That's a great idea. I reflected as below.
image

adrysn (commented):
LGTM.

adrysn (approved):

PR #1917, opened 2023-09-08
feat: update legacy key(`ram`) from device metadata

rapsealk:
This PR updates a legacy key `ram` to `mem`, following the manager.

https://github.com/lablup/backend.ai/blob/dc6b85868fe5cb8b5aea5ba58c62d579fd38d101/src/ai/backend/manager/api/etcd.py#L36-L43


inureyes (approved):
LGTM

PR #1920, opened 2023-09-11
ci: GitHub Action to run jest on `/react`

yomybaby:
Setup Jest for CI using GitHub Actions.

---

### Summary
🚀🧪🐛

This pull request adds a GitHub Actions workflow to run tests and coverage on the React codebase and fixes a dependency issue with Node.js 16 and macOS. The workflow file is `.github/workflows/jest.yml` and the dependency file is `react/package-lock.json`.

> _`react` codebase_
> _tests and coverage added_
> _autumn of quality_

### Walkthrough
* Add a new workflow file for GitHub Actions to run tests and coverage on the React codebase ([link](https://github.com/lablup/backend.ai-webui/pull/1920/files?diff=unified&w=0#diff-d717659d5699a02ed8f8fe63002776a6697fd148b13640c5faad798a46f9e68dR1-R38))
* Fix a compatibility issue with Node.js 16 and macOS for the `fsevents` dependency ([link](https://github.com/lablup/backend.ai-webui/pull/1920/files?diff=unified&w=0#diff-8c82bfd96b743d68691d36a5ebe566d90a8d5862b7617fc8a2038d44b929d9e5R10))

PR #1925, opened 2023-09-12
fix: errors in test code for iSizeToSize

yomybaby:
The initial test code of iSizeToSize in https://github.com/lablup/backend.ai-webui/pull/1920 had errors. This PR fixes that.

### DIFF
--- a/react/src/helper/index.test.tsx
+++ b/react/src/helper/index.test.tsx
@@ -26,27 +26,27 @@ describe('iSizeToSize', () => {
     });
   });
 
-  it('should convert iSize to Size with targetSizeUnit of "k"', () => {
-    const sizeWithUnit = '1M';
+  it('should convert iSize to Size with targetSizeUnit of "k"(lower case)', () => {
+    const sizeWithUnit = '1m';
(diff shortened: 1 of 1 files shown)


react/src/helper/index.test.tsx
```diff
@@ -26,27 +26,27 @@ describe('iSizeToSize', () => {
     });
   });
 
-  it('should convert iSize to Size with targetSizeUnit of "k"', () => {
-    const sizeWithUnit = '1M';
+  it('should convert iSize to Size with targetSizeUnit of "k"(lower case)', () => {
+    const sizeWithUnit = '1m';
     const targetSizeUnit = 'k';
```
agatha197, line 31:
```suggestion
    const targetSizeUnit = 'K';
```

Isn't it `K`?

agatha197 (changes requested):

react/src/helper/index.test.tsx
agatha197: ah my mistake

agatha197 (approved):
LGTM!

PR #1926, opened 2023-09-12
feat: increase memory of ssh image to 1g

agatha197:
To avoid OOM, increase memory of ssh image to 1g.


yomybaby (approved):
LGTM

adrysn:
Let's merge this after we further investigate there are other ways to prevent OOM due to upload buffer memory fill.

inureyes:
This issue can be resolved via `sync` option to NFS mount.

PR #1930, opened 2023-09-14
ci: Manage actions version with dependabot

Yaminyam:
Manage the version of github actions with dependabot.
First, only actions officially provided by the github actions team are applied first.

In the case of other external actions, there may be unknown update history, so I think it is necessary to check.


inureyes (approved):
LGTM

PR #1932, opened 2023-09-15
update: node.js packages

inureyes:
This PR updates node.js packages:

## Core component packages
 @babel/core                            ^7.22.15  →       ^7.22.19
 @babel/parser                          ^7.22.15  →       ^7.22.16
 @babel/types                           ^7.22.15  →       ^7.22.19
 @material/web                     ^1.0.0-pre.16  →  ^1.0.0-pre.17
 @typescript-eslint/eslint-plugin         ^6.6.0  →         ^6.7.0
 @typescript-eslint/parser                ^6.6.0  →         ^6.7.0
 @vaadin/combo-box                       ^24.1.6  →        ^24.1.7
 @vaadin/date-time-picker                ^24.1.6  →        ^24.1.7
 @vaadin/grid                            ^24.1.6  →        ^24.1.7
 @vaadin/icons                           ^24.1.6  →        ^24.1.7
 @vaadin/item                            ^24.1.6  →        ^24.1.7
 @vaadin/progress-bar                    ^24.1.6  →        ^24.1.7
 @vaadin/select                          ^24.1.6  →        ^24.1.7
 @vaadin/text-field                      ^24.1.6  →        ^24.1.7
 @vaadin/tooltip                         ^24.1.6  →        ^24.1.7
 electron                                ^26.1.0  →        ^26.2.1
 eslint                                  ^8.48.0  →        ^8.49.0
 marked                                   ^8.0.0  →         ^9.0.0
 rollup                                  ^3.29.0  →        ^3.29.1
 vis-data                                 ^7.1.6  →         ^7.1.7
 vis-util                                 ^5.0.3  →         ^5.0.4
 ws                                      ^8.13.0  →        ^8.14.1

## React component packages
 @testing-library/user-event  ^14.4.3  →  ^14.5.0
 @types/node                  ^20.5.9  →  ^20.6.0
 antd                          ^5.8.6  →   ^5.9.1
 i18next                      ^23.4.9  →  ^23.5.1
 react-router-dom             ^6.15.0  →  ^6.16.0

PR #1933, opened 2023-09-16
fix: broken `Select` UI by using inline-flex (since ant.d 5.9.1)

yomybaby:
After updating Antd to version 5.9.1 via #1932 , the block type label of a Select UI has broken.

# Before
image

# After
image


inureyes (approved):
LGTM

PR #1936, opened 2023-09-20
fix: resolve resource allocation related errors in session launchers when there are multiple AI accelerators.

agatha197:
resolve https://github.com/lablup/giftbox/issues/521
- fix: typo error so that displaying the allocated IPU
image

- When the user changed the resource group, the custom allocation items were updated to indicate the accelerators supported by the resource group.
- Resolved the mismatch between the selected resource preset and custom allocation and the requested resources when there are multiple accelerators.
- style: fix the broken UI of resource gauge (IPU)

| Before | After |
|--------|--------|
| image | image | 

> **Note**
The required resources are different based on the container image. For example, `python-pytorch` image needs IPU, but `ngc-pytorch` image needs GPU. So when you want to create a session with `python-pytorch` image, use the `ipu` resource group to use AI accelerator.

### DIFF
--- a/src/components/backend-ai-resource-broker.ts
+++ b/src/components/backend-ai-resource-broker.ts
@@ -96,7 +96,6 @@ export default class BackendAiResourceBroker extends BackendAIPage {
   // Flags
   @property({ type: Boolean }) _default_language_updated = false;
   @property({ type: Boolean }) _default_version_updated = false;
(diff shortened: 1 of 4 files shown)


inureyes (approved):
LGTM

lizable (approved):
LGTM!

PR #1941, opened 2023-09-24
fix: signing only macOS packages with apple signing system on GitHub Action

inureyes:
This PR changes signing behavior to bypass signing Windows / Linux packages with macOS signing key.


inureyes:
Note: it will be backported after merging.

kyujin-cho (approved):

PR #1945, opened 2023-09-26
fix: specify possible i18n static values in comments

yomybaby:
i18next-parser doesn't run the code, so it can't interpolate values in these expressions.
 ```
 t(key)
t('key' + id)
t(`key${id}`)
```


inureyes (approved):
LGTM

PR #1946, opened 2023-09-26
add: Spanish, Finnish translation

inureyes:
This PR adds Spanish, Finnish translation and missing texts for other languages.


adrysn (approved):
LGTM

PR #1948, opened 2023-09-26
add: multilingual support for 14 more languages

inureyes:
This PR adds language selector for Brazilian, Chinese, Finnish, German, Greek, Italian, Japanese, Polish, Portuguese, Spanish, Turkish, Vietnamese with machine-translated resources.

PR #1952, opened 2023-09-27
feat: replace translations in language selector with translations in each country

agatha197:
resolves https://github.com/lablup/backend.ai-webui/issues/1949
follows #1948 
image


inureyes (approved):
LGTM

PR #1966, opened 2023-10-11
feature: hide sftp resource group

agatha197:
- hide sftp resource group from `ResourceGroupSelect` and `backend-ai-resource-broker`.
- user cannot check the sftp resource group from resource monitor and resource group selector of creating model service modal.
- `ResourceGroupSelect`: replace resource group list with graphql

### DIFF
--- a/react/src/components/ResourceGroupSelect.tsx
+++ b/react/src/components/ResourceGroupSelect.tsx
@@ -1,4 +1,6 @@
+import { useBaiSignedRequestWithPromise } from '../helper';
 import { useCurrentProjectValue, useUpdatableState } from '../hooks';
+import { useTanQuery } from '../hooks/reactQueryAlias';
 import { ResourceGroupSelectorQuery } from './__generated__/ResourceGroupSelectorQuery.graphql';
 import { Select, SelectProps } from 'antd';
 import graphql from 'babel-plugin-relay/macro';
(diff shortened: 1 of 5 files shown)


react/src/components/ResourceGroupSelect.tsx
```diff
@@ -20,40 +22,61 @@ const ResourceGroupSelector: React.FC<ResourceGroupSelectorProps> = ({
...
-      }) as Promise<{ scaling_groups: { name: string }[] }>;
+        url: `/folders/_/hosts`,
+      }) as Promise<{
+        allowed: string[];
+        default: string;
+        volume_info: object;
+      }>;
     },
     staleTime: 0,
   });
```
yomybaby: You can set a type for the `useTanQuery` data like this. It's easier to read.
```suggestion
  const { data } = useTanQuery<{
    allowed: string[];
    default: string;
    volume_info: object;
  }>({
    queryKey: ['FolderHostsQuery', key],
    queryFn: () => {
      return baiRequestWithPromise({
        method: 'GET',
        url: `/folders/_/hosts`,
      });
    },
    staleTime: 0,
  });
```

react/src/components/ResourceGroupSelect.tsx
```diff
@@ -20,40 +22,61 @@ const ResourceGroupSelector: React.FC<ResourceGroupSelectorProps> = ({
...
-  //   {
-  //     projectId: projectId || currentProject.id,
-  //   }
-  // );
+  let { scaling_groups_for_user_group: resourceGroups } =
+    useLazyLoadQuery<ResourceGroupSelectorQuery>(
+      graphql`
+        query ResourceGroupSelectorQuery($user_group: String!) {
+          scaling_groups_for_user_group(user_group: $user_group) {
+            name
```
yomybaby: Should we add `is_active` or not?"

yomybaby (changes requested):

yomybaby (approved):

PR #2002, opened 2023-10-28
fix: hide inference app from app launcher

kyujin-cho:
This PR updates app launcher's icon display logic to hide inference app being shown at the UI. App launcher is meant to display interactive apps only, so printing out inference apps here can make confusion to the user. To test this PR you have to create a model service session, move on to sessions tab and click app launcher icon.

- [X] Minimum required manager version
- [X] Specific setting for review (eg., KB link, endpoint or how to setup)


inureyes (approved):
LGTM

PR #2003, opened 2023-10-28
fix: limit number of characters of model service name

kyujin-cho:
This PR updates model service constraint regex and its corresponding internationalization string to prevent model service session from being silently failing to be created. Due to the forming logic of model service session's `name` field (`<Model Service Name>-<UUID>`), the service name has to be smaller than 27 characters.


inureyes (approved):
LGTM

Issue #2006, opened 2023-10-30
Need to make a fallback routine for upload session creation works properly


lizable:
We need to delete `sftp-upload` key check on conditional statement since it throws an error when manager doesn't have metadata for `sftp-upload` tagged system image.

Here's the code link:
https://github.com/lablup/backend.ai-webui/blob/3b78ec24382aa8565f0f4aa8d5230706acc48b40/src/components/backend-ai-storage-list.ts#L3817

PR #2007, opened 2023-10-30
fix: remove image naming checking when selecting sftp image

agatha197:
resolves https://github.com/lablup/backend.ai-webui/issues/2006
## What's changed
By removing unnecessary logic that checks whether the image name contains `sftp-upload` or not, I have modified the fallback image to be properly selected even if `systemSSHImage` is not specified.

 (image: any) =>
        **~~image['name'].toLowerCase().includes('sftp-upload') &&~~**
        image['installed'],

- [ ] Test 1
1.  Remove the `systemSSHImage` option.
2. Open the ssh/sftp image by clicking the ssh icon on the folder dialog.
3. Check upload session is created without error.

- [ ] Test 2
1. Set `systemSSHImage` option.
2. Launch upload session in the same way as above.
3. Check the requested session image is the same as the set image.

- [ ] Documentation


lizable (approved):
Thanks a lot for defining test conditions!
All tests passed.
LGTM!

PR #2010, opened 2023-11-02
feature: set password to `undefined` when no change is desired.

yomybaby:
After https://github.com/lablup/backend.ai/pull/1670 merged, you can test.

There are two changes
- Set the password to `undefined` when no change is desired.
  - image
- display the masked password if the password is set.  
  - image

How to test
- Create a registry with a username and password
  - [ ] You can see the `****` in the password column of the created registry.
- Click the "Edit" button in the registry. Check "비밀번호 변경" and leave the password field blank. Finally, click "Save".
  - [ ]  You can see the empty password column of the registry. 

- [ ] Documentation


inureyes (approved):
LGTM

PR #2011, opened 2023-11-03
bugfix: GraphQL variable type error

yomybaby:

- [ ] Documentation


react/src/components/UserSettingModal.tsx
```diff
@@ -233,7 +233,6 @@ const UserSettingModal: React.FC<Props> = ({
       confirmLoading={isInFlightCommitModifyUserSetting}
       {...baiModalProps}
     >
-      <div>{sudoSessionEnabledSupported ? 'true' : 'false'}</div>
```
inureyes, line 236: Please explain about this change.

inureyes (changes requested):
Please explain about deletion of `sudo session enable` option

inureyes (approved):
LGTM

PR #2012, opened 2023-11-03
fix: UI with multiple types of accelerators

adrysn:
Some fixes for UI with multiple types of accelerators. The changes are far from optimal solutions, but rather points out that they require changes. Please check if the updates are compatible with other environments.

**Broken accelerator allocation bar layout in the summary page**
image

**Broken resource monitor texts**
image

**Hide sFTP upload resource group from session launcher dialog**
image

**User's max session concurrency is displayed after 20 seconds**
image


inureyes (approved):
LGTM

PR #2013, opened 2023-11-06
update: stand-alone local proxy packaging

inureyes:
Add stand-alone binary for easy installation / runtime / setup

### How to build

> **Warning**
> You should run `make` without parallelization. Set `-j1` or `\make` if aliased.

```
npm i
\make mac_arm64
```
or
```
npm i
\make linux_arm64
```

In Linux, install `zip` first.

- [ ] Documentation


achimnol (approved):
Confirmed build in Linux arm64 and macOS arm 64.

PR #2016, opened 2023-11-07
Fix: hide `max_vfolder_count` setting according to the api version

agatha197:
follows https://github.com/lablup/backend.ai-webui/pull/2011

## Problem
Updating `max_vfolder_count` in keypair is allowed only up to API 23 version or lower.
But the `max_vfolder_count` setting UI is always shown on the resource policy modification dialog.

## What's changed
- If the core version is greater than 23, hide the `max_vfolder_count` setting UI in the Modify Resource Policy dialog.
- Update schema.

## How to check
1. Go to the Users - Resource Policies page.
2. Check the `Max. # of folder` setting UI is displayed when only the API version is 23 or lower.
3. 23.09 endpoint: http://10.82.230.94:8090

## Checklist
- [ ] Main branch(greater than 23): `Max. # of folder` UI disappears on the create and update resource policy dialog.
- [ ] 23 or lower: `Max. # of folder` UI appears.

- [ ] Documentation

### DIFF
--- a/react/data/schema.graphql
+++ b/react/data/schema.graphql
@@ -1,3 +1,5 @@
+# PR #1688 79ef241dff642618d97e8d087c2a4d36e19fb27b
+
 schema {
   query: Queries
   mutation: Mutations
--- a/react/data/schema.graphql
+++ b/react/data/schema.graphql
@@ -1,21 +1,37 @@
-# PR #1680 27ab15e5297862f0bf2717b6b0b98d99eb609e26
-
 schema {
   query: Queries
   mutation: Mutations
 }
 
-"""All available GraphQL queries."""
+"""
+All available GraphQL queries.
+"""
 type Queries {
   node(
-    """The ID of the object"""
+    """
+    The ID of the object
+    """
     id: ID!
   ): Node
   agent(agent_id: String!): Agent
-  agent_list(limit: Int!, offset: Int!, filter: String, order: String, scaling_group: String, status: String): AgentList
+  agent_list(
+    limit: Int!
+    offset: Int!
+    filter: String
+    order: String
+    scaling_group: String
+    status: String
+  ): AgentList
   agents(scaling_group: String, status: String): [Agent]
   agent_summary(agent_id: String!): AgentSummary
-  agent_summary_list(limit: Int!, offset: Int!, filter: String, order: String, scaling_group: String, status: String): AgentSummaryList
+  agent_summary_list(
+    limit: Int!
+    offset: Int!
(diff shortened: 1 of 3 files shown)


react/data/schema.graphql
```diff
@@ -1,21 +1,37 @@
-# PR #1680 27ab15e5297862f0bf2717b6b0b98d99eb609e26
-
 schema {
```
yomybaby, line 3: Please add the PR number and commit hash to the first line.

yomybaby: Except for this, everything else is LGTM.

agatha197: Done! Thanks :)

src/components/backend-ai-credential-view.ts
```diff
@@ -1475,14 +1475,25 @@ export default class BackendAICredentialView extends BackendAIPage {
...
-              <div class="vertical layout flex">
-                <mwc-textfield label="${_t(
-                  'credential.Max#',
-                )}" class="discrete" id="vfolder-count-limit" type="number" min="0" max="50"
-                    @change="${(e) =>
-                      this._validateResourceInput(e)}"></mwc-textfield>
-              </div>
+            <div
+              class="horizontal layout justified"
+              style=${
```
lizable, line 1489: During rendering period, global variable (globalThis.backendaiclient.###) is set to undefined, which triggers runtime errors.  Please add additional property or variable to check whether manager version supports `max-vfolder-count-in-user-resource-policy`.
You may refer similar cases in here:
- [`container-registry-gql` support checking property in backend-ai-environment-view](https://github.com/lablup/backend.ai-webui/blob/51078d876a568db640dd6d5af2c760304de2fbed/src/components/backend-ai-environment-view.ts#L115-L116C31)
- [`inference-workload`, `sftp-scaling-group` support checking property in backend-ai-sessiojn-view](https://github.com/lablup/backend.ai-webui/blob/51078d876a568db640dd6d5af2c760304de2fbed/src/components/backend-ai-session-view.ts#L187-L190)

lizable (changes requested):
I got runtime error when testing on upstream version (`24.03.0dev0`) of Backend.AI Core. Please refer to the comments below and fix it. I think it's easy to handle.

src/components/backend-ai-credential-view.ts
agatha197: Good point! Thank you. I reflected it.

lizable (approved):
LGTM.

PR #2019, opened 2023-11-08
fix: add optional chaining to call support

inureyes:
copilot:all

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-environment-view.ts
+++ b/src/components/backend-ai-environment-view.ts
@@ -113,7 +113,7 @@ export default class BackendAIEnvironmentView extends BackendAIPage {
         () => {
           this.is_superadmin = globalThis.backendaiclient.is_superadmin;
           this.isSupportContainerRegistryGraphQL =
-            globalThis.backendaiclient.supports('container-registry-gql');
+            globalThis.backendaiclient?.supports('container-registry-gql');
         },
         true,
       );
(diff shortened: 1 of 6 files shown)


src/components/backend-ai-environment-view.ts
```diff
@@ -113,15 +113,15 @@ export default class BackendAIEnvironmentView extends BackendAIPage {
...
+            globalThis.backendaiclient?.supports('container-registry-gql');
         },
         true,
       );
     } else {
       // already connected
       this.is_superadmin = globalThis.backendaiclient.is_superadmin;
       this.isSupportContainerRegistryGraphQL =
-        globalThis.backendaiclient.supports('container-registry-gql');
+        globalThis.backendaiclient?.supports('container-registry-gql');
```
yomybaby, line 124: In this line, `?` is not required because it's already connected.
```suggestion
        globalThis.backendaiclient.supports('container-registry-gql');
```

yomybaby (changes requested):

src/components/backend-ai-environment-view.ts
inureyes: @yomybaby Logically you are right. However it is converted without predefine check routing after bundling.

yomybaby (approved):
LGTM

PR #2021, opened 2023-11-09
feature: hide size of directory by option

### LINKED ISSUE #1988: Hide the size value when filetype is directory in folder explorer
### Main idea

Displaying the size of directory is dependent on host OS, Which may cause confusing situations. For example, it may show the same directory size regardless of the sum of files(including sub-directory) inside the directory. Therefore It's better not to show the size of the file if the file type is directory. Just displaying with -(hyphen) will do.

### Alternative ideas

_No response_

### Anything else?

_No response_

lizable:
This PR resolves #1988 

- [ ] Documentation

#### Prerequisites
-  Backend.AI Core 23.09 or later version(just try the latest commit of main branch would be fine.)

- [ ] Check you can see the size of directory in vfolder explorer when `isDirectorySizeVisible` key doesn't exist in your config file.
- [ ] Check you can see `-` instead of the size of directory in vfolder explorer when `isDirectorySizeVisible` is set in your config file and set to false.

### Screenshot(s)
| After | Before  |
|------|-------|
| Screenshot 2023-11-09 at 7 26 22 PM | Screenshot 2023-11-09 at 7 28 27 PM |


src/components/backend-ai-storage-list.ts
```diff
@@ -2460,6 +2519,8 @@ export default class BackendAiStorageList extends BackendAIPage {
             globalThis.backendaiclient._config.maxFileUploadSize;
           this.directoryBasedUsage =
             globalThis.backendaiclient._config.directoryBasedUsage;
+          this._isDirectorySizeVisible =
+            globalThis.backendaiclient._config.isDirectorySizeVisible;
```
yomybaby, line 2523: You have to add this code to `else` block (line number 2544).

yomybaby (changes requested):

src/components/backend-ai-storage-list.ts
yomybaby: Without this modification, there is a bug where the folder size is displayed. To reproduce this issue:

- Visit another page first and then click the `data & folder` menu
- Click on the folder name
- You can see the folder size

yomybaby (approved):
LGTM

PR #2024, opened 2023-11-14
feat: VFolderPermissionTag component

yomybaby:
This component can display vFolder permissions by using a permission string or GraphQL fragment.
image

- [ ] Documentation

### DIFF
--- a/react/src/components/VFolderPermissionTag.tsx
+++ b/react/src/components/VFolderPermissionTag.tsx
@@ -1,0 +1,63 @@
+import DoubleTag, { DoubleTagObjectValue } from './DoubleTag';
+import {
+  VFolderPermissionTag_VFolder$data,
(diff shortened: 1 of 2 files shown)


react/src/components/VFolderPermissionTag.tsx
```diff
@@ -0,0 +1,60 @@
...
+  | {
+      vFolderFrgmt: VFolderPermissionTag_VFolder$key;
+      permission?: never;
+    };
+
+const VFolderPermissionTag: React.FC<VFolderPermissionTagProps> = ({
+  vFolderFrgmt = null,
+  permission,
+}) => {
+  const vFolder = useFragment(
```
rapsealk, line 32: To use this component in other proejct, such as FastTrack, it still have to depend on `VirtualFolder`.

rapsealk (approved):
LGTM.
image

Issue #2025, opened 2023-11-14
Neo Session Launcher - Beta


yomybaby:
Let's move forward to BETA! We need to complete the following tasks in the [alpha version](https://github.com/lablup/backend.ai-webui/pull/1953).

> [!IMPORTANT]
> If you add a task to the list below, please also leave your name so others know who created it.

```[tasklist]
### 💪 Tasks
- [ ] https://github.com/lablup/backend.ai-webui/pull/2026
- [x] `ResouceAllocationFormItems`: Consideration of the `alwaysEnqueueComputeSession` option https://github.com/lablup/backend.ai-webui/pull/2030
- [x] `ResouceAllocationFormItems`: improve UX related to min/max limit and remaining (min shmem 64 and image limit, etc.) https://github.com/lablup/backend.ai-webui/pull/2030
- [x] `ResouceAllocationFormItems`: setting values in accelerator value to zero(0) considering the image doesn't support AI accelerato, [comment by @lizable ](https://github.com/lablup/backend.ai-webui/pull/2026#discussion_r1394062070) https://github.com/lablup/backend.ai-webui/pull/2030
- [x] [Preset] Auto selection
- [x] [Preset] Config option to hide custom preset
- [ ] Session template & recent history (including handling not available resources)
- [x] "No mount vfolder" confirm modal
- [ ] Form item to change session owner (admin only)
- [ ] Session launcher modes: Inference and Import
- [x] Display remain time until batch scheduledDate
- [x] Display an appropriate error message when a session fails to start.
```

```[tasklist]
### 🏝️ Lower-priority tasks
- [ ] VFolder Table(In `Form.Item`) using GraphQL pagination, filter, order, etc. (filter and setting components)
- [ ] Save as a session template
```

```[tasklist]
### 🐞 Bugs
- [x] If a URL has a selected image, the image type does not change automatically based on the image limit. (reported by jongeun)
- [ ] https://github.com/lablup/backend.ai-webui/pull/2029
```

yomybaby:
Since 24.03, the Neo session launcher is set as the default launcher. Remaining tasks in the above list will be addressed in other issues.

PR #2026, opened 2023-11-14
feat: cluster mode in Neo Session launcher 

yomybaby:
ref [Task list ](https://github.com/lablup/backend.ai-webui/issues/2025)

- Cluster form items
  image

- Cluster & total allocations
  image

Please test cluster mode using Backend.AI server supports cluster mode such as `xxx.xx.64.15` test server(with cluster size 1 and 2).

- [ ] Documentation

### DIFF
--- a/react/src/components/ResourceAllocationFormItems.tsx
+++ b/react/src/components/ResourceAllocationFormItems.tsx
@@ -10,7 +10,7 @@ import ResourceGroupSelect from './ResourceGroupSelect';
 import { ACCELERATOR_UNIT_MAP } from './ResourceNumber';
 import ResourcePolicyCard from './ResourcePolicyCard';
 import ResourcePresetSelect from './ResourcePresetSelect';
-import SliderInputItem from './SliderInputFormItem';
+import SliderInputFormItem from './SliderInputFormItem';
 import {
   Card,
   Col,
@@ -531,7 +531,7 @@ const ResourceAllocationFormItems: React.FC<
               // getFieldValue('allocationPreset') === 'custom' && (
               <>
                 {resourceSlots?.cpu && (
-                  <SliderInputItem
+                  <SliderInputFormItem
                     name={['resource', 'cpu']}
                     // initialValue={0}
                     label={t('session.launcher.CPU')}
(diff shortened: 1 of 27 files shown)


react/src/components/ResourceAllocationFormItems.tsx
```diff
@@ -391,7 +379,7 @@ const ResourceAllocationFormItems: React.FC<
         if (nextImageSelectorType) {
```
lizable, line 379: I think we should add else block with setting values in accelerator value to zero(`0`) considering the image doesn't support AI accelerator. If not, the last page of the session launcher would display AI accelerator value, but when after creating session, the accelerator does not allocated to session, which may confuse user.

Here's the session that I tried to create using simple python kernel.

### Screenshot(s)
- The last page of session launcher (1 Warboy included)
Screenshot 2023-11-15 at 8 15 17 PM

- just created session from session launcher in session list
Screenshot 2023-11-15 at 8 15 44 PM

react/src/components/ResourceAllocationFormItems.tsx
```diff
@@ -776,77 +747,125 @@ const ResourceAllocationFormItems: React.FC<
...
-                            min: sliderMinMax[currentAcceleratorType]?.min || 0,
+                            min:
+                              sliderMinMaxLimit[currentAcceleratorType]?.min ||
+                              0,
                           },
                           remainingValidationRules[currentAcceleratorType],
                         ]}
-                      />
+                      >
+                        <InputNumberWithSlider
```
lizable, line 757: And I think we need to disable the AI accelerator input field when image does not support AI accelerator (e.g. simple python kernel).

lizable (changes requested):
I checked multi-node cluster session creation and It works fine, but there are a few things need to be fixed at ResourceAllocationFormItems component, especially in AI accelerator input field. I left comments below.

lizable:
> I checked multi-node cluster session creation and It works fine, but there are a few things need to be fixed at ResourceAllocationFormItems component, especially in AI accelerator input field. I left comments below.

This will be handled as a separate issue.

lizable (approved):
LGTM.

yomybaby:
Thank you for reviewing. 💪 
The accelerator value and disabling issue that you mentioned have been added https://github.com/orgs/lablup/projects/24/views/16?filterQuery=-last-updated%3A7days+beta&pane=issue&itemId=44660836

PR #2028, opened 2023-11-15
feature: Response style for `ImageEnvironmentSelectFormItems`

yomybaby:
## Changes
- Display Tag on Select
- Add a gradient to the end of the Tags when width is small
  | Before | After |
  |--------|--------|
  | image | image |
  | Cell | Cell | 
- Add opacity to image and tag when `Select` is opened (check a PyTorch icon below)

  | Before | After |
  |--------|--------|
  | image | image |

- [ ] Documentation


lizable (approved):
LGTM!

PR #2030, opened 2023-11-17
feature: remaining marks in environment allocation slider

yomybaby:
Ref https://github.com/lablup/backend.ai-webui/issues/2025

This PR has
- change the layout of sliders and input 
- remove `shmem` slider
  image
- add marks for the remaining resource amount
  image
- change remaining warning to error when alwaysEnqueueComputeSession is false
- fix: accelerator size is missing on request ([ref](https://github.com/lablup/backend.ai-webui/pull/2030/files#diff-cb5aa3700f469823d540634a1acf2796ffcfa96416d6fa9333bfab0ff658b2c1R354-R359))

  | `alwaysEnqueueComputeSession` is true | `alwaysEnqueueComputeSession` is false |
  |--------|--------|
  |  image | image | 

- [ ] Documentation


lizable (changes requested):
Works fine but I think there's some value missing (`size`) in translation.

### Screenshot(s)
Screenshot 2023-12-05 at 12 07 16 PM

lizable (approved):
### Updates
- Added Automatic Shared memory switch
   Screenshot 2023-12-08 at 4 55 58 PM
- Adjust the amount of shared memory according to memory; set to 1 GiB if amount of memory to allocate is set to 4GiB or more, and shared memory is smaller than 1GiB

I also checked neo session launcher works fine on both conditions (when `AlwaysEnqueueSession` value is true/false).
It works fine. LGTM

PR #2020, opened 2023-11-08
feat: Link FastTrack button to frontend server

rapsealk:
This PR is a counterpart to lablup/backend.ai#1692

This PR updates the `FastTrack` button to use `pipeline.frontendEndpoint`(frontend server) rather than `pipeline.endpoint`(API server) to redirect users to the right service.

Required manager version: >= 23.09.**6**

**Checklist:** (if applicable)

- [x] Mention to the original issue
- [ ] Documentation
- [x] Minium required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [ ] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

Issue #1938, opened 2023-09-21
Let’s add inactivelist


agatha197:
### Main idea

Currently, the `blocklist` option is disabled in the sidebar menu and is supposed to be taken to the error page when accessing the page in url. However, some people may want to hide the options on the blocklist from the sidebar menu altogether. So, let's add an inactivelist so that pages belonging to the blocklist are completely hidden in the sidebar menu, and pages belonging to the inactivelist are disabled like the existing blocklist.

```[tasklist]
### Tasks
- [ ] add inactivelist
- [ ] hide blocklist page in the blocklist
- [ ] disable inactive page in the inactivelist
- [ ] both of them, it is disallowed to access by url.
```

PR #2036, opened 2023-11-21
fix: Use `blockedMenuItem` to hide and show pipeline button

### LINKED ISSUE #2035: Remove hideSideMenuButton property and related statement to keep integrity of menu visibility.
### Main idea

Since we've been using "blocklist" in configuration to set visibility of menu in sidebar, I think it's better to remove `hideSideMenuButton` property, which only works for FastTrack link in the menu.

### Alternative ideas

_No response_

### Anything else?

_No response_

rapsealk:
This PR is a follow-up of #2020 and resolves #2035 by using `blockedMenuItem` to manage various menu buttons in a single point.

NOTE: Add `pipeline` to `[ui].menu_blocklist` from `webserver.conf` to hide, or it will be shown otherwise.

Refs
- #1938

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-webui.ts
+++ b/src/components/backend-ai-webui.ts
@@ -1348,26 +1348,24 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
                   <span class="bold">Backend</span>
                   .AI
                 </div>
-                ${
-                  this.siteDescription
-                    ? html`
-                        <div
-                          class="site-name"
-                          style="font-size:13px;text-align:left;"
-                        >
-                          ${this.siteDescription}
-                        </div>
-                      `
-                    : html``
-                }
+                ${this.siteDescription
+                  ? html`
+                      <div
+                        class="site-name"
+                        style="font-size:13px;text-align:left;"
+                      >
+                        ${this.siteDescription}
+                      </div>
+                    `
+                  : html``}
               </div>
               <span class="flex"></span>
             </div>
           </div>
           <div
-            class="${
-              this.mini_ui ? 'vertical' : 'horizontal'
-            } center-justified center layout flex"
+            class="${this.mini_ui
+              ? 'vertical'
+              : 'horizontal'} center-justified center layout flex"
           >
             <mwc-icon-button
               id="mini-ui-toggle-button"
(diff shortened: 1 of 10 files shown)


lizable:
Let's add `hideList` as a new config setting and rename `blockList` to `disableList`. And It's better to apply `hideList` option to all menu in sidebar.

src/components/backend-ai-webui.ts
```diff
@@ -451,6 +449,8 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
       this.blockedMenuItem = config.menu.blocklist
         .split(',')
         .map((x: string) => x.trim());
+      // Remove summary from accidentally added in config.menu.blocklist
+      this.blockedMenuItem.filter((menuTitle) => menuTitle !== 'summary');
```
lizable, line 453: Since fallback page is `summary`, we cannot block summary menu.

src/components/backend-ai-webui.ts
```diff
@@ -1452,24 +1440,17 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
...
-              ? html`
+            <!--<mwc-list-item graphic="icon" ?selected="${this._page ===
+            'session'}" @click="${() =>
+              this._moveTo(
+                '/session',
+              )}" ?disabled="${this.inactiveMenuItem.includes('session')}">
+              <i class="fas fa-list-alt" slot="graphic" id="sessions-menu-icon"></i>
+              <span class="full-menu">${_t('webui.menu.Sessions')} new</span>
+            </mwc-list-item>-->
+            ${!this.supportServing || this.blockedMenuItem.includes('serving')
```
lizable, line 1451: I also set the visibility of serving page not only dependent to Core version supports but also checking whether `serving` menu is included to blocklist.

lizable (approved):
LGTM.
I also touched up the visibility of menu according to blocklist, except for `summary` menu. When user tries to access a wrong page(hidden page),[ there's a fallback page which provides "go back to summary"](https://github.com/lablup/backend.ai-webui/blob/3cd9bd3639577db480fe1adadc10cbc5d63cfca0/src/components/backend-ai-error-view.ts#L96). Therefore I added simple filter to exclude `summary` page from blocklist in case of input error by user.

Issue #2032, opened 2023-11-20
fix: when user try to signup, errors is shown


gee05053:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

24.03.0a2

### Describe the bug

if user try to signup, errors is occurred.

![스크린샷 2023-11-20 오후 5 06 38](https://github.com/lablup/backend.ai-webui/assets/28584221/caf54f00-ac58-4287-9565-cadf7325f5aa)

this error is oeccurred because condition is not specified when **signupSupport** and **allowSignupWithoutConfirmation** is set true at config.toml.

### To Reproduce

Add condition when **allowSignupWithoutConfirmation** at config.toml is set true.

### Expected Behavior

user can siginup normally

PR #2037, opened 2023-11-22
Fix/user signup without confirmation

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui/issues/2032
- Add condition that token is not set when allowSignupWithoutConfirmation is set true
- Add EmailInput, passwordInput, passwordConfirmInput can't set during signup
  -> Previously, only userNameInput, tokenInput, signupInput is disabled
- Removed innerHTML
- userNameInput clear value after signup

- [ ] Documentation


gee05053:
This PR is cloed because user config is wrong.

Issue #2038, opened 2023-11-22
Feat: Expose endpoint in model serving list


gee05053:
### Main idea

<img width="999" alt="MicrosoftTeams-image" src="https://github.com/lablup/backend.ai-webui/assets/28584221/5056f51d-b3a9-4325-b440-5cdafbf8f7a0">

Add endpoint URL column beside of endpoint ID column
User can see endpoint URL more easily.

yomybaby:
Let's add `created_user` column only for the admin ccount

PR #2041, opened 2023-11-23
Feat: Expose endpoint and created user in model serving list

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui/issues/2038

User can see endpoint URL and admin can see created user.

Admin screen
스크린샷 2023-11-23 오후 2 57 13

User screen
스크린샷 2023-11-23 오후 2 57 45

- [ ] Documentation


react/src/pages/ServingListPage.tsx
```diff
@@ -146,6 +147,7 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
 
   const sortedEndpointList = _.sortBy(modelServiceList?.items, 'name');
 
+  console.log(baiClient);
```
rapsealk: Please remove any debugging logs prior to the code review.

gee05053: Sorry.. I missed that

react/src/pages/ServingListPage.tsx
```diff
@@ -360,9 +359,7 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
                   ? {
                       title: t('modelService.CreatedUser'),
                       dataIndex: 'created_user',
-                      render: (created_user) => (
-                        <EndpointCreatedUserText userId={created_user} />
-                      ),
+                      render: (created_user) => created_user,
                     }
                   : {},
```
yomybaby: When `is_admin` is `false, `{}` is added to the array. It's not a good idea even `Table` doesn't show it. 
One of the solution is
```js
onst array = [
  1,
  2,
  ...(is_admin ? [3] : []),
  4
];
```

yomybaby (changes requested):

react/src/pages/ServingListPage.tsx
gee05053: Oh.. I didn't know that!😳

resources/i18n/en.json
```diff
@@ -426,7 +426,8 @@
     "ServiceNameRule": "Only allow 4 to 24 characters of alphanumeric, underscore(_), hyphen(-), and period(.) and it must end with an alphanumeric character.",
     "FormValidationFailed": "Form validation failed",
     "resources": "Resources",
-    "Image": "Image"
+    "Image": "Image",
+    "CreatedUser": "Created User"
```
agatha197:
```suggestion
    "Owner": "Owner"
```

agatha197 (approved):
LGTM

gee05053:
This PR is closed because user config is wrong

PR #2044, opened 2023-11-23
fix: change default translated string to `''`

agatha197:
Set default translated string to `''` so that making easier to translate using Ally i18n extension.

- [ ] Documentation


gee05053 (approved):

agatha197:
Leave the implementation as it is now.

PR #2045, opened 2023-11-23
Feat:Cancelled session's log button is disabled

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui/issues/2022

CANCELLED session's log button is disabled
스크린샷 2023-11-23 오후 3 41 37

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -2928,6 +2928,7 @@ ${rowData.item[this.sessionNameField]}</pre
                   class="fg blue controls-running"
                   id="${rowData.index + '-assignment'}"
                   icon="assignment"
+                  ?disabled="${rowData.item.status == 'CANCELLED'}"
                   @click="${(e) => this._showLogs(e)}"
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-session-list.ts
```diff
@@ -2928,6 +2928,7 @@ ${rowData.item[this.sessionNameField]}</pre
                   class="fg blue controls-running"
                   id="${rowData.index + '-assignment'}"
                   icon="assignment"
+                  ?disabled="${rowData.item.status === 'CANCELLED'}"
```
lizable, line 2931: This is a minor thing, but I think we shall use triple equal (`===`) instead of `==`. Please refer [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#strict_equality_using).

lizable: And I just changed the operator to triple equal sign.

lizable (approved):
LGTM.

PR #2046, opened 2023-11-24
fix: user signup without confirmation

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui/issues/2032

- Add condition that token is not set when allowSignupWithoutConfirmation is set true
- Add EmailInput, passwordInput, passwordConfirmInput can't set during signup
  -> Previously, only userNameInput, tokenInput, signupInput is disabled
- Removed innerHTML
- userNameInput clear value after signup

- [ ] Documentation


src/components/backend-ai-signup.ts
```diff
@@ -308,7 +317,10 @@ export default class BackendAiSignup extends BackendAIPage {
     if (inputFieldsValidity.includes(false)) {
       return;
     }
-    const token = this.tokenInput.value;
+    let token;
+    if (!this.allowSignupWithoutConfirmation) {
+      token = this.tokenInput.value;
+    }
```
yomybaby: You don't need to do this because 335 line does something.  Please revert this part

yomybaby (changes requested):

src/components/backend-ai-signup.ts
gee05053: I added this code because token input is appered when allowSignupWithoutConfirmation is false like below code.
https://github.com/lablup/backend.ai-webui/blob/0c0e445222506b25f1b8c3dd11a264b8f6bf5e5f/src/components/backend-ai-signup.ts#L531-L543

Before code token value is always get input value.
But I think It is better to change to add token at body value when allowSignupWithoutConfirmation is false.

yomybaby (approved):
LGTM

PR #2047, opened 2023-11-24
feat: Expose endpoint and created user in model serving list

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui/issues/2038

User can see endpoint URL and admin can see created user.

Admin screen
스크린샷 2023-11-24 오전 9 35 37

User screean
스크린샷 2023-11-23 오후 2 57 45

- [ ] Documentation


react/src/pages/ServingListPage.tsx
```diff
@@ -288,6 +288,16 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
...
+                {
+                  title: t('modelService.ServiceEndpoint'),
+                  dataIndex: 'endpoint_id',
+                  render: (endpoint_id, row) =>
+                    row.url ? (
+                      <Typography.Text copyable>{row.url}</Typography.Text>
+                    ) : (
+                      <Tag>{t('modelService.NoServiceEndpoint')}</Tag>
+                    ),
+                },
```
yomybaby: `Tag` is not good for showing messages.  It's better to use `-`.
```suggestion
                    row.url ? (
                      <Typography.Text copyable>{row.url}</Typography.Text>
                    ) :  '-'
                },
```

react/src/pages/ServingListPage.tsx
```diff
@@ -288,6 +288,16 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
                     <Typography.Text code>{endpoint_id}</Typography.Text>
                   ),
                 },
+                {
+                  title: t('modelService.ServiceEndpoint'),
+                  dataIndex: 'endpoint_id',
+                  render: (endpoint_id, row) =>
+                    row.url ? (
+                      <Typography.Text copyable>{row.url}</Typography.Text>
```
yomybaby: It's better to make a link
```suggestion
                      <Typography.Link copyable href={row.url} target="_blank">
                        {row.url}
                      </Typography.Link>
```

yomybaby (changes requested):

react/src/pages/ServingListPage.tsx
gee05053: I think so too.🤔 I will request review again after resolving your suggestion

Issue #2033, opened 2023-11-20
Fix: When user name is blank, user can't change own profile


gee05053:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

24.03.0a2

### Describe the bug

When user try to change profile without name, error is occurred

![스크린샷 2023-11-20 오후 5 28 52](https://github.com/lablup/backend.ai-webui/assets/28584221/39e88a05-b143-4a1b-b475-55e128d71f38)

### To Reproduce

- We need to add rule when user try to change profile without user name

PR #2048, opened 2023-11-24
fix: remove `unhandled promise rejection` in UserProfileSettingModal 

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui/issues/2033

User can't change blank user name.

- [ ] Documentation

### DIFF
--- a/react/src/components/UserProfileSettingModal.tsx
+++ b/react/src/components/UserProfileSettingModal.tsx
@@ -166,7 +166,7 @@ const UserProfileSettingModal: React.FC<Props> = ({
           onRequestClose();
         }}
         confirmLoading={userInfo.isPendingMutation}
-        onOk={() => onSubmit()}
+        onOk={form.submit}
         centered
         destroyOnClose
         title={t('webui.menu.MyAccountInformation')}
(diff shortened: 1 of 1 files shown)


yomybaby (changes requested):
There are several ways to submit a form. In most cases, the React component in WebUI uses `form.validateFields`. It's easy to add extra handling for validation errors and send mutation requests if validation passes. The only downside is that it's easy to forget to catch. :)
Let's revert your changes and just add a `catch(()=>{})` to the end of `form.validateFields`.

react/src/components/UserProfileSettingModal.tsx
```diff
@@ -104,55 +104,58 @@ const UserProfileSettingModal: React.FC<Props> = ({
...
+          },
+          onError: (e) => {
+            messageApi.open({
+              type: 'error',
+              content: e.message,
+            });
+          },
+        });
+      })
+      .catch(() => {});
```
inureyes, line 162: Therefore, this is the only changed line. Is it correct? @gee05053

inureyes (approved):
LGTM now

PR #2052, opened 2023-11-26
fix: Correct hyphen usage in input validator regex

adrysn:
Place the hyphen (`-`) in the input element's regex pattern at the start of the character class to prevent it from being misinterpreted as a range indicator, which raises the following error in the browser's inspector console.
image


inureyes (approved):
LGTM

Issue #2050, opened 2023-11-25
Fix:The Credential View displays a list that does not match the sub-tab


gee05053:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

24.03.0a1

### Describe the bug

In the user menu (Credential View) that can only be accessed by administrators, when a user selects the sub-tab as inactive and then navigates to a different main-tab menu, and then returns to the main-tab menu where the sub-tab was selected again, the sub-tab remains selected as inactive, but the visible list shows active items.

In my opinion, it seems that when a main-tab is clicked, it always displays active items. Therefore, there may be a need to modify it so that the sub-tab is also selected

![스크린샷 2023-11-25 오후 2 39 05](https://github.com/lablup/backend.ai-webui/assets/28584221/627d7bfd-c18f-42ad-8112-e9d8c41e98e4)

PR #2055, opened 2023-11-27
fix:The Credential View displays a list that does not match the sub-tab

gee05053:
This PR resolves issue https://github.com/lablup/backend.ai-webui/issues/2050
Admin can see right active list-view whether admin move to other main menu.

- [ ] Documentation


lizable (approved):
LGTM.

PR #2064, opened 2023-11-29
feat: add model store tab to storage page

inureyes:

- [ ] Documentation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -95,7 +95,7 @@
         "concurrently": "^8.2.1",
         "cors": "^2.8.5",
         "crypto-es": "^2.0.4",
-        "electron": "^26.2.2",
+        "electron": "^26.2.4",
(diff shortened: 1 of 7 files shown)


react/src/components/ModelCloneModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+            name="target_host"
+            label={t('data.Host')}
+            tooltip={t('data.CloningIsOnlyPossibleSameHost')}
+            rules={[
+              {
+                required: true,
+              },
+            ]}
+          >
+            <StorageSelector value={sourceFolderHost} disabled />
```
yomybaby: You don't need to set value here again because there is an initialValue.
```suggestion
            <StorageSelector disabled />
```

yomybaby (changes requested):

yomybaby (approved):
LGTM.

PR #2066, opened 2023-11-30
fix: revive clone feature

lizable:
This PR revives "clone" feature which was commented out due to timing issue on large vfolder cloning ( > 10GB), which seems to be resolved already.

- [ ] Documentation

### Tests
- [ ] You can see Cloneable icon for each vfolder in Cloneable column, for each row. (check icon will show up only when the vfolder is cloneable.)
- [ ] You can change cloneable flag in "Update folder option", which can be triggered when you click "gear" icon in Control column.
- [ ] When you try to clone vfolder which has more than 10GB size for total, It do take time but eventually cloning operation will be finished succesfully.

### DIFF
--- a/src/components/backend-ai-data-view.ts
+++ b/src/components/backend-ai-data-view.ts
@@ -549,13 +549,15 @@ export default class BackendAIData extends BackendAIPage {
             : html``}
           ${this.enableStorageProxy
             ? html`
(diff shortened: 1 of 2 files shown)


lizable:
Since clone feature on backend part doesn't support cloning vfolders between different storages yet, therefore this PR will be on hold.

PR #2076, opened 2023-12-05
fix: rename to mode_card

yomybaby:
This PR 
- follows up on changes to the GraphQL Schema, which involve renaming from model_info to model_card.
- prevent sending a model metadata GraphQL query if the manager doesn't support the model store.

- [ ] Documentation


lizable (approved):
LGTM.

PR #2083, opened 2023-12-06
feature: Disable primary keypair deletion

lizable:
- related: [backend.ai#1761](https://github.com/lablup/backend.ai/pull/1761)
- required Core version: `24.03.0` or newer version
   - You need to upgrade alembic version to the latest by executing the following command in Core directory: 
      ```
       ./py -m alembic upgrade head
      ```

> FYI:
> User management in Backend.AI has two groups, user and keypair.
> The relationship btw user and keypair is 1:N relationship, but we regarded as 1:1 relationship.
> Both user and keypair has each status and works independently.

In this This PR, most of implementation is mainly focused on "main_access_key", considering the relationship between user and keypair (user can have multiple keypairs.)

- [ ] Prevent from deleting main access key by disabling "purge" button (red button in controls column at credentials table)
   - [ ] Even if user change the status of icon and requests purge, server returns error.
- [ ] Add `Main Access Key` tag in Access key column where access key is the main access key
- [ ] Add `Main Access Key` Column in users tab
- [ ] Show `Main Access Key` in User detail dialog

- [ ] Documentation

### Screenshot(s)
Screenshot 2023-12-06 at 6 42 18 PM
Screenshot 2023-12-06 at 6 42 23 PM
Screenshot 2023-12-06 at 6 42 31 PM

### DIFF
--- a/src/lib/backend.ai-client-esm.ts
+++ b/src/lib/backend.ai-client-esm.ts
@@ -4372,7 +4372,7 @@ class User {
       const limit = 100;
       const users = [] as any;
       if (!this.client.supports('main-access-key')) {
-        fields.splice(fields.indexOf('main_access_key'), 1);
+        fields = fields.filter(field => field !== 'main_access_key');
       }
       q = this.client.is_admin
         ? `
--- a/src/lib/backend.ai-client-esm.ts
+++ b/src/lib/backend.ai-client-esm.ts
@@ -669,6 +669,7 @@ class Client {
     if (this.isManagerVersionCompatibleWith('24.03.0')) {
       this._features['max-vfolder-count-in-user-resource-policy'] = true;
       this._features['model-store'] = true;
+      this._features['main-access-key'] = true;
     }
   }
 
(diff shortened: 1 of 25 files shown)


src/lib/backend.ai-client-esm.ts
```diff
@@ -4369,6 +4371,9 @@ class User {
       // we iterate pages to gather all users for client-side compability.
       const limit = 100;
       const users = [] as any;
+      if (!this.client.supports('main-access-key')) {
```
agatha197, line 4374:
```suggestion
      if (!this.client.supports('main-access-key') && 'main_access_key' in fields) {
```

agatha197 (changes requested):
I suggested just one additional condition. Other things are LGTM.

src/lib/backend.ai-client-esm.ts
```diff
@@ -4369,6 +4371,9 @@ class User {
...
+        fields = fields.filter(field => field !== 'main_access_key');
```
agatha197, line 4375:
```suggestion
        fields = fields.filter(field => !['main_access_key'].includes(field));
```

agatha197 (changes requested):
How about using `includes` to increase compatibility?

agatha197 (approved):
LGTM

PR #2086, opened 2023-12-07
fix: calculate liveStat by considering all containers

agatha197:
The liveStat of the current session only calculates the stat of the first container. It will miscalculate the CPU usage, etc. So I modified to calculate the liveStat considering all containers. And the calculation method followed the control panel.

ref: https://github.com/lablup/backend.ai-control-panel/blob/main/frontend/src/pages/hub-session-list.js#L842-L928

## How to test
- core: main
- Check the progress of usage column is same as control panel's session utilization.
  - If it's a 2-core allocation and only one core is 100% occupied, the progress should be 100% taken, but if you want to see the CPU usage for the entire session, you're only using 100% of the 200% available. So, it should be **50%** size

---

<Stress test>

1. Install package
   ```Python
   pip install stress
   ```
2. CPU test (compare with `top`)
   ```Python
    stress -c <number of core>
   ```
3. Mem test (compare with `top`)
   ```Python
    stress –m <size>
    # ex) stress -m 2048
   ```
4. GPU test (compare with `watch nvidia-smi`)
   > I tested with `library/colab-python:pytorch2.1` in http://10.100.64.15:8090
   ```Shell
	git clone https://github.com/wilicc/gpu-burn
	cd gpu-burn
	make
	./gpu_burn 60 # will run gpu_burn for 60 seconds
    ```
   

## Updated
For better testing, I updated utilization column UI like control panel.
image

- [ ] Documentation


lizable (approved):
I checked on both conditions (single-node single container, single-node multi container) and the CPU gauge in session usage seems to match with one in control panel.

lizable:
Confirmed gauge works as same as control panel.
Screenshot 2023-12-15 at 3 27 47 PM

Screenshot 2023-12-15 at 3 27 55 PM

PR #2087, opened 2023-12-07
feature: re-arrange my account dialog

lizable:
* related: #2083 

- required Core version: `24.03.0` or newer version
   - You need to upgrade alembic version to the latest by executing the following command in Core directory: 
      ```
       ./py -m alembic upgrade head
      ```

In this PR, keypair in user account information has been moved to user settings menu, since keypair is not maintained by user, but only for admin, which may confuses user if the fields is remaining in user account modal. 
Therefore, there are two things to check during this PR review:
- [x] Check access key and secret key is removed from user information(so called "My Account" modal)
- [x] Check the new menu "My keypair Info" in user settings menu
- [x] Check all the keypairs of current user in "My keypair info modal"
- [x] Check the main access key tag is attached correctly.

- [ ] Documentation

### Screenshot(s)
Screenshot 2023-12-07 at 6 41 56 PM
Screenshot 2023-12-07 at 6 41 41 PM
Screenshot 2023-12-07 at 7 36 14 PM

### DIFF
--- a/react/src/components/KeypairInfoModal.tsx
+++ b/react/src/components/KeypairInfoModal.tsx
@@ -6,10 +6,13 @@ import { useSuspendedBackendaiClient } from '../hooks';
 import { useCurrentUserInfo } from '../hooks/backendai';
 import { useTanQuery } from '../hooks/reactQueryAlias';
 import BAIModal, { BAIModalProps } from './BAIModal';
-import { Button, theme, Table, Typography } from 'antd';
-import _ from 'lodash';
+import Flex from './Flex';
+import { KeypairInfoModalQuery } from './__generated__/KeypairInfoModalQuery.graphql';
+import { Button, theme, Table, Typography, Tag } from 'antd';
+import graphql from 'babel-plugin-relay/macro';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { useLazyLoadQuery } from 'react-relay';
 
 interface KeypairInfoModalProps extends BAIModalProps {
   onRequestClose: () => void;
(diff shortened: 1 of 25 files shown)


react/src/components/KeypairInfoModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+        columns={[
+          {
+            title: '#',
+            fixed: 'left',
+            render: (id, record, index) => {
+              ++index;
+              return index;
+            },
+            showSorterTooltip: false,
+          },
```
agatha197:
```suggestion
            showSorterTooltip: false,
            rowScope: 'row',
          },
```

agatha197: control panel use `rowScope` to the index column. I added it to the style guide (KB).
ref: https://ant.design/components/table#components-table-demo-colspan-rowspan

react/src/components/KeypairInfoModal.tsx
```diff
@@ -0,0 +1,123 @@
...
+            dataIndex: 'access_key',
+            fixed: 'left',
+            render: (value) => (
+              <Flex direction="column" align="start">
+                <Typography.Text ellipsis copyable style={{ width: 150 }}>
+                  {value}
+                </Typography.Text>
+                {supportMainAccessKey && value === user?.main_access_key ? (
+                  <Tag color="red">{t('credential.MainAccessKey')}</Tag>
+                ) : null}
```
agatha197:
```suggestion
                {supportMainAccessKey && value === user?.main_access_key && (
                  <Tag color="red">{t('credential.MainAccessKey')}</Tag>
                )}
```

lizable:
Just updated few things:
- Refresh(fetching) keypair every time the modal opens.
- Enlarge Keypair info modal to show both access key and secret key clearly, with out ellipsis.

## Screenshot(s)
Screenshot 2023-12-08 at 12 38 48 PM

Issue #2084, opened 2023-12-07
Whitespace in welcome-message is missing.


Yaminyam:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

23.09.3

### Describe the bug

<img width="199" alt="CleanShot 2023-12-07 at 15 39 49@2x" src="https://github.com/lablup/backend.ai-webui/assets/31057849/4a7929f5-0fb6-4cde-ba55-66df948ee9d1">

It's awkward that there is no space after 'Welcome,'.
The actual code implementation includes spaces, but those spaces are omitted when output.
We need to figure out the cause.
https://github.com/lablup/backend.ai-webui/blob/afa8acaa56419121739aacf544ba51ebc2b5aedc/resources/i18n/en.json#L4

PR #2095, opened 2023-12-11
fix:Add white-space style

gee05053:
This PR resolves https://github.com/lablup/backend.ai-webui/issues/2084

Before
스크린샷 2023-12-11 오후 7 18 38

After
스크린샷 2023-12-11 오후 7 17 39

- [ ] Documentation

### DIFF
--- a/react/src/components/UserDropdownMenu.tsx
+++ b/react/src/components/UserDropdownMenu.tsx
@@ -148,17 +148,18 @@ const UserDropdownMenu: React.FC = () => {
       >
         <Flex direction="row" gap="sm" style={{ cursor: 'pointer' }}>
           {screens.md && (
-            <Typography.Text
-              strong
-              ellipsis
-              style={{
-                maxWidth: '15vw',
-                whiteSpace: 'nowrap',
-                overflow: 'hidden',
-              }}
-            >
-              {userInfo.username}
-            </Typography.Text>
+            <span style={{ maxWidth: '15vw' }}>
+              <Typography.Text
+                strong
+                ellipsis
+                style={{
+                  whiteSpace: 'nowrap',
+                  overflow: 'hidden',
+                }}
+              >
+                {userInfo.username}
+              </Typography.Text>
+            </span>
           )}
           <Avatar icon={<UserOutlined />} />
         </Flex>
--- a/react/src/components/UserDropdownMenu.tsx
+++ b/react/src/components/UserDropdownMenu.tsx
@@ -55,6 +55,10 @@ const UserDropdownMenu: React.FC = () => {
       style: {
         color: '#1f1f1f',
         cursor: 'default',
+        maxWidth: '15vw',
+        whiteSpace: 'nowrap',
+        overflow: 'hidden',
+        textOverflow: 'ellipsis',
       },
     },
     {
@@ -144,7 +148,17 @@ const UserDropdownMenu: React.FC = () => {
       >
         <Flex direction="row" gap="sm" style={{ cursor: 'pointer' }}>
           {screens.md && (
-            <Typography.Text strong>{userInfo.username}</Typography.Text>
+            <Typography.Text
+              strong
+              style={{
+                maxWidth: '15vw',
+                whiteSpace: 'nowrap',
+                overflow: 'hidden',
+                textOverflow: 'ellipsis',
+              }}
+            >
+              {userInfo.username}
+            </Typography.Text>
           )}
           <Avatar icon={<UserOutlined />} />
         </Flex>
--- a/react/src/components/UserDropdownMenu.tsx
+++ b/react/src/components/UserDropdownMenu.tsx
@@ -48,7 +48,7 @@ const UserDropdownMenu: React.FC = () => {
 
   const items: MenuProps['items'] = [
     {
-      label: userInfo.username,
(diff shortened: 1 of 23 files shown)


yomybaby (changes requested):
In this case, use interpolation
- lit-translate: https://github.com/stefanholzapfel/lit-translate#interpolation
- i18n-react: https://www.i18next.com/translation-function/interpolation

lizable (approved):
LGTM.

resources/i18n/de.json
```diff
@@ -1,7 +1,7 @@
 {
   "webui": {
     "menu": {
-      "WelcomeMessage": "Herzlich willkommen,",
+      "WelcomeMessage": "Willkommen, {{ userName }}",
       "WelcomeMessage_2": " .",
```
yomybaby: Is there any reason to keep the `WelcomeMessage_2`? You can put `WelcomeMessage_2` to the `WelcomeMessage` such as `.` and `님`.

yomybaby (changes requested):

resources/i18n/de.json
gee05053: Sorry. I deleted WelcomeMessage2 but I forgot commit that. I will commit right now

src/components/backend-ai-webui.ts
```diff
@@ -1862,9 +1862,9 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
                 </div>
                 <div slot="navigationIcon" class="vertical-line" style="height:20px;margin:0;"></div>
                 <div class="horizontal center layout" slot="title" id="welcome-message" style="font-size:12px;margin-left:10px;" class="draggable">
-                  <p>${_t('webui.menu.WelcomeMessage')}</p>
-                  <p class="user-name">${this._getUsername()}</p>
```
yomybaby, line 1866: please check  https://github.com/lablup/backend.ai-webui/pull/1008/commits/94bfc08edeaedca389560d4d4726f66cb9ddea80 commit related to `user-name` class.

yomybaby (changes requested):

src/components/backend-ai-webui.ts
```diff
@@ -1862,9 +1863,12 @@ export default class BackendAIWebUI extends connect(store)(LitElement) {
                 </div>
                 <div slot="navigationIcon" class="vertical-line" style="height:20px;margin:0;"></div>
                 <div class="horizontal center layout" slot="title" id="welcome-message" style="font-size:12px;margin-left:10px;" class="draggable">
-                  <p>${_t('webui.menu.WelcomeMessage', {
-                    userName: this._getUsername(),
-                  })}</p>
+                  <p style="display:flex;align-items:center;white-space:pre-wrap;">${_tr(
+                    'webui.menu.WelcomeMessage',
+                    {
+                      userName: `<p class="user-name">fdsfmsdklfnsklfnlksnfkslnflkknllnnkl</p>`,
```
yomybaby: Please check this part.
`fdsfmsdklfnsklfnlksnfkslnflkknllnnkl`

src/components/backend-ai-webui.ts
```diff
@@ -51,6 +51,7 @@ import {
   registerTranslateConfig,
   translate as _t,
   use as setLanguage,
+  translateUnsafeHTML as _tr,
```
yomybaby: Is there a lit-translate option to keep specific HTML like`transKeepBasicHtmlNodesFor` option of `react-i18next`?
Allowing all HTML is not safe.

yomybaby (changes requested):

src/components/backend-ai-webui.ts
gee05053: @yomybaby 
There is no option..
But I found that I can customize registerTranslateConfig()'s interplate option and I think that is work well.
I will apply and request review again

gee05053: @yomybaby 
I've searched for that feature, but lit-translate doesn't support it.
Therefore, I solved it using the method we discussed in the meeting

yomybaby (changes requested):
Please check again. It looks like this when the browser width is around 1100px.
image

gee05053:
@yomybaby 
I forgot to delete test style,,
I delete that and UserDropdownMenu component is applied user-name style too.

react/src/components/UserDropdownMenu.tsx
```diff
@@ -55,6 +55,10 @@ const UserDropdownMenu: React.FC = () => {
       style: {
         color: '#1f1f1f',
         cursor: 'default',
+        maxWidth: '15vw',
+        whiteSpace: 'nowrap',
+        overflow: 'hidden',
+        textOverflow: 'ellipsis',
```
yomybaby: How about applying this using the `ellipsis` prop of `Typhora.Text`?

yomybaby (changes requested):

react/src/components/UserDropdownMenu.tsx
yomybaby: If possible, setting style without explicit size(`15vw`) and color is better for maintenance.

yomybaby (approved):
I just push a commit. Please check this.
Thak you for contribution.
LGTM!

Issue #2089, opened 2023-12-08
Registry user config validation does not work properly.


agatha197:
### What Operating System(s) are you seeing this problem on?

Mac (Apple)

### Backend.AI version

1fcf5c9

### Describe the bug

Even if I changed the username or password of registry, I got confirm dialog that checks username or password is empty.
<img width="548" alt="image" src="https://github.com/lablup/backend.ai-webui/assets/28584164/50c6aedb-29f8-49b4-a99b-6a7d99e0f8ce">

### To Reproduce

update registry user info in the environments - registries.

### Expected Behavior

Show confirm dialog when only username and password is not changed.

PR #2099, opened 2023-12-13
fix:modiy condition of registry user config validation

gee05053:
This PR resolves https://github.com/lablup/backend.ai-webui/issues/2089

Modal.confirm is pop up when user isn't set username or check isChangePassword but user isn't write password

- [ ] Documentation


agatha197 (commented):
When I tried to modify registry, the registry info of the modify dialog did not match sometimes. Can you handle this issue too?

yomybaby:
We need to show the confirmation where the type is `habor`. I just pushed a commit.

yomybaby (approved):
LGTM

PR #2108, opened 2023-12-18
feature: `useResourceLimitAndRemaining` hook

yomybaby:
This PR:
 - introduces `useResourceLimitAndRemaining`, which includes bug fixes for calculating limit and remaining.
- makes `ResourceAllocationFormItems` uses `useResourceLimitAndRemaining`

- [ ] Documentation


react/src/components/ResourceAllocationFormItems.tsx
```diff
@@ -951,7 +738,10 @@ const ResourceAllocationFormItems: React.FC<
                             },
                           }}
                           min={0}
```
lizable, line 759: I think we need to apply disabled prop according to `currentImageAcceleratorLimits.length <= 0` since AI accelerator is dependent to image. (please refer [here](https://github.com/lablup/backend.ai-webui/pull/2108/files#diff-19c600075bfc029d497f5d6514affbea172d6e5d9c2a194a05a755241e72636fL111-L115))

react/src/components/ResourceAllocationFormItems.tsx
```diff
@@ -100,269 +89,51 @@ const ResourceAllocationFormItems: React.FC<
...
-        (keypairResourcePolicy.max_concurrent_sessions || 3) -
-        (keypair.concurrency_used || 0),
-    },
+  const sessionSliderLimitAndRemaining = {
+    min: 1,
+    max: _.min([
+      keypairResourcePolicy.max_concurrent_sessions,
+      3, //BackendAiResourceBroker.DEFAULT_CONCURRENT_SESSION_COUNT
+    ]) as number,
+    remaining:
```
lizable, line 120: I think sessionSliderLimitAndRemaining also need to consider remaining resources as well as `derivedClusterSizeMaxLimit`, such as following CPU cores. Since max_concurrent_session goes more than max value, then remaining pin will go far away from slider.

Here's an example:
> - Current max_concurrent_sessions of user's resource policy is set to **30**.
> - User just created one session that allocates 1 CPU Core.

The containing values of sessionSliderLimitAndRemaining goes to,
```
{ 
   min: 1,
   max: 3, // since max_concurrent_sessions is 30, so max will be set to 3
   remaining: 29, // since max_concurrent_session exists, therefore 30 - 1(concurrency_used) goes to 29
}
```

### Screenshot(s)
Screenshot 2023-12-18 at 11 50 49 PM
Screenshot 2023-12-18 at 11 50 53 PM

lizable (changes requested):
I found some minor changes, including remaining value of `InputNumberSlider` need to be fixed. Other seems working fine, whether `AlwaysEnqueueSession` is enabled or not.
Please refer to the comments above:

lizable (approved):
LGTM.

PR #2112, opened 2023-12-20
feature: Set resource limitation in service launcher

### LINKED ISSUE #1967: Need to set resource limitation (especially on max) at resource allocator in model-service launcher
### Main idea
related: [giftbox#552](https://github.com/lablup/giftbox/issues/552)

Currently, we don't have any limitations in the service launcher already applied in session launcher resource allocation.
It is crucial when there's no limit on resource allocation, it triggers continuous failure on service creation.

I think we could reuse [the resource allocation section in neo-session-launcher](https://github.com/lablup/backend.ai-webui/blob/main/react/src/components/ResourceAllocationFormItems.tsx) that reflects multi-layer resource limitation on the user.

### Alternative ideas

_No response_

### Anything else?

_No response_

lizable:
This PR resolves #1967 by replacing resource allocation form items to ResourceAllocationFormItems component.
After applying this changes, users can...
- [x] Notice the maximum value of resource according to intersect of the resource policy, resource group, and project
- [x] Set cluster size and cluster mode freely (⚠️ Though if corresponding resource group(and agents) doesn't support multi-node environment, request will fail. This issue need to be handled as a separate issue.)
- [x] Set the amount of shared memory to allocate in both automatic/manual way.

- [ ] Documentation

| After | Before |
|------|--------|
| Screenshot 2023-12-20 at 4 15 15 PM | Screenshot 2023-12-20 at 4 15 52 PM |

### DIFF
--- a/react/src/components/ResourceAllocationFormItems.tsx
+++ b/react/src/components/ResourceAllocationFormItems.tsx
@@ -906,7 +906,10 @@ const ResourceAllocationFormItems: React.FC<
                               baiClient._config?.always_enqueue_compute_session,
                             validator: async (rule: any, value: number) => {
                               if (
-                                sliderMinMaxLimit[currentAcceleratorType] &&
+                                _.isNumber(
+                                  sliderMinMaxLimit[currentAcceleratorType]
+                                    ?.remaining,
+                                ) &&
                                 value >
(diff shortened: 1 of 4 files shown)


yomybaby (approved):
LGTM

Issue #1951, opened 2023-09-27
Let's add validation to session name of the session launcher


agatha197:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

a42c0f1

### Describe the bug

If the user creates a new session with an invalid pattern session name, it causes an input validation error and cannot create a session.
Let's add validation to the session name of the session launcher and prohibit the user to move next progress if the user inputs an invalid session name.

### To Reproduce

Launch a new session launcher with special symbols like #,$,%, etc.

### Expected Behavior

Session name validation is shown and cannot proceed to the next progress.

### Anything else?

The session rename pattern: `^(?:[a-zA-Z0-9][a-zA-Z0-9._-]{2,}[a-zA-Z0-9])?$`
User can't move the next progress with an empty startup command if the session type is `batch`. 
![image](https://github.com/lablup/backend.ai-webui/assets/28584164/7f087733-082c-4ca4-ad68-e90d7045e3c9)

adrysn:
Another feedback from other customers.

PR #2123, opened 2024-01-02
feat: add session name validator to the session launcher

### LINKED ISSUE #2101: When creating a session, the name must be checked for validation

Linux (x86-64)

### Backend.AI version

main

### Describe the bug

image
When creating a session, the name must be set to 4 to 64 characters.
If this rule is not followed, session creation will fail.
However, in the session creation menu, even if you do not follow the above rules, you will proceed to the next step and an input validation error will occur.
I think it would be a good idea to verify this at the UI level as well so that it cannot proceed to the next step.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

agatha197:
resolve https://github.com/lablup/backend.ai-webui/issues/1951
- Fix the wrong pattern of session name field (session launcher) and session rename field (session list).
  - Pattern meaning (`^[a-zA-Z0-9]([a-zA-Z0-9\\-_\\.]{2,})[a-zA-Z0-9]$`)
  - `^[a-zA-Z0-9]`: The string starts with any alphanumeric character.
  - `([a-zA-Z0-9_.-]{2,62})`: Followed by any alphanumeric character or a dot . or a dash - , or underbar _ repeated between 2 and 62 times. This range satisfies the requirement that the entire string be at least 4 characters and a maximum of 64 characters.
  - `[a-zA-Z0-9]$`: The string ends with any alphanumeric.

- It allows alphanumeric and dash(-), underbar(_), dot(.) only.
  image

- It doesn't allow duplicated session name. 
  image
- If one of the validations of current progress of session launcher is failed, the user cannot move to the next progress. 

- [ ] Documentation
- [x] Minimum required manager version : 23.09
- [x] Specific setting for review (eg., KB link, endpoint or how to setup) : launch a new session, and check the session name. 


lizable (changes requested):
@agatha197 I tried to proceed to the next page of session launcher but It doesn't work. Could you check it please?
Here's the recording:
https://github.com/lablup/backend.ai-webui/assets/46954439/931d6f5a-91ee-4915-94ed-1d26d674e884

agatha197:
> @agatha197 I tried to proceed to the next page of session launcher but It doesn't work. Could you check it please? Here's the recording: [lablup/backend.ai-webui/assets/46954439/931d6f5a-91ee-4915-94ed-1d26d674e884](https://github.com/lablup/backend.ai-webui/assets/46954439/931d6f5a-91ee-4915-94ed-1d26d674e884)

Oops, it's my mistake. I resolved it! Thank you :)

yomybaby (approved):
LGTM!

PR #2138, opened 2024-01-10
fix: model service owner always represented as logged in user

kyujin-cho:
This PR resolves owner of model session always being shown as the email of current logged in user, by pulling out user information from the new GQL fields introduced at lablup/backend.ai#1831. Previous approach can cause confusion to the system admin, as they are possible to also take a look at the status of model service created by other users. 
Screenshot 2024-01-10 at 12 04 47 PM
Now this displays the true information correctly.
Screenshot 2024-01-10 at 12 02 01 PM
Whenever there's a difference between model service creator and actual model session owner, by hovering the cursor over question mark webui will show a tooltip mentioning about these facts. 
Testing this PR will require a Backend.AI server with lablup/backend.ai#1831 patch applied.

### DIFF
--- a/react/src/pages/RoutingListPage.tsx
+++ b/react/src/pages/RoutingListPage.tsx
@@ -126,8 +126,6 @@ const RoutingListPage: React.FC<RoutingListPageProps> = () => {
             resource_group
             resource_slots
             resource_opts
-            created_user_email
-            session_owner_email
             routings {
               routing_id
               session
--- a/react/src/pages/RoutingListPage.tsx
+++ b/react/src/pages/RoutingListPage.tsx
@@ -227,7 +227,13 @@ const RoutingListPage: React.FC<RoutingListPageProps> = () => {
           </Tooltip>
         </>
       );
-  }, [endpoint?.created_user_email, endpoint?.session_owner_email]);
+  }, [
+    endpoint?.created_user_email,
+    endpoint?.session_owner_email,
+    baiClient,
+    t,
+    token.colorTextSecondary,
+  ]);
 
   const resource_opts = JSON.parse(endpoint?.resource_opts || '{}');
   return (
--- a/react/src/pages/RoutingListPage.tsx
+++ b/react/src/pages/RoutingListPage.tsx
@@ -1,4 +1,5 @@
 import CopyableCodeText from '../components/CopyableCodeText';
+import EndpointOwnerInfo from '../components/EndpointOwnerInfo';
 import EndpointStatusTag from '../components/EndpointStatusTag';
 import EndpointTokenGenerationModal from '../components/EndpointTokenGenerationModal';
 import Flex from '../components/Flex';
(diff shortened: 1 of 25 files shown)


react/src/pages/RoutingListPage.tsx
```diff
@@ -201,6 +203,38 @@ const RoutingListPage: React.FC<RoutingListPageProps> = () => {
...
+        </>
+      );
+  }, [
+    endpoint?.created_user_email,
+    endpoint?.session_owner_email,
+    baiClient,
+    t,
+    token.colorTextSecondary,
+  ]);
+
```
yomybaby: Let's make this part as an individual component (eg. `<ServiceSessionOwner serviceFrgmt={service}/>`) using a GraphQL fragment. After that, we can reuse this component in [`ServingListPage.tsx`](https://github.com/lablup/backend.ai-webui/blob/cbc4bde074ed8f6562281af044a558eb0eddf0b7/react/src/pages/ServingListPage.tsx#L176-L185)

You can refer [`EndpointStatusTag.tsx`](https://github.com/lablup/backend.ai-webui/blob/cbc4bde074ed8f6562281af044a558eb0eddf0b7/react/src/components/EndpointStatusTag.tsx#L13C1-L20) which uses GraphQL fragment.

yomybaby (changes requested):

react/src/pages/RoutingListPage.tsx
yomybaby: Oops. I selected wrong line. The code what I want to mention is [this](https://github.com/lablup/backend.ai-webui/pull/2138/files/cbc4bde074ed8f6562281af044a558eb0eddf0b7#diff-b95c8773f0011bade5e876018f34889f34d54b495ccdabe90c2d81a7ca5334e8R206-R236)

react/src/pages/RoutingListPage.tsx
```diff
@@ -125,13 +126,16 @@ const RoutingListPage: React.FC<RoutingListPageProps> = () => {
             resource_group
             resource_slots
             resource_opts
+            created_user_email @since(version: "23.09.8")
+            session_owner_email @since(version: "23.09.8")
```
yomybaby: We can remove this because these fields aren't used in RoutingListPage. It is used EndpointOwenInfo.
```suggestion

```

react/src/pages/ServingListPage.tsx
```diff
@@ -280,7 +283,8 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
               url
               open_to_public
               created_at @since(version: "23.09.0")
-              created_user
+              created_user_email @since(version: "23.09.8")
+              session_owner_email @since(version: "23.09.8")
```
yomybaby: Same here: We can remove this because these fields aren't used in RoutingListPage. It is used EndpointOwenInfo.
```suggestion

```

yomybaby (changes requested):

yomybaby (approved):
LGTM! 🚀  Thank you!

PR #2152, opened 2024-01-19
fix: omit the mark if it is greater than the max

yomybaby:
### Bug
- The sliders of `InputNumberWithSilder` and `DynamicUnitInputNumberWithSlider` display the mark outside the slider when the mark value is greater than the max of the slider.
image

### How to reproduce:
Open the service launcher modal in a local dev environment, you can see this bug on the slider for cluster size.

- [ ] Documentation


agatha197 (approved):
LGTM

Issue #2165, opened 2024-01-23
bug: ResourceGroupSelect is out of resource monitor card


gee05053:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

24.03.0a1

### Describe the bug

When ResourceGroupSelect's value is too long, select is out of resource monitor card
![스크린샷 2024-01-23 오후 1 15 53](https://github.com/lablup/backend.ai-webui/assets/28584221/99e782ca-334d-4581-b4dd-08dca4c5f753)
![스크린샷 2024-01-23 오후 1 24 50](https://github.com/lablup/backend.ai-webui/assets/28584221/0aa012a0-f1d1-4982-bca1-02d2a54e54cd)

PR #2166, opened 2024-01-23
fix: ResourceGroupSelect is layout and support search highlight

gee05053:
This PR resolves https://github.com/lablup/backend.ai-webui/issues/2165

Now ResourceGroupSelect is set maxWidth and user can search in `Select`

Before
스크린샷 2024-01-23 오후 2 03 58
스크린샷 2024-01-23 오후 2 03 40

After
스크린샷 2024-01-29 오후 6 08 44
스크린샷 2024-01-29 오후 6 08 50
스크린샷 2024-01-29 오후 6 08 58
스크린샷 2024-01-29 오후 6 09 01

- [ ] Documentation


react/src/index.tsx
```diff
@@ -159,7 +159,11 @@ customElements.define(
 
     return (
       <DefaultProviders {...props}>
-        <Flex direction="column" align="stretch" style={{ minWidth: 200 }}>
+        <Flex
+          direction="column"
+          align="stretch"
+          style={{ minWidth: 200, maxWidth: 310 }}
```
yomybaby: You don't need to set the maxWidth. When you set an explicit number in your code, there might be a better way to do the same thing without a hardcoded number. @gee05053  @ironAiken2 

```suggestion
          style={{ minWidth: 200, flex:1 }}
```

react/src/components/ResourceGroupSelect.tsx
```diff
@@ -100,9 +102,19 @@ const ResourceGroupSelect: React.FC<ResourceGroupSelectProps> = ({
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [autoSelectDefault]);
+
+  const filteredResourceGroups = useMemo(() => {
+    return _.filter(resourceGroups, (resourceGroup) => {
+      if (!groupSearch) return true;
+      return resourceGroup.name.includes(groupSearch);
+    });
+  }, [groupSearch, resourceGroups]);
```
yomybaby: You can use `Select`'s filter feature instead of createing filtered data.

react/src/components/ResourceGroupSelect.tsx
```diff
@@ -21,6 +22,7 @@ const ResourceGroupSelect: React.FC<ResourceGroupSelectProps> = ({
   const baiRequestWithPromise = useBaiSignedRequestWithPromise();
   const currentProject = useCurrentProjectValue();
   const [key, checkUpdate] = useUpdatableState('first');
+  const [groupSearch, setGroupSearch] = useState('');
```
yomybaby: Can you use `useControllableValue` for this part? Your changes cause the parent componento to lose control over the `search`.

yomybaby (changes requested):

react/src/index.tsx
yomybaby: Oh! My above review was incorrect. The suggestion I provided doesn't work. The explicit width applies only to `backend-ai-react-resource-group-select`, which is used only in `backend-ai-resource-monitor`, so the current code is okay. @gee05053  @ironAiken2  I resolved this conversation.

gee05053: @yomybaby 
Yes😂 I tried yesterday to modify code according your suggestion, and I checked It was not work.
But I agreed your suggestion too about `there might be a better way without a hardcoded number`
I searched other way to solve this issue, and I found solution .
when I removed `wrap` in this [code](https://github.com/lablup/backend.ai-webui/blob/14c9bb563f928fd225c8950c811665f193524984/src/components/backend-ai-resource-monitor.ts#L661), this issue was resolved.
Is this way not good instead of using maxWidth?

gee05053: @yomybaby If you want to check code, please check recent code that I pushed😊

yomybaby (approved):
I pushed two commits. Please check that. @gee05053 Thank you.

PR #2162, opened 2024-01-22
feat: separate the formatting and linting configuration

ironAiken2:
### Changes
- Eslint and Prettier version integration between React and Lit
- Separate formatting and linting scripts
- Ignore eslint rules that conflict with prettier formatting rules
   - "object-curly-spacing": "off": Disables the rule that enforces spacing inside curly braces {} in object literals.
   - "operator-linebreak": "off": Disables the rule that enforces consistent linebreaks for operators.
   - "indent": "off": Disables the rule that enforces a specific indentation style.
   - "no-unused-vars": "off": Disables warnings for unused variables.
   - "no-undef": "off": Disables warnings for using undeclared variables.
   - "no-constant-condition": "off": Disables warnings for constant expressions in conditions.
   - "no-inner-declarations": "off": Disables the rule that forbids function declarations in nested blocks.
   - "no-irregular-whitespace": "off": Disables warnings for irregular whitespace outside of strings and comments.
   - "max-len": "off": Disables the rule that enforces a maximum line length.
   - "@typescript-eslint/no-explicit-any": "off": Disables warnings for using the any type in TypeScript.
   - "require-jsdoc": "off": Disables the rule that requires JSDoc comments for every function.
   - "camelcase": "off": Disables the rule that enforces camelCase naming convention for variable and function names.
   - "spaced-comment": "off": Disables the rule that enforces spacing after the comment start symbol.
   - "quotes": "off": Disables the rule that enforces the use of a specific quote type (e.g., single or double quotes) for strings.

**Checklist:** (if applicable)

- [ ] Mention to the original issue
- [ ] Documentation
- [ ] Minium required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [ ] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

PR #2168, opened 2024-01-24
feature: provide UI for updating resource allocation in model service

lizable:

- [ ] Documentation

### DIFF
--- a/react/src/components/ModelServiceSettingModal.tsx
+++ b/react/src/components/ModelServiceSettingModal.tsx
@@ -3,12 +3,18 @@ import { useSuspendedBackendaiClient } from '../hooks';
 import { useTanMutation } from '../hooks/reactQueryAlias';
 import BAIModal, { BAIModalProps } from './BAIModal';
 import Flex from './Flex';
+import ImageEnvironmentSelectFormItems from './ImageEnvironmentSelectFormItems';
+import ResourceAllocationFormItems, {
+  RESOURCE_ALLOCATION_INITIAL_FORM_VALUES,
+} from './ResourceAllocationFormItems';
+import SliderInputFormItem from './SliderInputFormItem';
+import { ModelServiceSettingModalModifyMutation } from './__generated__/ModelServiceSettingModalModifyMutation.graphql';
 import { ModelServiceSettingModal_endpoint$key } from './__generated__/ModelServiceSettingModal_endpoint.graphql';
 import { Form, InputNumber, theme } from 'antd';
 import graphql from 'babel-plugin-relay/macro';
-import React from 'react';
+import React, { Suspense } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useFragment } from 'react-relay';
+import { useFragment, useMutation } from 'react-relay';
 
 interface Props extends BAIModalProps {
   endpointFrgmt: ModelServiceSettingModal_endpoint$key | null;
@@ -34,6 +40,11 @@ const ModelServiceSettingModal: React.FC<Props> = ({
       fragment ModelServiceSettingModal_endpoint on Endpoint {
         endpoint_id
         desired_session_count
+        resource_slots
+        resource_opts
+        cluster_mode
+        cluster_size
+        image
       }
     `,
     endpointFrgmt,
@@ -53,21 +64,61 @@ const ModelServiceSettingModal: React.FC<Props> = ({
     },
   });
 
+  const [
+    commitModifyEndpoint,
+    // inInFlightCommitModifyEndpoint
+  ] = useMutation<ModelServiceSettingModalModifyMutation>(graphql`
+    mutation ModelServiceSettingModalModifyMutation(
+      $endpoint_id: UUID!
+      $props: ModifyEndpointInput!
+    ) {
+      modify_endpoint(endpoint_id: $endpoint_id, props: $props) {
+        ok
+        msg
+      }
+    }
+  `);
+
   // Apply any operation after clicking OK button
   const handleOk = (e: React.MouseEvent<HTMLElement>) => {
     form
       .validateFields()
       .then((values) => {
-        mutationToUpdateService.mutate(values, {
-          onSuccess: () => {
-            console.log('service updated');
+        commitModifyEndpoint({
+          variables: {
+            // @ts-ignore
+            endpoint_id: endpoint?.endpoint_id,
+            props: {
+              // TODO: setting resource_slots based on form values
+              resource_slots: JSON.stringify({
+                cpu: 2,
+                mem: '4g',
+              }),
+            },
+          },
+          onCompleted(response) {
+            if (response?.modify_endpoint?.ok) {
+              console.log('successfully modified');
+            } else {
+              console.log('error occurred');
+            }
             onRequestClose(true);
           },
-          onError: (error) => {
-            console.log(error);
-            // TODO: show error message
+          onError: (err) => {
+            // console.log(err);
+            throw err;
           },
         });
+        // mutationToUpdateService.mutate(values, {
+        //   onSuccess: () => {
+        //     console.log('service updated');
+        //     onRequestClose(true);
+        //   },
+        //   onError: (error) => {
+        //     console.log(error);
+        //     // TODO: show error message
+        //   },
+        // });
       })
       .catch((err) => {
         console.log(err);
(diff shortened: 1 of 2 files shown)


react/src/components/ServiceLauncherModal.tsx
```diff
@@ -151,27 +176,93 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
...
+    // inInFlightCommitModifyEndpoint
+  ] = useMutation<ServiceLauncherModalModifyMutation>(graphql`
+    mutation ServiceLauncherModalModifyMutation(
+      $endpoint_id: UUID!
+      $props: ModifyEndpointInput!
+    ) {
+      modify_endpoint(endpoint_id: $endpoint_id, props: $props) {
+        ok
+        msg
+      }
```
yomybaby: (This feedback is related to core's GraphQL Schema, not WebUI)
The return type of `modfy_endpoint` GraphQL query needs to be changed. It should return `Endpoint` type.

react/src/pages/ServingListPage.tsx
```diff
@@ -59,9 +58,8 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
   const curProject = useCurrentProjectValue();
   const [isOpenServiceLauncher, setIsOpenServiceLauncher] = useState(false);
   const [isOpenColumnsSetting, setIsOpenColumnsSetting] = useState(false);
-  const [selectedModelService, setSelectedModelService] = useState<Endpoint>();
-  const [isOpenModelServiceSettingModal, setIsOpenModelServiceSettingModal] =
-    useState(false);
+  const [selectedModelService, setSelectedModelService] =
+    useState<Endpoint | null>(null);
```
yomybaby: Let's seperate the `selectedModelService` into `editingModelService` and `deletingModelService`.

It also can be used to determine wether modal is open or not.

For example,
```
 <BAIModal
        open={!deletingModelService}
        title={t('dialog.title.LetsDouble-Check')}
```

yomybaby (changes requested):
https://github.com/lablup/backend.ai-webui/blob/b6286435a41162c8a4643d2c0d69faa0a70e044a/react/src/pages/ServingListPage.tsx#L331-L337

This [code](https://github.com/lablup/backend.ai-webui/blob/b6286435a41162c8a4643d2c0d69faa0a70e044a/react/src/pages/ServingListPage.tsx#L331-L337) is not modified in this PR, but it's better to improve.

Use an argument instead of referring to selectedModelService.
```js
>({
    // TODO: Use an argument instead of referring to selectedModelService.
    mutationFn: (endpoint_id:string) => {
      return baiSignedRequestWithPromise({
        method: 'DELETE',
        url: '/services/' + endpoint_id,
        client: baiClient,
      });
    },
    ```

File: react/src/components/ServiceLauncherModal.tsx (hunk 1 above)
Comment (lizable) [reply react/src/components/ServiceLauncherModal.tsx]: I just updated query in server-side but somehow generated graphql file didn't change. I need to dig it more...
https://github.com/lablup/backend.ai/pull/1859/files#diff-09158c3c58d5e1ee85d64c91e1805be9cdbcc9ba856492cb8a8e759760a9fd5dR667

File: react/src/pages/ServingListPage.tsx (hunk 2 above)
Comment (lizable) [reply react/src/pages/ServingListPage.tsx:63]: Hmm, I think It's already set according to row in table data, and I think deletingModelService is already set as a name of "`isOpenModelServiceTerminatingModal`"

File: react/src/components/ServiceLauncherModal.tsx (hunk 1 above)
Comment (lizable) [reply react/src/components/ServiceLauncherModal.tsx]: Resolved!

**Review by agatha197 — CHANGES_REQUESTED** [review]
- Can you add a margin to the bottom buttons?
image

- When modifying the service, the GPU input is still initialized.
image
image

**Comment by lizable** [comment]
> * Can you add a margin to the bottom buttons?
>   image
> * When modifying the service, the GPU input is still initialized.
>   image
>   image

I just fixed AI accelerator initial value setting issue that you mentioned above and also add some spaces between create/update button and cancel button. Could you check it again @agatha197 ?

Screenshot 2024-02-01 at 11 38 41 AM

File: react/src/components/ServiceLauncherModal.tsx [hunk 3]
```diff
@@ -184,16 +297,46 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
...
+      title={
+        endpoint
+          ? t('modelService.EditModelService')
+          : t('modelService.StartNewServing')
+      }
+      destroyOnClose
       maskClosable={false}
+      footer={() => (
+        <Flex direction="row" justify="end" align="end">
+          <Space size="middle">
```
agatha197:
```suggestion
          <Space size="small">
```
The service modal
image
Generate token modal
image

If you change the size to small, it looks like below.
image

react/src/components/ImageEnvironmentSelectFormItems.tsx
```diff
@@ -129,18 +129,42 @@ const ImageEnvironmentSelectFormItems: React.FC<
...
   // auto select when relative field is changed
   useEffect(() => {
     if (!_.isEmpty(environments?.manual)) {
       return;
     }
+
+    let matchedEnvironmentByVersion:
+      | ImageGroup['environmentGroups'][0]
+      | undefined;
+    // let matchedImageByVersion;
```
agatha197:
```suggestion
```

react/src/components/ImageEnvironmentSelectFormItems.tsx
```diff
@@ -129,18 +129,42 @@ const ImageEnvironmentSelectFormItems: React.FC<
...
+      _.find(imageGroups, (group) => {
+        return _.find(group.environmentGroups, (environment) => {
+          return _.find(environment.images, (image) => {
+            const matched =
+              getImageFullName(image) ===
+              form.getFieldValue('environments')?.version;
+
+            if (matched) {
+              matchedEnvironmentByVersion = environment;
+              // matchedImageByVersion = image;
```
agatha197:
```suggestion
```

agatha197 (dismissed):
It works well 👏  I left just minor things. Please reflect them :)

resources/i18n/ko.json
```diff
@@ -424,7 +424,11 @@
     "Image": "이미지",
     "Owner": "소유자",
     "ServiceDelegatedFrom": "모델 서비스는 {{ createdUser }}에 의해 생성되지만 세션 소유권은 {{ sessionOwner }}에 위임됩니다.",
-    "FailedToTerminateService": "서비스 삭제 요청이 실패하였습니다."
+    "FailedToTerminateService": "서비스 삭제 요청이 실패하였습니다.",
+    "FailedToUpdateService": "서비스를 업데이트하지 못했습니다.",
+    "ServiceCreated": "모델 서비스 {{ 이름 }}이(가) 성공적으로 생성되었습니다.",
```
yomybaby: It should be "name" instead of "이름", right? Please check other languages.

react/src/components/ImageEnvironmentSelectFormItems.tsx
```diff
@@ -298,6 +331,20 @@ const ImageEnvironmentSelectFormItems: React.FC<
...
             }
+            // NOTE: when user set environment only then set the version to the first item
+            else {
+              const firstInListImage: Image = imageGroups
+                .flatMap((group) => group.environmentGroups)
+                .filter((envGroup) => envGroup.environmentName === value)[0]
+                .images[0];
+              form.setFieldsValue({
+                environments: {
+                  environment: firstInListImage?.name || '',
```
yomybaby, line 341: You don't need to set `environment` again. It's already `value`.

react/src/components/ImageEnvironmentSelectFormItems.tsx
```diff
@@ -163,19 +188,27 @@ const ImageEnvironmentSelectFormItems: React.FC<
...
-      }
+      nextNewImage = nextEnvironmentGroup?.images[0];
+    } else {
+      // if image info is partially set, fill out image info
+      nextNewImage = _.find(
+        nextEnvironmentGroup?.images,
+        (image) =>
+          form.getFieldValue('environments')?.version ===
+          getImageFullName(image),
+      );
```
yomybaby: Could you let me know how to reproduce the bug you solved by adding these lines??

yomybaby (changes requested):

react/src/components/ModelServiceSettingModal.tsx
yomybaby, line 1: Let's restore this previous version and use it only for backward compatibility(~23.09.8).

yomybaby (changes requested):

react/src/components/ModelServiceSettingModal.tsx
yomybaby: Or you can make `ServiceLauncherModal` compatible with the previous version. 👍

react/src/components/ServiceLauncherModal.tsx
```diff
@@ -151,27 +237,118 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
...
-            } else {
-              message.error(t('modelService.FailedToStartService'));
-            }
-          },
-        });
+              if (error) {
+                message.error(t('dialog.ErrorOccurred'));
+              } else {
+                onRequestClose(true);
+              }
```
yomybaby: When an error exists, both message.success and message.error are executed.
```suggestion
              if (error) {
                message.error(t('dialog.ErrorOccurred'));
              } else {
                message.success(
                  t('modelService.ServiceUpdated', {
                    name: updatedEndpoint?.name,
                  }),
                );
                onRequestClose(true);
              }
```

yomybaby: image

yomybaby (changes requested):
Please check my requests.

src/components/backend-ai-environment-list.ts
```diff
@@ -965,9 +965,8 @@ export default class BackendAIEnvironmentList extends BackendAIPage {
         );
     } else {
       this.modifyImageCudaGpu.label = _t('environment.Disabled') as string;
-      (
-        this.shadowRoot?.querySelector('mwc-slider#cuda-gpu') as Slider
-      ).value = 0;
+      (this.shadowRoot?.querySelector('mwc-slider#cuda-gpu') as Slider).value =
+        0;
```
yomybaby, line 969: This PR includes several changes to update the code format only. I believe this was caused by https://github.com/lablup/backend.ai-webui/pull/2162. After this, the formatting rules have slightly changed. In that PR, we should have reapplied the rules to all files. I will create a separate issue to reapply r format rule to all files to accommodate the slightly altered rules.

PR #2121, opened 2023-12-28
feat: React root including MainLayout, useBAINotification

yomybaby:
This PR modifies the structure to have React as the root, with the existing backend.ai-webui components nested within. This change includes managing the Main layout, which corresponds to the Sider and Menu, within React.

```[tasklist]
### Feature
- [x] config options related to Sider menus (such as [menu_blocklist](https://github.com/lablup/backend.ai/pull/1727), [menu_inactivelist](https://github.com/lablup/backend.ai/pull/1733)...)
- [x] Documentation link on header ([#9f757](https://github.com/lablup/backend.ai-webui/pull/2121/commits/9f75721b9f951e04f333d403d6c42388b293f749), [#62781](https://github.com/lablup/backend.ai-webui/pull/2121/commits/62781c7141c1856a5dae22d9a2220cd763538adf))
- [x] Notification list
- [x] Task list
- [x] Plugin menu ([#3f3e1](https://github.com/lablup/backend.ai-webui/pull/2121/commits/3f3e11fa28e734f181ac8491377b681a5d67fec6))
- [x] Custom theme for sider configuration ([#f6bf4](https://github.com/lablup/backend.ai-webui/pull/2164/commits/f6bf42607ee585b525a47f3abe2229b900b24d1f))
- [x] ~~Apply light bg color to mwc-tab-bar~~ => https://github.com/lablup/backend.ai-webui/issues/2160
- [x] Collapse sidebar config (`Use Compact Sidebar by default`) ([#2288c](https://github.com/lablup/backend.ai-webui/pull/2121/commits/2288ccfd3d0191d05b1c3cfb760d072988e321bc))
- [x] Add notification / task filtering
- [x] 404 page
- [x] ~~routing 401, 404 and plugin pages in React part (react-router redirect)~~ => https://github.com/lablup/backend.ai-webui/issues/2160
- [x] ~~Timeout & off-line notification~~ => https://github.com/lablup/backend.ai-webui/issues/2160
- [x] Display/Hide menu item on Sider depends on manage version (sujin) ([#bd94355](https://github.com/lablup/backend.ai-webui/pull/2121/commits/bd94355b8bfc0e5c4f18a56bd169339eb83931fc))
```

```[tasklist]
### UI/Layout changes and bugs
- [x] Add padding to the notification title (or close button) so that there is no overlapping.
- [x] `bg_task_failed` or `task_failed` of background-task SSE? (handle both)
[...]

PR #2178, opened 2024-01-26
fix: `initialValues` props of `<Form/>` under Modal doesn't work in Electron

yomybaby:
This bug was found during the review of #2121.

The `initialValues` props don't work in the specific case below:
- In Electron, the modal has the `<Form/>` with `initialValues` prop. This is the case even when `destroyOnClose` of Modal and `preserve=false` of Form are set.

This PR has the following changes:
- Use `ref` of Form instead of `form` if available. (FYI, `Form.useWatch` cannot be used with `ref` because `ref` can be null)
- Set the row key to the table of ContainerRegistryList

### How to test
- Please check the electron environment and browser environment. For the election environment, you can build `make clean` and `make mac_arm64` (other [platforms](https://github.com/lablup/backend.ai-webui/blob/main/Makefile#L54-L60) are available.)
- Please check the modals that have been changed.

- [ ] Documentation


agatha197 (approved):
LGTM

PR #2180, opened 2024-01-26
feat: autoSelectDefault and showUsageStatus props of StorageSelector

### LINKED ISSUE #2151: feat: `autoSelectDefault` and `showUsageStatus` props of `StorageSelector`
### Main idea

The host select control of `backend-ai-data-view` has a feature to automatically select based on usage and display the status of usage.
image
image

Let's make `StorageSelector.tsx` has same features: 
- the prop name are `autoSelectDefault` and `showUsageStatus`
  ```
  <StorageSelect autoSelectDefault showUsageStatus />
  ```
  - `autoSelectDefault` use `useEffect` to select rendering after getting status data.
  - `showUsageStatus`
- rename `StorageSelector` to `StorageSelect`

### Alternative ideas

_No response_

### Anything else?

_No response_

ironAiken2:
### This PR resolves #2151 

feature:
- autoSelectDefault
:  set default selected host, "default" | "usage" or no use. 
- showUsageSelect
: "true" for showing usage in each host, "false" for disabling the feature.
- useControllableValue
: StorageSelect component can be controllable or not. If the parent passes state, it uses it; otherwise, it uses its own state.

- [ ] Documentation


ironAiken2:
But I thing that the upgrading version issue covers a large area and should be fixed by addressing the issue in the current version of the project.

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+        value={value?.id || value}
+        onChange={(host) => {
+          onChange?.(host, {
+            id: host,
+            ...(vhostInfo?.volume_info[host] || {}),
+          });
+        }}
+        options={options}
+        optionRender={(option) => (
+          <Space>
```
yomybaby: Please use 'Flex' instead of `<Space/>` if available.

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+            {option.label}
+            {showUsageStatus && (
+              <div
```
yomybaby: How about '[Badge](https://ant.design/components/badge#badge-demo-status)'?

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+  sftp_scaling_groups: string[];
+};
+interface Props extends Omit<SelectProps, 'value' | 'onChange'> {
+  value?: string | VolumeInfo;
+  onChange?: (hostName: string, volumeInfo: VolumeInfo) => void;
+  autoSelectDefault?: boolean;
+  showUsageStatus?: boolean;
+}
+const StorageSelect: React.FC<Props> = ({
+  value,
```
yomybaby, line 32: Let's use [ahooks's useControllerbleValue](https://ahooks.js.org/hooks/use-controllable-value)

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const idx = percentage < 70 ? 0 : percentage < 90 ? 1 : 2;
+    const type = ['adequate', 'caution', 'insufficient'][idx];
+
+    return type;
+  };
+
+  return (
+    <>
+      <style>{rawSelectCss}</style>
+      <Select
```
yomybaby: Form controls such as Select, Input, etc should return one element for `Form.Item`.

yomybaby (changes requested):

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+      const percentage =
+        vhostInfo?.volume_info[option.label]?.usage?.percentage;
+      const idx = percentage < 70 ? 0 : percentage < 90 ? 1 : 2;
+
+      if (showUsageStatus) {
+        return (
+          <Flex justify="between" align="center">
+            <Badge
+              // @ts-ignore
+              status={status[idx]}
```
lizable: We already have color setting function in index. Please refer to this code below and update Badge component props.
https://github.com/lablup/backend.ai-webui/blob/ff29a586a1e7af517289fa1999450692f638282e/react/src/helper/index.tsx#L227-L233

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+      selectedHost = lowestUsageHost?.host || selectedHost;
+    }
+    setState({
+      id: selectedHost,
+      ...(vhostInfo?.volume_info[selectedHost] || {}),
+    });
+  }, [vhostInfo]);
+
+  const optionRender = useMemo(() => {
+    const status = ['success', 'warning', 'error'];
```
lizable: And I think it's not necessary if we use color instead.

lizable (approved):
I just found some minor test failures in github action, originated from dependency array in hooks([useEffect](https://github.com/lablup/backend.ai-webui/pull/2180/files/a8537491593776c31e4485e01c4163b2d5fd87e5#diff-bec9f3b247068290d015fe25aac131df101c7055e4a8d8276085abd1f90ed04cR86) and [useMemo](https://github.com/lablup/backend.ai-webui/pull/2180/files/a8537491593776c31e4485e01c4163b2d5fd87e5#diff-bec9f3b247068290d015fe25aac131df101c7055e4a8d8276085abd1f90ed04cR66)). You could either follow the guidance from github action or just input right above the line if there has certain reason (https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528).

lizable (changes requested):
Please resolve test failure on github action following the comments above.

ironAiken2:
I thought the "vhostInfo" data is used directly within the useEffect, and changes in it have a significant impact on determining the selectd host. So I added vhostInfo to the dependencies array. 

However, the other items are only configurable in code, so I removed them from the Dependencies array because I thought they were less likely to be changed because the user can't change their settings.

lizable (approved):
LGTM.

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+    allowed: Array<string>;
+    volume_info: any;
+  }>('vhostInfo', () => {
+    return baiClient.vfolder.list_hosts();
+  });
+
+  const [state, setState] = useControllableValue({ value, onChange });
+
+  useEffect(() => {
+    console.log('!');
```
yomybaby: Please remove this line.

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }, [vhostInfo]);
+
+  return (
+    <Select
+      filterOption={true}
+      placeholder={t('data.SelectStorageHost')}
+      loading={isLoadingVhostInfo}
+      style={{
+        minWidth: 165,
+        direction: 'ltr',
```
yomybaby: Is there any reason to set the direction here?

react/src/components/StorageSelect.tsx
```diff
@@ -0,0 +1,120 @@
...
+      optionLabelProp="value"
+      options={_.map(vhostInfo?.allowed, (host) => ({
+        label: showUsageStatus ? (
+          <Flex align="center" gap={'xs'}>
+            {/* TODO: add tooltip for '여유/주의/부족' */}
+            {vhostInfo?.volume_info[host]?.usage && (
+              <Tooltip
+                title="여유 or 주의 or 부족"
+                // @ts-ignore
+                getPopupContainer={() => shadowRoot}
```
yomybaby: I push a commit to add a tooltip with `getPopupContainer`. Please add a proper text for each case. @ironAiken2

yomybaby (changes requested):

yomybaby:
I pushed commits to display search highlights and update type definition. @ironAiken2 , Please check.
I merged this PR. Thank you!

PR #2184, opened 2024-01-29
Fix/draggable header for electron

yomybaby:
After #2121, there is a bug related to the draggable area to move an Electron app window.

The blue box area is the draggable area, except for the red box area.
image

Please build an app using `make` command (below command for `make mac_arm64`) and test the draggable area.
```
make clean
make mac_arm64
```

- [ ] Documentation


gee05053 (approved):
Please check commit that I fixed lint error.
LGTM!

PR #2188, opened 2024-01-29
feat: expose TCP apps from browser-based WebUI

kyujin-cho:
This PR adds support for non-auth TCP proxy mode, which allows TCP apps to be spawned directly without arbitrary HTTP authentication mechanism. When setting `[resources].allowNonAuthTCP` to `true`, WebUI will now display TCP apps not only from electron-based desktop app but also from browser. Requires updated AppProxy to test this feature.


lizable (approved):
VSCode app execution(restricted mode) on local machine and ssh/sftp execution works fine. LGTM.


### Screenshot(s)
Screenshot 2024-04-05 at 12 58 11 AM
Screenshot 2024-04-05 at 12 58 19 AM

PR #2192, opened 2024-01-30
feat: set RegExpression for validating URL pattern

ironAiken2:
### Main Idea
There are some chances that user could use customized hostname with port number, but antd input form validation on URL type doesn't work properly on these cases. URL validation need to accept the addresses like "bai-repo:####".

### Changes
1. use new URL for url validation
2. Set Error Message for Invalid type URL

- [ ] Documentation

### DIFF
--- a/react/src/components/ContainerRegistryEditorModal.tsx
+++ b/react/src/components/ContainerRegistryEditorModal.tsx
@@ -254,9 +254,6 @@ const ContainerRegistryEditorModal: React.FC<
               required: true,
             },
             {
-              type: 'url',
-            },
-            {
               validator: (_, value) => {
                 if (
                   value &&
@@ -265,6 +262,14 @@ const ContainerRegistryEditorModal: React.FC<
                 ) {
                   return Promise.reject(t('registry.DescURLStartString'));
                 }
+                const pattern = new RegExp(
+                  '^(https?|ftp)://[a-zA-Z0-9-_.]+(:[0-9]{1,5})?(/[a-zA-Z0-9-_./]*)?$',
+                );
+
+                if (value && !pattern.test(value)) {
+                  return Promise.reject(t('registry.DescURLFormat'));
+                }
+
                 return Promise.resolve();
               },
             },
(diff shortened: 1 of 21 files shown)


react/src/components/ContainerRegistryEditorModal.tsx
```diff
@@ -265,6 +262,14 @@ const ContainerRegistryEditorModal: React.FC<
                 ) {
                   return Promise.reject(t('registry.DescURLStartString'));
                 }
+                const pattern = new RegExp(
```
yomybaby: Is there any reason not to use [`new URL(url)`](https://developer.mozilla.org/ko/docs/Web/API/URL/URL) to validate URL?

yomybaby: Oh sorry, you mentioned in the PR description. I'll check

yomybaby: @ironAiken2  I think it's better to use `new URL()` instead of creating and managing a custom regular expression. Even if `new URL()` sees a URL with spaces as okay, it's better to have a more flexible validation that considers all valid URLs.

yomybaby (changes requested):

react/src/components/ContainerRegistryEditorModal.tsx
lizable: I agree with @yomybaby. Besides, the performance of new URL is better than RegEx.

lizable (changes requested):
Please update the regex pattern to new URL by calling origin prop (reference: https://developer.mozilla.org/en-US/docs/Web/API/URL/origin)

yomybaby (approved):
LGTM

react/src/components/ContainerRegistryEditorModal.tsx
```diff
@@ -253,17 +253,19 @@ const ContainerRegistryEditorModal: React.FC<
...
               validator: (_, value) => {
-                if (
-                  value &&
-                  !value.startsWith('http://') &&
-                  !value.startsWith('https://')
-                ) {
-                  return Promise.reject(t('registry.DescURLStartString'));
+                if (value) {
+                  if (
+                    !value.startsWith('http://') &&
```
yomybaby, line 260: You can use `protocol` property of `new URL()`.

PR #2197, opened 2024-02-01
add an `undefined` type to `imageInfo` and a conditional chain

yomybaby:

- [ ] Documentation


yomybaby:
I merged this without review because it's a hotfix and doesn't change the logic at all.

PR #2199, opened 2024-02-01
feat: apply ellipsis into notification

### LINKED ISSUE #2196: Added exception handling for long text in BAINotificationItem

Linux (x86-64)

### Backend.AI version

24.03.0a1

### Describe the bug

Exception handling for long text is required. The close button disappears, and can't see the entire message. 

image

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

ironAiken2:
### This PR resolves #2196 

feature :
- Added allowEllipsis prop to control ellipsis property. (allowEllipsis (boolean))
- Users can use the ellipsis feature by clicking on the text.

- [ ] Documentation

### DIFF
--- a/react/src/components/BAINotificationItem.tsx
+++ b/react/src/components/BAINotificationItem.tsx
@@ -5,7 +5,6 @@ import {
   ClockCircleOutlined,
   CloseCircleOutlined,
 } from '@ant-design/icons';
-import { useControllableValue } from 'ahooks';
 import { Card, List, Progress, Typography, theme } from 'antd';
 import dayjs from 'dayjs';
 import _ from 'lodash';
@@ -24,9 +23,7 @@ const BAINotificationItem: React.FC<{
   const { t } = useTranslation();
   const { token } = theme.useToken();
   const [showExtraDescription, setShowExtraDescription] = useState(false);
-  const [ellipsis, setEllipsis] = useControllableValue({
-    defaultValue: allowEllipsis,
-  });
+  const [ellipsis, setEllipsis] = useState(allowEllipsis ?? false);
   return (
     <List.Item>
       <Flex direction="column" align="stretch" gap={'xxs'}>
@@ -63,7 +60,7 @@ const BAINotificationItem: React.FC<{
             }}
             ellipsis={ellipsis ? { rows: 3 } : false}
             onClick={() => {
-              setEllipsis(!ellipsis);
+              !allowEllipsis && setEllipsis(!ellipsis);
             }}
           >
             {notification.message}
@@ -74,7 +71,7 @@ const BAINotificationItem: React.FC<{
           <Typography.Paragraph
             ellipsis={ellipsis ? { rows: 3 } : false}
             onClick={() => {
-              setEllipsis(!ellipsis);
+              !allowEllipsis && setEllipsis(!ellipsis);
             }}
           >
             {notification.description}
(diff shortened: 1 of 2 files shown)


react/src/components/BAINotificationItem.tsx
```diff
@@ -18,10 +19,14 @@ const BAINotificationItem: React.FC<{
...
   showDate?: boolean;
-}> = ({ notification, onClickAction, showDate }) => {
+  allowEllipsis?: boolean;
+}> = ({ notification, onClickAction, showDate, allowEllipsis }) => {
   const { t } = useTranslation();
   const { token } = theme.useToken();
   const [showExtraDescription, setShowExtraDescription] = useState(false);
+  const [ellipsis, setEllipsis] = useControllableValue({
+    defaultValue: allowEllipsis,
+  });
```
yomybaby: `useControllableValue` helps manage the state to be either self-managed or controlled by its parent. 
`allowEllipsis` is not applicable in this case. `allowEllipsis` determines whether ellipsis is allowed or not. Ellipsis refers to the current state of being ellipsised or not, which is not related to the allowance of ellipsis.

yomybaby: Also, you might not need to manage the ellipsis status and don't need to handle it using the `click` event. Typography has a feature to manage it, such as the `expandable` property. Please check [this](https://ant.design/components/typography#components-typography-demo-ellipsis).

yomybaby (changes requested):

react/src/components/BAINotificationItem.tsx
ironAiken2: @yomybaby 

The `allowEllipsi` is simply a property that turns ellipsis on/off, so I didn't think it was right to use a `useControllableValue`, so I removed it. 
The `expendable` property only provides the ability to expand the omitted text again, not to omit the expanded text again. So in the sider that holds all the notifications, I want to support both collapsing and expanding, so I'm controlling the `ellipsis` property with a click event.

yomybaby: It's okay if it can't use ellipsis again. The main points are:
- The clickable content area is good for the main action. For example, clicking the "success creating session" notification takes you to the session detail page. It's better not to use the clicking action for toggling. Let's reserve clicking for another feature later.
- The UX for toggling ellipsis by clicking again is uncommon and not explicit.

yomybaby: Could you let me know what the longest notification you have found is?

ironAiken2: @yomybaby 
I decided that scalability would be an afterthought, so I modified the code to focus on fixing the mentioned issues.

[ref](https://teams.microsoft.com/l/message/19:14c484402d874dafb15806d093b95a82@thread.skype/1706621647590?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=74ae2c4d-ec4d-4fdf-b2c2-f5041d1e8631&parentMessageId=1706621647590&teamName=devops&channelName=Frontend&createdTime=1706621647590) The error occurs when turn off storage and access the web.  I didn't find the error in my test environment, so I tested it by hardcoding the notification text for the usage calculation part. 

image

yomybaby (approved):
LGTM

PR #2206, opened 2024-02-12
fix:  main layout header appears behind fixed column table

yomybaby:
(cherry picked from commit fa84a860dae37fc3fd3e16a2155abca92e441fba)

This PR resolves UI bug where main layout header appears behind fixed column.

How to reproduce this bug:
- Go to the page that has a fixed header table, such as `/session/start?step=2` and `/serving`.
- Reduce the browser window size until it has a scroll bar.
- Scroll to the bottom until the table header overlaps the table content.
image

- [ ] Documentation


gee05053 (approved):
LGTM!

ironAiken2 (approved):

PR #2211, opened 2024-02-14
fix: general setting page is empty after navigation from log tab

yomybaby:
###  How to Reproduce:
- Refresh the browser on the /summary page.
- Click the user dropdown in the top right corner.
- Click the "log / error" item.
- After navigating to the "Settings & Logs" page, click the "General" tab.

You will see that the tab content is blank.

- [ ] Documentation


gee05053 (approved):
LGTM!

Issue #2191, opened 2024-01-30
bug: Tooltip explanning mounted VFolders is not appeared perfectly at NeoSession


gee05053:
### What Operating System(s) are you seeing this problem on?

macOS (Apple Silicon)

### Backend.AI version

24.03.0a1

### Describe the bug

In NeoSession, if mounted Vfolder is too many, Tooptip is appeared to explan what is mounted VFolders.
Howerver after merging React-root, Tooltip is not appeared perfectly like below screenshot.

![스크린샷 2024-01-30 오후 2 47 16](https://github.com/lablup/backend.ai-webui/assets/28584221/a051da21-bcbd-4116-9c05-68e6a7922209)

yomybaby:
This bug is handled in #2220

PR #2220, opened 2024-02-20
feat: Neo session launcher as a default

yomybaby:
This PR:
- handles the Neo session launcher/list by react root (resolve https://github.com/lablup/backend.ai-webui/issues/2191)
- runs the `runShmemAutomationRule` when the preset is changed. 
- enables the allocatable resource preset only automatically.
- prevents session name autocompletion and fixes the session name validation. 
- displays the infinite symbol for session limit is 1000000.
- adds a range compare helper function and test code. 
- adds tests for handling undefined limits in the `isOutsideRange(WithUnits)` function. 
- include draft version of component will be used in Neo session list (You don't need to review these components yet because is Neo session list is in draft also)
  - `ResourceAvailableGageBar`, `AvailableResourcesCard.tsx`
  
You don't need to review the components below because they are in the draft version and not used yet.
- `react/src/components/AvailableResourcesCard.tsx`
-  `react/src/components/ResourceAvailableGageBar.tsx`
-  `react/src/components/SessionList.tsx`
-  `react/src/components/SessionListColums/SessionInfoCell.tsx`
-  `react/src/pages/SessionListPage.tsx`

- [ ] Documentation

---

copliot:all


react/src/components/NeoSessionLauncherSwitchAlert.tsx
```diff
@@ -0,0 +1,83 @@
...
+              {t('session.launcher.NeoSwitchDescription', {
+                launcher: is2409Launcher
+                  ? t('session.launcher.PreviousLauncher')
+                  : t('session.launcher.NeoLauncher'),
+              })}
+            </Typography.Text>
+            <Segmented
+              options={[
+                {
+                  label: t('session.launcher.PreviousLauncher'),
```
agatha197:
```suggestion
                  label: (
                    <Typography.Text
                      style={{
                        color: !is2409Launcher ? undefined : token.colorPrimary,
                      }}
                    >
                      {t('session.launcher.PreviousLauncher')}
                    </Typography.Text>
                  ),
```

image

↓
image
For dark mode, I recommend changing the previous version to the default color as well :)

react/src/hooks/useBAISetting.tsx
```diff
@@ -0,0 +1,142 @@
...
+} from 'recoil';
+
+export const localStorageEffect =
+  (
+    key: string,
+    onInitialize: (savedValue: any) => any = (v) => v,
+  ): AtomEffect<any> =>
+  ({ setSelf, onSet }) => {
+    const savedValue = localStorage.getItem(key);
+    setSelf(onInitialize(savedValue != null ? JSON.parse(savedValue) : null));
```
agatha197:
```suggestion
    setSelf(onInitialize(savedValue !== null ? JSON.parse(savedValue) : null));
```

src/components/backend-ai-session-launcher.ts
```diff
@@ -1443,6 +1446,33 @@ export default class BackendAiSessionLauncher extends BackendAIPage {
    * Else, launch session dialog.
    * */
   async _launchSessionDialog() {
+    let shouldNeo = false;
+    try {
+      shouldNeo = !JSON.parse(
+        localStorage.getItem('backendaiwebui.settings.user') || '{}',
+      )['use_2409_session_launcher'];
```
agatha197, line 1453:
```suggestion
      )['use_2403_session_launcher'];
```

src/plugins.gitignore/README.md
agatha197, line 1: Did you add this file on purpose?

agatha197 (changes requested):
## 1
Would it be better to message that the session name should not end in special characters?
image

## 2
It's better to automatically modify the custom input value like the number of sessions.
1. I got the error message when I wrote 0 to CPU at first.
	image

2. But it automatically changes to the minimum value of 1 when I put the mouse on the input and then turn it off again.
	image


On the other hand, the sessions number is automatically changed to 1 which is the minimum value.
image

## 3
The left time of starting the batch session is shown with delay. How about using `BatchSessionScheduledTimeSetting`  like original session launcher?
image

This is the previous session launcher's scheduled time selector.
image

src/plugins.gitignore/README.md
yomybaby: I removed files. 😉

react/src/hooks/useBAISetting.tsx
yomybaby: I pushed a commit to fix this.

react/src/components/NeoSessionLauncherSwitchAlert.tsx
yomybaby: I accept your suggestion. Thank you.

agatha197 (changes requested):
## 4
image
If there isn't supported AI accelerator, user can see the '0' string without any suffix. It would be better to represent as a 'No accelerator' or remove it.

## 5
image
When I set the batch session without the start command, the input field can be active. But I can't input any string. I think it's better to make readonly the input field or just represent it with strings like the session name.

## 6
I cannot confirm the setting options of `High-Performance Computing Optimizations` on the last page. Is it intended?

yomybaby:
@agatha197 [6233a8b](https://github.com/lablup/backend.ai-webui/pull/2220/commits/6233a8b55a6c44f3167889dba520010e52e8e13a) resolves your number 3 feedback.

yomybaby:
Please check recent commit to resolve your feedback 
- feedback 4: https://github.com/lablup/backend.ai-webui/pull/2220/commits/55c0609cecb136f666e7a976fda8b83931c0a2a6 
- feedback 5: https://github.com/lablup/backend.ai-webui/pull/2220/commits/bfb2df41cd078b37a88f8f2d61393cc9fe26395d

yomybaby:
@agatha197 How can I reproduce feedback 2?

yomybaby:
@agatha197  I pushed several commits related to `useBAISetting.txs` which are used by `NeoSessionLauncherSwitchAlert`.

agatha197:
> kimsujin (@agatha197) How can I reproduce feedback 2? I think it is fixed already. Please check again. Thank you.

Drag and input 0 and click outside of the input :) I know, it's no big deal.
https://github.com/lablup/backend.ai-webui/assets/28584164/f623a7fb-af56-46b9-898b-3fd398acd4ed

agatha197:
> Please check recent commit to resolve your feedback
> 
> * feedback 4: [55c0609](https://github.com/lablup/backend.ai-webui/commit/55c0609cecb136f666e7a976fda8b83931c0a2a6)
> * feedback 5: [bfb2df4](https://github.com/lablup/backend.ai-webui/commit/bfb2df41cd078b37a88f8f2d61393cc9fe26395d)
> * feedback 6: The previous session launcher doesn't display "High-Performance Computing Optimizations" in the last step. Should we add it?

Ah, I checked it. We don't need to.

agatha197 (changes requested):
## 7
Can you make the startup command to the center?
image

## 8
In the batch session, if the start time is not set, we don't care about it. The session ends after running the startup command. 
But even if I didn't set the start time, the start time seems like being set on the last page. So it's better to represent the start time as '-'.

yomybaby:
Response of Feedback 7 & 8:
In the last step, if there is no input, the item will display the text `None` with the secondary style. https://github.com/lablup/backend.ai-webui/pull/2220/commits/e620228e26a96600bd1245c6334e2c7e8226435c
image
image

yomybaby:
https://github.com/lablup/backend.ai-webui/pull/2220/commits/7cfe5307cf1db7a269951417d1e4ffad5a1ec447 resolves the feedback 2 😉 ✌️

ironAiken2 (changes requested):
- The environment variable value validation does not have i18n applied to it.
- The variable field in environment step performs the `required` validation on focus out with no input after focus, while a value field does not perform the `required` validation until a value is entered.

image

agatha197 (changes requested):
https://github.com/lablup/backend.ai-webui/blob/7cfe5307cf1db7a269951417d1e4ffad5a1ec447/react/src/pages/SessionLauncherPage.tsx#L1120

Please add `column={24}` to prevent this warning.
```
warning.js:30 Warning: [antd: Descriptions] Sum of column `span` in a line not match `column` of Descriptions.`.
```

yomybaby:
> * The environment variable value validation does not have i18n applied to it.
> * The variable field in environment step performs the `required` validation on focus out with no input after focus, while a value field does not perform the `required` validation until a value is entered.
> 
> image

resolved by https://github.com/lablup/backend.ai-webui/pull/2220/commits/f5454551a109c2c705e6e360229e948aca1c85ea and https://github.com/lablup/backend.ai-webui/pull/2220/commits/a3cfb0e42f5a79d8af235009b74776a08dff876e
I could not find a reason why the value field isn't validated on the blur event. So I set the validationTrigger explicitly. @ironAiken2

yomybaby:
@agatha197  @ironAiken2 All requested changes have been resolved. Please review them again.

agatha197 (approved):
LGTM!

ironAiken2 (approved):
LGTM!

PR #2230, opened 2024-02-27
refactor: use `useApp` instead of `contextHolder`

agatha197:
Replace `contextHolder` with `useApp` because we could use static methods of message/notification/Modal form useApp without writing contextHolder manually.
ref: https://ant.design/components/app

- [ ] Documentation


agatha197:
Don't need to update anymore. Close the issue.

PR #2238, opened 2024-03-04
feat: add password change request alert to summary page

agatha197:
resolves #2229 
image

## How to test
1. Log in with admin account.
2. Click Users tab.
3. Click the setting icon of the control column.
4. Enable `Require password change?`.
image

5. Log in with the account you just set up.

## Checklist for reviewer
- [ ] Is an alert displayed if `Require password change?` is true?

**Checklist:** (if applicable)

- [x] Mention to the original issue
- [ ] Documentation
- [x] Minium required manager version: 23.09
- [x] Specific setting for review (eg., KB link, endpoint or how to setup)
- [x] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

PR #2239, opened 2024-03-05
feat: display announcement alert to every page.

agatha197:
The password change request alert is displayed on all pages, according to the traditional implementation method.
follows https://github.com/lablup/backend.ai-webui/pull/2238

- [ ] Documentation


yomybaby (approved):
I pushed a commit to add `Suspense` to prevent flicking of `PasswordChangeRequestAlert`.

LGTM.

PR #2244, opened 2024-03-06
hotfix: regression of image info display in serving page

lizable:

- [ ] Documentation

This PR fixes two thing,
 - [x] Show full-image string in Routing page, as a regression of replacing deprecated `image` key in `endpoint` to `image_object`.
 - [x] Supports architecture on same image string (multiarch) by extracting from full-image string

- Core(manager) version: `23.9.9~`

How to test:
1. Go to Serving page
2. Click `Start service` button.
3. Service launcher modal pops up.
4. Create a service with two cases:
    - [x] Select from environment/version
    - [x] Input image string (`cr.backend.ai/multiarch/python:3.9-ubuntu20.04@aarch64`) in manual image input field
5. Click `Create` button.
6. Click Endpoint Name you just created from step 4.
7. See image info in Routing page.

| After | Before |
|------|--------|
| Screenshot 2024-03-06 at 12 02 14 PM | Screenshot 2024-03-06 at 12 24 07 PM |

### DIFF
--- a/react/src/pages/RoutingListPage.tsx
+++ b/react/src/pages/RoutingListPage.tsx
@@ -224,7 +224,11 @@ const RoutingListPage: React.FC<RoutingListPageProps> = () => {
     return color;
   };
 
-  const fullImageString = `${endpoint?.image_object?.registry}/${endpoint?.image_object?.name}:${endpoint?.image_object?.tag}@${endpoint?.image_object?.architecture}`;
+  const fullImageString: string = (
+    baiClient.supports('modify-endpoint')
+      ? `${endpoint?.image_object?.registry}/${endpoint?.image_object?.name}:${endpoint?.image_object?.tag}@${endpoint?.image_object?.architecture}`
+      : endpoint?.image
+  ) as string;
 
   const resource_opts = JSON.parse(endpoint?.resource_opts || '{}');
   return (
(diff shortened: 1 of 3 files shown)


react/src/pages/RoutingListPage.tsx
```diff
@@ -113,7 +113,26 @@ const RoutingListPage: React.FC<RoutingListPageProps> = () => {
             name
             status
             endpoint_id
-            image
```
yomybaby, line 116: To ensure backward compatibility, we should use different images based on the version.

yomybaby (changes requested):

react/src/pages/RoutingListPage.tsx
lizable: I just added conditional statement to check whether current core version supports `image_object` key in endpoint.

yomybaby: Please add  `@deprecatedSince` like this:
`image @deprecatedSince(version: "23.09.9")`

yomybaby (changes requested):

react/src/pages/ServingListPage.tsx
```diff
@@ -273,6 +273,26 @@ const ServingListPage: React.FC<PropsWithChildren> = ({ children }) => {
               name
               endpoint_id
               image
+              image_object @since(version: "23.09.9") {
```
yomybaby: Thia page doesn't use `image` and `image_object`, right?. If yes, please remove these in this file.

yomybaby (changes requested):

PR #2259, opened 2024-03-12
feat: Quota Scope ID in data page's storage status panel

### LINKED ISSUE #2251: Let's add quota scope id of Vfolder in storage list
### Main idea

There are some requests that need `quota scope id` in order to use it as a partial path of vfolder destination path. Since we already provides VFolder v3 architecture which manages every user-created folder in quota scope id directory, It's better to expose quota scope id into the list in data & storage folder.

### Alternative ideas

_No response_

### Anything else?

_No response_

ironAiken2:
### Feature
- show quota scope id (user id) in storage status panel.
- change antd's descriptions column because 2 line representation is visually better than 3 line representation.
(I've tried showing text at the bottom of the card without using descriptions.item, but it doesn't feel consistent.)
- modify the way render descriptions items that is appropriate for the current antd version. The old way was used in versions 5.8.0 and earlier.

image

- [ ] Documentation

### DIFF
--- a/react/src/components/StorageStatusPanel.tsx
+++ b/react/src/components/StorageStatusPanel.tsx
@@ -291,7 +291,7 @@ const StorageStatusPanel: React.FC<{
       key: 'userQuotaScopeId',
       label: t('data.userQuotaScopeId'),
       children: (
-        <Typography.Text>
+        <Typography.Text copyable>
           {addQuotaScopeTypePrefix('user', user?.id || '')}
         </Typography.Text>
       ),
--- a/react/src/components/StorageStatusPanel.tsx
+++ b/react/src/components/StorageStatusPanel.tsx
@@ -170,11 +170,12 @@ const StorageStatusPanel: React.FC<{
       ? ((createdCount / maxVfolderCount) * 100)?.toFixed(2)
       : 0
   ) as number;
-
-  return (
-    <Card size="small" title={t('data.StorageStatus')}>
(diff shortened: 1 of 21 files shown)


react/src/components/StorageStatusPanel.tsx
```diff
@@ -170,11 +170,12 @@ const StorageStatusPanel: React.FC<{
       ? ((createdCount / maxVfolderCount) * 100)?.toFixed(2)
       : 0
   ) as number;
-
-  return (
-    <Card size="small" title={t('data.StorageStatus')}>
-      <Descriptions bordered column={columnSetting} size="small">
-        <Descriptions.Item label={t('data.NumberOfFolders')}>
+  const descriptionItem: DescriptionsProps['items'] = [
```
yomybaby: Please rename `descriptionItem` to `descriptionItems` because it's array.

react/src/components/StorageStatusPanel.tsx
```diff
@@ -170,11 +170,12 @@ const StorageStatusPanel: React.FC<{
...
+    {
+      key: '1',
```
yomybaby: Please change keys to more contextual names. Using `1` can easily lead to duplication and can be difficult to manage if ordering is required.

yomybaby (changes requested):

lizable (approved):
LGTM.

yomybaby (approved):
LGTM

PR #2262, opened 2024-03-15
fix: modify husky config to format both lit and react components

agatha197:
Original pre-commit doesn't affect the react components.
Fix the pre-commit command to format both lit(web) components and react components.

- [ ] Documentation


ironAiken2 (approved):
LGTM!

PR #2264, opened 2024-03-16
revert: lint-staged setting of #2262

agatha197:
This reverts part of commit 90d912c.

- [ ] Documentation


yomybaby (approved):
I just pushed some commits to match file extensions where the Prettier command was applied.

LGTM!

PR #2265, opened 2024-03-18
feat: apply react managing apps modal into web component

### LINKED ISSUE #2261: bug: modifications to the image in the environment are not applied.

Linux (x86-64)

### Backend.AI version

23.09.10rc3

### Describe the bug

When I press the control button on the image tab in the environment, I can't save the modification in the popup (Edit limit value, Manage app).

image

### To Reproduce

1. Go to environment page
2. Select any image and edit it with the control buttons
3. No modifications are applied

### Expected Behavior

_No response_

### Anything else?

and also apply `ManageAppsModal` into `backend-ai-envrionment-list`.

ironAiken2:
> [!CAUTION]
> I've checked compatibility with version under 23.09.11, but some changes will be required once the [Backend.ai Core #2028 PR](https://github.com/lablup/backend.ai/pull/2028/files) PR is released.

> PR content:  replacing None in the validation code with Undefined [conf](
https://github.com/lablup/backend.ai/blob/5d69e7291ee386c7aff46c492dcd2e723c82ec2a/src/ai/backend/manager/models/image.py#L1181)

### Feature
- Minor bug-fix in mwc-slider.
- Apply `ManageAppsModal` into `backend-ai-environment-list`.
   - Includes validation for each field. Validation results are displayed in the space to the left of the confirm button.
- Modified to use ImageRow in mgr DB instead of sending modification requests to ETCD. (Currently, we have no logic to reflect ETCD changes).
- After modifying the service port, reload image values.

### Modify App Modal
|modal|validation|
|---|---|
|image|image|

### Modify Resource Limit Modal
|modal||
|---|---|
|image||

- [ ] Documentation


src/components/backend-ai-environment-list.ts
```diff
@@ -1916,6 +1691,18 @@ export default class BackendAIEnvironmentList extends BackendAIPage {
           ></mwc-button>
         </div>
       </backend-ai-dialog>
+      ${this.openManageAppModal
```
yomybaby: This pattern of managing the open state is not ideal because it does not show the closing transition.

react/src/components/ManageAppsModal.tsx
```diff
@@ -1,101 +1,210 @@
...
-        <Flex direction="row" justify="between">
-          <Button type="text" danger>
-            {t('button.Reset')}
-          </Button>
-          <Button type="primary" icon={<CheckOutlined />}>
-            {t('button.Save')}
-          </Button>
-        </Flex>,
-      ]}
+      footer={
```
yomybaby: You can implement without using a footer if there are only two buttons. Please change them to "Cancel" and "Save".

yomybaby (changes requested):

agatha197 (dismissed):
1. make port type to number to modify value with direction keys
2. changes are not reflected.
3. can you change the bottom border color to `token.colorBorder`? it's too bright.
image

src/components/backend-ai-environment-list.ts
ironAiken2: For the closing transition, I modified the web component's open state change to occur after the modal is closed.

react/src/components/ManageAppsModal.tsx
ironAiken2: Since the footer shows the validation result of each field, so only the buttons in the footer used antd's default modal buttons.

yomybaby (changes requested):
Adding this filter causes the modal to display information of a completely different image, rather than the one currently clicked.

image

react/src/components/ManageAppsModal.tsx
```diff
@@ -1,101 +1,256 @@
...
+        .join(',');
+
+      const INPUT: KVPairInput[] = [];
+      Object.entries(image.labels).forEach(([key, value]) => {
+        if (key.includes('service-ports')) {
+          INPUT.push({ key: key, value: values });
+        } else {
+          INPUT.push({ key: key, value: value?.toString() || '' });
+        }
+      });
```
agatha197: How about using lodash?

```suggestion
const INPUT = _.map(image.labels, (value, key) => {
  if (key.includes('service-ports')) {
    return { key, value: values };
  } else {
    return { key, value: _.toString(value) || '' };
  }
});
```

react/src/components/ManageAppsModal.tsx
```diff
@@ -1,101 +1,256 @@
...
 
   return (
     <BAIModal
+      destroyOnClose
       open={open}
-      onCancel={() => {
-        dispatchEvent('cancel', null);
-      }}
-      centered
+      onOk={handleOnclick}
```
agatha197: Please rename like this :)
```suggestion
      onOk={handleOnClick}
```

react/src/components/ManageAppsModal.tsx
```diff
@@ -1,101 +1,256 @@
...
-                    <Space.Compact block key={field.key}>
-                      <Form.Item
-                        {...field}
-                        name={[field.name, 'app']}
-                        rules={[
-                          {
-                            required: true,
-                            message: t('environment.AppNameMustNotBeEmpty'),
+              <Flex direction="column" style={{ width: '100%' }}>
+                {fields.map((field, index) => (
```
agatha197: Please use lodash when you use `map`

react/src/components/ManageImageResourceLimitModal.tsx
```diff
@@ -0,0 +1,150 @@
+import { useSuspendedBackendaiClient } from '../hooks';
+import BAIModal, { BAIModalProps } from './BAIModal';
+import { useWebComponentInfo } from './DefaultProviders';
+import ImageResourceFormItem from './ImageResourceFormItem';
+import { ManageImageResourceLimitModalMutation } from './__generated__/ManageImageResourceLimitModalMutation.graphql';
+import { ResourceLimitInput } from './__generated__/ManageImageResourceModalMutation.graphql';
```
ironAiken2, line 6: This code will not be auto-generated because it does not fit the GQL auto-generated code rules.
```typescript
import {
  ManageImageResourceLimitModalMutation,
  ResourceLimitInput,
} from './__generated__/ManageImageResourceLimitModalMutation.graphql';
```

yomybaby: It follows that rule. I have already renamed the component. Please check again.

PR #2272, opened 2024-03-25
feat: 'make clean' removes build files in the react folder

### LINKED ISSUE #2271: bug: `make clean` cli does not delete react build files

Linux (x86-64)

### Backend.AI version

23.09.10rc3

### Describe the bug

Packaging is executed with the existing success file even if the build fails because there is no command to clean the react-side buildfile in the make clean command.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

ironAiken2:
### Feature:
- add cli to remove build files in react folder.

- [ ] Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,3 +207,4 @@ i18n:
 clean:
 	@cd app;	rm -rf ./backend*; rm -rf ./Backend*
 	@cd build;rm -rf ./unbundle ./bundle ./rollup ./electron-app
+	@cd react/build; rm -rf ./*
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -207,3 +207,4 @@ i18n:
 clean:
 	@cd app;	rm -rf ./backend*; rm -rf ./Backend*
 	@cd build;rm -rf ./unbundle ./bundle ./rollup ./electron-app
+	@cd react/build; rm -rf ./*
```
yomybaby:
```suggestion
	@rm -rf ./react/build
```
How about this? Using `rm -rf ./*` has a higher risk of accidentally deleting files or directories compared to `rm -rf ./react/build`.

yomybaby (changes requested):

Makefile
ironAiken2: I think it's better to set the exact path to reduce the risk of deletion. Fixed.

yomybaby (approved):

PR #2268, opened 2024-03-23
update: node.js packages

inureyes:
> [!NOTE]  
> Need to reinstall node packages.
> If you have trouble when running `npm run i`, remove node_modules folder and try again.

**Checklist:** (if applicable)

- [ ] Mention to the original issue
- [ ] Documentation
- [ ] Minium required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [ ] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

PR #2280, opened 2024-03-28
style: fix the broken select UI of the new storage folder dialog

agatha197:
follows #2268 

Before
image

After
image

- [ ] Documentation


patches/@material+mwc-select+0.27.0.patch
```diff
@@ -1,23 +1,25 @@
...
 +        }
 +      </style>
        <div
            class="mdc-select ${classMap(classes)}">
          <input
-@@ -247,7 +255,7 @@ export class SelectBase extends FormElement {
+@@ -247,7 +257,7 @@ export class SelectBase extends FormElement {
          if (!this.icon) {
              return nothing;
          }
```
yomybaby: Without this patch file, it works properly with `backend-ai-data-view.ts` changes. Is this paths is required? @agatha197

agatha197: You are right. I reverted it.

yomybaby (approved):
LGTM

PR #2285, opened 2024-04-01
feat: display `Quota per storage volume` info if quota is available

agatha197:
fix: remove `quotaSupportStorageBackends` from backend-ai-data-view because it is not used.
resolves https://github.com/lablup/giftbox/issues/642
follows #2180

## Descriptions
Problem: A problem that does not show `Quota per storage volume` even though a storage supports quota.
Cause: The value of `selectedVolumeInfo` changed to a string (single string of hostname) instead of an object.
Resolution: Set the value of `selectedVolumeInfo` as the second factor of `onChange` in `StorageSelect`

## How to test
- test endpoint: http://10.82.230.101:8090

## Checklist 
- [ ] If selected storage supports quota, can you see the `Quota per storage volume` ?
- [ ] Highlighting/Searching storage host is available?

- [ ] Documentation


yomybaby (approved):
I push a commit to fix TypeScript type definition for `onChange`. 
After that, LGTM!

PR #2286, opened 2024-04-01
fix: regression of listing connected agent with null-check on live_stat from response

lizable:
Sometimes there has a chance of receiving NULL value in `live_stat` in agent status, but since now It's been treated as if always has value, which caused runtime-error. Therefore I added conditional statement so that parsing `live_stat` value is only available when it's not NULL.

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-agent-list.ts
+++ b/src/components/backend-ai-agent-list.ts
@@ -1225,7 +1225,7 @@ export default class BackendAIAgentList extends BackendAIPage {
         cpu_util: { capacity: 0, current: 0, ratio: 0 },
         mem_util: { capacity: 0, current: 0, ratio: 0 },
       };
-      if (rowData.item.live_stat && rowData.item.live_stat.node.cuda_util) {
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-agent-list.ts
```diff
@@ -1225,7 +1225,7 @@ export default class BackendAIAgentList extends BackendAIPage {
         cpu_util: { capacity: 0, current: 0, ratio: 0 },
         mem_util: { capacity: 0, current: 0, ratio: 0 },
       };
-      if (rowData.item.live_stat.node.cuda_util) {
+      if (rowData.item.live_stat && rowData.item.live_stat.node.cuda_util) {
```
yomybaby:
```suggestion
      if (rowData.item.live_stat?.node?.cuda_util) {
```
How about this? @lizable

yomybaby (changes requested):

yomybaby (approved):
LGTM!

PR #2289, opened 2024-04-02
feat: destroy-all function in stacked notification

### LINKED ISSUE #2205: Add feature for closing all notification at once
### Main idea

If notifications stack too many, it is hard to close all one by one. Let's add a more icon next to the `See Detail` button, and add a menu to close all notifications when only notifications are stacked.
image

### Alternative ideas

_No response_

### Anything else?

_No response_

### LINKED ISSUE #2281: feat: the "close all" button in Notification
We need a "Close All" button in Notification.
- "Close All" Button is shown only when it stacked
- Consider [`maxCount` option](https://ant.design/components/notification#api)

Image

ironAiken2:
### Feature
- you can delete all notification in active, by clicking clear button in notification drawer.
image

- [ ] Documentation

### DIFF
--- a/react/src/components/BAINotificationItem.tsx
+++ b/react/src/components/BAINotificationItem.tsx
@@ -5,7 +5,7 @@ import {
   ClockCircleOutlined,
   CloseCircleOutlined,
 } from '@ant-design/icons';
-import { Card, List, Progress, Typography, theme } from 'antd';
+import { Button, Card, List, Progress, Typography, theme } from 'antd';
 import dayjs from 'dayjs';
 import _ from 'lodash';
 import React, { useState } from 'react';
@@ -18,7 +18,8 @@ const BAINotificationItem: React.FC<{
     notification: NotificationState,
   ) => void;
   showDate?: boolean;
-}> = ({ notification, onClickAction, showDate }) => {
+  destroyAll?: () => void | undefined;
+}> = ({ notification, onClickAction, showDate, destroyAll }) => {
   const { t } = useTranslation();
   const { token } = theme.useToken();
   const [showExtraDescription, setShowExtraDescription] = useState(false);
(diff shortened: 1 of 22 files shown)


react/src/components/BAINotificationItem.tsx
agatha197, line 1: How about displaying `Destroy All` button when only `destroyAll` is not undefined?

agatha197 (changes requested):

react/src/components/BAINotificationItem.tsx
ironAiken2: @agatha197 Resolved.

agatha197 (approved):
LGTM

react/src/components/BAINotificationItem.tsx
```diff
@@ -18,7 +18,8 @@ const BAINotificationItem: React.FC<{
     notification: NotificationState,
   ) => void;
   showDate?: boolean;
-}> = ({ notification, onClickAction, showDate }) => {
+  destroyAll?: () => void | undefined;
```
yomybaby: This `prop` is function type, it's better to use `onClickDestroyAll` name patter

resources/i18n/en.json
```diff
@@ -1276,7 +1276,8 @@
     "Notifications": "Notifications",
     "NoNotification": "No Notification.",
     "ClearNotifications": "Clear Notifications",
-    "AreYouSureToClearAllNotifications": "Are you sure to clear all notifications?"
+    "AreYouSureToClearAllNotifications": "Are you sure to clear all notifications?",
+    "DestroyAll": "Destroy All"
```
yomybaby: I know the API name is `destroy`, How about using `Close` instead of `Destroy` for a user? @agatha197  @ironAiken2

yomybaby (changes requested):
The position of the "Destroy all" button is not appropriate.

image

resources/i18n/en.json
agatha197: image  

How about clear? lol

ironAiken2: 'clear' is fine, but I think it feels more like 'closing' an open popup, rather than clearing space, so 'close' seems more better. How about it?

react/src/components/BAINotificationItem.tsx
ironAiken2: Resolved

resources/i18n/en.json
ironAiken2: Resolved

ironAiken2:
If antd config provider is provided ([codes](https://github.com/lablup/backend.ai-webui/blob/bd357fa22737901c84a91a270d7c3d8e9f277bf8/react/src/components/DefaultProviders.tsx#L158)) and configured via notification.config(), it seems to refer to that config when opening the notification, but it doesn't work properly. 

Related links : https://github.com/ant-design/ant-design/issues/42799

yomybaby (approved):
LGTM!

PR #2294, opened 2024-04-03
feat: display accelerator key instead of empty string

agatha197:
follows #1651 
- Add IPU translations.

Before
image

After
image
If the translation string is not included, it shows the key. (For example, IPU)
image

- [ ] Documentation


lizable (approved):
LGTM.

PR #2295, opened 2024-04-04
fix: disable download button when download perm is not allowed

lizable:
This PR disallows downloading files by clicking download button in folder explorer when the permissions of corresponding storage host doesn't include `download-file`.

> ⚠️ NOTE: You should refresh the page when permissions are updated.

Please check,
- [x] Download file/folder in folder-explorer only if `download-file` permission is allowed.
- [x] Not allowed to download file/folder when there's no `download-file` permission in corresponding storage host.

### Screenshot(s) of when `download-file` permission is not allowed
| After | Before |
|------|--------|
| Screenshot 2024-04-04 at 12 54 37 PM | Screenshot 2024-04-04 at 11 23 11 AM |

- [ ] Documentation


agatha197 (approved):
LGTM

PR #2299, opened 2024-04-04
Fix: TOTP activation modal not opening for non-TOTP users with `forceTotp`

yomybaby:
This PR introduces `ForceTOTPChecker` to open `TOTPActivateModal` for users who have not activated TOTP when the `forceTotp` configuration is true.
This part was missing during the transition to React.

image

### How to test
- Install the TOTP plugin.
- Set the following config.toml variables:
  ```
  [general]
  enable2FA = true
  force2FA = true
  ```
- Try to log in using a user who has not activated TOTP.
- You will see the TOTPActivateModal (refer to the screenshot above) right after logging in.

- [ ] Documentation


agatha197 (approved):
LGTM

PR #2301, opened 2024-04-05
misc: update description of type of login account in login panel

lizable:
Sometimes users login their own ID account instead of email address, which may confuse them in login panel, which only shows "Login with E-mail" or "Login with IAM". This PR will fix this situation by adding "Username" to description in there.

| After | Before |
|------|--------|
| Screenshot 2024-04-05 at 10 41 39 AM | Screenshot 2024-04-05 at 10 21 09 AM |

- [ ] Documentation


yomybaby (approved):
LGTM

PR #2302, opened 2024-04-05
fix: created count in `StorageStatusPanel`

yomybaby:
This PR fix the bug related to the count of created vFolders in `StorageStatusPanel`

Before this PR,
- the count does not include vfolders in the trash bin with 'delete-pending' status.
- the count is not updated right after creating/deleting(forever) vfolders.

- [ ] Documentation


agatha197 (approved):
LGTM

PR #2306, opened 2024-04-05
fix: vFolder field has incorrect value in editing

yomybaby:
In Service endpoint editing,
- The vFolder field value is always the first option.

This PR 
- displays the correct value as read-only.
- uses `VFolderLazyView` and show Skeleton when `VFolderLazyView` is suspended.

| Before | After |
|--------|--------|
| image | image | 

- [ ] Documentation


lizable (approved):
Works fine. LGTM!

PR #2307, opened 2024-04-05
fix: session list fetch error

agatha197:
fix: fetch `image_object` when supporting `per-user-image` to fix the error that occurs fetching the session list in 23.09

- [ ] Documentation


yomybaby (approved):
LGTM

PR #2322, opened 2024-04-08
feat: set color for text 'only'

### LINKED ISSUE #2309: Bug: VFolder invitation permission icon is not visible in the light mode

Linux (x86-64)

### Backend.AI version

24.03.0-beta.1

### Describe the bug

The "O" mark is not visible at all in light mode.
I think it'd be better to change the color in dark mode as well.

### To Reproduce

After creating a VFolder of the "User" type, share it with a specific user as "RO" permission. 
Then, log out and log in with that account.

### Expected Behavior

_No response_

### Anything else?

_No response_

ironAiken2:
### Feature
|light|dark|
|---|---|
|image|image|

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-summary-view.ts
+++ b/src/components/backend-ai-summary-view.ts
@@ -604,9 +604,6 @@ export default class BackendAISummary extends BackendAIPage {
                                     ></lablup-shields>
                                   `,
                                 )}
-                                ${[...invitation.perm].map((c) =>
-                                  console.log(c),
-                                )}
                               </div>
                               <div
                                 style="margin:15px auto;"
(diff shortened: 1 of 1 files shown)


src/components/backend-ai-summary-view.ts
```diff
@@ -593,14 +593,20 @@ export default class BackendAISummary extends BackendAIPage {
...
+                                        'yellow',
+                                      ][['r', 'w', 'd', 'o'].indexOf(c)]}"
                                       description="${c.toUpperCase()}"
                                       ui="flat"
                                     ></lablup-shields>
                                   `,
                                 )}
+                                ${[...invitation.perm].map((c) =>
+                                  console.log(c),
+                                )}
```
agatha197:
```suggestion
```

agatha197 (changes requested):

agatha197 (approved):
LGTM

src/components/backend-ai-summary-view.ts
ironAiken2: Oops... Reflected!

PR #2324, opened 2024-04-08
fix: check isNaN before set value of DynamicUnitInputNumber

yomybaby:
When you input `mem` like this:
image

You can see the error like this:
image

- [ ] Documentation

### DIFF
--- a/react/src/components/DynamicUnitInputNumber.tsx
+++ b/react/src/components/DynamicUnitInputNumber.tsx
@@ -72,6 +72,14 @@ const DynamicUnitInputNumber: React.FC<DynamicUnitInputNumberProps> = ({
       onBlur={() => {
         if (_.isNumber(roundStep) && roundStep > 0) {
           const decimalCount = roundStep.toString().split('.')[1]?.length || 0;
+          if (
+            isNaN(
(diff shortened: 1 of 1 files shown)


yomybaby:
To monkey patch to apply this PR to 24.03.1,  replace 
```
onBlur:()=>{if(p().isNumber(u)&&u>0){var e,n;const i=(null===(e=u.toString().split(".")[1])||void 0===e?void 0:e.length)||0;x("".concat((Math.round(p().toNumber((null===(n=C.current)||void 0===n?void 0:n.value)||"0")/u)*u).toFixed(i)).concat(A))}}
``` 
to 
```
onBlur:()=>{if(p().isNumber(u)&&u>0){var e,n;const i=(null===(e=u.toString().split(".")[1])||void 0===e?void 0:e.length)||0;if(isNaN(Math.round(p().toNumber((null === (n = C.current) || void 0 === n ? void 0 : n.value) ||"0") / u) * u)) return;x("".concat((Math.round(p().toNumber((null===(n=C.current)||void 0===n?void 0:n.value)||"0")/u)*u).toFixed(i)).concat(A))}}
``` 
of `491.f714d492.chunk.js`

lizable (approved):
Works fine. LGTM.

PR #2325, opened 2024-04-09
style: add css variable to vaadin grid for plugin page

agatha197:
For plugin pages that use `backend-ai-storage-proxy-list.ts`, the grid size is too large. So I added CSS variable to `vaadin-grid`. The `--list-height` is already used in many plugin pages.

- [ ] Documentation


inureyes (approved):
LGTM

PR #2330, opened 2024-04-11
fix: unkown scailing group error when user change the project

yomybaby:
This PR includes the following changes:
- Ignore the `/resource-preset` response error.
- Clear the value of `ResourceGroupSelect` when the value is not in the options.
- Validate and handle validation errors for no resource group.
- change the position of the resource item in the Neo Session Launcher preview step.

| Before | After |
|--------|--------|
| image | image | 

This PR affects the pages that use `ResourceGroupSelect` and `ResourceAllocationFormItems`, such as the Neo Session Launcher, Service Launcher, and Resource Monitor.

- [ ] Documentation


lizable (approved):
Though user need to select resource group explicitly after changing project, It works fine.

PR #2332, opened 2024-04-11
feat: add MyEnvironment page to manage per user commit images

agatha197:
> [!Important]
>	should be merge after https://github.com/lablup/backend.ai/pull/2010 is reflected.

- resolves https://github.com/lablup/giftbox/issues/653
- related PR: https://github.com/lablup/backend.ai/pull/2010

## Checklist
- [ ] New page: `My Environments` (> 24.03.0). If the core version is under 24.03.0 and user tries to access the `my-environment` page through the URL, it redirects to error page.
- [ ] Display the list of session committed images
- [ ] Users can delete each image by clicking the trash icon button.
- [ ] Users can filter the display columns.

## Screenshots
- list
	image

- style: border color changed
	image
	image

- style: add `cyan` color to `lablup-shields` and display constraint with Customized tag.
image

- style: use token font color to Session Name, Session ID for dark mode supporting
image

- [ ] Documentation


react/src/pages/MyEnvironmentPage.tsx
```diff
@@ -0,0 +1,355 @@
...
+                displayedColumnKeys?.includes(_.toString(column.key)),
+              )}
+              dataSource={processedImages as CustomizedImages[]}
+              scroll={{ x: 'max-content' }}
+            />
+            <Flex
+              justify="end"
+              style={{
+                padding: token.paddingXXS,
+              }}
```
yomybaby, line 299: If no pagination, please add `pagination={false}` to Table.

react/src/pages/MyEnvironmentPage.tsx
```diff
@@ -0,0 +1,355 @@
...
+        <Flex direction="row" align="stretch">
+          <Popconfirm
+            title={t('dialog.ask.DoYouWantToProceed')}
+            description={t('dialog.warning.CannotBeUndone')}
+            okType="danger"
+            onConfirm={() => {
+              if (row?.id) {
+                commitForgetAndUntag({
+                  variables: {
+                    image_id: row.id,
```
yomybaby: Please set Popconfirm's `okText` prop to "Delete".

react/src/components/MainLayout/WebUISider.tsx
```diff
@@ -85,6 +87,11 @@ const WebUISider: React.FC<WebUISiderProps> = (props) => {
       icon: <CloudUploadOutlined />,
       key: 'data',
     },
+    supportPendingSessionCount && {
```
yomybaby: `supportPendingSessionCount` is not related to user-commited image, right? It's better to add a support key for user-commited image support.

yomybaby (changes requested):

react/src/components/MainLayout/WebUISider.tsx
agatha197: Oops, I was confused with another PR. Rename completed!

react/src/pages/MyEnvironmentPage.tsx
```diff
@@ -0,0 +1,321 @@
...
+            okText={t('button.Delete')}
+            onConfirm={() => {
+              console.log(row.key);
+              if (row.key) {
+                commitForgetAndUntag({
+                  variables: {
+                    image_id: row.key,
+                    id: row.key,
+                  },
+                  onCompleted(data) {
```
yomybaby: You also need to handle `errors` in `onCompleted(data, errors)`.

react/src/pages/MyEnvironmentPage.tsx
```diff
@@ -0,0 +1,321 @@
...
+          : 'network-only',
+      fetchKey: myEnvironmentFetchKey,
+    },
+  );
+
+  const [commitForgetAndUntag, isInflightForgetAndUntag] =
+    useMutation<MyEnvironmentPageForgetAndUntagMutation>(graphql`
+      mutation MyEnvironmentPageForgetAndUntagMutation(
+        $image_id: String!
+        $id: String!
```
yomybaby: `$id` and `$image_id` are the same. We can use just one of them.

react/src/pages/MyEnvironmentPage.tsx
```diff
@@ -0,0 +1,321 @@
...
+      key,
+      imageFullName,
+      namespace,
+      lang,
+      baseversion,
+      baseimage,
+      constraint,
+    };
+  });
+
```
yomybaby: In most cases in Relay, it's better to use GraphQL data without generating a new type. Transformation functions like `getImageFullName` can be called in React components or in the `render` function of columns.

yomybaby (changes requested):

react/src/hooks/index.tsx
```diff
@@ -178,6 +180,12 @@ interface ImageMetadata {
   }[];
 }
 
+export const getImageFullName = (image: Image | CommittedImage) => {
+  return image
+    ? `${image.registry}/${image.name}:${image.tag}@${image.architecture}`
+    : undefined;
+};
```
yomybaby: Let's move this function to `helper/index.tsx` because this is not a react hook.

ironAiken2:
The sort logic in `constraint` is not working.
image

yomybaby (approved):
LGTM 💪

PR #2337, opened 2024-04-14
fix: regression of applying default session environment

### LINKED ISSUE #2336: Cannot apply default session environment in Neo session launcher

Linux (x86-64)

### Backend.AI version

24.09.0dev1

### Describe the bug

Compare to the "Classic" version of session launcher, It seems like "Neo" session launcher doesn't support ["default session environment"](https://github.com/lablup/backend.ai-webui/blob/16974db484321c6e9bd25dc2e1fceb2dc44ecff4/config.toml.sample#L4), which can be applied in config file.

### To Reproduce

1. Set `defaultSessionEnvironment` value to certain value in config file. (e.g. `"cr.backend.ai/stable/python-tensorflow:2.7-py38-cuda11.3"`)
2. Go to "Summary" or "Session" page.
3. Click `START` button. (If session launcher is set to "Classic", please select "Neo" version)
4. Directs to "Session launcher page".
5. Click "Environment & Resource Allocation" section in right pane
6. See the error.

### Expected Behavior

If defaultSessionEnvironment is set, then environment selection initially selects environment and version section according to its value.

### Anything else?

_No response_

lizable:
This PR resolves #2336.

- [ ] Documentation

### DIFF
--- a/react/src/components/ImageEnvironmentSelectFormItems.tsx
+++ b/react/src/components/ImageEnvironmentSelectFormItems.tsx
@@ -216,6 +216,15 @@ const ImageEnvironmentSelectFormItems: React.FC<
     } else if (baiClient._config.allow_manual_image_name_for_session) {
       // if no image is available, only set manual if it's allowed
       form.setFieldValue(['environments', 'manual'], version);
+    } else {
+      form.setFieldsValue({
+        environments: {
+          environment: nextEnvironment.environmentName,
+          version: getImageFullName(nextImage),
+          image: nextImage,
+          customizedTag: customizedImageTag ?? undefined,
+        },
+      });
     }
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/react/src/components/ImageEnvironmentSelectFormItems.tsx
+++ b/react/src/components/ImageEnvironmentSelectFormItems.tsx
@@ -151,8 +151,11 @@ const ImageEnvironmentSelectFormItems: React.FC<
       | ImageGroup['environmentGroups'][0]
       | undefined;
     let matchedImageByVersion: Image | undefined;
-    const version = form.getFieldValue('environments')?.version;
-
+    let version = form.getFieldValue('environments')?.version;
+    // FIXME: manually add architecture based on amd64
+    if (version && version.indexOf('@') < 0) {
+      version += '@x86_64';
+    }
     version &&
       _.find(imageGroups, (group) => {
         matchedEnvironmentByVersion = _.find(
--- a/react/src/components/ImageEnvironmentSelectFormItems.tsx
+++ b/react/src/components/ImageEnvironmentSelectFormItems.tsx
@@ -216,15 +216,6 @@ const ImageEnvironmentSelectFormItems: React.FC<
     } else if (baiClient._config.allow_manual_image_name_for_session) {
       // if no image is available, only set manual if it's allowed
       form.setFieldValue(['environments', 'manual'], version);
-    } else {
-      form.setFieldsValue({
-        environments: {
-          environment: nextEnvironment.environmentName,
-          version: getImageFullName(nextImage),
-          image: nextImage,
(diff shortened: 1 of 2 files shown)


react/src/pages/SessionLauncherPage.tsx
```diff
@@ -639,7 +639,14 @@ const SessionLauncherPage = () => {
...
               layout="vertical"
               requiredMark="optional"
-              initialValues={INITIAL_FORM_VALUES}
+              initialValues={_.merge(
+                INITIAL_FORM_VALUES,
+                baiClient._config?.default_session_environment && {
+                  environments: {
+                    version: baiClient._config?.default_session_environment,
+                  },
+                },
```
yomybaby: `INITIAL_FORM_VALUES` is used [to determine whether to apply form values](https://github.com/lablup/backend.ai-webui/pull/2337/files#diff-cb5aa3700f469823d540634a1acf2796ffcfa96416d6fa9333bfab0ff658b2c1R232-R243) using query parameters. Therefore, `INITIAL_FORM_VALUES` should be the initial values for a `<Form/>` without modification. Let's create a function outside of the component to handle the default image of `config.toml`.
`getInitialFormValues()` instead of `_.merge(INITIAL_FORM_VALUES...`

yomybaby (changes requested):

react/src/pages/SessionLauncherPage.tsx
lizable: In order to use "config" settings from `config.toml`, I need to use baiClient hook, which can be called only inside of React function. Therefore, instead of creating a function like `getInitialFormValues()` I would like to rearrange `INITIAL_FORM_VALUES` inside React function with default_session_environment value.

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1216,7 +1221,7 @@ const SessionLauncherPage = () => {
                             <Flex direction="row">
                               {form.getFieldValue('environments')?.manual ? (
                                 <Typography.Text copyable code>
```
lizable, line 1223: I also found that there's a cosmetic issue when showing manual image in the last step of Neo session launcher. I already fixed it by wrapping with bracket. See the difference in the screenshot below:

### Screenshot(s)
| After | Before |
|------|--------|
| Screenshot 2024-04-16 at 12 42 23 PM | Screenshot 2024-04-16 at 12 42 34 PM |

react/src/components/ImageEnvironmentSelectFormItems.tsx
```diff
@@ -213,6 +216,15 @@ const ImageEnvironmentSelectFormItems: React.FC<
     } else if (baiClient._config.allow_manual_image_name_for_session) {
       // if no image is available, only set manual if it's allowed
       form.setFieldValue(['environments', 'manual'], version);
```
lizable: Also I found that there's an another case when the defaultSessionEnvironment is set, but it's not installed or available for now, and the configuration doesn't allow manual image. We didn't thought the case, but I think it's possible. Therefore, when it comes to having a situation like this, we need to apply the first environment and version from the list as I implemented in here.

react/src/pages/SessionLauncherPage.tsx
yomybaby: 🦅👀👍

react/src/components/ImageEnvironmentSelectFormItems.tsx
yomybaby: But this case already handled in https://github.com/lablup/backend.ai-webui/blob/ef0adef4e521a96dbb2b0dd6a4c3eb576a03e3a8/react/src/components/ImageEnvironmentSelectFormItems.tsx#L195-L209.
I pushed a commit to remove a duplicated case.

yomybaby (approved):
LGTM

PR #2339, opened 2024-04-15
fix: `_viewStateChanged` is called for same `active` value in `attributeChangedCallback`

yomybaby:
This PR avoid calling `_viewStateChanged` for same `active` value in `attributeChangedCallback`.

Before the PR, every component extending the BackendAIPage component runs `_viewStateChanged` twice with the same active value, leading to multiple duplicated remote requests.

- [ ] Documentation


agatha197 (approved):
LGTM 🚀

PR #2338, opened 2024-04-15
style: label wrapping, font resizing and numeric token formatting in login panel

ironAiken2:
### Features
- Make modal's label text automatically wrap when it gets long.
- Adjust the font size of the MWC button.
- If the value type of antd token value is numeric, append 'px' after it. (e.g. fontSize)

image

**Checklist:** (if applicable)

- [ ] Mention to the original issue
- [ ] Documentation
- [ ] Minium required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [ ] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

PR #2344, opened 2024-04-16
style: remove `px` when calculating margin of WebUIHeader wrapper

agatha197:
`--token-xxx` has already px from #2338, I removed the duplicated definition.

- [ ] Documentation


ironAiken2 (approved):
LGTM

PR #2345, opened 2024-04-16
fix: use `defaultSessionEnvironment` as a initial `environments.environment`

### LINKED ISSUE #2336: Cannot apply default session environment in Neo session launcher

yomybaby:
This PR follows up on #2337 and resolves #2336.

After this PR, the `defaultSessionEnvironment` will be used as the initial `environments.environment` form value for Neo Session Launcher.

- [ ] Documentation

### DIFF
--- a/react/src/components/ImageEnvironmentSelectFormItems.tsx
+++ b/react/src/components/ImageEnvironmentSelectFormItems.tsx
@@ -172,7 +172,20 @@ const ImageEnvironmentSelectFormItems: React.FC<
     if (matchedEnvironmentByVersion) {
       nextEnvironment = matchedEnvironmentByVersion;
       nextImage = matchedImageByVersion;
-    } else {
+    } else if (form.getFieldValue(['environments', 'environment'])) {
+      _.find(imageGroups, (group) => {
+        nextEnvironment = _.find(group.environmentGroups, (environment) => {
+          return (
+            environment.environmentName ===
+            form.getFieldValue(['environments', 'environment'])
+          );
+        });
+        nextImage = nextEnvironment?.images[0];
+        return !!nextEnvironment;
+      });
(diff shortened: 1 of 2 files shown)


react/src/components/ImageEnvironmentSelectFormItems.tsx
```diff
@@ -242,7 +255,7 @@ const ImageEnvironmentSelectFormItems: React.FC<
                   // metadata?.imageInfo[
                   //   getImageMeta(getImageFullName(image) || "").key
                   // ]?.name || image?.name
-                  image?.name
+                  image?.registry + ':' + image?.name
```
lizable: I think this `:` should be changed to `/`.

lizable (approved):
Works fine on both web browser and app.
LGTM.

PR #2346, opened 2024-04-16
hotfix: enable non-accelerator preset when accelerator image

yomybaby:
- If there is no accelerator requirement in the image:
  - Exclude accelerators presets.
  - Disable the accelerator input field.
  
- If there is an accelerator requirement in the image:
  - (After this PR) **Show non-accelerator presets also.** 
  - Enable the accelerator input field.
  
 

- [ ] Documentation


lizable (approved):
Now Resource Preset Select in Neo session launcher works as same as "Classic" one.
LGTM.

PR #2334, opened 2024-04-12
fix: vFolder form item validation in Neo session launcher

yomybaby:
This PR contains changes related to vFolder validation in the Neo session launcher. Please review the commits one by one.
- remove console.log warning about `<form> cannot appear as a descendant of <form>`
- add missing name rule for vFolder alias map for preview step
- [exclude auto-mounted vFolders from the vFolder list in Neo session launcer](https://github.com/lablup/backend.ai-webui/commit/d6c8308bb8d032d5ff7ed4076027196cd63650d9)
- validate form after step changed using `useEffect` and set the last validation error time to re-render.

**Checklist:** (if applicable)

- [ ] Mention to the original issue
- [ ] Documentation
- [ ] Minium required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [ ] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

PR #2359, opened 2024-04-19
fix: wrong vFolder validation result occasionally.

yomybaby:
This is related to https://github.com/lablup/backend.ai-webui/pull/2334. But it doesn't fix all bugs.

In the final step of Neo Session Launcher, the bug that shows an error in vFolder validation when there is none. 

How to reproduce:
It occurs occasionally (you might see this after several tries.)
-  Start with the form that has no validation error
-  Select one vfolder and type a alias name to `ㄱ`)
-  Go to final step.
- And go back to the Data vFolder and unselect vfolder and go to the final step.
- You can see the error sometimes.

https://github.com/lablup/backend.ai-webui/assets/621215/9c6723e1-3c00-4861-8211-b786787cc916

- [ ] Documentation


lizable (approved):
LGTM. I think it's been fixed.

PR #2363, opened 2024-04-23
fix: prevent delete in shared vfolder by permission

### LINKED ISSUE #2362: Need to disable "delete from trash bin" operation on deleted(moved-to-trash-bin) vfolder when user has only "read" permission

Linux (x86-64)

### Backend.AI version

24.09.0dev1

### Describe the bug

> NOTE: This issue is reported by @jopemachine.

When user has read permission from another user's vfolder and the sharer deleted vfolder after sharing, It becomes deletable in trash bin. It need to be fixed.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

ironAiken2:
### Feature
- Prevent delete/restore by users without delete permission.

### How to test
1. Share vfolders to other users with different permissions.
2. After the sharing is successful, log in as the owner user and move the folders to the trash bin.
3. log in as the shared user account, and make sure that the permanent delete and restore features are enabled according to your permissions. (Only accounts with the delete permission can be permanently deleted and restored.)

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-storage-list.ts
+++ b/src/components/backend-ai-storage-list.ts
@@ -2154,7 +2154,9 @@ export default class BackendAiStorageList extends BackendAIPage {
             text="${_t('data.folders.FolderInfo')}"
             position="top-start"
           ></vaadin-tooltip>
-          ${rowData.item.is_owner || rowData.item.permission === 'wd'
+          ${rowData.item.is_owner ||
+          this._hasPermission(rowData.item, 'd') ||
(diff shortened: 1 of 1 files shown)


yomybaby (changes requested):
Please add a "초대된 폴더에서 나가기" control button to the Trashbin tab list.

yomybaby (approved):
LGTM!

PR #2374, opened 2024-04-26
fix: preventing sharing of auto-mounted folders

### LINKED ISSUE #2373: Let's disable automount folder share button in VFolder list
### Main idea

AutoMounted folders are not shareable, but the icons are enabled. The server sends an error message when requested, but it would be better to disable it.

image

### Alternative ideas

_No response_

### Anything else?

_No response_

ironAiken2:
### Feature
`auto mounted folder` is a folder that name begins with '.' in the folder name. I added folder name validation to the `isSharingAllowed` validation logic in the control button lander logic.

- [ ] Documentation


lizable (approved):
LGTM.

PR #2376, opened 2024-04-26
fix: Change '/session' path to be used as sider's selectedKey 'job'

### LINKED ISSUE #2372: Bug: when launching a new session with Neo session launcher, the tab for sider is deselected

macOS (Apple Silicon)

### Backend.AI version

24.09.0dev1

### Describe the bug

When launching a new session with neo session laucher, navigating to the session/start query will deselect tabs in the sider.

|path: session|path: session/start|
|---|---|
|image|image|

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

ironAiken2:
### Why the bug occurred?
The sider selection key value for the session page of the component's initial render is 'job'. Since `neo session launcher` uses 'session/start' as the query path, it needs to be changed to recognize it as the 'job' key. 

### Feature 
If the first query path is 'session', change it to 'job' so that it can be recognized as the key of the sider.

- [ ] Documentation


yomybaby (approved):
LGTM

PR #2383, opened 2024-04-29
fix: remove comments on config values

### LINKED ISSUE #2382: bug: The toml parser library needs to be replaced.

macOS (Apple Silicon)

### Backend.AI version

24.09.0dev1

### Describe the bug

Summary page app download link doesn't work. I've figured out that there's a problem with the current TOML parsing process. Additionally, A syntax error occurs if the TOML file contains an empty string during preprocessing. [ref](https://github.com/lablup/backend.ai-webui/blob/9f4ce7c6ccd3b2b064399e8e97dc9d3b1d73c673/src/components/backend-ai-webui.ts#L661-L672)

Currently, webui is using the 'markty-toml' parser. This library currently has some issues with comment handling, which are not yet resolved. [github ref](https://github.com/Jonarod/markty-TOML/issues/15#issue-1318877785)

In addition to this, I've tested it myself and found that it doesn't load the value correctlyBecause of this, I think a replacement for the toml parser library is needed.

```typescript
const example = `
          [general]
          commentExample = "This is a comment."     # This is a comment.
          stringExample = "Hello, World!"           # Hello, World!
          booleanExample = true                     # true
          numberExample = 123                       # 123
          floatExample = 123.456                    # 123.456
        `;
```

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Anything else?

_No response_

JIRA Issue: FR-66

ironAiken2:
### Why the bug occurred
Inside the config.toml file, a description of each value is included as a comment. These comments were being read together to prevent the default value from being used even when no value was provided. 

### Feature
- Remove comments when reading config.toml values
- Adjust the alert margin at the top of the summary page

- [ ] Documentation

### DIFF
--- a/react/src/components/AnnouncementAlert.tsx
+++ b/react/src/components/AnnouncementAlert.tsx
@@ -21,8 +21,8 @@ const AnnouncementAlert: React.FC<Props> = ({ style, ...otherProps }) => {
     <Alert
       banner
       style={{
-        // alignItems: 'flex-start',
         // overflow: 'auto',
+        alignItems: 'center',
         ...style,
       }}
       // icon={
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -467,13 +467,16 @@ export default class BackendAILogin extends BackendAIPage {
   }
 
   private _getConfigValueByExists(parentsKey, valueObj: ConfigValueObject) {
+    const uncommentedValue =
+      typeof valueObj.value === 'string' && valueObj.value.includes('#')
+        ? valueObj.value.split('#')[0].trim()
+        : '';
     const defaultConditions: boolean =
       parentsKey === undefined ||
       valueObj.value === undefined ||
       typeof valueObj.value === 'undefined' ||
-      valueObj.value === '' ||
-      valueObj.value === '""' ||
-      valueObj.value === null;
+      ['', '""', null].includes(uncommentedValue);
+
     let extraConditions;
     switch (typeof valueObj.defaultValue) {
       case 'number':


react/src/components/AnnouncementAlert.tsx
```diff
@@ -21,8 +21,8 @@ const AnnouncementAlert: React.FC<Props> = ({ style, ...otherProps }) => {
     <Alert
       banner
       style={{
-        // alignItems: 'flex-start',
         // overflow: 'auto',
+        alignItems: 'center',
```
yomybaby, line 25: Please remove this. When the announcement has multiple lines, it's better to place a close button at the top instead of the middle.

src/components/backend-ai-login.ts
```diff
@@ -467,13 +467,16 @@ export default class BackendAILogin extends BackendAIPage {
   }
 
   private _getConfigValueByExists(parentsKey, valueObj: ConfigValueObject) {
+    const uncommentedValue =
+      typeof valueObj.value === 'string' && valueObj.value.includes('#')
+        ? valueObj.value.split('#')[0].trim()
+        : '';
```
yomybaby, line 473: This is not a good idea because the value can have a #. 
```
some="This#Value#has#"
```

yomybaby: If the current TOML parser does not support inline comments, please search for an alternative one that supports inline comments.

yomybaby (changes requested):

ironAiken2:
* **#2383** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #2395, opened 2024-04-30
hotfix: set cuda share first as an ai accelerator when both cuda device and shares available

lizable:
This PR resolves setting `cuda.shares` as the first AI accelerator when multiple setting (`cuda.device`, `cuda.shares`) applied in server-side.

- [ ] Documentation


yomybaby (approved):
LGTM

PR #2402, opened 2024-05-07
feat: introduce SourceCodeViewer component

### LINKED ISSUE #2399: Let's create a source code viewer component so that users can easily understand and copy the source code.
### Main idea

Currently, the code copy feature in the webui is not easy for users to understand because the code is heavily reformatted and does not provide highlighting. 

To improve this, let's create a component that can copy the code without deformation and provide code highlighting. 

[ref](https://github.com/umijs/dumi/blob/e5241906fb976d69662a5f27eef2950bd0647ec6/src/client/theme-default/builtins/SourceCode/index.tsx#L28)
~~[ref2](https://ui.lobehub.com/components/highlighter)~~
ref2 : LobeHub/UI seems to be unusable because there is an issue that UI is not drawn properly when converted to lit web component. 
|web component|react component|
|---|---|
|||

### Alternative ideas

_No response_

### Anything else?

_No response_

ironAiken2:
### Featue 
Utilizing the `prism-react-renderer` and `react-copy-to-clipboard` libraries, it provides highlighted code blocks and copy functionality. 

> [!CAUTION]
> **Warning:** When writing code that will pass children to the `code viewer`, code inside the `` should be written with zero indentation. If there is indentation, the `code viewer` will interpret that space as being provided "intentionally". 

### How To Use
1. Write the desired code in the codeBlock. Basically, you can use it the same way you would write code. If you don't want line breaks, write it as a single line of code. The UI will automatically wrap it, but if you copy the code, it will copy the original single line of code. 
2. Set the type of the code. Different types of code will have different highlighting. 
3. Set additional options. 
(language, wordWrap : if false, scroll bar will be shown)

### Screen Shots
|Light Mode|Dark Mode|
|---|---|
|||

- [ ] Documentation


ironAiken2:
* **#2402** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2402?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+
+interface SourceCodeViewerProps {
+  // In case of using this component in a web component, the props should be parsed from the web component.
+  codeBlock?: string;
+  language?: string;
+  fontSize?: string;
+}
+
+const SourceCodeViewer: React.FC<SourceCodeViewerProps> = ({
```
yomybaby, line 18: `SourceCodeViewer` itself should NOT be implemented assuming it will be used within a web component. Implement SourceCodeViewer to receive necessary input through props instead of `useWebComponentInfo`. For web component, we can make it High order component like [this](https://github.com/lablup/backend.ai-webui/blob/065f3e2d88a0ccb4c86b12d62e9d4adf976fab8a/react/src/index.tsx#L172-L184). @ironAiken2

ironAiken2:
Understood. The React component has changed its code to fit the React environment, and the web component's compatibility has been wrapped once more during the web component conversion process. 
> 
> 
> `SourceCodeViewer` itself should NOT be implemented assuming it will be used within a web component. Implement SourceCodeViewer to receive necessary input through props instead of `useWebComponentInfo`. For web component, we can make it High order component like [this](https://github.com/lablup/backend.ai-webui/blob/065f3e2d88a0ccb4c86b12d62e9d4adf976fab8a/react/src/index.tsx#L172-L184). @ironAiken2

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { CheckOutlined, CopyOutlined } from '@ant-design/icons';
+import { Button } from 'antd';
+import { Highlight, themes } from 'prism-react-renderer';
+import { useRef, useState } from 'react';
+// @ts-ignore
+import { CopyToClipboard } from 'react-copy-to-clipboard';
+
+interface SourceCodeViewerProps {
+  // In case of using this component in a web component, the props should be parsed from the web component.
+  codeBlock: string;
```
yomybaby: Let's handle code using `children` instead of another prop.

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,87 @@
...
+  language: string;
+  fontSize?: string;
+  minHeight?: string;
+}
+
+const SourceCodeViewer: React.FC<SourceCodeViewerProps> = ({
+  codeBlock,
+  language,
+  fontSize,
+  minHeight,
```
yomybaby: Do we need to use 'minHeight' and 'fontSize'?

yomybaby (changes requested):
Please add padding and border radius to the code box like this. It would be better to show the copy button  when hovering over the code block. ([ref](https://ant.design/docs/react/use-with-next)). Please change the default theme to a more general one as the current theme is too vivid.

ironAiken2:
Applied

> Please add padding and border radius to the code box like this. It would be better to show the copy button  when hovering over the code block. ([ref](https://ant.design/docs/react/use-with-next)). Please change the default theme to a more general one as the current theme is too vivid.
> 
> 
> 
> 
> 
> Let's handle code using `children` instead of another prop.
> 
> Do we need to use 'minHeight' and 'fontSize'?

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+        <pre
+          style={{
+            ...style,
+            width: '100%',
+            minHeight: minHeight,
+            fontSize: `${fontSize}`,
+            padding: token.padding,
+            borderRadius: token.borderRadius,
+            whiteSpace: 'pre-wrap',
+            wordBreak: 'break-word',
```
agatha197: how about overwriting style with children's style?

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+          }}
+        >
+          {tokens.map((line, i) => (
```
agatha197: please use lodash's map

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+  minHeight?: string;
+}
+
+const SourceCodeViewer: React.FC<SourceCodeViewerProps> = ({
+  codeBlock,
+  language,
+  fontSize,
+  minHeight,
+}) => {
+  const timer = useRef<number>();
```
yomybaby: When you use ref please add a prefix 'Ref'. 
```suggestion
  const timerRef = useRef<number>();
```

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+          timer.current = window.setTimeout(() => {
+            setIsCopied(false);
+          }, 2000);
+        }}
+      >
+        <Button
+          style={{
+            position: 'absolute',
+            bottom: 24,
+            right: 10,
```
yomybaby: do not use numbers explicitly, please use token's size.

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+            cursor: 'pointer',
+            zIndex: 1000,
```
yomybaby: Please remove this if it is not needed. If it is needed, please use `token.zIndexBase` to set the zIndex, for example, `token.zIndexBase + 1`. Setting a zIndex too high, even if not necessary, can break the UI related to the z-index stack (modals, popup, ...).

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+            opacity: isMouseEntered ? 1 : 0,
+            transition: 'opacity 0.2s ease-in-out',
+          }}
+          type="default"
```
yomybaby: You can remove type="default"

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,97 @@
...
+      >
+        <Button
+          style={{
+            position: 'absolute',
+            bottom: 24,
+            right: 10,
+            cursor: 'pointer',
+            zIndex: 1000,
+            opacity: isMouseEntered ? 1 : 0,
+            transition: 'opacity 0.2s ease-in-out',
```
yomybaby: Please set a backgroundColor to `token.colorFillSecondary` to make it look like a secondary button.

react/src/components/SourceCodeViewer.tsx
ironAiken2: Applied

yomybaby (changes requested):
Please check my comments and below:
- Please remove the margin from the component as it makes it hard to control the gap. In most cases, no margin is better, except for documentation purposes, because documentation is always a vertical layout. This component can be used in both vertical and horizontal layouts.
- It's not a good idea to use `wordBreak: 'break-word'` as defult. Let's add a prop to toggle it. Docusaurus also uses Prism React Renderer. Please refer to it.  When you add a long code and set 'language="javascript", you can see below:

react/src/components/SourceCodeViewer.tsx
ironAiken2: I removed that props.
I thought it to provide the control over the UI of the code viewer, but it would be more good to have the code viewer use the height and width of the parent and control the UI in the component that uses the code viewer.

yomybaby (changes requested):
Please move the copy button to top right of the box.

src/components/backend-ai-app-launcher.ts
```diff
@@ -1839,50 +1852,31 @@ export default class BackendAiAppLauncher extends BackendAIPage {
...
-                /path/to/source/
-                work@${this.tcpHost}:~/${this.mountedVfolderName}/
-              </span>
-              <mwc-icon-button
-                class="sftp-session-connection-copy"
-                icon="content_copy"
-                @click="${() =>
-                  this._copySSHConnectionExample('#rsync-string')}"
-              ></mwc-icon-button>
+            <div style="display: flex; flex-direction: column; gap: 6px; width: -webkit-fill-available;">
```
yomybaby:
```suggestion
            <div class="vertical layout flex" style="gap: var(--token-sizeXS); width: 100%;">
```

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,112 @@
...
+            opacity: isMouseEntered ? 1 : 0,
+            backgroundColor: token.colorFillSecondary,
+            cursor: 'pointer',
+            position: 'absolute',
+            transition: 'opacity 0.2s ease-in-out',
+          }}
+          size="small"
+          icon={isCopied ? <CheckOutlined /> : <CopyOutlined />}
+        ></Button>
+      </CopyToClipboard>
```
yomybaby: Lets' provide tooltip using Tooltip component to make it same ux of [copyable feature]([url](https://ant.design/components/typography#components-typography-demo-copyable)) of `Typhography` component.

yomybaby (changes requested):

react/src/components/SourceCodeViewer.tsx
ironAiken2: Applied

src/components/backend-ai-app-launcher.ts
ironAiken2: Applied

react/src/components/SourceCodeViewer.tsx
```diff
@@ -0,0 +1,114 @@
...
+        text={children}
+        onCopy={() => {
+          setIsCopied(true);
+          clearTimeout(timerRef.current);
+          timerRef.current = window.setTimeout(() => {
+            setIsCopied(false);
+          }, 2000);
+        }}
+      >
+        <Tooltip title={isCopied ? 'Copied' : 'Copy'}>
```
yomybaby: Please use i18n. If it does not exist, please create it.

yomybaby (changes requested):

ironAiken2:
Applied.
> 
> 
> Please use i18n. If it does not exist, please create it.

yomybaby (approved):
LGTM!

PR #2433, opened 2024-05-27
fix: temporally disable session commit to tar file in commit session dialog

lizable:
### TL;DR

This PR temporally disables commit session to `tar.gz` file in agent node, since there's no way to manage it as images in My Environment page.

### What changed?
Back then, commit session dialog has two feature, 1) Simply commit session to agent node and 2) Commit session to agent node and convert it to executable image. We disabled the first feature in this PR.

### How to test?
1. Create a session.
2. Click container commit icon in Controls column.
3. Dialog pops up. 
4. See commit button is disabled.
5. See the commit button changed to clickable status when you click the checkbox and input title of customized image at the input field.

### Why make this change?
For now we don't have any kind of management UI to limit and delete user-commit file, which may bring shortage of storage. Therefore It should be blocked.

---

- [ ] Documentation


lizable:
* **#2433** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2433?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @lizable and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

ironAiken2 (changes requested):
Everything works fine but If we don't provide the first feature, how about unifying the button's text?

yomybaby (changes requested):
After this PR, there is only one feature. So let's 
1. rename "Conver Sesssion to Image" label to other. (eg. Session name)
1. remove checkbox
1. set button title "push session to customized image" always.
image

yomybaby (approved):
LGTM.
We need to improve the management of i18n files. It seems that many keys are occasionally lost during the merging process. Thank you for checking and correcting the i18n keys.

ironAiken2 (approved):

PR #2437, opened 2024-05-28
add: providing additional vfolder mounts and arbitrary model definition path in model service launcher

lizable:
- Related: [backend.ai#2218](https://github.com/lablup/backend.ai/pull/2218)
- Prerequisites:
   - all-set configuration to creating / modifying model-service.
   - wsproxy v2 configuration
   - model-definition.yaml file

This PR adds two features in model service launcher, supporting arbitrary path of model definition file (previously set to `/models` and `model-definition.yaml(or yml)` file, and providing additional mounts and aliasing on other vfolders.

> ⚠️ NOTE: When updating the model service, it doesn't applied immediately. It applies when there's no corresponding session 
in running status, which represents "PROVISIONING" or "READY" status of model service.

### Test
- [x] Check whether you can create model service with no additional configuration, just model folder with `model-definition.yml` file inside. (default model folder mount path would be `/models`)
- [x] Check whether you can create model service with setting arbitrary path of mount destination for model folder (e.g. `/test`)
- [x] Check whether you can create model service with setting user-defined model definition file in model folder to mount by renaming it.
- [x] Check whether you can mount a general type folder with aliasing if needed

### Screenshot(s)
* Creating model service
   | After | Before |
   |------|--------|
   | Screenshot 2024-05-29 at 10 22 41 PM | Screenshot 2024-05-29 at 10 14 15 PM |

* Modifying model service
   | After | Before |
   |------|--------|
   | Screenshot 2024-05-29 at 10 18 46 PM |  Screenshot 2024-05-29 at 10 21 50 PM |

- [ ] Documentation
- [] Test case(s) to demonstrate the difference of before/after


lizable:
* **#2437** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2437?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @lizable and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

react/src/components/VFolderSelect.tsx
```diff
@@ -9,6 +9,7 @@ import React, { startTransition, useEffect } from 'react';
 export type VFolder = {
   name: string;
   id: string;
+  row_id: string;
```
yomybaby: There is no `row_id` in the API response of `/folders?groupd_id=someid`.
```suggestion
```

react/src/components/VFolderLazyView.tsx
```diff
@@ -45,11 +49,11 @@ const VFolderLazyView: React.FC<VFolderLazyViewProps> = ({
           });
         }}
       >
-        <FolderOutlined /> {vFolder.name}
+        <FolderOutlined /> {`${vFolder.name}`}
```
yomybaby: If you don't use the `mountDestination` you added, please revert all changes related to that.
```suggestion
        <FolderOutlined /> {vFolder.name}
```

react/src/components/VFolderTableFormItem.tsx
```diff
@@ -8,17 +9,21 @@ import _ from 'lodash';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 
-interface VFolderTableFromItemProps extends Omit<FormItemProps, 'name'> {
+interface VFolderTableFormItemProps extends Omit<FormItemProps, 'name'> {
   filter?: VFolderTableProps['filter'];
+  rowKey?: keyof VFolder;
```
yomybaby, line 14: 👍

react/src/components/ServiceLauncherModal.tsx
```diff
@@ -636,11 +763,64 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
...
+                  filter={(vf) =>
+                    vf.name !== selectedModelFolder &&
+                    vf.status === 'ready' &&
+                    vf.usage_mode !== 'model'
+                  }
+                  tableProps={{
+                    size: 'small',
+                  }}
+                />
               </>
```
yomybaby: You don't need an extra state to update some component based on the form's value. Instead of using an extra state, you can wrap the part you want to update using `<Form.Item noStyle shouldUpdate ...>`. After that, you can delete the `selectedModelFolder` state, `updateSelectedModelFolder`, and the `useEffect` for that.

Some like this(I have not tested below code.):
```suggestion
                <Form.Item noStyle shouldUpdate={(prev,cur)=>prev.vFolderName !== cur.vFolderName}>
                  {()=>{
                    return <VFolderTableFormItem
                      rowKey={'id'}
                      label={t('modelService.AdditionalMounts')}
                      filter={(vf) =>
                        vf.name !== formRef.current?.getFieldValue('vFolderName') &&
                        vf.status === 'ready' &&
                        vf.usage_mode !== 'model'
                      }
                      tableProps={{
                        size: 'small',
                      }}
                    />
                  }}
                </Form.Item>
```

react/src/components/ServiceLauncherModal.tsx
```diff
@@ -609,25 +725,36 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
...
-                    />
-                  </Form.Item>
+                  <>
+                    <Flex
+                      direction="column"
+                      gap={'xxs'}
+                      align="stretch"
+                      justify="between"
+                    >
+                      <Form.Item
```
yomybaby: Can I ask why we need to wrap using `<Flex>` even if it has only one child?

react/src/components/ServiceLauncherModal.tsx
```diff
@@ -636,11 +763,64 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
                       required
                     >
                       <Suspense fallback={<Skeleton.Input active />}>
-                        <VFolderLazyView uuid={endpoint?.model} />
+                        <VFolderLazyView
+                          uuid={endpoint?.model}
+                          mountDestination={endpoint?.model_mount_destination}
```
yomybaby: If you want to display the `mountDestination` path alongside with the vFolder name, it's better to display it using one of the methods below:

As a sibling:
```
<Flex>
  <FolderLazyView/>
  <TyphoGraphy.Text>{endpoint?.model_mount_destination}</TyphoGraphy.Text>
</Flex>
```

As a child:
```
<FolderLazyView>
  <TyphoGraphy.Text>{endpoint?.model_mount_destination}</TyphoGraphy.Text>
</FolderLazyView>
```

react/src/components/ServiceLauncherModal.tsx
```diff
@@ -636,11 +763,64 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
...
+                      allowClear
+                      placeholder={'/models'}
+                      disabled={endpoint ? true : false}
+                    />
+                  </Form.Item>
+                  <MinusOutlined
+                    style={{ fontSize: token.fontSizeXL }}
+                    rotate={290}
+                  />
+                  <Form.Item
```
yomybaby: If this is only for the dividing, you can use `<Divider type="vertical"/>`. [ref](https://ant.design/components/divider#divider-demo-vertical)

yomybaby (changes requested):

react/src/components/ServiceLauncherModal.tsx
lizable: I just added as if looks like a path divider, (`/`)
But would It be better to replace with `<Divider>` component?

react/src/components/ServiceLauncherModal.tsx
lizable: Right. I just rearranged the component but forgot to remove it...

react/src/components/ServiceLauncherModal.tsx
lizable: Thanks for the suggestion! I think it's okay to get rid of model_mount_destination since it's already applied as Input form with disabled status...

react/src/components/ServiceLauncherModal.tsx
```diff
@@ -239,7 +287,21 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
         open_to_public: values.openToPublic,
         config: {
           model: values.vFolderName,
-          model_mount_destination: '/models', // FIXME: hardcoded. change it with option later
+          model_version: 1, // FIXME: hardcoded. change it with option later
+          model_mount_destination: values.modelMountDestination,
+          model_definition_path: values.modelDefinitionPath,
+          extra_mounts: values.mounts.reduce(
```
yomybaby: The `values.mounts` can be `undefined` when the user does not change any extra mounts. Please use `_.reduce` to make this safe.

yomybaby (changes requested):
Please check below:
- When the user enters `Mount Destination For Model Folder` and then makes it empty again, the user gets an error when submitting:
```
{
    "type": "https://api.backend.ai/probs/invalid-api-params",
    "title": "Missing or invalid API parameters.",
    "msg": "Alias name cannot be empty."
}
```

yomybaby (changes requested):
Please check another bug, you can reproduce it:
- When creating a service, input a alias path like this:
image
- After that, click modify button. You can not see the the alias path what you enter above step.
image

lizable:
For now, there's no way to import alias of extra mounted vfolders, unless fetching from server.

react/src/components/ServiceLauncherModal.tsx
```diff
@@ -641,6 +739,59 @@ const ServiceLauncherModal: React.FC<ServiceLauncherProps> = ({
...
+                  label={t('modelService.AdditionalMounts')}
+                  filter={(vf) =>
+                    vf.name !== formRef.current?.getFieldValue('vFolderName') &&
+                    vf.status === 'ready' &&
+                    vf.usage_mode !== 'model'
+                  }
+                  tableProps={{
+                    size: 'small',
+                  }}
+                />
```
yomybaby: If you do not wrap the `<Form.Item noStyle shouldUpdate={...}>`, this part will not re-render when `vFolderName` is changed. Please check [my previous comment](https://github.com/lablup/backend.ai-webui/pull/2437#discussion_r1619562472) again.

yomybaby (changes requested):
This [feedback](https://github.com/lablup/backend.ai-webui/pull/2437#pullrequestreview-2087015822) is also not resolved.

yomybaby (approved):
LGTM, but I'll merge this after reviewing the top of the stack. #2447

yomybaby:
### Merge activity

* **May 31, 2:51 AM EDT**: @yomybaby started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2437).

PR #2441, opened 2024-05-29
feat: memorize board's items by using local storage in SummaryPage

ironAiken2:
### Feature
The scope of this PR stack is to allow changes to Summary page items to be saved via using local-storage and to display item loading indicators via using Suspense.
- Enable local storage so that changes to items on the Summary page can be saved. 
- Use Suspense to show an indicator for slow rendering of Board Item Components.
- Restore the Commented out SummaryPage component in App.tsx.
- Remove <backend-ai-summary-view> components in backend-ai-webui.ts
- If you can't see Summary Page Items, delete localStorage.

**Checklist:** (if applicable)

- [ ] Mention to the original issue
- [ ] Documentation
- [ ] Minium required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [ ] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

PR #2450, opened 2024-05-31
refactor: Change the items on the summary page to match BAIBoard

ironAiken2:
### Feature
The scope of this PR is to reflect on each item in the BAIBoard. Each item name starts with SummaryItem. All items will behave responsively in preparation for resize. 

We need a new component design for the BAIBoard. 
- StartMenu
  - The size of the logo changes depending on the container size. 
- Invitation 
  - The default UI has changed. 
- App Download 
  - Same as before. 
- ShortCut 
  - The default UI has changed. 
  - The number of items in a single ROW depends on the component size. 
  - Description added.

- Restore the Commented out SummaryPage component in App.tsx.
- Remove `<backend-ai-summary-view>` components in backend-ai-webui.ts
- If you can't see Summary Page Items, delete localStorage.

- [ ] Documentation

### DIFF
--- a/react/src/components/SummaryPageItems/SummaryItemShortCut.tsx
+++ b/react/src/components/SummaryPageItems/SummaryItemShortCut.tsx
@@ -1,0 +1,187 @@
+import { useWebUINavigate } from '../../hooks';
+import Flex from '../Flex';
+import {
+  FileOutlined,
+  InboxOutlined,
+  SettingOutlined,
+  ToolOutlined,
+} from '@ant-design/icons';
+import { Typography, theme } from 'antd';
+import { useTheme } from 'antd-style';
+import _ from 'lodash';
+import { useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+
+interface itemProps {
+  id: string;
+  title: string;
+  content: string;
+  icon?: JSX.Element;
+  onClick?: () => void;
+}
+
+const SummaryItemShortCut: React.FC = () => {
+  const { t } = useTranslation();
+  const { token } = theme.useToken();
+  const customTheme = useTheme();
+  const webuiNavigate = useWebUINavigate();
+  const [items] = useState<itemProps[]>([
+    {
+      id: 'systemSetting',
+      title: t('summary.ChangeSystemSetting'),
+      content: t('summary.ChangeSystemSettingDescription'),
+      icon: (
+        <SettingOutlined
+          style={{
+            fontSize: token.fontSizeHeading3,
+            color: token.colorTextDescription,
+          }}
+        />
+      ),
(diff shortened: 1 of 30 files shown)


react/src/components/SummaryPageItems/SummaryItemShortCut.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const { t } = useTranslation();
+  const { token } = theme.useToken();
+  const webuiNavigate = useWebUINavigate();
+  // In case of adding new items, implemented with useState
+  const [items] = useState<itemProps[]>([
+    {
+      id: 1,
+      title: t('summary.ChangeSystemSetting'),
+      content:
+        '계산환경 이미지, 그래픽 환경, 스케일링 및 플러그인을 변경합니다. ',
```
agatha197: Please use i18n

react/src/components/SummaryPageItems/SummaryItemShortCut.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const container = containerRef.current;
+    if (!container) {
+      return;
+    }
+    const updateImageSize = () => {
+      const containerWidth = container.clientWidth;
+
+      if (containerWidth <= 410) {
+        setChunkSize(1);
+      } else if (containerWidth <= 820) {
```
agatha197: How about replacing it with token's screen size?

agatha197 (changes requested):

react/src/components/SummaryPageItems/SummaryItemShortCut.tsx
ironAiken2: The screen size should be based on the screen size of the page, but BAIBoard doesn't provide a ref property. 

And the size of the screen size token is too large to use.

lizable: Maybe we can set "screenXS" value to 410.

lizable: or we can just create custom-token for that.
https://ant-design.github.io/antd-style/guide/custom-theme#custom-token

lizable (changes requested):
Works fine, but I think we'd better make most of "token" rather than using magic numbers.

react/src/components/SummaryPageItems/SummaryItemShortCut.tsx
ironAiken2: I applied it using antd's custom token.

yomybaby (changes requested):
Please reflect changes in #2441 and hide and block the link  on `SummaryItemShortCut` only admin can access.

PR #2480, opened 2024-06-14
fix: run form validation after changing preset

yomybaby:
This pull request ensures that when the 'resource' field is validated when preset is changed, the validation is conducted recursively. It addresses an issue where non-recursive validation could have missed nested fields under 'resource'. The change involves modifying the validateFields method to include the 'recursive: true' option. This update impacts the ResourceAllocationFormItems component in the React codebase.

---

- [ ] Documentation


ironAiken2 (approved):
LGTM!

PR #2509, opened 2024-07-02
fix: change login input validation logic to support both email and non-email types

ironAiken2:
### TL;DR

Added input validation for the login form to ensure that the user ID is either a valid email address or a username that matches specified patterns.

### What changed?

- Introduced `_validateInput` function in `backend-ai-login.ts`.
- Updated the `<mwc-textfield>` element to use the `_validateInput` function for validation.
- Modified the input field to adhere to the validation patterns specified for email and username.

### How to test?

1. Open the login form.
2. Try entering an invalid email or username. Ensure that the validation message is shown.
3. Enter a valid email or username. Ensure that no validation message is shown and the form can be submitted.

`email type` : Beginning with uppercase and lowercase letters of the alphabet, numbers, dots, underscores, and minus signs. The @ symbol followed by uppercase and lowercase letters of the alphabet, numbers, dots, and minus signs.
`username type` : Text between 3 and 20 characters, consisting of uppercase and lowercase letters of the alphabet, numbers, dots, underscores, and minuses.

### Why make this change?

To improve user experience by providing immediate feedback on input errors in the login form, thereby preventing users from submitting invalid data.

---

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -2070,7 +2070,7 @@ export default class BackendAILogin extends BackendAIPage {
                     disabled
                   ></mwc-icon-button>
                   <mwc-textfield
-                    type="email"
+                    required
                     id="id_user_id"
                     maxlength="64"
                     autocomplete="username"


ironAiken2:
* **#2509** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/components/backend-ai-login.ts
```diff
@@ -1979,6 +1979,25 @@ export default class BackendAILogin extends BackendAIPage {
       : password.setAttribute('type', 'password');
   }
 
+  private _validateInput() {
+    // Email validation pattern
+    const emailPattern = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
+    // Username validation pattern
+    const usernamePattern = /^[a-zA-Z0-9._-]{3,20}$/;
```
yomybaby: Can you include the reference to this rule in the description?

yomybaby (changes requested):
Even when the email/username validation fails, WebUI sends a login request when you click the login button.

src/components/backend-ai-login.ts
lizable: Agree. I found that username is allowed to 64 characters, but is it really restricted to 20 characters?
(https://github.com/lablup/backend.ai/blob/df57a091147454fed5c4563142e19a8d7c86dae4/src/ai/backend/manager/models/user.py#L120)

lizable (changes requested):
Could you check the conditions of username and email pattern from Backend.AI Core so that we could verify whether the regex is acceptable?

src/components/backend-ai-login.ts
ironAiken2: I couldn't find rules for the input email and username, so I set them arbitrarily: for email, I searched a popular expression and applied it, and for username, I removed the max_length rule and changed it to apply only min_length (max_length is defined in mwc-textfield). 

In the description, I've written down what type of input each regular expression accepts.

src/components/backend-ai-login.ts
```diff
@@ -1979,6 +1979,25 @@ export default class BackendAILogin extends BackendAIPage {
       : password.setAttribute('type', 'password');
   }
 
+  private _validateInput() {
+    // Email validation pattern
+    const emailPattern = /^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
```
yomybaby: Please use simpler client-side validation programmatically because it is difficult to validate perfectly.
/^[^\s@]+@[^\s@]+\.[^\s@]+$/;

yomybaby (changes requested):

ironAiken2:
Currently, the webui does not specify any other validation for username, so username can be any type of input.

For this reason, I've changed it to not set any other validation on username or email inputs, but only set the `required` and `max_length` properties.

@yomybaby @lizable

yomybaby (approved):
LGTM

PR #2555, opened 2024-07-17
refactor: use controllable value in VFolderSelect

gahyuun:
### Changes
- Use 'useControllableValue' hook to allow uncontrolled as well 

- [ ] Documentation

### DIFF
--- a/react/src/components/VFolderSelect.tsx
+++ b/react/src/components/VFolderSelect.tsx
@@ -94,7 +94,6 @@ const VFolderSelect: React.FC<VFolderSelectProps> = ({
         value: _.first(filteredVFolders)?.[valuePropName],
       }
     : undefined;
-  // TODO: use controllable value
 
   const [value, setValue] = useControllableValue<string | undefined>({
     value: selectProps.value,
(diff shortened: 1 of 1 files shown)


react/src/components/VFolderSelect.tsx
```diff
@@ -93,17 +94,25 @@ const VFolderSelect: React.FC<VFolderSelectProps> = ({
         value: _.first(filteredVFolders)?.[valuePropName],
       }
     : undefined;
-  // TODO: use controllable value
+
+  const [value, setValue] = useControllableValue<string | undefined>({
+    value: selectProps.value,
+    onChange: selectProps.onChange,
+  });
```
yomybaby, line 101: This is an incorrect way to pass arguments to useControllableValues. Please check [this code](https://github.com/alibaba/hooks/blob/master/packages/hooks/src/useControllableValue/index.ts#L38). If the props object has a 'value' property, even if it is undefined, it works in controlled mode. 
There are many components use `useControllableValue` in this way in WebUI. we need to fix all of them.

yomybaby (changes requested):

Issue #2568, opened 2024-07-22
It is hard to reach the Controls in the session list


ggstargame45:
### Main idea

When I try to use backend.ai in my phones or small screens.
It is hard to press the terminal or so in the controls.
So, I thought it would be better if the first three columns of the session list didn't have the frozen property.
This is a video to help you understand.
<table>
    <tr align='center'>
        <td>
            <h3>Current</h3>
        </td>
        <td>
            <h3>Fixed</h3>
        </td>
    </tr>
    <tr align='center'>
        <td>
            <video src="https://github.com/user-attachments/assets/0397b820-e019-448f-a4de-1b7c7bfd8333">
        </td>
        <td>
            <video src="https://github.com/user-attachments/assets/3d74a22b-6f7c-4abc-b1e8-32f444b80e01">
        </td>
    </tr>
</table>

### Anything else?

I understand that the current script made in lit will be obsolete after it is changed to react.
However, I thought it would help the accessability until it is replaced.

PR #2569, opened 2024-07-22
fix: get rid of the frozen properties in session list component

ggstargame45:
### This is a follow up of #2568

Added this for users to easily to reach the Controls in the session list in small screens like smartphones.

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -4321,7 +4321,6 @@ ${rowData.item[this.sessionNameField]}</pre
             this._isRunning
               ? html`
                   <vaadin-grid-column
-                    frozen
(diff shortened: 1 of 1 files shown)


agatha197:
How about making the `frozen` properties disappear based on the screen size? This information is important, so we decided to freeze them to ensure visibility even when scrolling horizontally.

ggstargame45:
> How about making the `frozen` properties disappear based on the screen size? The frozen columns are important information, so we decided to freeze them to ensure visibility even when scrolling horizontally.

I understand. I'll apply it and commit it right away.

yomybaby (changes requested):
It's better to remove `frozen` only for small screen.

PR #2572, opened 2024-07-23
feat: introduce `useControllableState`

gahyuun:
### TL;DR

implement useControllableState hook 

### useControllableState
The overall implementation is based on [useControllableValue](https://github.com/alibaba/hooks/blob/master/packages/hooks/src/useControllableValue/index.ts), and the code to determine whether it is controlled is based on [useControlledState](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/utils/src/useControlledState.ts)

1.  isControlled :  `useControllableValue` of ahooks treats a component as a controlled component if the value exists for props object . But `useControllableState` treats a component as a uncontrolled component  If value is undefined, even if value exists like `useControlledState` of react-spectrum hook 
so I changed the code to determine whether component is controlled 

2. display a warning if the value of isControlled changes. ( console.warn ~ )

### change from useControllableValue to useControllableState

1. BAICodeEditor : I changed the value from being passed to children to being passed to value prop and
The below code was modified because {...CodeMirrorProps} could potentially overwrite value and defaultValue 
```
 <CodeMirror
      value={script}
      onChange={(value, viewUpdate) => setScript(value)}
       {...CodeMirrorProps}/>

```

2. NonLinearSlider : Removed defaultValue in the second argument because the presence of defaultValue in the first argument makes the defaultValue in the second argument meaningless 
```
  const [controlledValue, setControlledValue] = useControllableState(
    {
      value,
      defaultValue,
      onChange,
    },
    {
      defaultValue: defaultValue ?? normalizedSteps[0]?.value,
    },
  );

```

3. ResourceGroupSelect , ResourceGroupSelectForCurrentProject , StorageSelect : I removed _.omitBy because component can be uncontrolled component even if the value exists on prop
```
 useControllableValue<string>(
      _.omitBy(
        {
          value: searchValue,
          onChange: onSearch,
        },
        _.isUndefined,
      ),
```

and DatePickerISO also used useControllableValue, but I didn't change it due to a bug issue. 

### How to test?

I used the test code from useControllableValue and added a few more tests  
The test I added  : `test valuePropName of options`, `test defaultValuePropName of options`,`test trigger of options`,`warning is logged when changed from uncontrolled to controlled`,`defaultValue should be applied when value is undefined`

### Why make this change?

Even if the value is explicitly set to undefined, it will be treated as uncontrolled, just like React's built-in components.
so I implement useControllableState hook to replace useControllableValue

### DIFF
--- a/react/src/components/BAICodeEditor.tsx
+++ b/react/src/components/BAICodeEditor.tsx
@@ -30,9 +30,6 @@ const BAICodeEditor: React.FC<BAICodeEditorProps> = ({
 
   return (
     <CodeMirror
-      {...CodeMirrorProps}
-      value={script}
-      onChange={(value, viewUpdate) => setScript(value)}
       theme={'dark'}
       extensions={
         lineWrapping ? [EditorView.lineWrapping, ...extensions] : extensions
@@ -42,6 +39,9 @@ const BAICodeEditor: React.FC<BAICodeEditorProps> = ({
       basicSetup={{
         lineNumbers: showLineNumbers,
       }}
+      {...CodeMirrorProps}
+      value={script}
+      onChange={(value, viewUpdate) => setScript(value)}
     />
   );
 };
--- a/react/src/components/BAICodeEditor.tsx
+++ b/react/src/components/BAICodeEditor.tsx
@@ -1,10 +1,10 @@
+import useControllableState from '../hooks/useControllableState';
 import { loadLanguage, LanguageName } from '@uiw/codemirror-extensions-langs';
 import CodeMirror, { ReactCodeMirrorProps } from '@uiw/react-codemirror';
 import { EditorView } from '@uiw/react-codemirror';
-import { useControllableValue } from 'ahooks';
 
 interface BAICodeEditorProps extends Omit<ReactCodeMirrorProps, 'language'> {
-  children: string;
+  value: string;
   onChange: (value: string) => void;
   language: LanguageName;
   editable?: boolean;
@@ -13,7 +13,7 @@ interface BAICodeEditorProps extends Omit<ReactCodeMirrorProps, 'language'> {
 }
 
 const BAICodeEditor: React.FC<BAICodeEditorProps> = ({
-  children,
+  value,
   onChange,
   language = 'shell',
   editable = false,
@@ -21,15 +21,16 @@ const BAICodeEditor: React.FC<BAICodeEditorProps> = ({
   lineWrapping = false,
   ...CodeMirrorProps
 }) => {
-  const [script, setScript] = useControllableValue<string>({
+  const [script, setScript] = useControllableState<string>({
     defaultValue: '',
-    value: children,
+    value,
     onChange,
   });
   const extensions = [loadLanguage(language)!];
 
   return (
     <CodeMirror
+      {...CodeMirrorProps}
       value={script}
       onChange={(value, viewUpdate) => setScript(value)}
       theme={'dark'}
@@ -41,7 +42,6 @@ const BAICodeEditor: React.FC<BAICodeEditorProps> = ({
       basicSetup={{
         lineNumbers: showLineNumbers,
       }}
-      {...CodeMirrorProps}
     />
   );
 };
--- a/react/src/components/ResourceGroupSelect.tsx
+++ b/react/src/components/ResourceGroupSelect.tsx
@@ -1,9 +1,9 @@
 import { useBaiSignedRequestWithPromise } from '../helper';
 import { useUpdatableState } from '../hooks';
 import { useTanQuery } from '../hooks/reactQueryAlias';
+import useControllableState from '../hooks/useControllableState';
 import { useCurrentProjectValue } from '../hooks/useCurrentProject';
 import TextHighlighter from './TextHighlighter';
-import { useControllableValue } from 'ahooks';
 import { Select, SelectProps } from 'antd';
 import _ from 'lodash';
 import React, { useEffect, useTransition } from 'react';
@@ -27,18 +27,13 @@ const ResourceGroupSelect: React.FC<ResourceGroupSelectProps> = ({
   const currentProject = useCurrentProjectValue();
   const [key, checkUpdate] = useUpdatableState('first');
   const [controllableSearchValue, setControllableSearchValue] =
-    useControllableValue<string>(
-      _.omitBy(
-        {
-          value: searchValue,
-          onChange: onSearch,
-        },
-        _.isUndefined,
-      ),
-    );
+    useControllableState<string>({
+      value: searchValue,
+      onChange: onSearch,
+    });
 
   const [controllableValue, setControllableValue] =
-    useControllableValue(selectProps);
+    useControllableState(selectProps);
 
   const [isPendingLoading, startLoadingTransition] = useTransition();
   const { data: resourceGroupSelectQueryResult } = useTanQuery<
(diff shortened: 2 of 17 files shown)


yomybaby (changes requested):
If you started with an open-source hook, please add links of it. Also, please let others know what parts you have changed.

react/src/components/BAICodeEditor.tsx
```diff
@@ -1,35 +1,36 @@
...
     defaultValue: '',
-    value: children,
+    value,
     onChange,
   });
   const extensions = [loadLanguage(language)!];
 
   return (
     <CodeMirror
+      {...CodeMirrorProps}
```
agatha197: The properties below, such as `basicSetup`, `theme`, etc., cannot be overridden. Move the items that should not be overridden below `{...CodeMirrorProps}`, and move the items that can be overridden above `{...CodeMirrorProps}`.

react/src/components/ResourceGroupSelect.tsx
```diff
@@ -27,18 +27,12 @@ const ResourceGroupSelect: React.FC<ResourceGroupSelectProps> = ({
...
-          value: searchValue,
-          onChange: onSearch,
-        },
-        _.isUndefined,
-      ),
-    );
-
+    useControllableState<string>({
+      value: searchValue,
+      onChange: onSearch,
```
agatha197, line 32: Doesn't it need `_.omitBy`?

react/src/components/ResourceGroupSelectForCurrentProject.tsx
```diff
@@ -27,15 +27,10 @@ const ResourceGroupSelectForCurrentProject: React.FC<
...
-        {
-          value: searchValue,
-          onChange: onSearch,
-        },
-        _.isUndefined,
-      ),
-    );
+    useControllableState<string>({
+      value: searchValue,
+      onChange: onSearch,
```
agatha197, line 32: same as above.

react/src/components/StorageSelect.tsx
```diff
@@ -58,13 +58,13 @@ const StorageSelect: React.FC<Props> = ({
...
+  const [controllableState, setControllableState] = useControllableState({
+    value,
+    onChange,
+    defaultValue,
+  });
   const [controllableSearchValue, setControllableSearchValue] =
-    useControllableValue(
-      _.omitBy({ value: searchValue, onChange: onSearch }, _.isUndefined),
-    );
+    useControllableState({ value: searchValue, onChange: onSearch });
```
agatha197, line 67: same as above.

react/src/components/VFolderSelect.tsx
```diff
@@ -96,14 +98,17 @@ const VFolderSelect: React.FC<VFolderSelectProps> = ({
   // TODO: use controllable value
```
agatha197: Let's remove this line.

react/src/hooks/useControllableState.ts
```diff
@@ -0,0 +1,135 @@
...
+  value: T;
+  defaultValue?: T;
+  onChange: (val: T) => void;
+}
+/**
+ * Custom hook to  manage this kind of state. managed by itself or controlled by it's parent
+ * useControllableState is based on useControllableValue.
+ * However if the value is undefined, the component is treated as an uncontrolled component
+ */
+// params 설명 더 작성하기
```
agatha197:
```suggestion
```

react/src/hooks/useControllableState.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  return [stateRef.current, useMemoizedFn(setState)] as const;
+}
+
+export default useControllableState;
+
+function isFunction(value: unknown): value is (...args: any) => any {
+  return typeof value === 'function';
+}
```
agatha197: you can use lodash's [isFunction](https://lodash.com/docs/4.17.15#isFunction)

agatha197 (changes requested):

react/src/components/ResourceGroupSelect.tsx
gahyuun: we wanted to treat the component as uncontrolled component when value was set to undefined,
but useControllableValue treated component as controlled component  if the value property existed in the props,  so we used _.omitBy to not assign the value property when value was undefined.

but now, No need for _.omitBy because I replaced useControllableValue with useControllableState  ( useControllableState treats component as uncontrolled component when value is set to undefined  )

react/src/components/BAICodeEditor.tsx
gahyuun: I modified it ! https://github.com/lablup/backend.ai-webui/pull/2572/files#diff-a6f793908a50d0ec4d6fec1664904acc0e6eb3ba89bb374f697c3ea3ff739ddeR42

react/src/hooks/useControllableState.ts
gahyuun: I modified it !  https://github.com/lablup/backend.ai-webui/pull/2572/files#diff-e98538d8ef63e2ebd590961cc47a0a8badea1d72e19f7cf0cbe5a3501e0a97bbR20-R23

react/src/hooks/useControllableState.ts
gahyuun: I modified it https://github.com/lablup/backend.ai-webui/pull/2572/files#diff-e98538d8ef63e2ebd590961cc47a0a8badea1d72e19f7cf0cbe5a3501e0a97bbR79

react/src/components/VFolderSelect.tsx
gahyuun: I modified it ! https://github.com/lablup/backend.ai-webui/pull/2572/files#diff-d191efffefe30deea0936aba97b2645040e05e1276691079ca15866550abbdefL96

react/src/hooks/useControllableState.ts
```diff
@@ -0,0 +1,131 @@
...
+    defaultValue,
+    defaultValuePropName = 'defaultValue',
+    valuePropName = 'value',
+    trigger = 'onChange',
+  } = options;
+
+  const value = props[valuePropName] as T;
+
+  let isControlledRef = useRef(value !== undefined);
+  let isControlled = value !== undefined;
```
yomybaby:
```suggestion
  const isControlledRef = useRef(value !== undefined);
  const isControlled = value !== undefined;
```

react/src/hooks/useControllableState.ts
```diff
@@ -0,0 +1,131 @@
...
+    let wasControlled = isControlledRef.current;
+    if (wasControlled !== isControlled) {
+      console.warn(
+        `WARN: A component changed from ${wasControlled ? 'controlled' : 'uncontrolled'} to ${
+          isControlled ? 'controlled' : 'uncontrolled'
+        }.`,
+      );
+    }
+    isControlledRef.current = isControlled;
+  }, [isControlled]);
```
yomybaby: This is for the development. It's better to run `console.warn` only in the `development` mode.

yomybaby: Don't forget to remove a related test code.

yomybaby (changes requested):

react/src/hooks/useControllableState.ts
gahyuun: I modified it 
https://github.com/lablup/backend.ai-webui/pull/2572/commits/f3d10d20efe6d9993cd0359e000c53499f0b912d

yomybaby (commented):
LGTM

yomybaby (approved):

PR #2584, opened 2024-07-29
fix: use starts_at instead of created_at for session reservation & elapsed time

gahyuun:
### TL;DR

This PR resolves issue that session reservation time and session elapsed time were incorrectly displayed 

###  Why make this change?

session reservation time should mean session start time, not session creation time
session elapsed time should be  caculated by session start time,  not session creation time

### What changed?

session reservation time means session start time
session elapsed time is calculated based on the session start time and session terminate time
session elapsed time is displayed only after the session starts

### How to test?

1. set a session start time
2. start a session
3. check that reservation time is displayed correctly
4. check that elapsed time is displayed after the session starts and elapsed time is calculated correctly

### DIFF
--- a/react/src/index.tsx
+++ b/react/src/index.tsx
@@ -13,6 +13,7 @@ import { useSuspendedBackendaiClient } from './hooks';
 import { useCurrentResourceGroupValue } from './hooks/useCurrentProject';
 import { ThemeModeProvider } from './hooks/useThemeMode';
 import { Tag, theme } from 'antd';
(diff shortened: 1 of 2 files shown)


ironAiken2 (changes requested):
Elapsed time and date are not displayed correctly.

ironAiken2 (approved):
LGTM

agatha197 (approved):
LGTM

PR #2615, opened 2024-08-07
feat: migrate image list 

gahyuun:
### TL;DR

migrate ImageList 

###  Why make this change?

To migrate from lit component to react component

### What changed?

- ImageInstallModal
    - Component that install images
    - I deselected the rows when clicking the install button in the modal because I felt the flow of deselecting the rows was more natural, whereas before it would not deselect the rows.
    - The newly added phrases are written in en and ko only
    - However, I didn't proceed with replacing "download" with "install". This was because there were more download phrases than I thought, and I thought it might be more confusing if I replaced a lot of them

- ImageList
    - I used suspense to prevent the entire screen from loading when the image list is first rendered

- ManageAppsModal, ManageImageResourceLimitModal
    - I modified it because it's called from a React component, not a lit component.
    - Implemented the getServicePorts function in ManageAppsModal without passing service port as a prop 

- ResourceNumber
    - I added ranges via the range and max properties ( ex 1~∞ Core )
    - The ResourceNumber component made rendering the table too slow, so I used Memo

### How to test?

- imageList
1. enter the environment page 
2. select the row
3. sort the column

- ManageImageResourceLimitModal
1. click button
3. change resource
4. click Okay button
5. check if the resource has changed

- ManageAppsModal
1. click button
2. add port
3. delete port
4. check if the ports has changed

- ImageInstallModal
1. select uninstalled images
2. click install button
3. check uninstalled images info
4. click install button and check if an notification appears

1. select installed images
2. check warn message

1. select uninstalled images and installed images
2. check exclude message


react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,345 @@
...
+        );
+      },
+    },
+  ];
+
+  return (
+    <>
+      <Flex justify="end" style={{ height: '3rem', padding: '.5rem 1rem' }}>
+        <Button
+          size="middle"
```
ironAiken2: ant'd button's default size is 'middle'

react/src/index.tsx
```diff
@@ -284,13 +256,11 @@ const root = ReactDOM.createRoot(
   document.getElementById('react-root') as HTMLElement,
 );
 root.render(
-  <React.StrictMode>
```
ironAiken2: Why did you remove this code?

react/src/pages/EnvironmentPage.tsx
```diff
@@ -45,38 +62,40 @@ const EnvironmentPage = () => {
         },
       }}
     >
+      {curTabKey === 'image' ? (
+        <Flex
+          style={{
+            display: 'block',
+            paddingTop: token.paddingContentVerticalSM,
```
ironAiken2:
```suggestion
```

This code seems to break the consistency of the css styles. It would be better to control it inside the `ImageList`.

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,345 @@
...
+import _ from 'lodash';
+import { Key, useMemo, useState, useTransition } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useLazyLoadQuery } from 'react-relay';
+
+export type EnvironmentImage = NonNullable<
+  ImageListQuery$data['images']
+>[number];
+
+export default function ImageList() {
```
ironAiken2: I think it would be better to use the arrow function. And it would be nice to be able to pass the style to props in case we want to change the style of that component elsewhere. 
```suggestion
const ImageList: React.FC<{ style?: React.CSSProperties;}> = ({ style }) => {
```

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,345 @@
...
+            </Flex>
+          </>
+        );
+      },
+    },
+  ];
+
+  return (
+    <>
+      <Flex justify="end" style={{ height: '3rem', padding: '.5rem 1rem' }}>
```
ironAiken2:
```suggestion
    <Flex
      direction="column"
      align="stretch"
      style={{
        flex: 1,
        ...style,
      }}
    >
      <Flex justify="end" style={{ padding: token.paddingSM }}>
```

It seems to be changed to maintain stylistic consistency with other components on the same page.

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,345 @@
...
+          onClick={() => {
+            setIsOpenInstallModal(true);
+          }}
+        >
+          {t('environment.Install')}
+        </Button>
+      </Flex>
+      <Table
+        rowKey="id"
+        scroll={{ x: '100vw', y: 'calc(100vh - (9.0625rem + 3rem + 4.5rem))' }}
```
ironAiken2:
```suggestion
scroll={{ x: 'max-content' }}
```

It seems to be affecting the styling of the table columns. The following modifications should fix the issue.

ironAiken2 (changes requested):
Currently, we don't have pagination, so It's getting too much data and it's causing lag on page navigation and other interactions. Can you take a look at this as well?

react/src/index.tsx
gahyuun: my misake .. i modified it

react/src/components/ImageList.tsx
gahyuun: I fixed style related issues and added the style prop

gahyuun: I used function declaration to write the declaration and default export at once. can you explain why arrow function is better in react component?

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+  };
+  const { hasInstalledImage, imagesToInstall } = mapImages();
+
+  if (imagesToInstall.length === 0) {
+    message.warning(t('environment.AlreadyInstalledImage'));
+    onRequestClose();
+    return null;
+  }
+
+  const handleClick = () => {
```
gahyuun, line 39: handleClick references the _installImage function in the lit code 
https://github.com/lablup/backend.ai-webui/blob/main/src/components/backend-ai-environment-list.ts#L618

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,369 @@
...
+    [images],
+  );
+  const columns: ColumnsType<EnvironmentImage> = [
+    {
+      title: t('environment.Status'),
+      dataIndex: 'installed',
+      key: 'installed',
+      sorter: sortBasedOnInstalled,
+      render: (text, row) => {
+        if (row?.id && installingImages.includes(row.id))
```
gahyuun: I used the installingImages state to show the installing tag

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,369 @@
...
+            }}
+          >
+            {t('environment.Install')}
+          </Button>
+        </Flex>
+        <Table
+          rowKey="id"
+          scroll={{
+            x: 'max-content',
+            y: 'calc(100vh - (9.0625rem + 3rem + 4.5rem))',
```
gahyuun, line 327: The reason I specified scroll on the y-axis is that I thought the colum header and install button should be visible while scrolling.

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,369 @@
...
+  };
+
+  const sortBasedOnInstalled = (a: EnvironmentImage, b: EnvironmentImage) => {
+    return a?.installed && !b?.installed
+      ? -1
+      : !a?.installed && b?.installed
+        ? 1
+        : 0;
+  };
+
```
gahyuun: https://github.com/lablup/backend.ai-webui/blob/main/src/components/backend-ai-environment-list.ts#L461
lit code for reference

react/src/components/ImageTags.tsx
```diff
@@ -85,48 +84,49 @@ interface LangTagsProps extends TagProps {
 }
 export const LangTags: React.FC<LangTagsProps> = ({ image, ...props }) => {
   image = image || '';
-  const [, { getImageLang, tagAlias }] = useBackendAIImageMetaData();
+  const [, { getImageLang }] = useBackendAIImageMetaData();
   return (
     <Tag color="green" {...props}>
-      {tagAlias(getImageLang(image))}
+      {getImageLang(image)}
```
gahyuun, line 90: The reason for removing tagalias is that sometimes it is necessary to show the language without using the LangTags component, so I think it is better to include tagAlias inside the getImageLang function.

resources/image_metadata.json
gahyuun, line 1: While checking ImageList, I added the parts that didn't have metaData.

react/src/hooks/index.tsx
```diff
@@ -251,7 +260,49 @@ export const useBackendAIImageMetaData = () => {
         const { tags } = getImageMeta(imageName);
         return tags[1];
       },
+      getBaseImages: (tag: string, name: string) => {
```
gahyuun, line 268: `getBaseImages`,`getConstraints`,`getLangs` implemented a new function without modifying the `getBaseImage`,`getCustomTag`,`getFilteredRequirementsTag`,`getImageLang` because of side effects ( ex. `SessionKernelTags` uses `getBaseImage`. )
Also, because of the side effect, I didn't modify `MyEnvironmentPage` to any values except the constraints. I think it's better to do this in a new task.

react/src/hooks/index.tsx
gahyuun, line 1: https://github.com/lablup/backend.ai-webui/blob/main/src/components/backend-ai-environment-list.ts#L345
lit code for reference

ironAiken2 (changes requested):
Please resolve conflict :)

react/src/hooks/useBackendAIImageMetaData.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+            r: 'R',
+            'r-base': 'R',
+            py3: 'Python3',
+            ngc: 'Nvidia GPU Cloud',
+            py310: 'Python3',
+          },
+          tagReplace: {},
+        }),
+    }),
+  );
```
gahyuun, line 31: I used jest.fn because suspense can't wait for the test to complete while fetching data

react/src/pages/EnvironmentPage.tsx
```diff
@@ -50,12 +51,15 @@ const EnvironmentPage = () => {
     >
       <Flex
         style={{
-          display: curTabKey === 'image' ? 'block' : 'none',
-          paddingTop: token.paddingContentVerticalSM,
+          display: 'block',
         }}
       >
```
ironAiken2: This seems to be a remnant of code written to manage the display of the web component, how about removing this line?

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+    return { hasInstalledImage, imagesToInstall };
+  };
+  const { hasInstalledImage, imagesToInstall } = mapImages();
+
+  if (imagesToInstall.length === 0) {
+    message.warning(t('environment.AlreadyInstalledImage'));
+    onRequestClose();
+    return null;
+  }
```
ironAiken2: If the modal's openness were managed inside the modal, it would consume unnecessary resources, and since it is not the role of the component, it seems better to manage it in the parent component.

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,369 @@
...
+          ...style,
+        }}
+      >
+        <Flex justify="end" style={{ padding: token.paddingSM }}>
+          <Button
+            icon={<VerticalAlignBottomOutlined />}
+            style={{ backgroundColor: token.colorPrimary, color: 'white' }}
+            onClick={() => {
+              setIsOpenInstallModal(true);
+            }}
```
ironAiken2: [related comment](https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2615/feat-migrate-image-list#comment-PRRC_kwDOCRTcws5mG_cB)

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const { upsertNotification } = useSetBAINotification();
+  const painKiller = usePainKiller();
+
+  if (!open) return null;
+
+  if (selectedRowKeys.length === 0) {
+    message.warning(t('environment.NoImagesAreSelected'));
+    onRequestClose();
+    return null;
+  }
```
ironAiken2: Same as [related comment](https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2615/feat-migrate-image-list#comment-PRRC_kwDOCRTcws5mG_cB)

react/src/components/ImageInstallModal.tsx
gahyuun: However, if the parent component manages the determination of whether an image is already installed, then mapImages() must be implemented in the parent component as well. 
What do you think about this resource?

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+        .map((item) => item?.id)
+        .filter((id): id is string => id !== null && id !== undefined),
+    );
+    setSelectedRowKeys([]);
+  };
+
+  return (
+    <BAIModal
+      destroyOnClose
+      open={open}
```
ironAiken2:
```suggestion

```

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+  setSelectedRowKeys: Dispatch<SetStateAction<Key[]>>;
+}
+export default function ImageInstallModal({
+  open,
+  onRequestClose,
+  selectedRowKeys,
+  images,
+  setInstallingImages,
+  setSelectedRowKeys,
+}: ImageInstallModalInterface) {
```
ironAiken2:
```suggestion
export default function ImageInstallModal({
  onRequestClose,
  selectedRowKeys,
  images,
  setInstallingImages,
  setSelectedRowKeys,
  ...modalProps
}: ImageInstallModalInterface) {
```

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+  return (
+    <BAIModal
+      destroyOnClose
+      open={open}
+      maskClosable={false}
+      onCancel={() => onRequestClose()}
+      title={t('dialog.title.LetsDouble-Check')}
+      okText={t('environment.Install')}
+      onOk={handleClick}
+    >
```
ironAiken2, line 153:
```suggestion
      onOk={handleClick}
      {...modalProps}
    >
```

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { useTranslation } from 'react-i18next';
+
+interface ImageInstallModalInterface extends BAIModalProps {
+  open: boolean;
+  onRequestClose: () => void;
+  selectedRowKeys: Key[];
+  images: EnvironmentImage[];
+  setInstallingImages: Dispatch<SetStateAction<string[]>>;
+  setSelectedRowKeys: Dispatch<SetStateAction<Key[]>>;
+}
```
ironAiken2: If you declare a new interface by extending antd's default modal, the fields included in the existing modal interface can be removed from the declaration and also other fields in the main modal interface are also available.[[related comment]](https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2615/feat-migrate-image-list#comment-PRRC_kwDOCRTcws5mHAhD) [[related comment]](https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2615/feat-migrate-image-list#comment-PRRC_kwDOCRTcws5mHAoW) [[related comment]](https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2615/feat-migrate-image-list#comment-PRRC_kwDOCRTcws5mHAtJ)

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+              ),
+          );
+          // @ts-ignore
+          globalThis.lablupNotification.text = painKiller.relieve(error.title);
+          // @ts-ignore
+          globalThis.lablupNotification.detail = error.message;
+          // @ts-ignore
+          globalThis.lablupNotification.show(true, error);
+          indicator.set(100, t('environment.DescProblemOccurred'));
+          indicator.end(1000);
```
ironAiken2: Is there a reason why you didn't use `upsertNotification` in this part?

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+        .then(() => {
+          indicator.end(1000);
+        })
+        .catch((error: any) => {
+          setInstallingImages(
+            imagesToInstall
+              .map((item) => item?.id)
+              .filter(
+                (id): id is string =>
+                  id !== null && id !== undefined && id !== image?.id,
```
ironAiken2, line 146: [ref](https://github.com/lablup/backend.ai-webui/pull/2615/files#diff-8ee05aa0711da52a475b27ba894a607046f46a5c4cf64903702679637cefbef4R141-R148) Isn't this duplicated?

react/src/components/ImageInstallModal.tsx
```diff
@@ -0,0 +1,197 @@
...
+          globalThis.lablupNotification.show(true, error);
+          indicator.set(100, t('environment.DescProblemOccurred'));
+          indicator.end(1000);
+        });
+    });
+    setInstallingImages(
+      imagesToInstall
+        .map((item) => item?.id)
+        .filter((id): id is string => id !== null && id !== undefined),
+    );
```
ironAiken2: Isn't it duplicated with [ref](https://github.com/lablup/backend.ai-webui/pull/2615/files#diff-8ee05aa0711da52a475b27ba894a607046f46a5c4cf64903702679637cefbef4R159-R164)? And it seems that the problem will occur if you download another image additionally without refreshing the page.

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,369 @@
...
+        <Table
+          rowKey="id"
+          scroll={{
+            x: 'max-content',
+            y: 'calc(100vh - (9.0625rem + 3rem + 4.5rem))',
+          }}
+          pagination={false}
+          dataSource={sortedImages}
+          columns={columns as ColumnType<AnyObject>[]}
+          rowSelection={rowSelection}
```
ironAiken2:
```suggestion
          rowSelection={{
            type: 'checkbox',
            onChange: (_, selectedRows: DataType[]) => {
              setSelectedImage(selectedRows);
            },
          }}
```

I think we could save the selected image as a state, and pass only that image to installmodal, to reduce unnecessary filtering logic.

react/src/components/ImageInstallModal.tsx
gahyuun:
```
 <BAIModal
      destroyOnClose
      open
..
```
Is this code suggested?

react/src/components/FlexActivityIndicator.tsx
gahyuun, line 1: I modified it because I wanted to resize the spinner

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,414 @@
...
+        display: 'flex',
+        alignItems: 'center',
+        height: '100%',
+        ...style,
+      }}
+    >
+      {children}
+    </div>
+  );
+};
```
gahyuun: In Ant Design's Table component, when using virtual mode and specifying a scroll.y value, the vertical center alignment does not seem to be applied correctly because the cells have a fixed height. 
So, I implemented a component called CellWrapper and did the vertical alignment myself.  

Since I was doing a custom render, the ellipsis property was not applied, so I applied ellipsis directly as well

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,414 @@
...
+            }}
+          >
+            {t('environment.Install')}
+          </Button>
+        </Flex>
+        <Table<EnvironmentImage>
+          rowKey="id"
+          scroll={{
+            x: 1600,
+            y: window.innerHeight - 145 - 56 - 54,
```
gahyuun: 145: ref ) https://github.com/lablup/backend.ai-webui/pull/2615/files#diff-7513e7364d1bbef624ab929ba47fa51f2fcb5c0090f36db4f56e3837bb29bac5R78

56 : The space where the install button exists 

54: column title of table

react/src/components/ImageInstallModal.tsx
gahyuun: - The way to display the error in upsertNotification seemed to be to set the backgroundTask status to rejected.
But there is no backgroundTask.
- I also referenced the container Registry List code 
ref) https://github.com/lablup/backend.ai-webui/blob/main/react/src/components/ContainerRegistryList.tsx#L159

react/src/components/ImageInstallModal.tsx
gahyuun: I modified the rendering to only show the successful API image after clicking the install button. 
I made the requests in parallel and only updated the state of the successful images.  

https://github.com/lablup/backend.ai-webui/pull/2615/commits/ded9e4901c466f945092ca04d01f11ab7cd34d21

rapsealk:
Please resolve the conflicts!

yomybaby (changes requested):
I'm working to resolve blow list:
- [ ] Make the entire row clickable.
- [x] Resource Limitation Column: The rows in this table do not have a fixed height, so 'virtual' is not appropriate. Changing to a fixed height is not possible due to the resource limitation UI column.
- [ ] Base and Constraint Columns: Display them in one line.
- [ ] When you modify an image, the order can be changed. To prevent this, let's set the order using 'humanized_name'.

I'll share soon.

react/src/components/ImageInstallModal.tsx
yomybaby: @gahyuun  we can change this part using `upsertNotification` with `promise`. This is an [example](https://github.com/lablup/backend.ai-webui/blob/81c1dacb35a17b74da5c1e053097f511fd9e382d/react/src/pages/SessionLauncherPage.tsx#L502-L515).

yomybaby (changes requested):
I got an error when I try to update using Manage app modal.  Please check it again.

yomybaby (approved):
LGTM

react/src/components/ImageList.tsx
```diff
@@ -0,0 +1,512 @@
...
+              if (selectedRows.find((row) => row.id === record.id)) {
+                setSelectedRows((rows) =>
+                  rows.filter((row) => row.id !== record.id),
+                );
+              } else {
+                setSelectedRows((rows) => [...rows, record]);
+              }
+            },
+          })}
+          showSorterTooltip={false}
```
ironAiken2, line 475:
```suggestion
          showSorterTooltip={false}
          sortDirections={['descend', 'ascend', 'descend']}
```
Logic to limit data sorting to ascending and descending order is not needed in the webui?

ironAiken2 (approved):
LGTM

PR #2637, opened 2024-08-16
feat: implement image list filtering and refresh

gahyuun:
### TL;DR

Implement image list filtering and refresh

###  Why make this change?

To add filtering and refreshing to the list of images on the `EnvironmentPage`, 

### What changed?

- search
When searching for images, filteredImageData checks if the processed data (the data you see in the table) matches your search terms and render with a TextHighlighter 

- refresh 
When you click the refresh button, use the environmentkey to get the data again

### How to test?

- search 
Retrieve the image and verify that it was rendered correctly  

- refresh
After editing the information on the other tabs, make sure that the changed information is displayed when you click Refresh.


react/src/components/ImageList.tsx
```diff
@@ -320,6 +363,57 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
...
+            getNamespace(getImageFullName(image) || ''),
+          );
+          return (
+            baseVersionMatch ||
+            baseImagesMatch ||
+            constraintsMatch ||
+            langMatch ||
+            namespaceMatch ||
+            customizedMatch
+          );
```
gahyuun: If the `filteredImageData` doesn't go through the process of comparing it to processed data rather than image data, the `imageList` will render the wrong data

react/src/components/ImageList.tsx
```diff
@@ -252,6 +282,15 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
...
               labels={row?.labels as { key: string; value: string }[]}
+              wrapper={(constraint: unknown) =>
+                typeof constraint === 'string' ? (
+                  <TextHighlighter keyword={imageSearch}>
+                    {constraint}
+                  </TextHighlighter>
+                ) : (
+                  <></>
+                )
+              }
```
gahyuun: The TextHighlighter component needs to wrap the value, but the constraint column is using a constraint tag, so I was wondering how to extend it. I solved it by adding a wrapper attribute, but I'd like to get some feedback on what you think would be a better way.

react/src/components/ImageList.tsx
```diff
@@ -330,7 +424,27 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
...
         }}
       >
-        <Flex justify="end" style={{ padding: token.paddingSM }}>
+        <Flex justify="end" style={{ padding: token.paddingSM, gap: 8 }}>
+          <Input
+            allowClear
+            prefix={<SearchOutlined />}
+            placeholder={t('environment.SearchImages')}
+            onChange={(e) => {
+              startTransition(() => setImageSearch(e.target.value));
```
gahyuun: I didn't use loading when updating information by search or control. I thought the updates were fast enough that I didn't need to use loading, and I found the UI with loading to be more cluttered.

react/src/components/DoubleTag.tsx
```diff
@@ -11,7 +11,8 @@ export type DoubleTagObjectValue = {
 type ValueType = string | React.ReactNode;
 const DoubleTag: React.FC<{
   values?: ValueType[] | DoubleTagObjectValue[];
-}> = ({ values = [] }) => {
+  wrapper?: (value: unknown) => React.ReactNode;
+}> = ({ values = [], wrapper }) => {
```
yomybaby: It's better to composite component without adding `wrapper` prop. The `ValueType` can be a React node. 
Something like this
```
<Doubletags values= {[ label: <TextHighlighter>{somthing}</TextHighlighter>]} .. />
```

react/src/components/ImageTags.tsx
```diff
@@ -95,10 +95,12 @@ export const LangTags: React.FC<LangTagsProps> = ({ image, ...props }) => {
 interface ConstraintTagsProps extends TagProps {
   tag: string;
   labels: { key: string; value: string }[];
+  wrapper?: (constraint: unknown) => React.ReactNode;
```
yomybaby: For `ConstraintTagsProps`, it would be better to add a `highlightKeyword` prop and use `TextHighlighter` here.

react/src/components/ImageList.tsx
yomybaby: Please add a loading indicator to the table when searching and updating the table. It's better if there is a possibility of the UI displaying inconsistent data.

yomybaby (changes requested):
`

react/src/components/ImageTags.tsx
gahyuun: I modified it https://github.com/lablup/backend.ai-webui/pull/2637/commits/5d1e367da437a15159cab6474b23d992c56895f3
and I have a question about the loading UI. 
It shows search results so fast that the loading UI is invisible. So I don't know the need for loading UI.
I wonder if the reason for using the loading UI is for when there is a lot of data.

yomybaby: > It shows search results so fast that the loading UI is invisible. 

Yes, you are right.  However, if a specific column is suspensive, loading may take a long time at any time. Therefore, I think there is no need to remove the indication of loading.

gahyuun: Thank you for reviewing . I modified it

react/src/components/ImageList.tsx
yomybaby: Every `imageSearch` change generates text for filter comparison. It would be better to separate the following concerns:
1. `imageFilterValues`: text for filter comparison when `images` change.
2. `filteredImages`: filtered images based on `imageSearch` and `imageFilterValues`.

Somthing like this:
```
const imageFilterValues = useMemo(()=>{

},[image]);
const filteredImages = useMemo(()=>{
  return _.filter(images, (img, idx)=>{
    // check imageFilterValues[idx] has imageSearch
  })
},[imageFilterValues, imageSearch]);

<Table dataSource={filteredImages}/>
```

yomybaby (changes requested):
I pushed commits and add a comment. Please check it

react/src/components/ImageList.tsx
gahyuun: Thank you for reviewing  I modified it
https://github.com/lablup/backend.ai-webui/pull/2637/commits/3edc4c6983718a23133d91367d15595443cd60c0

yomybaby (approved):
I pushed a commit to refactor codes related to filtering and ordering. Please check it before merge this.
LGTM now.

react/src/components/ImageList.tsx
```diff
@@ -319,53 +325,39 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
...
+      };
+    });
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [images]);
+  }, [defaultSortedImages]);
 
   const filteredImageData = useMemo(() => {
-    const { sortedImages, imageFilterValues } = processedImages;
-    if (_.isEmpty(imageSearch)) return sortedImages;
+    if (_.isEmpty(imageSearch) || imageFilterValues === undefined)
```
gahyuun: Is there a case that ImageFilterValues is undefined?

yomybaby: Yes, you are right. But we need to handle some `undefined` cases. Please check my latest commit.

PR #2648, opened 2024-08-21
fix: `pnpm install` does not use `/react/.npmrc`.

yomybaby:
### TL;DR

Updated the postinstall script in package.json to use pnpm recursive install for react dependencies. Before this PR, when you run `pnpm install` in the `/react` directory, you would see the following logs.

> ? The modules directory at "/Users/_______/webui/react/node_modules" will be removed and reinstalled from scratch. Proceed? (Y/n) › true

### What changed?

The postinstall script in package.json has been modified. The command to install react dependencies has been changed from `cd react && pnpm install` to `pnpm recursive install --filter ./react`.

### How to test?

1. Delete the node_modules folder in the project root and the react subdirectory.
2. Run `pnpm install` in the project root.
3. Verify that all dependencies, including those in the react subdirectory, are installed correctly.
4. Run `pnpm install` in the `/react` directory to verify that it uses `react/.npmrc`. If there are no errors, it means that it works properly.
5. Ensure that the project builds and runs without any errors related to missing dependencies.

### Why make this change?

This change improves the dependency installation process by using pnpm's recursive install feature with a filter. It allows for a more efficient and targeted installation of dependencies in the react subdirectory without changing the working directory. This approach is more consistent with pnpm best practices and can potentially lead to faster and more reliable installations, especially in monorepo setups.


yomybaby:
* **#2648** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @yomybaby and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

ironAiken2 (changes requested):
Installing the package does not cause any errors, but running a browser with `pnpm run build:d` and `pnpm run server:d` will cause the error `[eslint] Failed to load config "react-app" to extend from.`. This error is resolved by updating the package with the `pnpm upgrade` command.

yomybaby:
@ironAiken2  I added two more dev dependencies to `/react`. Please don't forget a below guide before you test again.
> Delete the node_modules folder in the project root and the react subdirectory.

ironAiken2 (approved):
LGTM!

PR #2659, opened 2024-08-23
fix: modify the locale of reservation time to the language of Backend.AI

gahyuun:
### TL;DR

modify the locale of reservation time to the language of Backend.AI

###  Why make this change?

The locale of the reservation time in the session list is the OS default language, not the Backend.AI set language. 
So when the OS default is Korean, no matter how I set the language, the reservation time displays Korean. 
image

### What changed?

Set the toLocaleString function's parameter Locales to the general.language of Backend.AI

### How to test?
1. change language to ko  in settings
2. Verify reservation time looks like this `2024. 8. 21. 오전 2:45:26`
3. change language to en in settings
4. Verify reservation time looks like this `8/21/2024, 2:45:26 AM`

### DIFF
--- a/src/components/backend-ai-session-list.ts
+++ b/src/components/backend-ai-session-list.ts
@@ -1366,7 +1366,9 @@ export default class BackendAISessionList extends BackendAIPage {
    */
   _humanReadableTime(d: any) {
     d = new Date(d);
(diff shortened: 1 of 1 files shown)


agatha197 (approved):
LGTM

PR #2665, opened 2024-08-28
e2e: test NEO session launcher and imageList 

gahyuun:
### TL;DR

e2e test for NEO session creation and `imageList`

### What changed?
Test:
 - create and terminate a session without setting anything except the session name.  
 - rendering image list
 - install image
 - modify image resource limit
 - manage apps

### DIFF
--- a/e2e/environment.test.ts
+++ b/e2e/environment.test.ts
@@ -1,0 +1,190 @@
+import { loginAsAdmin, navigateTo } from './test-util';
+import { findColumnIndex } from './test-util-antd';
+import { expect, test } from '@playwright/test';
+
+test.describe('environment ', () => {
+  let imageListTable;
+  test.beforeEach(async ({ page }) => {
+    imageListTable = await page.locator('.ant-table');
+  });
+  test('Rendering Image List', async ({ page }) => {
+    await loginAsAdmin(page);
+    await navigateTo(page, 'environment');
+    const table = page.locator('.ant-table-content');
+    await expect(table).toBeVisible();
+  });
+
+  // skip this test because there is no way to uninstall the image in WebUI
+  test.skip('user can install image', async ({ page }) => {
+    await loginAsAdmin(page);
+    await navigateTo(page, 'environment');
+    await expect(imageListTable).toBeVisible();
+    // Sort installation status
+    await page.locator('.ant-table-cell.ant-table-column-sort').first().click();
+
+    // Find uninstalled item and select
+    const uninstalledImage = page
+      .locator('.ant-table-cell.ant-table-column-sort')
+      .filter({
+        hasNot: page.locator('.ant-tag-gold'),
+      })
+      .nth(1);
+    // If all images are installed, skip the test
+    const count = await uninstalledImage.count();
+    if (count === 0) {
+      test.skip();
+    }
+    await uninstalledImage.click();
+
+    // Install image
+    await page
+      .getByRole('button', { name: 'vertical-align-bottom Install' })
+      .click();
+    await page.getByRole('button', { name: 'Install', exact: true }).click();
+    await expect(
+      page.getByText('It takes time so have a cup of coffee!'),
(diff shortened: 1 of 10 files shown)


e2e/test-util.ts
```diff
@@ -94,3 +94,43 @@ export async function deleteVFolderAndVerify(page: Page, folderName: string) {
...
+  const sessionNameInput =  page.locator('#sessionName');
+  await sessionNameInput.fill(sessionName);
+  await page.getByRole("button",{name:"Skip to review"}).click();
+
+  await page.locator("button").filter({hasText:"Launch"}).click();
+  await expect(page.locator(".ant-modal-confirm-title")).toHaveText("No storage folder is mounted");
+  await page.getByRole("button",{name:"Start"}).click();
+
+  // Wait for a session to be launched
+  await page.waitForTimeout(10000);
```
gahyuun: It was difficult to me that the sessions created were not immediately reflected on the screen.

So I tried a few things

1. refresh ( use `navigateTo`) -> Not suitable for user scenario testing.
2. “Create session with session name e2e test -> Recreate session with session name e2e test. Check the failure statement that a session has already been created”
I tried to run the test with the above scenario. But It isn't a good fit for user scenario testing either.
3. Increasing the `waitForTimeout` time

I thought option 3 works best and I applied that method

yomybaby: Instead of page.waitForTimeout(10000), we can use locator's wait. Please check my recent commit.
```
await expect(session).toBeVisible({timeout: 10000});
```

e2e/environment.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      .first();
+    await expect(installingItem.getByText('installing')).toBeVisible();
+  });
+
+  test('user can modify image resource limit', async ({ page }) => {
+    await loginAsAdmin(page);
+    await navigateTo(page, 'environment');
+    // Click resource limit button
+    await page
+      .locator('.ant-table-row > td:nth-child(12)')
```
yomybaby: Please use `findColumnIndex` function instead of `css selector`

e2e/environment.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      await page.getByRole('button', { name: 'OK' }).nth(1).click();
+    }
+  });
+
+  test('user can manage apps', async ({ page }) => {
+    await loginAsAdmin(page);
+    await navigateTo(page, 'environment');
+    // Click manage apps button
+    await page
+      .locator('.ant-table-row > td:nth-child(12)')
```
yomybaby: Please use `findColumnIndex` here too.

e2e/environment.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      .first()
+      .getByRole('button')
+      .nth(1)
```
yomybaby: It's better to use Buttons's icon for finding the button.
```
<span role="img" aria-label="setting" 
```

yomybaby (changes requested):
Please check my comments and commits.

e2e/environment.test.ts
gahyuun: I modified it 
https://github.com/lablup/backend.ai-webui/pull/2665/commits/3467f2a0a3a7d6991acb40db33574fa021e7c060

e2e/test-util.ts
gahyuun: I check it. Create session returns a passed.  But session test fails. Because the timeout (`await page.waitForTimeout(5000);`)was removed and the session termination was not immediately reflected on the screen. 
So I think that Delete session test needs Timeout

yomybaby (approved):
I pushed a [commit](https://github.com/lablup/backend.ai-webui/pull/2665/commits/630ca503b78cf7404b4a7843ecf3c6657bb55b09) please check it. 
Now, LGTM! Thank you for your contribution.

PR #2687, opened 2024-09-04
fix: modification of "model service's resource"

yomybaby:
### TL;DR

This PR resolves the issue where the endpoint's resource allocation values are not displayed when modifying the model service, and the first allocatable preset is automatically selected.

### What changed?
The logic was changed as follows:
- case 1: `allowCustomResourceAllocation = false`
  - Hide resource group select and preset select. and display current value like below: 

  - If the user click the "Change" button, the preset select component is shown with automatic selection of preset
- case 2: `allowCustomResourceAllocation = true`
  - When modifying, the `allocationPreset` field is set to custom.

### DIFF
--- a/react/src/components/ResourceAllocationFormItems.tsx
+++ b/react/src/components/ResourceAllocationFormItems.tsx
@@ -106,6 +106,10 @@ const ResourceAllocationFormItems: React.FC<
     form,
     preserve: true,
   });
+  const currentAllocationPreset = Form.useWatch(['allocationPreset'], {
+    form,
+    preserve: true,
+  });
   const [{ currentImageMinM, remaining, resourceLimits, checkPresetInfo }] =
     useResourceLimitAndRemaining({
       currentProjectName: currentProject.name,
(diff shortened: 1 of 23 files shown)


yomybaby:
* **#2687** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @yomybaby and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

ironAiken2 (changes requested):
It works fine, but it seems like there needs to be some consideration for when the maximum quota for a resource group has changed.

[If the maximum allocatable amount is reduced by a resource policy from what it was before].

yomybaby:
In that case, the user must change the value to be smaller than the slider max. Is the current behavior not like that?

ironAiken2 (changes requested):
Yes, that's how the behavior is implemented now. 

However, if the user doesn't modify the resource, the request can be sent with an existing set value (greater than the resource policy).

and please resolve the conflict!

ironAiken2 (approved):
LGTM

PR #2743, opened 2024-10-10
feat: name-based query search feature on `EndpointSelect`

ironAiken2:
**Changes:**

The current EndpointList only fetches up to 100 endpoints via gql, and if you have more than 100 endpoints, it is not possible to search the entire data.

This PR enhances the `EndpointSelect` component to allow searching over the entire data using gql filters:

1. Introduced `useState` and `useTransition` hooks for managing search state and transitions.
2. Added a search filter to the GraphQL query when the backend supports 'endpoint-lifecycle-stage-filter'.
3. Implemented `onSearch` handler for the Select component to update the search string.
4. Added loading indicator during search transitions.

**How to test**
  1. move to LLM Playground page.
  2. Verify that endpoint select requests a new gql based on search string

**Rationale:**

These changes improve the user experience by allowing users to search for specific endpoints within the select component, especially useful when dealing with a large number of endpoints.

**Effects:**

- Users can now search for endpoints by name in the select dropdown.
- The component will show a loading state while searching, providing visual feedback.
- The search is performed on the backend, potentially improving performance for large datasets.

**Checklist:**

- [ ] Documentation

### DIFF
--- a/react/src/components/EndpointSelect.tsx
+++ b/react/src/components/EndpointSelect.tsx
@@ -1,5 +1,4 @@
 import { useSuspendedBackendaiClient } from '../hooks';
-import { useBAIPaginationOptionState } from '../hooks/reactPaginationQueryOptions';
 import {
   EndpointSelectQuery,
   EndpointSelectQuery$data,
@@ -8,33 +7,46 @@ import { useControllableValue } from 'ahooks';
 import { Select, SelectProps } from 'antd';
 import graphql from 'babel-plugin-relay/macro';
 import _ from 'lodash';
-import React from 'react';
+import React, { useState, useTransition } from 'react';
 import { useLazyLoadQuery } from 'react-relay';
 
 interface EndpointSelectProps extends Omit<SelectProps, 'options'> {
   fetchKey?: string;
+  lifecycleStageFilter?: LifecycleStage[];
 }
 
 export type Endpoint = NonNullableItem<
   EndpointSelectQuery$data['endpoint_list']
 >;
 
+type LifecycleStage = 'created' | 'destroying' | 'destroyed';
+
 const EndpointSelect: React.FC<EndpointSelectProps> = ({
   fetchKey,
-  ...selectProps
+  lifecycleStageFilter = ['created'],
+  loading,
+  ...selectPropsWithoutLoading
 }) => {
   const baiClient = useSuspendedBackendaiClient();
-  const { baiPaginationOption } = useBAIPaginationOptionState({
-    current: 1,
-    pageSize: 100,
-  });
-  const { endpoint_list } = useLazyLoadQuery<EndpointSelectQuery>(
+  const [controllableValue, setControllableValue] = useControllableValue(
+    selectPropsWithoutLoading,
+  );
+  const [searchStr, setSearchStr] = useState<string>();
+  const [isSearchPending, startSearchTransition] = useTransition();
+
+  const lifecycleStageFilterStr = lifecycleStageFilter
+    .map((v) => `lifecycle_stage == "${v}"`)
+    .join(' | ');
+
+  const { endpoint_list, endpoint } = useLazyLoadQuery<EndpointSelectQuery>(
     graphql`
       query EndpointSelectQuery(
         $offset: Int!
         $limit: Int!
         $projectID: UUID
         $filter: String
+        $endpoint_id: UUID!
+        $skipEndpoint: Boolean!
       ) {
         endpoint_list(
           offset: $offset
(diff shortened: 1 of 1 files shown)


ironAiken2:
* **#2743** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2743?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

react/src/components/EndpointSelect.tsx
```diff
@@ -24,6 +24,8 @@ const EndpointSelect: React.FC<EndpointSelectProps> = ({
   ...selectProps
 }) => {
   const baiClient = useSuspendedBackendaiClient();
+  const [searchStr, setSearchStr] = useState<string>();
+  const [isSearchPending, startSearchTransition] = useTransition();
   const { baiPaginationOption } = useBAIPaginationOptionState({
     current: 1,
     pageSize: 100,
```
yomybaby: Please remove unnecessary code such as pagination hook. You can set `limit` and `skip` directly.  Also 100 is too big. Let change it to 10.

react/src/components/EndpointSelect.tsx
```diff
@@ -56,7 +58,12 @@ const EndpointSelect: React.FC<EndpointSelectProps> = ({
       limit: baiPaginationOption.limit,
       offset: baiPaginationOption.offset,
       filter: baiClient.supports('endpoint-lifecycle-stage-filter')
-        ? `lifecycle_stage == "created"`
+        ? [
+            `(lifecycle_stage == "created")`,
+            searchStr && `(name ilike "%${searchStr}%")`,
+          ]
+            .filter(Boolean)
+            .join(' & ')
```
yomybaby: One more condition is needed. You also need to fetch the endpoint information for `controllableValue`. Without this, this component may display `endpoint_id` instead of the endpoint's name if the endpoint not in the first page of query.

yomybaby (changes requested):

react/src/components/EndpointSelect.tsx
ironAiken2: Currently, the endpoint_list query does not provide filtering based on endpoint_id, but the controllableValue is passing id. Is backend.ai core need to add endpoint_id to the filter options? Or should I change it to descend the name from the parent?@yomybaby

yomybaby: There is another option to query the endpoint for `value`. It's adding the `endpoint(endpoint_id: UUID!): Endpoint` query to the existing `EndpointSelectQuery`. Don't forget to use `@skip` to skip the `endpoint` query when the `value` is undefined.

react/src/components/EndpointSelect.tsx
```diff
@@ -50,21 +52,31 @@ const EndpointSelect: React.FC<EndpointSelectProps> = ({
...
         : undefined,
+      endpoint_id: controllableValue,
+      skipEndpoint: !!!controllableValue,
```
yomybaby:
```suggestion
      skipEndpoint: !controllableValue,
```

react/src/components/EndpointSelect.tsx
```diff
@@ -50,21 +52,31 @@ const EndpointSelect: React.FC<EndpointSelectProps> = ({
...
+      limit: 10,
+      offset: 0,
       filter: baiClient.supports('endpoint-lifecycle-stage-filter')
-        ? `lifecycle_stage == "created"`
+        ? [
+            `(lifecycle_stage == "created")`,
+            searchStr && `(name ilike "%${searchStr}%")`,
+          ]
+            .filter(Boolean)
+            .join(' & ')
```
yomybaby: 1. Please wrap with `()` before join like below:
```
              .filter(Boolean)
              .map((v) => `(${v})`)
              .join(' & ')
```
2. Please add the `filter` prop to this component. Its default value is `lifecycle_stage == "created"` for more usability.

yomybaby (changes requested):

agatha197 (changes requested):
I got runtime errors :(

react/src/components/EndpointSelect.tsx
```diff
@@ -50,21 +52,31 @@ const EndpointSelect: React.FC<EndpointSelectProps> = ({
             ...EndpointLLMChatCard_endpoint
           }
         }
+        endpoint(endpoint_id: $endpoint_id) @skip(if: $skipEndpoint) {
```
agatha197, line 65:
```suggestion
        endpoint(endpoint_id: $endpoint_id) @skipOnClient(if: $skipEndpoint) {
```

agatha197 (changes requested):

agatha197 (approved):
LGTM

react/src/components/EndpointSelect.tsx
yomybaby: In this case, it's better to use `@skip`. `@skipOnClient` is a workaround to maintain compatibility for multiple schema versions. @agatha197  @ironAiken2  I'll edit and merge this PR.

yomybaby (approved):
I push a commit. @ironAiken2  Please check it. 
LGTM, now. Thank you.

agatha197:
> I push a commit. @ironAiken2 Please check it. LGTM, now. Thank you.

It still causes conflict.

yomybaby (commented):
Actually, we need to implement infinite scroll when the option list is scrolled to the end. However, it's easier to implement it after the `endpoint` query meets the Relay spec. Let's handle it later in another PR.

PR #2762, opened 2024-10-21
fix: hide "compare with other models" in LLM Playground

yomybaby:
**Changes:**
Hide "Compare with Other Models" menu item" in LLM playground:

- Added `showCompareMenuItem` prop to `LLMChatCard` component
- Implemented conditional rendering of "Compare with Other Models" menu item based on `showCompareMenuItem` prop
- Utilized `filterEmptyItem` helper function to remove empty items from the menu

**Rationale:**

This change allows for more flexible control over the visibility of the "Compare with Other Models" menu item in the LLM Chat interface. By making it conditional, we can show or hide this option based on the context or user permissions.

**Effects:**

- Developers can now control the visibility of the comparison feature by passing the `showCompareMenuItem` prop
- Users may or may not see the "Compare with Other Models" option depending on how the component is configured

PR #2771, opened 2024-10-23
fix: hide divider in LLMChatCard

ironAiken2:
This PR is for the conditional divider render that is missing from the https://github.com/lablup/backend.ai-webui/pull/2762 PR.

|before|after|
|---|---|
|||

- [ ] Documentation


yomybaby (approved):
LGTM

PR #2778, opened 2024-10-28
hotfix: container registry password caanot be set when creating

agatha197:
related PR: https://github.com/lablup/backend.ai-control-panel/pull/892

**Changes:**

This PR modifies the password handling logic in the `ContainerRegistryEditorModal` component. The condition for setting the password has been updated to include cases where a new container registry is being created (`!containerRegistry`). This ensures that the password is properly set for both new and existing registries.

**Rationale:**

The previous implementation only considered the `isChangedPassword` flag, which could lead to issues when creating a new container registry. By including the `!containerRegistry` condition, we ensure that the password is always set correctly for new registries, even if the `isChangedPassword` flag is not explicitly set.

**Impact:**

This change improves the reliability of password handling when creating or updating container registries. Users should now experience more consistent behavior when managing registry credentials.

**Checklist:**

  - Create a new container registry and verify that the password is set correctly
  - Update an existing container registry with and without changing the password

### DIFF
--- a/react/src/components/ContainerRegistryEditorModal.tsx
+++ b/react/src/components/ContainerRegistryEditorModal.tsx
@@ -110,11 +110,12 @@ const ContainerRegistryEditorModal: React.FC<
           type: values.type,
           project: values.type === 'docker' ? 'library' : values.project,
           username: _.isEmpty(values.username) ? null : values.username,
-          password: values.isChangedPassword
-            ? _.isEmpty(values.password)
-              ? null // unset
-              : values.password
-            : undefined, // no change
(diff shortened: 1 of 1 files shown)


agatha197:
### Merge activity

* **Oct 27, 9:25 PM EDT**: The merge label 'flow:hotfix' was detected. This PR will be added to the [Graphite merge queue](https://app.graphite.dev/merges?org=lablup&repo=backend.ai-webui) once it meets the requirements.

yomybaby (approved):
LGTM

Issue #2803, opened 2024-11-01
Session usage component in React


yomybaby:
![https://github.com/user-attachments/assets/36b58919-26e6-4069-b63e-ade66023d1e7](https://github.com/user-attachments/assets/36b58919-26e6-4069-b63e-ade66023d1e7)

For the NEO session list and session detail panel(#2784), we need to implement the session usage component in React using the `ComputeSessionNode` fragment.

JIRA Issue: FR-52

PR #2814, opened 2024-11-05
[FR-52] feat: Session usage monitor in react

ironAiken2:
### This PR resolves [#2803](https://github.com/lablup/backend.ai-webui/issues/2803) issue.

**Changes:**
- Displays the resources allocated to the session. Currently, the compute session node does not include a live stat field, which shows the live stat information of the main kernel associated with the session.
- Since each device has different usage, I've shown the usage below the progress bar to bring consistency to the design.

**How to test:**
- Access the Detail page of the endpoint that is currently in service. 
- Verify that resource usage looks normal.

|normal|small screen|
|---|---|
|||

- [ ] Documentation

### DIFF
--- a/react/src/components/SessionUsageMonitor.tsx
+++ b/react/src/components/SessionUsageMonitor.tsx
@@ -1,0 +1,311 @@
+import {
+  convertBinarySizeUnit,
+  convertDecimalSizeUnit,
+  toFixedFloorWithoutTrailingZeros,
+} from '../helper';
+import { useResourceSlotsDetails } from '../hooks/backendai';
+import BAIProgressWithLabel from './BAIProgressWithLabel';
+import Flex from './Flex';
+import { SessionUsageMonitorFragment$key } from './__generated__/SessionUsageMonitorFragment.graphql';
+import {
+  Progress,
+  ProgressProps,
+  Tooltip,
+  Typography,
+  theme,
+  Row,
+  Col,
+} from 'antd';
+import graphql from 'babel-plugin-relay/macro';
+import _ from 'lodash';
+import { useMemo } from 'react';
+import { useFragment } from 'react-relay';
+
+interface SessionUsageMonitorProps extends ProgressProps {
+  sessionFrgmt: SessionUsageMonitorFragment$key | null;
+  size?: 'small' | 'default';
+}
+
+const SessionUsageMonitor: React.FC<SessionUsageMonitorProps> = ({
+  sessionFrgmt,
+  size = 'default',
+}) => {
+  const { token } = theme.useToken();
+  const { mergedResourceSlots } = useResourceSlotsDetails();
+
+  const sessionNode = useFragment(
+    graphql`
+      fragment SessionUsageMonitorFragment on ComputeSessionNode {
+        kernel_nodes {
+          edges {
+            node {
+              live_stat
+              occupied_slots
+            }
+          }
+        }
+      }
+    `,
+    sessionFrgmt,
+  );
+
+  const firstKernelNode = _.get(sessionNode, 'kernel_nodes.edges[0].node');
+  const occupiedSlots = useMemo(
+    () => JSON.parse(firstKernelNode?.occupied_slots ?? '{}'),
+    [firstKernelNode?.occupied_slots],
+  );
+  const resourceSlotNames = _.keysIn(occupiedSlots);
+  const liveStat = JSON.parse(
+    _.get(sessionNode, 'kernel_nodes.edges[0].node.live_stat') ?? '{}',
+  );
+
+  // to display util first, mem second
+  const sortedLiveStat = useMemo(
+    () =>
+      Object.keys(liveStat)
+        .sort((a, b) => {
+          const aUtil = a.includes('_util');
+          const bUtil = b.includes('_util');
+          const aMem = a.includes('_mem');
+          const bMem = b.includes('_mem');
+
+          if (aUtil && !bUtil) return -1;
+          if (!aUtil && bUtil) return 1;
+          if (aMem && !bMem) return -1;
+          if (!aMem && bMem) return 1;
+
+          return 0;
+        })
+        .reduce((acc: { [key: string]: any }, key) => {
+          acc[key] = liveStat[key];
+          return acc;
+        }, {}),
+    [liveStat],
+  );
+
+  const displayMemoryUsage = (
+    current: string,
+    capacity: string,
+    decimalSize: number = 2,
+  ) => {
+    return `${convertBinarySizeUnit(current, 'g', decimalSize)?.numberFixed ?? '-'} / ${
+      convertBinarySizeUnit(capacity, 'g', decimalSize)?.numberFixed ?? '-'
+    } GiB`;
+  };
+
+  return (
+    <Row gutter={[16, 16]}>
+      {sortedLiveStat?.cpu_util ? (
+        <Col xs={24} sm={12}>
+          <Flex direction="column" align="stretch">
+            {size === 'default' ? (
+              <>
+                <Typography.Text>
+                  {mergedResourceSlots?.['cpu']?.human_readable_name}
+                </Typography.Text>
+                <BAIProgressWithLabel
+                  percent={sortedLiveStat?.cpu_util?.pct || 0}
+                  valueLabel={
+                    toFixedFloorWithoutTrailingZeros(
+                      liveStat?.cpu_util?.pct || 0,
+                      1,
+                    ) + '%'
+                  }
+                  strokeColor="#BFBFBF"
+                  progressStyle={{ border: 'none' }}
+                />
+              </>
+            ) : (
+              <Tooltip
+                title={mergedResourceSlots?.['cpu']?.human_readable_name}
+              >
+                <Progress
+                  format={(percent) => (
+                    <Typography.Text style={{ fontSize: token.fontSizeSM }}>
+                      {percent + '%'}
+                    </Typography.Text>
+                  )}
+                  percent={
+                    _.toNumber(
+                      toFixedFloorWithoutTrailingZeros(
+                        sortedLiveStat?.cpu_util?.pct,
+                        1,
+                      ),
+                    ) || 0
+                  }
+                  strokeColor="#BFBFBF"
+                  strokeLinecap="butt"
+                />
+              </Tooltip>
+            )}
+          </Flex>
+        </Col>
+      ) : null}
+
+      {sortedLiveStat?.mem ? (
+        <Col xs={24} sm={12}>
+          <Flex direction="column" align="stretch">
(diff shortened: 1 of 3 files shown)


ironAiken2:
* **#2814** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2814?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,254 @@
...
+          return 0;
+        })
+        .reduce((acc: { [key: string]: any }, key) => {
+          acc[key] = liveStat[key];
+          return acc;
+        }, {}),
+    [liveStat],
+  );
+
+  console.log('#liveStat: ', sortedLiveStat);
```
agatha197:
```suggestion
```

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,254 @@
...
+                      1,
+                    ),
+                  )}
+                  strokeLinecap="butt"
+                  size={{ height: 20 }}
+                  strokeColor={token.colorPrimary}
+                ></Progress>
+              </Flex>
+              <Typography.Text>
+                {`${bytesToGB(sortedLiveStat?.mem?.current)} / ${bytesToGB(sortedLiveStat?.mem?.capacity)} GB`}
```
agatha197: Let's use `bytesToBinarySize` for memory. And we have to use `GiB`.

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,254 @@
...
+                  showInfo
+                  status="normal"
+                  percent={_.toNumber(
+                    toFixedFloorWithoutTrailingZeros(
+                      sortedLiveStat?.mem?.pct || 0,
+                      1,
+                    ),
+                  )}
+                  strokeLinecap="butt"
+                  size={{ height: 20 }}
```
agatha197:
```suggestion
                  size={{ height: token.sizeMD }}
```

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,254 @@
...
+              </Flex>
+              <Typography.Text>
+                {`${bytesToGB(sortedLiveStat?.mem?.current)} / ${bytesToGB(sortedLiveStat?.mem?.capacity)} GB`}
+              </Typography.Text>
+            </Flex>
+          </Col>
+        ) : null}
+        {/* accelerator type is in the format of `[device_name]_[util|mem]` */}
+        {_.map(sortedLiveStat, (value, key) => {
+          const device = key.split('_')[0] as keyof typeof resourceTypeMap;
```
agatha197:
```suggestion
          const device = key.split('_')?.[0] as keyof typeof resourceTypeMap;
```

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,254 @@
...
+                          ),
+                        )}
+                        strokeLinecap="butt"
+                        size={{ height: 20 }}
+                        strokeColor={token.colorPrimary}
+                      ></Progress>
+                    </Flex>
+                    <Typography.Text>
+                      {_.includes(key, 'mem') &&
+                        `${device} (mem) ${bytesToGB(sortedLiveStat?.mem?.current)} / ${bytesToGB(sortedLiveStat?.mem?.capacity)} GB`}
```
agatha197: Same as above.

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+interface SessionUsageMonitorProps extends ProgressProps {
+  sessionFrgmt: SessionUsageMonitorFragment$key | null;
+  col?: 1 | 2 | 3 | 4;
+  size?: ProgressProps['size'];
+}
+
+// compute session node query returns accelerator type as `[device_name]_[util|mem]`
+// so we need to split via '_' and get the first part to map to the resource type in `resourceTypeIconSrcMap`
+type acceleratorType =
```
agatha197: this is the same code with `knownAcceleratorResourceSlotNames` of hooks/backendai.tsx

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,254 @@
...
+                        showInfo
+                        status="normal"
+                        percent={_.toNumber(
+                          toFixedFloorWithoutTrailingZeros(
+                            sortedLiveStat?.[device]?.pct || 0,
+                            1,
+                          ),
+                        )}
+                        strokeLinecap="butt"
+                        size={{ height: 20 }}
```
agatha197:
```suggestion
                        size={{ height: token.sizeMD }}
```

agatha197 (changes requested):
Memory doesn't have label, but CUDA  memory has label.

react/src/components/SessionUsageMonitor.tsx
ironAiken2: The `bytesToBinarySize` cannot select a unit (division by 1024 automatically selects the unit). The session list also uses the `bytesToGB` calculation, but I'll change it to `iSizeToSize` for a more accurate calculation.

react/src/components/SessionUsageMonitor.tsx
ironAiken2: yes but the device names coming into live stat are not in the form of `cuda.device` (they come in the form of cuda_mem), so we need logic to extract and map only the device names based on the '_'.

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,264 @@
...
+                  </Flex>
+                </Col>
+              )}
+            </>
+          ) : null;
+        })}
+      </Row>
+      <Flex justify="end" style={{ width: '100%' }}>
+        <Typography.Text>
+          {`I/O Read: ${iSizeToSize(sortedLiveStat?.io_read?.current, 'g')?.numberFixed} GiB / Write: ${iSizeToSize(sortedLiveStat?.io_write?.current, 'g')?.numberFixed} GiB`}
```
agatha197: For I/O read and write, please use `GB`.

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,264 @@
...
+        }
+      }
+    `,
+    sessionFrgmt,
+  );
+
+  const liveStat = JSON.parse(
+    _.get(kernel_nodes, 'kernel_nodes.edges[0].node.live_stat') ?? '{}',
+  );
+  console.log(liveStat);
```
agatha197:
```suggestion
```

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,264 @@
...
+                {`(mem) ${
+                  iSizeToSize(sortedLiveStat?.mem?.current, 'g')?.numberFixed ??
+                  0
+                }
+                /
+                ${
+                  iSizeToSize(sortedLiveStat?.mem?.capacity, 'g')
+                    ?.numberFixed ?? 0
+                }
+                GiB`}
```
agatha197: How about use `numberUnit` instead of `numberFixed`. Use can refer https://github.com/lablup/backend.ai-webui/blob/0479273e50fe3726916f803650d30eae83778b59/react/src/components/AgentDetailModal.tsx#L104

react/src/components/SessionUsageMonitor.tsx
agatha197: Ah I see.

agatha197 (approved):
LGTM

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  </Flex>
+                </Col>
+              )}
+            </>
+          ) : null;
+        })}
+      </Row>
+      <Flex justify="end" style={{ width: '100%' }}>
+        <Typography.Text>
+          {`I/O Read: ${iSizeToSize(sortedLiveStat?.io_read?.current, 'g')?.numberUnit}B / Write: ${iSizeToSize(sortedLiveStat?.io_write?.current, 'g')?.numberUnit}B`}
```
yomybaby: The iSizeToSize function is for binary size conversion, not for I/O. Please check the original version and ensure that the display matches it. See the reference here: https://github.com/lablup/backend.ai-webui/blob/main/src/components/backend-ai-session-list.ts#L4318-L4326

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,260 @@
...
+  rocm: 'rocm.device',
+  tpu: 'tpu.device',
+  ipu: 'ipu.device',
+  atom: 'atom.device',
+  'atom-plus': 'atom-plus.device',
+  gaudi2: 'gaudi2.device',
+  warboy: 'warboy.device',
+  rngd: 'rngd.device',
+  'hyperaccel-lpu': 'hyperaccel-lpu.device',
+};
```
yomybaby: It's better to create a function to add `device` without using hardcoded accelerator names.

react/src/components/SessionUsageMonitor.tsx
```diff
@@ -0,0 +1,260 @@
...
+          </Col>
+        ) : null}
+        {/* accelerator type is in the format of `[device_name]_[util|mem]` */}
+        {_.map(sortedLiveStat, (value, key) => {
+          const device = key.split('_')?.[0] as keyof typeof resourceTypeMap;
+
+          return _.includes(
+            _.keysIn(_.omit(resourceTypeMap, 'cpu', 'mem')),
+            device,
+          ) ? (
```
yomybaby: Let's refactor this part. `filter`(or pickBy) first and map later.

yomybaby (changes requested):

react/src/components/SessionUsageMonitor.tsx
ironAiken2: In that code, the API request value comes in as a MB value, but not in ComputeSessionNode, so I modified it to show it in MB, the same as the session list usage column, using `convertDecimalSizeUnit`.

yomybaby (changes requested):
I found several things that need improvement and discussion. Let's discuss this PR tomorrow.
- Should we include an icon?
- Should the progress bar be larger?
- How can we make it usable in the columns of the session list neo?
- There are incorrect labels and units.

ironAiken2:
The size property of progressProps or showIcon can be used to modify the visibility of the icon or the size of the progress bar.

yomybaby (approved):
LGTM

Issue #2812, opened 2024-11-05
User can set a batch session timeout


yomybaby:
In session launcher, user can set a batch session timeout.
- [Core Issue](https://github.com/lablup/backend.ai/issues/2357)

Let's start implementing the UI even though the API is not ready yet.

PR #2824, opened 2024-11-06
feat: add batch job timeout duration

agatha197:
resolves https://github.com/lablup/backend.ai-webui/issues/2812
related PR: https://github.com/lablup/backend.ai/pull/3066

Add batch job timeout duration to the session launcher.
User can select the time unit and input the timeout duration.

**How to test:**
> test endpoint: 10.82.230.49
1. Modify the support version of `batch-timeout` in `backend.ai-client-esm` for testing. (24.09)
2. Set session type to the batch.
3. Enable `Batch Job Timeout Duration`
  - eg  
    start command: sleep 20 && echo \"hello world\"
    timeout: 3s

**Checklist:**
- [ ] Batch job timeout value remains after refreshing
- [ ] The 'Confirm and Launch' page has the same value if you set the 'Batch job timeout duration'.
- [ ] (API is not implemented yet) Create a new session with batch session timeout option.

**Screentshots:**

**Checklist:**

- [ ] Documentation
- [ ] Test case: Verify timeout can be enabled/disabled
- [ ] Test case: Confirm time picker accepts valid duration formats
- [ ] Test case: Check default value initialization
- [ ] Test case: Validate timeout enforcement on batch sessions


agatha197:
* **#2824** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2824?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

 Join @agatha197 and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -979,6 +999,150 @@ const SessionLauncherPage = () => {
...
+                              {t('session.launcher.BatchJobTimeoutDuration')}
+                            </Typography.Text>
+                            <Tooltip
+                              title={t(
+                                'session.launcher.BatchJobTimeoutDurationDesc',
+                              )}
+                            >
+                              <QuestionCircleOutlined />
+                            </Tooltip>
+                          </Flex>
```
yomybaby: `Form.Item` has `tooltip` prop. You don't need to add it in `label`

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -979,6 +999,150 @@ const SessionLauncherPage = () => {
...
+                                'timeout',
+                              ]);
+                              const unit = form.getFieldValue([
+                                'batch',
+                                'timeoutUnit',
+                              ]);
+                              const humanizedDuration = timeout
+                                ? dayjs
+                                    .duration(timeout, unit ?? 'ms')
+                                    .humanize(true)
```
yomybaby: Using the `humanize` function does not provide an accurate duration, so it would be better to use an alternative way. How about `elapsedTime` of baiClient?

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -979,6 +999,150 @@ const SessionLauncherPage = () => {
...
+                                    rules={[
+                                      {
+                                        min: 0,
+                                        type: 'number',
+                                        message: t(
+                                          'error.AllowsPositiveNumberOnly',
+                                        ),
+                                      },
+                                    ]}
+                                  >
```
yomybaby: Please set `required` to true when `timeoutEnabled` is true.

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -979,6 +999,150 @@ const SessionLauncherPage = () => {
...
+                            </Tooltip>
+                          </Flex>
+                        }
+                        extra={
+                          <Form.Item
+                            noStyle
+                            shouldUpdate={(prev, next) =>
+                              prev.batch.timeout !== next.batch.timeout ||
+                              prev.batch.timeoutUnit !== next.batch.timeoutUnit
+                            }
```
yomybaby: If `extra` is a React element, `Form.Item` creates a bottom space for it, which causes unnecessary space like this:



I'll change and submit, if you are ok.

yomybaby (changes requested):

react/src/pages/SessionLauncherPage.tsx
yomybaby: We can use `dependencies={[['batch', 'timeoutEnabled'], ['batch', 'timeoutUnit']]}` instead of `shouldUpdate`

yomybaby (approved):
I push a commit to apply related to my reviews. LGTM, now!

PR #2869, opened 2024-11-21
fix: Add the missing session `type` argument to the SFTP session creation API

jopemachine:
This PR resolves an issue where the SFTP session is created as an interactive session rather than an upload session due to the missing session `type` argument.


yomybaby (changes requested):
LGTM

yomybaby (approved):

PR #2900, opened 2024-11-29
fix: compare time is expired by using `asSeconds()` instead of `seconds()`

lizable:
The PR resolves [backend.ai#3180](https://github.com/lablup/backend.ai/issues/3180) changes the condition for detecting an expired session from checking if seconds are exactly 0 to checking if the total duration in seconds is less than or equal to 0. This fixes an edge case where the session could appear valid even when expired.

**Checklist:**

- [ ] Documentation

1. Login with any active user account.
2. Leave the browser window open without any user interaction for about a minute.
3. Observe that the browser refreshes automatically after one minute.

### DIFF
--- a/react/src/components/LoginSessionExtendButton.tsx
+++ b/react/src/components/LoginSessionExtendButton.tsx
@@ -41,8 +41,8 @@ const LoginSessionExtendButton: React.FC<
         callback={() => {
           const diff = dayjs(data?.expires).diff(dayjs(), 'seconds');
           const duration = dayjs.duration(Math.max(0, diff), 'seconds');
-          const days = duration.days();
-          if (duration.seconds() === 0) {
+          const days = Math.floor(duration.asDays());
+          if (duration.asSeconds() <= 0) {
             // @ts-ignore
             if (globalThis.isElectron) {
               // @ts-ignore


lizable:
* **#2900** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2900?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2900?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

react/src/components/LoginSessionExtendButton.tsx
```diff
@@ -42,7 +42,7 @@ const LoginSessionExtendButton: React.FC<
           const diff = dayjs(data?.expires).diff(dayjs(), 'seconds');
           const duration = dayjs.duration(Math.max(0, diff), 'seconds');
           const days = duration.days();
```
yomybaby: We need to change this to 'asDays' also.

yomybaby (changes requested):

yomybaby (approved):
LGTM

PR #2912, opened 2024-12-02
fix:  Resource group change does not update related data in form

### LINKED ISSUE #2911: Resource slot selector does not update when changing resource groups

Linux (x86-64)

### Backend.AI version

24.12.0-alpha.2.1

### Describe the bug

In the session launcher and service launcher, when a user changes the resource group, the options in the resource slot selector do not reflect the available resources in the new resource group.

### To Reproduce

1. Open the session launcher or service launcher in Backend.AI Web UI.
1. Change the resource group to one with a different type of accelerator.
1. Observe the resource slot selector options.
   ### Expected Behavior

The resource slot selector should update to reflect the available resources in the newly selected resource group.

### Anything else?

_No response_

JIRA Issue: FR-33

yomybaby:
Resolves #2911

### Description
When a user changes the resource group, the related data is not updated as expected. This issue occurs in the following scenario:

### Steps to Reproduce
1. The user navigates to the session launcher or service launcher.
2. The user changes the resource group
3. The expected behavior is that all related data and configurations should update to reflect the new resource group such as accelerator types.
4. However, the data remains unchanged, leading to inconsistencies.

**Changes:**
- Replaced global resource group state with form-level resource group state in ResourceAllocationFormItems
- Added transition handling for resource group selection changes to prevent UI jitters
- Added validation to ensure selected resource group exists before checking presets
- Improved auto-selection behavior for resource groups
- Added loading and disabled states during resource group transitions
- Use BAISelect with `autoSelectOption` for accelerator type

**Rationale:**
The changes improve form stability by using form-level state instead of global state for resource group selection. This prevents unwanted side effects when multiple forms are present. The addition of transitions makes the UI more stable during resource group changes.

**Effects:**
- More reliable resource group selection behavior in forms
- Smoother UI transitions during resource group changes
- Better handling of invalid resource group selections
- Improved auto-selection behavior for default resource groups

**Review Focus:**
- Verify resource group selection behavior in forms
- Check transition states during resource group changes
- Validate auto-selection of default resource groups
- Confirm proper handling of invalid resource group selections

### DIFF
--- a/react/src/components/ResourceAllocationFormItems.tsx
+++ b/react/src/components/ResourceAllocationFormItems.tsx
@@ -6,10 +6,7 @@ import {
 import { useSuspendedBackendaiClient, useUpdatableState } from '../hooks';
 import { useResourceSlotsDetails } from '../hooks/backendai';
 import { useCurrentKeyPairResourcePolicyLazyLoadQuery } from '../hooks/hooksUsingRelay';
-import {
-  useCurrentProjectValue,
-  useCurrentResourceGroupValue,
-} from '../hooks/useCurrentProject';
+import { useCurrentProjectValue } from '../hooks/useCurrentProject';
 import { useEventNotStable } from '../hooks/useEventNotStable';
 import {
   MergedResourceLimits,
@@ -17,6 +14,7 @@ import {
   useResourceLimitAndRemaining,
 } from '../hooks/useResourceLimitAndRemaining';
 import AgentSelect from './AgentSelect';
+import BAISelect from './BAISelect';
 import DynamicUnitInputNumberWithSlider from './DynamicUnitInputNumberWithSlider';
 import Flex from './Flex';
 import {
@@ -35,7 +33,6 @@ import {
   Form,
   Radio,
   Row,
-  Select,
   Switch,
   theme,
 } from 'antd';
@@ -113,7 +110,11 @@ const ResourceAllocationFormItems: React.FC<
   const [isPendingAgentList, startAgentListTransition] = useTransition();
 
   const currentProject = useCurrentProjectValue();
-  const currentResourceGroup = useCurrentResourceGroupValue(); // use global state
+  const currentResourceGroupInForm =
+    Form.useWatch(['resourceGroup'], {
+      form,
+      preserve: true,
+    }) || form.getFieldValue('resourceGroup');
 
   const currentImage = Form.useWatch(['environments', 'image'], {
     form,
@@ -126,12 +127,12 @@ const ResourceAllocationFormItems: React.FC<
   const [{ currentImageMinM, remaining, resourceLimits, checkPresetInfo }] =
     useResourceLimitAndRemaining({
       currentProjectName: currentProject.name,
-      currentResourceGroup: currentResourceGroup || undefined, // global currentResourceGroup can be null
+      currentResourceGroup: currentResourceGroupInForm || undefined, // global currentResourceGroup can be null
       currentImage: currentImage,
     });
 
   const { mergedResourceSlots, resourceSlotsInRG: resourceSlots } =
-    useResourceSlotsDetails(currentResourceGroup || undefined);
+    useResourceSlotsDetails(currentResourceGroupInForm || undefined);
 
   const acceleratorSlots = _.omitBy(resourceSlots, (value, key) => {
     if (['cpu', 'mem', 'shmem'].includes(key)) return true;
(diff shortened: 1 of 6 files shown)


yomybaby:
* **#2912** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2912?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2912?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

react/src/components/ResourceAllocationFormItems.tsx
```diff
@@ -113,7 +110,11 @@ const ResourceAllocationFormItems: React.FC<
   const [isPendingAgentList, startAgentListTransition] = useTransition();
 
   const currentProject = useCurrentProjectValue();
-  const currentResourceGroup = useCurrentResourceGroupValue(); // use global state
+  const currentResourceGroupInFrom =
```
ironAiken2: There's a typo.

```suggestion
  const currentResourceGroupInForm =
```

ironAiken2: I'm not sure if this is the intended logic, but backend.ai currently allows an empty string for the resource group name.

react/src/components/ResourceAllocationFormItems.tsx
```diff
@@ -113,7 +110,11 @@ const ResourceAllocationFormItems: React.FC<
...
+    Form.useWatch(['resourceGroup'], {
+      form,
+      preserve: true,
+    }) || form.getFieldValue('resourceGroup');
```
ironAiken2, line 117: I'm not sure if this is the intended logic, but backend.ai currently allows an empty string for the resource group name.

I've created an issue and will work on adding validation logic to the webui side. [Issue](https://github.com/lablup/backend.ai-webui/issues/2915)

ironAiken2 (changes requested):
It works fine but I left few comments :)

ironAiken2:
> Currently, it seems that when I change the project in the webui, the resource group does not display to match the project.

The above discussion has been resolved in Teams below.
[teams](https://teams.microsoft.com/l/message/19:14c484402d874dafb15806d093b95a82@thread.skype/1733203156283?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=74ae2c4d-ec4d-4fdf-b2c2-f5041d1e8631&parentMessageId=1733203156283&teamName=devops&channelName=Frontend&createdTime=1733203156283)

react/src/components/ResourceAllocationFormItems.tsx
yomybaby: An empty string is different from a string containing a single whitespace (" "). 😉

ironAiken2: When I first realized the issue, I was only reporting on empty strings. 
Later, I realized that there were more to consider, such as characters starting with a space or containing a single whitespace, so I created an issue and worked on adding regular expressions. Thank you for notice! :)

react/src/components/ResourceAllocationFormItems.tsx
yomybaby: Thank you for reporting. I fixed it.

ironAiken2 (approved):
LGTM

lizable (commented):
I'm not sure whether is intended or not, but when it comes to allowing none of resource groups in certain project(group), It still lists resource group list which is received from the server. the control panel describes that it's "allow-list basis", but I have no idea whether is bug or not.
@yomybaby @ironAiken2 We need to know this is intended before merging.

### Screenshot(s)
* Control Panel


* WebUI

lizable:
My mistakes, server-side doesn't allow none of resource group is coupled with project(group). Therefore we don't need to consider that cases.

lizable (commented):
Btw, After changing the project in the header while the main layout is located in session launcher page, resource allocation points out resource group is empty. It's okay, but wouldn't be better to select "available" resource group (such as auto-selecting the first one from the resource group list) by default? 
Here's the video clip that shows resource group is set to none when changing project(group) in session launcher page.

<div class='graphite__hidden'>
          <div>🎥 Video uploaded on Graphite:</div>
            <a href="https://app.graphite.dev/media/video/Dgoz5XqHffJdivyccymr/ef650a63-e1ce-4fa6-bce1-8da1af915d98.mov">
              
            </a>
          </div>
<video src="https://graphite-user-uploaded-assets-prod.s3.amazonaws.com/Dgoz5XqHffJdivyccymr/ef650a63-e1ce-4fa6-bce1-8da1af915d98.mov">Screen Recording 2024-12-04 at 6.30.07 PM.mov</video>

yomybaby:
@lizable  This is the intended behavior which I changed. When the project is changed, if the previously selected resource group does not exist in the new project, the user is required to explicitly select a resource group. While automation could be convenient, the resource group is one of the critical selections, so I excluded it from automation. We will observe further and make a decision regarding automatic selection.

lizable:
Understood. Let's apply this for now and think about the better option as a separate issue.

lizable (approved):
LGTM.

Issue #2915, opened 2024-12-03
Resource group with only whitespace name can be created


ironAiken2:
### What Operating System(s) are you seeing this problem on?

Linux (x86-64)

### Backend.AI version

24.12.0a1

### Description
There is an issue where a resource group with a name consisting only of whitespace can be created. However, when attempting to retrieve this group using the API, an error occurs and the group is not listed.

### Expected Behavior

The system should prevent the creation of resource groups with names consisting only of whitespace.

PR #2916, opened 2024-12-03
fix: enhance resource group name validation logic for spaces and special characters

ironAiken2:
### This PR resolves [#2915](https://github.com/lablup/backend.ai-webui/issues/2915) issue

Related PR: [control-panel#930](https://github.com/lablup/backend.ai-control-panel/pull/930)

**Changes:**

Currently, validation rules are not applied in Backend.AI Core, but plan to add them to the scope of domain and project in the future. Therefore, I added validation rules (allow dot and unicode) to match the rules on the core side.

**Invalid strings:**
- contains special characters 
- consisting of leading, trailing, or only spaces, or more than one space between characters.
- empty string

**How to test:**
- Move to resources page and click resource group tab
- Create a new resource group and verify that the validation rule works correctly.

### DIFF
--- a/src/components/backend-ai-resource-group-list.ts
+++ b/src/components/backend-ai-resource-group-list.ts
@@ -365,6 +365,14 @@ export default class BackendAIResourceGroupList extends BackendAIPage {
             valid: nativeValidity.valid,
             valueMissing: !nativeValidity.valid,
           };
+        } else if (nativeValidity.patternMismatch) {
+          this.resourceGroupNameInput.validationMessage = _text(
+            'resourceGroup.EnterValidResourceGroupName',
+          );
+          return {
+            valid: nativeValidity.valid,
+            patternMismatch: !nativeValidity.valid,
+          };
         } else {
           this.resourceGroupNameInput.validationMessage = _text(
             'resourceGroup.EnterValidResourceGroupName',
@@ -459,7 +467,8 @@ export default class BackendAIResourceGroupList extends BackendAIPage {
           this.notification.show(true, err);
         });
     } else {
-      // ResourceGroupNameEl.reportValidity();
+      this._validateResourceGroupName();
+      this.resourceGroupNameInput.reportValidity();
       return;
     }
   }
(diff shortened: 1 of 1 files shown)


ironAiken2:
* **#2916** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2916?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2916?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`

<h2></h2>

lizable (dismissed):
Let's add a few test cases for this changes.
Please take a look at the test cases code below:
https://github.com/lablup/backend.ai/pull/1663

src/components/backend-ai-resource-group-list.ts
```diff
@@ -848,6 +856,7 @@ export default class BackendAIResourceGroupList extends BackendAIPage {
                   validationMessage="${_t('data.explorer.ValueRequired')}"
                   required
                   autoValidate
+                  pattern="^[ㄱ-ㅎ가-힣a-zA-Z0-9]+(\\s[ㄱ-ㅎ가-힣a-zA-Z0-9]+)*$"
```
lizable: And we need to consider unicode, not just for English and Korean. In Server-side, Resource Group name allows unicode, which contains special characters.

### Screenshot(s):

* WebUI:



* CLI:

ironAiken2:
> Let's add a few test cases for this changes.
> Please take a look at the test cases code below:
> https://github.com/lablup/backend.ai/pull/1663

Since it's webcomponent code, I don't think it would be possible to write separate test code. We can use playwrite to test it, but I think it defeats the purpose since it's a unit functional test.

src/components/backend-ai-resource-group-list.ts
ironAiken2: Fixed to allow Unicode.

agatha197 (changes requested):
When I entered an invalid resource group name, I got a required message. Not a pattern-related message. Please refer backend-ai-page codes.

ironAiken2:
> 
> When I entered an invalid resource group name, I got a required message. Not a pattern-related message. Please refer backend-ai-page codes.

Changed the @change event to use the @input event instead so that we can validate on input of a value regardless of focus.

agatha197 (approved):
LGTM

Issue #2957, opened 2024-12-16
Update`desired_session_count` field name to `replicas` of a endpoint


yomybaby:
In the recent pull request [#3257](https://github.com/lablup/backend.ai/pull/3257), the field name `desired_session_count` has been changed to `replicas`. This issue is to update the Backend.AI WebUI to reflect this change.

### Affected Areas

- Any WebUI components or pages that display or interact with the `desired_session_count` field should be updated to use the new `replicas` field name.

JIRA Issue: FR-17

PR #2963, opened 2024-12-17
feat(FR-17): update `desired_session_count` field name to `replicas`

ironAiken2:
### This PR resolves [#2957](https://github.com/lablup/backend.ai-webui/issues/2957) issue

Since Backend.AI Core version 24.12.0, the desired_session_count field has been replaced with `replicas`. 

**Changes:**
- Added `replicas` field to service creation and update interfaces
- Deprecated `desired_session_count` field for manager version >= 24.12.0
- Updated GraphQL queries to handle both `desired_session_count` and `replicas` fields
- Added version-based field selection using `@deprecatedSince` and `@since` directives
- Added `replicas` feature flag check in client capabilities

**How to test:**
- Check whether `desired_session_count` or `replicas` is used when creating or modifying a service, depending on the manager version. (From 24.12.0 onwards, replicas is used.) 
- Check whether desired_session_count is displayed correctly on the Service page and Detail page.

### DIFF
--- a/react/src/components/ServiceLauncherPageContent.tsx
+++ b/react/src/components/ServiceLauncherPageContent.tsx
@@ -88,7 +88,8 @@ interface ServiceCreateConfigType {
 }
 export interface ServiceCreateType {
   name: string;
-  desired_session_count: number;
+  desired_session_count?: number;
+  replicas?: number;
   image: string;
   runtime_variant: string;
   architecture: string;
@@ -106,7 +107,7 @@ export interface ServiceCreateType {
 interface ServiceLauncherInput extends ImageEnvironmentFormInput {
   serviceName: string;
   vFolderID: string;
-  desiredRoutingCount: number;
+  replicas: number;
   openToPublic: boolean;
   modelMountDestination: string;
   modelDefinitionPath: string;
@@ -153,7 +154,8 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
     graphql`
       fragment ServiceLauncherPageContentFragment on Endpoint {
         endpoint_id
-        desired_session_count
+        desired_session_count @deprecatedSince(version: "24.12.0")
+        replicas @since(version: "24.12.0")
         resource_group
         resource_slots
         resource_opts
@@ -264,7 +266,7 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
   const legacyMutationToUpdateService = useTanMutation({
     mutationFn: (values: ServiceLauncherFormValue) => {
       const body = {
-        to: values.desiredRoutingCount,
+        to: values.replicas,
       };
       return baiSignedRequestWithPromise({
         method: 'POST',
@@ -289,7 +291,8 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
       }
       const body: ServiceCreateType = {
         name: values.serviceName,
-        desired_session_count: values.desiredRoutingCount,
+        // REST API does not support `replicas` field. To use `replicas` field, we need `create_endpoint` mutation.
+        desired_session_count: values.replicas,
         ...getImageInfoFromInputInCreating(
           checkManualImageAllowed(
             baiClient._config.allow_manual_image_name_for_session,
@@ -407,7 +410,8 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
         msg
         endpoint {
           endpoint_id
-          desired_session_count
+          desired_session_count @deprecatedSince(version: "24.12.0")
+          replicas @since(version: "24.12.0")
           resource_group
           resource_slots
           resource_opts
@@ -500,7 +504,11 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
                       ? 'SINGLE_NODE'
                       : 'MULTI_NODE',
                   cluster_size: values.cluster_size,
-                  desired_session_count: values.desiredRoutingCount,
+                  ...(baiClient.supports('replicas')
+                    ? { replicas: values.replicas }
+                    : {
+                        desired_session_count: values.replicas,
+                      }),
                   ...getImageInfoFromInputInEditing(
                     checkManualImageAllowed(
                       baiClient._config.allow_manual_image_name_for_session,
@@ -643,7 +651,7 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
         serviceName: endpoint?.name,
         resourceGroup: endpoint?.resource_group,
         allocationPreset: 'custom',
-        desiredRoutingCount: endpoint?.desired_session_count ?? 1,
+        replicas: endpoint?.replicas ?? endpoint?.desired_session_count ?? 1,
         // FIXME: memory doesn't applied to resource allocation
         resource: {
           cpu: parseInt(JSON.parse(endpoint?.resource_slots || '{}')?.cpu),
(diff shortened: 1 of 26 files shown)


react/src/pages/EndpointDetailPage.tsx
```diff
@@ -298,7 +296,7 @@ const EndpointDetailPage: React.FC<EndpointDetailPageProps> = () => {
     },
     {
       label: t('modelService.DesiredSessionCount'),
-      children: endpoint?.desired_session_count,
+      children: endpoint?.desired_session_count ?? endpoint?.replicas,
```
agatha197: Is there any reason to consider desired_session_count first? In `ServiceLauncherPageContent`, it uses replicas value first like `replicas: endpoint?.replicas ?? endpoint?.desired_session_count ?? 1,`

same as below.

agatha197 (changes requested):

react/src/pages/EndpointDetailPage.tsx
ironAiken2: When `desired_session_count` deprecated since 24.12.0 and `replicas` supported since 24.12.0, it is unlikely that both values will come in at the same time.

But I think replicas should comes first because It's more recent fields and `desired_session_count` is for backward compatibility.

react/src/pages/EndpointListPage.tsx
```diff
@@ -153,9 +150,12 @@ const EndpointListPage: React.FC<PropsWithChildren> = ({ children }) => {
             type="text"
             icon={<SettingOutlined />}
             style={
-              isDestroyingStatus(row?.desired_session_count, row?.status) ||
-              (!!row.created_user_email &&
-                row.created_user_email !== currentUser.email)
+              isDestroyingStatus(
+                row?.desired_session_count ?? row?.replicas,
```
agatha197:
```suggestion
                row?.replicas ?? row?.desired_session_count,
```
I think so it's better to consider replicas first. Please apply it below similar codes too.

agatha197 (changes requested):

react/src/pages/EndpointDetailPage.tsx
```diff
@@ -107,14 +111,8 @@ const EndpointDetailPage: React.FC<EndpointDetailPageProps> = () => {
...
-  const [paginationState] = useState<{
-    current: number;
-    pageSize: number;
-  }>({
-    current: 1,
-    pageSize: 100,
-  });
-  const { message } = App.useApp();
+  const baiClient = useSuspendedBackendaiClient();
+  const navigate = useNavigate();
```
yomybaby: Is there any specific reason to use `useNavigate()`? If not, please use `webuiNavigate` instead.

react/src/pages/EndpointDetailPage.tsx
```diff
@@ -598,7 +596,7 @@ const EndpointDetailPage: React.FC<EndpointDetailPageProps> = () => {
               title: 'Created at',
               dataIndex: 'created_at',
               render: (text, row) => (
-                <span>{dayjs(row.created_at).format('ll LT')}</span>
+                <span>{dayjs(row?.created_at).format('ll LT')}</span>
```
yomybaby: Row can not be `undefined`. Is there any reason to add `?.`?

react/src/pages/EndpointListPage.tsx
```diff
@@ -205,10 +211,10 @@ const EndpointListPage: React.FC<PropsWithChildren> = ({ children }) => {
                   type: 'primary',
                 },
                 onOk: () => {
-                  setOptimisticDeletingId(row.endpoint_id);
+                  setOptimisticDeletingId(row?.endpoint_id);
```
yomybaby, line 214: Same here, It can not be `undefined`

react/src/pages/EndpointListPage.tsx
```diff
@@ -528,30 +538,36 @@ const EndpointListPage: React.FC<PropsWithChildren> = ({ children }) => {
...
       <Table
         loading={{
-          spinning: isPendingPaginationAndFilter || isFilterPending,
+          spinning: isFilterPending || isPendingPageChange,
           indicator: <LoadingOutlined />,
         }}
         scroll={{ x: 'max-content' }}
         rowKey={'endpoint_id'}
-        dataSource={(sortedEndpointList || []) as Endpoint[]}
+        dataSource={(modelServiceList?.items || []) as Endpoint[]}
```
yomybaby: Please use `filterNonNullItems` for `modelServiceList?.items`. and remove `as Endpoint[]`

yomybaby (changes requested):

react/src/pages/EndpointDetailPage.tsx
ironAiken2: I've changed the order of the declarations in this section, and haven't added or deleted anything new. I was using navigate in the existing code, but I'll change that as well.

react/src/pages/EndpointDetailPage.tsx
ironAiken2: I thought `created_at` should be null or undefined. I'll check query data type and apply optional chaining if have null or undefined value.

react/src/pages/EndpointListPage.tsx
ironAiken2: https://app.graphite.dev/github/pr/lablup/backend.ai-webui/2963/FR-17-feat-update-desired_session_count-field-name-to-replicas#comment-PRRC_kwDOCRTcws5yANWE

react/src/pages/EndpointDetailPage.tsx
ironAiken2: Arrays for undefined or null are deleted when using filterNonNullItems, so rows can't be undefined. I'll fix this.

yomybaby (approved):
LGTM!

react/src/components/ServiceLauncherPageContent.tsx
```diff
@@ -899,8 +907,8 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
                     </>
                   )}
                   <Form.Item
-                    label={t('modelService.DesiredRoutingCount')}
-                    name="desiredRoutingCount"
+                    label={t('modelService.NumberOfReplicas')}
```
agatha197, line 910: Let's check `baiClient.supports('replicas')` for this too. Users can confused.

react/src/pages/EndpointListPage.tsx
```diff
@@ -274,7 +281,9 @@ const EndpointListPage: React.FC<PropsWithChildren> = ({ children }) => {
     },
     {
       title: t('modelService.DesiredSessionCount'),
```
agatha197: Same as above.

agatha197 (changes requested):

react/src/components/ServiceLauncherPageContent.tsx
ironAiken2: This field is also required for older versions that use desired_session_count.

agatha197: Then does it need to represent it to original text (desired session count)?

agatha197 (changes requested):

react/src/pages/EndpointListPage.tsx
```diff
@@ -272,14 +279,20 @@ const EndpointListPage: React.FC<PropsWithChildren> = ({ children }) => {
...
     },
-    {
-      title: t('modelService.DesiredSessionCount'),
-      dataIndex: 'desired_session_count',
-      key: 'desiredSessionCount',
-      render: (desired_session_count) => {
-        return desired_session_count < 0 ? '-' : desired_session_count;
-      },
-    },
+    ...(baiClient.supports('replicas')
```
agatha197: Please use `filterEmptyItem`

agatha197 (changes requested):

react/src/pages/EndpointListPage.tsx
ironAiken2: This is used the same as desired_session_count, so I don't think it needs to be version specific.

agatha197: I just suggested this for better readability, but it's a minor issue so it doesn't need to reflect it.

agatha197 (approved):

agatha197 (approved):
LGTM 🚀

Issue #2976, opened 2024-12-20
Resetting the Neo Session Launcher in the review stage disables accelerator resource control when revisited


achimnol:
If I click this:

The accelerator input control is disabled when I restart the wizard, while the configured value is applied if I proceed to create the session:

![https://github.com/user-attachments/assets/e340ba81-fe39-4aaa-83f3-cbff13ecc4bd](https://github.com/user-attachments/assets/e340ba81-fe39-4aaa-83f3-cbff13ecc4bd)

Please update the milestone label as needed if this requires backport.

JIRA Issue: FR-7

PR #3002, opened 2025-01-06
[FR-7] fix: neo session launcher reset function behavior

ironAiken2:
### This PR resolves [#2976](https://github.com/lablup/backend.ai-webui/issues/2976) issue

**Changes:**
- Adds default empty string parameter to `preserveDotStartCase` function to prevent undefined errors
- Initialize the query string via setQuery on reset. currentTabKey will change, so it will be moved to the first step.
- use form.resetFields() method to initialize the form value to the Initial value. If the form value changes, `syncFormToURLWithDebounce` is executed by the provider and the initialization value is saved as a query string.

**Checklist:**
- [ ] Documentation

### DIFF
--- a/react/src/pages/SessionLauncherPage.tsx
+++ b/react/src/pages/SessionLauncherPage.tsx
@@ -84,7 +84,7 @@ import {
 import dayjs from 'dayjs';
 import { useAtomValue } from 'jotai';
 import _ from 'lodash';
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useLayoutEffect, useMemo, useState } from 'react';
 import { ErrorBoundary } from 'react-error-boundary';
 import { Trans, useTranslation } from 'react-i18next';
 import { useLocation } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


react/src/pages/SessionLauncherPage.tsx
```diff
@@ -690,6 +690,15 @@ const SessionLauncherPage = () => {
...
 
+  const [isQueryReset, setIsQueryReset] = useState(false);
+  useEffect(() => {
+    if (isQueryReset) {
+      form.resetFields();
+      setIsQueryReset(false);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isQueryReset]);
+
```
yomybaby:
```suggestion
  const [isQueryReset, setIsQueryReset] = useState(false);
  useLayoutEffect(() => {
    if (isQueryReset) {
      form.resetFields();
      setIsQueryReset(false);
    }
  }, [isQueryReset, form]);
```

yomybaby (changes requested):
When you reset, it moves to the first step, but there is an issue where the previously set session name is briefly visible. This can be resolved by using useLayoutEffect.

yomybaby (approved):
LGTM! 🚢

Issue #3006, opened 2025-01-08
NEO maintenance page using SettingList


yomybaby:
The current `/maintenance` page is implemented using Lit(Web component). We need to refactor this page to use the `SettingList` React component. Additionally, ensure that the `useSetBAINotification` hook is used directly in React to properly display notifications for promise or `bg_task` state changes.

### References

- Refer to `react/src/pages/UserSettingsPage.tsx` and `react/src/components/ContainerRegistryList.tsx` for guidance on implementation.

### Additional Information

Please ensure that the refactored page maintains the existing functionality and improves the user experience with clear notifications.

### Plan

- The `/maintenance` page currently consists of two cards, but more options are expected to be added in the future. Considering the potential integration with the admin settings page, I plan to refactor the page using the `SettingList` component as follows:
  ![image-20250109-010542.png](image-20250109-010542.png){width=1468 height=790}
  - It seems scalable because more items can be easily added to the list, and tabs can be used to organize content if combined with other pages.
- I’m thinking of dividing the tasks into sub-issues: ~~migrating the page written in~~ `lit-element` ~~to~~ `React` and moving the cards currently in the `maintenance` page into the `SettingList` items.
  add) I just use `MaintenancePage` component. So, I don’t need to make another page component for maintenance
  - I plan to create a single issue that includes adding items to the `SettingList` and verifying that the notification feature works correctly using `useSetBAINotification`.
    - To maintain the design and text of the existing cards, I plan to work with the following design.
      !스크린샷 2025-01-09 오전 10.30.27.png|width=392,height=104,alt="스크린샷 2025-01-09 오전 10.30.27.png"!

JIRA Issue: FR-343

PR #3009, opened 2025-01-09
refactor(FR-343): migrate maintenance page to react by using SettingList

nowgnuesLee:
### This PR resolves [#3006](https://github.com/lablup/backend.ai-webui/issues/3006) [(FR-343)](https://lablup.atlassian.net/browse/FR-343)

**Changes**
* Updated routing from LitElement to React components.
* From now on, the `/maintenance` page will operate exclusively with React
* Updated `SettingList` component
	* `isShowChangedOptionFilter`, `isShowResetButton`, and `isShowSearchBar` have been added as props to SettingList. Now, these features can be disabled when they are unnecessary.
* Updated i18n.

- [ ] Documentation

[FR-343]: https://lablup.atlassian.net/browse/FR-343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yomybaby (changes requested):
If `backend-ai-maintenance-view` is not used anymore, please delete it.

react/src/components/SettingList.tsx
```diff
@@ -24,11 +24,17 @@ const useStyles = createStyles(({ css }) => ({
 interface SettingPageProps {
   settingGroup: { title: string; settingItems: SettingItemProps[] }[];
   tabDirection?: 'top' | 'left';
+  isShowChangedOptionFilter?: boolean;
+  isShowResetButton?: boolean;
+  isShowSearchBar?: boolean;
```
yomybaby: For the prop name, it's better to use `show***` instead of `isShow***`.
<SomeComponent showResetButton />

Also, if the default is `show`, it's better to use `hide***` instead of `show***`.

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,220 @@
...
+        statusDescriptions: {
+          pending: t('maintenance.Recalculating'),
+          resolved: t('maintenance.RecalculationFinished'),
+          rejected: t('maintenance.RecalculationFailed'),
+        },
+      },
+      duration: 0,
+    });
+    try {
+      await baiClient.maintenance.recalculate_usage();
```
yomybaby: Without handling promise status, we can pass the promise like this. 
```
const notiKey = upsertNotification({
      message: t('maintenance.RecalculateUsage'),
      description: t('maintenance.Recalculating'),
      open: true,
      backgroundTask: {
        status: 'pending',
        promise: baiClient.maintenance.recalculate_usage(),
        statusDescriptions: {
          pending: t('maintenance.Recalculating'),
          resolved: t('maintenance.RecalculationFinished'),
          rejected: t('maintenance.RecalculationFailed'),
        },
      },
    });
```
Also we don't need to set duration to 0.

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,220 @@
...
+  // hooks
+  const { t } = useTranslation();
+  const [curTabKey, setCurTabKey] = useQueryParam('tab', tabParam);
+  // funcs
+  const getCardItemByTabKey = () => {
+    if (curTabKey === 'maintenance') {
+      return <MaintenanceSettingList />;
+    }
+    return null;
+  };
```
yomybaby: For better readability, it would be better to move the body of this function to line 215.

yomybaby (changes requested):

react/src/components/SettingList.tsx
nowgnuesLee: Thank you! I'm going to check it.

react/src/pages/MaintenancePage.tsx
nowgnuesLee: I tried to use `backgroundTask.promise` for handling async function, but if I use it like that, I can't update `percent` value when it resolved. In order to resolve this issue, I suggested this [issue](https://lablup.atlassian.net/browse/FR-352?atlOrigin=eyJpIjoiYjMxYWFhZDFlZjU3NDcyOGFhNTc0ZTYwNmIzZDFjOTQiLCJwIjoiaiJ9). However, if this percentage feature is not immediately necessary, I will go ahead and adjust it as you suggested. And I will remove duration '0'.

react/src/pages/MaintenancePage.tsx
nowgnuesLee: Thank you, I'll check it out.

react/src/pages/MaintenancePage.tsx
yomybaby: The absence of percent display when handling promises is intentional. This is because promises cannot measure percent completion. Improvements related to the progress display of promises will be discussed in a separate issue.
Therefore, please remove  specifying percent.

nowgnuesLee: I understand. I will remove an re-request issue review. Thank you!

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,203 @@
...
+            upsertNotification({
+              message: err.message,
+              description: painKiller.relieve(err.title),
+              open: true,
+            });
+          }
+          reject(err);
+        })
+        .finally(() => setIsRecalculating(false));
+    });
```
yomybaby, line 50: `baiClient.maintenance` already returns a promise, so you don't need to create a `new Promise`.

yomybaby (changes requested):

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,203 @@
...
-              }}
-            >
-              {t('maintenance.MatchDatabase')}
-            </Typography.Title>
-            <Trans>{t('maintenance.DescMatchDatabase')}</Trans>
+  const rescanImages = async () => {
+    setIsRescanning(true);
+    const notiKey = upsertNotification({
+      message: t('maintenance.rescanImages'),
+      description: t('maintenance.RescanImageScanning'),
```
yomybaby: Key name has changed
```suggestion
      message: t('maintenance.RescanImages'),
```

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,203 @@
...
+            upsertNotification({
+              message: err.message,
+              description: painKiller.relieve(err.title),
+              open: true,
+            });
+          }
+          reject(err);
+        })
+        .finally(() => setIsRecalculating(false));
+    });
```
yomybaby: `baiClient.maintenance` already returns a promise, so you don't need to create a `new Promise`.

yomybaby (changes requested):

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,194 @@
...
+  // state
+  const [isRecalculating, setIsRecalculating] = useState(false);
+  const [isRescanning, setIsRescanning] = useState(false);
+  // hooks
   const { t } = useTranslation();
+  const { upsertNotification } = useSetBAINotification();
+  const baiClient = useSuspendedBackendaiClient();
+  const painKiller = usePainKiller();
+  // funcs
+  const recalculateUsage = async () => {
```
ironAiken2: This function doesn't seem to need to be async.

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,194 @@
...
+        status: 'pending',
+        promise: baiClient.maintenance
+          .recalculate_usage()
+          .catch((err: any) => {
+            if (err && err.message) {
+              upsertNotification({
+                message: err.message,
+                description: painKiller.relieve(err.title),
+                open: true,
+              });
```
ironAiken2: This part shows two notifications. You can also deep merge the same notification by setting the key and adjusting the status for the backgroundTask.

```suggestion
              upsertNotification({
                key: <key_string>
                message: err.message,
                description: painKiller.relieve(err.title),
                open: true,
                backgroundTask: {
                    status: 'rejected'
                }
              });
```

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,194 @@
...
+      upsertNotification({
+        key: notiKey,
+        backgroundTask: {
+          status: 'rejected',
+        },
+        duration: 1,
+      });
+    } finally {
+      setIsRescanning(false);
+    }
```
ironAiken2: This is when the backgroundTask id is returned, not when the backgroundTask ends. Therefore, another method is needed to activate the button after the backgroundTask is completely finished (rescan image).

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,194 @@
...
+          description: t('maintenance.DescMatchDatabase'),
+          children: (
             <Button
-              block
-              disabled={recalculating}
               icon={<RedoOutlined />}
-              onClick={() => dispatchEvent('recalculate', null)}
-              style={{ marginTop: '12px' }}
+              onClick={recalculateUsage}
+              disabled={isRecalculating}
```
ironAiken2: In WebUI, we are using loading to indicate loading.

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,194 @@
...
+import { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { StringParam, useQueryParam, withDefault } from 'use-query-params';
 
-const MaintenancePage: React.FC = () => {
+type TabKey = 'maintenance';
+type SettingGroup = {
+  title: string;
+  settingItems: SettingItemProps[];
+};
```
ironAiken2: This type doesn't seem to be of interest to this component. it seems better to declare it in a component that uses it directly, then import it and use it.

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,194 @@
...
+
+const tabParam = withDefault(StringParam, 'maintenance');
+
+const MaintenanceSettingList = () => {
+  // state
```
ironAiken2: Uncomment unless you really need to explain something further :)

react/src/pages/MaintenancePage.tsx
```diff
@@ -1,89 +1,194 @@
...
+          {
+            key: 'maintenance',
+            tab: t('webui.menu.Maintenance'),
+          },
+        ]}
+        styles={{ body: { padding: 0 } }}
+      >
+        {curTabKey === 'maintenance' && <MaintenanceSettingList />}
+      </Card>
+    </>
```
ironAiken2: Please separate component's interests in case the maintenance page gets a lot of other features added. 

[ref](https://github.com/lablup/backend.ai-webui/blob/0de11e40119782ddba4828a655c0020b0d8ad4a0/react/src/pages/EnvironmentPage.tsx)

ironAiken2 (changes requested):

react/src/pages/MaintenancePage.tsx
nowgnuesLee: It doesn't. I dont't use `await` anymore. Thank you!

react/src/pages/MaintenancePage.tsx
nowgnuesLee: I got it. I'll remove it. Thank you!

react/src/pages/MaintenancePage.tsx
nowgnuesLee: I'm sorry, but I don't get it. Could you tell me how can I separate component's interests? Did you mean that I need to make a another file for MaintenanceSettingList?

react/src/pages/MaintenancePage.tsx
nowgnuesLee: It seems like. I'll move it, Thank you!

react/src/pages/MaintenancePage.tsx
nowgnuesLee: Thank you, but there is no key to use. And I can't pass specific value through `reject`. I should first check if this notification is really necessary before making any changes. (Because I programmed it based on the legacy code.)

react/src/pages/MaintenancePage.tsx
nowgnuesLee: It is, but it seems that there is currently no way to track the state of the `backgroundTask`. Also, this feature is still preventing the request from being made before the `baiClient.maintenance.rescan_images()` function finishes. So, I think it would be better to address [this issue](https://lablup.atlassian.net/browse/FR-352) together with the current one. What do you think?

react/src/pages/MaintenancePage.tsx
nowgnuesLee: Oh, Thank you! I'll change it later.

agatha197:
please resolve the conflicts.

react/src/pages/MaintenancePage.tsx
ironAiken2: key can be arbitrarily specified. Simply remove the rejected string from the first notification declaration and deep merge it to the rejected state.

react/src/pages/MaintenancePage.tsx
ironAiken2: Yes, currently there is a part that shows the tab card and a part that shows the maintenance setting group settings.

In my opinion, for maintenance, code readability, and scalability, it's better to separate the files as you mentioned.

react/src/pages/MaintenancePage.tsx
nowgnuesLee: The keys will be merged, and in `useNotification`, the `description` will be replaced with the `err.message` value in the `catch`. Also, since the notification seems to display the same content twice, it might be better to remove the `catch` for that promise.

react/src/components/MaintenanceSettingList.tsx
```diff
@@ -0,0 +1,150 @@
...
+        statusDescriptions: {
+          pending: t('maintenance.Recalculating'),
+          resolved: t('maintenance.RecalculationFinished'),
+          rejected: t('maintenance.RecalculationFailed'),
+        },
+      },
+    });
+  };
+
+  const rescanImages = async () => {
```
ironAiken2: Since rescan_images() displays the error via ok and msg, it doesn't seem to need to be determined by a try_catch statement.

```suggestion
  const rescanImages = async () => {
    setIsRescanning(true);

    const { rescan_images } = await baiClient.maintenance
      .rescan_images()
      .finally(() => setIsRescanning(false));

    if (!rescan_images.ok) {
      upsertNotification({
        message: t('maintenance.RescanImages'),
        description: rescan_images.msg ?? t('maintenance.RescanFailed'),
      });
      return;
    }

    upsertNotification({
      key: rescan_images.task_id,
      message: t('maintenance.RescanImages'),
      open: true,
      duration: 0,
      backgroundTask: {
        status: 'pending',
        percent: 0,
        taskId: rescan_images.task_id,
        statusDescriptions: {
          pending: t('maintenance.RescanImageScanning'),
          resolved: t('maintenance.RescanImageFinished'),
          rejected: t('maintenance.RescanFailed'),
        },
      },
    });
  };
```

react/src/components/MaintenanceSettingList.tsx
```diff
@@ -0,0 +1,150 @@
...
+        duration: 1,
+      });
+    } finally {
+      setIsRescanning(false);
+    }
+  };
+
+  const settingGroupList: SettingGroup = [
+    {
+      title: t('maintenance.Fix'),
```
ironAiken2, line 95: I'm not sure that Fix is sufficiently representative of those features (image rescan, resource recalculation). Asking @YEONFEEL96 for advice seems like a good idea.

ironAiken2 (changes requested):

agatha197 (changes requested):
please use showXXX instead of hideXXX.

ironAiken2 (changes requested):
The type of showSearchBar is boolean, so you don't need to cast it if you declare an initial value.

ironAiken2 (approved):
LGTM!

react/src/components/MaintenanceSettingList.tsx
```diff
@@ -0,0 +1,140 @@
...
+            key: notiKey,
+            backgroundTask: {
+              status: 'pending',
+              percent: 0,
+              taskId: rescan_images.task_id,
+              statusDescriptions: {
+                pending: t('maintenance.RescanImageScanning'),
+                resolved: t('maintenance.RescanImageFinished'),
+                rejected: t('maintenance.RescanFailed'),
+              },
```
agatha197, line 66:
```suggestion
              },
              to: '/environment',
              toText: 'Go to environment page'
```
How about setting `to` and `toText` props? This is just an example, so feel free to set the toText as you like. Don't forget to apply i18n.

agatha197 (dismissed):

react/src/components/MaintenanceSettingList.tsx
nowgnuesLee: Then, I think it would be better to use `to` and `toText` when the task finishes successfully. However, there is currently no way to check if it has finished or not. So, I will consider using `to` and `toText` when the `notification` can track the state of the task. What do you think?

yomybaby: Yes, It's better to improve after adding new feature to BAINotification in separated issue.

yomybaby (approved):
LGTM

agatha197 (approved):
LGTM

agatha197 (approved):
LGTM

PR #3050, opened 2025-01-20
fix: add max value for User/Project setting modal

nowgnuesLee:
resolves #NNN (FR-MMM)

- [ ] Documentation


react/src/components/KeypairResourcePolicySettingModal.tsx
```diff
@@ -218,17 +221,18 @@ const KeypairResourcePolicySettingModal: React.FC<
...
-              }
-              if (errors && errors?.length > 0) {
-                const errorMsgList = _.map(errors, (error) => error.message);
-                for (const error of errorMsgList) {
-                  message.error(error, 2.5);
+              if (!res?.create_keypair_resource_policy?.ok || errors) {
+                if (res.create_keypair_resource_policy?.msg) {
+                  message.error(
+                    painKiller.relieve(res.create_keypair_resource_policy.msg),
+                  );
```
ironAiken2: Basically, I think we've been talking about returning a massage that shows the graphql error message directly to the user, so we don't need to include the message handling logic in painkiller. 

I looked through the history, but I didn't find anything, so please check this again.

react/src/components/KeypairResourcePolicySettingModal.tsx
```diff
@@ -218,17 +221,18 @@ const KeypairResourcePolicySettingModal: React.FC<
...
+                } else if (errors && errors.length > 0) {
+                  // if res.create_keypair_resource_policy is null
```
ironAiken2: Remove any comments that aren't absolutely necessary.

react/src/components/KeypairResourcePolicySettingModal.tsx
```diff
@@ -329,65 +334,74 @@ const KeypairResourcePolicySettingModal: React.FC<
...
-                          {
-                            validator(__, value) {
-                              if (
-                                _.includes(resourceSlotKey, 'mem') &&
-                                value &&
-                                // @ts-ignore
-                                convertBinarySizeUnit(value, 'p').number >
+                  {_.chain(resourceSlotKeys)
+                    .map((resourceSlotKey) => (
+                      <Col
```
ironAiken2: Is there a reason for this change?

react/src/components/KeypairResourcePolicySettingModal.tsx
```diff
@@ -245,19 +249,20 @@ const KeypairResourcePolicySettingModal: React.FC<
...
+                    painKiller.relieve(res.modify_keypair_resource_policy.msg),
+                  );
+                } else if (errors && errors.length > 0) {
+                  // if res.create_keypair_resource_policy is null
+                  errors.forEach((error) =>
+                    message.error(painKiller.relieve(error.message), 2.5),
+                  );
                 }
-                onRequestClose();
+                onRequestClose(false);
```
ironAiken2: It looks like the condition for the if statement is set incorrectly. And if you have multiple nested conditions in an if statement, an if-return statement may be more readable than an if-else statement. This is a code styling difference, so just give it some thought.

```suggestion
              if (
                !res?.create_keypair_resource_policy?.ok &&
                res.create_keypair_resource_policy?.msg
              ) {
                message.error(
                  painKiller.relieve(res.create_keypair_resource_policy.msg),
                );
                return;
              }

              if (errors && errors.length > 0) {
                errors.forEach((error) =>
                  message.error(painKiller.relieve(error.message), 2.5),
                );
                return;
              }
              message.success(t('resourcePolicy.SuccessfullyCreated'));
              onRequestClose(true);
```

react/src/components/KeypairResourcePolicySettingModal.tsx
```diff
@@ -329,65 +334,74 @@ const KeypairResourcePolicySettingModal: React.FC<
...
                             },
-                          },
-                        ]}
-                      >
-                        {_.includes(resourceSlotKey, 'mem') ? (
-                          <DynamicUnitInputNumber />
-                        ) : (
-                          <InputNumber
-                            min={0}
-                            step={
```
ironAiken2, line 376: It seems that values larger than the MAX_CPU_QUOTA value are allowed, please double check the max value for each form item.

ironAiken2 (changes requested):

react/src/components/KeypairResourcePolicySettingModal.tsx
nowgnuesLee: It was. Thank you!

react/src/components/KeypairResourcePolicySettingModal.tsx
nowgnuesLee: Oh, thank you. I'll remove it.

react/src/components/KeypairResourcePolicySettingModal.tsx
nowgnuesLee: The structure of HOF (Higher-Order Functions) often requires reading the code from right to left, which I believe negatively impacts readability. (And honestly, I switched to the chaining approach because I found the code difficult to read.)

react/src/components/KeypairResourcePolicySettingModal.tsx
nowgnuesLee: I agree with that. That’s why I plan to use a `painkiller` only when the error message might be `undefined`, as I think it’s strange not to show any message for an error. What do you think?

react/src/components/KeypairResourcePolicySettingModal.tsx
nowgnuesLee: Since the CPU value is of type `Float`, values larger than `1e16` are technically allowed. However, after discussing this with @fregataa , we decided to limit the maximum value to `1e16`. (Is what you asked related to this?)

react/src/components/KeypairResourcePolicySettingModal.tsx
ironAiken2: Yes. For the future, I think fixing the error message is the right way to go. For now, It would be better to output a specific phrase via i18n when the error message is undefined, rather than via painkiller.

[Teams](https://teams.microsoft.com/l/message/19:14c484402d874dafb15806d093b95a82@thread.skype/1737525257159?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=74ae2c4d-ec4d-4fdf-b2c2-f5041d1e8631&parentMessageId=1737360561715&teamName=devops&channelName=Frontend&createdTime=1737525257159)

react/src/components/KeypairResourcePolicySettingModal.tsx
ironAiken2: In my opinion, that code doesn't use lodash methods enough to use chains, so the addition of chains actually makes it less readable. The second chain only uses the map method, so there's no need to use them too.

ironAiken2 (changes requested):
please resolve the conflicts

ironAiken2 (changes requested):
There seems to be a conflict in the const-vars.ts file.

nowgnuesLee:
I resolved that conflict in latest commit.

lizable:
@nowgnuesLee Still I have conflicts. Would you resolve it and re-request review again?

lizable (changes requested):
Please resolve the conflicts.

resources/i18n/en.json
```diff
@@ -1194,6 +1194,7 @@
     "Name": "Name",
     "AllowedHosts": "Allowed hosts",
     "CannotCreateResourcePolicy": "Cannot create Resource Policy. Please check input values.",
+    "CannotUpdateResourcePolicy": "Cannot update Resource Policy. Please check input values.",
```
yomybaby: We couldn't say "Please check input values" because this PR might display this message even when the user input is correct.

react/src/components/KeypairResourcePolicySettingModal.tsx
```diff
@@ -218,24 +219,27 @@ const KeypairResourcePolicySettingModal: React.FC<
...
-                const errorMsgList = _.map(errors, (error) => error.message);
-                for (const error of errorMsgList) {
-                  message.error(error, 2.5);
-                }
-                onRequestClose();
-              } else {
-                message.success(t('resourcePolicy.SuccessfullyCreated'));
-                onRequestClose(true);
+              if (errors && errors.length > 0) {
+                errors.forEach((error) => message.error(error.message, 2.5));
```
yomybaby, line 231: Is there any reason to set it to 2.5 only here?

yomybaby (changes requested):

react/src/components/KeypairResourcePolicySettingModal.tsx
nowgnuesLee: I set it to 2.5 seconds since it was set that way in the UserResourcePolicyList. What would be an appropriate duration?

resources/i18n/en.json
nowgnuesLee: I expect the detailed error message to be provided in `err.message` from the `response`. If not, I assume it would be an error that occurs in such cases. So, what do you think about simply displaying ‘Cannot update Resource Policy.’? I will check whether the backend provides sufficient error explanations.

nowgnuesLee (commented):
I couldn’t find any conflicts on my side. I have updated the branch, but if you still encounter any conflicts, please let me know.

nowgnuesLee:
* **#3050** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/3050?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/3050?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=3050).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/UserResourcePolicySettingModal.tsx
```diff
@@ -285,7 +303,12 @@ const UserResourcePolicySettingModal: React.FC<Props> = ({
               label={t('storageHost.MaxFolderSize')}
               style={{ width: '100%', margin: 0 }}
             >
-              <InputNumber min={0} addonAfter="GB" style={{ width: '100%' }} />
+              <InputNumber
+                min={0}
+                max={Number.MAX_SAFE_INTEGER}
+                addonAfter="GB"
+                style={{ width: '100%' }}
+              />
```
ironAiken2: I get an error when I put a max value.

react/src/components/ProjectResourcePolicySettingModal.tsx
```diff
@@ -276,7 +293,12 @@ const ProjectResourcePolicySettingModal: React.FC<Props> = ({
               label={t('storageHost.MaxFolderSize')}
               style={{ width: '100%', margin: 0 }}
             >
-              <InputNumber min={0} addonAfter="GB" style={{ width: '100%' }} />
+              <InputNumber
+                min={0}
+                max={Number.MAX_SAFE_INTEGER}
+                addonAfter="GB"
+                style={{ width: '100%' }}
+              />
```
ironAiken2: Same as above

ironAiken2 (changes requested):

react/src/components/UserResourcePolicySettingModal.tsx
nowgnuesLee: The value was later converted to a bytes value, causing an overflow. So I set the max value to `Number.MAX_SAFE_INTEGER` divided by `10^9` to prevent that error.

ironAiken2 (approved):
LGTM

yomybaby (approved):
LGTM

PR #3065, opened 2025-01-21
feature(FR-352): update-useBAINotification

### LINKED ISSUE #3060: Update the value upon receiving the resolved value of a background task’s promise in usBAINotification.
## Description

Currently, `useBAINotification` tracks the state of `backgroundTask.promise` to update the `notification`’s status. However, in cases where the `notification`’s status needs to be updated after the promise is resolved, it cannot be updated directly through the promise. Therefore, I plan to pass the value to be updated through the `resolve` or `reject` of the promise to update the `notification`’s status.

## Plans

In `resolve()` or `reject()`, when passing a `value` of type `Partial<Omit<NotificationState, 'created'>>`, it performs a deep merge using `lodash.merge`. Then, using the updated value, the `upsertNotification` function is executed.

(Previously, no value was passed in resolve() or reject(), so it was expected that no side effects would occur.)

- If this is applied, we can handle promise more cool
  - **example**
    ```javascript
    upsertNotification({
      ...,
      backgroundTask: {
        ...,
        promise: asyncFunc().then((res) => {
          Promise.resolve({
            ...,
            taskId: res.taskId,
            status: 'pending',
          })
        })
      }
    })
    ```

nowgnuesLee:
resolves #3060, [(FR-352)](https://lablup.atlassian.net/browse/FR-352)

I added functionality to handle `task's status` based on the notification in `useNotification`. Since the existing approach was not changed, no additional modifications are needed.

**changes**
* update `useBAINotification` to handle task's status
* refactor `MaintenanceSetttingList` component (I have confirmed that it is working properly.)

- [ ] Documentation

### DIFF
--- a/react/src/components/ComputeSessionNodeItems/ContainerCommitModal.tsx
+++ b/react/src/components/ComputeSessionNodeItems/ContainerCommitModal.tsx
@@ -58,7 +58,7 @@ const ContainerCommitModal: React.FC<ContainerCommitModalProps> = ({
               backgroundTask: {
                 taskId: res.task_id,
                 status: 'pending',
-                statusDescriptions: {
+                onChange: {
                   pending: t('session.CommitOnGoing'),
                   resolved: t('session.CommitFinished'),
                   rejected: t('session.CommitFailed'),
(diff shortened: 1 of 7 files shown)


yomybaby (changes requested):
I push a refactoring commit. Please check it.

nowgnuesLee:
I couldn’t find a better way to handle typing. Although I did come across a method, the more I used it, the more confusing it became. So, I’ve decided to stick with my current approach.

yomybaby (approved):
LGTM

PR #3111, opened 2025-02-07
misc(FR-473): missing i18n update and remove table token value

### LINKED ISSUE #2974: Add UI for Model Service Autoscaling
This UI is related to the changes made in https://github.com/lablup/backend.ai/pull/3277.

We need to update the UI to reflect the metrics and their usage once they are defined. Additionally, documentation updates are required in the `backend.ai-docs-webui` repository to guide users on the new autoscaling feature.

agatha197:
resolves #2974 (FR-473)

Added internationalization support for auto-scaling rule table headers in the endpoint detail page. The following columns are now translated across all supported languages:
- Metric Name
- Metric Source
- Min/Max Replicas
- Last Triggered
- Created At

**Checklist:**
- [ ] Documentation

### DIFF
--- a/react/src/pages/EndpointDetailPage.tsx
+++ b/react/src/pages/EndpointDetailPage.tsx
@@ -640,7 +640,7 @@ const EndpointDetailPage: React.FC<EndpointDetailPageProps> = () => {
               },
             },
             {
-              title: 'Metric Name',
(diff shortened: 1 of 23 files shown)


yomybaby (approved):
LGTM

PR #3301, opened 2025-03-04
fix(FR-620): folder permission refresh folder list after updating folder permission in FolderExplorer

### LINKED ISSUE #3330: Chore: clean up unused lit components
There are lit components that seem to no longer be in use since migrating to React. It is necessary to confirm whether they are truly unused before proceeding with deletion.  
Components that appear to no longer be in use:

- backend-ai-resource-preset-list
- backend-ai-environment-list

When deleting the files of these components, even if they are not currently in use, they may still be loaded, so the related code must be carefully checked before deletion.

yomybaby:
resolves #3330 (FR-620)

Adds a new event listener to refresh the folder list when folder permissions are modified in the legacy folder explorer. This ensures the storage list view stays synchronized with permission changes made through the UI.

**Checklist:**
- [ ] Documentation

### DIFF
--- a/react/src/components/InviteFolderSettingModal.tsx
+++ b/react/src/components/InviteFolderSettingModal.tsx
@@ -270,8 +270,8 @@ const InviteFolderSettingModal: React.FC<InviteFolderSettingModalProps> = ({
                             { label: t('data.folders.Edit'), value: 'rw' },
                           ]}
                           defaultValue={perm}
-                          onChange={() => {
-                            handlePermission(record.shared_to.uuid, perm);
+                          onChange={(nextPerm) => {
+                            handlePermission(record.shared_to.uuid, nextPerm);
                           }}
                         />
                         <Popconfirm
--- a/react/src/components/LegacyFolderExplorer.tsx
+++ b/react/src/components/LegacyFolderExplorer.tsx
@@ -1,10 +1,12 @@
 import { toGlobalId } from '../helper';
 import { useSuspendedBackendaiClient } from '../hooks';
+import { useCurrentUserInfo } from '../hooks/backendai';
 import { useTanMutation } from '../hooks/reactQueryAlias';
 import { usePainKiller } from '../hooks/usePainKiller';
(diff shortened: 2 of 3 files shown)


ironAiken2 (changes requested):
I think the refetch should also happen on `VFolderNodeListPage` that uses vfolder node list, when mount permissions or share permissions are changed.

yomybaby:
Instead of refreshing the list of loaded vfolder nodes, I modified it to fetch the updated data of vfolder nodes so that it automatically updates in the Relay environment.

ironAiken2:
I got an error on `LegacyFolderExplorerPermissionRefreshQuery`

ironAiken2 (changes requested):
I got an error on `LegacyFolderExplorerPermissionRefreshQuery`

ironAiken2 (changes requested):
Updates by relay environment sometimes don't seem to run in time. Mount permissions are not displayed correctly in the list if the permissions are modified multiple times.

lizable (changes requested):
It doesn't seem to be working properly. After I changed the permission and refresh both inviter and invitee, but the permission didn't change.

How to reproduce:
1. User B shares Vfolder to user A(admin) with Read-only permission.
2. User A(admin) accept the invitation.
3. User B changes permission to Read/Write.
4. Refresh Both User A and User B login session.
5. See there's no changes.

Please refer to recording below:

[Screen Recording 2025-03-05 at 5.10.09 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/media/video/Dgoz5XqHffJdivyccymr/1d7e3383-f6ea-4410-80e0-433b86126841.mov)

ironAiken2 (approved):
LGTM!

The issue I mentioned is addressed in Backend.AI Core [#3900PR](https://github.com/lablup/backend.ai/pull/3900).

lizable (approved):
Works fine. LGTM.

PR #3420, opened 2025-03-27
fix(FR-727): Modify ChatSender's style to follow Antd's design system

### LINKED ISSUE #3419: Dark mode is not applied to the ChatSender component

nowgnuesLee:
resolves #3419 (FR-727)
This PR resolves dark mode bug in the `ChatSender` component.

**changes**
* remove the `background` style to enable dark mode
* remove unnecessary `border-radius` style

|before|after|
|------|-----|
| |  |

- [ ] Documentation

### DIFF
--- a/react/src/components/Chat/ChatInput.tsx
+++ b/react/src/components/Chat/ChatInput.tsx
@@ -44,9 +44,7 @@ const ChatInput: React.FC<ChatInputProps> = ({
   const { token } = theme.useToken();
 
   const ChatInputStyle = {
-    background: '#ffffff',
-    borderTop: '1px solid #f0f0f0',
-    borderRadius: '0 0 8px 8px',
+    borderTop: `1px solid ${token.colorBorderSecondary}`,
     padding: token.paddingContentVertical,
   };
 


react/src/components/Chat/ChatInput.tsx
```diff
@@ -44,9 +44,7 @@ const ChatInput: React.FC<ChatInputProps> = ({
   const { token } = theme.useToken();
 
   const ChatInputStyle = {
-    background: '#ffffff',
-    borderTop: '1px solid #f0f0f0',
-    borderRadius: '0 0 8px 8px',
```
ragingwind, line 49: Why removed radius too?

yomybaby: If borderRadius revert needed, please use `token.borderRadiusLG`. It's default value is `8`. Don't forget to put the `px` when using token in `css`. `${token.borderRadiusLG}px`

ragingwind: It was moved from origin style (sorry that I couldn't find where was). It can be kept removed If you think it was unnecessary. It's not a big deal.

nowgnuesLee: Thank you! I removed it because it doesn't affect the UI. But if you notice anything strange, feel free to let me know.

ragingwind (approved):

yomybaby (approved):
LGTM

PR #3438, opened 2025-04-02
fix(FR-750): improve resource slot display in SessionSlotCell component

### LINKED ISSUE #3437: Formatting of resource slot cell in Session list
### Fixed number of memory

- Set memory display with fixed decimal precision (3 digits)

{width=680 height=64}

### New accelerator type

- If the accelerator value is 0, it should not be displayed.
- Accelerators that are registered on the server but not added to device-metadata.json should also be displayed.

{width=536 height=229}

JIRA Issue: FR-750

yomybaby:
Resolves #3437 (FR-750)

# Enhance Resource Slot Display in Session Slot Cell

This PR improves the display of resource slots in the SessionSlotCell component:

1. Replaces `deviceMetadata` with `mergedResourceSlots` from the `useResourceSlotsDetails` hook
2. Enhances memory display with fixed decimal precision (3 digits)
3. Filters out accelerators with zero or invalid values
4. Uses `display_unit` from resource slots data instead of `human_readable_name`
5. Modifies the `useResourceSlotsDetails` hook to:
   - Accept empty resource group name to fetch all resource slots
   - Handle URL parameter construction more robustly
   - Return an empty object instead of undefined when manager version is incompatible

Before 1: New accelerator type and 0 accelerator type

After 1: New accelerator type and 0 accelerator type

Before 2: 

After 2:

### DIFF
--- a/react/src/components/ComputeSessionNodeItems/SessionSlotCell.tsx
+++ b/react/src/components/ComputeSessionNodeItems/SessionSlotCell.tsx
@@ -17,7 +17,7 @@ const SessionSlotCell: React.FC<OccupiedSlotViewProps> = ({
   sessionFrgmt,
   mode = 'requested',
 }) => {
-  const { deviceMetadata } = useResourceSlotsDetails();
+  const { mergedResourceSlots } = useResourceSlotsDetails('');
   const session = useFragment(
     graphql`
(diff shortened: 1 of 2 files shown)


nowgnuesLee (approved):
LGTM

PR #3555, opened 2025-04-21
feat(FR-624): Add chat history

### LINKED ISSUE #3305: Add chat tab for multiple chat session

ragingwind:
resolves #3305 (FR-624)

# Add chat conversation management with local storage persistence

This PR adds a new chat conversation management system with local storage persistence. Key features include:

- Implements a `ChatCacheProvider` that stores chat data in localStorage
- Adds conversation management with ability to create, switch between, and delete conversations
- Creates unique IDs for chats and conversations using nanoid
- Refactors chat components to work with the new data model
- Replaces the single chat view with a tabbed interface for multiple conversations
- Improves loading states with better suspense fallbacks

**Checklist:**

- [ ] Documentation


react/src/App.tsx
```diff
@@ -142,12 +143,9 @@ const router = createBrowserRouter([
         path: '/chat',
         handle: { labelKey: 'webui.menu.Chat' },
         Component: () => {
-          const { t } = useTranslation();
           useSuspendedBackendaiClient();
           return (
-            <Suspense
-              fallback={<BAICard title={t('webui.menu.Chat')} loading />}
-            >
+            <Suspense fallback={<FlexActivityIndicator spinSize="large" />}>
```
ragingwind, line 151: Prevent display same fallback component

react/src/components/Chat/ChatCacheProvider.tsx
```diff
@@ -0,0 +1,276 @@
...
+import {
+  createContext,
+  useContext,
+  useEffect,
+  type ReactNode,
+  useCallback,
+  useRef,
+} from 'react';
+
+const createIdGenerator = () => {
```
ragingwind: Random chat, conversation id from chatbot sdk(swr)

react/src/components/Chat/ChatCacheProvider.tsx
```diff
@@ -0,0 +1,276 @@
...
+type ChatCacheDataType =
+  | ChatData
+  | ChatConversationData
+  | ChatConversationData[]
+  | string
+  | string[];
+
+type ChatCacheDataTypeSet = Map<string, ChatCacheDataType>;
+
+function chatLocalStorageProvider() {
```
ragingwind: Cache use localstorage as a primary storage provider

react/src/components/Chat/ChatCacheProvider.tsx
```diff
@@ -0,0 +1,276 @@
...
+    set(cache: ChatCacheDataTypeSet) {
+      localStorage.setItem(
+        CHAT_CACHE_KEY,
+        JSON.stringify(Array.from(cache.entries())),
+      );
+    },
+  };
+}
+
+function createChatLocalStorageCache(
```
ragingwind: In app, map will be using for save chat, conversation state

react/src/components/Chat/ChatCacheProvider.tsx
```diff
@@ -0,0 +1,276 @@
...
+}
+
+export const ChatCacheProvider: React.FC<ChatCacheProviderProps> = ({
+  children,
+}) => {
+  const cache = useRef<ChatCacheDataTypeSet>(
+    new Map<string, ChatCacheDataType>(),
+  );
+
+  useEffect(() => {
```
ragingwind: Cache will be loaded while component mounted

react/src/components/Chat/ChatCacheProvider.tsx
```diff
@@ -0,0 +1,276 @@
...
+export function useConversations() {
+  const { cache } = useChatCache();
+  const { list, remove, push, resetList } =
+    useDynamicList<ChatConversationData>();
+
+  const addConversation = useCallback(
+    (provider: ChatProviderData) => {
+      const conversationId = creatChatModelId('/conversation');
+      const chatId = creatChatModelId('/chat');
+      const maxLabelIndex =
```
ragingwind: Always use biggest index in the label in tabs

react/src/components/Chat/ChatCard.tsx
```diff
@@ -122,17 +120,20 @@ function useModels(
 
       if (res.ok) {
         return await res.json();
-      } else {
-        return { data: [], error: res?.status };
       }
+
+      return { data: [], error: res?.status };
     },
+    select: (res) => ({
```
ragingwind, line 144: Use select to migrate data of result

react/src/components/Chat/ChatCacheProvider.tsx
```diff
@@ -0,0 +1,276 @@
...
+    addConversation,
+    removeConversation,
+    reset,
+    isEmptyCache,
+    conversations: list,
+  };
+}
+
+export function useConversation(conversationId: string) {
+  const { cache } = useChatCache();
```
ragingwind: Access cache via Context

react/src/components/Chat/ChatCard.tsx
```diff
@@ -308,11 +341,15 @@ const ChatCard: React.FC<ChatCardProps> = ({
...
+          onClickCreate={() => {
+            onCreateNewChat?.(chat);
+          }}
           onClickClose={() => {
             onRequestClose?.(chat);
           }}
-          onClickDeleteChatHistory={() => {
+          onClickClearChatMessages={() => {
+            onClickClearChatMessages?.(chat);
             setMessages([]);
```
ragingwind, line 353: We should call setMessage manually 


```suggestion
            setMessages([]); // Directly clear local messages state for immediate UI update
```

ragingwind, line 368: We still need to call setMessage manually to clear chat message. useChat manage own message cache inside

react/src/components/Chat/ChatCacheProvider.tsx
```diff
@@ -0,0 +1,391 @@
...
+  const { list, remove, push, resetList } =
+    useDynamicList<ChatConversationData>();
+
+  const addConversation = useCallback(
+    (provider: ChatProviderData) => {
+      const conversationId = creatChatModelId('/conversation');
+      const chatId = creatChatModelId('/chat');
+      const maxLabelIndex =
+        list.length > 0
+          ? Math.max(...list.map((c) => extarctLabelIndex(c.label)))
```
agatha197: typo error: extarctLabelIndex -> extractLabelIndex

react/src/pages/ChatPage.tsx
```diff
@@ -42,33 +40,123 @@ function useDefaultEndpointId() {
...
+    };
+  });
+
+  const handleTabEdit = useCallback(
+    (e: TabClickEvent, action: 'add' | 'remove') => {
+      if (action === 'add') {
+        if (conversations.length < 10) {
+          addConversation(provider);
+        } else {
+          appMessage.error(`You can't add more than 10 conversations`, 5);
```
agatha197: please apply i18n.

react/src/components/Chat/ChatHeader.tsx
```diff
@@ -199,7 +200,9 @@ const ChatHeader: React.FC<ChatHeaderProps> = ({
             }}
           />
         )}
-        <Button onClick={() => onClickCreate?.()} icon={<PlusOutlined />} />
+        {clonable && (
+          <Button onClick={() => onClickCreate?.()} icon={<PlusOutlined />} />
```
agatha197: Please add a tooltip such as "Add model for comparison", "Add chat", ...

agatha197 (changes requested):
1. What about inserting a new conversation instead of pushing it? We could seamlessly add it right next to the current chat.
2. If base path is too long, it is not possible to write base path. (wide screen with many conversations)

3. Only first chatting saves added card. Is it intendeed?
[Screen Recording 2025-04-30 at 5.24.27 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/media/video/2HueYSdFvL8pOB5mgrUQ/4963a4c0-068f-428e-90e4-f570e33c7e9d.mov)
4. It would be better to save chatting index to the url so that keep current tab after refreshing. You can handle it in another PR.

ragingwind:
1. good idea, I will update it
2. it was fixed. only show input only when size is under to lg



3. it look like a bug but I couldn't reproduce it. I will check more. @agatha197 Could you capture 
the data in local-storage if you can? 



4. Will be added

ragingwind:
Regarding to 4 suggestion, as following to your idea, it would be nice to move to another task. Using subpath is better way I think. ex) chat(s)/conversation-id/chat-id

react/src/components/Chat/CustomModelForm.tsx
```diff
@@ -72,7 +75,7 @@ const CustomModelForm: React.FC<CustomModelFormProps> = ({
         <Form.Item label={t('modelService.BasePath')} name="basePath">
           <Input
             placeholder="v1"
-            addonBefore={origin}
+            addonBefore={toggleURLAddon ? undefined : endpointUrl}
             defaultValue={basePath}
           />
         </Form.Item>
```
agatha197, line 81: <p class='graphite__hidden'><i>[Re: line +83]</i></p>
Let's change the name of toggleURLAddon to others and also apply a width style to EndpointTokenSelect.

```gt_suggestion
   <EndpointTokenSelect
            loading={loading}
            endpointId={endpointId}
            style={{
              width: toggleURLAddon ? '100%' : 200,
            }}
    />
```
Before

After



<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/3555?utm_source=unchanged-line-comment">Graphite</a>.</p>

agatha197 (changes requested):
Overall, it works well. However, there are still a few minor suggestions that haven't been applied yet, so please check again.

react/src/components/Chat/ChatHeader.tsx
ragingwind: The other buttons on the chat page doesn't show any tooltip now. I don't feel that it's necessary ui.

agatha197 (approved):
LGTM!

ragingwind:
Sometimes, I couldn't get errors on the terminal. I will fix it

ragingwind:
Confirm that it's same problem. insert() makes ghost chat.

agatha197 (changes requested):
Currently, to add a new chat, I need to press the "+" button at the end of the last chat. If I press the "+" button in the middle of chats, the new chat overwrites the next one, causing the previously existing conversation to disappear unexpectedly. Doesn't this feel a bit awkward?

lizable (commented):
Works fine, but there's a one minor suggestion.
How about adding confirmation dialog when clicking "+" button at right-top of the layout which is for opening a new chat? It may help user to notify that current chat (if is on-going) will be in pause, due to user-requested operation.

react/src/pages/ChatPage.tsx
```diff
@@ -42,33 +64,259 @@ function useDefaultEndpointId() {
...
 
   return (
-    <BAICard title={t('webui.menu.Chat')} styles={ChatPageStyle}>
-      <Conversation conversation={conversation} provider={provider} />
+    <BAICard
+      title={
+        <Typography.Text
+          className={styles.fixEditableVerticalAlign}
+          editable={{
+            icon: <EditIcon style={{ color: 'black' }} />,
```
yomybaby: Let's use the default icon and style to match the other pages. We can consider changing this later all at once on all pages.
```suggestion
```

react/src/pages/ChatPage.tsx
```diff
@@ -42,33 +64,259 @@ function useDefaultEndpointId() {
...
+            }}
+          />
+          <Button
+            type="text"
+            icon={<HistoryIcon />}
+            onClick={() => {
+              setOpenHistory(!openHistory);
+            }}
+          />
+        </>
```
yomybaby: Let's add tooltip for these buttons.

react/src/pages/ChatPage.tsx
```diff
@@ -42,33 +64,259 @@ function useDefaultEndpointId() {
...
+  onClickRemove,
+  onClickHistory,
+}: ChatHistoryDrawerProps) => {
+  return (
+    <Drawer
+      getContainer={false}
+      open={open}
+      onClose={onClickClose}
+      mask={false}
+      title="History"
```
yomybaby: Please support i18n.

react/src/pages/ChatPage.tsx
```diff
@@ -42,33 +64,259 @@ function useDefaultEndpointId() {
...
+                  />
+                }
+              />,
+            ]}
+            style={{ cursor: 'pointer' }}
+            onClick={() => onClickHistory(item.id)}
+          >
+            <List.Item.Meta
+              title={item.title}
+              description={dayjs().format('YYYY-MM-DD HH:mm:ss')}
```
yomybaby: Please check this part. All item has same updated date value.
alt text

react/src/pages/ChatPage.tsx
```diff
@@ -42,33 +64,259 @@ function useDefaultEndpointId() {
...
+              navigate({ id: chat?.id, provider });
+
+              // Close history drawer if no more history items left
+              if (history.length < 2) {
+                setOpenHistory(false);
+              }
+            }}
+            onClickHistory={(historyId) => {
+              navigate({ id: historyId, provider });
+              setOpenHistory(false);
```
yomybaby: How about keeping the drawer open when clicking on history to allow users to switch history items easily?

react/src/pages/ChatPage.tsx
yomybaby, line 1: When navigating to another menu and returning to the chat, two history entries are created.

ragingwind: Because of strict mode. in production, it will be ok

react/src/pages/ChatPage.tsx
ragingwind: My bad. missed updatedAt

react/src/pages/ChatPage.tsx
ragingwind: History will be created as user do action like, change model, parameter, sync options, add chat, send messages

yomybaby (changes requested):
Please check below:
- The "New Chat" button on the chat page doesn't work.
- The chat title (history name) remains "Chat" even after sending a message. After revisiting the current history, the chat title displays correctly.

ragingwind:
Fixed the missing navigation code and resolved conflicts between user data updates and message sending in the history. I also missed testing the model change functionality and message sending that updates the chat label.

react/src/pages/ChatPage.tsx
```diff
@@ -41,32 +63,233 @@ function useDefaultEndpointId() {
...
+
+function isClonable(chatLength: number) {
+  return chatLength <= 10;
+}
+
+const ChatPage: React.FC = () => {
+  const { id } = useParams();
+
+  if (id && !getChatById(id)) {
+    window.location.href = `/chat`;
```
agatha197: How about using the useNavigate hook for this to avoid twinkling when removing last history?

resources/i18n/de.json
```diff
@@ -1,5 +1,7 @@
 {
   "$schema": "../i18n.schema.json",
+  "History": "Geschichte",
+  "New Chat": "Neuer Chat",
```
agatha197: How about replacing this to `chatui` group?
And please remove space from key.
```suggestion
  "NewChat": "Neuer Chat",
```

react/src/pages/ChatPage.tsx
ragingwind: It's for handling invalid ID, not for removing the last history. It might need other way to manage navigation for removing the last history

ragingwind: as history is empty, navigate to '/chat'

agatha197 (changes requested):
NewChat is missing in fi.json. And I left minor UI suggestion. Except for that, works fine.

react/src/pages/ChatPage.tsx
```diff
@@ -41,32 +63,237 @@ function useDefaultEndpointId() {
...
```
yomybaby:
```suggestion
    webuiNavigate(`/chat`, { replace: true });
```

react/src/pages/ChatPage.tsx
```diff
@@ -41,32 +63,237 @@ function useDefaultEndpointId() {
...
+      </BAICard>
+    )
+  );
+};
+
+function isClosable(chatLength: number) {
+  return chatLength > 1;
+}
+
+function isClonable(chatLength: number) {
```
yomybaby, line 285:
```suggestion
function isCloneable(chatLength: number) {
```

yomybaby (changes requested):

react/src/pages/ChatPage.tsx
```diff
@@ -41,32 +63,239 @@ function useDefaultEndpointId() {
...
+}: ChatHistoryDrawerProps) => {
+  return (
+    <Drawer
+      getContainer={false}
+      open={open}
+      onClose={onClickClose}
+      mask={false}
+      maskClosable={true}
+      title={t('chatui.History')}
+    >
```
agatha197, line 110: Trash button looks too far off the right.
```suggestion
     styles={{
        body: { paddingRight: 0 },
      }}
    >
```
Before:


After:

agatha197 (approved):

yomybaby (approved):
LGTM

PR #3687, opened 2025-05-21
fix(FR-956): enable desktop notifications feature

ironAiken2:
### Desktop Notification Support

This PR adds desktop notification functionality to complement the existing in-app notifications. When enabled in user settings, notifications will appear as native OS notifications, allowing users to stay informed even when the browser is not in focus.

Key changes:
- Added desktop notification implementation in `useBAINotification` hook
- Updated notification options to support both in-app and desktop notifications
- Added permission handling in user settings page
- Added error messages for unsupported browsers and denied permissions
- Updated translations for all supported languages

**Checklist:**
- [ ] Documentation


ironAiken2:
* **#3687** <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/3687?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/3687?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=3687).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/hooks/useBAINotification.tsx
```diff
@@ -442,3 +478,17 @@ function generateOverrideByStatus(
...
+ * @param message
+ * @returns string | undefined
+ * @description Extracts text from the message used in the antd's notification. If the message's type is ReactNode, it will return undefined.
+ */
+function extractTextFromMessageNode(message: ReactNode): string | undefined {
+  if (typeof message === 'string' || typeof message === 'number') {
+    return _.toString(message);
+  }
+
+  return undefined;
```
nowgnuesLee: What do you think about renaming the function to `extractTextFromReactNode` and changing it to return an empty string rather than `undefined`? Even the body type of the Notification is string. So, this function needs to make sure to return a string. And since the name of the function contains text, it would be better to return a string.

react/src/hooks/useBAINotification.tsx
```diff
@@ -400,6 +414,28 @@ export const useSetBAINotification = () => {
...
+        body: extractTextFromMessageNode(params.description),
+        icon: '/manifest/backend-ai.iconset/icon_32x32@2x.png',
+        tag: _.toString(params.key),
+      };
+
+      const notification = new Notification(title, options);
+      notification.onclick = () => {
+        if (params.to) {
+          window.focus();
+          window.location.href = _.toString(params.to);
```
nowgnuesLee: `params.to` can be `To` type. When I pass that value to the toString function, it doesn't seem to return a string as it should.  What do you think about replacing it with 
```
typeof params.to === 'string'
              ? params.to
              : createPath(params.to);
```
And let's change type of the `to` to `To`.

react/src/hooks/useBAINotification.tsx
```diff
@@ -4,7 +4,8 @@ import { App } from 'antd';
 import { ArgsProps } from 'antd/lib/notification';
 import { atom, useAtomValue, useSetAtom } from 'jotai';
 import _ from 'lodash';
-import { Key, useCallback, useEffect, useRef } from 'react';
+import { Key, ReactNode, useCallback, useEffect, useRef } from 'react';
+import { useTranslation } from 'react-i18next';
 import { To } from 'react-router-dom';
 import { v4 as uuidv4 } from 'uuid';
```
nowgnuesLee, line 11: <p class='graphite__hidden'><i>[Re: line +20]</i></p>
It seems like a good idea to change the `to` type to only have `To`, since it's already `string | Partial<Path>`, just to be clear.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.dev/github/pr/lablup/backend.ai-webui/3687?utm_source=unchanged-line-comment">Graphite</a>.</p>

react/src/hooks/useBAINotification.tsx
```diff
@@ -104,10 +110,9 @@ export const useBAINotificationEffect = () => {
               updatedNotification,
               data,
             );
-            upsertNotification(
-              _.merge({}, updatedNotification, overrideData),
-              true,
-            );
+            upsertNotification(_.merge({}, updatedNotification, overrideData), {
+              skipOverrideByStatus: true,
```
nowgnuesLee, line 114: Can you elaborate on this part? I don’t fully understand why you put skipOverrideByStatus as true. Why don't you add comments here.

nowgnuesLee (changes requested):

react/src/hooks/useBAINotification.tsx
ironAiken2: Before my modification, that value already existed and was set to true. The modification I made was to make the parameter an object type to make it clearer what was being set with the addition of the useDesktopNotification value.

react/src/hooks/useBAINotification.tsx
ironAiken2: The extracted text will be used as the title of the Notification object. Title must contain a string and does not accept the undefined type. If the extract function allowed an empty string, it could lead to the mistake of using an empty string as title. If you return undefined instead, it will raise a type error to prevent the mistake of using an empty string as the title.

I'll change the name, thank you.

nowgnuesLee (approved):
LGTM

agatha197 (dismissed):
please resolve the conflicts

lizable (approved):
Works fine in Chrome, Firefox and Safari. LGTM.

PR #3797, opened 2025-06-17
fix(FR-957): Display device name in upper case on prometheus metrics

ironAiken2:
### Improve metric title display in SessionMetricGraph component

This PR enhances the `getMetricTitle` function in the SessionMetricGraph component to better handle different types of metrics. It adds special handling for 'cuda', 'rocm', and 'io' metrics to display them in uppercase, and improves the resource slot key lookup by searching for partial matches instead of exact matches.

**How to test:**
- Connect to endpoints that support devices like CUDA or ROCm or etc.
- Access the STATISTICS page and verify that the device name is capitalized on the card in USER SESSION HISTORY.

- [ ] Documentation

### DIFF
--- a/react/src/components/SessionMetricGraph.tsx
+++ b/react/src/components/SessionMetricGraph.tsx
@@ -12,6 +12,7 @@ import { Empty, theme } from 'antd';
 import { createStyles } from 'antd-style';
 import dayjs from 'dayjs';
 import _ from 'lodash';
+import { useMemo } from 'react';
 import { graphql, useLazyLoadQuery } from 'react-relay';
 import {
   LineChart,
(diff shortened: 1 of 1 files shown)


agatha197 (changes requested):
How about using `description` of device_metadata.json? 

suggestions: 
```
mergedResourceSlots[resourceSlotKey]?.description || `${_.startCase(metricName.replaceAll('_', ' '))}`
```

nowgnuesLee (approved):
LGTM

react/src/components/SessionMetricGraph.tsx
```diff
@@ -143,17 +144,29 @@ const SessionMetricGraph: React.FC<PrometheusMetricGraphProps> = ({
...
-      return `${mergedResourceSlots[key]?.human_readable_name} ${restLabel}`;
-    }
-    if (_.includes(metricName, 'io')) {
-      return `${_.upperCase(key)} ${restLabel}`;
+    // TODO: Modify to use display name when display name is added to device metadata.
+    // Currently, cuda and rocm have the same human_readable_name in device_metadata.
+    if (deviceDescription) {
+      return `${deviceDescription} ${restLabel}`;
+    } else if (_.includes(metricName, 'io')) {
+      return `IO ${restLabel}`;
```
agatha197: It might be better to simply replace "io" with "IO" since there is no guarantee that "io" will always appear at the beginning.

agatha197 (dismissed):

lizable (approved):
A bit long but, acceptible. LGTM.

nowgnuesLee (approved):
LGTM

PR #3802, opened 2025-06-18
fix(FR-1106): prevent automatic shmem override when resource preset has specific shmem configuration

### LINKED ISSUE #3808: Prevent automati shmem override when resource preset has specific shmem
[https://lablup.atlassian.net/browse/GF-381](https://lablup.atlassian.net/browse/GF-381)

JIRA Issue: FR-1106

yomybaby:
Resolves #3808 (FR-1106)

## Summary

This PR fixes a bug where the automatic shared memory (shmem) setting was incorrectly overriding resource preset conf
igurations. When a resource preset with a specific shmem value was selected, the automatic shmem rule would always set it to 1GB when memory >= 4GB, ignoring the preset's configured value.

## Problem

When users selected a resource preset that included a specific shared memory configuration:

1. **Expected behavior**: The preset's shmem value should be used
2. **Actual behavior**: The automatic shmem rule would override the preset value with 1GB when memory >= 4GB

### Root Cause

In `ResourceAllocationFormItems.tsx`:
- Line 538: `enabledAutomaticShmem` was always set to `true` when selecting any preset
- Lines 398-399: The preset's shmem value was correctly applied to the form  
- Line 403: `runShmemAutomationRule()` was called unconditionally, overriding the preset's shmem value

### 1. Enhanced preset detection logic
- Added logic to detect when a preset has a specific shmem setting (`preset?.shared_memory && Number(preset.shared_memory) > 0`)
- Only run automatic shmem rule if the preset doesn't have a specific shmem configuration

### 2. Improved preset selection behavior
- When selecting a preset with specific shmem: disable automatic shmem (`enabledAutomaticShmem: false`)
- When selecting a preset without specific shmem: enable automatic shmem (`enabledAutomaticShmem: true`)

### `react/src/components/ResourceAllocationFormItems.tsx`

**In `updateResourceFieldsBasedOnPreset()` function:**
```typescript
// Check if preset has a specific shmem setting
const hasPresetShmem = preset?.shared_memory && Number(preset.shared_memory) > 0;

// Only run automatic shmem rule if preset doesn't have a specific shmem setting
if (!hasPresetShmem) {
  runShmemAutomationRule(mem || '0g');
}
```

**In preset selection onChange handler:**
```typescript
// Check if the selected preset has a specific shmem setting
const selectedPreset = _.find(
  checkPresetInfo?.presets,
  (preset) => preset.name === value,
);
const hasPresetShmem = selectedPreset?.shared_memory && Number(selectedPreset.shared_memory) > 0;

// If preset has specific shmem, disable automatic shmem; otherwise enable it
form.setFieldValue('enabledAutomaticShmem', !hasPresetShmem);
```

## Testing

- ✅ TypeScript compilation
- ✅ ESLint validation  
- ✅ React build process
- ✅ No breaking changes to existing functionality

## Impact

- **Users with custom shmem presets**: Will now see their configured shmem values respected
- **Users without custom shmem presets**: No change in behavior (automatic shmem still works)
- **Backward compatibility**: Fully maintained

## Related Issues

This fix ensures that resource preset configurations are properly respected and not overridden by automatic settings when explicit values are provided.

### DIFF
--- a/react/src/components/ResourceAllocationFormItems.tsx
+++ b/react/src/components/ResourceAllocationFormItems.tsx
@@ -389,18 +389,26 @@ const ResourceAllocationFormItems: React.FC<
           accelerator: Number(acceleratorObj[firstMatchedAcceleratorType] || 0),
         };
       }
+
+      // Check if preset has a specific shmem setting
+      const hasPresetShmem =
+        preset?.shared_memory && Number(preset.shared_memory) > 0;
+
       form.setFieldsValue({
         resource: {
           // ...slots,
           ...acceleratorSetting,
           // transform to GB based on preset values
           mem,
-          shmem: convertToBinaryUnit(preset?.shared_memory || 0, 'g', 2)
-            ?.displayValue,
+          shmem: convertToBinaryUnit(preset?.shared_memory || 0, 'g', 2)?.value,
           cpu: parseInt(slots?.cpu || '0') || 0,
         },
       });
-      runShmemAutomationRule(mem || '0g');
+
+      // Only run automatic shmem rule if preset doesn't have a specific shmem setting
+      if (!hasPresetShmem) {
+        runShmemAutomationRule(mem || '0g');
+      }
 
       form
         .validateFields(['resource'], {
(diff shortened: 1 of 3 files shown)


lizable (approved):
I just fixed minor thing, which defines more accurate expression about "custom allocation" not only for CPU/Mem/AI accelerator, but also shared memory(shmem).

Presume that resource preset do have shared memory configuration and it doesn't match with automatic shared memory. User tries to change the status of auto-allocate shared memory after they select resource preset that has shared memory value, It could not be just as "same" as resource preset, which should be regarded as  "custom" resource preset.

Therefore, I just added couple of code so that user can recognize after they enable/disable auto-allocate shared memory.

[Screen Recording 2025-06-19 at 1.19.37 AM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.dev/media/video/Dgoz5XqHffJdivyccymr/7a23d598-7ae8-4b7b-bf32-53007f72c49b.mov)

agatha197 (approved):
LGTM

PR #3809, opened 2025-06-18
feat(FR-1105): Improve session metrics layout and remove CPU used metric

ironAiken2:
### Enhanced User Sessions Metrics Display

This PR improves the user sessions metrics display by:

1. Implementing a responsive grid layout using Ant Design's Row and Col components to display metrics in a two-column format on larger screens
2. Enhancing metric sorting logic to group related metrics together:
   - Prioritizing CPU utilization and memory metrics
   - Grouping accelerator metrics by type (e.g., GPU)
   - Sorting accelerator metrics to show utilization before memory metrics
3. Excluding 'cpu_used' metric from display

The new layout provides a more organized view of metrics, making better use of screen space while maintaining readability on both desktop and mobile devices.

- [ ] Documentation

### DIFF
--- a/react/src/components/UserSessionsMetrics.tsx
+++ b/react/src/components/UserSessionsMetrics.tsx
@@ -4,7 +4,7 @@ import { useCurrentUserInfo } from '../hooks/backendai';
 import BAIFetchKeyButton from './BAIFetchKeyButton';
 import Flex from './Flex';
 import SessionMetricGraph from './SessionMetricGraph';
-import { Alert, DatePicker, Empty, Skeleton, theme } from 'antd';
+import { Alert, Col, DatePicker, Empty, Row, Skeleton, theme } from 'antd';
 import dayjs from 'dayjs';
 import _ from 'lodash';
 import { Suspense, useTransition } from 'react';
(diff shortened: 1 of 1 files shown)


react/src/components/UserSessionsMetrics.tsx
```diff
@@ -47,21 +47,53 @@ const UserSessionsMetrics: React.FC<PrometheusMetricProps> = () => {
...
+    const memory = _.filter(filteredMetrics, (metric) => metric === 'mem');
+
+    const acceleratorMetrics = _.filter(
+      filteredMetrics,
+      (metric) =>
+        metric &&
+        (metric.endsWith('_util') || metric.endsWith('_mem')) &&
+        !_.startsWith(metric, 'cpu') &&
+        !_.startsWith(metric, 'mem'),
+    );
```
nowgnuesLee: How about using `map` instead of `filter`.

react/src/components/UserSessionsMetrics.tsx
```diff
@@ -47,21 +47,53 @@ const UserSessionsMetrics: React.FC<PrometheusMetricProps> = () => {
...
+      filteredMetrics,
       (metric) =>
-        metric && !_.startsWith(metric, 'cpu') && !_.startsWith(metric, 'mem'),
-    ),
-  ];
+        metric &&
+        !_.includes(cpuUtil, metric) &&
+        !_.includes(memory, metric) &&
+        !_.includes(acceleratorMetrics, metric),
+    );
```
nowgnuesLee: If you use `map` above, I think `rest` would also be categorized within that function.

nowgnuesLee (changes requested):

react/src/components/UserSessionsMetrics.tsx
ironAiken2: I fixed it so that the logic is executed in a single navigation, but I think the function is less readable and understandable to others than before, is that okay?

nowgnuesLee: t doesn’t seem worse to me. I think it would be better if someone else could review it as well. :)

nowgnuesLee (approved):
LGTM

lizable (changes requested):
One minor thing, the original notation of NVIDIA CUDA platform is "CUDA", not "Cuda". It's better to be changed.

PR #3851, opened 2025-06-24
fix: React web bundle overrided by electron bundle

### LINKED ISSUE #3852: React web bundle overridden by electron bundle
The React web bundle is being overridden by the electron bundle during the build process. This needs to be fixed to ensure proper bundle separation and prevent conflicts between web and electron builds.

JIRA Issue: FR-1135

yomybaby:
Resolves #3852 (FR-1135)

# Update React build script

### Bug description:
The React web bundle is being overridden by the electron bundle during the build process. This needs to be fixed to ensure proper
  bundle separation and prevent conflicts between web and electron builds.

### Changes
This PR changes the React build script to use `build:only` instead of `build`. This also adds the `--stream` flag to the build command, which will display the output of each package's build process in real-time.


nowgnuesLee (approved):
LGTM

PR #3856, opened 2025-06-25
feat(FR-1107): Apply BAIBoard to customize prometheus metrics display

### LINKED ISSUE #3810: Add Feature to Customize Prometheus Metrics Display for User Session History
As a user, I want the ability to customize which Prometheus metrics I can view in the User Session History. This feature will allow me to select specific metrics that are relevant to my needs, enhancing my ability to analyze and monitor user sessions effectively.

JIRA Issue: FR-1107

ironAiken2:
resolves #3810 (FR-1107)

> **!Notice**
There is a bug with changing the vertical size of a card. If you increase the card vertically and then decrease the size vertically again, the graphs inside the card do not shrink.

This PR implements a draggable and resizable metrics board in the User Sessions Metrics page. Users can now customize the layout of their metrics graphs by dragging and resizing them according to their preferences. The layout is persisted in local storage so it's maintained between sessions.

- [ ] Documentation

### DIFF
--- a/react/src/components/SessionMetricGraph.tsx
+++ b/react/src/components/SessionMetricGraph.tsx
@@ -8,8 +8,8 @@ import {
   toFixedFloorWithoutTrailingZeros,
 } from '../helper';
 import { useResourceSlotsDetails } from '../hooks/backendai';
-import BAICard from './BAICard';
-import { Empty, theme } from 'antd';
+import Flex from './Flex';
+import { Empty, Typography, theme } from 'antd';
 import { createStyles } from 'antd-style';
 import dayjs from 'dayjs';
 import _ from 'lodash';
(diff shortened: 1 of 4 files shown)


nowgnuesLee (changes requested):
It's a minor issue. If there is no data to display, it will not be centered due to padding.

agatha197 (changes requested):
Can you also set the vertical center?

react/src/components/SessionMetricGraph.tsx
```diff
@@ -171,20 +171,28 @@ const SessionMetricGraph: React.FC<PrometheusMetricGraphProps> = ({
...
+    <Flex
+      direction="column"
+      align="stretch"
+      gap="sm"
+      style={{
+        height: '100%',
+        overflow: 'hidden',
       }}
     >
+      <Flex align="center" style={{ height: 55, marginLeft: 52 }}>
```
agatha197: What is the basis for this size? I'm just asking because it ends with 55 instead of 56.

ironAiken2: The handle (6 dots) that repositions the card is implemented as an absolute position, so the title should not encroach on it. I'll modify the 55 px to 56 px.

agatha197 (approved):
LGTM

nowgnuesLee (approved):
LGTM

PR #3891, opened 2025-07-03
fix(FR-1184, FR-1182): enable desktop notification when local storage setting is true

### LINKED ISSUE #3886: Desktop Notification Persists Even After Disabling Options
JIRA Issue: FR-1184

ironAiken2:
resolves #3886 (FR-1184)

### Add Desktop Notification Setting Control

This PR enhances the BAI notification system by adding desktop notification control based on user settings. Now, desktop notifications will only be shown when:

1. The notification is explicitly marked to use desktop notifications
2. The user has enabled desktop notifications in their settings
3. Remove unnecessary logo image

The implementation:
- Imports and uses the `useBAISettingUserState` hook to check if desktop notifications are enabled
- Adds a second parameter to `upsertNotification` calls with a `useDesktopNotification` flag
- Sets this flag to `false` for progress updates (to avoid excessive notifications)
- Sets this flag to `true` for completion and error notifications
- Adds a condition to check both the flag and user setting before showing desktop notifications

**Checklist:**

- [ ] Documentation

### DIFF
--- a/react/src/hooks/useBAINotification.tsx
+++ b/react/src/hooks/useBAINotification.tsx
@@ -1,6 +1,7 @@
 import { useWebUINavigate } from '.';
 import BAINotificationItem from '../components/BAINotificationItem';
 import { SSEEventHandlerTypes, listenToBackgroundTask } from '../helper';
+import { useBAISettingUserState } from './useBAISetting';
 import { App } from 'antd';
 import { ArgsProps } from 'antd/lib/notification';
 import { atom, useAtomValue, useSetAtom } from 'jotai';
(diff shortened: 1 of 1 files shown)


nowgnuesLee (approved):
LGTM

agatha197 (changes requested):
how about adding desktop notification on onFailed too?

yomybaby (approved):
-

yomybaby (changes requested):
Please change `useDesktopNotification` to `skipDesktopNotification`. By default, the WebUI sends a desktop notification when the releated user setting is true. However, this can be skipped if `skipDesktopNotification` is set to true.
```
interface NotificationOptions {
  skipOverrideByStatus?: boolean;
  useDesktopNotification?: boolean;
}
```

agatha197 (approved):
LGTM. But in my environment, it's hard to test it. Please merge it after @yomybaby reviews.

PR #3899, opened 2025-07-07
chore(FR-1195): modify column setting logic

### LINKED ISSUE #3893: Modify column setting logic to select at least one column
## description

Currently, there is no limit on the minimum number of column selections, so when you don't select anything, it looks like the table is empty. I’ll fix that problem by modifying the logic to select at least one.

hummingbbird:
resolves #3893 (FR-1195)

## description
The logic in TableColumnsSettingModal has been updated to require at least one checkbox to be selected. Previously, selecting zero checkboxes caused the table to appear empty. Now, when attempting to confirm with zero selections, a validation message is displayed.

Screenshot 2025-07-07 at 09 43 22

<br><br>

- [ ] Documentation

### DIFF
--- a/react/src/components/TableColumnsSettingModal.tsx
+++ b/react/src/components/TableColumnsSettingModal.tsx
@@ -126,6 +126,12 @@ const TableColumnsSettingModal: React.FC<TableColumnsSettingProps> = ({
                   height: 220,
                   overflowY: 'auto',
                 }}
+                rules={[
+                  {
+                    required: true,
+                    message: t('general.validation.PleaseSelectOptions'),
+                  },
+                ]}
               >
                 <Checkbox.Group
                   options={filteredColumns}
--- a/resources/i18n/de.json
+++ b/resources/i18n/de.json
@@ -210,17 +210,7 @@
     "UserStatus": "Benutzer-Status",
     "Users": "Benutzer",
     "WarningLessRateLimit": "Wenn der Wert für das Ratenlimit klein ist, können API-Vorgänge wie die Anmeldung blockiert werden.",
-    "YouAreAboutToDeleteCredential": "Sie sind dabei, die Anmeldeinformationen dieses Benutzers zu löschen:",
-    "validation": {
-      "InvalidEmailAddress": "Ungültige E-Mail-Adresse.",
-      "LetterNumber-_dot": "Erlaubt Buchstaben, Zahlen und -_.",
-      "NameAlreadyExists": "Richtlinienname ist bereits vorhanden!",
-      "NumbersOnly": "Erlaubt nur Zahlen",
-      "PleaseSelectOption": "Bitte wählen Sie eine Option",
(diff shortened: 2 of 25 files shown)


react/src/components/TableColumnsSettingModal.tsx
```diff
@@ -126,6 +126,13 @@ const TableColumnsSettingModal: React.FC<TableColumnsSettingProps> = ({
                   height: 220,
                   overflowY: 'auto',
                 }}
+                validateTrigger="onChange"
```
nowgnuesLee: I don't think we need this.

react/src/components/TableColumnsSettingModal.tsx
```diff
@@ -126,6 +126,13 @@ const TableColumnsSettingModal: React.FC<TableColumnsSettingProps> = ({
...
+                rules={[
+                  {
+                    required: true,
+                    message: t('credential.validation.PleaseSelectOptions'),
```
nowgnuesLee: Referencing key values from other components in i18n doesn't look good. What do you think about creating a key with `validation` and using it, ex. `validation.PleaseSelectOptions` cc @yomybaby

nowgnuesLee (changes requested):

react/src/components/TableColumnsSettingModal.tsx
nowgnuesLee: After removing, plz check it.

hummingbbird: It's okay to remove that. thank you!

react/src/components/TableColumnsSettingModal.tsx
hummingbbird: I agree. I think validation texts can be used in various situations, not just for credentials, so how about moving the validation section outside of credential?

nowgnuesLee: That's what i meant. But I dont think that is good idea to move the existing validation key to outside of section.

hummingbbird: I understand what you mean. I'll create a new validation section. That should make it easier to reuse for different validation cases.

resources/i18n/de.json
```diff
@@ -679,7 +669,17 @@
     "StorageProxies": "Speicher",
     "TotalItems": "Insgesamt {{Gesamt}} Artikel",
     "Username": "Nutzername",
-    "ValueRequired": "Bitte geben Sie {{ Name }}"
+    "ValueRequired": "Bitte geben Sie {{ Name }}",
+    "validation": {
+      "InvalidEmailAddress": "Ungültige E -Mail -Adresse.",
+      "LetterNumber-_dot": "",
```
nowgnuesLee: I think you missed that key value, please check again.

nowgnuesLee (changes requested):

yomybaby (approved):
LGTM

PR #3939, opened 2025-07-11
feat(FR-1230): My resource usage/capacity within the resource group

### LINKED ISSUE #3930: My resource usage/capacity within the resource group
## Summary

The issue is about tracking resource usage and capacity within a resource group.

## Context

The current setup does not provide a progress bar indicator for resource usage. The usage and remaining values should reflect the preset limits.

## Acceptance criteria

- Use the existing values for using/remaining from the preset.
- Ensure that all relevant limits are considered in the numbers.
- Implement a progress bar indicator.

## Other information

N/A

JIRA Issue: FR-1230

agatha197:
Resolves #3930 ([FR-1230](https://lablup.atlassian.net/browse/FR-1230))

# Add "My Resources in Resource Group" card to dashboard

This PR adds a new card to the dashboard that shows resource usage and available capacity within a specific resource group. The card allows users to:

- Select a specific resource group to view
- Toggle between usage and capacity views
- See CPU, memory, and accelerator resources available in the selected group
- You can check each data in the check-preset's scaling groups. (see network tabs)

The implementation includes:
- A new `MyResourceWithinResourceGroupCard` component
- Integration with the dashboard layout
- Localization strings for all supported languages

**Checklist:**
- [x] Documentation
- [x] Specific setting for review (eg., KB link, endpoint or how to setup): dogbowl

[FR-1230]: https://lablup.atlassian.net/browse/FR-1230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


react/src/components/MyResourceWithinResourceGroupCard.tsx
```diff
@@ -0,0 +1,320 @@
...
+    (values: { current: number }, resourceSlot: ResourceSlotDetail) => (
+      <BAIResourceWithSteppedProgress
+        current={values.current}
+        title={resourceSlot.human_readable_name}
+        unit={resourceSlot.display_unit}
+        showProgress={false}
+      />
+    ),
+    [],
+  );
```
nowgnuesLee: It looks better to declare it outside instead of useCallback.

react/src/components/MyResourceWithinResourceGroupCard.tsx
```diff
@@ -0,0 +1,320 @@
...
+            textAlign: 'center',
+            height: 84,
+            padding: 0,
+          }}
+        >
+          <Divider type="vertical" style={{ height: '100%' }} />
+        </Col>
+      ),
+    [xs],
+  );
```
nowgnuesLee: Simple functions look better without useCallback.

react/src/components/MyResourceWithinResourceGroupCard.tsx
```diff
@@ -0,0 +1,320 @@
...
+  ]);
+
+  const hasMemoryData = useMemo(
+    () =>
+      !_.isNaN(
+        convertToBinaryUnit(checkPresetInfo?.keypair_using.mem, 'g')
+          ?.numberFixed,
+      ),
+    [checkPresetInfo?.keypair_using.mem],
+  );
```
nowgnuesLee: Non-complex calculations are better off without useMemo.

react/src/components/MyResourceWithinResourceGroupCard.tsx
```diff
@@ -0,0 +1,320 @@
...
+    const remaining = _.get(
+      checkPresetInfo?.scaling_groups?.[resourceGroup]?.remaining,
+      resource,
+    );
+    if (remaining === Number.MAX_SAFE_INTEGER) return Number.MAX_SAFE_INTEGER;
+
+    return isMemory
+      ? _.toNumber(convertToBinaryUnit(remaining, 'g')?.numberFixed)
+      : _.toNumber(remaining);
+  };
```
nowgnuesLee: using 'auto' instead of 'g' would be better.

resources/i18n/en.json
```diff
@@ -1963,6 +1963,9 @@
       "MyEnvironments": "My Environments",
       "MyResources": "My Resources",
       "MyResourcesDescription": "<p>This panel shows your total resource usage and available capacity across all projects and resource groups.</p><p>When there are multiple limits (such as domain, project, or keypair), the most restrictive limit is applied to show your usage and remaining capacity.</p>\n",
+      "MyResourcesIn": "My Resources in",
+      "MyResourcesInResourceGroupDescription": "This panel shows your current resource usage and available capacity in the selected resource group.",
+      "MyResourcesWithinTheResourceGroup": "",
```
nowgnuesLee: is it necessary? it's a empty value.

agatha197: It is added unexpectedly. I removed it.

yomybaby (changes requested):
When auto-refresh occurs, it displays the suspense fallback. It would be better to retain the existing information instead of showing a fallback like "My sessions".

ironAiken2 (dismissed):
On smaller screen sizes, the card looks awkward with no margin. How about apply small size to the select and segment components and give them a margin? And how about sticking the refresh button in the top right corner for consistency when wrapping lines?

agatha197:
Thanks for good idea. But, if we want to show all title and extra, the Segment and refresh button should be set as a extra button. So it's hard to fix only refresh button to the top right corner.

agatha197:
I modified top margin though.

PR #3979, opened 2025-07-21
chore(FR-1153): replace CPU and MEM icons in React

### LINKED ISSUE #3860: Use the CPU and MEM icons of Lucide instead of MWC.
The current CPU and MEM icons are using the MWC Icon. Since we will soon be removing the Lit components, these icons need to be replaced with different ones. As the existing session list utilizes the MWC Icon, let's plan to change the icons when the NEO session list transitions from experimental to official.

JIRA Issue: FR-1153

yomybaby:
Resolves #3860 ([FR-1153](https://lablup.atlassian.net/browse/FR-1153))

# Replace Material Web Component Icons with Lucide Icons

|  Before  | After    |
|---------|--------|
|                |             |

This PR replaces the deprecated Material Web Component icons with Lucide icons for CPU and memory resources. Specifically:

- Removes the `MWCIconWrap` component that was using the deprecated `mwc-icon` element
- Imports `CpuIcon` and `MemoryStickIcon` from Lucide React
- Updates the resource type icons to use these Lucide components instead of the MWC icons

[FR-1153]: https://lablup.atlassian.net/browse/FR-1153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react/src/components/ResourceNumber.tsx
+++ b/react/src/components/ResourceNumber.tsx
@@ -6,7 +6,7 @@ import ImageWithFallback from './ImageWithFallback';
 import NumberWithUnit from './NumberWithUnit';
 import { Tooltip, Typography, theme } from 'antd';
 import _ from 'lodash';
-import { MicrochipIcon } from 'lucide-react';
+import { CpuIcon, MemoryStickIcon, MicrochipIcon } from 'lucide-react';
 import React, { ReactElement } from 'react';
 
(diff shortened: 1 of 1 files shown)


nowgnuesLee (approved):
LGTM

PR #3996, opened 2025-07-24
fix(FR-1282): modify service name rule to validate only when create and apply field name change

hummingbbird:
resolves FR-1282

This PR fixes the validation rule to only be applied when creating a service. It also addresses the issue by reflecting the changed field names.
(`vfoldersAliasMap` -> `mount_id_map`)

- [ ] Documentation

### DIFF
--- a/react/src/components/ServiceLauncherPageContent.tsx
+++ b/react/src/components/ServiceLauncherPageContent.tsx
@@ -116,7 +116,7 @@ interface ServiceLauncherInput extends ImageEnvironmentFormInput {
   openToPublic: boolean;
   modelMountDestination: string;
   modelDefinitionPath: string;
-  vfoldersAliasMap: Record<string, string>;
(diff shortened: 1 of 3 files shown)


react/src/components/ServiceLauncherPageContent.tsx
```diff
@@ -737,7 +737,8 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
                         label={t('modelService.ServiceName')}
                         name="serviceName"
                         validateDebounce={500}
-                        rules={validationRules}
+                        rules={!!endpoint ? [] : validationRules}
+                        // rules={!!endpoint? []:validationRules}
                       >
```
yomybaby: Please remove unused code.
```suggestion
                        rules={!!endpoint ? [] : validationRules}
                      >
```

PR #3997, opened 2025-07-24
feat(FR-1287): refactoing Session detail panel data fetching logic

yomybaby:
# Refactor ConnectedKernelList to use fragments and improve component unmounting

This PR refactors the ConnectedKernelList component to use Relay fragments instead of direct queries, improving performance and code organization. It also renames UnmountModalAfterClose to UnmountAfterClose and extends its functionality to support Drawer components.

Key changes:
- Convert ConnectedKernelList to use fragments instead of direct queries
- Rename UnmountModalAfterClose to UnmountAfterClose for better semantics
- Add support for Drawer components in UnmountAfterClose
- Optimize fetch policies in SessionDetailContent
- Add useFetchKey hook with INITIAL_FETCH_KEY constant
- Apply UnmountAfterClose to SessionDetailDrawer for proper cleanup


react/src/components/UnmountAfterClose.tsx
```diff
@@ -42,10 +42,17 @@ const UnmountModalAfterClose: React.FC<UnmountModalAfterCloseProps> = ({
 
   // Clone the child element, keeping the open prop and replacing afterClose with the new handler
   const clonedChild = React.cloneElement(modalElement, {
+    // for Modal
     afterClose: handleModalAfterClose,
+    // for Drawer
+    afterOpenChange: (open: boolean) => {
```
nowgnuesLee: I think the copilot review makes sense. what do u think about it?

react/src/components/SessionDetailContent.tsx
```diff
@@ -427,7 +436,12 @@ const SessionDetailContent: React.FC<{
           <Typography.Title level={4} style={{ margin: 0 }}>
             {t('kernel.Kernels')}
           </Typography.Title>
-          <ConnectedKernelList id={id} fetchKey={fetchKey} />
+          <ConnectedKernelList
+            kernelsFrgmt={filterNonNullItems(
```
nowgnuesLee, line 440: Just a question, it seems like the item in the map function could be undefined, but since filterNonNullItems only filters nulls, wouldn't filterEmptyItem make more sense?

ironAiken2 (changes requested):
Does the SessionDetailDrawer used in EndpointDetailPage and RecentlyCreatedSession not need to be wrapped in an UnmountAfterDrawer?

react/src/components/SessionDetailContent.tsx
yomybaby: I'll already handle in https://app.graphite.dev/github/pr/lablup/backend.ai-webui/4038

react/src/components/UnmountAfterClose.tsx
yomybaby: Already applied

PR #4033, opened 2025-07-31
fix(FR-1308): Await token login result

### LINKED ISSUE #4032: token login is not awaited
`token_login()`  function is used when a user tries to login using token

The function does not wait for the result of login-check API call, which causes multiple href “/” and it blocks token-login

JIRA Issue: FR-1308

fregataa:
resolves #4032 (FR-1308)

- [ ] Documentation


agatha197 (approved):
LGTM!

PR #4055, opened 2025-08-04
fix(FR-1320): prevent sending empty system prompt in Chat API requests

### LINKED ISSUE #4052: Fix empty system prompt causing errors in Chat with vision models
Chat feature sends empty string as system prompt when no agent is selected, causing errors with models like Llama 3 Vision.

The system prompt field should be omitted entirely when empty instead of sending an empty string.

JIRA Issue: FR-1320

yomybaby:
Resolves #4052 ([FR-1320](https://lablup.atlassian.net/browse/FR-1320))

## Summary
Fixes Chat system prompt compatibility issue with vision models by preventing empty system prompt from being sent in API requests.

## Changes
- Modified react/src/components/Chat/ChatCard.tsx to use undefined instead of empty string for system prompt when no agent is selected
- This ensures the system field is omitted from the API request payload entirely when empty

## Background
When no agent is selected in the Chat feature, an empty string was being sent as the system prompt parameter, causing compatibility issues with certain AI models like Llama 3 Vision which throw errors when receiving empty system prompts.

[FR-1320]: https://lablup.atlassian.net/browse/FR-1320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react/src/components/Chat/ChatCard.tsx
+++ b/react/src/components/Chat/ChatCard.tsx
@@ -260,7 +260,7 @@ const PureChatCard: React.FC<ChatCardProps> = ({
             middleware: extractReasoningMiddleware({ tagName: 'think' }),
           }),
           messages: body?.messages,
-          system: agent ? (agent.config.system_prompt ?? '') : '',
+          system: agent?.config.system_prompt || undefined,
           ...(chat.usingParameters ? chat.parameters : {}),
         });
 


react/src/components/Chat/ChatCard.tsx
```diff
@@ -260,7 +260,7 @@ const PureChatCard: React.FC<ChatCardProps> = ({
             middleware: extractReasoningMiddleware({ tagName: 'think' }),
           }),
           messages: body?.messages,
-          system: agent ? (agent.config.system_prompt ?? '') : '',
+          system: agent ? (agent.config.system_prompt ?? '') : undefined,
```
nowgnuesLee: Shouldn't it be `agent ? (agent.config.system_prompt ?? undefined) : undefined`?

nowgnuesLee (changes requested):

react/src/components/Chat/ChatCard.tsx
yomybaby: I change from `??` to `||` also because empt string `""` should be handle as `undefined`.

nowgnuesLee (approved):
LGTM

PR #4061, opened 2025-08-05
fix: apply babel relay diffrently for each project

nowgnuesLee:
resolves #NNN (FR-MMM)

- [ ] Documentation


react/craco.config.js
```diff
@@ -54,6 +46,53 @@ module.exports = {
...
+        babelLoader.options.overrides = [
+          {
+            include: [
+              (filePath) => filePath.includes(path.resolve(__dirname, 'src')),
+            ], // include only react/src folder
+            plugins: [
+              [
+                'relay',
+                {
+                  artifactDirectory: './src/__generated__',
```
yomybaby: Please use `path.resolve` here too.

yomybaby (approved):
LGTM

PR #4076, opened 2025-08-06
fix(FR-1332): remove try-catch wrapper from use hook in useConnectedBAIClient

ironAiken2:
### Simplify error handling in useConnectedBAIClient hook

Removes the try-catch block from the `useConnectedBAIClient` hook, simplifying the code while maintaining the same functionality. The error handling was redundant since React's error boundaries would catch any errors thrown during the hook's execution.

- [ ] Documentation


nowgnuesLee (approved):
LGTM

PR #4104, opened 2025-08-08
fix(FR-1351): replace useUpdatableState with useFetchKey in ComputeSessionListPage

yomybaby:
# Replace `useUpdatableState` with `useFetchKey` in ComputeSessionListPage

This PR replaces the usage of `useUpdatableState` with the more specific `useFetchKey` hook in the ComputeSessionListPage component. The change simplifies the code by using a dedicated hook for managing fetch key state.


agatha197 (approved):
LGTM

PR #4122, opened 2025-08-11
refactor(FR-1358): separate filtering concerns in VFolderTable to fix auto-mount detection

### LINKED ISSUE #4121: Fix row filter affecting auto-mounted folder detection in VFolderTable
A structural issue in the VFolderTable component caused auto-mounted folder lists to be passed as empty arrays to parent components.

## Background

The VFolderTable component handles multiple types of filtering:

1. Permission-based filtering (domain/group/keypair policies)
1. Project-accessible folder filtering
1. Display filtering (search keywords + external rowFilter prop)
1. Auto-mounted folder extraction (ready status folders starting with `.`)

## Problem Description

- Previously, the display rowFilter was also applied to auto-mounted folder extraction logic, causing the auto-mounted folder list passed to parent components via the `onChangeAutoMountedFolders` callback to become empty arrays
- This appears to be a side effect from refactoring other logic

## Solution Approach

- Clearly separate each filtering concern into independent useMemo implementations
- Extract auto-mounted folders only from permission-filtered folders
- Apply display filters (rowFilter + search keywords) only in displayingFolders

## Affected Components

- `@react/src/components/ServiceLauncherPageContent.tsx`
- `@react/src/pages/SessionLauncherPage.tsx`

This issue resolves the problem where auto-mounted folders are not properly detected when launching sessions and services.

JIRA Issue: FR-1358

yomybaby:
Resolves #4121 ([FR-1358](https://lablup.atlassian.net/browse/FR-1358))

## Summary

Fixed a structural issue in the VFolderTable component that caused auto-mounted folder lists to be incorrectly filtered by the display rowFilter, resulting in empty arrays being passed to parent components via the onChangeAutoMountedFolders callback.

### Filtering Concern Separation
- **Permission-based filtering**: Separated into , , and 
- **Auto-mount folder extraction**: Now only depends on permission-filtered folders, not display filters
- **Display filtering**: Isolated to  with both external rowFilter and search keywords

### Props Refactoring  
- Renamed  prop to  to clarify its purpose as a display-only filter
- Updated prop destructuring to maintain backward compatibility

### Variable Naming Improvements
-  →  (clearer purpose)
-  →  (more concise)

## Impact

This fix ensures that auto-mounted folders (folders starting with  in ready status) are correctly detected and reported to parent components in:
- 
- 

## Technical Details

The previous implementation applied the display rowFilter to all filtering operations, including auto-mount detection. This caused issues where the rowFilter would exclude auto-mountable folders from being detected, even though they should be identified regardless of display preferences.

The refactored implementation clearly separates:
1. **Data access layer**: Permission and project-based filtering
2. **Business logic layer**: Auto-mount folder identification
3. **Presentation layer**: Display filtering with search and external filters

[FR-1358]: https://lablup.atlassian.net/browse/FR-1358?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react/src/components/ServiceLauncherPageContent.tsx
+++ b/react/src/components/ServiceLauncherPageContent.tsx
@@ -933,7 +933,7 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
                               <VFolderTableFormItem
                                 rowKey={'id'}
                                 label={t('modelService.AdditionalMounts')}
-                                filter={(vf) =>
+                                rowFilter={(vf) =>
                                   vf.id !== getFieldValue('vFolderID') &&
                                   vf.status === 'ready' &&
                                   vf.usage_mode !== 'model' &&
--- a/react/src/components/VFolderTable.tsx
+++ b/react/src/components/VFolderTable.tsx
@@ -75,7 +75,7 @@ export interface VFolderTableProps extends Omit<TableProps<VFolder>, 'rowKey'> {
   aliasBasePath?: string;
   aliasMap?: AliasMap;
   onChangeAliasMap?: (aliasMap: AliasMap) => void;
-  filter?: (vFolder: VFolder) => boolean;
+  rowFilter?: (vFolder: VFolder) => boolean;
   rowKey: string | number;
   onChangeAutoMountedFolders?: (names: Array<string>) => void;
   showAutoMountedFoldersSection?: boolean;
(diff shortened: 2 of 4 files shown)


agatha197 (changes requested):
Is it necessary to update `filter` to `rowFilter` in `VFolderSelect` as well?

agatha197 (approved):
LGTM

PR #4150, opened 2025-08-14
feat(FR-1367): implement flexible CSS-based system for UI element customization

ironAiken2:
resolves #NNN (FR-MMM)

- [ ] Documentation


react/src/components/BAIBoard.tsx
```diff
@@ -99,9 +100,10 @@ const BAIBoard = <T extends BAIBoardDataType>({
         !resizable && styles.disableResize,
       )}
       empty
-      renderItem={(item: BoardProps.Item<T>) => {
+      renderItem={(item: Partial<StartPageBoardItem>) => {
         return (
           <BoardItem
+            data-testid={item?.['data-testid']}
```
yomybaby: - BoardItem doesn't set the data-testid to HTML element: The data-testid prop is being passed to the BoardItem component, but this is a Cloudscape component that likely doesn't forward this prop to the actual DOM element.
- CSS display: none won't work with Board layout: Even if the data-testid worked, using display: none on a BoardItem would hide it visually but the Board component would still treat it as existing, causing layout issues with empty spaces.

The functionality to hide the "Start model service" button based on baiClient?._config?.enableModelFolders is already implemented.

PR #4183, opened 2025-08-26
fix(FR-1404): ensure BAIUnmountAfterClose always unmounts after close animations

yomybaby:
Resolves #4183 ([FR-1404](https://lablup.atlassian.net/browse/FR-1404))

## Summary

Fixes the BAIUnmountAfterClose component to ensure it always properly unmounts Modal and Drawer components after close animations complete, regardless of whether parent components provide afterClose or afterOpenChange callback props.

## Problem

The component was using type guards to conditionally add unmounting handlers only when original props existed. This caused the component to fail to unmount when parents didn't provide these callbacks, leading to memory leaks.

## Solution

- Removed conditional type guards (`hasAfterClose`, `hasAfterOpenChange`)
- Always add our own unmounting handlers (`afterClose`, `afterOpenChange`)
- Preserve any existing callbacks while ensuring proper unmounting behavior
- Ensures component unmounts after close animations in all scenarios

## Changes

- Simplified prop handling by always applying unmounting handlers
- Reduced code complexity by removing unnecessary conditional logic
- Maintains backward compatibility with existing callback preservation

[FR-1404]: https://lablup.atlassian.net/browse/FR-1404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/backend.ai-ui/src/components/BAIUnmountAfterClose.tsx
+++ b/packages/backend.ai-ui/src/components/BAIUnmountAfterClose.tsx
@@ -28,32 +28,26 @@ const BAIUnmountAfterClose: React.FC<BAIUnmountModalAfterCloseProps> = ({
   children,
 }) => {
   // Ensure there is only one child element
-  const modalElement = React.Children.only(children);
-  const isModalOpen = modalElement.props.open;
+  const childElement = React.Children.only(children);
+  const isOpen = childElement.props.open;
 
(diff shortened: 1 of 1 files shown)


ironAiken2 (approved):
LGTM

PR #4269, opened 2025-09-12
fix(FR-1455): correct the incorrect values shown on the dashboard

nowgnuesLee:
# Fix resource allocation validation and type safety

This PR addresses issues with resource allocation validation and type safety in the React components:

1. Added null coalescing operators (`?? 0`) in `MyResourceWithinResourceGroup.tsx` to handle potentially undefined values when accessing resource usage data
2. Updated `RemainingSlots` type to make `cpu` and `mem` properties optional, and allow `undefined` values in the accelerators map
3. Fixed a potential null reference in `ResourceAllocationFormItems.tsx` by adding a fallback value when `remaining.mem` is undefined
4. Removed unnecessary lodash import in the test file
5. Modified the resource limit calculation logic to return `undefined` instead of `Number.MAX_SAFE_INTEGER` when no valid limits are found

These changes improve type safety and prevent potential runtime errors when handling resource allocation limits.

- [ ] Documentation

### DIFF
--- a/packages/backend.ai-ui/src/components/ResourceStatistics.tsx
+++ b/packages/backend.ai-ui/src/components/ResourceStatistics.tsx
@@ -71,7 +71,7 @@ const ResourceStatistics: React.FC<ResourceStatisticsProps> = ({
     return (
       <Empty
         image={Empty.PRESENTED_IMAGE_SIMPLE}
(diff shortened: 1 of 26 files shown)


yomybaby (commented):
LGTM

PR #4321, opened 2025-09-23
feat(FR-1503):Adopt pnpm minimumReleaseAge for security

nowgnuesLee:
# Add minimumReleaseAge to pnpm-workspace.yaml 

This PR adds a `minimumReleaseAge` setting of 1440 minutes (24 hours) to the pnpm-workspace.yaml file. This setting helps ensure we only use package versions that have been available for at least 24 hours, reducing the risk of using packages that might be quickly yanked due to issues.

- [ ] Documentation


yomybaby (approved):
LGTM

PR #4356, opened 2025-09-29
feat(FR-1522): add a validator to the project

agatha197:
resolves #NNN (FR-MMM)

- [ ] Documentation


react/src/components/ContainerRegistryEditorModal.tsx
```diff
@@ -397,6 +397,24 @@ const ContainerRegistryEditorModal: React.FC<
                     max: 255,
                     message: t('maxLength.255chars'),
                   },
+                  {
+                    type: 'string',
+                    min: 1,
+                    message: t('minLength.1char'),
+                  },
```
nowgnuesLee: As the required rule already exists, it appears unnecessary.

nowgnuesLee (changes requested):

nowgnuesLee (approved):
LGTM

react/src/components/ContainerRegistryEditorModal.tsx
```diff
@@ -397,6 +397,19 @@ const ContainerRegistryEditorModal: React.FC<
...
                   },
+                  {
+                    validator: (_, value) => {
+                      if (value) {
+                        // Allowed characters: lowercase letters, numbers, and separators '.', '_', '-' that are not consecutive
+                        const restrictedNamePattern =
+                          /^[a-z0-9]+(?:[._-][a-z0-9]+)*$/;
+                        if (!restrictedNamePattern.test(value)) {
+                          return Promise.reject(t('validation.SluggedString'));
+                        }
```
yomybaby:
```suggestion
                        if (!restrictedNamePattern.test(value)) {
                          if (/[._-]$/.test(value)) {
                            return Promise.reject(t('validation.SluggedStringTrailing'));
                          } else {
                            return Promise.reject(t('validation.SluggedString'));
                          }
                        }
```

PR #4389, opened 2025-10-02
fix(FR-1552): fix infinite re-rendering in useInterval hook

yomybaby:
# Refactor `useInterval` hook for better performance and readability

**Changes:**
- Fixed dependency array in the callback effect to prevent unnecessary re-renders
- Simplified page visibility handling logic
- Created an `effectiveDelay` variable to centralize the delay calculation logic
- Removed redundant conditional checks in the interval effect
- Improved code organization with clearer separation of concerns


nowgnuesLee (approved):
LGTM

PR #4640, opened 2025-11-14
feat(FR-1679): add auto refresh to container logs

agatha197:
resolves FR-1679

# Add auto-refresh feature to container logs with configurable interval

This PR enhances the container log modal with an auto-refresh feature that allows users to automatically refresh logs at configurable intervals. Users can toggle auto-refresh on/off and set their preferred refresh interval using a time picker.

The implementation includes:

- Enhanced `AutoRefreshSwitch` component to accept additional props and use a configurable interval
- Added auto-refresh controls to the container log modal with a time picker for interval selection
- Stored user preferences for auto-refresh settings in user settings
- Added internationalization support for "Auto refresh" text in all language files

image

**Checklist:**

- [x] Documentation


yomybaby (changes requested):
- It doesn't work correctly on the first activation. You can reproduce this by clearing the relevant `localStorage` keys and trying again.
- Instead of using a `TimePicker`, what do you think about accepting just a single number representing seconds?

yomybaby (approved):
LGTM

PR #4671, opened 2025-11-19
fix(FR-1698): fix session creation failure when agent is set to auto

### LINKED ISSUE #4670: fix session creation failure when agent is set to auto
JIRA Issue: FR-1698

nowgnuesLee:
resolves #4670 (FR-1698)

This PR fixes an issue in the `useStartSession` hook where the agent selection logic wasn't properly handling both array and non-array values. The fix uses `_.castArray()` to ensure the agent value is always treated as an array before comparing it to `['auto']`.

- [ ] Documentation


yomybaby (approved):
LGTM

PR #4724, opened 2025-11-28
feat(FR-1734): Add 'system' type to resource group setting modal

agatha197:
resolves FR-1734

# Add "System" option to resource group type selection

This PR adds a new "System" option to the resource group type dropdown in the ResourceGroupSettingModal component. Users can now select "System" as a resource group type alongside the existing options (Batch, Interactive, and Inference).

**Checklist:**

- [ ] Documentation


nowgnuesLee (approved):
LGTM

PR #4785, opened 2025-12-08
feat(FR-1720): implement theme color reset functionality

### LINKED ISSUE #4689: Add individual initialization feature for setting items
We are looking to implement a feature that allows for the individual initialization of setting items. This enhancement will enable users to reset specific settings to their default values without affecting other configurations. The goal is to improve user experience by providing more granular control over settings management.

JIRA Issue: FR-1720

ironAiken2:
Resolves #4689 ([FR-1720](https://lablup.atlassian.net/browse/FR-1720))

## Summary
- Add `resetThemeConfig` function to reset individual theme color tokens back to their default values
- Support functional updates in `useBAISettingUserState` to prevent race conditions during bulk resets
- Fix theme color reset functionality to properly restore default theme.json values

## Test plan
- [ ] Verify individual theme color reset works correctly for each color picker
- [ ] Verify "Reset All" button properly resets all theme colors at once
- [ ] Confirm reset values match the default theme.json configuration

[FR-1720]: https://lablup.atlassian.net/browse/FR-1720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react/src/components/SettingItem.tsx
+++ b/react/src/components/SettingItem.tsx
@@ -1,8 +1,20 @@
-import { Badge, Checkbox, Select, SelectProps, Typography, theme } from 'antd';
+import { SettingOutlined } from '@ant-design/icons';
+import { useToggle } from 'ahooks';
+import {
+  Alert,
+  Badge,
+  Checkbox,
+  CheckboxProps,
+  Dropdown,
+  Select,
+  Typography,
+  theme,
+} from 'antd';
 import { createStyles } from 'antd-style';
-import { BAIFlex } from 'backend.ai-ui';
+import { BAIButton, BAIFlex, BAIModal, BAISelectProps } from 'backend.ai-ui';
+import { t } from 'i18next';
 import _ from 'lodash';
-import React, { ReactElement, ReactNode } from 'react';
+import React, { ReactElement, ReactNode, useState } from 'react';
 
 export interface SettingItemProps {
   'data-testid'?: string;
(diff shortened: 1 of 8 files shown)


react/src/components/SettingItem.tsx
```diff
@@ -35,33 +49,83 @@ const SettingItem: React.FC<SettingItemProps> = ({
...
+            !checkboxProps?.disabled &&
+            showResetButton && (
+              <Dropdown
+                menu={{
+                  items: [
+                    {
+                      key: 'reset',
+                      label: t('button.Reset'),
+                      onClick: () => setIsOpenResetChangesModal(),
+                      style: { color: token.colorError },
```
nowgnuesLee: Let's use `danger` option.

react/src/components/SettingItem.tsx
```diff
@@ -102,6 +177,24 @@ const SettingItem: React.FC<SettingItemProps> = ({
...
+        }}
+        cancelText={t('button.Cancel')}
+        onCancel={() => setIsOpenResetChangesModal()}
+      >
+        <Alert
+          showIcon
+          message={t('dialog.warning.CannotBeUndone')}
+          type="warning"
+        />
+      </BAIModal>
```
nowgnuesLee, line 197: How about using `Modal.confirm`?? what do you think?

ironAiken2: When confirming a button that is always visible, using the button's action with `await modal.confirm` has the advantage of displaying a loading interface. However, in this case, I think there doesn't seem to be much difference between the two approaches.

PR #4817, opened 2025-12-11
fix(FR-1781): fixed a bug where the endpoint change in chat was not being applied correctly

### LINKED ISSUE #4818: Chat messages are not being displayed correctly
{width=482}

dogbowl

I think the chat system doesn’t work properly.

JIRA Issue: FR-1782

nowgnuesLee:
resolves #4818 (FR-1782)

In this PR, I resolved the issue where endpoint changes were not being applied correctly by adding an ID to useChat. A new transport will be created as the endpoint changes.

- [ ] Documentation

### DIFF
--- a/react/src/components/Chat/ChatCard.tsx
+++ b/react/src/components/Chat/ChatCard.tsx
@@ -193,6 +193,7 @@ const PureChatCard: React.FC<ChatCardProps> = ({
   ...otherCardProps
 }) => {
   'use memo';
+
   const { t } = useTranslation();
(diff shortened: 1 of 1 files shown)


yomybaby (changes requested):
If useChat's id does not change, it seems that even if you pass a new transport, a change does not occur. Considering that multiple ChatCards are displayed on one screen, it would be good to create an id.

yomybaby (approved):
LTGM

PR #4856, opened 2025-12-18
fix(FR-1801): add ellipsis to DoubleTag component

nowgnuesLee:
resolves #NNN (FR-1801)

This PR adds text truncation with ellipsis to the tag labels in the DoubleTag component. Long tag labels are now limited to a maximum width of 150px, with overflow text being truncated and replaced with an ellipsis. The full label text is still accessible via a tooltip when hovering over the truncated label.

**Checklist:**

- [ ] Documentation

### DIFF
--- a/react/src/components/DoubleTag.tsx
+++ b/react/src/components/DoubleTag.tsx
@@ -1,6 +1,6 @@
 import TextHighlighter from './TextHighlighter';
 import { Tag } from 'antd';
-import { BAIFlex } from 'backend.ai-ui';
+import { BAIFlex, BAIText } from 'backend.ai-ui';
 import _ from 'lodash';
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


react/src/components/DoubleTag.tsx
```diff
@@ -45,13 +45,24 @@ const DoubleTag: React.FC<{
...
           >
-            {!_.isUndefined(highlightKeyword) ? (
-              <TextHighlighter keyword={highlightKeyword}>
-                {objValue.label}
-              </TextHighlighter>
-            ) : (
-              objValue.label
-            )}
+            <div
+              // Since BAIText has own style for fontSize, I decided to not use BAIText here for ellipsis.
```
yomybaby: You can use `inherit` for `fontSize`. It would be better to use BAIText to manage global changes later.

PR #4861, opened 2025-12-19
fix(FR-1803): handle missing keypair information during login to clear session ID

### LINKED ISSUE #4860: when login fails, the previous session ID is retained
After check_login api, a GQL request for the keypair is sent, but there is a bug where the session ID is not cleared when that request fails

JIRA Issue: FR-1803

nowgnuesLee:
resolves #4860 (FR-1803)

This PR adds a check for missing keypair information during the login process. If the keypair data is missing from the response, the client will now automatically log out and throw an appropriate error message, preventing potential issues that could occur when trying to access undefined keypair properties.

- [ ] Documentation

### DIFF
--- a/src/components/backend-ai-login.ts
+++ b/src/components/backend-ai-login.ts
@@ -1765,6 +1765,11 @@ export default class BackendAILogin extends BackendAIPage {
       .then(async (response) => {
         this.is_connected = true;
         globalThis.backendaiclient = this.client;
+        if (!response['keypair']) {
+          // if no keypair info, logout and throw error
+          await this.client?.logout();
+          throw new Error('Keypair information is missing.');
+        }
         const resource_policy = response['keypair'].resource_policy;
         globalThis.backendaiclient.resource_policy = resource_policy;
         this.user = response['keypair'].user;


src/components/backend-ai-login.ts
```diff
@@ -1765,6 +1765,11 @@ export default class BackendAILogin extends BackendAIPage {
       .then(async (response) => {
         this.is_connected = true;
         globalThis.backendaiclient = this.client;
+        if (!response['keypair']) {
+          // if no keypair info, logout and throw error
+          await this.client?.logout();
+          throw new Error('Keypair information is missing.');
+        }
```
ironAiken2, line 1772: Is it possible for a login request to succeed even when there is no keypair information? How can I test this?

nowgnuesLee:
> If the reason for a login request failure is passed as data.detail, it can be displayed to the user as a message.  
>   
> https://github.com/lablup/backend.ai-webui/blob/b46ac74a5a8ac0da3a0c5a6bb26e3f96ffc7e734/src/components/backend-ai-login.ts#L1557

It’s not failing in that function. It fails inside `this._connectGQL()`, and it becomes an issue because there’s no logout logic in that code path. We’re planning to refactor the login flow itself when we move it over to React.

PR #4925, opened 2026-01-02
mics(FR-1833): Modified BAIAdminResourceGroupSelect to use scaling group name as value

### LINKED ISSUE #4900: add name fetching logic to BAIAdminResourceGroupSelect to display labels correctly
`BAIAdminResourceGroupSelect` cannot display the label when the resource ID set as its value is not included in the first page of options.

JIRA Issue: FR-1833

nowgnuesLee:
resolves #4900 (FR-1833)

Changed the value of resource group select options from `id` to `name` in the `BAIAdminResourceGroupSelect` component. This ensures that the resource group name is used as the value instead of its ID when selecting resource groups in the admin interface.

- [ ] Documentation


agatha197 (changes requested):
How can I test it? There's only storybook where uses `BAIAdminResourceGroupSelect` in webui. But it occurs errors. Please update PR description details.

ironAiken2 (changes requested):
Could you also add testing for fetching data via scrolling?   
Additionally, it appears a minimum width specification is required.

ironAiken2 (approved):
LGTM

PR #4966, opened 2026-01-07
feat(FR-1014): implement text file editor for NEO File Explorer

### LINKED ISSUE #3686: NEO File Explorer: Edit and save text files
As a user, I want to edit and save small text files, such as .txt and .yaml formats, so that I can make quick changes without needing a separate text editor.

JIRA Issue: FR-1014

limshyun:
Resolves #3686 ([FR-1014](https://lablup.atlassian.net/browse/FR-1014))

## Summary

Implement the ability to directly edit and save simple text files within the NEO File Explorer.

## Background

Users frequently need to modify configuration files such as yaml, toml, and json within vfolders for model serving. Currently, this requires downloading the file → editing locally → re-uploading, which results in a poor user experience.

### Key Features
- **Edit button in More menu**: Added dropdown menu with edit option in File Explorer controls
- **Text file detection**: Supports .txt, .md, .json, .yaml, .yml, .xml, .csv, .js, .ts, .jsx, .tsx, .py, .sh, .bash, .html, .css, .scss, .less, .sql, .log, .env, .conf, .config, .ini, .toml
- **CodeMirror editor**: Full-featured text editor with syntax highlighting
- **File operations**: Download → Edit → Upload workflow using existing APIs

### Components Added
- `TextFileEditorModal.tsx`: Main editor modal with BAICodeEditor
- Enhanced `FileItemControls.tsx`: Added More menu with Edit option
- Enhanced `BAIFileExplorer.tsx`: Added `enableEdit` and `onClickEditFile` props
- Updated `FolderExplorerModal.tsx`: Integration with TextFileEditorModal

### Technical Details
- **File Reading**: `request_download_token` → `fetch` → `blob.text()`
- **File Saving**: Modified content → `new File()` → `FileUploadManager.uploadFiles()` (overwrites)
- **Permissions**: Requires `write_content` permission
- **Size Limit**: `baiClient._config.maxFileUploadSize`

### i18n Support
- `comp:FileExplorer.EditFile`: "Edit"
- `data.explorer.EditFile`: "Edit File" 
- `data.explorer.FailedToLoadFile": "Failed to load file"

**Checklist:**
- [x] Documentation (PR description)
- [ ] Minimum required manager version (N/A - client-side feature)
- [ ] Specific setting for review: Test with various text files in File Explorer
- [x] Minimum requirements: `write_content` permission on vfolder
- [x] Test cases: Edit .txt, .yaml, .json files; verify save/cancel functionality

[FR-1014]: https://lablup.atlassian.net/browse/FR-1014

### test

https://github.com/user-attachments/assets/f067e671-999a-4f20-9f7f-1c13f1d92e00


packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -1,29 +1,35 @@
...
 const FileItemControls: React.FC<FileItemControlsProps> = ({
   selectedItem,
   onClickDelete,
+  onClickEdit,
   enableDownload = false,
   enableDelete = false,
+  enableEdit = false,
   downloadButtonProps,
   deleteButtonProps,
 }) => {
```
nowgnuesLee, line 36: <p class='graphite__hidden'><i>[Re: lines +35 to +36]</i></p>
Let's use 'use memo' here.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/4966?utm_source=unchanged-line-comment">Graphite</a>.</p>

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -108,6 +114,38 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
...
+              onClick: (e) => {
+                e.domEvent.stopPropagation();
+                onClickEdit?.();
+              },
+            },
+          ],
+        }}
+        trigger={['click']}
+      >
+        <Button
```
nowgnuesLee: Let's use BAIButton. Please change others too.

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -108,6 +114,38 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
...
+          type="text"
+          size="small"
+          onClick={(e) => {
+            e.preventDefault();
+            e.stopPropagation();
+          }}
+          icon={<MoreOutlined />}
+          aria-label={t('comp:FileExplorer.MoreOptions')}
```
nowgnuesLee, line 148: how about using Tooltip component or title option?

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -108,6 +114,38 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
         }}
         {...deleteButtonProps}
       />
+      <Dropdown
+        menu={{
+          items: [
+            {
+              key: 'fileEdit',
+              icon: <EditIcon size={14} />,
```
nowgnuesLee: Do we need to specify the size?

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      const errorMessage =
+        error instanceof Error ? error.message : 'Unknown error';
+      setLoadError(errorMessage);
+    } finally {
+      setIsSaving(false);
+    }
+  };
+
+  const handleCancel = () => {
+    onRequestClose(false);
```
nowgnuesLee: Since it didn’t fail, passing **`false`** doesn’t seem appropriate. Would it be better to make the **`success`** parameter optional?   
Please don’t create a new function—use it directly in **`onCancel`**

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      okText={t('button.Save')}
+      cancelText={t('button.Cancel')}
+    >
+      <BAIFlex direction="column" gap="md" style={{ minHeight: 400 }}>
+        {isLoading ? (
+          <Skeleton active paragraph={{ rows: 10 }} />
+        ) : loadError ? (
+          <Alert
+            type="error"
+            message={t('data.explorer.FailedToLoadFile')}
```
nowgnuesLee: message is deprecated. Please use title instead of message option.

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+
+const getLanguageFromExtension = (fileName: string): LanguageName => {
+  const extension = _.last(fileName.split('.'))?.toLowerCase() || '';
+  return extension || 'sh';
+};
+
+interface TextFileEditorModalProps extends Omit<BAIModalProps, 'children'> {
+  open: boolean;
```
nowgnuesLee: You don't need to open optoin.

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <BAIFlex direction="column" gap="md" style={{ minHeight: 400 }}>
+        {isLoading ? (
+          <Skeleton active paragraph={{ rows: 10 }} />
+        ) : loadError ? (
+          <Alert
+            type="error"
+            message={t('data.explorer.FailedToLoadFile')}
+            description={loadError}
+            showIcon
+          />
```
nowgnuesLee: Let's use BAIAlert and set width 100%.

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      const blob = new Blob([content], { type: 'text/plain' });
+      const file = new File([blob], fileInfo.name, {
+        type: 'text/plain',
+      }) as RcFile;
+
+      await uploadFiles([file], targetVFolderId, currentPath);
+      onRequestClose(true);
+    } catch (error) {
+      const errorMessage =
+        error instanceof Error ? error.message : 'Unknown error';
```
nowgnuesLee: Let's use useErrorMessageResolver

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+      setLoadError(errorMessage);
+    } finally {
+      setIsSaving(false);
+    }
+  };
```
nowgnuesLee: how about using useTanMutation??

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+        const text = await blob.text();
+        setContent(text);
+      } catch (error) {
+        const errorMessage =
+          error instanceof Error ? error.message : 'Unknown error';
+        setLoadError(errorMessage);
+      } finally {
+        setIsLoading(false);
+      }
+    };
```
nowgnuesLee: how about using useTanQuery instead of using useEffect??

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const handleCancel = () => {
+    onRequestClose(false);
+  };
+
+  return (
+    <BAIModal
+      open={open}
+      title={
+        <BAIFlex align="center" gap="xs">
+          <Typography.Text>{t('data.explorer.EditFile')}</Typography.Text>
```
nowgnuesLee: Let's use BAIText

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+  convertToDecimalUnit,
+  useConnectedBAIClient,
+} from 'backend.ai-ui';
+import _ from 'lodash';
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+
+const getLanguageFromExtension = (fileName: string): LanguageName => {
+  const extension = _.last(fileName.split('.'))?.toLowerCase() || '';
+  return extension || 'sh';
```
nowgnuesLee:
```suggestion
  return _.last(fileName.split('.'))?.toLowerCase() || 'sh';
```

nowgnuesLee (changes requested):

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
limshyun: I applied both Tooltip and aria-label. The Tooltip displays "More Options" text visually when hovering with the mouse, while aria-label is kept for screen reader users with visual impairments. Since these serve different purposes, it's better to use both together. Korean

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -1,32 +1,40 @@
...
   enableDelete = false,
+  enableEdit = false,
   downloadButtonProps,
   deleteButtonProps,
 }) => {
+  'use memo';
+
   const { t } = useTranslation();
   const { token } = theme.useToken();
   const { message } = App.useApp();
```
ironAiken2, line 41: <p class='graphite__hidden'><i>[Re: lines +44 to +45]</i></p>
Although not included in this PR, we are using `useTanMutation`—which wraps Tanstack's `useMutation`—to differentiate it from `useMutation`, as relay also supports `useMutation`.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/4966?utm_source=unchanged-line-comment">Graphite</a>.</p>

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -85,7 +93,7 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
 
   return (
     <BAIFlex gap="xs">
-      <Button
+      <BAIButton
         type="text"
         size="small"
         icon={<DownloadIcon color={token.colorInfo} />}
```
ironAiken2, line 111: If you use BAIButton, you can control the button's loading state through the action.   
  
To control loading with an action, ensure asynchronous processing works correctly within the action's await function. When mutating with Tanstack, use mutateAsync instead of mutate to ensure the promise resolves within the then block.

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -108,6 +116,40 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
         }}
         {...deleteButtonProps}
       />
+      <Tooltip title={t('comp:FileExplorer.MoreOptions')}>
```
ironAiken2: Since the MoreOutlined icon is typically used to provide additional options, it probably isn't necessary to provide further guidance via a tooltip. Instead, how about providing guidance for files that cannot be edited? For example, “This file format does not support editing.”

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -108,6 +116,40 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
...
+              },
+            ],
+          }}
+          trigger={['click']}
+        >
+          <BAIButton
+            type="text"
+            size="small"
+            onClick={(e) => {
+              e.preventDefault();
```
ironAiken2: It doesn't seem necessary.

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+  onRequestClose,
+  uploadFiles,
+}) => {
+  const { t } = useTranslation();
+  const { message } = App.useApp();
+  const baiClient = useConnectedBAIClient();
+  const { getErrorMessage } = useErrorMessageResolver();
+
+  const [content, setContent] = useState<string>('');
+  const [isEdited, setIsEdited] = useState(false);
```
ironAiken2: If you use `fileContent` as the initial value for `content`, you don't need to use the `isEdited` state.

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+        {isLoading ? (
+          <Skeleton active paragraph={{ rows: 10 }} />
+        ) : loadError ? (
+          <BAIAlert
+            type="error"
+            title={t('data.explorer.FailedToLoadFile')}
+            description={getErrorMessage(loadError as Error)}
+            showIcon
+            style={{ width: '100%' }}
+          />
```
ironAiken2, line 167: The error message doesn't seem user-friendly. It would be better to inform the user that an error occurred and explain the next steps for resolution. For example: 'The file contents cannot be verified. Please confirm that this is the correct file.'   
  
Additionally, it seems unnecessary to expose error codes to users. Using BAILogger appears to be the better approach.

ironAiken2 (changes requested):
When I edit one file, the editor shows the changes made to other files. And I think it would be good to include the text file editing feature in the BUI. What do you think?

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
limshyun: The disabled state handles non-editable files, but users might not understand why the option is grayed out. Would it be helpful to add a tooltip explaining "This file format does not support editing" or "Only text files can be edited"?

agatha197 (changes requested):
https://github.com/user-attachments/assets/a4087506-a2ed-4c22-9ad3-a59c1b8b7269



When I edit the text file, the file browser is closed and frozen. I cannot re-open the file browser before refreshing. I made the text file by this way.
1. In the terminal, `touch test.txt` 
2. Add some contents
3. Upload that file

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -108,6 +110,46 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
...
+        <BAIButton
+          type="text"
+          size="small"
+          onClick={(e) => {
+            e.stopPropagation();
+          }}
+          icon={<MoreOutlined />}
+          aria-label={t('comp:FileExplorer.MoreOptions')}
+          style={{ color: token.colorTextSecondary }}
+        />
```
nowgnuesLee, line 150: Since you said you applied tooltip here, but I can't see anything like Tooltip component.

react/src/components/TextFileEditorModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+        type: 'text/plain',
+      }) as RcFile;
+
+      await uploadFiles([file], targetVFolderId, currentPath);
+    },
+    onSuccess: () => {
+      onRequestClose(true);
+    },
+    onError: (error) => {
+      message.error(getErrorMessage(error as Error));
```
nowgnuesLee: error is already type of Error

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
limshyun: I initially applied both Tooltip and aria-label together, where Tooltip shows "More Options" text visually on hover while aria-label serves screen reader users.

However, after consideration, I decided this provides too much information. Since MoreOutlined is commonly understood as an icon for additional options, adding a "More" tooltip feels redundant and visually cluttered.

Instead, I'll only add a tooltip for unsupported files to show "Unsupported file type" - this provides meaningful information where it's actually needed.

Would it be acceptable to apply it like https://github.com/lablup/backend.ai-webui/pull/4966#discussion_r2675124748

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,161 @@
...
+    },
+  });
+
+  const handleContentChange = (value: string | undefined) => {
+    setContent(value ?? '');
+  };
+
+  return (
+    <BAIModal
+      open={open}
```
nowgnuesLee:
```suggestion
      {...modalProps}
```

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,164 @@
...
```
nowgnuesLee: we don't need it.

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+        </BAIFlex>
+      }
+      width={'100%'}
+      destroyOnHidden
+      onCancel={() => onRequestClose()}
+      onOk={() => saveMutation.mutate()}
+      confirmLoading={saveMutation.isPending}
+      okText={t('button.Save')}
+      cancelText={t('button.Cancel')}
+      {...modalProps}
```
nowgnuesLee: **If you place it here, the parent can override** **`title`,** **`onOk`,** **`onCancel`, etc. So please change the order and add the options that shouldn’t be overridden to** **`Omit`.**

react/src/components/FolderExplorerModal.tsx
```diff
@@ -297,6 +308,23 @@ const FolderExplorerModal: React.FC<FolderExplorerProps> = ({
           </BAIFlex>
         )}
       </Suspense>
+      <Suspense fallback={null}>
```
nowgnuesLee: Do we need to put Suspense here?

packages/backend.ai-ui/src/helper/index.ts
```diff
@@ -426,3 +426,46 @@ export const isValidUUID = (uuid: string) => {
...
+  css: 'css',
+  scss: 'sass',
+  less: 'less',
+  sql: 'sql',
+  log: 'shell',
+  env: 'shell',
+  conf: 'shell',
+  config: 'shell',
+  ini: 'toml',
+  toml: 'toml',
```
agatha197: Please check again with official codes. Some extension doesn't support highlighting properly.

agatha197 (changes requested):

limshyun:
> I cannot modify the txt file still.

This was addressed in commit 7cd637e67. You can verify the supported language extensions at https://github.com/uiwjs/react-codemirror/blob/master/extensions/langs/src/index.ts

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,168 @@
...
+                style={{ width: '100%' }}
+              />
+            );
+          })()
+        ) : (
+          <BAICodeEditor
+            value={content ?? fileContent ?? ''}
+            onChange={handleContentChange}
+            editable={true}
+            language={getCodeMirrorLanguage(fileInfo?.name || '')}
```
nowgnuesLee: I got type error here.

nowgnuesLee (changes requested):

ironAiken2 (changes requested):
Please enable the `fixedOverflowWidgets` option to allow the popup to render outside the modal area.

packages/backend.ai-ui/src/components/baiClient/FileExplorer/FileItemControls.tsx
```diff
@@ -108,6 +111,46 @@ const FileItemControls: React.FC<FileItemControlsProps> = ({
...
+                  t('comp:FileExplorer.EditFile')
+                ),
+                disabled: isEditDisabled,
+                onClick: (e) => {
+                  e.domEvent.stopPropagation();
+                  onClickEdit?.();
+                },
+              },
+            ];
+          })(),
```
ironAiken2: Making this part a function might reduce readability when adding other items. Please consider declaring it as an array and using `filterOutEmpty`.

react/src/components/FolderExplorerModal.tsx
```diff
@@ -64,6 +68,20 @@ const FolderExplorerModal: React.FC<FolderExplorerProps> = ({
...
+    loader.init().then((monacoInstance) => {
+      if (isMounted) {
+        setMonaco(monacoInstance);
+      }
+    });
+    return () => {
+      isMounted = false;
+    };
+  }, []);
+
```
ironAiken2: Whether Monaco is mounted or available does not seem to be something this component should consider. It appears preferable to determine Monaco's availability within VFolderTextFileEditorModal and provide guidance via an alert or similar method.

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+    },
+    enabled: !!modalProps.open && !!fileInfo && !!targetVFolderId,
+  });
+
+  const saveMutation = useTanMutation({
+    mutationFn: async () => {
+      if (!fileInfo) return;
+
+      const contentToSave = editorRef.current?.getValue() ?? fileContent ?? '';
+      const blob = new Blob([contentToSave], { type: 'text/plain' });
```
ironAiken2: All file types supported by Monaco are allowed. Is it okay to save it as text/plain? Is it necessary to match the file extension?

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+    },
+    onError: (error) => {
+      message.error(getErrorMessage(error));
+    },
+  });
+
+  return (
+    <BAIModal
+      width={'100%'}
+      destroyOnHidden
```
ironAiken2, line 135: The Monaco Editor may require the use of BAIUnmountAfterClose because it loads all settings when the component is mounted.

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+        type: 'text/plain',
+      }) as RcFile;
+
+      await uploadFiles([file], targetVFolderId, currentPath);
+    },
+    onSuccess: () => {
+      queryClient.removeQueries({
+        queryKey: ['textFileContent', targetVFolderId, filePath],
+      });
+      onRequestClose(true);
```
ironAiken2: Since the modal fetches new content when it opens, it seems unnecessary.

ironAiken2 (changes requested):

react/src/components/VFolderTextFileEditorModal.tsx
limshyun: `BAIUnmountAfterClose` is already being used in `FolderExplorerModal.tsx` where `VFolderTextFileEditorModal` is rendered:

```tsx
<BAIUnmountAfterClose>
  <VFolderTextFileEditorModal
    open={!!editingFile}
    ...
  />
</BAIUnmountAfterClose>
```

react/src/components/VFolderTextFileEditorModal.tsx
limshyun: Without removeQueries, when a user saves and reopens the same file, React Query returns the cached (old) content instead of fetching the updated content from the server.

https://github.com/user-attachments/assets/dfb10a02-d6eb-4a2c-8627-63ca197417b4

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+      const contentToSave = editorRef.current?.getValue() ?? fileContent ?? '';
+      const blob = new Blob([contentToSave], { type: 'text/plain' });
+      const file = new File([blob], fileInfo.name, {
+        type: 'text/plain',
+      }) as RcFile;
+
+      await uploadFiles([file], targetVFolderId, currentPath);
+    },
+    onSuccess: () => {
+      queryClient.removeQueries({
```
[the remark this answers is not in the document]
agatha197: `removeQueries` completely removes the cache from memory, while `invalidateQueries` marks it as stale and triggers a background refetch. This is the standard React Query pattern used elsewhere in the codebase.



**Benefits:**

- Better UX: When reopening the same file, users see the previous content instantly while it refetches in the background
- Standard pattern: Matches the approach in `reactQueryAlias.tsx` and other mutations
- Clearer intent: Signals "this data needs updating" rather than "delete this data"

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,205 @@
...
+    error: loadError,
+    isLoading,
+  } = useTanQuery({
+    queryKey: ['textFileContent', targetVFolderId, filePath],
+    queryFn: async () => {
+      const tokenResponse = await baiClient.vfolder.request_download_token(
+        filePath!,
+        targetVFolderId,
+        false,
+      );
```
agatha197, line 90: If file size is too large, the browser can be frozen. Can you check file size first?

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,205 @@
...
+            type="error"
+            showIcon
+            title={t('data.explorer.FailedToLoadFile')}
+            description={t('data.explorer.FailedToLoadFileDescription')}
+          />
+        ) : isLoading ? (
+          <Skeleton active />
+        ) : (
+          <MonacoEditor
+            defaultValue={fileContent ?? ''}
```
agatha197, line 172: The reason for [the issue](https://app.graphite.com/github/pr/lablup/backend.ai-webui/4966/feat()-implement-text-file-editor-for-NEO-File-Explorer#review-PRR_kwDOCRTcws7aFyhB) might be that you used **`defaultValue`** instead of **`value`**. Can you check it? (I didn't test it myself, but the AI suggested that. Can you verify it?)

agatha197 (changes requested):

react/src/components/VFolderTextFileEditorModal.tsx
limshyun: I initially implemented it using `value` and `onChange`, but encountered an issue where the cursor position would jump to the end of the text after a short delay following the initial render.

This happens because `@monaco-editor/react`'s `value` prop internally calls `editor.setValue(value)` when the value is synchronized after the editor mounts. Combined with React Query's data fetching timing, when data loads and `value` updates while the editor is already mounted, `setValue()` is called which replaces the entire content and moves the cursor to the end.

Using `defaultValue` resolves this because it only sets the value during editor initialization, and the editor manages its own state afterwards.

react/src/components/VFolderTextFileEditorModal.tsx
limshyun: I want to add error handling when the file size is too large. What size limit would be appropriate?

The upload logic has a 4GB limit, but for file editing, should we set it to around 10MB considering browser memory and Monaco Editor rendering performance?

react/src/components/VFolderTextFileEditorModal.tsx
ironAiken2: Using react-query's staleTime or gcTime options along with isFetching allows you to display newly loaded data without removing the query. isLoading and isFetching behave differently.

react/src/components/VFolderTextFileEditorModal.tsx
```diff
@@ -0,0 +1,205 @@
+'use memo';
```
ironAiken2: The `useMemo` hook must be declared inside the function. Please review the components again. If any parts are missing, add them.

react/src/components/VFolderTextFileEditorModal.tsx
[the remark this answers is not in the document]
limshyun: The issue has been resolved in commit 
a754ee7

react/src/components/VFolderTextFileEditorModal.tsx
agatha197: Please use `maxFileUploadSize` of config.toml.

PR #5055, opened 2026-01-20
feat(FR-1859): Create a BAIUserSelect

### LINKED ISSUE #4931: Create BAIUserSelect
JIRA Issue: FR-1859

nowgnuesLee:
resolves #4931 (FR-1859)

This PR improves the Relay Infinite Scroll Select skill documentation and implementation by:

1. Introducing a clearer decision tree based on dynamic query parameter needs
2. Adding a new Pattern B (Dynamic) that provides full control over query parameters
3. Implementing `BAIUserSelect` component as a reference for email-based selection with dynamic parameters
4. Adding dynamic `first` parameter to ensure all selected values are fetched efficiently
5. Updating `BAIVFolderSelect` to use dynamic `first` parameter for better data completeness
6. Providing detailed comparison between Simple and Dynamic patterns
7. Adding comprehensive examples and best practices for both patterns

The key innovation is the dynamic `first` parameter that calculates exactly how many items to fetch based on selection count, preventing both over-fetching and under-fetching of data.

**Checklist:**

- [x] Documentation

### DIFF
--- a/.claude/skills/relay-infinite-scroll-select/SKILL.md
+++ b/.claude/skills/relay-infinite-scroll-select/SKILL.md
@@ -19,42 +19,54 @@ description: >
 ## Quick Start Decision Tree
 
 ```
-START: What type of value does your select use?
+START: Does your select need dynamic query parameters?
 
 ┌─────────────────────────────────────────────────────────────┐
-│ Q: Is the select value the same as the display label?      │
-│    (e.g., entity's name field is both value and label)     │
+│ Q: Do you need dynamic control over query parameters?      │
+│    (filter, limit, first, order, etc.)                     │
+│    OR need external refetch capability?                    │
 └─────────────────────────────────────────────────────────────┘
                          │
           ┌──────────────┴──────────────┐
           │                             │
-         YES                           NO
+         NO                            YES
           │                             │
           ▼                             ▼
    ┌─────────────┐             ┌─────────────┐
    │  Pattern A  │             │  Pattern B  │
-   │  (Simple)   │             │  (Complex)  │
+   │  (Simple)   │             │ (Dynamic)   │
    └─────────────┘             └─────────────┘
           │                             │
           ▼                             ▼
-Reference:                    Reference:
-BAIAdminResourceGroupSelect   BAIVFolderSelect
+Reference:                    References:
+BAIAdminResourceGroupSelect   BAIUserSelect (email-based)
+                              BAIVFolderSelect (id-based)
 ```
 
 ## Pattern Comparison
 
-| Criteria | Pattern A (Name-Based) | Pattern B (ID-Based) |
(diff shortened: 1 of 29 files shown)


ironAiken2 (changes requested):
It appears network request optimization is needed, such as applying debounce during input for search. There also seems to be no need to use two queries internally.

packages/backend.ai-ui/src/components/fragments/BAIUserSelect.tsx
yomybaby, line 1: When a defaultValue is specified for `BAIUserSelect` and the browser is refreshed, the first rendering occurs with the select popup open. Since defaultValue is not being utilized, it is not a critical issue, but please resolve it if it can be fixed simply.

yomybaby, line 1: Can the value not be used as a user id (or local id) instead of an email?

yomybaby, line 1: If you are going to handle the value with email and ID, it would be better to implement it so that the value TS type of this component's onChange is a `string`.

yomybaby (changes requested):
a

.claude/skills/relay-infinite-scroll-select/SKILL.md
```diff
@@ -258,6 +383,71 @@ const [open, setOpen] = useControllableValue(props, {
...
+- ✅ Dynamic pagination size
+- ✅ Custom sort order
+
+**Example: Name-based but needs Pattern B**
+```typescript
+// Even though value is 'name', we need Pattern B for dynamic features
+<YourEntitySelect
+  value={selectedNames}           // Name-based value
+  filter={externalFilter}         // 🔑 Dynamic filter
+  pageSize={20}                   // 🔑 Dynamic limit
```
agatha197, line 607: It saids dynamic limit, but static value is used.

.claude/skills/relay-infinite-scroll-select/SKILL.md
```diff
@@ -258,6 +383,71 @@ const [open, setOpen] = useControllableValue(props, {
...
+  sortBy="name"                   // 🔑 Dynamic order
```
agatha197, line 608: this too.

packages/backend.ai-ui/src/components/fragments/BAIUserSelect.tsx
agatha197, line 1:
```
<BAIUserSelect showSearch={false} />
```

`showSearch` doesn't work like upper. It is always possible to search.

ironAiken2 (approved):
LGTM

yomybaby (approved):

PR #5104, opened 2026-01-23
misc(FR-1940): Modify the upload dropdown button to open only when clicked

### LINKED ISSUE #5089: Modify the upload dropdown button to open only when clicked
### Description

In the current Folder Explorer, the **Upload** button’s dropdown is displayed on **hover**. However, most users tend to **click** instead of hover, and there is a bug where the dropdown **closes immediately when clicked**.

To improve usability and avoid the click-to-close issue, we plan to change the behavior so that the dropdown is shown **only on click** (not on hover).

JIRA Issue: FR-1940

nowgnuesLee:
resolves #5089 (FR-1940)

Added `trigger={['click']}` to the Dropdown component in ExplorerActionControls to ensure the dropdown menu only opens on click events rather than hover.

**Checklist:**

- [ ] Documentation


yomybaby (approved):
LGTM

PR #5106, opened 2026-01-23
feat(FR-1933): Implement domain tab content in fair share page

### LINKED ISSUE #5073: Implement Domain Tab in Fair Share Page
We need to develop a Domain tab that allows users to verify and modify the fair share allocations for each domain. This feature will enhance our ability to manage domain resources effectively and ensure equitable distribution.

JIRA Issue: FR-1933

ironAiken2:
resolves #5073 ([FR-1933](https://lablup.atlassian.net/browse/FR-1933))

> test server: 10.122.10.253

### Add Domain Fair Share Table to Fair Share UI

This PR adds a new Domain Fair Share Table component to the Fair Share UI, allowing users to view and manage fair share weights at the domain level. The implementation includes:

- New `DomainFairShareTable` component that displays domain fair share data including weights, fair share factors, and usage metrics
- `FairShareWeightSettingModal` for configuring domain weights individually or in bulk
- Integration with the existing Fair Share navigation flow
- Support for sorting, filtering, and pagination of domain data

The domain table appears after selecting a resource group and shows the configured weights and cumulative usage during the aggregation period.

**Checklist:**

- [x] Documentation

[FR-1933]: https://lablup.atlassian.net/browse/FR-1933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


react/src/components/FairShareItems/DomainFairShareTable.tsx
nowgnuesLee, line 1: Is there any reason why we didn’t create `BAIDomainFairShareNodes`?

react/src/components/FairShareItems/DomainFairShareTable.tsx
```diff
@@ -0,0 +1,417 @@
...
+    setTablePaginationOption,
+  } = useBAIPaginationOptionStateOnSearchParam({
+    current: 1,
+    pageSize: 10,
+  });
+
+  const [queryParams, setQueryParams] = useQueryStates(
+    {
+      order: parseAsStringLiteral(availableDomainFairShareSorterValues),
+      filter: parseAsString,
```
nowgnuesLee: how about using parseAsJSON?

resources/i18n/ko.json
```diff
@@ -738,20 +738,25 @@
...
     "DefaultWeightDescription": "가중치가 지정되지 않은 도메인, 프로젝트, 사용자에 적용되는 가중치 기본값입니다.",
     "Domain": "도메인",
+    "DomainDescription": "도메인 에서는 도메인에 설정된 가중치와 집계 기간동안의 누적 사용량을 확인할 수 있습니다. <br /> 도메인 가중치는 해당 도메인에 속한 사용자의 우선순위 계수 계산에 사용됩니다. 서로 다른 도메인에 속한 사용자들은 서로의 우선순위에 영향을 주지 않습니다.",
+    "FairShareFactor": "우선 순위 계수",
+    "FairShareFactorDescription": "Fair Share Scheduler에 의해 계산된 작업 실행 우선순위입니다. 값이 높을수록 높은 우선순위를 갖습니다.",
     "FairShareSettingDescription": "변경사항은 Fair Share 계산에 바로 반영되지 않으며, 계산 주기에 따라 약 5분정도 소요될 수 있습니다.",
     "FairShareSettingTitleWithName": "Fair Share 설정: {{name}}",
-    "FairShareSettingsSuccessfullyUpdated": "{{resourceGroupName}} 자원 그룹의 fair share 설정이 수정되었습니다.",
+    "FairShareSettingsSuccessfullyUpdated": "Fair share 설정이 수정되었습니다.",
     "GoToSubComponent": "하위 {{sub}} 확인",
```
nowgnuesLee, line 761: The word “확인” feels a bit awkward here. How about using “이동” or “선” instead?

react/src/components/FairShareItems/DomainFairShareTable.tsx
```diff
@@ -0,0 +1,417 @@
...
+          {t('fairShare.FairShareSettingTitleWithName', {
+            name: t('fairShare.Domain'),
+          })}
+        </Typography.Title>
+        <Typography.Text type="secondary">
+          {<Trans i18nKey={t('fairShare.DomainDescription')} />}
+        </Typography.Text>
+      </BAIFlex>
+
+      <BAIFlex justify="between" align="center">
```
nowgnuesLee:
```suggestion
      <BAIFlex justify="between" align="center" wrap="wrap" gap="sm">
```

react/src/components/FairShareItems/FairShareWeightSettingModal.tsx
```diff
@@ -0,0 +1,244 @@
...
+                  domainName: name,
+                  weight: response?.weight,
+                })),
+              },
+            },
+            onCompleted: (res, errors) => {
+              if (!res?.bulkUpsertDomainFairShareWeight) {
+                message.error(t('dialog.ErrorOccurred'));
+                return;
+              }
```
nowgnuesLee: As far as I remember, `res` becomes `null`, so it always hits that conditional. Because of that, even if an error occurs, it doesn’t reach the next condition. Could you please check?

nowgnuesLee (changes requested):

react/src/components/FairShareItems/FairShareWeightSettingModal.tsx
ironAiken2: No, the response comes in as follows: {“data”:{‘bulkUpsertDomainFairShareWeight’:{“upsertedCount”:1}}} If there is no response, it is indeed an error.

resources/i18n/ko.json
ironAiken2: It looked a bit strange, and since it was checking the table, so I did it that way.   
  
하위 도메인 (으)로 이동 // 하위 프로젝트(으)로 이동

react/src/components/FairShareItems/DomainFairShareTable.tsx
ironAiken2: I did not consider the fair share scheduler configuration to be universally applicable. If necessary, I will handle it as a separate stack.

react/src/components/FairShareItems/DomainFairShareTable.tsx
```diff
@@ -0,0 +1,274 @@
...
+        createdAt
+        updatedAt
+
+        ...FairShareWeightSettingModal_DomainFragment
+      }
+    `,
+    domainFairShareNodeFragment,
+  );
+
+  const columns: ColumnsType<DomainFairShare> = [
```
nowgnuesLee, line 109: Let's use BAIColumnsType

react/src/components/FairShareItems/DomainFairShareTable.tsx
```diff
@@ -0,0 +1,274 @@
...
+          selectedRowKeys: _.map(selectedRows, 'domainName'),
+        }}
+        order={queryParams.order}
+        onChangeOrder={(order) => {
+          setQueryParams({
+            order:
+              (order as (typeof availableDomainFairShareSorterValues)[number]) ||
+              null,
+          });
+        }}
```
nowgnuesLee, line 246: It doesn’t seem necessary to configure this option here. I think the correct approach would be to have the FairShareList component configure it.

react/src/components/FairShareItems/FairShareList.tsx
```diff
@@ -50,24 +85,71 @@ const FairShareList: React.FC = () => {
...
+          : 'user';
+
   const [queryParams, setQueryParams] = useQueryStates(
     {
-      order: parseAsStringLiteral(availableResourceGroupSorterValues),
+      order: parseAsStringLiteral([
+        ...availableResourceGroupSorterValues,
+        ...availableDomainFairShareSorterValues,
+      ]),
       filter: parseAsString,
```
nowgnuesLee, line 114: how about using parseAsJSON instead?

react/src/components/FairShareItems/FairShareList.tsx
```diff
@@ -208,16 +335,35 @@ const FairShareList: React.FC = () => {
...
-            }}
-          />
+          <BAIFlex gap="sm">
+            {selectedRows?.length > 0 && (
+              <>
+                {t('general.NSelected', {
+                  count: selectedRows.length,
+                })}
+                <Tooltip title={t('general.BulkEdit')} placement="topLeft">
+                  <Button
```
nowgnuesLee, line 345: Please use BAIButton instead.

react/src/components/FairShareItems/FairShareList.tsx
```diff
@@ -208,16 +335,35 @@ const FairShareList: React.FC = () => {
...
+              </>
+            )}
+            <BAIFetchKeyButton
+              autoUpdateDelay={7_000}
+              loading={isPendingRefetch}
+              value=""
+              onChange={() => {
+                startRefetchTransition(() => {
+                  updateFetchKey();
+                });
```
nowgnuesLee, line 363: Since the fetchKey is already managed with `deferredValue`, is there any reason to use a transition as well?

react/src/components/FairShareItems/FairShareList.tsx
```diff
@@ -252,6 +403,59 @@ const FairShareList: React.FC = () => {
             />
           </Suspense>
         )}
+        {currentStep === 'domain' && (
+          <Suspense fallback={<Skeleton active />}>
```
nowgnuesLee, line 407: Suspense doesn’t seem necessary here, does it?

react/src/components/FairShareItems/FairShareWeightSettingModal.tsx
```diff
@@ -0,0 +1,237 @@
...
+        name: t('fairShare.Domain'),
+      })}
+      onCancel={() => onRequestClose?.(false)}
+      okButtonProps={{
+        loading:
+          isInflightCommitModifyDomainWeight ||
+          isInflightBulkModifyDomainWeight,
+      }}
+      onOk={handleOk}
+      {...modalProps}
```
nowgnuesLee, line 178: To use `modalProps` here, it seems we’ll need to omit some values from the type or change the order.

react/src/components/FairShareItems/DomainFairShareTable.tsx
ironAiken2: It modified in PR 5139.

PR #5156, opened 2026-01-29
fix(FR-1970): apply an optimistic update to the BAIVFolderSelect

### LINKED ISSUE #5142: apply an optimistic update to the BAIVFolderSelect
### Description

In `BAIVFolderSelect`, updating the `search` value currently feels laggy because it is not handled with an optimistic update. This causes noticeable UI stuttering during typing and degrades the overall user experience.

To improve UX, we will apply an optimistic update to the `search` value using React’s `useOptimistic`, so the UI updates immediately while the underlying state/data flow is processed.

JIRA Issue: FR-1970

nowgnuesLee:
resolves #5142 (FR-1970)

This PR addresses two issues:

1. Fixes the `showSearch` behavior in the `BAISelect` component:
   - Previously, when `showSearch` was a simple `true` value, it was not applied correctly because the component only handled the object form of `showSearch`
   - Now `showSearch` works properly for both the boolean (`true/false`) and object configurations

2. Improves search experience in `BAIVFolderSelect`:
   - Implements optimistic UI updates for search using React's `useOptimistic` hook
   - Updates the search value immediately in the UI before the actual search action completes
   - Maintains the actual search state separately from the optimistic display value

These changes provide a more responsive user experience when searching in folder selection dropdowns.

- [ ] Documentation

### DIFF
--- a/packages/backend.ai-ui/src/components/BAISelect.tsx
+++ b/packages/backend.ai-ui/src/components/BAISelect.tsx
@@ -180,9 +180,12 @@ function BAISelect<
         {...selectProps}
         loading={isPending || selectProps.loading}
         showSearch={
-          selectProps.showSearch && _.isObject(selectProps.showSearch)
-            ? {
-                ...selectProps.showSearch,
+          selectProps.showSearch === false
+            ? false
+            : {
+                ...(_.isObject(selectProps.showSearch)
+                  ? selectProps.showSearch
+                  : {}),
                 onSearch: async (value) => {
                   _.get(selectProps.showSearch, 'onSearch')?.(value);
                   startTransition(async () => {
@@ -190,7 +193,6 @@ function BAISelect<
                   });
                 },
               }
-            : false
         }
         ref={ref}
         className={classNames(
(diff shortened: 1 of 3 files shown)


packages/backend.ai-ui/src/components/fragments/BAIVFolderSelect.tsx
```diff
@@ -70,6 +71,12 @@ const BAIVFolderSelect: React.FC<BAIVFolderSelectProps> = ({
   );
   const deferredOpen = useDeferredValue(controllableOpen);
   const [searchStr, setSearchStr] = useState<string>();
+  const [optimisticSearchStr, setOptimisticSearchStr] = useOptimistic(
+    searchStr,
+    (_s, action: string) => {
+      return action;
+    },
```
yomybaby: You can remove sencond argument in this case.

```
[optimisticSearchStr, setOptimisticSearchStr] = useOptimistic(
    searchStr);
```

packages/backend.ai-ui/src/components/BAISelect.tsx
agatha197, line 1: `showSearch={false}` doesn't work.

packages/backend.ai-ui/src/components/fragments/BAIVFolderSelect.tsx
```diff
@@ -240,21 +248,27 @@ const BAIVFolderSelect: React.FC<BAIVFolderSelectProps> = ({
...
+      showSearch={
+        selectProps.showSearch === false
+          ? false
+          : {
+              searchValue: optimisticSearchStr,
+              autoClearSearchValue: true,
+              filterOption: false,
+              ...(_.isObject(selectProps.showSearch)
+                ? _.omit(selectProps.showSearch, ['searchValue'])
+                : {}),
```
ironAiken2, line 268: To properly use searchAction during searches, it seems filterOption must always be false. We should either move the expansion operation to the top or modify it so that filterOption is also omitted.

PR #5185, opened 2026-01-30
refactor(FR-2001): remove server-side duplicate name validation from session and service name hooks

### LINKED ISSUE #5184: Remove server-side duplicate name validation from session and service name hooks
Simplify client-side validation by removing GraphQL queries that check for duplicate session and service names, as the server already handles duplicate detection during creation requests.

## Why This Work is Needed

The current implementation performs duplicate name checks via GraphQL queries in useValidateSessionName and useValidateServiceName hooks. This creates validation logic inconsistencies between client and server, and the debounce delays are unnecessary since the server provides authoritative duplicate detection during creation.

## Expected Outcomes

- Validation hooks only perform client-side format validation (length, characters, patterns)
- Removed unnecessary debounce delays on form validation
- Server handles all duplicate name detection through creation error responses
- Simplified code using React Compiler 'use memo' directive instead of manual useMemo

JIRA Issue: FR-2001

yomybaby:
Resolves #5184 ([FR-2001](https://lablup.atlassian.net/browse/FR-2001))

Simplify client-side validation by removing GraphQL queries that check for duplicate session and service names, as the server already handles duplicate detection during creation requests.

## Why This Work is Needed

The current implementation performs duplicate name checks via GraphQL queries in useValidateSessionName and useValidateServiceName hooks. This creates validation logic inconsistencies between client and server, and the debounce delays are unnecessary since the server provides authoritative duplicate detection during creation.

## Changes

- Remove server-side validation queries from both hooks
- Remove validateDebounce from form items using these hooks  
- Simplify hooks using React Compiler 'use memo' directive
- Keep only client-side format validation (length, characters, patterns)

[FR-2001]: https://lablup.atlassian.net/browse/FR-2001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react/src/components/ComputeSessionNodeItems/EditableSessionName.tsx
+++ b/react/src/components/ComputeSessionNodeItems/EditableSessionName.tsx
@@ -181,11 +181,7 @@ const EditableSessionName: React.FC<EditableSessionNameProps> = ({
             flex: 1,
           }}
         >
-          <Form.Item
-            name="sessionName"
-            rules={validationRules}
-            validateDebounce={200}
-          >
+          <Form.Item name="sessionName" rules={validationRules}>
             <Input
               size="large"
               value={optimisticName || ''}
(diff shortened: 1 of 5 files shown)


nowgnuesLee (approved):
LGTM but please resolves the conflicts

PR #5233, opened 2026-02-04
style(FR-667): apply word-break keep-all for CJK languages

limshyun:
Resolves [FR-667](https://lablup.atlassian.net/browse/FR-667)

## Changes
- Set `document.documentElement.lang` on language initialization and change events
- Add CSS rule `:lang(ko|ja|zh) * { word-break: keep-all; }` to apply proper word breaking for CJK languages

## Implementation
1. **DefaultProviders.tsx**: Set `lang` attribute on `<html>` element
   - On mount: Initialize with current language
   - On change: Update when 'langChanged' event fires
2. **index.css**: Apply `word-break: keep-all` using CSS `:lang()` selector
   - Targets Korean (ko), Japanese (ja), and Chinese (zh) languages
   - Prevents awkward mid-word line breaks in CJK text

## Testing
- [x] Verify `<html lang="ko">` is set when using Korean
- [x] Verify text breaks at word boundaries (spaces) in Korean
- [x] Verify same behavior for Japanese and Chinese
- [x] Verify English and other languages are not affected

[FR-667]: https://lablup.atlassian.net/browse/FR-667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react/src/index.css
+++ b/react/src/index.css
@@ -4,3 +4,9 @@
   height: 1em !important;
   vertical-align: -0.125em !important;
 }
+
+:lang(ko) *,
+:lang(ja) *,
+:lang(zh) * {
+  word-break: keep-all;
+}
--- a/react/src/components/DefaultProviders.tsx
+++ b/react/src/components/DefaultProviders.tsx
@@ -174,6 +174,8 @@ export const useCurrentLanguage = () => {
     // For changing locale globally, use dayjs.locale instead of dayjs().locale
     dayjs.locale(lang);
     // eslint-disable-next-line react-hooks/exhaustive-deps
+    document.documentElement.lang = lang;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   useEffect(() => {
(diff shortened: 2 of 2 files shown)


react/src/index.css
agatha197, line 1: There're more language we supported. Is any reason to add only these languages?

limshyun: > There're more language we supported. Is any reason to add only these languages?

The word-break: keep-all CSS property is specifically designed for CJK (Chinese, Japanese, Korean) languages only.

Why only ko, ja, zh?

CJK languages don't use spaces between words, so browsers by default break text at any character boundary, which can split words awkwardly mid-syllable
word-break: keep-all tells the browser to keep semantic word units together and only break at appropriate boundaries (like punctuation or natural word breaks)

Why not other languages?

Western languages (en, fr, de, etc.) already use spaces as natural word boundaries
Applying keep-all to Western languages could actually cause problems - long words without hyphens would never break, potentially causing overflow issues
This is a CSS best practice for CJK text handling. See [MDN word-break documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#keep-all) which specifically mentions this property is for CJK text.

PR #5533, opened 2026-02-24
fix(FR-2114): move @ts-ignore to correct import line in global-stores.ts

### LINKED ISSUE #5537: Fix misplaced @ts-ignore in global-stores.ts causing TS2307 error
The `@ts-ignore` comment in `react/src/global-stores.ts` was placed on the wrong line — it suppressed errors for the `generateUUID` import (which doesn't need it) instead of the `backend.ai-client-esm` import (which does).\n\nThis caused a TS2307 error: `Cannot find module 'backend.ai-client-esm' or its corresponding type declarations`.\n\n\*Fix:\* Move the `@ts-ignore` comment to the correct line.

JIRA Issue: FR-2114

agatha197:
Resolves #5537 (FR-2114)

## Summary

- Move `@ts-ignore` comment from the `generateUUID` import (which doesn't need it) to the `backend.ai-client-esm` import (which does)
- Fixes TS2307 error: Cannot find module 'backend.ai-client-esm'

### DIFF
--- a/react/src/global-stores.ts
+++ b/react/src/global-stores.ts
@@ -21,8 +21,8 @@
 // Backend.AI client classes on globalThis.
 // Must be available before any component calls createBackendAIClient().
 // ---------------------------------------------------------------------------
-// @ts-ignore - resolved via craco webpack alias to dist/lib/backend.ai-client-esm.js
 import { generateUUID } from './helper/uuid';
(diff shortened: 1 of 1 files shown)


ironAiken2 (approved):
LGTM

PR #5658, opened 2026-02-26
chore(FR-2175): fix plugin loader to load from dist/plugins

nowgnuesLee:
Resolves [FR-2175](https://lablup.atlassian.net/browse/FR-2175)

## Summary
- Fix stale `configRef.current` in `doGQLConnect` by reading directly from Jotai store (`jotaiStore.get(loginConfigState)`) to get the latest config after `loadConfigFromWebServer` merges endpoint config
- Fix stale `apiEndpoint` closure in `doGQLConnect` by reading from `globalThis.backendaiclient._config.endpoint` instead of the captured closure value
- Add missing `loadConfigFromWebServer(ep)` call in `handleLogin` for Electron SESSION mode (was only present in auto-login path)
- Move plugin config detection from `useInitializeConfig` to `PluginLoader` — PluginLoader now derives `pluginConfigString` directly from `rawConfigState`, so it automatically picks up plugin pages from endpoint config merged by `loadConfigFromWebServer`
- Prevent double config.toml fetch from two LoginView instances (eager + lazy) using module-level `configInitialized` flag instead of per-instance `useRef`

## Test plan
- [ ] Build Electron app (`make mac`) and verify plugins load after login
- [ ] Verify web deployment still loads plugins correctly
- [ ] Verify login works in both SESSION and API modes
- [ ] Verify config.toml from endpoint is properly merged in Electron

[FR-2175]: https://lablup.atlassian.net/browse/FR-2175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nowgnuesLee:
* **#5658** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/5658?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/5658?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=5658).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

yomybaby (changes requested):
Please check a Copilot review.

react/src/components/PluginLoader.tsx
```diff
@@ -84,10 +84,10 @@ function PluginLoader() {
           const pluginUrl =
             (globalThis as Record<string, unknown>).isElectron && apiEndpoint
               ? `${apiEndpoint}/dist/plugins/${page}.js`
-              : `../plugins/${page}.js`;
+              : `/dist/plugins/${page}.js`;
```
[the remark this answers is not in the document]
nowgnuesLee: Good point. Added plugin name sanitization matching the existing pattern in LoginView.tsx. Invalid names (containing characters other than alphanumeric, hyphens, and underscores) are now rejected with a warning log before the import attempt.

react/src/components/LoginView.tsx
```diff
@@ -576,6 +589,9 @@ const LoginView: React.FC = () => {
         setIsLoading(false);
         return;
       }
+      if ((globalThis as Record<string, unknown>).isElectron) {
+        await loadConfigFromWebServer(ep);
+      }
       await connectUsingSession(true, ep);
```
yomybaby, line 596: `loadConfigFromWebServer` is called only in the `SESSION` branch, but it loads server-specific settings that apply regardless of auth mode — including resource limits (`maxCPU/MemoryPerContainer`, CUDA/ROCm/TPU device limits), `blockList`/`inactiveList` (menu access), `allow_image_list`, `pluginPages`, and many UI flags.

An Electron user connecting via `API` mode will silently miss all these per-server settings.

Please move the Electron check above the `SESSION`/`API` branch so it runs in both modes:
```ts
if ((globalThis as Record<string, unknown>).isElectron) {
  await loadConfigFromWebServer(ep);
}

if (connectionMode === 'SESSION') {
  ...
  await connectUsingSession(true, ep);
} else {
  ...
  await connectUsingAPI(true, ep);
}
```
Same fix is needed in the `login` callback (line ~642).

yomybaby (changes requested):
The `loadConfigFromWebServer` call in Electron mode is placed inside the `SESSION` branch in both `handleLogin` and the `login` callback, so it is never invoked when the user connects via `API` mode.

However, `loadConfigFromWebServer` loads much more than login-specific settings — via `refreshConfigFromToml` it also populates:
- **Resource limits** (`maxCPUCoresPerContainer`, `maxMemoryPerContainer`, all CUDA/ROCm/TPU/etc. device limits, `maxShmPerContainer`, `maxFileUploadSize`)
- **Menu access control** (`blockList`, `inactiveList`)
- **Image policy** (`allow_image_list`, `showNonInstalledImages`)
- **Plugin pages** (`pluginPages`)
- **General UI flags** (`maskUserInfo`, `hideAgents`, `enableContainerCommit`, `allowAppDownloadPanel`, etc.)

All of these are server-specific and apply regardless of the authentication mode. An Electron user connecting to a different server via `API` mode will silently miss all these per-server settings.

Please move the Electron check before the `SESSION`/`API` branch in both `handleLogin` and the `login` callback:

```ts
// Move this block outside/above the SESSION branch
if ((globalThis as Record<string, unknown>).isElectron) {
  await loadConfigFromWebServer(ep);
}

if (connectionMode === 'SESSION') {
  ...
  await connectUsingSession(true, ep);
} else {
  ...
  await connectUsingAPI(true, ep);
}
```

react/src/components/LoginView.tsx
nowgnuesLee: Good catch. Moved `loadConfigFromWebServer` above the SESSION/API branch in all three places (handleSubmit, login, and checkLogin callbacks) so Electron users in API mode also get server-specific settings.

PR #5705, opened 2026-03-03
fix(FR-2194): add missing @codemirror/lang-liquid dependency

### LINKED ISSUE #5704: Fix missing @codemirror/lang-liquid dependency causing webpack build failure
Adds @codemirror/lang-liquid as an explicit direct dependency to resolve a webpack build error.

## Why This Work is Needed

The @uiw/codemirror-extensions-langs package imports @codemirror/lang-liquid in its ESM bundle but does not declare it as a peer or optional dependency. Under pnpm's strict isolation model, transitive packages are not automatically available to consumers, so webpack fails to resolve the module at build time.

## Expected Outcomes

- Build succeeds without the @codemirror/lang-liquid module resolution error
- @codemirror/lang-liquid is explicitly tracked as a direct dependency in react/package.json

JIRA Issue: FR-2194

ironAiken2:
Resolves #5704 ([FR-2194](https://lablup.atlassian.net/browse/FR-2194))

## Summary

- Add `@codemirror/lang-liquid` as an explicit direct dependency to `react/package.json`
- Fixes webpack build failure caused by pnpm's strict module isolation

## Problem

`@uiw/codemirror-extensions-langs` imports `@codemirror/lang-liquid` in its ESM bundle but does not declare it as a peer or optional dependency. Under pnpm's strict isolation model, transitive packages are not hoisted, so webpack cannot resolve the module:

```
Module not found: Error: Can't resolve '@codemirror/lang-liquid' in
'.../@uiw/codemirror-extensions-langs/esm'
```

## Fix

Adding `@codemirror/lang-liquid@6.3.2` as an explicit dependency makes pnpm link it properly so webpack can resolve the import.

[FR-2194]: https://lablup.atlassian.net/browse/FR-2194?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ironAiken2:
Closing this PR — the build error could not be reproduced. The `@codemirror/lang-liquid` module resolution failure was likely a transient local issue (e.g., stale pnpm store or incomplete `node_modules` from a prior install) rather than a missing dependency in the project. A clean `pnpm install` resolves it without adding the explicit dependency.

PR #5955, opened 2026-03-13
fix: fix registry delete cancel E2E test strict mode violation

### LINKED ISSUE #5954: Fix registry delete cancel E2E test strict mode violation
The registry delete cancel E2E test fails because `filter({ hasText: registryName })` matches multiple table rows sharing the same registry name (e.g., `cr.backend.ai`). Playwright's strict mode throws when `toBeVisible()` resolves to more than one element.

Fix: Add `.first()` after the filter to check that at least one matching row remains visible after canceling deletion.

---
Captured while working on branch: 03-12-fix_fr-2279_fix_user-crud_e2e_test_fragile_radio_assertion_and_keypair_modal_name

agatha197:
Resolves #5954

## Summary
- Fix registry delete cancel E2E test failing due to Playwright strict mode violation
- `filter({ hasText: registryName })` matched multiple table rows sharing the same registry name (e.g., `cr.backend.ai`)
- Added `.first()` after the filter to verify at least one matching row remains visible after canceling deletion

### DIFF
--- a/e2e/environment/registry.spec.ts
+++ b/e2e/environment/registry.spec.ts
@@ -551,10 +551,11 @@ test.describe(
       await confirmDialog.getByRole('button', { name: 'Cancel' }).click();
       await expect(confirmDialog).toBeHidden();
 
-      // Registry row is still present
+      // Registry row is still present (use first() since multiple rows may share the same registry name)
       const rowAfterCancel = page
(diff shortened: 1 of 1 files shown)


yomybaby (approved):
LGTM

Issue #5966, opened 2026-03-13
Diagnostics System Enhancement


agatha197:
Enhance the WebUI diagnostics system with new diagnostic checks (CORS, auto-notification), result export capability, and address logical gaps and quality improvements across the diagnostics architecture.

Spec: .specs/FR-XXXX-diagnostics-enhancement/spec.md

Covers:

- 3 new features (export, auto-notification, CORS check)
- 5 logical gap fixes (SSL mismatch relocation, configRules tests, 404 handling, configurable threshold, frame-src)
- 4 improvements (category field, toggle passed, dedup data sources, smarter refresh)

JIRA Issue: FR-2300

PR #5972, opened 2026-03-13
feat(FR-2300): add 404 warning for endpoint health check

agatha197:
Resolves #5966 (FR-2300)

## Summary
- Add HTTP 404 detection to endpoint health check, returning a warning-level diagnostic instead of silently passing
- Pass `statusCode` from the fetch response through `useEndpointDiagnostics` to `checkEndpointReachability`
- Add new i18n keys for the "Endpoint Not Found" diagnostic (title, description, remediation)
- Remove unused SSL mismatch i18n keys (`SslMismatch`, `SslMismatchDesc`, `SslMismatchFix`)
- Expand unit tests covering 200, 401, 403, 404, 500, and network error scenarios

## Changed Files
- `react/src/diagnostics/rules/endpointRules.ts` — add `statusCode` param and 404 warning logic
- `react/src/diagnostics/rules/__tests__/endpointRules.test.ts` — add tests for status code scenarios
- `react/src/hooks/useEndpointDiagnostics.ts` — pass `statusCode` from fetch to the rule function
- `resources/i18n/en.json` — add `EndpointNotFound*` keys, remove unused `SslMismatch*` keys


resources/i18n/en.json
```diff
@@ -710,9 +713,6 @@
     "ResourceLimitOutOfRangeFix": "Review the {{section}} section in config.toml and set {{field}} to a value between {{min}} and {{max}}.",
     "ResourceLimitsPassed": "Resource Limits OK",
     "ResourceLimitsPassedDesc": "All resource limit configurations are within reasonable ranges.",
```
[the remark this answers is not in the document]
agatha197: Good catch. Added `ConfigSslMismatchFix` key to `en.json` and overridden `remediationKey` in the hook to use it. The original `SslMismatchFix` remains in other locale files for backward compatibility.

resources/i18n/en.json
```diff
@@ -669,6 +669,9 @@
     "CspWsProxyBlockedDesc": "The WebSocket proxy URL ({{proxyUrl}}) is not included in the Content-Security-Policy connect-src directive.",
     "CspWsProxyBlockedFix": "Add {{proxyUrl}} to the connect-src directive in your CSP meta tag or server headers.",
     "EndpointConnectivity": "Endpoint Connectivity",
+    "EndpointNotFound": "API Endpoint Not Found",
+    "EndpointNotFoundDesc": "The API endpoint ({{endpoint}}) returned a 404 response. This may indicate an incorrect endpoint URL.",
+    "EndpointNotFoundFix": "Check the API endpoint URL in config.toml and ensure it points to the correct Backend.AI server.",
```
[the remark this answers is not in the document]
agatha197: Translations for non-English locales are handled in a separate i18n batch workflow. These keys will be translated when the feature set stabilizes.

react/src/hooks/useEndpointDiagnostics.ts
```diff
@@ -37,7 +38,10 @@ export function useEndpointDiagnostics(): {
           method: 'GET',
           signal: AbortSignal.timeout(10000),
         });
-        return { isReachable: response.ok || response.status < 500 };
+        return {
+          isReachable: response.status < 500,
+          statusCode: response.status,
```
ironAiken2: Even when the endpoint is invalid, a 200 response is returned, so if we interpret this, it would seem to be treated as an endpoint that is always reachable.

resources/i18n/en.json
```diff
@@ -633,6 +633,7 @@
     "ConfigSslMatchPassedDesc": "API endpoint and proxy URL use the same protocol.",
     "ConfigSslMismatch": "Config SSL Protocol Mismatch",
     "ConfigSslMismatchDesc": "The API endpoint protocol ({{apiProtocol}}) and proxy URL protocol ({{proxyProtocol}}) in your configuration use different SSL settings.",
+    "ConfigSslMismatchFix": "Ensure both endpoints use the same protocol (both HTTPS or both HTTP).",
```
ironAiken2, line 636: Translations for other languages are missing.

react/src/hooks/useEndpointDiagnostics.ts
```diff
@@ -37,7 +38,10 @@ export function useEndpointDiagnostics(): {
...
+        };
       } catch (e) {
         return {
           isReachable: false,
```
ironAiken2, line 43: <p class='graphite__hidden'><i>[Re: line +53]</i></p>
If `staleTime` is set to 1 minute, the information may not be updated even if you perform a re-query. How about setting it to 0 so that new data is requested every time?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/5972?utm_source=unchanged-line-comment">Graphite</a>.</p>

react/src/hooks/useEndpointDiagnostics.ts
agatha197: The Backend.AI Manager typically returns 200 for the root path, but this 404 check guards against misconfigured reverse proxies or incorrect endpoint paths. It's a defensive check for edge cases rather than the standard path.

resources/i18n/en.json
agatha197: Translations for all supported languages are added in the stacked PR #5980.

react/src/hooks/useEndpointDiagnostics.ts
agatha197: Good point. Updated staleTime to 0 so fresh data is always fetched on re-query.

ironAiken2 (changes requested):
If the API endpoint specified in `config.toml` is incorrect, the login will fail, and you will not be able to access the diagnostics page.

react/src/hooks/useEndpointDiagnostics.ts
agatha197: Agreed. The 404 check has been removed entirely — if the API endpoint is wrong, login won't succeed in the first place, so this diagnostic is unnecessary.

resources/i18n/en.json
agatha197: The EndpointNotFound i18n keys have been removed along with the 404 check feature, so translations are no longer needed.

react/src/hooks/useEndpointDiagnostics.ts
agatha197: staleTime is now set to 0 as suggested.

PR #6116, opened 2026-03-23
chore: fix untranslated toTextKey in auto-diagnostics notification

### LINKED ISSUE #6115: Fix untranslated toTextKey in auto-diagnostics notification
The auto-diagnostics notification passes a raw i18n key string instead of the translated value to `toTextKey`, causing the notification link to display the key literal instead of the localized text.

## Why This Work is Needed

The `toTextKey` prop in `useAutoDiagnostics` was receiving `'diagnostics.ViewDiagnostics'` as a plain string instead of the translated value via `t()`. This means users see the raw key instead of the properly translated label.

## Expected Outcomes

- The diagnostics notification link displays the correct translated text
- Consistent with other notification calls that use `t()` for text values

agatha197:
Resolves #6115

## Summary
- Fix `toTextKey` in auto-diagnostics notification to pass translated string via `t()` instead of raw i18n key

## Test plan
- [ ] Verify diagnostics notification link displays translated text instead of raw key

### DIFF
--- a/react/src/hooks/useAutoDiagnostics.ts
+++ b/react/src/hooks/useAutoDiagnostics.ts
@@ -64,7 +64,7 @@ export function useAutoDiagnostics(): void {
       duration: 0,
       open: true,
       to: '/diagnostics',
-      toTextKey: 'diagnostics.ViewDiagnostics',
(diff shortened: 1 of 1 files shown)


ironAiken2 (approved):
LGTM

PR #6178, opened 2026-03-26
fix(FR-2384): enable LLM Chat Test button when any route is HEALTHY

### LINKED ISSUE #6176: Enable LLM Chat Test button when any route status is HEALTHY
## Problem

The LLM Chat Test button on the Endpoint Detail page is currently disabled based on `endpoint?.status !== 'HEALTHY'`. This means the button is only enabled when the overall endpoint status is HEALTHY.

## Expected Behavior

The LLM Chat Test button should be enabled when at least one route's status is HEALTHY, regardless of the overall endpoint status. This allows users to test the chat functionality as soon as any route becomes available.

## Implementation

- Check `endpoint.routings` array for any route with `status === 'HEALTHY'`
- Use this condition instead of `endpoint?.status !== 'HEALTHY'` for the button disabled state

JIRA Issue: FR-2384

nowgnuesLee:
Resolves #6176 (FR-2384)

## Summary

- The "LLM Chat Test" button was previously disabled unless the endpoint's overall status was `HEALTHY`
- Changed to check if **any individual route** has `HEALTHY` status by adding a separate `healthyRoutes` query with `filter: { status: [HEALTHY] }`
- This allows users to test LLM chat as long as at least one route is healthy, even if the endpoint itself isn't fully healthy
- Falls back to endpoint status check for servers that don't support `route-node`

## Changes

- `react/src/pages/EndpointDetailPage.tsx`
  - Added `healthyRoutes` alias query to fetch count of HEALTHY routes
  - Added `hasAnyHealthyRoute` derived state
  - Changed button disabled condition from `endpoint?.status !== 'HEALTHY'` to `!hasAnyHealthyRoute`

## Test plan

- [ ] Deploy an endpoint with multiple routes where some are HEALTHY and some are not
- [ ] Verify the "LLM Chat Test" button is enabled when at least one route is HEALTHY
- [ ] Verify the button is disabled when no routes are HEALTHY
- [ ] Verify behavior on older backends without `route-node` support (should fall back to endpoint status)


react/src/pages/EndpointDetailPage.tsx
```diff
@@ -167,210 +167,225 @@ const EndpointDetailPage: React.FC<EndpointDetailPageProps> = () => {
...
+        $routeOffset: Int
+        $skipRouteNodes: Boolean!
+        $skipRoutings: Boolean!
+      ) {
+        endpoint(endpoint_id: $endpointId) {
+          name
+          status
+          endpoint_id
+          project
+          image_object @since(version: "23.09.9") {
```
ironAiken2: Since the new release only needs to maintain backward compatibility up to version 26.2.x, it seems safe to remove any unused directives.

react/src/pages/EndpointDetailPage.tsx
```diff
@@ -167,210 +167,225 @@ const EndpointDetailPage: React.FC<EndpointDetailPageProps> = () => {
...
+            }
+            labels {
+              key
+              value
+            }
+            size_bytes
+            supported_accelerators
+            ...ImageNodeSimpleTagFragment
+          }
+          desired_session_count @deprecatedSince(version: "24.12.0")
```
ironAiken2: Since this part has been completely deprecated, you can probably remove it from your code.

ironAiken2 (changes requested):

PR #6452, opened 2026-04-07
fix(FR-2467): disable session commit button until session is Running

### LINKED ISSUE #6443: Disable session commit button in detail drawer until session is Running
From Jira FR-2467: The session commit button in the session detail drawer should be disabled when the session is not in RUNNING state.

ironAiken2:
Resolves #6443 (FR-2467)

## Summary
- Disable commit button when session status is not RUNNING
- Previously only checked isActive() which allowed non-running states like PREPARING, PENDING, PULLING to enable the button

## Test plan
- [ ] Verify commit button is disabled when session is in PREPARING state
- [ ] Verify commit button is disabled when session is in PENDING state
- [ ] Verify commit button is enabled when session is in RUNNING state
- [ ] Verify commit button remains disabled for non-owners regardless of session status

### DIFF
--- a/react/src/components/ComputeSessionNodeItems/SessionActionButtons.tsx
+++ b/react/src/components/ComputeSessionNodeItems/SessionActionButtons.tsx
@@ -329,7 +329,7 @@ const SessionActionButtons: React.FC<SessionActionButtonsProps> = ({
           >
             <Button
               size={size}
-              disabled={!isActive(session) || !isOwner}
+              disabled={session?.status !== 'RUNNING' || !isOwner}
               icon={<BAIContainerCommitIcon />}
               onClick={() => {
                 onAction?.('containerCommit');


react/src/components/ComputeSessionNodeItems/SessionActionButtons.tsx
```diff
@@ -329,7 +329,7 @@ const SessionActionButtons: React.FC<SessionActionButtonsProps> = ({
           >
             <Button
               size={size}
-              disabled={!isActive(session) || !isOwner}
+              disabled={session?.status !== 'RUNNING' || !isOwner}
```
nowgnuesLee, line 332:
```suggestion
              disabled={!session?.status || (session.status !== 'RUNNING' || !isOwner)}
```

ironAiken2: session.status !== 'RUNNING' already covers the case where session?.status is
undefined or null, since a falsy value can never equal 'RUNNING'.

So disabled={session?.status !== 'RUNNING' || !isOwner} is equivalent.

nowgnuesLee (dismissed):
LGTM

PR #6668, opened 2026-04-14
refactor(FR-2493): centralize resource slot metadata into BAIMetaDataProvider

nowgnuesLee:
Resolves #6484 (FR-2493)

## Summary

Move the `/config/resource-slots/details` fetch out of `backend.ai-ui` and into the webui, and let package components read the result from context.

The metadata splits along an **auth boundary**, so the providers do too:

| | `BAIMetaDataProvider` | `BAIResourceSlotsProvider` |
| --- | --- | --- |
| Carries | `device_metadata.json`, `image_metadata.json`, `imagePath` | the server's configured resource slots |
| Request | none — static files bundled with the app | authenticated (`/config/resource-slots/details`) |
| Mounted at | app root (`DefaultProvidersForReactRoot`) | inside `MainLayout`'s routed subtree |

Keeping the static half app-wide means the login screen, header and notification drawer still render under it, and no signed request fires before sign-in.

`useBAIResourceSlots()` reads both and returns `deviceMetaData` / `resourceSlots` / `mergedResourceSlots`. The merge (`server slots over static metadata`) happens **once inside the provider**, not in every consumer — an agent list mounts this hook hundreds of times.

The hook is null-safe like `useBAIImageMetaData`: outside `BAIResourceSlotsProvider` — anonymous routes, Storybook — it falls back to the static metadata instead of throwing, so a host that supplies only `deviceMetaData` still renders units and icons.

### `backend.ai-ui`

- **`BAIMetaDataProvider`** — static metadata only (`deviceMetaData`, `imageMetaData`, `imagePath`).
- **`BAIResourceSlotsProvider`** (new) — takes `resourceSlots`, reads `deviceMetaData` from the outer context, and publishes both the raw and merged tables.
- **`useBAIResourceSlots`** (new, replaces `useBAIDeviceMetaData`) — reads both contexts via React 19 `use`, exported from the provider barrel.
- **`useBAIIconPath`** (new) — owns the `imagePath` + filename join, so the package resolves icon URLs in exactly one place and never hardcodes an app asset path. Both `getImageIcon` and the resource type icon use it.
- **`BAIImageWithFallback`** — moved here from `react/src/components/ImageWithFallback.tsx`. A server-configured `display_icon` the package doesn't bundle now loads from the host's icon directory, falling back to a generic glyph when the file is missing or no `imagePath` was provided.
- `useResourceSlotsDetails` is removed from `hooks/index.ts` — the package no longer fetches.

### `react` (host)

- `DefaultProviders.tsx` — unchanged from `main`; it keeps providing the static metadata app-wide.
- `MainLayout.tsx` — new `ResourceSlotsWrapper` calls `useResourceSlotsDetails()` and feeds `BAIResourceSlotsProvider` around `<Outlet />`. It shares the existing react-query cache entry, so this warms the cache rather than adding a request.
- 21 components now import `useResourceSlotsDetails` from `src/hooks/backendai` instead of `backend.ai-ui`, including `SessionReclamationStatusCell`/`Popover`, which landed on `main` while this PR was open.
- `ResourceNumber.tsx` is deleted. Its default export had already been superseded by `BAIResourceNumberWithIcon`; the surviving `ResourceTypeIcon` kept its own icon lookup and a leading-slash `/resources/icons/...` path, which breaks under subpath and Electron `file://` serving. Its one live call site (`SessionLauncherPage.tsx`) now uses the package's equivalent, so no resource-slot icon path is built with a leading slash anywhere.

### Instructions

- `.github/instructions/react.instructions.md` — extends the `'use memo'` policy from components to custom hooks (`use*`), and makes it Required rather than Recommended. Strictly this is **outside the PR's goal**; it is included because this PR adds `'use memo'` to several new hooks and the existing wording only covered components. Happy to split it out if reviewers prefer.

## Verification

`bash scripts/verify.sh` → **ALL PASS** (Relay, Lint, Format, TypeScript, Vite warmup paths, Terminology).

Unit tests: **960 passed** (`react`), **363 passed / 1 skipped** (`backend.ai-ui`).

## How to test

- [ ] Login screen renders without blocking, and no `/config/resource-slots/details` request fires before sign-in.
- [ ] Resource slot icons, units and names render on the session launcher, agent list, and resource preset pages.
- [ ] A server-configured accelerator whose `display_icon` has no bundled component renders its SVG from `resources/icons/`; one with no matching file falls back to the generic chip glyph.
- [ ] Agent detail modal shows correct resource metadata.
- [ ] My Resource and Fair Share pages work with the new hook imports.
- [ ] Storybook: `BAIAgentTable` and `BAIResourceNumberWithIcon` stories render units and icons with only `deviceMetaData` supplied.


nowgnuesLee:
* **#6669** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6669?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6668** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6668?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7552** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7552?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7533** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7533?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7526** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7526?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7519** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7519?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7505** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=6668).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend.ai-ui/src/components/provider/BAIMetaDataProvider/BAIMetaDataProvider.tsx
```diff
@@ -1,18 +1,37 @@
...
   children,
 }: BAIMetaDataProviderProps) => {
+  'use memo';
+  const value: BAIDeviceMetaDataContextValue = {
+    deviceMetaData,
+    resourceSlotsInRG,
+    mergedResourceSlots,
+    refresh,
+    isLoading,
+  };
```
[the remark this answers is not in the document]
nowgnuesLee: The `BAIMetaDataProvider` component has the `'use memo'` directive, so the React Compiler automatically memoizes the `value` object — no manual `useMemo` needed.

packages/backend.ai-ui/src/components/provider/BAIMetaDataProvider/hooks/useBAIMetaData.ts
```diff
@@ -0,0 +1,26 @@
...
+ * - `resourceSlotsInRG`: server-side resource slots for the current resource group
+ * - `mergedResourceSlots`: merged result of the above two (always a defined object)
+ * - `refresh`: re-fetches resource slot details
+ * - `isLoading`: whether the resource slot details are currently loading
+ */
+const useBAIMetaData = () => {
+  const context = useContext(BAIDeviceMetaDataContext);
+  if (!context) {
+    throw new Error('useBAIMetaData must be used within a BAIMetaDataProvider');
+  }
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in the latest commit.

packages/backend.ai-ui/src/components/provider/BAIMetaDataProvider/hooks/useBAIMetaData.ts
```diff
@@ -0,0 +1,26 @@
+import { BAIDeviceMetaDataContext } from '../context';
+import { useContext } from 'react';
+
+/**
+ * Returns all resource slot metadata from the BAIMetaDataProvider context.
+ * - `deviceMetaData`: raw static device metadata from `resources/device_metadata.json`
+ * - `resourceSlotsInRG`: server-side resource slots for the current resource group
+ * - `mergedResourceSlots`: merged result of the above two (always a defined object)
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in the latest commit — docstring now clarifies the scoping behavior.

nowgnuesLee:
Addressed the two suppressed Copilot comments for `BAIImageWithFallback`:

1. **`hasError` not resetting on `src` change** — Replaced `useState(false)` + `useEffect` reset pattern with derived state: `errorSrc` tracks which `src` failed, and `hasError` is derived as `errorSrc === src`. When `src` changes, the error state clears automatically without an effect.

2. **Missing `'use memo'` directive** — Added `'use memo'` as the first statement in the component body.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
```diff
@@ -72,7 +72,7 @@ const BAIAgentTable: React.FC<BAIAgentTableProps> = ({
   const { token } = theme.useToken();
```
agatha197, line 72: Component body is missing the `'use memo'` directive. Please add it as the first statement inside the function body.

packages/backend.ai-ui/src/components/fragments/BAIProjectSettingModal.tsx
```diff
@@ -69,7 +69,7 @@ const BAIProjectSettingModal = ({
   const { token } = theme.useToken();
```
agatha197, line 69: Component body is missing the `'use memo'` directive. Please add it as the first statement inside the function body.

packages/backend.ai-ui/src/components/provider/BAIMetaDataProvider/BAIMetaDataProvider.tsx
```diff
@@ -1,18 +1,37 @@
...
 export interface BAIMetaDataProviderProps {
   deviceMetaData?: DeviceMetaData;
+  resourceSlotsInRG?: DeviceMetaData;
+  mergedResourceSlots?: DeviceMetaData;
+  refresh?: () => void;
+  isLoading?: boolean;
   children?: ReactNode;
 }
 
 const BAIMetaDataProvider = ({
```
agatha197, line 17: Component body is missing the `'use memo'` directive. Please add it as the first statement inside the function body. Also, please add an explicit `React.FC<BAIMetaDataProviderProps>` type annotation on the component for consistency with the project convention.

packages/backend.ai-ui/src/components/provider/BAIMetaDataProvider/index.ts
```diff
@@ -1,5 +1,6 @@
 export { default as BAIMetaDataProvider } from './BAIMetaDataProvider';
 export type { BAIMetaDataProviderProps } from './BAIMetaDataProvider';
 export { BAIDeviceMetaDataContext } from './context';
```
agatha197, line 3: The context and its value type are named `BAIDeviceMetaDataContext` / `BAIDeviceMetaDataContextValue` (with "Device"), but the public-facing API — `BAIMetaDataProvider` and `useBAIMetaData` — does not include "Device". This naming inconsistency leaks an implementation detail into the public API.

Two options:
1. Rename to `BAIMetaDataContext` / `BAIMetaDataContextValue` to align with the provider and hook names.
2. Remove the context export entirely — it is an internal implementation detail since `useBAIMetaData` is the intended access point.

react/src/components/AgentDetailModal.tsx
```diff
@@ -7,9 +7,9 @@ import {
   convertToDecimalUnit,
   toFixedFloorWithoutTrailingZeros,
 } from '../helper';
+import { useResourceSlotsDetails } from '../hooks/backendai';
```
agatha197, line 10: The component body (line 43) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

react/src/components/AgentStats.tsx
```diff
@@ -2,10 +2,10 @@
  @license
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
+import { useResourceSlotsDetails } from '../hooks/backendai';
```
agatha197, line 5: The component body (line 37) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

react/src/components/AgentSummaryList.tsx
```diff
@@ -11,7 +11,7 @@ import {
   convertToBinaryUnit,
   toFixedFloorWithoutTrailingZeros,
 } from '../helper';
-import { ResourceSlotName } from '../hooks/backendai';
+import { ResourceSlotName, useResourceSlotsDetails } from '../hooks/backendai';
```
agatha197, line 14: The component body (line 60) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

react/src/components/ComputeSessionNodeItems/SessionIdleChecks.tsx
```diff
@@ -8,14 +8,10 @@ import {
   formatDurationAsDays,
   toFixedFloorWithoutTrailingZeros,
 } from '../../helper';
+import { useResourceSlotsDetails } from '../../hooks/backendai';
```
agatha197, line 11: The component body (line 131) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

react/src/components/ComputeSessionNodeItems/SessionSlotCell.tsx
```diff
@@ -4,17 +4,15 @@
  */
 import { SessionSlotCellFragment$key } from '../../__generated__/SessionSlotCellFragment.graphql';
 import { convertToBinaryUnit } from '../../helper';
-import { ResourceSlotName } from '../../hooks/backendai';
+import {
+  ResourceSlotName,
+  useResourceSlotsDetails,
```
agatha197, line 9: The component body (line 28) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

react/src/components/KeypairResourcePolicySettingModal.tsx
```diff
@@ -14,7 +14,7 @@ import {
 import { convertToBinaryUnit } from '../helper';
 import { MAX_CPU_QUOTA, SIGNED_32BIT_MAX_INT } from '../helper/const-vars';
 import { useSuspendedBackendaiClient } from '../hooks';
-import { useResourceSlots } from '../hooks/backendai';
+import { useResourceSlots, useResourceSlotsDetails } from '../hooks/backendai';
```
agatha197, line 17: The component body (line 65) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

react/src/components/ManageImageResourceLimitModal.tsx
```diff
@@ -8,6 +8,7 @@ import {
 } from '../__generated__/ManageImageResourceLimitModalMutation.graphql';
 import { ManageImageResourceLimitModal_image$key } from '../__generated__/ManageImageResourceLimitModal_image.graphql';
 import { compareNumberWithUnits } from '../helper';
+import { useResourceSlotsDetails } from '../hooks/backendai';
```
agatha197, line 11: The component body (line 47) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

react/src/components/MyResource.tsx
```diff
@@ -2,7 +2,7 @@
  @license
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
-import { ResourceSlotName } from '../hooks/backendai';
+import { ResourceSlotName, useResourceSlotsDetails } from '../hooks/backendai';
```
agatha197, line 5: The component body (line 35) is missing the `'use memo'` directive. Please add `'use memo';` as the first statement inside the function body to enable React Compiler optimization.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

packages/backend.ai-ui/src/components/fragments/BAIProjectSettingModal.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/AgentDetailModal.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/AgentStats.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/AgentSummaryList.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/ComputeSessionNodeItems/SessionIdleChecks.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/ComputeSessionNodeItems/SessionSlotCell.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/KeypairResourcePolicySettingModal.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/ManageImageResourceLimitModal.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

react/src/components/MyResource.tsx
nowgnuesLee: Fixed in the latest commit — added `'use memo'` as the first statement in the component body.

packages/backend.ai-ui/src/components/provider/BAIMetaDataProvider/BAIMetaDataProvider.tsx
nowgnuesLee: The `'use memo'` directive is already present as the first statement in the component body (line 25). Regarding the `React.FC<BAIMetaDataProviderProps>` annotation — I'll leave that open for discussion as the project has been moving away from `React.FC` in newer code; happy to align either way based on team preference.

nowgnuesLee: `React.FC` is not a project convention — newer components in this codebase use plain function signatures with destructured-prop typing (e.g. `({ ... }: Props) => { ... }`). Leaving it as-is. The `'use memo'` directive is already in place.

packages/backend.ai-ui/src/components/provider/BAIMetaDataProvider/index.ts
nowgnuesLee: Renamed `BAIDeviceMetaDataContext` / `BAIDeviceMetaDataContextValue` to `BAIMetaDataContext` / `BAIMetaDataContextValue` to align with the provider and hook naming.

agatha197 (dismissed):
please resolve conflicts

.claude/rules/use-memo-directive.md
```diff
@@ -0,0 +1,46 @@
+---
```
ironAiken2: Should this be added?

.github/instructions/react.instructions.md
```diff
@@ -8,10 +8,10 @@ These instructions apply to React components in the `/react` directory.
```
ironAiken2, line 8: Should this be added?

ironAiken2 (changes requested):

react/src/components/ComputeSessionNodeItems/SessionSlotCell.tsx
```diff
@@ -27,6 +25,7 @@ const SessionSlotCell: React.FC<OccupiedSlotViewProps> = ({
   type,
   sessionFrgmt,
 }) => {
+  'use memo';
```
agatha197, line 29: `useResourceSlotsDetails('')` passes an empty string as `resourceGroupName`. Because the hook guards the `sgroup` query param with `resourceGroupName && search.set(...)`, an empty string is falsy and the parameter is silently dropped — making this call equivalent to `useResourceSlotsDetails(undefined)` (all-group, unscoped).

If the intent is **no resource-group filter**, this component should read from the context instead:

```tsx
// Preferred — reads already-fetched data from BAIMetaDataWrapper context
const { mergedResourceSlots } = useBAIMetaData();
```

If the intent is to scope to the **current resource group**, the actual group name needs to be passed (similar to `MyResourceWithinResourceGroup` which passes `deferredCurrentResourceGroup`). Please clarify the intent.

react/src/components/AgentDetailModal.tsx
```diff
@@ -40,6 +40,7 @@ const AgentDetailModal: React.FC<AgentDetailModalProps> = ({
   onRequestClose,
   ...modalProps
 }) => {
+  'use memo';
```
agatha197, line 43: **Design question: why do webui components call `useResourceSlotsDetails()` directly instead of reading from context?**

After this PR, BUI components (`BAIAgentTable`, `BAIProjectSettingModal`, `BAIResourceNumberWithIcon`) correctly read merged metadata from `useBAIMetaData()` (populated once by `BAIMetaDataWrapper` in `MainLayout`). However, most webui components — this one included, along with `AgentStats`, `AgentSummaryList`, `SessionIdleChecks`, `SessionUsageMonitor`, `MyResource`, `SessionMetricGraph`, `ResourceNumber`, `KeypairResourcePolicySettingModal`, `ManageImageResourceLimitModal` — still call `useResourceSlotsDetails()` directly without a resource group argument.

React Query deduplicates the network request, but each call site creates its own hook instance with an independent `useUpdatableState` key, which means:
1. A `refresh()` triggered via the context does **not** invalidate these direct-call instances.
2. Calling `useResourceSlotsDetails(undefined)` from many components when the same data is already in context is unnecessary.

Components that genuinely need a resource-group-scoped fetch (e.g., `MyResourceWithinResourceGroup`, `ResourceAllocationFormItems`) should keep the direct call. For everything else without a resource group argument, switching to `useBAIMetaData()` would make the data flow consistent:

```tsx
// Instead of:
const { mergedResourceSlots } = useResourceSlotsDetails();

// Prefer (reads from BAIMetaDataWrapper context):
const { mergedResourceSlots } = useBAIMetaData();
```

Happy to discuss if there is a reason to keep the direct calls — for example, if these components also need `refresh()` tied to their own lifecycle.

agatha197 (commented):
Two design questions worth addressing before merge (details in inline comments below).

react/src/components/AgentDetailModal.tsx
nowgnuesLee: The purpose of this PR is to make the existing `useResourceSlotsDetails` hook accessible from BUI as well — by routing its result through the provider. For webui components that do not need to introduce a BUI dependency, there is no reason to switch to `useBAIMetaData`; `useResourceSlotsDetails` is the correct hook to use directly. Reaching for `useBAIMetaData` in webui in place of `useResourceSlotsDetails` would actually be an incorrect usage pattern — the hook is intended to be consumed only within BUI.

`useBAIMetaData` was previously exported from the module's public surface (along with `BAIMetaDataContext`), which made the misuse possible. I've now restricted them so that `useBAIMetaData` can only be used within BUI.

.claude/rules/use-memo-directive.md
nowgnuesLee: No, I will remove this file. thanks!

.github/instructions/react.instructions.md
nowgnuesLee: I think this file should be added so that `use memo` can be applied to custom hooks.

agatha197 (approved):
LGTM

yomybaby (changes requested):
The direction is right and consistent with #7552's philosophy (host app fetches, package reads from context). A few things should be unified with the conventions settled in #7552's latest revisions, plus one regression.

### Requested changes

**1. Hardcoded asset path in the package (`BAIResourceNumberWithIcon.tsx:214`)**
`src={`/resources/icons/${displayIcon}.svg`}` resolves an app asset path inside `backend.ai-ui`. #7552 introduced `BAIImagePathContext` exactly to avoid this ("The package never bundles or resolves app asset paths on its own"). Please build the URL from the provided `imagePath` instead. Also note the leading-slash absolute path breaks under subpath/Electron `file://` serving, whereas #7552 uses a relative `resources/icons`.

**2. Compute `mergedResourceSlots` inside the provider (fixes a Storybook regression)**
The provider now receives a pre-merged `mergedResourceSlots` prop, so every host must compute it. `BAIAgentTable.stories.tsx:120` still passes only `deviceMetaData`, so `mergedResourceSlots` becomes `{}` and the story's table loses all units/icons/names (previously the BUI hook merged `deviceMetaData` itself). If `BAIMetaDataProvider` computes `merge({}, deviceMetaData, resourceSlotsInRG)` internally, this story is fixed without changes, the duplicated `deviceMetaData`/`mergedResourceSlots` props in `BAIResourceNumberWithIcon.stories.tsx` go away, and `BAIMetaDataWrapper` gets simpler.

**3. `useContext` → `use`**
#7552 migrated the repo to React 19 `use`; the new `useBAIMetaData.ts:16` reintroduces `useContext`.

**4. Export `useBAIMetaData` from the provider barrel**
`useBAIImageMetaData` is exported from `provider/BAIMetaDataProvider/index.ts`, but `useBAIMetaData` is not — 3 call sites (`BAIAgentTable.tsx:24`, etc.) use deep relative imports. Please add the barrel export and unify imports.

**5. Provider placement: keep static metadata app-wide**
Moving `BAIMetaDataWrapper` inside `MainLayout`'s `<Outlet />` fixes the login blocking, but it also pulls the auth-free static metadata (`deviceMetaData`, `imageMetaData`, `imagePath`) out of the header / notification drawer / login scope, where #7552 provided it app-wide. Nothing outside the Outlet consumes it *today*, but `useBAIImageMetaData` silently renders nothing there, and a throw from `useBAIMetaData` would be swallowed by `ErrorBoundaryWithNullFallback` (whole subtree silently disappears). Suggestion: keep the static-metadata provider at `DefaultProvidersForReactRoot`, and add only the authenticated resource-slots layer as a nested provider in `MainLayout`.

**6. Null-safety policy mismatch**
#7552's `useBAIImageMetaData` is null-safe with documented fallbacks; `useBAIMetaData` throws, while its `refresh = () => {}` default silently masks missing wiring. Please pick one policy for both hooks (and note the null-fallback error boundaries above make throwing less loud than it looks).

**7. Naming**
- `resourceSlotsInRG` in the context is actually **unscoped** (the wrapper calls `useResourceSlotsDetails()` with no `sgroup`, returning slots across all resource groups). The name suggests current-RG scoping and will mislead readers of e.g. `BAIProjectSettingModal`. Rename (e.g. `resourceSlots`) or make the scope explicit.
- `useBAIMetaData` is a generic name but returns only resource-slot data, while image metadata has its own `useBAIImageMetaData`. Suggest a domain-specific name (e.g. `useBAIResourceSlots`) to match #7552's naming style.

### Nits
- Stray double blank line at `BAIAgentTable.tsx:606-608`.
- The `react.instructions.md` 'use memo' policy change is unrelated to the PR's goal (fine to keep since the PR adds `'use memo'` to hooks, but worth a mention in the description).

Nice touches: shared cache key between the wrapper's unscoped query and component-level calls, the `useResourceSlotsDetails('')` → `()` cache unification in `SessionSlotCell`, and the `importOriginal`-based partial mocks in `MyResourceWithinResourceGroup.test.tsx`.

nowgnuesLee:
@yomybaby Thanks — the `#7552` comparison was the right frame, and following it resolved a regression I had not noticed. All seven items are addressed in `4437b285e`, and the PR description has been rewritten to match (it still described the old `BAIMetaDataWrapper` / `useBAIMetaData` shape).

One piece of context first: this branch was stacked on a pre-merge revision of `#7552`, which renamed `BAIMetaDataWrapper` → `BAIMetaDataProviderWrapper` shortly before it landed. Rebasing onto `main` therefore hit a rename-vs-modify conflict whose resolution silently produced a broken `DefaultProviders.tsx`. Your item 5 turned out to be both the diagnosis and the fix, so the recovery and the review response are the same change.

### 1. Hardcoded asset path

Fixed. Icon URLs now come from the host-provided `imagePath`, via a new `useBAIIconPath` hook. That hook owns the join and the "never guess an app asset URL" rule, so `getImageIcon` and the resource type icon share one implementation instead of two copies of `trimEnd(imagePath, '/') + '/' + filename`. When no `imagePath` is provided the icon falls through to the generic glyph rather than fabricating a path. No leading slash anywhere in the package.

### 2. Compute `mergedResourceSlots` inside the provider

Done, and it fixed the Storybook regression exactly as you predicted — `BAIAgentTable.stories.tsx` still passes only `deviceMetaData` and now renders units, icons and names again. The duplicated props in `BAIResourceNumberWithIcon.stories.tsx` are gone.

Two changes beyond the literal request:

- The merge lives in `BAIResourceSlotsProvider` rather than `BAIMetaDataProvider`, because of item 5 — the two halves mount at different points, and the authenticated one is where the server slots arrive.
- Moving it into the provider also removes real work. `useBAIResourceSlots` is instantiated hundreds of times on an agent list (per-cell and per-slot-icon), and each instance was deep-merging the whole slot table and retaining its own copy. It is now one merge and one copy per data change.

### 3. `useContext` → `use`

Fixed; the new provider module has no `useContext` left.

### 4. Barrel export

Fixed. `useBAIResourceSlots` is exported from `provider/BAIMetaDataProvider/index.ts` and all call sites import from the barrel — no deep relative imports remain.

### 5. Provider placement

Implemented as you suggested. `BAIMetaDataProvider` stays at `DefaultProvidersForReactRoot` with the auth-free static metadata (`deviceMetaData`, `imageMetaData`, `imagePath`), so the header, notification drawer and login screen are all still covered. Only the authenticated layer is nested in `MainLayout` around `<Outlet />`. `DefaultProviders.tsx` is now byte-identical to `main`.

### 6. Null-safety policy

Unified on the null-safe policy, matching `useBAIImageMetaData`. `useBAIResourceSlots` no longer throws; it degrades `merged → static → frozen empty object`, which is also what makes the Storybook case in item 2 work.

**Deviation worth flagging:** rather than making `refresh` loud, I removed `refresh` and `isLoading` from the context entirely. No BUI component consumed either, and the `refresh = () => {}` default was precisely the silent-masking you described — deleting the unused surface seemed better than keeping a footgun with a warning attached. The webui side keeps `useResourceSlotsDetails().refresh` for its own callers. Happy to put them back if you'd rather the package expose them for future use.

### 7. Naming

Both renamed: the context field `resourceSlotsInRG` → `resourceSlots`, and `useBAIMetaData` → `useBAIResourceSlots`.

**Partial:** I left the webui hook's own return field named `resourceSlotsInRG` (`react/src/hooks/backendai.tsx`). It has the same accuracy problem, but renaming it reaches 23 call sites and felt like scope creep for this PR. Can follow up separately if you want it consistent.

### Nits

- Stray double blank line in `BAIAgentTable.tsx` — removed.
- The `react.instructions.md` change is now called out in the PR description, including the note that it is outside this PR's goal and can be split out if preferred.

### One item no longer applies

`SessionIdleChecks` appears in your item 5 list of components calling `useResourceSlotsDetails()` directly. It was rewritten on `main` (FR-3318, `SessionReclamationStatusCell`) while this PR was open and no longer touches resource slot metadata at all, so the rebase took `main`'s version. Conversely, `SessionReclamationStatusPopover` arrived on `main` importing `useResourceSlotsDetails` from `backend.ai-ui` — which this PR removes — so it now imports from `src/hooks/backendai`.

### Known leftover, not in this PR's scope

`react/src/components/ResourceNumber.tsx` exports its own `ResourceTypeIcon` that still builds `/resources/icons/${displayIcon}.svg` with a leading slash. Your item 1 was about the package, and this is the host's parallel component, used in exactly one place (`SessionLauncherPage.tsx`). The rest of that file is already superseded by `BAIResourceNumberWithIcon`, so the clean fix is to point that import at the package and delete the file. Tell me if you'd like it folded in here rather than as a follow-up.

### Verification

`bash scripts/verify.sh` → ALL PASS. Unit tests: 960 (`react`), 363 + 1 skipped (`backend.ai-ui`).

nowgnuesLee:
Correction to my previous comment: the "known leftover" is no longer a leftover — it is in this PR now (`51d4875e2`).

### `react/src/components/ResourceNumber.tsx` is deleted

I had said the host's parallel `ResourceTypeIcon`, with its leading-slash `/resources/icons/...`, was out of scope. On a closer look the file was only half-migrated: its default export had already been superseded by `BAIResourceNumberWithIcon`, its one remaining live consumer was a single import in `SessionLauncherPage.tsx`, and the package already exports an equivalent `ResourceTypeIcon` that takes the props that call site passes. So the file is gone and the call site imports from `backend.ai-ui`.

**Item 1 is now complete repo-wide**: no resource-slot icon path is built with a leading slash anywhere. (The remaining `/resources/icons/...` literals — `FileBrowserButton`, `SFTPServerButton`, `AgentList` — are app/platform logos, a different domain.)

### Also since that comment

- **`UnifiedAcceleratorChip` reads from context.** It rendered the package's context-backed `ResourceTypeIcon` while calling `useResourceSlotsDetails()` two lines above for the description — the same unscoped query key and the same merge `BAIResourceSlotsProvider` already publishes. Both now come from `useBAIResourceSlots()`, which also retired that file's last import of the host hook and a rationale comment the provider had made false.
- **Provider dependency documented.** `BAIResourceNumberWithIcon` and `ResourceTypeIcon` degrade to the static metadata outside `BAIResourceSlotsProvider`. That is intentional and the stories rely on it, but nothing said so — now both carry it in their JSDoc.
- **Dead optional chaining dropped** on the nine `useBAIResourceSlots().mergedResourceSlots` reads, since that value is always defined. The `?.` after the property stays — a requested slot may genuinely be absent. The ~39 equivalent guards on `useResourceSlotsDetails().mergedResourceSlots` predate this PR and are untouched.
- **Storybook docs corrected.** The `ServerConfiguredIcon` story still described fetching `/resources/icons/{display_icon}.svg` — the exact absolute form this PR removes. It now documents `{imagePath}/{display_icon}.svg`.
- **Comments trimmed.** Several of the docs I added restated mechanics visible a line or two below (where the merge runs, what the guard returns, which branch is taken). Cut those; kept what the code cannot say — the auth boundary behind each provider's mount point, that `mergedResourceSlots` is always defined, that the package never guesses an app asset URL, and that a slot's description exists only in the server response.

`bash scripts/verify.sh` ALL PASS; 960 `react` and 363 `backend.ai-ui` tests pass.

nowgnuesLee:
One note on item 4, since it reverses something said earlier in this PR.

Addressing it means `useBAIResourceSlots` (formerly `useBAIMetaData`) is exported from `provider/BAIMetaDataProvider/index.ts`, and `components/index.ts` re-exports that barrel with `export *` — so the hook is back on the package's public surface. That undoes the restriction described in [an earlier reply](https://github.com/lablup/backend.ai-webui/pull/6668#discussion_r3287825238), where it was pulled off the public surface specifically to stop host components from reaching for it instead of `useResourceSlotsDetails`.

Keeping it public, for consistency: `useBAIImageMetaData` sits on the same surface, and item 4 asked for exactly that parity. The BUI-only rule stays a convention rather than a module-system guarantee — the three package call sites now import from the barrel, and no host component uses it.

That convention is load-bearing, so to be concrete about what it forbids: a host component reading slot metadata should call `useResourceSlotsDetails` from `src/hooks/backendai`, not `useBAIResourceSlots`. The package hook reads from context, so calling it from the host both inverts the fetch/read split this PR sets up and makes the host component depend on `BAIResourceSlotsProvider` being mounted above it. I had a `SessionLauncherPage` call site briefly do this while tidying up a duplicate query; it is reverted.

Happy to switch `components/index.ts` to an explicit export list instead, if enforcing the boundary in the module system is worth the maintenance.

yomybaby (approved):
LGTM

ironAiken2 (approved):
LGTM

PR #6759, opened 2026-04-16
feat(FR-2593): implement VFolder creation modal using V2 GraphQL mutation

### LINKED ISSUE #6755: Implement VFolder creation modal using Strawberry V2 GraphQL API
## Overview

Create a new VFolder creation modal component that uses the Strawberry V2 GraphQL mutation instead of the current REST API (`POST /folders`).

## Current Implementation

- **File**: `react/src/components/FolderCreateModal.tsx`
- Uses `baiRequestWithPromise` (REST API) to create folders via `POST /folders`
- Uses `useTanQuery` for fetching allowed types (`baiClient.vfolder.list_allowed_types()`)

## Requirements

- Create a new V2 modal component file alongside the existing one (per epic guidelines: do not modify existing files)
- Replace REST API call with Strawberry V2 GraphQL mutation for folder creation
- Replace `list_allowed_types()` with equivalent V2 GraphQL query
- Maintain the same form fields and validation logic:
  - Usage mode (general / model / automount)
  - Folder name (with dot-prefix rules for automount)
  - Storage location (host)
  - Type (user / project)
  - Permission (read-write / read-only)
  - Cloneable toggle (for model folders)
- Maintain backward compatibility with the legacy API through 26.4.x

## Acceptance Criteria

- [ ] New V2 component created (e.g., `FolderCreateModalV2.tsx`)
- [ ] Folder creation uses V2 GraphQL mutation
- [ ] Allowed types fetched via V2 GraphQL query
- [ ] All existing form validations preserved
- [ ] Legacy component remains untouched

JIRA Issue: FR-2593

ironAiken2:
Resolves #6755 (FR-2593)

## Summary

Introduces `FolderCreateModalV2`, a rewrite of the VFolder creation modal on the V2 GraphQL `createVfolderV2` mutation. The bigger structural change in this PR is **moving caller-intent (user vs. admin context) from implicit role inference into explicit, declarative props**, and **replacing the open-ended `hiddenFormItems` escape hatch with a typed `folderType` preset**.

## User / admin separation

Previously the original `FolderCreateModal` decided whether to expose the project-folder option purely by reading the current user's role:

```tsx
{(userRole === 'admin' || userRole === 'superadmin') && _.includes(allowedTypes, 'group') && ...}
```

This leaks admin UI when a superadmin user opens the same modal from a user-facing page (e.g. their personal Data page) — they would see project folder creation that doesn't belong in that context.

`FolderCreateModalV2` flips this: the **page** declares its context, and the modal honors it.

### `allowCreateProjectFolder?: boolean` (default `false`)

Caller-controlled flag indicating the modal is being opened from an admin context where project folder creation should be exposed.

- User-facing pages omit it → project option is never shown, even for superadmins.
- Admin pages opt in by passing `allowCreateProjectFolder` → project option becomes available.
- The user role check (`admin`/`superadmin`) is **kept as defense-in-depth**, so a non-admin who somehow lands on an admin context (route-level permission misconfig) still cannot see the option.

Effective condition for showing the project option:

```
allowCreateProjectFolder && isAdmin && allowedTypes.includes('group')
```

### `folderType?: 'model_project'`

Replaces the prior pattern of `initialValues={...} + hiddenFormItems=[...]` for "create a folder of a fixed shape" call sites. When set:

- Pre-selects `usage_mode='model'`, `type='project'`, `permission='ro'`, `cloneable=true` (`cloneable` remains user-editable).
- Disables the structural radio groups so the user can see they are locked but cannot change them.
- Suppresses the existing warning tooltips/icons on disabled options for a clean read-only appearance.
- Caller still controls `name` and `host`.

`folderType` is intentionally a single-value union for now — extend it (e.g. `'model_user'`) when the next preset is needed instead of adding more boolean flags.

### Removal of `hiddenFormItems`

The previous `hiddenFormItems: HiddenFormItemsType[]` prop is **removed**. It was a leaky abstraction: callers had to know the modal's internal field names and individually opt out of each option. All real uses were really one of two intents — "this is an admin context" or "this is a model_project preset" — both now expressed declaratively.

## Caller updates

| Caller | Before | After |
|---|---|---|
| `AdminModelCardSettingModal` | `initialValues={{ usage_mode: 'model', type: 'project', permission: 'ro', cloneable: true }}` + 4 `hiddenFormItems` entries | `folderType=\"model_project\"` |
| `ImportArtifactRevisionToFolderModal` | Same as above | `folderType=\"model_project\"` |
| `AdminVFolderNodeListPage` | Create Folder button was a TODO | Wired with `allowCreateProjectFolder` |
| `VFolderMountFormItem` | Used `hiddenFormItems` | Currently unused component; `hiddenFormItems` block left as a TODO comment |

Other callers (`StartPage`, `VFolderNodeListPage`, `VFolderTable`, `VFolderSelect`) only used `initialValues` and are unaffected.

## V2 mutation

Folder creation now uses the `createVfolderV2` mutation. The response is mapped back to the existing `FolderCreationResponse` shape so downstream code (notification, event dispatch, callers consuming `result.id`) is unchanged. The REST `list_allowed_types` call is retained as there is no V2 equivalent yet.

- [ ] Documentation
- [ ] Minimum required manager version — verify `createVfolderV2` mutation availability
  - User-facing page (e.g. user Data page): superadmin should **not** see the project type option.
  - Admin page (`AdminVFolderNodeListPage`): admin should see the project type option via the new Create Folder button.
  - Admin Model Card / Import Artifact: opening the create-folder dialog should show the form locked to model+project+ro, with cloneable still toggleable.

### DIFF
--- a/react/src/components/FolderCreateModalV2.tsx
+++ b/react/src/components/FolderCreateModalV2.tsx
@@ -1,0 +1,638 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+import { FolderCreateModalV2Mutation } from '../__generated__/FolderCreateModalV2Mutation.graphql';
+import { useSuspendedBackendaiClient } from '../hooks';
+import { useCurrentUserRole } from '../hooks/backendai';
+import { useTanQuery } from '../hooks/reactQueryAlias';
+import { useSetBAINotification } from '../hooks/useBAINotification';
+import { useCurrentProjectValue } from '../hooks/useCurrentProject';
+import QuestionIconWithTooltip from './QuestionIconWithTooltip';
+import StorageSelect from './StorageSelect';
+import {
+  App,
+  Divider,
+  Form,
+  Input,
+  Radio,
+  Skeleton,
+  Switch,
+  theme,
+  Tooltip,
+} from 'antd';
+import { createStyles } from 'antd-style';
+import { FormInstance } from 'antd/lib';
+import {
+  BAIButton,
+  BAIFlex,
+  BAIModal,
+  BAIModalProps,
+  toLocalId,
+  useBAILogger,
+  useErrorMessageResolver,
+  useMutationWithPromise,
+} from 'backend.ai-ui';
+import * as _ from 'lodash-es';
+import { TriangleAlertIcon } from 'lucide-react';
+import { Suspense, useRef } from 'react';
+import { useTranslation } from 'react-i18next';
+import { graphql } from 'react-relay';
+
+// Constants
+const MODEL_STORE_PROJECT_NAME = 'model-store';
+const FOLDER_NAME_MAX_LENGTH = 64;
+const MODAL_WIDTH = 650;
+
+const useStyles = createStyles(({ css }) => ({
+  modal: css`
+    .ant-modal-body {
+      padding-top: 24px !important;
+      padding-bottom: 0 !important;
+    }
+  `,
+  form: css`
+    .ant-form-item-label {
+      display: flex;
+      align-items: start;
+      padding-left: var(--token-paddingSM);
+    }
+    .ant-form-item-control {
+      padding-right: var(--token-paddingSM);
+    }
+    .ant-form-item-label > label::after {
+      display: none !important;
+    }
+  `,
+}));
+
+interface FolderCreateFormItemsType {
+  name: string;
+  host: string | undefined;
+  group: string | undefined;
+  usage_mode: 'general' | 'model' | 'automount';
+  type: 'user' | 'project';
+  permission: 'rw' | 'ro';
+  cloneable: boolean;
+}
+
+export interface FolderCreateModalProps extends BAIModalProps {
+  onRequestClose: (response?: FolderCreationResponse) => void;
+  initialValidate?: boolean;
+  initialValues?: Partial<FolderCreateFormItemsType>;
+  /**
+   * Whether to allow creating project-type folders. Should be enabled only on
+   * admin pages — leaving it off ensures project folder creation UI is not
+   * exposed when the same modal is used from user-facing pages, even if the
+   * current user has admin privileges.
+   */
+  allowCreateProjectFolder?: boolean;
+  /**
+   * When set, locks the form to a specific folder shape by pre-selecting and
+   * disabling the structural fields (usage_mode, type, permission). The user
+   * can still edit name, host, and cloneable. Used when a caller needs a folder
+   * with a fixed shape — e.g. 'model_project' for the Model Store admin flow.
+   */
+  folderType?: 'model_project';
+}
+
+export interface FolderCreationResponse {
+  id: string;
+  name: string;
+  quota_scope_id: string;
+  host: string;
+  usage_mode: 'general' | 'model' | 'automount';
+  permission: 'rw' | 'ro';
+  creator: string;
+  ownership_type: 'user' | 'project';
+  user: string;
+  group: string | null;
+  cloneable: boolean;
+  status: string;
+}
+
+const FolderCreateModalV2: React.FC<FolderCreateModalProps> = ({
+  onRequestClose,
+  initialValidate = false,
+  initialValues: initialValuesFromProps = {},
+  allowCreateProjectFolder = false,
+  folderType,
+  ...modalProps
+}) => {
+  'use memo';
+  const { t } = useTranslation();
+  const { styles } = useStyles();
+  const { token } = theme.useToken();
+  const { message } = App.useApp();
+  const { logger } = useBAILogger();
+  const { getErrorMessage } = useErrorMessageResolver();
+
+  const formRef = useRef<FormInstance>(null);
+  const baiClient = useSuspendedBackendaiClient();
+  const userRole = useCurrentUserRole();
+  const currentProject = useCurrentProjectValue();
+
+  const { upsertNotification } = useSetBAINotification();
+
+  const INITIAL_FORM_VALUES: FolderCreateFormItemsType = {
+    name: '',
+    host: undefined,
+    group: currentProject.id || undefined,
+    usage_mode: 'general',
+    type: 'user',
+    permission: 'rw',
+    cloneable: false,
+  };
+
+  const isFolderTypeLocked = folderType !== undefined;
+
+  // When folderType locks the form to a specific shape, these preset values
+  // override any user-passed initialValues for the locked fields.
+  const folderTypePreset: Partial<FolderCreateFormItemsType> | undefined =
+    folderType === 'model_project'
+      ? {
+          usage_mode: 'model',
+          type: 'project',
+          permission: 'ro',
+          cloneable: true,
+        }
+      : undefined;
+
+  const mergedInitialValues: FolderCreateFormItemsType = {
+    ...INITIAL_FORM_VALUES,
+    ...initialValuesFromProps,
+    ...folderTypePreset,
+  };
+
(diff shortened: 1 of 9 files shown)


ironAiken2:
* **#6933** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6933?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6827** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6827?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6842** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6842?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6934** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6934?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6962** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6962?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6941** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6941?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6817** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6817?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6759** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6759?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/6759?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=6759).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/FolderCreateModalV2.tsx
```diff
@@ -0,0 +1,658 @@
...
+                usage_mode: isAutomount
+                  ? 'automount'
+                  : vfolder.metadata.usageMode === 'MODEL'
+                    ? 'model'
+                    : 'general',
+                permission:
+                  vfolder.accessControl.permission === 'READ_ONLY'
+                    ? 'ro'
+                    : 'rw',
+                max_size: 0,
```
agatha197: `max_size` is hardcoded to `0` because the V2 mutation doesn't return this field. Please verify whether any downstream consumer of `FolderCreationResponse.max_size` relies on the actual value. If the field is unused, consider making it optional (`max_size?: number`) or removing it from the interface to avoid silently propagating a wrong value.

react/src/components/FolderCreateModalV2.tsx
```diff
@@ -0,0 +1,658 @@
...
+          {({ getFieldValue }) => {
+            return (
+              getFieldValue('usage_mode') === 'model' && (
+                <>
+                  <Divider />
+                  <Form.Item
+                    label={t('data.folders.Cloneable')}
+                    name={'cloneable'}
+                  >
+                    <Switch defaultChecked={false} />
```
agatha197: `defaultChecked={false}` is redundant here. `Form.Item` controls the Switch via `initialValues.cloneable = false`, so `defaultChecked` is ignored. Remove it to keep the controlled/uncontrolled boundary clean.

react/src/components/VFolderMountFormItem.tsx
```diff
@@ -302,13 +302,15 @@ const VFolderMountFormItem: React.FC<VFolderMountFormItemProps> = ({
...
-          hiddenFormItems={[
-            'usage_mode',
-            'usage_mode_model',
-            'usage_mode_automount',
-          ]}
+          // TODO: hiddenFormItems prop was removed from FolderCreateModalV2.
+          // This component is currently unused; revisit if it gets re-enabled.
+          // hiddenFormItems={[
+          //   'usage_mode',
+          //   'usage_mode_model',
```
agatha197, line 311: The commented-out `hiddenFormItems` block leaves this modal without the usage-mode restriction that the old `FolderCreateModal` enforced. With only `initialValues={{ usage_mode: 'general' }}`, the user can still switch to `model` or `automount` — which was not the intent. If this call site is truly unused right now, it's safer to keep `FolderCreateModal` here until V2 supports restricting visible options, and track the migration in a follow-up issue.

react/src/pages/AdminVFolderNodeListPage.tsx
```diff
@@ -441,6 +444,15 @@ const AdminVFolderNodeListPage: React.FC = (props) => {
                   updateFetchKey(newFetchKey);
                 }}
               />
+              <Button
```
agatha197: Prefer `BAIButton` over Ant Design's `Button` for consistency with the rest of the codebase.

react/src/components/FolderCreateModalV2.tsx
ironAiken2: Good point. Since `max_size` isn't returned by the V2 mutation and no downstream consumer reads `response.max_size` (grep confirmed), I've removed the field from the `FolderCreationResponse` interface entirely rather than keeping a misleading hardcoded `0`.

react/src/components/FolderCreateModalV2.tsx
ironAiken2: Good catch. Removed `defaultChecked={false}` in the latest commit.

react/src/components/VFolderMountFormItem.tsx
ironAiken2: Thanks for flagging this. This call site in `VFolderMountFormItem` is currently unused — the inline TODO comment documents the removed `hiddenFormItems` behavior so it can be restored when V2 gains support for restricting visible options. I'll track adding that capability to `FolderCreateModalV2` as a follow-up; until then the file keeps the usage-mode restriction context visible for future re-enablement.

react/src/pages/AdminVFolderNodeListPage.tsx
ironAiken2: Updated — replaced `Button` with `BAIButton` on both call sites in this file (the primary "Create Folder" button and the "Restore" button) to align with the rest of the codebase.

agatha197 (approved):
LGTM

react/src/components/FolderCreateModalV2.tsx
```diff
@@ -0,0 +1,656 @@
...
+          name: folderName,
+          host: values.host ?? null,
+          usageMode: isAutomount ? 'general' : values.usage_mode,
+          permission: values.permission,
+          cloneable: !!values.cloneable,
+          // If type is 'project', send the current project ID; otherwise null for user-owned
+          projectId: values.type === 'project' ? values.group : null,
+        };
+
+        return new Promise<void>((resolve, reject) => {
```
nowgnuesLee: you can use useMutationWithPromise

nowgnuesLee (changes requested):

react/src/components/FolderCreateModalV2.tsx
ironAiken2: Good suggestion. Replaced `useMutation` with `useMutationWithPromise` and refactored `handleOk` to use async/await directly, eliminating the manual `new Promise()` wrapping. Also dropped the now-unused `isInFlight` state since `BAIButton.action` handles loading via `useTransition`.

nowgnuesLee (approved):
LGTM

yomybaby (approved):

PR #7262, opened 2026-05-06
fix(FR-2823): migrate Environment > Images table to server-side ordering

### LINKED ISSUE #7257: Environment > Images table — switch from client-side sorters to server-side ordering (image_nodes already supports $order)
## Summary

The Images tab on `/environment` (`?tab=image`) sorts the table entirely in the browser via per-column `sorter: (a, b) => ...` callbacks. Because the table is paginated server-side, these client sorters can only reorder the rows on the **current page** — they cannot produce a globally-sorted view, so column header clicks give the user a misleading partial sort.

The backing GraphQL query (`image_nodes`) already exposes an `$order: String` argument, but the page hard-codes `order: undefined` and never wires the tables sort state through to the server. This should be migrated to server-side ordering so column headers produce a true global sort.

## Where (frontend pointer)

`react/src/components/ImageList.tsx`:

- Query variables (around lines 80-86): `order: undefined` is hard-coded; the `$order: String` argument exists in the operation declaration (lines 97, 104) and is therefore wasted.
- Column definitions (around lines 155-229): every sortable column declares an in-memory comparator, e.g.

```ts

{ key: "installed", defaultSortOrder: "descend", sorter: (a, b) => ... }

{ key: "name",      sorter: (a, b) => localeCompare(getImageFullName(a), getImageFullName(b)) }

// ...registry / architecture / namespace / base_image_name / version / digest

```

These should be reduced to `sorter: true` (or the explicit BAITable equivalent) so the table emits a sort change but does not attempt to sort locally.

- BAITable rendering (around line 526): no `onChangeOrder` / `order` is plumbed through. The component currently has no way to inform the parent of a sort change.

## Suggested fix

1. **Confirm what the server expects.** `image_nodes(order: $order)` is a `String`. Inspect the schema / how other queries in this codebase encode multi-column order to ensure the right format is used (e.g., comma-separated `"-installed_at,name"` style). The `/project` page already plumbs server-side ordering through `BAIProjectTable` and is a good reference for both the URL-state plumbing and the encoding.
1. **Add URL-persisted sort state** in `EnvironmentPage` / `ImageList` using `nuqs` (`useQueryStates` + `parseAsString`) keyed `order`, mirroring how `ProjectPage` and other listing pages do it.
1. **Pass the order through to the query**: replace `order: undefined` with `order: queryParams.order` (and add `order` to the deferred-value dependency so a refetch is triggered when it changes).
1. **Strip the in-memory comparators**: change each per-column `sorter: (a, b) => ...` to the BAITable server-sort signal (`sorter: true` or the projects standard equivalent), and translate the tables `onChange` sort payload into the encoded `order` string before writing it back to URL state.
1. **Verify column ↔ field mapping**: only declare a column sortable on the client side if the corresponding GraphQL field is sortable on the server. If `image_nodes` does not accept ordering on a given field (e.g., `humanized_name` is computed), drop the sortable affordance for that column rather than silently ignoring the click.
1. **Confirm pagination + sort interaction**: changing the sort should reset to page 1 (matches existing pattern on other listing pages).
1. Run `pnpm run relay` (no schema change expected) and `bash scripts/verify.sh`.

## Steps to reproduce (current behavior)

1. Open `/environment?tab=image`.
1. Click any sortable column header — e.g., `Installed`, `Registry`, `Version`.
1. Observe that only the rows on the current page reorder; navigating to page 2 / 3 reveals a different ordering, proving the sort is not global.

## Expected

Column header clicks issue a server-side ordering request via `image_nodes(order: ...)`, the URL reflects the new order, and the rows are globally sorted across all pages.

## Environment

- Test server: `10.122.10.97`
- Core: `26.4.4rc2`
- Release train: FR-2811 (26.5.0)

## Notes

- This is one specific instance of a broader anti-pattern. While migrating, it is worth a quick grep for other paginated `BAITable` consumers that still use in-memory `sorter: (a, b) => ...` and capture them as follow-up work (out of scope for this issue).
- Reference pattern (URL-state-driven server order) lives in `react/src/pages/ProjectPage.tsx` (`order: queryParams.order` + `onChangeOrder`).

JIRA Issue: FR-2823

ironAiken2:
Resolves #7257(FR-2823)

## Summary

The Images tab on `/environment?tab=image` paginates server-side but sorted client-side via per-column `sorter: (a, b) => ...` callbacks, which can only reorder rows on the current page — column header clicks gave a misleading partial sort. The backing `image_nodes` GraphQL query already exposes `$order: String`, but the page hard-coded `order: undefined` and never plumbed the table's sort state through.

## Changes

- **URL-persisted sort state**: add `useQueryStates({ order })` from `nuqs`. No default order is set; the server returns its natural order until the user picks a sortable column.
- **Pass `order` through to GraphQL**: replace `order: undefined` with `order: queryParams.order || undefined` so the server actually receives the sort key.
- **Server-orderable columns only**: the `image_nodes` ordering parser only accepts a fixed set of fields (`id`, `name`, `namespace`, `tag`, `status`, `project`, `image`, `base_image_name`, `created_at`, `registry`, `registry_id`, `architecture`, `is_local`, `type`, `accelerators`). Of the columns currently rendered, only **Registry**, **Architecture**, **Namespace**, and **BaseImageName** map to supported fields and get `sorter: true`. Status (`installed`), Version, Digest, and FullImagePath are not server-orderable, so the sort affordance is dropped on those columns.
- **Wire BAITable**: pass `order={queryParams.order}` and `onChangeOrder={...}`. The handler updates URL state and resets pagination to page 1 (matching `ProjectPage`'s convention).
- **Loading indicator**: switch the table's `loading` prop from `isPendingFilterTransition` (which only tracked filter transitions) to the deferred-value diff pattern used in `ProjectPage` — `deferredFetchKey !== fetchKey || deferredQueryVariables !== queryVariables`. This shows the spinner during filter, order, page, and refresh transitions uniformly.
- Remove now-unused `localeCompare` import.

## Pattern reference

Follows the URL-state-driven server-ordering pattern already used in `react/src/pages/ProjectPage.tsx` (`order: queryParams.order` + `onChangeOrder` + deferred-value loading).

## Verification

- Relay: PASS (no schema change)
- Lint: PASS
- Format: PASS
- TypeScript: pre-existing failures only on `main` (unrelated; tracked separately under FR-2816).

## Manual smoke test

- Open `/environment?tab=image`.
- Click any sortable column header (Registry, Architecture, Namespace, BaseImageName).
- The URL `?order=` reflects the new key (e.g., `registry`, `-namespace`).
- Loading spinner appears on the table during the order/filter/page transition.
- Navigate between pages — rows are now globally sorted across all pages, not partially per page.
- Status, Version, Digest, and FullImagePath column headers are no longer clickable to sort (these are not exposed as orderable fields by the backend).

## Notes

- Out of scope: other paginated `BAITable` consumers in the codebase still use in-memory comparators. The issue description acknowledges this and recommends capturing them as follow-up work.

### DIFF
--- a/react/src/components/ImageList.tsx
+++ b/react/src/components/ImageList.tsx
@@ -7,7 +7,7 @@ import {
   ImageListQuery$data,
   ImageListQuery$variables,
 } from '../__generated__/ImageListQuery.graphql';
-import { getImageFullName, localeCompare } from '../helper';
+import { getImageFullName } from '../helper';
 import { useBackendAIImageMetaData } from '../hooks';
 import { useBAIPaginationOptionStateOnSearchParam } from '../hooks/reactPaginationQueryOptions';
 import { useCurrentProjectValue } from '../hooks/useCurrentProject';
@@ -38,6 +38,7 @@ import {
   INITIAL_FETCH_KEY,
 } from 'backend.ai-ui';
 import * as _ from 'lodash-es';
+import { parseAsStringLiteral, useQueryStates } from 'nuqs';
 import { Key, useDeferredValue, useState, useTransition } from 'react';
 import { useTranslation } from 'react-i18next';
 import { graphql, useLazyLoadQuery } from 'react-relay';
@@ -45,6 +46,19 @@ import { graphql, useLazyLoadQuery } from 'react-relay';
 export type EnvironmentImage = NonNullableNodeOnEdges<
   ImageListQuery$data['image_nodes']
 >;
+
+const availableImageSorterKeys = [
+  'registry',
+  'architecture',
+  'namespace',
+  'base_image_name',
+] as const;
+const availableImageSorterValues = [
+  ...availableImageSorterKeys,
+  ...availableImageSorterKeys.map((key) => `-${key}` as const),
+] as const;
+const isEnableSorter = (key: string) =>
+  _.includes(availableImageSorterKeys, key);
 
 const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
   'use memo';
@@ -65,7 +79,6 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
   const [visibleColumnSettingModal, { toggle: toggleColumnSettingModal }] =
     useToggle();
   const [isPendingRefreshTransition, startRefreshTransition] = useTransition();
-  const [isPendingFilterTransition, startFilterTransition] = useTransition();
   const currentProject = useCurrentProjectValue();
 
   const {
@@ -77,12 +90,19 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
     pageSize: 20,
   });
 
+  const [queryParams, setQueryParams] = useQueryStates(
+    {
+      order: parseAsStringLiteral(availableImageSorterValues),
+    },
+    { history: 'replace' },
+  );
+
   const queryVariables: ImageListQuery$variables = {
     scopeId: `project:${currentProject.id}`,
     offset: baiPaginationOption.offset,
     first: baiPaginationOption.first,
     filter: imageFilter || undefined,
-    order: undefined,
+    order: queryParams.order || undefined,
   };
   const deferredQueryVariables = useDeferredValue(queryVariables);
   const deferredFetchKey = useDeferredValue(fetchKey);
@@ -157,10 +177,6 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
       title: t('environment.Status'),
       dataIndex: 'installed',
       key: 'installed',
-      defaultSortOrder: 'descend',
-      sorter: (a, b) => {
-        return _.toNumber(a?.installed || 0) - _.toNumber(b?.installed || 0);
-      },
       render: (_text, row) =>
         row?.id && installingImages.includes(row.id) ? (
           <Tag color="gold">{t('environment.Installing')}</Tag>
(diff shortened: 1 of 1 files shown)


ironAiken2:
* **#7262** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7262?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7262?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7262).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/ImageList.tsx
```diff
@@ -77,12 +78,19 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
     pageSize: 20,
   });
 
+  const [queryParams, setQueryParams] = useQueryStates(
+    {
+      order: parseAsString.withDefault('-installed'),
+    },
+    { history: 'replace' },
+  );
```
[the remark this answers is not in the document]
ironAiken2: The description never contained the `#NNN (FR-MMM)` placeholder text or the unchecked checklist template — it has had `Resolves #7257(FR-2823)` and a Summary/Changes/Verification/Manual smoke test breakdown since the PR was opened. Refreshed the body anyway so it reflects the current amended state (no `-installed` default; only the four backend-supported columns expose `sorter: true`; loading indicator switched to the deferred-value pattern).

react/src/components/ImageList.tsx
```diff
@@ -458,7 +460,15 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
...
           )}
-          loading={isPendingFilterTransition}
+          loading={
+            deferredFetchKey !== fetchKey ||
+            deferredQueryVariables !== queryVariables
+          }
+          order={queryParams.order}
+          onChangeOrder={(order) => {
+            setQueryParams({ order: order ?? null });
+            setTablePaginationOption({ current: 1 });
```
nowgnuesLee, line 484: Shouldn't we set the page to 1 when changing the order?

react/src/components/ImageList.tsx
```diff
@@ -180,39 +184,38 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
...
-      sorter: (a, b) => localeCompare(getImageFullName(a), getImageFullName(b)),
+      // Computed (`getImageFullName`) — not orderable on the server.
       width: token.screenXS,
     },
     {
       title: t('environment.Registry'),
       dataIndex: 'registry',
       key: 'registry',
-      sorter: (a, b) => localeCompare(a?.registry, b?.registry),
+      sorter: true,
```
nowgnuesLee: Wouldn't it be better to manage this with `availableSorterKey`?

nowgnuesLee (changes requested):

react/src/components/ImageList.tsx
ironAiken2: Resetting the page to 1 on order change isn't applied in our other table components either. I'll keep the current behavior here for consistency and address the page-reset-on-sort behavior across tables in a separate stack after we discuss the desired UX.

react/src/components/ImageList.tsx
ironAiken2: Good suggestion — applied in the latest commit. Introduced `availableImageSorterKeys` / `availableImageSorterValues` and an `isEnableSorter` helper, narrowed the URL `order` param via `parseAsStringLiteral(availableImageSorterValues)`, and gated each `sorter: true` column through `isEnableSorter('<key>')` so the sortable keys are a single source of truth (matching the SessionNodes / DeploymentReplicasTab pattern).

react/src/components/ImageList.tsx
```diff
@@ -65,7 +79,7 @@ const ImageList: React.FC<{ style?: React.CSSProperties }> = ({ style }) => {
   const [visibleColumnSettingModal, { toggle: toggleColumnSettingModal }] =
     useToggle();
   const [isPendingRefreshTransition, startRefreshTransition] = useTransition();
-  const [isPendingFilterTransition, startFilterTransition] = useTransition();
+  const [, startFilterTransition] = useTransition();
```
nowgnuesLee: I think we could remove this transition.

nowgnuesLee (changes requested):

nowgnuesLee (approved):
LGTM

PR #7315, opened 2026-05-08
fix(FR-2832): send null for definitionPath when not customized

### LINKED ISSUE #7283: Revision creation hardcodes "model-definition.yaml"; should send null to allow both .yaml and .yml
## Summary

When creating a deployment revision, the WebUI sends the model-definition filename as the literal string `model-definition.yaml`. This forces the backend to look for that exact filename and breaks model folders that use `model-definition.yml`.

The frontend should send the field as `null` (omit / no override) when the user has not explicitly customized the filename, so the backend can fall back to its own resolution logic and accept either `model-definition.yaml` or `model-definition.yml`.

## Steps to reproduce

1. Open "Add revision" for an existing model deployment whose model folder contains `model-definition.yml` (not `.yaml`).
1. Submit the form without changing the model-definition filename field.
1. The mutation payload sets the model-definition filename to `"model-definition.yaml"`, and the revision fails / loads the wrong file.

## Expected

When the user does not customize the model-definition filename, the WebUI sends `null` (or omits the field) so the backend can resolve the file using both supported names (`model-definition.yaml` and `model-definition.yml`).

## Actual

The WebUI always sends `"model-definition.yaml"` as the default value, hardcoding the extension and breaking folders that ship the file as `.yml`.

## Fix

- In the revision-creation form, treat the model-definition filename input as optional. When the user has not entered a custom value, submit `null` (or omit the field) instead of defaulting the payload to `"model-definition.yaml"`.
- Coordinate with the backend so it accepts `null` and resolves both `.yaml` and `.yml` extensions on the server side.

## Related

- Blocks: FR-2811 (26.5.0 deployment UX)

—

Captured while working on branch: (detached HEAD on backend.ai-webui)

JIRA Issue: FR-2832

ironAiken2:
Resolves #7283 (FR-2832)

## Summary

When creating a deployment revision, the WebUI was sending the literal string `"model-definition.yaml"` as `modelMountConfig.definitionPath`. This forced the backend to look for that exact filename and broke model folders that use `model-definition.yml`. After this change, the frontend **omits** `definitionPath` when the user has not explicitly customized the filename, so the backend can fall back to its own resolution logic and accept either extension.

## Changes (single file: `react/src/components/DeploymentAddRevisionModal.tsx`)

1. **`initialValues`** — dropped the `definitionPath: 'model-definition.yaml'` seed. The form field now starts empty; the existing `placeholder="model-definition.yaml"` on the Input still hints the default.
2. **Submit handler** — for the non-custom / non-command-mode branch, **omit** `definitionPath` from the `modelMountConfig` payload when the trimmed user input is empty. The `isCustom && isCommandMode` branch still sends the literal `'model-definition.yaml'` because we generate that file ourselves before submission.
3. **Edit-mode pre-population** — dropped the `?? 'model-definition.yaml'` fallback so when an existing revision had `definitionPath: null/empty`, the form leaves the field empty and a re-submit continues to omit the field.
4. **Required rule removed** — the `definitionPath` Form.Item no longer has `rules={[{ required: true }]}` so users can submit an empty value (which is now the explicit "use server default" signal).

## Backend dependency

> **Requires backend to accept an absent / nullable `modelMountConfig.definitionPath` and resolve both `.yaml`/`.yml` server-side.**

The current GraphQL schema declares `ModelMountConfigInput.definitionPath: String!` (non-null). The submit handler uses a conditional spread to omit the field when not customized; this works at runtime because most GraphQL backends treat an absent input field as "use default." A small `as { definitionPath: string }` cast on the spread is needed to satisfy the current strict type — flagged with `TODO(needs-backend): FR-2832`. The backend must relax the schema field to nullable and update the resolver to try `model-definition.yaml` then `model-definition.yml` (or vice versa) when this field is absent.

## Out of scope

Other call sites of `'model-definition.yaml'` (`ServiceLauncherPageContent.tsx`, `LegacyModelTryContentButton.tsx`, `DeploymentLauncherPageContent.tsx`, `VFolderTextFileEditorModal.tsx`, `DeploymentLauncherPage.tsx`) are intentionally untouched per the issue triage. A follow-up issue tracks applying the same pattern to the launcher components.

## Verification

`bash scripts/verify.sh`:

- Relay: PASS
- Lint: PASS
- Format: PASS
- TypeScript: only **pre-existing** errors remain (`packages/backend.ai-client/src/client.ts`, `src/components/DeleteForeverVFolderModalV2.tsx`) — confirmed identical against `main` before this change.

## Test plan

- [ ] Create a new deployment revision **without** typing anything in the "Model Definition Path" field → mutation payload should **omit** `modelMountConfig.definitionPath`.
- [ ] Create a new deployment revision with **whitespace only** (e.g. `"   "`) → mutation payload should **omit** `modelMountConfig.definitionPath` (whitespace is trimmed).
- [ ] Create a new deployment revision **with** an explicit value (e.g. `custom.yaml`) → mutation payload should send `modelMountConfig.definitionPath: "custom.yaml"`.
- [ ] Edit an existing revision whose previous `definitionPath` is empty/null → form field stays empty, re-submit keeps omitting the field.
- [ ] Edit an existing revision whose previous `definitionPath` is `"model-definition.yaml"` → form field shows that value, re-submit sends it back unchanged.
- [ ] Custom variant + command mode → literal `"model-definition.yaml"` is still sent (since we generate that file ourselves).
- [ ] Once backend lands its companion change, verify that a model folder containing only `model-definition.yml` deploys successfully via this UI.

## Review feedback applied

- **Replaced `as string` cast with conditional spread.** The submit handler no longer lies to TypeScript by asserting `null` is a `string`. Instead, the field is omitted entirely from the input object when the user has not customized it (`...(resolvedDefinitionPath ? { definitionPath: resolvedDefinitionPath } : {})`). A narrower `as { definitionPath: string }` cast on the spread expression remains until the schema is relaxed to nullable.
- **Trim whitespace before truthiness check.** `values.definitionPath?.trim()` is computed once into `trimmedDefinitionPath`, so whitespace-only inputs (`"   "`) collapse to "not customized" and the backend default applies.
- **Comment consolidation.** The two multi-paragraph FR-2832 comments now share a single `TODO(needs-backend)` block at the spread site; the resolver site has a one-line back-reference.

### DIFF
--- a/react/src/components/DeploymentAddRevisionModal.tsx
+++ b/react/src/components/DeploymentAddRevisionModal.tsx
@@ -502,8 +502,11 @@ const DeploymentAddRevisionModalFormBody: React.FC<
         ? rev.modelMountConfig.vfolderId.replace(/-/g, '')
         : undefined,
       mountDestination: rev.modelMountConfig?.mountDestination ?? '/models',
-      definitionPath:
-        rev.modelMountConfig?.definitionPath ?? 'model-definition.yaml',
+      // FR-2832: don't fall back to `'model-definition.yaml'` here — when
+      // the previous revision left `definitionPath` empty/null, keep the
+      // form field empty so a re-submit continues to send `null` instead
+      // of re-introducing the hardcoded default.
+      definitionPath: rev.modelMountConfig?.definitionPath,
       // `ImageEnvironmentSelectFormItems` matches the form's
       // `environments.version` against its image catalog by canonical
       // name; setting this is enough to drive the rest of the
@@ -771,10 +774,15 @@ const DeploymentAddRevisionModalFormBody: React.FC<
       isCustom && isCommandMode
         ? (values.commandModelMount ?? '/models')
         : values.mountDestination || '/models';
-    const definitionPath =
+    // FR-2832: see TODO at the modelMountConfig spread below for backend
+    // coordination details. Trim user input so whitespace-only values
+    // (`'   '`) collapse to "not customized" and let the backend pick
+    // the default filename.
+    const trimmedDefinitionPath = values.definitionPath?.trim();
+    const resolvedDefinitionPath =
       isCustom && isCommandMode
         ? 'model-definition.yaml'
-        : values.definitionPath || 'model-definition.yaml';
+        : trimmedDefinitionPath || null;
 
     onIsAddingChange(true);
     commitAdd({
@@ -804,7 +812,12 @@ const DeploymentAddRevisionModalFormBody: React.FC<
             // normalize before submitting (same as `extraMounts`).
             vfolderId: convertToUUID(values.modelFolderId),
             mountDestination,
-            definitionPath,
+            // FR-2832: send `null` for `definitionPath` when the user
+            // hasn't customized it so the backend can resolve either
+            // `model-definition.yaml` or `model-definition.yml`.
+            // `ModelMountConfigInput.definitionPath` is nullable in the
+            // schema (default `null`).
+            definitionPath: resolvedDefinitionPath,
           },
           modelDefinition,
           extraMounts: extraMounts.length > 0 ? extraMounts : null,
@@ -888,7 +901,11 @@ const DeploymentAddRevisionModalFormBody: React.FC<
       onFinishFailed={handleFinishFailed}
       initialValues={_.merge({}, RESOURCE_ALLOCATION_INITIAL_FORM_VALUES, {
         mountDestination: '/models',
-        definitionPath: 'model-definition.yaml',
+        // FR-2832: don't seed `definitionPath` — leave it empty so the
+        // submit handler can send `null` and let the backend resolve
+        // either `model-definition.yaml` or `model-definition.yml`.
+        // The Input below shows `model-definition.yaml` as a placeholder
+        // so users still see the default hint.
         customDefinitionMode: 'command',
         commandPort: 8000,
         commandHealthCheck: '/health',
@@ -1083,7 +1100,6 @@ const DeploymentAddRevisionModalFormBody: React.FC<
                       <Form.Item
                         name="definitionPath"
                         label={t('deployment.ModelDefinitionPath')}
-                        rules={[{ required: true }]}
                         style={{ flex: 1 }}
                       >
                         <Input allowClear placeholder="model-definition.yaml" />
--- a/data/schema.graphql
+++ b/data/schema.graphql
@@ -9235,10 +9235,10 @@ input ModelConfigInput
   @join__type(graph: STRAWBERRY)
 {
   """Name of the model."""
-  name: String!
+  name: String = null
 
   """Path to the model file."""
-  modelPath: String!
+  modelPath: String = null
 
   """Configuration for the model service."""
   service: ModelServiceConfigInput = null
@@ -9264,7 +9264,7 @@ input ModelDefinitionInput
   @join__type(graph: STRAWBERRY)
 {
   """List of models in the model definition."""
-  models: [ModelConfigInput!]!
+  models: [ModelConfigInput!] = null
 }
 
 """
@@ -9431,22 +9431,22 @@ input ModelHealthCheckInput
   @join__type(graph: STRAWBERRY)
 {
   """Interval in seconds between health checks."""
-  interval: Float! = 10
+  interval: Float = null
 
   """Path to check for health status."""
-  path: String!
+  path: String = null
 
   """Maximum number of retries for health check."""
-  maxRetries: Int! = 10
+  maxRetries: Int = null
 
   """Maximum time in seconds to wait for a health check response."""
-  maxWaitTime: Float! = 15
+  maxWaitTime: Float = null
 
   """Expected HTTP status code for a healthy response."""
-  expectedStatusCode: Int! = 200
+  expectedStatusCode: Int = null
 
   """Initial delay in seconds before the first health check."""
-  initialDelay: Float! = 60
+  initialDelay: Float = null
 }
 
 """Added in 26.4.2. Metadata describing a model entry."""
@@ -9558,7 +9558,7 @@ input ModelMountConfigInput
 {
   vfolderId: ID!
   mountDestination: String!
-  definitionPath: String!
+  definitionPath: String = null
 }
 
 """
@@ -9808,16 +9808,16 @@ input ModelServiceConfigInput
   """
   List of pre-start actions to execute before starting the model service.
   """
-  preStartActions: [PreStartActionInput!]!
+  preStartActions: [PreStartActionInput!] = null
 
   """Command to start the model service."""
   startCommand: [String!] = null
 
   """Shell configured for the model service."""
-  shell: String! = "/bin/bash"
+  shell: String = null
 
-  """Port number for the model service. Must be greater than 1."""
-  port: Int!
+  """Port number for the model service."""
+  port: Int = null
 
   """Health check configuration for the model service."""
   healthCheck: ModelHealthCheckInput = null


ironAiken2:
* **#7315** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7315?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7315?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7315).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

ironAiken2:
Applied review feedback:

- Replaced `as string` cast with a conditional spread so we no longer lie to TypeScript about a `null` being a `string`. The handler now **omits** `definitionPath` from the `modelMountConfig` payload when the user has not customized it. A narrower `as { definitionPath: string }` cast on the spread expression remains until the schema is relaxed to nullable, with the `TODO(needs-backend): FR-2832` marker in place.
- Added a `.trim()` step before the truthiness check, so whitespace-only inputs (`'   '`) collapse to 'not customized' instead of being sent as a meaningless string.
- Consolidated the two FR-2832 comment blocks per L2 suggestion: the `TODO(needs-backend)` block now lives only at the modelMountConfig spread, and the resolver has a one-line back-reference.

The 3 sibling launcher files flagged in review (`ServiceLauncherPageContent.tsx`, `DeploymentLauncherPageContent.tsx`, `DeploymentLauncherPage.tsx`) are out of scope for this PR — tracked in a follow-up issue.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -771,10 +774,20 @@ const DeploymentAddRevisionModalFormBody: React.FC<
...
+    // generate that file ourselves above (see `new File([blob],
+    // 'model-definition.yaml', ...)`), so the backend must look for it
+    // by that exact name.
+    // TODO(needs-backend): FR-2832 — confirm backend resolves both .yaml
+    // and .yml when `modelMountConfig.definitionPath` is null.
     const definitionPath =
       isCustom && isCommandMode
         ? 'model-definition.yaml'
-        : values.definitionPath || 'model-definition.yaml';
+        : values.definitionPath || null;
```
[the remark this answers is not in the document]
ironAiken2: Addressed in a subsequent amendment. The submit handler now uses a **conditional spread** (`...(resolvedDefinitionPath ? { definitionPath: resolvedDefinitionPath } : {})`) instead of sending an explicit `null`, so the field is **omitted** from the input object when not customized. Most GraphQL backends — including Strawberry/pydantic which Backend.AI uses — treat an absent input field as "use default" rather than rejecting it as an empty `String!`. Backend coordination is still required to actually resolve both `.yaml` and `.yml` server-side, which is documented with a `TODO(needs-backend): FR-2832` marker. The PR description's "Backend dependency" section calls this out as a hard prerequisite for end-to-end behavior.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -805,7 +818,17 @@ const DeploymentAddRevisionModalFormBody: React.FC<
...
+            // we intentionally send `null` when the user hasn't
+            // customized the model definition filename so the backend
+            // can resolve either `model-definition.yaml` or
+            // `model-definition.yml`. The schema must be relaxed to
+            // `String` (nullable) once the backend lands the matching
+            // change.
+            // TODO(needs-backend): FR-2832 — relax
+            // `ModelMountConfigInput.definitionPath` to nullable and
+            // resolve `.yaml` / `.yml` server-side.
+            definitionPath: definitionPath as string,
```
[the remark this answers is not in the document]
ironAiken2: Narrowed in a subsequent amendment. The broad `as string` cast on a possibly-null value was replaced with a conditional spread that **omits** the field entirely when not customized. The remaining cast is `as { definitionPath: string }` on the spread shape — i.e., it asserts the shape of the spread expression rather than lying about a value being a string when it is null. The runtime no longer sends `null` over the wire; the field is absent. The cast can be removed once the schema relaxes `ModelMountConfigInput.definitionPath` to nullable (tracked by the `TODO(needs-backend): FR-2832` markers).

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -502,8 +502,11 @@ const DeploymentAddRevisionModalFormBody: React.FC<
...
         : undefined,
       mountDestination: rev.modelMountConfig?.mountDestination ?? '/models',
-      definitionPath:
-        rev.modelMountConfig?.definitionPath ?? 'model-definition.yaml',
+      // FR-2832: don't fall back to `'model-definition.yaml'` here — when
+      // the previous revision left `definitionPath` empty/null, keep the
+      // form field empty so a re-submit continues to send `null` instead
+      // of re-introducing the hardcoded default.
+      definitionPath: rev.modelMountConfig?.definitionPath ?? '',
       // `ImageEnvironmentSelectFormItems` matches the form's
```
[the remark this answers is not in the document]
ironAiken2: The PR description has been updated and contains the real issue/Jira references (`Resolves #7283 (FR-2832)`), a detailed Summary, Changes, Backend dependency, Out-of-scope, Verification, Test plan, and a Review-feedback-applied section. No template placeholders remain. This thread was opened on an earlier revision before the description was completed.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -502,8 +502,11 @@ const DeploymentAddRevisionModalFormBody: React.FC<
         ? rev.modelMountConfig.vfolderId.replace(/-/g, '')
         : undefined,
       mountDestination: rev.modelMountConfig?.mountDestination ?? '/models',
-      definitionPath:
-        rev.modelMountConfig?.definitionPath ?? 'model-definition.yaml',
+      // FR-2832: don't fall back to `'model-definition.yaml'` here — when
+      // the previous revision left `definitionPath` empty/null, keep the
+      // form field empty so a re-submit continues to send `null` instead
+      // of re-introducing the hardcoded default.
+      definitionPath: rev.modelMountConfig?.definitionPath ?? '',
```
nowgnuesLee:
```suggestion
      definitionPath: rev.modelMountConfig?.definitionPath,
```

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -804,7 +812,19 @@ const DeploymentAddRevisionModalFormBody: React.FC<
...
+            // The cast is needed because the current schema marks
+            // `definitionPath` as required; omission still works at
+            // runtime because most GraphQL backends treat an absent
+            // input field as "use default."
+            // TODO(needs-backend): FR-2832 — relax
+            // `ModelMountConfigInput.definitionPath` to nullable in the
+            // schema, then drop the cast.
+            ...((resolvedDefinitionPath
+              ? { definitionPath: resolvedDefinitionPath }
+              : {}) as { definitionPath: string }),
```
nowgnuesLee:
```suggestion
            definitionPath: resolvedDefinitionPath,
```

nowgnuesLee (changes requested):

react/src/components/DeploymentAddRevisionModal.tsx
ironAiken2: Good catch — applied as-is. Removed the `?? ''` fallback so the form value stays as the raw `string | undefined` from the query.

react/src/components/DeploymentAddRevisionModal.tsx
ironAiken2: Applied. Confirmed the schema is already nullable (`ModelMountConfigInput.definitionPath: String = null` in `data/schema.graphql:9556`) and the generated mutation type is `string | null | undefined`, so the conditional spread + cast workaround was unnecessary. Reverted to a direct `definitionPath: resolvedDefinitionPath` assignment and updated the comment to reflect the actual schema state.

nowgnuesLee (approved):
LGTM

PR #7455, opened 2026-05-18
feat(FR-26): show error state on ProjectSelect when no accessible projects

### LINKED ISSUE #2933: Need to handle errors if the user doesn't belong to a project
### Main idea

[teams](https://teams.microsoft.com/l/message/19:14c484402d874dafb15806d093b95a82@thread.skype/1733808363550?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=74ae2c4d-ec4d-4fdf-b2c2-f5041d1e8631&parentMessageId=1733797630870&teamName=devops&channelName=Frontend&createdTime=1733808363550)

Currently, WebUI does not provide error handling for cases where the user does not have a project to which the user belongs. It is likely that we need to provide error handling via error boundary fallback only where there is a dependency on the project.

### Alternative ideas

_No response_

### Anything else?

_No response_

JIRA Issue: FR-26

ironAiken2:
> **Stacked on** #7469 (`chore: remove unused Alert import in VFolderNodes.tsx`) — merge that first.

Resolves #2933(FR-26)

## Summary

When the current user has no accessible projects in the current domain, `ProjectSelect` now visually and **accessibly** communicates the empty state:

- The select border switches to the error color via `status="error"`.
- The suffix arrow is replaced with an `InfoCircleOutlined` icon (visual cue).
- The explanation **"No accessible projects available."** (ko: "선택 가능한 프로젝트가 없습니다.") is attached to the focusable select control itself — via BAISelect's `tooltip` (hover over the whole control) and a persistent `aria-label` — so keyboard and screen-reader users can discover *why* it is in an error state, not just *that* it is.

The error treatment is suppressed while the select is `disabled` or `loading`, so existing UX is preserved:

- `WebUIHeader.tsx` — during project switching (`isProjectChanging`), the select is disabled+loading, so the error never flashes mid-switch.
- `UpdateUsersModal.tsx` — when no domain is picked yet, the field is disabled and the form-item already shows a "Please select a domain" warning; we don't compete with that.

## Why

`ProjectSelect` previously rendered as an empty dropdown with no indication of *why* it was empty. Users with no project membership (a real case after onboarding, after losing a project, or in a fresh domain) saw a normal-looking selector with nothing inside it. This change makes the empty state legible at a glance without a heavy error boundary fallback.

### Ghost-mode border fix (`BAISelect`)

The header `ProjectSelect` is rendered with `ghost`. `BAISelect`'s `ghostSelect` style hard-overrides `border-color` with `!important`, which swallowed antd's `status="error"` red border — so in the header (the primary use case) only the info icon appeared and the border stayed the ghost color. Added a higher-specificity `&.ant-select.ant-select-status-error` rule inside `ghostSelect` that restores the error border + suffix color. Normal (non-error) ghost selects are unaffected; no other ghost select uses `status="error"` today.

### Accessibility (Copilot review feedback)

The first revision exposed the reason only through a tooltip on the **non-focusable** suffix icon, so keyboard / screen-reader users got the error state without the reason. The message is now attached to the focusable control: `tooltip` (antd Tooltip over the whole select) for pointer users and a persistent `aria-label` (independent of tooltip open state) for assistive tech. The icon stays as a pure visual affordance.

## Changes

- `react/src/components/ProjectSelect.tsx` — derive `showNoProjectError = !accessibleProjects?.length && !disabled && !loading`; in the empty state forward `status="error"`, the `InfoCircleOutlined` suffix, and the explanation via BAISelect `tooltip` + `aria-label`. Preserves caller-supplied `status` / `suffixIcon` / `tooltip` / `aria-label` when not in the empty state.
- `packages/backend.ai-ui/src/components/BAISelect.tsx` — add an error-status override to the `ghostSelect` style so a ghost select can still surface a red error border (root cause of the header border not turning red).
- `resources/i18n/en.json`, `resources/i18n/ko.json` — add `projectSelect.NoAccessibleProjects` key. Other languages will be filled in by `/fw:i18n`.

## Test plan

- [ ] Header `ProjectSelect` in a domain with no accessible projects → red error border + info icon; hovering anywhere on the select shows the tooltip; the control exposes the reason as its accessible name.
- [ ] Tab to the header `ProjectSelect` with a screen reader in the no-project state → the reason is announced (not just "invalid").
- [ ] Header `ProjectSelect` during project switching (loading) → no error styling.
- [ ] `UpdateUsersModal` with no domain selected → field disabled, existing "Please select a domain" warning, no project-select error styling.
- [ ] `UpdateUsersModal` with a domain that has projects → normal selector, no error styling.
- [ ] `UserSettingModal` admin-edit flow with no accessible projects → error border + accessible reason.

### DIFF
--- a/react/src/components/ProjectSelect.tsx
+++ b/react/src/components/ProjectSelect.tsx
@@ -7,6 +7,7 @@ import { useSuspendedBackendaiClient } from '../hooks';
 import { useCurrentUserInfo, useCurrentUserRole } from '../hooks/backendai';
 import useControllableState_deprecated from '../hooks/useControllableState';
 import { useCurrentUserProjectRoles } from '../hooks/useCurrentUserProjectRoles';
+import { InfoCircleOutlined } from '@ant-design/icons';
 import { theme, Tooltip } from 'antd';
 import { BAIFlex, BAISelect, BAISelectProps } from 'backend.ai-ui';
 import * as _ from 'lodash-es';
@@ -28,6 +29,7 @@ export interface ProjectSelectProps extends BAISelectProps {
   autoSelectDefault?: boolean;
   disableDefaultFilter?: boolean;
   lockedProjectTypes?: string[];
+  'aria-label'?: string;
 }
 
 const ProjectSelect: React.FC<ProjectSelectProps> = ({
@@ -35,6 +37,7 @@ const ProjectSelect: React.FC<ProjectSelectProps> = ({
   domain,
   disableDefaultFilter,
   lockedProjectTypes,
+  'aria-label': ariaLabel,
   ...selectProps
 }) => {
   const { t } = useTranslation();
@@ -153,6 +156,13 @@ const ProjectSelect: React.FC<ProjectSelectProps> = ({
     },
   );
 
+  const showNoProjectError =
+    !accessibleProjects?.length &&
+    !selectProps.disabled &&
+    !selectProps.loading;
+
+  const noAccessibleProjectsMessage = t('projectSelect.NoAccessibleProjects');
+
   return (
     <BAISelect
       onChange={(value, option) => {
@@ -169,6 +179,25 @@ const ProjectSelect: React.FC<ProjectSelectProps> = ({
       options={
         _.size(groupOptions) > 1 ? groupOptions : groupOptions[0]?.options
       }
+      status={showNoProjectError ? 'error' : selectProps.status}
+      // Surface the empty-state reason on the focusable control itself,
+      // not only on the non-focusable suffix icon. `tooltip` wraps the
+      // whole BAISelect in an antd Tooltip (hover over the entire control,
+      // not just the tiny icon), and `aria-label` gives keyboard /
+      // screen-reader users a persistent accessible description that does
+      // not depend on the tooltip being open.
+      tooltip={
+        showNoProjectError ? noAccessibleProjectsMessage : selectProps.tooltip
+      }
+      aria-label={showNoProjectError ? noAccessibleProjectsMessage : ariaLabel}
+      suffixIcon={
+        showNoProjectError ? <InfoCircleOutlined /> : selectProps.suffixIcon
+      }
+      // Prevent the dropdown from opening in the empty-error state so it
+      // does not visually overlap the explanation tooltip. The tooltip
+      // alone communicates why the control is unusable; an empty popup
+      // beneath it just looks broken.
+      open={showNoProjectError ? false : selectProps.open}
     />
   );
 };
--- a/packages/backend.ai-ui/src/components/BAISelect.tsx
+++ b/packages/backend.ai-ui/src/components/BAISelect.tsx
@@ -43,6 +43,24 @@ const useStyles = createStyles(({ css, token }) => ({
         color: ${token.colorBgBase};
       }
     }
+
+    /* In ghost mode the border-color is hard-overridden with !important,
+       which would otherwise swallow antd's status="error" red border.
+       Restore the error treatment with a higher-specificity rule so a
+       ghost select (e.g. the header ProjectSelect) can still surface an
+       error state. */
+    &.ant-select.ant-select-status-error {
+      border-color: ${token.colorError} !important;
+
+      .ant-select-suffix {
+        color: ${token.colorError};
+      }
+
+      &:hover .ant-select-suffix,
+      &:active .ant-select-suffix {
+        color: ${token.colorError};
+      }
+    }
   `,
   customStyle: css`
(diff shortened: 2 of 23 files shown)


ironAiken2:
* **#7455** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7455?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7469** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7455).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/ProjectSelect.tsx
```diff
@@ -169,6 +175,16 @@ const ProjectSelect: React.FC<ProjectSelectProps> = ({
       options={
         _.size(groupOptions) > 1 ? groupOptions : groupOptions[0]?.options
       }
+      status={showNoProjectError ? 'error' : selectProps.status}
+      suffixIcon={
+        showNoProjectError ? (
+          <Tooltip title={t('projectSelect.NoAccessibleProjects')}>
+            <InfoCircleOutlined />
```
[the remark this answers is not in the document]
ironAiken2: Fixed in the latest commit. The empty-state explanation is now attached to the focusable select control instead of the non-focusable suffix icon:

- BAISelect's `tooltip` prop wraps the whole select in an antd Tooltip, so the message is reachable by hovering anywhere on the control (not just the tiny icon).
- A persistent `aria-label` exposes the reason to keyboard and screen-reader users **independently of the tooltip's open state** (antd/rc-tooltip only sets `aria-describedby` while the tooltip is open and its default trigger is hover-only, so a tooltip alone would not have been discoverable for AT users).

The `InfoCircleOutlined` now serves purely as a visual affordance. Note this also required a `ghostSelect` error-status override in BAISelect, since the header renders ProjectSelect with `ghost` and the ghost style's `!important` border-color was swallowing the `status="error"` red border.

react/src/components/ProjectSelect.tsx
```diff
@@ -169,6 +177,24 @@ const ProjectSelect: React.FC<ProjectSelectProps> = ({
...
+      // screen-reader users a persistent accessible description that does
+      // not depend on the tooltip being open.
+      tooltip={
+        showNoProjectError ? noAccessibleProjectsMessage : selectProps.tooltip
+      }
+      aria-label={
+        showNoProjectError
+          ? noAccessibleProjectsMessage
+          : selectProps['aria-label']
+      }
```
nowgnuesLee: I got ts error from here.

nowgnuesLee (dismissed):

react/src/components/ProjectSelect.tsx
ironAiken2: Fixed in the latest commit. The error was `TS7053`: `selectProps['aria-label']` indexed the rest-props bag with the hyphenated string literal `'aria-label'`, which `BAISelectProps` does not expose as an indexable key (the other forwarded props use dot access — `selectProps.tooltip`/`.suffixIcon`/`.status` — so they were fine). Resolved it by destructuring `'aria-label': ariaLabel` from props explicitly and forwarding `ariaLabel` instead of bracket-indexing. `react` tsc no longer reports `ProjectSelect.tsx`; the remaining `BAISelect.tsx`/`*Select.tsx` RefObject diagnostics are pre-existing package-wide issues unrelated to this PR (its only `BAISelect.tsx` change is a CSS-only ghost-mode error-border block).

agatha197 (changes requested):
can you block to open project list if there's no accessible project? cuurrently, the tooltip and menu item are ovelapped.

ironAiken2:
Addressed @agatha197's feedback in 93d66d4f6:

The dropdown is now suppressed (`open={false}`) whenever the no-accessible-projects error state is active, so the empty popup no longer overlaps the explanation tooltip. Clicking / focusing the control still surfaces the tooltip and the persistent `aria-label`, but no empty menu appears beneath it.

Also tightened `ProjectSelectProps` to explicitly declare the optional `'aria-label'` prop so the rest-props destructure type-checks cleanly under `react/tsc` (this had been silently broken).

agatha197 (approved):
LGTM

PR #7509, opened 2026-05-20
fix(FR-2932): disable SW navigateFallback to allow OIDC/SAML callback navigations

### LINKED ISSUE #7508: SW navigateFallback default breaks OIDC/SAML callback after Vite migration
## Summary

FR-2611 (craco → Vite build migration, #6876) replaced the SW generator from `workbox-webpack-plugin` to `vite-plugin-pwa`. The new plugin silently introduced `workbox.navigateFallback: "index.html"` as a default, which the old plugin did not have. As a result, the SW intercepts every same-origin GET navigation and responds with the cached `index.html` — including the OIDC/SAML callback navigations (`/func/openid/redirect`, `/func/saml/redirect`) that the IdP redirects back to our origin. The webserver's callback handler is never invoked, no session cookie is set, and SSO login silently fails (user sees the login page again).

## Reproduction

- Environment: any non-localhost deployment (e.g., the internal dev environment). `index.html:122-131` gates SW registration to non-localhost hosts, so `pnpm dev` and Electron cannot reproduce.
- `config.toml`: `singleSignOnVendors = "openid"` or `"saml"`
- Steps:
  1. On the login page, click "Login with OpenID" (or SAML)
  1. Authenticate at the IdP (e.g., Microsoft Entra ID)
  1. IdP redirects to `https://<host>/func/openid/redirect?code=...&state=...`
  1. Browser receives only the SPA shell and re-renders the login page
- DevTools → Application → Service Workers → check "Bypass for network" → reload → login completes normally. This isolates the SW as the cause.

## Root cause

- `vite-plugin-pwa@1.2.0` `dist/index.js:838` defines `defaultWorkbox.navigateFallback: "index.html"`.
- It is merged via `Object.assign({}, defaultWorkbox, options.workbox || {})`, so unless we explicitly override, the default applies.
- The retired `workbox-webpack-plugin` `GenerateSW` had no such default (`undefined`) — see the deleted `react/craco.config.cjs` in commit `eaec71340`. The migration preserved every explicitly set option but missed the default difference between the two wrappers.
- workbox `NavigationRoute` matches `request.method === "GET"` + `request.mode === "navigate"`, and its handler is cache-only — the network is never even attempted, so being online does not save us.

## Proposed fix

Add `navigateFallback: null` to the `workbox` block in `react/vite.config.ts` to restore the pre-FR-2611 behavior:

```
workbox: {
  skipWaiting: true,
  clientsClaim: true,
  maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
  globIgnores: ['**/*.map', '**/asset-manifest.json'],
  navigateFallback: null,  // ← add
},
```

Alternative: `navigateFallbackDenylist: [/^\/func\//, /^\/server\//, /^\/openid\//, /^\/saml\//]` keeps SPA offline navigation fallback while letting backend routes pass through. The craco-era build never had offline navigation fallback anyway, so `null` is the smaller behavioral delta.

Note on rationale for `null` vs `undefined` vs `false`: workbox-build's schema is `string | null`. Setting `undefined` leaves the default `"index.html"` in place via `Object.assign`. `false` is rejected by the type checker. `null` is the correct way to opt out.

## Verification

- Cannot be reproduced locally (`index.html:122-131` gates SW registration to non-localhost). Verification requires building and deploying to a non-localhost environment.
- Force-refresh the SW on the deployed environment (DevTools → Application → Service Workers → Unregister, or "Update on reload").
- Retry the OIDC/SAML login flow.
- In the Network tab, the response to `/func/openid/redirect` should NOT have `(ServiceWorker)` in the Size column and should be a 302 redirect (not an HTML body).

## Impact

- Scope: every SSO flow (OIDC and SAML) on every non-localhost, non-Electron deployment.
- Regression introduced: FR-2611 (#6876), merged 2026-04-30.
- Severity: high — SSO login is completely broken in deployed environments configured to use it.

—

JIRA Issue: FR-2932

nowgnuesLee:
Resolves #7508 (FR-2932)

## Summary

The Vite migration (FR-2611, #6876) silently changed the SW's navigation behavior:

- `vite-plugin-pwa@1.2.0` defaults `workbox.navigateFallback` to `"index.html"` (`dist/index.js:838`).
- The retired `workbox-webpack-plugin` `GenerateSW` had no such default (`undefined`).
- The migration preserved every explicitly set option but didn't account for the differing implicit defaults between the two workbox wrappers.

In production, the registered SW now runs a `NavigationRoute` that responds to every same-origin GET navigation with the cached `index.html` — *including* the OIDC/SAML callbacks the IdP redirects back to our origin (`/func/openid/redirect`, `/func/saml/redirect`). The webserver's callback handler is never invoked, so no session cookie is set and SSO login silently fails (the user is re-rendered into the login page).

This is a cache-only handler, so being online does not save us — the network is never even attempted.

## Fix

Set `navigateFallback: null` in `react/vite.config.ts`'s `workbox` block to restore pre-FR-2611 behavior. With `navigateFallback` falsy, `workbox-build`'s SW template (`workbox-build/.../sw-template.ts:50`) gates out the `NavigationRoute` registration entirely, so navigation requests pass through to the network. Precaching of build assets is unaffected.

Value choice — workbox's schema is `string | null`. `undefined` leaves the plugin default `"index.html"` in place via `Object.assign`; `false` is rejected by the type checker; `null` is the correct opt-out.

Alternative considered: `navigateFallbackDenylist: [/^\/func\//, /^\/server\//, /^\/openid\//, /^\/saml\//]`. Rejected because the craco-era build never had offline navigation fallback anyway, so `null` is the smaller behavioral delta and matches the implicit contract this project already shipped on.

## Test plan

- [ ] Build the production bundle (`pnpm run build`) and deploy to a non-localhost environment. `index.html:122-131` gates SW registration to non-localhost hosts, so `pnpm dev` and Electron cannot reproduce.
- [ ] On the deployed environment, force-refresh the SW (DevTools → Application → Service Workers → Unregister, or "Update on reload").
- [ ] Attempt OIDC login (or SAML if configured). Authenticate at the IdP.
- [ ] In the Network tab, confirm the response to `/func/openid/redirect` does NOT show `(ServiceWorker)` in the Size column and is a 302 redirect (not an HTML body).
- [ ] Confirm login completes — session cookie is set and the user lands on the post-login page instead of the login page.
- [ ] Sanity-check that SPA routes (`/sessions`, `/folders`, etc.) still load normally on hard refresh in the same environment — the webserver's catch-all serves `index.html` on these, so behavior should be identical.

### DIFF
--- a/react/vite.config.ts
+++ b/react/vite.config.ts
@@ -788,6 +788,12 @@ export default defineConfig(({ command, mode }) => {
           clientsClaim: true,
           maximumFileSizeToCacheInBytes: 5 * 1024 * 1024,
           globIgnores: ['**/*.map', '**/asset-manifest.json'],
+          // vite-plugin-pwa defaults navigateFallback to 'index.html', which
+          // makes the SW serve cached HTML for ANY GET navigation that doesn't
+          // match a precached asset — including OIDC/SAML callbacks like
+          // /func/openid/redirect. The retired workbox-webpack-plugin had no
+          // such default, so this restores pre-FR-2611 behavior.
+          navigateFallback: null,
         },
       }),
     ],


nowgnuesLee:
* **#7509** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7509?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7509).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

yomybaby (approved):
LGTM

PR #7566, opened 2026-05-22
fix(FR-2954): wrap BAIAgentTable JSON-parsed cells in error boundary

nowgnuesLee:
Resolves FR-2954

## Summary

`BAIAgentTable` crashed with `TypeError: Cannot read properties of undefined (reading 'current')` when an agent reported a sparse GPU live_stat (e.g. Jetson Orin Nano + `cuda_open` plugin where NVML returns `NOT_SUPPORTED` for `nvmlDeviceGetMemoryInfo`). Because all cell renders ran inline in the parent table render, a single bad row took down the entire `/agent` page and any other view sharing the same ErrorBoundary scope (notably the Session Launcher's environment/agent dropdown).

## Changes

- Extract the four columns that parse JSON (`occupied_slots`, `available_slots`, `live_stat`, `compute_plugins`) into module-scope sub-components: `AllocationCell`, `UtilizationCell`, `DiskPctCell`, `StatusCell`. This is required because `JSON.parse` must run inside a child component's render for an `ErrorBoundary` to catch it — running inside the parent's `render: (...) => { ... }` throws before the boundary is mounted.
- Wrap each of those four cells with a new `CellErrorBoundary` (built on `react-error-boundary`). On error the cell renders `BAIAlertIconWithTooltip` (red alert icon with a tooltip showing a localized failure message) instead of crashing.
- Add `comp:AgentTable.FailedToLoadCellData` to all 22 locale files (ko: "데이터를 불러오는데 실패했습니다.", en: "Failed to load data.", plus 20 other languages auto-translated).

### Normal render (after fix, no regression)
normal

### Error fallback — alert icons in the 4 affected cells (simulated via temporary throws)
The page no longer crashes; only the affected cells degrade. ID / Endpoint, Region, Starts, Resource Group, Schedulable continue to render normally.

error fallback

### Tooltip on hover — "Failed to load data."
Hovering an alert icon reveals the localized failure message.

error tooltip

## How to verify

1. `pnpm dev` → log in → Admin → Resources → Agents.
2. With a cluster containing an agent that reports incomplete GPU metrics (e.g. Jetson Orin Nano on the cuda_open plugin), confirm the Agents table renders.
3. Affected cells in the misbehaving row show a red alert icon — hover to see "Failed to load data." (or the localized equivalent). Other rows and other cells in the same row continue to render normally.
4. Open the Session Launcher and confirm the environment/agent dropdown is no longer affected.

## Checklist

- [x] No backend / manager changes required
- [x] i18n keys added for all 22 supported locales
- [ ] Documentation

> Note: The Jira → GitHub webhook had not yet cloned FR-2954 to GitHub at PR creation time. Once the clone lands, this description will be updated to `Resolves #<N> (FR-2954)`.

### DIFF
--- a/packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
+++ b/packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
@@ -8,7 +8,8 @@ import {
   convertUnitValue,
   toFixedFloorWithoutTrailingZeros,
 } from '../../helper';
-import { useResourceSlotsDetails } from '../../hooks';
+import { useBAILogger, useResourceSlotsDetails } from '../../hooks';
+import BAIAlertIconWithTooltip from '../BAIAlertIconWithTooltip';
 import BAIDoubleTag from '../BAIDoubleTag';
 import BAIFlex from '../BAIFlex';
 import BAIIntervalView from '../BAIIntervalView';
@@ -24,6 +25,11 @@ import { theme, Typography } from 'antd';
 import type { ColumnsType } from 'antd/es/table';
 import dayjs from 'dayjs';
 import * as _ from 'lodash-es';
+import { type ErrorInfo } from 'react';
+import {
+  ErrorBoundary,
+  type ErrorBoundaryPropsWithRender,
+} from 'react-error-boundary';
 import { useTranslation } from 'react-i18next';
 import { graphql, useFragment } from 'react-relay';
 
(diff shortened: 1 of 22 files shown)


nowgnuesLee:
* **#7566** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7566?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7566?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7566).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
```diff
@@ -47,6 +50,520 @@ const isEnableSorter = (key: string) => {
   return _.includes(availableAgentSorterKeys, key);
 };
 
+const CellErrorBoundary: React.FC<PropsWithChildren> = ({ children }) => {
+  'use memo';
+  const { t } = useTranslation();
+  return (
+    <ErrorBoundary
+      fallbackRender={() => (
```
agatha197, line 71: **(suggestion) The fallback discards the caught `error` and shows only a static message.**

`fallbackRender={() => (...)}` ignores the `error` argument it receives, which has two side effects:

1. The tooltip always shows the generic `FailedToLoadCellData` ("Failed to load data."). The PR description says "a tooltip showing the failure reason", but the actual reason is never surfaced — please align the description with the code.
2. The error is silently swallowed. There is no trace of which agent / which field threw, which makes root-cause analysis hard in production. Since CLAUDE.md requires `useBAILogger` over `console.log`, consider logging via `<ErrorBoundary onError={(error) => logger.error(...)}>` (this would also help follow-up analysis of the Jetson sparse-stat case this PR addresses).

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
```diff
@@ -47,6 +50,520 @@ const isEnableSorter = (key: string) => {
   return _.includes(availableAgentSorterKeys, key);
 };
 
+const CellErrorBoundary: React.FC<PropsWithChildren> = ({ children }) => {
+  'use memo';
+  const { t } = useTranslation();
+  return (
+    <ErrorBoundary
```
agatha197, line 63: **(robustness) No `resetKeys`, so once a cell errors it stays stuck on the error icon.**

`react-error-boundary` keeps rendering the fallback after catching an error until `resetKeys` change or `reset()` is called. The agent table refreshes `live_stat` periodically, so if a row throws once on a transient/partial payload, that cell will remain on the error icon even after a later poll returns valid data.

For the targeted bug (a consistently sparse Jetson stat) keeping the error visible is the intended behavior, so this is not a problem here. But to recover from transient errors, passing something like `resetKeys={[value, record?.live_stat, record?.occupied_slots, record?.available_slots, record?.compute_plugins]}` would let the cell auto-recover when the data changes. Non-blocking.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
```diff
@@ -47,6 +50,520 @@ const isEnableSorter = (key: string) => {
...
+    >
+      {children}
+    </ErrorBoundary>
+  );
+};
+
+const AllocationCell: React.FC<{ record: AgentNodeInList }> = ({ record }) => {
+  'use memo';
+  const { token } = theme.useToken();
+  const { mergedResourceSlots } = useResourceSlotsDetails();
```
agatha197, line 87: **(nit, perf) `useResourceSlotsDetails()` now runs per-cell per-row instead of once in the parent.**

`AllocationCell` and `UtilizationCell` each call it, so it is roughly rows × 2 calls. That said, it is backed by `useTanQuery` (React Query), so calls with the same queryKey are deduped — no extra network requests — and `'use memo'` + React Query structural sharing keep the `_.merge` result memoized. So the real cost is small and this is not blocking; just a note in case the agent list ever grows very large.

agatha197 (commented):
## Review summary (FR-2954)

Overall this is a **well-targeted, low-risk fix**. The core idea — extracting the four cells (`AllocationCell` / `UtilizationCell` / `DiskPctCell` / `StatusCell`) into module-scope components wrapped in `CellErrorBoundary` — correctly relies on the fact that `JSON.parse` must run inside a child component's render for an `ErrorBoundary` to catch it. Comparing the diff line by line, the extracted logic is identical to the original `render` callbacks (the only difference is the non-ALIVE branch now wraps the string in a `<>{...}</>` fragment — behavior unchanged). `mergedResourceSlots` was removed from the parent and moved into the cells, while `token` is still used by the schedulable column, so there are no unused variables.

Verification:
- CI all green (lint / tsc (Analyze) / relay-drift / backend-ai-ui-vitest)
- `react-error-boundary` is declared as a direct dependency in `backend.ai-ui/package.json` (^6.1.1)
- i18n key `comp:AgentTable.FailedToLoadCellData` added to all 22 locales, in alphabetical position
- `BAIAlertIconWithTooltip` extends `TooltipProps`, so `title` / `type="error"` are valid props

Below are 3 non-blocking suggestions (see inline). I think the PR is good to merge as-is. 👍

agatha197 (commented):
.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
nowgnuesLee: Applied. `CellErrorBoundary` now calls `useBAILogger` and passes an `onError` handler that logs the error and React's `info.componentStack` (which identifies which of the four cell subcomponents threw). Also fixed the PR description wording — it now says "a tooltip showing a localized failure message" instead of "failure reason" to match what the code actually surfaces (the tooltip stays as a static localized string; the real reason now lives in the logger). Resolving.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
nowgnuesLee: Applied. `CellErrorBoundary` now accepts `Omit<ErrorBoundaryPropsWithRender, 'fallbackRender'>` and each column passes `resetKeys` scoped to only the fields that cell actually parses — Allocation: `[occupied_slots, available_slots]`, Utilization: `[value, available_slots]`, DiskPct: `[live_stat]`, Status: `[value, compute_plugins, available_slots]`. Transient errors will now auto-recover on the next `live_stat` poll. Resolving.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
nowgnuesLee: Agreed. For the current agent-list scale this is negligible — React Query's queryKey dedup keeps it to one HTTP call, and `'use memo'` + structural sharing keep the merged result memoized. Will revisit if the agent count grows much larger. Leaving as-is. Resolving.

agatha197 (changes requested):
Please review copilot's feedbacks

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
```diff
@@ -47,6 +53,531 @@ const isEnableSorter = (key: string) => {
...
+              <BAIFlex
+                key={statKey}
+                justify="between"
+                style={{ minWidth: 200, width: '100%' }}
+                gap="xxs"
+                data-testid={`live-stat-${statKey}`}
+              >
+                <Typography.Text>
+                  {mergedResourceSlots?.[deviceName]?.human_readable_name}
+                  (util)
```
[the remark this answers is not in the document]
nowgnuesLee: Decided to keep these as-is. `(util)` and `(mem)` are short metric-type suffixes appended after a device name (e.g. `cuda.device (util)`) — they are not natural-language phrases but column/metric identifiers used identically across all 22 locales, the same way `GPU`, `Rx`, or `Tx` are. They are also not introduced by this PR; they are pre-existing labels in the Utilization cell that we only moved into a module-scope subcomponent to enable the per-cell error boundary. The intentional i18n scope of FR-2954 is the new `comp:AgentTable.FailedToLoadCellData` fallback key. Relocalizing the agent table's existing labels (if we decide to) should be a separate task so it can be reviewed end-to-end. Resolving.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
```diff
@@ -47,6 +53,531 @@ const isEnableSorter = (key: string) => {
...
+            return (
+              <BAIFlex
+                key={statKey}
+                justify="between"
+                style={{ minWidth: 200, width: '100%' }}
+                gap="xxs"
+                data-testid={`live-stat-${statKey}`}
+              >
+                <BAIText>{statKey === 'net_rx' ? 'Net Rx' : 'Net Tx'}</BAIText>
+                <BAIText>
```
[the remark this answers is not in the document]
nowgnuesLee: Decided to keep these as-is. `Net Rx`, `Net Tx`, `Agent`, `CUDA`, `CUDA Plugin`, `CUDA Disabled`, and `Fractional GPU™` are technical / network / brand terms that render identically across all 22 locales — there is no natural-language translation for them. They are also not new in this PR; they are pre-existing labels in the Status / Utilization cells that this change only moved into module-scope subcomponents to enable the per-cell error boundary. Out of scope for FR-2954 (cell error boundary). If we want to relocalize the agent table's existing labels, that should be a separate task so it can be reviewed end-to-end. Resolving.

packages/backend.ai-ui/src/components/fragments/BAIAgentTable.tsx
```diff
@@ -47,6 +53,531 @@ const isEnableSorter = (key: string) => {
...
+        logger.error(
+          '[BAIAgentTable] Cell render failed:',
+          error,
+          info.componentStack,
+        )
+      }
+      fallbackRender={() => (
+        <BAIAlertIconWithTooltip
+          type="error"
+          title={t('comp:AgentTable.FailedToLoadCellData')}
```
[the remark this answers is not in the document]
nowgnuesLee: I think it's ok.

agatha197 (approved):
LGTM

PR #7577, opened 2026-05-24
fix(FR-2965): keep remaining filter tags when closing one tag

### LINKED ISSUE #7576: Property filter hides all tags when removing the first filter tag
The shared GraphQL property filter removes (hides) every filter tag when the user closes the first of multiple filter tags, instead of removing only that one condition.

## Why This Work is Needed

`BAIGraphQLPropertyFilter` is used across the app (deployment/route/session scheduling history modals, deployment and other list-page filters). With two or more filters applied, clicking the close (x) of the first tag makes all tags visually disappear, leaving the filter unusable until the page is reloaded. This is a regression in multi-filter UX and affects every page that relies on this component.

## Expected Outcomes

- Closing a single filter tag removes only that condition; all remaining filter tags stay visible.
- The fix applies uniformly to every consumer of `BAIGraphQLPropertyFilter` (scheduling history modals and list filters).
- No change to the underlying GraphQL filter serialization behavior.

JIRA Issue: FR-2965

yomybaby:
Resolves #7576 (FR-2965)

## Summary

Closing a filter tag could make **other** filter tags disappear. Both shared filter components keep antd `Tag`'s self-hide behavior on close (antd's `Tag` toggles its own internal `visible` state unless the close event is prevented) while re-keying tags with a positional component — so after a removal the just-hidden Tag instance gets reused for a surviving condition, hiding the wrong tag.

- **`BAIGraphQLPropertyFilter`** — tags are keyed positionally (`cond-${i}`). Closing the first of several tags hid **every** remaining tag (it became `ant-tag-hidden` / `display:none`), leaving the filter unusable. Affects the deployment / route / session scheduling history modals and list-page GraphQL filters.
- **`BAIPropertyFilter`** — tags are keyed as `index + value`. The same bug surfaces when two filters share the same parsed value (e.g. filtering two different properties by the same text): closing the first reused the just-hidden instance and hid the survivor.

### Fix

- Call `e.preventDefault()` in each `Tag`'s `onClose` so antd no longer self-hides the tag; removal is driven solely by the value / conditions re-render.
- Positional keys are kept intentionally (duplicate conditions are allowed, so content-based keys would collide); a cross-reference comment documents the coupling.

## Test plan

- [x] Added regression tests for both components (`BAIGraphQLPropertyFilter.test.tsx`, `BAIPropertyFilter.test.tsx`): closing one of two tags keeps the other visible and **not** `ant-tag-hidden`, plus onChange-payload and single-tag-clear cases. Verified each new "surviving tag" test **fails without the fix** and passes with it.
- [x] `vitest run` on both files: all green (BAIPropertyFilter 27 passed, BAIGraphQLPropertyFilter 3 passed).
- [x] Verified live on the dev server (`/deployments` GraphQL filter): add two filters, close the first → only the second remains visible (before the fix both vanished — confirmed via a before/after run measuring `.ant-tag:visible`).
- [x] `scripts/verify.sh`: Relay / Lint / Format pass; tsc reports only pre-existing `@types/react` phantom errors on unrelated files in fresh worktrees (none reference the changed files).

### DIFF
--- a/packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.test.tsx
+++ b/packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.test.tsx
@@ -1,0 +1,96 @@
+import BAIGraphQLPropertyFilter, {
+  type FilterProperty,
+  type GraphQLFilter,
+} from './BAIGraphQLPropertyFilter';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { useState } from 'react';
+
+const filterProperties: Array<FilterProperty> = [
+  {
+    key: 'name',
+    propertyLabel: 'Name',
+    type: 'string',
+    fixedOperator: 'iContains',
+  },
+];
+
+// Controlled wrapper so closing a tag triggers a real re-render with the
+// updated filter value (mirrors how consumers pass value/onChange).
+const Controlled = ({ initial }: { initial: GraphQLFilter | undefined }) => {
+  const [value, setValue] = useState<GraphQLFilter | undefined>(initial);
+  return (
+    <BAIGraphQLPropertyFilter
+      filterProperties={filterProperties}
+      value={value}
+      onChange={setValue}
+    />
+  );
+};
+
+describe('BAIGraphQLPropertyFilter tag removal', () => {
+  const twoFilters: GraphQLFilter = {
+    AND: [{ name: { iContains: 'alpha' } }, { name: { iContains: 'beta' } }],
+  };
+
+  it('removes only the closed tag and keeps the rest visible (FR-2965)', async () => {
+    render(<Controlled initial={twoFilters} />);
+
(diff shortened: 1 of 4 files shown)


yomybaby:
* **#7577** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7577?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7577?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7577).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #7587, opened 2026-05-26
feat(FR-2969): migrate storage host permission management to detail drawer

### LINKED ISSUE #7588: Implement storage host permission migration (drawer + permissions tab + cleanup)
Single-PR implementation of the FR-2969 Epic: port storage host permission management from control-panel into backend.ai-webui, absorb the standalone Storage Setting page into a tabbed detail drawer, and remove the legacy `/storage-settings/:hostname` route.

**Phases (all shipped in one PR):**

1. `StorageHostDetailDrawer` + `StorageHostDetailDrawerContent` + Capacity Setting tab (existing `StorageHostSettingsPanel` moved into the drawer).
1. `StorageHostPermissionPanel` — three Relay queries, three `modify_*` mutations, mount-in-session warning, superadmin gate, Select All / Unselect All per section.
1. Delete `StorageHostSettingPage.tsx`, route `/storage-settings/:hostname` becomes `<WebUINavigate to="/agent" replace />`, drop `storage-settings` special cases in WebUISider / useWebUIMenuItems.

**PR**: https://github.com/lablup/backend.ai-webui/pull/7587

**Spec**: `.specs/FR-2969-storage-host-permission-migration/spec.md`

Tracks the implementation surface of FR-2969; the Epic itself remains as the parent.

JIRA Issue: FR-2971

ironAiken2:
Resolves #7588 (FR-2971) — under Epic [FR-2969](https://lablup.atlassian.net/browse/FR-2969).

## Overview

Migrate the **storage host permission management** feature from `backend.ai-control-panel` (`/vfolders/?tab=permissions`) into `backend.ai-webui`'s Resources → Storage tab, and absorb the existing standalone **Storage Setting page** (`/storage-settings/:hostname`) into a tabbed detail drawer opened from the storage row's gear icon.

Before this change operators had to switch between control-panel (for storage host permissions) and webui (for storage host quotas). This PR consolidates both flows under the gear icon on every storage row, mirroring the project's existing detail-drawer pattern (`AgentDetailDrawer`, `RoleDetailDrawer`).

## Phase 1 — Drawer skeleton + Capacity Setting tab

- New `react/src/components/StorageHostDetailDrawer.tsx` and `StorageHostDetailDrawerContent.tsx`, mirroring `AgentDetailDrawer.tsx` / `AgentDetailDrawerContent.tsx`. Drawer header title = storage host id, `BAIFetchKeyButton` in `extra`, Suspense inside the content child.
- The Capacity Setting tab embeds `<StorageHostSettingsPanel/>` (rewritten — see below) with the "quota not supported" `<Empty/>` fallback wrapped in `BAICard` for visual consistency.
- Inner tabs use a bare `<Tabs>` inside `BAIFlex` (not `BAICard` with `tabList`), matching `AgentDetailDrawerContent`.
- `StorageProxyList.tsx` row link now opens the drawer (`setDrawerStorageHostId(record.id)`) instead of navigating to `/storage-settings/...`. The drawer is mounted at the list level inside `BAIUnmountAfterClose`.

### Capacity Setting panel rewrite (`QuotaScopeTable`)

`StorageHostSettingsPanel` was rewritten — the legacy `useQueryLoader` + `usePreloadedQuery` + `PreloadedQuotaScope` wrapper + `queryRef ? ... : ...` ternary chain is replaced with:

- A new self-contained `QuotaScopeTable` component (replacing `QuotaScopeCard`). Owns its own `useLazyLoadQuery` gated by `@skip(if: $skip)` + `fetchPolicy: 'store-only'` when no scope is picked (zero network at idle). Embeds `QuotaSettingModal` inline under `BAIUnmountAfterClose`.
- Panel-level `Form` removed — `userId` / `projectId` held in plain `useState` slots so switching the User/Project radio preserves each side's last pick.
- antd `Table` → `BAITable`. The `quota_scope_id` column uses `BAINameActionCell` with inline Edit + Unset actions (`type: 'danger'` + Popconfirm; the Promise-returning `onConfirm` makes the Popconfirm OK-button spinner reflect mutation state).
- Removes the unreachable `addQuotaConfigsWhenEmpty` branch — `quota_scope` is always assigned for any picked entity.

## Phase 2 — Permissions tab (rebuilt from the original PR design)

The original design used a single panel-level query that fetched **all** domains / groups (capped at 500) / keypair resource policies and rendered three full checkbox matrices with a panel-level Update button. After review feedback that scrolling 500+ rows and bulk-editing them was the wrong UX, this PR replaces that with **per-entity lazy tables driven by selectors**:

- **`StorageHostPermissionPanel`** is now a thin coordinator: one panel-level `useLazyLoadQuery` for `vfolder_host_permissions.vfolder_host_permission_list` (the canonical permission key list, used for column headers), three `useState` slots for the per-card selections, and three `BAICard`s — each card has a select in the header `extra` and a single-row table in the body.
- Top of the panel renders a `BAIAlert` (type=info) explaining that a user's effective permissions are the **union** of their domain, project, and keypair resource policy entries.

### Three per-entity tables (one row each)

- **`DomainStoragePermissionTable`** — V1 `domain(name: String): Domain` + `allowed_vfolder_hosts` JSONString. Domain stays on V1 because `DomainV2` does not expose `allowedVfolderHosts`.
- **`ProjectStoragePermissionTable`** — V2 `projectV2(projectId: UUID!): ProjectV2` + `storage.allowedVfolderHosts: [VFolderHostPermissionEntry!]!`. No `toGlobalId` dance — the V2 query takes the raw UUID returned by `BAIAdminProjectSelect`.
- **`KeypairResourcePolicyStoragePermissionTable`** — V2 `adminKeypairResourcePolicyV2(name: String!): KeypairResourcePolicyV2` + `allowedVfolderHosts`.

Each table:
- Issues its own `useLazyLoadQuery` gated by `@skip` + `fetchPolicy: 'store-only'` when no selection (no network call at idle).
- `network-only` + bumped `fetchKey` on save success → re-fetches the entity so the table reflects the just-saved permissions immediately. The V1 `modify_*` mutations return only `{ ok, msg }`, so Relay cannot auto-update the cached entry.
- `useDeferredValue` on query variables and fetchKey keeps the previous row visible (with `BAITable.loading` spinner) instead of falling through to a Suspense fallback during entity switch or post-save refetch.
- Extends `BAITableProps<RowType>` so the panel can pass `locale`, `scroll`, etc. through.
- Name column uses `BAINameActionCell` with an Edit action. Under the name, a small `Typography.Text type="secondary"` line shows **host-allowed status** with `CheckCircleOutlined` (green) / `CloseCircleOutlined` (red).
- Permission columns are **read-only icons** — `CheckCircleOutlined` (token.colorSuccess) when enabled, `CloseCircleOutlined` (token.colorTextDisabled) otherwise.

### Shared `StoragePermissionEditModal`

Edit action opens a single shared modal (`BAIUnmountAfterClose`-wrapped). Each table injects its own `onSave` callback that runs the appropriate `modify_*` mutation.

- Top row: master "전체" `Checkbox` with `indeterminate` state and `n / m` count on the right.
- A `Divider` and a secondary-color "권한" section label below the master, mirroring the antd Dropdown group-header style.
- Permission checkboxes use `PERMISSION_DISPLAY_MAP` for labels.
- **Mount-in-session warning** runs as `App.useApp().modal.confirm` (reversible save → not `BAIConfirmModalWithInput`, per `.claude/rules/destructive-confirmation.md`). Triggered when the user's edited set has `mount-in-session` without both `download-file` and `upload-file`.

### V1 ↔ V2 permission key mapping

V1 stores permissions as kebab keys in a JSONString (`allowed_vfolder_hosts`); V2 returns a typed `[VFolderHostPermissionEntry!]!` with `VFolderHostPermissionV2` enum values. Most enum values lowercase-and-dash cleanly (`CREATE_VFOLDER` ↔ `create-vfolder`), but **`SET_USER_PERM` ↔ `set-user-specific-permission` is asymmetric**. Without an explicit map, the naive conversion produced `set-user-perm`, which never matched the canonical column key from `vfolder_host_permissions.vfolder_host_permission_list` — the "권한 설정" column silently appeared unchecked even when the entity had the permission, and round-tripped saves would corrupt other hosts' entries. `react/src/helper/storageHostPermission.ts` defines a spelled-out `V2_TO_V1_PERMISSION` map with a naive fallback for forward-compat.

### Mutations still on V1

Reads use V2 (structured array) where available; writes stay on the V1 mutations (`modify_domain` / `modify_group` / `modify_keypair_resource_policy`) since the corresponding V2 mutations are not wired yet. `buildAllowedHostsPayloadFromV2` rebuilds the V1 JSONString from V2 entries, preserving all other hosts' entries verbatim.

### Drop the queryLoader plumbing

The eager `useQueryLoader(StorageHostPermissionsQuery)` + `loadPermissionsQuery({}, { fetchPolicy: 'store-and-network' })` on row open is gone:

- `StorageProxyList.tsx` drops `useQueryLoader`, the load call on row click, and the props that forwarded the ref to the drawer.
- `StorageHostDetailDrawer.tsx` and `StorageHostDetailDrawerContent.tsx` drop `permissionsQueryRef` and `onReloadPermissions` props.
- Each table's own `useLazyLoadQuery` replaces the panel-level preloaded query.

### Permissions tab gating

Still **superadmin only** via `useCurrentUserRole()` from `react/src/hooks/backendai.tsx`. Domain-admins and regular users see only the Capacity Setting tab.

## Phase 3 — Old standalone page removal

- Deletes `react/src/pages/StorageHostSettingPage.tsx`.
- `react/src/routes.tsx`: removes the lazy import; the `/storage-settings/:hostname` route now renders `<WebUINavigate to="/agent" replace />`.
- `react/src/components/MainLayout/WebUISider.tsx`: drops the `storage-settings` → `agent` ternary in `selectedKeys`.
- `react/src/hooks/useWebUIMenuItems.tsx`: drops the two `storage-settings` special cases.

## New `BAIAdminKeypairResourcePolicySelect` (backend.ai-ui)

Modeled on `BAIAdminProjectSelect`. Uses the new V2 `adminKeypairResourcePoliciesV2(filter: KeypairResourcePolicyV2Filter, orderBy: [KeypairResourcePolicyV2OrderBy!])` (added in 26.4.2) for the paginated dropdown with name-based search (`name: { contains }`). Drop-open gated `network-only` (no fetch at mount). KRP `name` is both identifier and display label, so no separate value query.

Exported from `packages/backend.ai-ui/src/components/fragments/index.ts` alongside the existing admin selects.

## Drive-by fixes (backend.ai-ui)

- **`BAICard.cloneElement` style merge** — the `extra` element's `style` was being **replaced** with `{ fontWeight: 'normal' }`, silently wiping any `style.width` a consumer set on a Select used as the card extra. Now merges so consumer-set sizing wins; `fontWeight: 'normal'` is the default, and consumer-set `fontWeight` still wins (spread after).
- **`BAIDomainSelect` Props** — `interface Props extends Omit<SelectProps, 'role'>` (matching `BAISelectProps`), removing the pre-existing baseline `Property 'role' is missing` TS error from this component and from the existing `UserSettingModal` call sites.

## i18n

New keys under `storageHost.permission.*`:

- `All`, `EditPermissions`, `EditPermissionsAction`, `EffectivePermissionsNote`, `HostAllowed`, `HostNotAllowed`, `NoDomainSelected`, `NoProjectSelected`, `NoKeypairResourcePolicySelected`, `Permissions`

New key under `comp:BAIAdminKeypairResourcePolicySelect.*`:

- `SelectKeypairResourcePolicy`

All keys propagated to **all 19 target locales** by the `i18n-translator` agent. Korean uses `자원` (not `리소스`) per the PR-2969 terminology decision.

### 1. Storage list — Resources › 스토리지

Storage list row's ID column is now the drawer trigger (replaces the old route nav to `/storage-settings/:hostname`).

storage-list

### 2. Detail drawer — Permissions tab (default for superadmin, no selection yet)

Top BAIAlert explains the union rule. Three cards (도메인 / 프로젝트 / 키페어 자원 정책) — each with its select in the card extra. Tables start empty with a "선택해주세요" hint and skip the network entirely until a selection is made (`@skip` + `store-only`).

detail-drawer-permissions

### 3. Permissions tab — domain selected ("default")

Picking a domain lazily fetches just that single entity (V1 `domain(name)` for Domain; V2 `projectV2` / `adminKeypairResourcePolicyV2` for the other two). Name column shows the host-allowed status under the name (✓ "호스트 허용됨" in this case); permission columns render as read-only ✓/✗ icons.

permissions-domain-selected

### 4. Edit modal — `권한 수정 — default`

Triggered by the edit icon in the row's `BAINameActionCell`. Master "전체" checkbox with indeterminate support + `n / m` count, "권한" section label below the divider, then per-permission checkboxes. Save runs the appropriate V1 `modify_*` mutation built from the V2 structured `allowedVfolderHosts` (preserving every other host's entries) and bumps the table's `fetchKey` so the new state is reflected immediately.

permissions-edit-modal

### 5. Capacity Setting tab — quota not supported

For storage backends without quota capability (here, the `vfs` test volume), the tab renders the "이 스토리지 백엔드는 가용량 설정을 지원하지 않습니다." `<Empty>` fallback wrapped in `BAICard` for visual consistency with the rest of the tab body.

capacity-tab-not-supported

## Verification

`bash scripts/verify.sh` results:

```
=== Relay ===          PASS
=== Lint ===           PASS
=== Format ===         PASS
=== TypeScript ===     FAIL (pre-existing baseline)
=== Vite warmup paths === PASS
```

The TypeScript failures are the same pre-existing baseline (15 errors in `SessionLauncherPage.tsx`, `StatisticsPage.tsx`, `UserCredentialsPage.tsx`, `UserSettingsPage.tsx`, `VFolderNodeListPage.tsx`) listed in the original PR description. Baseline `tsc --noEmit` (react/) — original PR reported 835; current PR reports **824** (net -11), thanks to (a) the deleted `StorageHostSettingPage.tsx`, (b) the rewrite removing several wrapper components, and (c) the `BAIDomainSelect` `Omit<'role'>` drive-by clearing 3 baseline errors from sibling call sites. No new TS errors introduced.

## Acceptance criteria

- [x] Clicking the gear icon on any storage row opens `StorageHostDetailDrawer` with the storage host id as the drawer title.
- [x] Drawer body renders `StorageHostResourcePanel` (Endpoint / Backend / Capabilities / Usage) as a top section.
- [x] Drawer body renders a `Tabs` with "Permissions" (superadmin, default) and "Capacity Setting".
- [x] Capacity Setting tab renders `<StorageHostSettingsPanel/>` with the "quota not supported" `<Empty/>` (wrapped in `BAICard`) fallback preserved.
- [x] Permissions tab renders three `BAICard`s (Domains / Projects / Keypair Resource Policies), each with a select in `extra` and a single-row table in the body.
- [x] Each table fetches its entity lazily only after a selection (no idle network).
- [x] Each table shows host-allowed status under the name and check/close icons in the permission columns.
- [x] Edit action opens `StoragePermissionEditModal` with a master "전체" + indeterminate + `n/m` count, "권한" section label, and per-permission checkboxes.
- [x] Save issues the correct V1 `modify_*` mutation with a JSONString payload that preserves permissions for other storage hosts (built from V2 structured entries for Project/KRP).
- [x] Save success bumps the table's `fetchKey` so the new permission state is reflected without a manual refresh.
- [x] Mount-in-session warning runs as `modal.confirm` (reversible action) inside the edit modal.
- [x] `StorageHostSettingPage.tsx` deleted; `/storage-settings/:hostname` redirects to `/agent`.
- [x] All new `.tsx` files start with `'use memo'`; no new `useMemo` / `useCallback`.
- [x] New i18n keys added and propagated to all 21 locales (en/ko hand-written, 19 others auto-translated).

[FR-2969]: https://lablup.atlassian.net/browse/FR-2969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


react/src/components/StorageHostDetailDrawerContent.tsx
```diff
@@ -0,0 +1,111 @@
...
+          {
+            key: 'permissions' as const,
+            label: t('storageHost.tab.Permissions'),
+            children: (
+              <ErrorBoundaryWithNullFallback>
+                <Suspense fallback={<Skeleton active />}>
+                  <StorageHostPermissionPanel
+                    storageHostId={storageHostId}
+                    fetchKey={fetchKey}
+                  />
```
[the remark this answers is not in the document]
ironAiken2: Good catch — wired `onAfterPermissionSave` from `StorageHostDetailDrawer` down through `StorageHostDetailDrawerContent` to `StorageHostPermissionPanel`'s `onAfterSave`. On successful save the drawer now calls `updateFetchKey()` inside `startTransition`, which propagates a fresh `fetchKey` to both the `storage_volume` query and the permission panel queries, replacing the stale `allowed_vfolder_hosts` strings in the Relay store. The drawer header refresh button continues to do the same thing manually. Fixed in e53500d7e.

react/src/components/StorageHostPermissionPanel.tsx
```diff
@@ -0,0 +1,628 @@
...
+          render: (_value: unknown, row: T) => {
+            const key = rowKey(row);
+            const set = state[key] ?? new Set<string>();
+            return (
+              <Checkbox
+                checked={set.has(permKey)}
+                onChange={(e: CheckboxChangeEvent) =>
+                  handleCellToggle(section, key, permKey, e.target.checked)
+                }
+              />
```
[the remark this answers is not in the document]
ironAiken2: Agreed — added `aria-label="{row name} — {permission label}"` to each permission checkbox. The matrix now identifies every cell to screen readers (e.g. "default — Mount" for the Mount column on the `default` domain row). Required threading a new `rowName` accessor through `buildPermissionColumns` alongside the existing `rowKey`, since for groups the row key is `id` but the human-readable name is `name`. Fixed in e53500d7e.

ironAiken2:
## Verify Spec Results

**Spec**: [`.specs/FR-2969-storage-host-permission-migration/spec.md`](https://github.com/lablup/backend.ai-webui/blob/fr-2969-storage-host-permission-migration/.specs/FR-2969-storage-host-permission-migration/spec.md)

| Category | PASS | PARTIAL | MISS |
|---|---|---|---|
| UI behavior | 13 | 0 | 0 |
| Migration / removal | 4 | 0 | 0 |
| Role gating | 2 | 0 | 0 |
| i18n | 3 | 0 | 0 |
| Code quality | 4 | 1 | 0 |
| **Total** | **26** | **1** | **0** |

**Spec compliance: 26/27 (96%)**

### The 1 PARTIAL

> `bash scripts/verify.sh` ends with `=== ALL PASS ===`

Relay / Lint / Format / Vite all PASS. TypeScript FAILs on **15 pre-existing baseline errors** in unrelated files (`SessionLauncherPage.tsx`, `StatisticsPage.tsx`, `UserCredentialsPage.tsx`, `UserSettingsPage.tsx`, `VFolderNodeListPage.tsx`). This PR net-removes 3 baseline errors (the deleted `StorageHostSettingPage.tsx` contributed 3 of them) and introduces zero new TS errors.

### Security

CRITICAL 0 / HIGH 0 / MEDIUM 0 / LOW 2 (already noted by pr-reviewer in batch-review). JSON merge preservation, defensive `JSON.parse`, superadmin gating, and mount-in-session warning trigger logic all verified against the spec's stated security concerns.

### Type B additions (impl beyond spec)

These came out of agent reviews and are improvements over the spec's literal requirements:

- `useEffectEvent`-based baseline + editable state separation, so dirty diffs survive a mid-edit refetch (`store-and-network`).
- `aria-label="{row name} — {permission label}"` on every checkbox in the permission matrix (Copilot a11y feedback).
- `NoChanges` info toast when the user clicks **Update** with no dirty rows.
- Per-row failure logging via `useBAILogger` so support has the full picture when multiple mutations partially fail.

No spec updates needed — these are implementation details below the spec's abstraction level.

### Follow-up issues

None. Every Must-Have acceptance criterion is addressed.

### Verdict

**PASS WITH NOTES** — ready for human review and merge.

---
🤖 Generated by /fw:pipeline verify-spec stage

ironAiken2:
* **#7672** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7672?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7658** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7658?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7626** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7626?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7587** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7587?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7587?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7587).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/MainLayout/WebUISider.tsx
```diff
@@ -248,12 +248,7 @@ const WebUISider: React.FC<WebUISiderProps> = (props) => {
             {adminHeader}
             <BAIMenu
               collapsed={props.collapsed}
-              selectedKeys={[
-                // TODO: After matching first path of 'storage-settings' and 'agent', remove this code
-                location.pathname.split('/')[1] === 'storage-settings'
-                  ? 'agent'
-                  : location.pathname.split('/')[1],
-              ]}
```
agatha197, line 258: Is it okay to remove? I think it's fallback router for legacy version.

agatha197 (changes requested):

react/src/components/MainLayout/WebUISider.tsx
ironAiken2: Safe to remove — this was a sidebar-highlight fallback that mapped the legacy `/storage-settings/:hostname` URL to the `agent` menu key, needed only while that was a standalone page. The route now redirects to `/agent` (see `routes.tsx`), so the path is never `storage-settings` at runtime and the ternary is unreachable dead code. Legacy bookmarks / deep links still resolve via that redirect.

packages/backend.ai-ui/src/components/fragments/BAIAdminKeypairResourcePolicySelect.tsx
```diff
@@ -0,0 +1,221 @@
...
+import * as _ from 'lodash-es';
+import {
+  useDeferredValue,
+  useImperativeHandle,
+  useOptimistic,
+  useRef,
+  useState,
+  useTransition,
+} from 'react';
+import { useTranslation } from 'react-i18next';
```
agatha197: please use `BAITrans` instead of `useTranslation`

agatha197 (changes requested):

react/src/components/StorageHostSettingsPanel.tsx
```diff
@@ -35,146 +31,59 @@ const StorageHostSettingsPanel: React.FC<StorageHostSettingsPanelProps> = ({
...
+        <BAIRadioGroup
+          options={[
+            { label: t('storageHost.ForUser'), value: 'user' },
+            { label: t('storageHost.ForProject'), value: 'project' },
+          ]}
+          value={currentSettingType}
+          onChange={(e) => {
+            setCurrentSettingType(e.target.value as QuotaScopeType);
+            setUserId(undefined);
+            setProjectId(undefined);
```
agatha197: This resets **both** `userId` and `projectId` to `undefined` on every radio switch, which contradicts the comment just above it ("Keep separate states per scope type so switching... preserves whatever the user previously picked") and the PR description ("switching the User/Project radio preserves each side's last pick"). As written, each side's previous selection is lost on switch. If preserving each side's pick is the intent, drop these two `setXxx(undefined)` calls; otherwise update the comment/PR body to match the reset behavior.

react/src/components/StorageHostDetailDrawerContent.tsx
```diff
@@ -0,0 +1,113 @@
...
+            children: isQuotaSupportedStorage ? (
+              <ErrorBoundaryWithNullFallback>
+                <Suspense fallback={<Skeleton active />}>
+                  <StorageHostSettingsPanel
+                    storageVolumeFrgmt={storageVolume}
+                  />
+                </Suspense>
+              </ErrorBoundaryWithNullFallback>
+            ) : (
+              <BAICard>
```
agatha197: Per the project's `use-bai-card.md` rule, every non-tabbed `BAICard` must pass `styles={{ body: { paddingTop: 0 } }}` so the body sits flush against the header. This fallback card has no `tabList`, so it should be `<BAICard styles={{ body: { paddingTop: 0 } }}>`.

react/src/components/QuotaScopeTable.tsx
```diff
@@ -0,0 +1,202 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+import { QuotaScopeTableQuery as QuotaScopeTableQueryType } from '../__generated__/QuotaScopeTableQuery.graphql';
+import { QuotaScopeTableUnsetMutation } from '../__generated__/QuotaScopeTableUnsetMutation.graphql';
+import { bytesToGB } from '../helper/index';
+import QuotaSettingModal from './QuotaSettingModal';
+import { CloseOutlined, EditFilled } from '@ant-design/icons';
+import { Empty, message } from 'antd';
```
agatha197: Prefer `const { message } = App.useApp()` over antd's static `message` import — the static API bypasses the app's ConfigProvider theme/context. The new permission components in this PR already use `App.useApp()`, so this is a consistency nit.

react/src/components/StoragePermissionEditModal.tsx
```diff
@@ -0,0 +1,143 @@
...
+      } else {
+        message.error(result.msg || t('storageHost.permission.SaveFailed'));
+      }
+    } finally {
+      setIsSaving(false);
+    }
+  };
+
+  const totalCount = permissionKeys.length;
+  const selectedCount = editedKeys.length;
```
agatha197, line 85: Minor hardening (LOW): `selectedCount` is `editedKeys.length` while `totalCount` is `permissionKeys.length`. If `editedKeys` ever contained a key not in `permissionKeys`, `selectedCount` could exceed `totalCount`, making both `allSelected` and `indeterminate` false (master checkbox blank despite items selected). Not reachable with the current data flow, but an `_.intersection(editedKeys, permissionKeys)` guard would make it robust.

agatha197 (commented):
Review notes — a few things worth addressing before merge. Great consolidation overall; the lazy-per-entity tables and idle-gating look solid. Comments below are inline. The one runtime check I couldn't verify statically (whether `buildAllowedHostsPayloadFromV2` drops zero-permission hosts) is noted on `storageHostPermission.ts` separately.

react/src/helper/storageHostPermission.ts
```diff
@@ -0,0 +1,162 @@
...
+  allowedVfolderHosts:
+    | ReadonlyArray<V2AllowedVfolderHostEntry>
+    | null
+    | undefined,
+  storageHostId: string,
+  enabledKeys: string[],
+): string => {
+  const merged: Record<string, string[]> = {};
+  for (const entry of allowedVfolderHosts ?? []) {
+    merged[entry.host] = entry.permissions.map(v2PermissionToKey);
```
agatha197, line 158: **Please verify with a live round-trip before merge (potential data-corruption).** The doc comment says other hosts are "preserved verbatim," but this loop actually *re-derives* every other host from the V2 `allowedVfolderHosts` array. That is lossless only if V2 returns **every** host present in the V1 JSON — including hosts allowed with an *empty* permission array (V1 distinguishes "absent key = not allowed" from "empty array = allowed, zero perms"). If V2 omits zero-permission hosts, saving any Project/KRP permission would silently drop those hosts' allow-entries — the same class of bug this PR set out to fix, shifted to the empty-array case.\n\nSuggested manual test: give an entity host A (some perms) **and** host B (zero perms, allowed), edit/save host A, and confirm host B's allow-entry survives. (The V1 Domain path via `buildAllowedHostsPayload` spreads the raw JSON and is not affected.)

packages/backend.ai-ui/src/components/fragments/BAIAdminKeypairResourcePolicySelect.tsx
ironAiken2: This BUI component still uses `useTranslation` because this stack branched **before** FR-2986 landed on `main` — `useBAIi18n` (and the ESLint rule that enforces it for `packages/backend.ai-ui/**`) does not exist on the stack's base yet, so switching now fails the type-check (`Cannot find module '../../hooks/useBAIi18n'`). I'll switch to `useBAIi18n()` when the stack is rebased onto `main` (which brings in the hook + its enforcing lint rule). Leaving this thread open as a tracked follow-up for that rebase.

react/src/components/StorageHostSettingsPanel.tsx
ironAiken2: Fixed in the latest push — removed the two `setUserId(undefined)` / `setProjectId(undefined)` calls from the radio `onChange`. `rawScopeEntityId` already selects the active side's state, so each side's previous pick is now preserved on switch, matching the comment and the PR description.

react/src/components/StorageHostDetailDrawerContent.tsx
ironAiken2: Fixed — the fallback `BAICard` (no `tabList`) now passes `styles={{ body: { paddingTop: 0 } }}` per the `use-bai-card.md` rule.

react/src/components/QuotaScopeTable.tsx
ironAiken2: Fixed — `QuotaScopeTable` now reads `const { message } = App.useApp()` instead of antd's static `message` import, so notifications respect the app's ConfigProvider theme/context.

react/src/components/StoragePermissionEditModal.tsx
ironAiken2: Good catch. This modal was reworked in the stacked PR #7658 (FR-3010), so the hardening lands there: `selectedCount` is now `_.intersection(editedKeys, permissionKeys).length`, which keeps the master checkbox correct even if a target's `enabled` set carries a stale key not present in `permissionKeys`. Resolving here since #7658 owns the current implementation of this component.

react/src/helper/storageHostPermission.ts
ironAiken2: Valid concern — keeping this **open** pending a backend round-trip I can't run solo. `buildAllowedHostsPayloadFromV2` re-derives every other host from the V2 `allowedVfolderHosts` array, which is lossless only if V2 returns hosts that are allowed with an **empty** permission list (V1 distinguishes absent-key = not-allowed from empty-array = allowed/zero-perms).

**Verification needed (before merge):** does `ProjectV2.storage.allowedVfolderHosts` / `KeypairResourcePolicyV2.allowedVfolderHosts` include zero-permission host entries, or filter them out? Live test: give an entity host A (some perms) **and** host B (zero perms, allowed), save host A, reload, and confirm host B's allow-entry survives. If V2 omits zero-perm hosts, the save flow must thread the original payload through (fetch+merge the V1 JSON, or pass it to the mutation) rather than re-derive from V2.

packages/backend.ai-ui/src/components/fragments/BAIAdminKeypairResourcePolicySelect.tsx
ironAiken2: Done. I synced the whole stack onto the latest `main` (a `gt restack` onto `origin/main`, 6 commits ahead), which brings in FR-2986's `useBAIi18n` hook and its enforcing lint rule. `BAIAdminKeypairResourcePolicySelect` now uses `import { useBAIi18n } from '../../hooks/useBAIi18n'` + `const { t } = useBAIi18n()` instead of `react-i18next`. `bash scripts/verify.sh` is green across the restacked stack (the FR-2986 `no-restricted-imports` rule now passes for this file).

react/src/helper/storageHostPermission.ts
ironAiken2: Verified live on a dev server (fr-2969 / PR #7587) — **no data loss; non-issue.**

Test on a Keypair Resource Policy: added a previously-not-allowed host -> shows "Host allowed" -> unchecked **all** permissions and saved (host now allowed with an empty permission set) -> still "Host allowed" -> reload -> still "Host allowed".

The zero-permission host survives the save->reload round-trip, which proves the V2 `allowedVfolderHosts` **returns** host entries that are allowed with an empty permission list rather than filtering them out. Since `buildAllowedHostsPayloadFromV2` re-derives the other hosts from exactly that V2 array, a zero-permission host is present in the array at edit time and is therefore preserved when saving a *different* host — the data-loss path described here cannot occur. Resolving.

agatha197 (changes requested):
I think it's better to leave redirect link for legacy url instead of removing.

react/src/routes.tsx
```diff
@@ -756,12 +753,7 @@ export const mainLayoutChildRoutes: RouteObject[] = [
   },
   {
     path: '/storage-settings/:hostname',
-    handle: { labelKey: 'storageHost.StorageSetting' },
-    element: (
-      <Suspense fallback={<Skeleton active />}>
-        <StorageHostSettingPage />
-      </Suspense>
-    ),
+    element: <WebUINavigate to="/agent" replace />,
```
agatha197:
```suggestion
    element: <WebUINavigate to="/agent?tab=storages" replace />,
```

yomybaby (approved):
LGTM! don't forget to resolve other reviews before merging.

react/src/routes.tsx
ironAiken2: Applied your suggestion — the legacy `/storage-settings/:hostname` route now redirects to `/agent?tab=storages` (was `/agent`), preserving the redirect for the legacy URL instead of removing it. Thanks!

agatha197 (approved):
LGTM

PR #7595, opened 2026-05-26
fix(FR-2987): restore host i18n Context inside BAIConfigProvider

yomybaby:
## Summary

`fix(FR-2925)` (PR #7491) upgraded TypeScript to `~6.0.3` for both the host app (`react/`) and `backend.ai-ui`. Before that change the two packages had different TS versions (host: 5.9.3, BUI: 5.7.3), which caused pnpm to keep **two separate physical copies** of `react-i18next` in the store — one per unique TypeScript peer signature. Two copies = two distinct React Context objects = proper isolation between host i18n and BUI i18n.

After aligning TS versions, pnpm deduplicates to **one physical copy** of `react-i18next`. There is now only one React Context object for i18n. `BAIConfigProvider` internally renders `<I18nextProvider i18n={buiI18n}>` which becomes the sole i18n provider for all its children. Host components calling `useTranslation()` resolve keys against BUI's resource set (`packages/backend.ai-ui/src/locale/`), which does not contain host app keys — so raw key strings like `login.Login`, `login.Password`, `login.Endpoint`, etc. are rendered on screen.

`vite.config.ts` already has a comment that documents this exact failure mode (the `optimizeDeps.exclude` block).

## Fix

Extract a `<HostI18nProvider>` helper that wraps children with `<I18nextProvider i18n={i18n}>` (the host i18next singleton, initialized with `i18next-http-backend` loading `/resources/i18n/{{lng}}.json`). Apply it immediately inside **every** `<BAIConfigProvider>` instance to restore the host context for host children.

```tsx
<BAIConfigProvider ...>
  <HostI18nProvider>  {/* re-assert host i18n to override BUI's leak */}
    ...{children}...
  </HostI18nProvider>
</BAIConfigProvider>
```

Two instances are patched:
1. `DefaultProvidersForReactRoot` in `react/src/components/DefaultProviders.tsx`
2. The inner `BAIConfigProvider` in `react/src/components/LoginView.tsx` that wraps `LoginFormPanel` (the cause of the remaining raw keys after the first patch — the LoginView creates its own BAIConfigProvider for the login overlay's message z-index)

## Root cause bisection

- **v26.4.8-rc.6** (before `64931aafc`): i18n works — pnpm keeps two separate `react-i18next` copies
- **latest main** (after `64931aafc`): raw i18n keys — pnpm collapses to one copy

The regression-introducing commit: `64931aafc fix(FR-2925): jsx type error and update typescript to v6`

## Verification

[...]

PR #7616, opened 2026-05-27
refactor(FR-2986): use explicit i18n instance binding in useBAIi18n

### LINKED ISSUE #7615: Refactor BUI i18n to use explicit instance binding (eliminate Context leak)
## Why this work is needed

BUI (`packages/backend.ai-ui`) currently exposes BUI translations through `<I18nextProvider i18n={buiI18n}>` wrapping inside `BAIConfigProvider`. This pattern depends on host and BUI ending up with two physically separate copies of `react-i18next` in the bundle — only then do they have distinct React Context objects, which is what previously kept host translations and BUI translations from leaking into each other.

That assumption broke in FR-2925 (TypeScript peer-version unification). pnpm now dedupes `react-i18next` into a single physical module, so there is only one React Context object — and `BAIConfigProvider`'s `<I18nextProvider>` ends up shadowing the host's i18n for every child, causing raw key strings like `login.Login` to render on the login screen.

FR-2987 / PR #7595 ships a fast workaround (`HostI18nProvider` re-asserts the host context inside every `BAIConfigProvider`). That keeps the user-facing bug from shipping, but it leaves the brittle two-Context-instance design in place — any future deduplication change or any place that re-instantiates `BAIConfigProvider` (e.g. `LoginView.tsx:1080`) will re-introduce the same class of bug.

This work removes the design dependency entirely by switching BUI components to explicitly bind to BUI's i18n instance, bypassing React Context for the BUI side.

## Expected outcomes

- New `useBAIi18n()` hook in BUI that calls `useTranslation(undefined, { i18n: buiI18n })` — explicit instance binding, no Context lookup.
- All 71 BUI components migrated from `useTranslation()` to `useBAIi18n()`.
- The single `<Trans>` usage in BUI (BAIDeleteConfirmModal) passes `t={t`} so it inherits the BUI-bound `t`.
- `BAIConfigProvider` drops its `<I18nextProvider i18n={buiI18n}>` wrap, keeping only the `i18n.changeLanguage(locale.lang)` `useEffect` that mirrors language changes onto BUI's instance.
- `vite.config.ts`: `i18next` and `react-i18next` removed from `optimizeDeps.exclude`, added to `resolve.dedupe`. The two-physical-copies hack is no longer required.
- No host-side changes: `DefaultProviders.tsx` does not need `ns`, `nsSeparator`, or `registerBAIResources`.
- `HostI18nProvider` workaround from PR #7595 is no longer needed (host translations are no longer shadowed).
- Login screen renders all i18n keys correctly (verified on dev server, zero raw keys).
- BUI is now usable from any host that does not run `react-i18next` at all — useful for future external distribution.

JIRA Issue: FR-2986

yomybaby:
Resolves #7615 (FR-2986)

## Summary

BUI (`packages/backend.ai-ui`) previously exposed its translations by wrapping children with `<I18nextProvider i18n={buiI18n}>` inside `BAIConfigProvider`. That design relied on host and BUI ending up with **two physically separate copies** of `react-i18next` so each side had a distinct React Context. After FR-2925 unified TypeScript peer versions, pnpm started deduping `react-i18next` into a single physical module → only one Context object → BUI's `<I18nextProvider>` shadowed the host's, causing raw key strings (`login.Login`, `login.Endpoint`, …) on the login page (FR-2987).

PR #7595 (FR-2987, **already merged**) ships a fast workaround that re-asserts the host i18n Context inside `BAIConfigProvider`. This PR is the **structural alternative** — instead of patching around the leak, it removes the React-Context dependency entirely on BUI's side.

## Approach (Option C: explicit instance binding)

BUI components now access BUI's i18n instance **directly** via `useTranslation(undefined, { i18n: buiI18n })`, exposed as an **internal-only** `useBAIi18n()` hook. That call signature tells react-i18next to skip Context lookup and use the supplied instance — so BUI is unaffected by whatever i18n provider (or absence of one) the host has.

### Code changes

- **New internal hook** `packages/backend.ai-ui/src/hooks/useBAIi18n.ts` — explicit-instance wrapper around `useTranslation`. **Not exported from the package's public surface** (`hooks/index.ts`); it is an implementation detail of how BUI binds to its own instance.
- **ESLint enforcement** in `packages/backend.ai-ui/eslint.config.js` — `no-restricted-imports` forbids `useTranslation` / `withTranslation` / `Translation` / `I18nextProvider` imports from `react-i18next` anywhere under `packages/backend.ai-ui/src/**`. `useBAIi18n.ts` itself is exempt. This prevents the original anti-pattern from reappearing in future BUI code.
- **`BAIConfigProvider`** — drops `<I18nextProvider i18n={buiI18n}>` wrap and the matching `react-i18next` import. Keeps the `useEffect` that mirrors `locale.lang` onto `buiI18n.changeLanguage(...)` and `dayjs.locale(...)` (this is the language-sync mechanism for the explicit-instance design).
- **72 BUI source files** migrated `useTranslation()` → `useBAIi18n()` with depth-aware relative import paths (including `src/tests/BAITestButton.tsx`).
- **`BAIDeleteConfirmModal`** — the single `<Trans>` usage now passes `t={t}` so it inherits the BUI-bound translation function (otherwise Trans would re-enter the Context path).
- **`react/vite.config.ts`** — drops `i18next` / `react-i18next` from `optimizeDeps.exclude`, adds them to `resolve.dedupe`. The two-physical-copies workaround is no longer required.
- **`react/src/components/DefaultProviders.tsx`** — removes the inline `<I18nextProvider i18n={i18n}>` wrap added by PR #7595, since BUI no longer leaks Context. Comment updated to point at `useBAIi18n.ts`.

## Comparison with PR #7595 (FR-2987 workaround, already merged)

| Aspect | PR #7595 (HostI18nProvider) | This PR (#7616) |
|---|---|---|
| Approach | Patch each `BAIConfigProvider` site by re-asserting host Context | Remove the leaking design at the source |
| BUI uses React Context for i18n | Yes (still) | **No** (explicit instance binding) |
| Future `BAIConfigProvider` instances | Need re-assertion added manually | Automatic — no per-site fix needed |
| `vite.config.ts` `optimizeDeps.exclude` hack | Still required | **Removed** |
| Anti-pattern recurrence prevention | Convention comment only | **ESLint rule enforced** |
| `useBAIi18n` exposed publicly | N/A | **No** — internal-only implementation detail |

## Verification

- `bash scripts/verify.sh` ends with `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Vite warmup all green).
- Dev server: `https://option-c.localhost:1356/` (worktree `option-c-explicit-instance`)
- Playwright headless check on the login page: **0 raw i18n keys** found in body text.
- ESLint guard verified: temporarily injecting `import { useTranslation } from "react-i18next"` into a BUI source file produces the expected `no-restricted-imports` error with the `useBAIi18n` migration message.

## Test plan

- [ ] Verify login page renders with no raw `*.key` strings in EN / KO / JA
- [ ] Verify language switching from user settings still propagates to BUI components (e.g. `BAITagList` Copy-All tooltip, `BAIModal` minimize tooltip)
- [ ] Verify Storybook still renders BUI components in isolation (BUI's own i18n instance + dayjs are still initialized in `locale/index.ts`)
- [ ] CI: `pnpm run -r --stream build` and `bash scripts/verify.sh` pass

## Architectural notes

- BUI is now usable from a host that does **not** use `react-i18next` at all (or uses a different version) — useful if BUI is ever distributed as an external npm package.
- Two physical i18n instances coexist (host + BUI) but they never compete because Context resolution is bypassed in BUI components. The cost is duplicated language-resource memory; given BUI bundles are small and JSON parsing is one-time, this is negligible.
- The `useBAIi18n` hook is also a step toward an Adapter-pattern future where the i18n backend behind it could be swapped without touching the 71 BUI call sites.

### DIFF
--- a/packages/backend.ai-ui/eslint.config.js
+++ b/packages/backend.ai-ui/eslint.config.js
@@ -28,6 +28,55 @@ export default [
     },
   },
 
+  // Enforce that BUI components access translations through the internal
+  // `useBAIi18n` hook only. Direct imports of i18n primitives from
+  // `react-i18next` would re-introduce React-Context-based lookup, which
+  // (under pnpm dedup) shadows the host app's i18n and surfaces raw keys
+  // on screen — the exact class of bug FR-2986 eliminates.
+  //
+  // All three banned bindings (`useTranslation` hook, `withTranslation` HOC,
+  // `<Translation>` render-prop) are different injection mechanisms for the
+  // same thing — a Context-derived `t`. The project is fully function-
+  // component / hook-based (`'use memo'` directive), so `useBAIi18n()` is
+  // the single replacement for all three; we do not maintain a separate
+  // BUI-instance-bound HOC or render-prop component.
+  //
+  // `useBAIi18n.ts` and `BAITrans.tsx` are exempted because they are the
+  // two places that are *allowed* to call into `react-i18next` directly
+  // (with explicit `{ i18n }` binding). Every other BUI source file routes
+  // i18n access through them.
+  {
+    files: ["src/**/*.{ts,tsx}"],
+    ignores: [
+      "src/hooks/useBAIi18n.ts",
+      "src/components/BAITrans.tsx",
+      "**/*.test.*",
+      "**/*.stories.*",
+      "**/__test__/**",
+    ],
+    rules: {
+      "no-restricted-imports": [
+        "error",
+        {
+          paths: [
+            {
+              name: "react-i18next",
+              importNames: [
+                "useTranslation",
+                "withTranslation",
+                "Translation",
+                "Trans",
+                "I18nextProvider",
+              ],
+              message:
+                "Use `useBAIi18n` from `<relative path>/hooks/useBAIi18n` (replaces useTranslation / withTranslation / Translation) or `BAITrans` from `<relative path>/components/BAITrans` (replaces <Trans>) instead. BUI components must bind explicitly to BUI's i18n instance — see FR-2986 / packages/backend.ai-ui/src/hooks/useBAIi18n.ts.",
+            },
+          ],
+        },
+      ],
+    },
+  },
+
   {
     files: ["**/*.test.*", "**/*.stories.*", "**/__test__/**"],
     languageOptions: {
--- a/packages/backend.ai-ui/src/components/BAIBoardItemErrorBoundary.tsx
+++ b/packages/backend.ai-ui/src/components/BAIBoardItemErrorBoundary.tsx
@@ -1,3 +1,4 @@
+import { useBAIi18n } from '../hooks/useBAIi18n';
 import BAIAlertIconWithTooltip from './BAIAlertIconWithTooltip';
 import BAIBoardItemTitle, {
   type BAIBoardItemTitleProps,
@@ -5,7 +6,6 @@ import BAIBoardItemTitle, {
 import { theme } from 'antd';
 import React, { type PropsWithChildren } from 'react';
 import { ErrorBoundary } from 'react-error-boundary';
-import { useTranslation } from 'react-i18next';
 
 interface BAIBoardItemErrorBoundaryProps extends PropsWithChildren {
   title?: BAIBoardItemTitleProps['title'];
@@ -19,7 +19,7 @@ const BAIBoardItemErrorBoundary: React.FC<BAIBoardItemErrorBoundaryProps> = ({
   children,
   style,
 }) => {
-  const { t } = useTranslation();
+  const { t } = useBAIi18n();
   const { token } = theme.useToken();
   return (
     <ErrorBoundary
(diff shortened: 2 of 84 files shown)


nowgnuesLee (changes requested):
Could you please remove the `buiTranslations` that was added to `bui-languages.ts`?

nowgnuesLee (approved):
LGTM

nowgnuesLee:
* **#7616** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7616?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7616?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7616).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #7648, opened 2026-05-29
fix(FR-3005): default initial delay to 1800s, show all revision detail fields, auto-open detail after create

### LINKED ISSUE #7643: Increase initial delay default to 1800s, show all revision detail fields, auto-open revision detail after creation
## Context

It is reported that model services fail during provisioning because the "Initial Delay Time" default value of 60 seconds is too short. Services are stopped and recreated mid-provisioning before the health check can succeed.

## Tasks

1. **Change default value of "Initial Delay Time" from 60 to 1800 seconds**
   - In `DeploymentAddRevisionModal.tsx`, the form field `commandInitialDelay` currently defaults to `60`.
   - Change to `1800` as a short-term fix while a proper feature-flag mechanism is developed.

1. **Show all available fields in revision detail view**
   - The revision detail drawer/view does not currently display "Initial Delay Time" and possibly other health check fields.
   - Show ALL displayable fields (initialDelay, maxRetries, interval, maxWaitTime, health check path, etc.) in the revision detail.

1. **Auto-open revision detail drawer after revision creation**
   - After a new revision is successfully created, the detail drawer for that revision should automatically open.

## Acceptance Criteria

- Creating a new revision shows "Initial Delay Time" pre-filled with 1800 (not 60)
- Revision detail drawer shows all health check fields including initialDelay
- After revision creation, the revision detail drawer opens automatically for the newly created revision

JIRA Issue: FR-3005

agatha197:
Resolves #7643 (FR-3005)

## Background

Model services were failing during provisioning: the **Initial Delay Time (초기 지연 시간)** default of **60s** was too short for large models to finish loading before the first health check, so services were dropped and re-created mid-provisioning. 

### 1. Default initial delay 60s → 1800s (short-term fix)
`DeploymentAddRevisionModal` custom form now defaults `commandInitialDelay` to **1800**. This is a stopgap until a shared backend/frontend default (feature flag) lands — see the thread discussion. The change is confined to the create form's `initialValues`; the edit/prefill path still reads the persisted value, so existing revisions are unaffected.

### 2. Show all displayable model-service fields in the revision detail
The revision detail view (`DeploymentRevisionDetail`) previously omitted several persisted fields. Added:
- `service.shell`
- `service.preStartActions` (action + args)
- `healthCheck.expectedStatusCode`
- extra-mount `mountPerm` (was already fetched, now rendered as a tag)
- `resourceConfig` shared memory (`shmem`) and `modelRuntimeConfig.inferenceRuntimeConfig` (runtime-specific JSON config)

Per review, model-card `metadata` (author/version/task/…) was intentionally **not** added — only the runtime config and shmem were requested. Image-catalog metadata (registry/permission/aliases) and the `UNRELEASED` `subpath` remain excluded for the same reasons as before.

This rounds out the health-check / service configuration block so users can confirm exactly what was persisted — directly addressing the "초기 지연 시간 등이 안 보인다" feedback. (Image-catalog metadata such as registry/permission/aliases and the `UNRELEASED` `subpath` field were intentionally left out — the former describes the image in the registry rather than this revision's config; the latter is not safe against current manager versions.)

### 3. Auto-open the revision detail drawer after creation
After a revision is created, its detail drawer now opens automatically. The `addModelRevision` mutation already returns the full `...DeploymentRevisionDetail_revision` fragment, so the drawer renders straight from the Relay store (no extra fetch). Wired via `onRequestClose(success, createdRevision)` → `DeploymentDetailPage`.

## i18n
Three new `modelService` keys (`ExpectedStatusCode`, `PreStartActions`, `Shell`) added to `en.json` and translated into all 20 other locales.

## Verification
`bash scripts/verify.sh` — Lint ✅, Format ✅, TypeScript ✅. (Relay artifacts regenerated and committed; the only "out of sync" report during the run was the expected uncommitted-then-committed `__generated__` diff.)

### 1. Add Revision modal — Initial Delay now defaults to 1800 (was 60)
initial-delay-1800

### 2. Revision detail — Shared Memory, Runtime Configuration, and extra-mount permission (`READ_ONLY`) now shown
revision-detail-shmem-runtimeconfig

### 3. Revision detail — Pre-Start Actions + Expected Status Code now shown
revision-detail-healthcheck

_The revision created via the modal auto-opens this same detail drawer (task 3); it renders the identical `DeploymentRevisionDetail` component shown above. Captured against the test backend with admin on an existing revision (Initial Delay 60 here predates this change)._


agatha197:
* **#7648** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7648?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7648?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7648).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/pages/DeploymentDetailPage.tsx
```diff
@@ -83,6 +85,12 @@ const DeploymentDetailPage: React.FC = () => {
 
   const revisionsSectionRef = useRef<HTMLDivElement>(null);
   const accessTokensSectionRef = useRef<HTMLDivElement>(null);
+  // Fragment ref of the revision just created via the Add Revision modal.
+  // When set, its detail drawer opens automatically so the user can confirm
+  // the configuration that was actually persisted (FR-3005) — notably fields
+  // like "Initial Delay" that the form prefills but does not echo back.
+  const [createdRevisionFrgmt, setCreatedRevisionFrgmt] =
+    useState<DeploymentRevisionDetail_revision$key | null>(null);
```
nowgnuesLee: you don't need to import another component's key type for typing.

react/src/pages/DeploymentDetailPage.tsx
```diff
@@ -210,9 +218,19 @@ const DeploymentDetailPage: React.FC = () => {
     startRefetchTransition(() => updateFetchKey());
   };
 
-  const handleAddRevisionRequestClose = (success?: boolean) => {
+  const handleAddRevisionRequestClose = (
+    success?: boolean,
+    createdRevision?: DeploymentRevisionDetail_revision$key | null,
```
nowgnuesLee: here too.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -131,7 +132,14 @@ export type PresetFormValues = {
...
 
 interface DeploymentAddRevisionModalProps extends BAIModalProps {
-  onRequestClose: (success?: boolean) => void;
+  // `createdRevision` is the fragment ref of the revision just added (taken
+  // straight from the `addModelRevision` mutation response). The caller uses
+  // it to open the revision detail drawer right after a successful create
+  // (FR-3005). It is undefined on cancel/close and on the create failure path.
+  onRequestClose: (
+    success?: boolean,
+    createdRevision?: DeploymentRevisionDetail_revision$key | null,
```
nowgnuesLee: You don't need to import another component's key type for typing.

nowgnuesLee (changes requested):
LGTM

react/src/pages/DeploymentDetailPage.tsx
agatha197: Fixed in latest commit. Removed the cross-component `DeploymentRevisionDetail_revision$key` import here. The page now types the created-revision ref with `DeploymentAddRevisionModalCreatedRevision`, derived from the `addModelRevision` mutation `$data` (which already spreads `DeploymentRevisionDetail_revision`).

react/src/pages/DeploymentDetailPage.tsx
agatha197: Fixed in latest commit. Same change applied to the `onRequestClose` handler param type — it now uses `DeploymentAddRevisionModalCreatedRevision` instead of importing the fragment's `$key`.

react/src/components/DeploymentAddRevisionModal.tsx
agatha197: Fixed in latest commit. The modal no longer imports `DeploymentRevisionDetail_revision$key`; it exports `DeploymentAddRevisionModalCreatedRevision = NonNullable<DeploymentAddRevisionModalAddMutation$data['addModelRevision']>['revision']` and types `onRequestClose` with that. The ref type travels with the mutation response, so no cross-component key-type import is needed (and it keeps working even when the fragment component lives in backend.ai-ui, where the `$key` isn't re-exported).

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1364,13 +1384,15 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
-                </Form.Item>
+                <Suspense fallback={<Skeleton.Input active block />}>
+                  <Form.Item
+                    name="revisionPresetId"
+                    noStyle
+                    rules={[{ required: true }]}
+                  >
+                    <BAIAvailablePresetSelect style={{ flex: 1 }} />
+                  </Form.Item>
+                </Suspense>
```
nowgnuesLee: Let's use <BAISelect loading/>

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1401,20 +1423,22 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
-                  <BAIVFolderSelect
-                    ref={presetVFolderSelectRef}
-                    currentProjectId={currentProjectId ?? undefined}
-                    disabled={!currentProjectId}
-                    excludeDeleted
-                    filter='usage_mode == "model"'
-                    style={{ flex: 1 }}
-                  />
-                </Form.Item>
+                <Suspense fallback={<Skeleton.Input active block />}>
```
nowgnuesLee: here too.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1486,20 +1513,22 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
-                <BAIVFolderSelect
-                  ref={customVFolderSelectRef}
-                  currentProjectId={currentProjectId ?? undefined}
-                  disabled={!currentProjectId}
-                  excludeDeleted
-                  filter='usage_mode == "model"'
-                  style={{ flex: 1 }}
-                />
-              </Form.Item>
+              <Suspense fallback={<Skeleton.Input active block />}>
```
nowgnuesLee: here too.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1538,32 +1567,36 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
-                warningOnly: true,
-                validator: async (_rule, value: string) => {
-                  const variantName = runtimeVariantNameMap[value];
-                  if (variantName && variantName !== 'custom') {
-                    return Promise.reject(
-                      t('modelService.RuntimeVariantDefaultCommandAppliedNote'),
-                    );
-                  }
-                  return Promise.resolve();
+          <Suspense fallback={<Skeleton.Input active block />}>
```
nowgnuesLee: here too.

react/src/pages/DeploymentDetailPage.tsx
nowgnuesLee, line 1: Please check overall whether types for fragments are being imported from other components.

If the current component is spreading the fragment, there is no need to import the type from another component.

nowgnuesLee (changes requested):

react/src/components/DeploymentAddRevisionModal.tsx
agatha197: Good call — done. Replaced the `<Skeleton.Input active block />` Suspense fallback with `<BAISelect loading style={{ flex: 1 }} />`, so the placeholder now matches the select's shape and width while the lazy query resolves (instead of a generic skeleton input). Pushed in the latest commit.

react/src/components/DeploymentAddRevisionModal.tsx
agatha197: Done — same fallback swap here: the Suspense fallback is now `<BAISelect loading style={{ flex: 1 }} />` instead of `<Skeleton.Input active block />`, matching the model-folder select's width. Pushed in the latest commit.

react/src/components/DeploymentAddRevisionModal.tsx
agatha197: Done — same fallback swap here: the Suspense fallback is now `<BAISelect loading style={{ flex: 1 }} />` instead of `<Skeleton.Input active block />`, matching the model-folder select's width. Pushed in the latest commit.

react/src/components/DeploymentAddRevisionModal.tsx
agatha197: Done — the Runtime Variant select's Suspense fallback is now `<BAISelect loading style={{ width: '100%' }} />` instead of `<Skeleton.Input active block />`, so the placeholder matches the full-width select. Pushed in the latest commit.

react/src/pages/DeploymentDetailPage.tsx
agatha197: Checked both files — no cross-component fragment `$key` imports remain. `DeploymentDetailPage` types the created revision with `DeploymentAddRevisionModalCreatedRevision` (derived from the `addModelRevision` mutation `$data`, which spreads the detail fragment), and `DeploymentAddRevisionModal` only imports its own fragment `$key`s. Nothing left to change here.

nowgnuesLee (approved):
LGTN

PR #7650, opened 2026-05-29
feat(FR-3008): auto-expand non-success rows and add expand/collapse-all toggle in scheduling history

### LINKED ISSUE #7649: Auto-expand non-success rows and add expand-all/collapse-all header toggle in scheduling history modals
## Background

Originates from a Teams thread (devops / Frontend): scheduling history lists should surface problems automatically. The original idea (Jonghyun Park): "non-success items should be shown auto-expanded".

This **supersedes FR-2958 / FR-2959**, which were closed as **Not Planned** because they targeted the wrong surface (validation / check result views). The correct target is the **scheduling history modals**.

## Scope

Apply the same behavior to all three scheduling history tables:

- `BAISchedulingHistoryNodes` (session scheduling history)
- `BAIDeploymentSchedulingHistoryNodes` (deployment scheduling history)
- `BAIRouteSchedulingHistoryNodeTable` (route scheduling history)

All three are `BAITable` with expandable rows (`expandable.expandedRowRender` shows `BAISubStepNodes`). A row is expandable only when it has `subSteps`.

## Behavior

1. **Initial state — "collapse success only":** rows that have `subSteps` and whose `result !== 'SUCCESS'` are expanded by default; `SUCCESS` rows stay collapsed. This makes failures / retries / expirations visible at a glance.
1. **Header master toggle:** a toggle in the expand-icon column header. Clicking it switches between **expand all ↔ collapse all** (binary). Direction is derived from current state (if all expandable rows are open → next click collapses all; otherwise → expand all).
1. **Individual rows** remain manually expandable/collapsible as today.
1. **On refetch / data reload:** re-apply the default ("collapse success only"). So after toggling, a refresh returns to the smart default.

## Out of scope

- A persistent 3-mode selector (expand all / collapse all / collapse success only) with last-choice memory (localStorage). Considered but intentionally deferred to keep the control to a single compact header toggle.

## Implementation notes

- Make `expandable.expandedRowKeys` controlled in each table; track via state and `onExpandedRowsChange`.
- Extract shared logic into a small reusable hook in `backend.ai-ui` (accessors for row key, is-expandable, expand-by-default predicate) + the header toggle control, so all three tables reuse it.
- "Success" is defined strictly as `result === 'SUCCESS'`; every other result value (`FAILURE`, `NEED_RETRY`, `EXPIRED`, `GIVE_UP`, `STALE`, `SKIPPED`) expands by default.
- i18n: add labels/tooltips for the toggle (expand all / collapse all).

## Acceptance criteria

- [ ] On opening any of the three scheduling history modals, non-success rows with sub-steps are expanded and success rows are collapsed.
- [ ] The header toggle expands all expandable rows, and toggles back to collapse all.
- [ ] Individual row expand/collapse still works.
- [ ] Refetching the list restores the "collapse success only" default.
- [ ] Behavior is consistent across session / deployment / route history.
- [ ] `bash scripts/verify.sh` passes.

JIRA Issue: FR-3008

agatha197:
Resolves #7649 (FR-3008)

## Summary

Scheduling-history tables — **session**, **deployment**, and **route** — make scheduling problems easy to scan. A kebab (⋮) menu in the expand-column header offers three view actions — **expand all / collapse all / expand errors only** — with the chosen mode persisted per user (default **expand errors only**). In errors-only mode, non-success rows auto-expand and their nested sub-step table is filtered to just the failing / retried steps.

Originates from a Teams thread in devops/Frontend (auto-expand non-success scheduling entries). Supersedes FR-2958 / FR-2959, which were closed as *Not Planned* because they targeted the wrong surface (validation / check-result views).

## Behavior

- **Header kebab menu (⋮):** a vertical-ellipsis hover menu in the expand-column header with three actions — **Expand all**, **Collapse all**, **Expand errors only**. It reads as an action menu, not a stateful toggle: no per-item icons and no "active" highlight.
- **Persisted mode:** the selected mode is stored per user via `useBAISettingUserState('schedulingHistoryExpandMode')` and reused as the default on subsequent opens. Initial default is **expand errors only**.
- **Expand errors only:** rows whose `result !== 'SUCCESS'` are expanded; `SUCCESS` rows stay collapsed. The expanded row's **nested sub-step table hides `SUCCESS` sub-steps**, surfacing only the failing / retried steps.
- **Per-row expand icons** use Ant Design's default `+` / `−` (full expand / collapse). Individual rows remain manually expandable.
- **Result column** shows the result badge only (the previous hover info-icon/tooltip was removed).
- **On refetch:** the current mode is re-applied.

## Table presentation

- History tables use BAITable's default muted header style, which renders correctly in **both light and dark mode** (token-based, dark-mode aware). The status transition is shown as two **flat** columns — **From Status** / **To Status** — reusing the same labels as the property filter for those fields (no grouped two-level header, no cell borders).
- **Resizable** columns, **drag-to-reorder**, and **column show/hide** via `tableSettings`, with the per-column overrides persisted per user (`table_column_overrides.*`).
- **Nested sub-step table:** `Error Code` column now precedes `Message`.

## Implementation

- `useSchedulingHistoryExpandable(dataSource, { mode, onModeChange })` owns `expandedRowKeys` / `onExpandedRowsChange`, exposes the effective `mode`, and renders the kebab header menu. Reused across all three tables.
- Feature lives in dedicated wrappers — `BAISchedulingHistoryTable`, `BAIDeploymentSchedulingHistoryTable`, `BAIRouteSchedulingHistoryTable` — each owning a fragment that spreads the corresponding `*NodesFragment`, runs the hook, and supplies the `expandable` config; the base `*Nodes` components stay pure fragment tables.
- `BAISubStepNodes` gains an `errorsOnly` prop that filters out `SUCCESS` sub-steps.
- The three host modals (`SessionSchedulingHistoryModal`, `DeploymentSchedulingHistoryModal`, `RouteSchedulingHistoryModal`) wire the persisted expand-mode and column-override settings.
- The expanded-row background in `BAITable` is token-based (`colorFillSecondary`, dark-mode aware).
- i18n: `comp:BAITable.ExpandAll` / `CollapseAll` / `ExpandErrorsOnly` / `ExpandOptions` across all languages.

## Verification

`bash scripts/verify.sh` → **ALL PASS** (Relay, Lint, Format, TypeScript).

## Screenshots

Captured (light mode) against mock scheduling-history data — a `NEED_RETRY` and a `FAILURE` row, each with sub-steps mixing `SUCCESS` / `FAILURE` / `NEED_RETRY`, plus `SUCCESS` rows.

### Default — "expand errors only"
Non-success rows (`NEED_RETRY`, `FAILURE`) auto-expanded, `SUCCESS` rows collapsed. The nested sub-step tables show only non-success steps (the `FIFOSequencer` / `PrepareContainer` `SUCCESS` sub-steps are filtered out). Flat **From Status** / **To Status** columns, default muted header, no cell borders, `Error Code` before `Message`, result badge only, kebab (⋮) in the header.

default — expand errors only

### Header kebab menu
The ⋮ hover menu with the three plain-text actions — no icons, no active highlight.

kebab menu

### Expand all
expand all

### DIFF
--- a/packages/backend.ai-ui/src/components/Table/BAITable.tsx
+++ b/packages/backend.ai-ui/src/components/Table/BAITable.tsx
@@ -672,9 +672,9 @@ const useStyles = createStyles(({ token, css }) => ({
       font-weight: 500;
       color: ${token.colorTextTertiary};
     }
-    body:not(.dark-theme) & .ant-table-expanded-row > .ant-table-cell,
-    body:not(.dark-theme) & .ant-table-expanded-row:hover > .ant-table-cell {
-      background: #e3e3e3;
+    & .ant-table-expanded-row > .ant-table-cell,
+    & .ant-table-expanded-row:hover > .ant-table-cell {
+      background: ${token.colorFillSecondary};
     }
   `,
   zeroWithSelectionColumn: css`
--- a/packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryNodes.tsx
+++ b/packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryNodes.tsx
@@ -18,8 +18,6 @@ import {
   BAITable,
   BAITableProps,
 } from '../Table';
-import { BAIColumnGroupType } from '../Table/BAITable';
-import BAISubStepNodes from './BAISubStepNodes';
 import dayjs from 'dayjs';
 import * as _ from 'lodash-es';
 import { graphql, useFragment } from 'react-relay';
@@ -76,9 +74,6 @@ const BAIDeploymentSchedulingHistoryNodes = ({
           result
           errorCode
           message
-          subSteps {
-            ...BAISubStepNodesFragment
-          }
           attempts
           createdAt
           updatedAt
@@ -88,10 +83,7 @@ const BAIDeploymentSchedulingHistoryNodes = ({
     );
 
   const baseColumns = _.map(
-    filterOutEmpty<
-      | BAIColumnType<DeploymentSchedulingHistoryNodeInList>
-      | BAIColumnGroupType<DeploymentSchedulingHistoryNodeInList>
-    >([
+    filterOutEmpty<BAIColumnType<DeploymentSchedulingHistoryNodeInList>>([
       {
         dataIndex: 'updatedAt',
         title: t('comp:BAIDeploymentSchedulingHistoryNodes.UpdatedAt'),
@@ -132,22 +124,16 @@ const BAIDeploymentSchedulingHistoryNodes = ({
         sorter: isEnableSorter('category'),
       },
       {
-        title: t('comp:BAIDeploymentSchedulingHistoryNodes.StatusTransition'),
-        key: 'statusTransition',
-        children: [
-          {
-            key: 'fromStatus',
-            title: t('comp:BAIDeploymentSchedulingHistoryNodes.From'),
-            dataIndex: 'fromStatus',
-            sorter: isEnableSorter('from_status'),
-          },
-          {
-            key: 'toStatus',
-            title: t('comp:BAIDeploymentSchedulingHistoryNodes.To'),
-            dataIndex: 'toStatus',
-            sorter: isEnableSorter('to_status'),
-          },
-        ],
+        key: 'fromStatus',
+        title: t('comp:BAIDeploymentSchedulingHistoryNodes.From'),
+        dataIndex: 'fromStatus',
+        sorter: isEnableSorter('from_status'),
+      },
+      {
+        key: 'toStatus',
+        title: t('comp:BAIDeploymentSchedulingHistoryNodes.To'),
+        dataIndex: 'toStatus',
+        sorter: isEnableSorter('to_status'),
       },
       {
         dataIndex: 'attempts',
@@ -203,18 +189,6 @@ const BAIDeploymentSchedulingHistoryNodes = ({
           (order as (typeof availableDeploymentHistorySorterValues)[number]) ||
             null,
         );
-      }}
-      expandable={{
-        rowExpandable: (record) => !_.isEmpty(record.subSteps),
-        expandedRowRender: (record) => {
-          return (
-            <BAISubStepNodes
-              resizable
-              subStepsFrgmt={record.subSteps}
-              pagination={false}
-            />
-          );
-        },
       }}
       {...tableProps}
     />
--- a/packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryTable.tsx
+++ b/packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryTable.tsx
@@ -1,0 +1,85 @@
+import { BAIDeploymentSchedulingHistoryTableFragment$key } from '../../__generated__/BAIDeploymentSchedulingHistoryTableFragment.graphql';
+import { filterOutNullAndUndefined } from '../../helper';
+import {
+  type SchedulingHistoryExpandMode,
+  useSchedulingHistoryExpandable,
+} from '../../hooks/useSchedulingHistoryExpandable';
+import BAIDeploymentSchedulingHistoryNodes, {
+  BAIDeploymentSchedulingHistoryNodesProps,
+  DeploymentSchedulingHistoryNodeInList,
+} from './BAIDeploymentSchedulingHistoryNodes';
+import BAISubStepNodes from './BAISubStepNodes';
+import * as _ from 'lodash-es';
+import { graphql, useFragment } from 'react-relay';
+
+export interface BAIDeploymentSchedulingHistoryTableProps extends Omit<
+  BAIDeploymentSchedulingHistoryNodesProps,
+  'schedulingHistoryFrgmt' | 'expandable'
+> {
+  schedulingHistoryFrgmt: BAIDeploymentSchedulingHistoryTableFragment$key;
+  expandMode?: SchedulingHistoryExpandMode;
+  onExpandModeChange?: (mode: SchedulingHistoryExpandMode) => void;
+}
+
+/**
+ * Thin wrapper around the pure `BAIDeploymentSchedulingHistoryNodes` fragment
+ * table that adds the optional expand/collapse-all sub-step feature. It owns
+ * its own fragment (selecting the fields the expandable hook needs) and feeds
+ * the resolved rows straight through to the nodes component as the nodes
+ * fragment ref — valid because this fragment spreads the nodes fragment.
+ */
+const BAIDeploymentSchedulingHistoryTable = ({
+  schedulingHistoryFrgmt,
+  expandMode,
+  onExpandModeChange,
+  ...rest
+}: BAIDeploymentSchedulingHistoryTableProps) => {
+  'use memo';
+  const histories = useFragment(
+    graphql`
+      fragment BAIDeploymentSchedulingHistoryTableFragment on DeploymentHistory
+      @relay(plural: true) {
+        id
+        result
+        subSteps {
+          ...BAISubStepNodesFragment
+        }
+        ...BAIDeploymentSchedulingHistoryNodesFragment
+      }
+    `,
+    schedulingHistoryFrgmt,
+  );
+
+  const dataSource = filterOutNullAndUndefined(histories);
+  const { mode, expandedRowKeys, onExpandedRowsChange, expandColumnTitle } =
+    useSchedulingHistoryExpandable(dataSource, {
+      mode: expandMode,
+      onModeChange: onExpandModeChange,
+    });
+
+  return (
+    <BAIDeploymentSchedulingHistoryNodes
+      schedulingHistoryFrgmt={histories}
+      expandable={{
+        columnTitle: expandColumnTitle,
+        expandedRowKeys,
+        onExpandedRowsChange,
+        rowExpandable: (record: DeploymentSchedulingHistoryNodeInList) =>
+          !_.isEmpty(dataSource.find((h) => h.id === record.id)?.subSteps),
+        expandedRowRender: (record: DeploymentSchedulingHistoryNodeInList) => (
+          <BAISubStepNodes
+            resizable
+            subStepsFrgmt={
+              dataSource.find((h) => h.id === record.id)?.subSteps ?? []
+            }
+            pagination={false}
+            errorsOnly={mode === 'errors-only'}
+          />
+        ),
+      }}
+      {...rest}
+    />
+  );
+};
+
+export default BAIDeploymentSchedulingHistoryTable;
--- a/packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryNodeTable.tsx
+++ b/packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryNodeTable.tsx
@@ -18,8 +18,6 @@ import {
   BAITable,
   BAITableProps,
 } from '../Table';
-import { BAIColumnGroupType } from '../Table/BAITable';
-import BAISubStepNodes from './BAISubStepNodes';
 import dayjs from 'dayjs';
 import * as _ from 'lodash-es';
 import { graphql, useFragment } from 'react-relay';
@@ -68,8 +66,6 @@ const BAIRouteSchedulingHistoryNodeTable = ({
       fragment BAIRouteSchedulingHistoryNodeTableFragment on RouteHistory
       @relay(plural: true) {
         id
-        routeId
-        deploymentId
         category
         phase
         fromStatus
@@ -77,9 +73,6 @@ const BAIRouteSchedulingHistoryNodeTable = ({
         result
         errorCode
         message
-        subSteps {
-          ...BAISubStepNodesFragment
-        }
         attempts
         createdAt
         updatedAt
@@ -89,10 +82,7 @@ const BAIRouteSchedulingHistoryNodeTable = ({
   );
 
   const baseColumns = _.map(
-    filterOutEmpty<
-      | BAIColumnType<RouteSchedulingHistoryNodeInList>
-      | BAIColumnGroupType<RouteSchedulingHistoryNodeInList>
-    >([
+    filterOutEmpty<BAIColumnType<RouteSchedulingHistoryNodeInList>>([
       {
         dataIndex: 'updatedAt',
         title: t('comp:BAIRouteSchedulingHistoryNodes.UpdatedAt'),
@@ -133,22 +123,16 @@ const BAIRouteSchedulingHistoryNodeTable = ({
         sorter: isEnableSorter('category'),
       },
       {
-        title: t('comp:BAIRouteSchedulingHistoryNodes.StatusTransition'),
-        key: 'statusTransition',
-        children: [
-          {
-            key: 'fromStatus',
-            title: t('comp:BAIRouteSchedulingHistoryNodes.From'),
-            dataIndex: 'fromStatus',
-            sorter: isEnableSorter('from_status'),
-          },
-          {
-            key: 'toStatus',
-            title: t('comp:BAIRouteSchedulingHistoryNodes.To'),
-            dataIndex: 'toStatus',
-            sorter: isEnableSorter('to_status'),
-          },
-        ],
+        key: 'fromStatus',
+        title: t('comp:BAIRouteSchedulingHistoryNodes.From'),
+        dataIndex: 'fromStatus',
+        sorter: isEnableSorter('from_status'),
+      },
+      {
+        key: 'toStatus',
+        title: t('comp:BAIRouteSchedulingHistoryNodes.To'),
+        dataIndex: 'toStatus',
+        sorter: isEnableSorter('to_status'),
       },
       {
         dataIndex: 'attempts',
@@ -203,18 +187,6 @@ const BAIRouteSchedulingHistoryNodeTable = ({
         onChangeOrder?.(
           (order as (typeof availableRouteHistorySorterValues)[number]) || null,
         );
-      }}
-      expandable={{
-        rowExpandable: (record) => !_.isEmpty(record.subSteps),
-        expandedRowRender: (record) => {
-          return (
-            <BAISubStepNodes
-              resizable
-              subStepsFrgmt={record.subSteps}
-              pagination={false}
-            />
-          );
-        },
       }}
       {...tableProps}
     />
--- a/packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryTable.tsx
+++ b/packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryTable.tsx
@@ -1,0 +1,85 @@
+import { BAIRouteSchedulingHistoryTableFragment$key } from '../../__generated__/BAIRouteSchedulingHistoryTableFragment.graphql';
+import { filterOutNullAndUndefined } from '../../helper';
+import {
+  type SchedulingHistoryExpandMode,
+  useSchedulingHistoryExpandable,
+} from '../../hooks/useSchedulingHistoryExpandable';
+import BAIRouteSchedulingHistoryNodeTable, {
+  BAIRouteSchedulingHistoryNodesProps,
+  RouteSchedulingHistoryNodeInList,
+} from './BAIRouteSchedulingHistoryNodeTable';
+import BAISubStepNodes from './BAISubStepNodes';
+import * as _ from 'lodash-es';
+import { graphql, useFragment } from 'react-relay';
+
+export interface BAIRouteSchedulingHistoryTableProps extends Omit<
+  BAIRouteSchedulingHistoryNodesProps,
+  'schedulingHistoryFrgmt' | 'expandable'
+> {
+  schedulingHistoryFrgmt: BAIRouteSchedulingHistoryTableFragment$key;
+  expandMode?: SchedulingHistoryExpandMode;
+  onExpandModeChange?: (mode: SchedulingHistoryExpandMode) => void;
+}
+
+/**
+ * Thin wrapper around the pure `BAIRouteSchedulingHistoryNodeTable` fragment
+ * table that adds the optional expand/collapse-all sub-step feature. It owns
+ * its own fragment (selecting the fields the expandable hook needs) and feeds
+ * the resolved rows straight through to the nodes component as the nodes
+ * fragment ref — valid because this fragment spreads the nodes fragment.
+ */
+const BAIRouteSchedulingHistoryTable = ({
+  schedulingHistoryFrgmt,
+  expandMode,
+  onExpandModeChange,
+  ...rest
+}: BAIRouteSchedulingHistoryTableProps) => {
+  'use memo';
+  const histories = useFragment(
+    graphql`
+      fragment BAIRouteSchedulingHistoryTableFragment on RouteHistory
+      @relay(plural: true) {
+        id
+        result
+        subSteps {
+          ...BAISubStepNodesFragment
+        }
+        ...BAIRouteSchedulingHistoryNodeTableFragment
+      }
+    `,
+    schedulingHistoryFrgmt,
+  );
+
+  const dataSource = filterOutNullAndUndefined(histories);
+  const { mode, expandedRowKeys, onExpandedRowsChange, expandColumnTitle } =
+    useSchedulingHistoryExpandable(dataSource, {
+      mode: expandMode,
+      onModeChange: onExpandModeChange,
+    });
+
+  return (
+    <BAIRouteSchedulingHistoryNodeTable
+      schedulingHistoryFrgmt={histories}
+      expandable={{
+        columnTitle: expandColumnTitle,
+        expandedRowKeys,
+        onExpandedRowsChange,
+        rowExpandable: (record: RouteSchedulingHistoryNodeInList) =>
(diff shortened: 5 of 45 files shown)


agatha197:
* **#7650** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7650?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7758](https://github.com/lablup/backend.ai-webui/pull/7758) <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7758?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7794](https://github.com/lablup/backend.ai-webui/pull/7794) <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7794?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7650?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7650).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryNodes.tsx
```diff
@@ -183,6 +189,9 @@ const BAISchedulingHistoryNodes = ({
         );
       }}
       expandable={{
+        columnTitle: expandColumnTitle,
+        expandedRowKeys,
+        onExpandedRowsChange,
         rowExpandable: (record) => !_.isEmpty(record.subSteps),
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit: moved the `{...tableProps}` spread to the top of <BAITable> so the component's controlled `expandable`/`rowKey` props are authoritative and can no longer be silently overridden by a passed-through prop. Added an explanatory comment as well.

packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryNodes.tsx
```diff
@@ -205,6 +210,9 @@ const BAIDeploymentSchedulingHistoryNodes = ({
         );
       }}
       expandable={{
+        columnTitle: expandColumnTitle,
+        expandedRowKeys,
+        onExpandedRowsChange,
         rowExpandable: (record) => !_.isEmpty(record.subSteps),
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit: moved the `{...tableProps}` spread to the top of <BAITable> so the component's controlled `expandable`/`rowKey` props are authoritative and can no longer be silently overridden by a passed-through prop. Added an explanatory comment as well.

packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryNodeTable.tsx
```diff
@@ -205,6 +210,9 @@ const BAIRouteSchedulingHistoryNodeTable = ({
         );
       }}
       expandable={{
+        columnTitle: expandColumnTitle,
+        expandedRowKeys,
+        onExpandedRowsChange,
         rowExpandable: (record) => !_.isEmpty(record.subSteps),
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit: moved the `{...tableProps}` spread to the top of <BAITable> so the component's controlled `expandable`/`rowKey` props are authoritative and can no longer be silently overridden by a passed-through prop. Added an explanatory comment as well.

packages/backend.ai-ui/src/hooks/useSchedulingHistoryExpandable.tsx
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Minimal shape shared by every scheduling-history row type
+ * (session / deployment / route). A row is expandable when it has sub-steps,
+ * and it is expanded by default unless its result is a success.
+ */
+export interface SchedulingHistoryExpandableRow {
+  readonly id: string;
+  readonly result?: string | null;
+  readonly subSteps?: ReadonlyArray<unknown> | null;
+}
```
nowgnuesLee, line 19: Please use exact type.

packages/backend.ai-ui/src/components/fragments/useSchedulingHistoryExpandable.tsx
```diff
@@ -0,0 +1,128 @@
...
+    : t('comp:BAITable.ExpandAll');
+
+  // Reuse Ant Design's native row-expand-icon classes so the master toggle is
+  // visually identical to the per-row +/- expand icons in the nested table.
+  // The +/- glyph is drawn by antd's CSS pseudo-elements on these classes.
+  const expandColumnTitle =
+    expandableRowKeys.length > 0 ? (
+      // Center the toggle in the header cell so it lines up with the
+      // per-row expand icons, which Ant Design centers in their column.
+      <div style={{ display: 'flex', justifyContent: 'center' }}>
```
nowgnuesLee: use BAIFlex.

packages/backend.ai-ui/src/components/BAISchedulingResultCell.tsx
```diff
@@ -0,0 +1,85 @@
...
+                <div key={`${subStep.step ?? 'step'}-${idx}`}>
+                  {subStep.step ? <strong>{subStep.step}: </strong> : null}
+                  {subStep.message
+                    ? newLineToBrElement(subStep.message)
+                    : subStep.errorCode}
+                </div>
+              ))}
+            </BAIFlex>
+          }
+        >
```
nowgnuesLee: Since this message is already displayed in another cell, is the tooltip necessary?

nowgnuesLee (changes requested):
Instead of separating BAISchedulingHistoryNodes and useSchedulingHistoryExpandable, I think we should create BAISchedulingHistoryTable and use BAISchedulingHistoryNodes inside that component. The expand-all feature and the useSchedulingHistoryExpandable hook should also live in that component.

packages/backend.ai-ui/src/hooks/useSchedulingHistoryExpandable.tsx
agatha197: Fixed — typed `result` as `SchedulingResult | '%future added value' | null` (the exact scheduling-result union exported from BAISchedulingResultBadge) instead of bare `string`. Pushed in the latest commit.

packages/backend.ai-ui/src/components/fragments/useSchedulingHistoryExpandable.tsx
agatha197: Done — replaced the raw `<div style={{ display: 'flex', justifyContent: 'center' }}>` with `<BAIFlex justify="center">`. Pushed in the latest commit.

packages/backend.ai-ui/src/components/BAISchedulingResultCell.tsx
agatha197: The tooltip isn't redundant with the Message column. The Message column shows the **row-level** message, whereas this tooltip aggregates the **per-sub-step** reasons (step name + message/errorCode) for the non-success sub-steps — which are otherwise only visible after expanding the row. So it lets the user see *why* a row is non-success without expanding. Happy to drop it if you'd rather force the expand-only path, but I'd lean toward keeping it.

agatha197:
@nowgnuesLee thanks for the review. I've applied the two inline nits (exact `SchedulingResult` type on `SchedulingHistoryExpandableRow.result`, and `BAIFlex` for the expand-all header), pushed in the latest commit.

On the structural suggestion — extracting a `BAISchedulingHistoryTable` that wraps `BAISchedulingHistoryNodes` and owns the expand-all feature + `useSchedulingHistoryExpandable`: I want to flag a wrinkle before doing it. There are actually **three** sibling table components today (`BAISchedulingHistoryNodes` for SessionSchedulingHistory, `BAIDeploymentSchedulingHistoryNodes` for DeploymentHistory, `BAIRouteSchedulingHistoryNodeTable` for routes), each with its own Relay fragment/columns but all sharing the `useSchedulingHistoryExpandable` hook and an identical `<BAITable expandable={...}>` shell. So a single wrapper would need to be generic over the three row types, and there's an API-shape decision (does it take a fragment ref + column config, or compose the existing Nodes component?).

Could you confirm the intended shape? Two options I see:
1. A thin generic `BAISchedulingHistoryTable` that takes `dataSource` + `columns` and internally wires the hook + expandable shell, with the three Nodes components composing it (removes the duplicated expandable block).
2. Keep the per-type Nodes components but move the expand-all wiring into a shared internal table so the hook isn't called at each call site.

I'd rather land the structure you have in mind than guess. Happy to do it as a follow-up commit on this PR once you confirm.

packages/backend.ai-ui/src/components/BAISchedulingResultCell.tsx
agatha197: Good point — changed it to show the sub-step **error code** instead of the message. Since the message is already surfaced elsewhere (the Message column), the tooltip now lists the per-sub-step error codes for non-success rows, which is the non-redundant signal. Pushed in the latest commit.

agatha197: Good point — changed it to show the sub-step error code instead of the message. Since the message is already surfaced in the Message column, the tooltip now lists the per-sub-step error codes for non-success rows. Pushed in the latest commit.

agatha197:
@nowgnuesLee done — restructured as you suggested. The expand/collapse-all feature + `useSchedulingHistoryExpandable` hook are now lifted out of the base nodes components into dedicated wrapper components, so the nodes components stay pure fragment tables.

- Added `BAISchedulingHistoryTable` / `BAIDeploymentSchedulingHistoryTable` / `BAIRouteSchedulingHistoryTable`. Each owns its own fragment that spreads the corresponding `*NodesFragment` and selects the fields the hook needs (`id`/`result`/`subSteps`), runs the hook, and passes the resolved rows straight through to the nodes component as the fragment ref while supplying the `expandable` config.
- `BAISchedulingHistoryNodes` / `BAIDeploymentSchedulingHistoryNodes` / `BAIRouteSchedulingHistoryNodeTable` no longer call the hook or hard-code `expandable`; they just spread `{...tableProps}`, so the feature is opt-in via the wrapper and the base components aren't polluted when expansion isn't needed.
- The 3 scheduling-history modals now use the wrappers (their queries spread the `*TableFragment`).

`scripts/verify.sh` → `=== ALL PASS ===`. Re-requesting review.

packages/backend.ai-ui/src/components/BAISchedulingResultCell.tsx
nowgnuesLee, line 1: Please remove this component.

packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryNodes.tsx
```diff
@@ -192,31 +199,26 @@ const BAIDeploymentSchedulingHistoryNodes = ({
...
+      // `dataSource`, `columns`, …) below stay authoritative. Optional features
+      // such as `expandable` are supplied by callers and flow through here.
+      {...tableProps}
       rowKey={'id'}
-      dataSource={filterOutNullAndUndefined(histories)}
+      dataSource={dataSource}
       columns={allColumns}
       scroll={{ x: 'max-content' }}
+      neoHeader={false}
+      bordered
```
nowgnuesLee: Please configure the corresponding value in `BAISchedulingHistoryTable`.

packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryNodes.tsx
```diff
@@ -78,6 +76,10 @@ const BAIDeploymentSchedulingHistoryNodes = ({
           message
           subSteps {
             ...BAISubStepNodesFragment
```
nowgnuesLee: we can remove this line.

packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+  'use memo';
+  const histories = useFragment(
+    graphql`
+      fragment BAIDeploymentSchedulingHistoryTableFragment on DeploymentHistory
+      @relay(plural: true) {
+        id
+        result
+        subSteps {
+          __typename
+        }
```
nowgnuesLee, line 46: Please include the BAISubStepNodes fragment here.

packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryTable.tsx
```diff
@@ -0,0 +1,83 @@
...
+        onExpandedRowsChange,
+        rowExpandable: (record: DeploymentSchedulingHistoryNodeInList) =>
+          !_.isEmpty(record.subSteps),
+        expandedRowRender: (record: DeploymentSchedulingHistoryNodeInList) => (
+          <BAISubStepNodes
+            resizable
+            subStepsFrgmt={record.subSteps}
+            pagination={false}
+            errorsOnly={mode === 'errors-only'}
+          />
```
nowgnuesLee, line 77: please add `neoHeader = {false}`

packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryNodeTable.tsx
```diff
@@ -79,6 +75,10 @@ const BAIRouteSchedulingHistoryNodeTable = ({
         message
         subSteps {
           ...BAISubStepNodesFragment
```
nowgnuesLee: remove this line too.

nowgnuesLee (changes requested):

packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+  'use memo';
+  const histories = useFragment(
+    graphql`
+      fragment BAIDeploymentSchedulingHistoryTableFragment on DeploymentHistory
+      @relay(plural: true) {
+        id
+        result
+        subSteps {
+          __typename
+        }
```
nowgnuesLee, line 46: please spread BAISubStepNodes here.

packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryNodeTable.tsx
```diff
@@ -68,8 +67,6 @@ const BAIRouteSchedulingHistoryNodeTable = ({
       fragment BAIRouteSchedulingHistoryNodeTableFragment on RouteHistory
       @relay(plural: true) {
         id
-        routeId
-        deploymentId
         category
         phase
         fromStatus
```
nowgnuesLee, line 71: <p class='graphite__hidden'><i>[Re: line +78]</i></p>
don't spread it here.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7650?utm_source=unchanged-line-comment">Graphite</a>.</p>

packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+  'use memo';
+  const histories = useFragment(
+    graphql`
+      fragment BAIRouteSchedulingHistoryTableFragment on RouteHistory
+      @relay(plural: true) {
+        id
+        result
+        subSteps {
+          __typename
+        }
```
nowgnuesLee, line 46: here too.

nowgnuesLee (changes requested):
please clean up useless comments.

packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryTable.tsx
agatha197: Done. Moved `...BAISubStepNodesFragment` from the inner `BAIDeploymentSchedulingHistoryNodesFragment` to the wrapper `BAIDeploymentSchedulingHistoryTableFragment`, and updated the expandable callbacks to look up `subSteps` from the wrapper's own `dataSource` by `id`.

packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryNodeTable.tsx
agatha197: Removed the `subSteps { ...BAISubStepNodesFragment }` from `BAIRouteSchedulingHistoryNodeTableFragment`. The wrapper now owns the spread.

packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryTable.tsx
agatha197: Done. Added `...BAISubStepNodesFragment` to `BAIRouteSchedulingHistoryTableFragment` and updated callbacks to look up `subSteps` from the wrapper's `dataSource`.

packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+}: BAISchedulingHistoryTableProps) => {
+  'use memo';
+  const histories = useFragment(
+    graphql`
+      fragment BAISchedulingHistoryTableFragment on SessionSchedulingHistory
+      @relay(plural: true) {
+        id
+        result
+        subSteps {
+          __typename
```
nowgnuesLee: Please spread baisubstepnodes here.

nowgnuesLee (changes requested):
Please clean up the useless comments.

packages/backend.ai-ui/src/hooks/useSchedulingHistoryExpandable.tsx
```diff
@@ -0,0 +1,202 @@
...
+          // item is shown as "active". The three modes read as actions you
+          // trigger, not a stateful toggle.
+          menu={{
+            items: menuItems,
+            onClick: onMenuClick,
+          }}
+        >
+          <button
+            type="button"
+            aria-label={t('comp:BAITable.ExpandOptions')}
```
ironAiken2: This `aria-label` is screen-reader-only, and I'm not sure screen-reader accessibility is a deliberate target for this PR. Let's drop the `aria-label` and instead surface the same text through an antd `Tooltip` on the kebab trigger — that makes the affordance discoverable for sighted users on hover, and keeps `comp:BAITable.ExpandOptions` in use (so the key and its translations don't become dead).

Note: `comp:BAITable.ExpandOptions` is currently referenced only here (this single `aria-label`), so if it were removed outright the key + its 21 locale entries would be orphaned. Moving it into a Tooltip resolves that.

Separately, if the goal was ever to give the test (Playwright) a stable hook, an i18n-bound value isn't the right tool — a locator tied to translated text breaks across locales. Prefer a locale-independent `data-testid` for that.

packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryNodes.tsx
```diff
@@ -171,29 +169,25 @@ const BAISchedulingHistoryNodes = ({
   const allColumns = customizeColumns
     ? customizeColumns(baseColumns)
     : baseColumns;
+
+  const dataSource = filterOutNullAndUndefined(histories);
```
ironAiken2: `dataSource` is used exactly once (the `dataSource={dataSource}` just below), so the local doesn't earn the indirection. Inlining `filterOutNullAndUndefined(histories)` straight into the prop keeps the data source visible at its point of use, without jumping back up:

```tsx
<BAITable
  rowKey={'id'}
  dataSource={filterOutNullAndUndefined(histories)}
  columns={allColumns}
  ...
```

Same single-use pattern in `BAIDeploymentSchedulingHistoryNodes.tsx` (L191) and `BAIRouteSchedulingHistoryNodeTable.tsx` (L190) — worth inlining those too.

Leaving two cases as-is on purpose: the `*SchedulingHistoryTable.tsx` wrappers use `dataSource` more than once (deployment/route reference it 3x via `useSchedulingHistoryExpandable(...)` + `dataSource.find(...)`), and `BAISubStepNodes.tsx` builds it from a 3-step chain (`...filter(...).reverse()`) where a named local still helps readability.

packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const dataSource = filterOutNullAndUndefined(histories);
+  const { mode, expandedRowKeys, onExpandedRowsChange, expandColumnTitle } =
+    useSchedulingHistoryExpandable(dataSource, {
+      mode: expandMode,
+      onModeChange: onExpandModeChange,
+    });
+
+  return (
+    <BAISchedulingHistoryNodes
```
ironAiken2, line 61: Looking at the current structure, each `*Table` contains a `*Nodes` component, but I don't think this separation is necessary — the `*Table` components just thinly wrap their `*Nodes` component.

ironAiken2 (dismissed):

packages/backend.ai-ui/src/components/Table/BAITable.tsx
```diff
@@ -459,7 +462,8 @@ const BAITable = <RecordType extends AnyObject = AnyObject>({
           showSorterTooltip={false}
           className={classNames(
             resizable && styles.resizableTable,
-            styles.neoHeader,
+            styles.expandedRowBackground,
+            neoHeader ? styles.neoHeaderTitle : styles.defaultHeaderTitle,
```
yomybaby: This line's purpose is to use Antd's default styles, so why is a separate CSS needed? Isn't it just enough to not apply the neoHeaderTitle style?

yomybaby (changes requested):

packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryTable.tsx
agatha197: Done — the session wrapper fragment (`BAISchedulingHistoryTableFragment`) now spreads `...BAISubStepNodesFragment` under `subSteps`, and the base `BAISchedulingHistoryNodesFragment` no longer selects `subSteps`. Sub-steps are read via `dataSource.find(...)` in the wrapper, matching the deployment/route wrappers. (commit 93f16773c)

packages/backend.ai-ui/src/hooks/useSchedulingHistoryExpandable.tsx
agatha197: Done — dropped the `aria-label` and wrapped the kebab trigger in an antd `Tooltip` titled `comp:BAITable.ExpandOptions`, so the affordance is discoverable on hover and the key (plus its locale entries) stays in use. Screen-reader support remains out of scope for this feature, as you noted. (commit 93f16773c)

packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryNodes.tsx
agatha197: Done — inlined `filterOutNullAndUndefined(histories)` straight into the `dataSource` prop in all three base `*Nodes` tables (session/deployment/route) and removed the single-use local. Left the `*Table` wrappers and `BAISubStepNodes` as-is, per your note about their multi-use `dataSource`. (commit 93f16773c)

packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryTable.tsx
agatha197: I'd like to keep the `*Table` / `*Nodes` split here. It was introduced at @nowgnuesLee's explicit request in an earlier review: the `*Table` wrapper owns the expand-all feature and the `useSchedulingHistoryExpandable` hook (plus its own fragment), while the base `*Nodes` stays a pure fragment table that can be reused without the expand behavior. Collapsing them would revert that structure, so I'd prefer not to change it in this PR. Happy to revisit if you and @nowgnuesLee agree on merging them.

packages/backend.ai-ui/src/components/Table/BAITable.tsx
agatha197: Good catch — you're right. I re-checked the theme: `resources/theme.json` defines `Table.headerBg` only for light (there is no `Table` entry under `dark`), but Ant Design's dark algorithm already supplies a default header background, so the extra class wasn't actually needed. I removed `defaultHeaderTitle` entirely; `neoHeader={false}` now applies no header-title style and lets antd render its default header (the themed `#E3E3E3` in light, antd's default header fill in dark). (commit 93f16773c)

nowgnuesLee (approved):
LGTM

yomybaby (approved):
LGTM

PR #7673, opened 2026-06-01
refactor(FR-3019): migrate Purge/Update user modals to UserV2 fragment

### LINKED ISSUE #7669: Migrate Purge/Update user modals to receive a v2 UserV2 fragment
## Background

In FR-3016, the super-admin user list (`AdminUserManagement`) was migrated from the v1 `user_nodes` query to the v2 `adminUsersV2` query. To keep that PR scoped to the list, the modals it opens were intentionally **kept on their v1 implementation**.

`PurgeUsersModal` and `UpdateUsersModal` previously received a Relay fragment on the v1 `UserNode`. They now receive a plain `{ id, email }[]` `users` prop instead (they only ever used id/email), so the v2 list can drive them without a v1 fragment.

## Task

Now that the list is on v2, these modals should be updated to receive data via a **v2 `UserV2` fragment** instead of the plain `{ id, email }[]` prop, consistent with the rest of the v2 migration.

- `react/src/components/PurgeUsersModal.tsx`
- `react/src/components/UpdateUsersModal.tsx`

## Acceptance Criteria

- Both modals consume a `UserV2` Relay fragment (passed from `AdminUserManagement`) rather than a plain `{ id, email }[]` prop.
- No regression in the purge / update flows.

JIRA Issue: FR-3019

ironAiken2:
Resolves #7669 (FR-3019)

Builds on FR-3016 (#7666, merged), which moved the super-admin user list to `adminUsersV2`.

## Summary

This PR migrates **all modals used by `AdminUserManagement`** — together with the shared TOTP / profile / forced-2FA flows — onto the v2 GraphQL surface, and **removes the legacy v1 fallbacks** now that the WebUI targets manager **>= 26.4.3** only. Modals receive their data via `UserV2` Relay fragments and drive the `admin*V2` mutations, replacing the `{ id, email }[]` props / v1 `user(email:)` queries / `modify_user`-style mutations. User creation (single and bulk) also moves to v2 now that the v2 create payload exposes the generated keypair.

## Changes

**Fragment-driven data (UserV2)**
- `PurgeUsersModal` / `UpdateUsersModal`: plural `UserV2` fragments (`@relay(plural: true)`) read with `useFragment`, replacing the plain `{ id, email }[]` prop.
- `UserInfoModal`: replaced its own v1 `user(email:)` query with a `UserV2` fragment. `AdminUserManagement` spreads `...UserInfoModalFragment` on the list query's edges and looks up the selected node in the already-loaded `adminUsersV2.edges` — no second query, no `limit: 1` filter.
- `UserSettingModal` (edit): same fragment-on-list-query pattern via `...UserSettingModalFragment` on `adminUsersV2.edges`; create / bulk-create render the modal with no fragment.

**Shared TOTP modal on a single `UserV2` fragment**
- `TOTPActivateModalFragment` lives `on UserV2` (`basicInfo.email`, `security.totpActivated`) and is the single source of data for the modal across **all** consumers: admin (`UserSettingModal`, via `adminUsersV2`) and self / profile / login (`UserProfileSettingModal`, `ForceTOTPChecker`).
- The self / profile / login flows now read the current user via the **`myUserV2`** root field — it returns a `UserV2` and is **not** superadmin-gated — instead of the v1 `user(email:)` query. `UserDropdownMenu` and `ForceTOTPChecker` query `myUserV2`. Because both `adminUsersV2` and `myUserV2` yield `UserV2` nodes, one colocated `on UserV2` fragment serves every caller (no scalar prop bridge, no broken colocation).

**Removed all `update-user-v2` v1 fallbacks** (manager >= 26.4.3 only)
- Dropped every `supports('update-user-v2')` branch and its legacy `modify_user` path, so the v2 mutations always run:
  - `UpdateUsersModal` (bulk): `adminBulkUpdateUsersV2`.
  - `UserSettingModal` (edit): `adminUpdateUserV2`.
  - `AdminUserManagement` (activate/deactivate toggle): `adminUpdateUserV2`.
  - `UserProfileSettingModal` (full-name / password): `updateUserV2`.
- Purge (`adminBulkPurgeUsersV2`) was already v2; the legacy `purge_user` loop and `supports('bulk-purge-users')` gate are now removed too.

**User creation on v2 (keypair reveal)**
- Single create now runs `adminCreateUserV2`: its `CreateUserV2Payload` returns the user together with the generated `keypair` (one-time `secretKey`).
- Bulk create runs `adminBulkCreateUsersWithKeypairV2`, which replaces the now-**deprecated** `adminBulkCreateUsersV2` and returns each created user's keypair — so bulk create now reveals the generated keypairs too (previously it did not). Keypairs are surfaced on both full-success and partial-failure so one-time secret keys are never silently lost.
- `GeneratedKeypairListModal` reads a `CreateKeypairPayload` fragment (`secretKey`, `keypair.accessKey`, `keypair.user.basicInfo.email`) instead of the v1 `KeyPair` fragment. It is used by `UserSettingModal` and `BulkCreateUserFromCSVModal` (FR-1818, merged on main during review); the CSV flow was migrated to `adminBulkCreateUsersWithKeypairV2` accordingly, dropping its per-email `keypairs(email:)` re-fetch.

## Notable decisions

- **User creation now runs on v2 (revised).** Create was initially left on v1 `create_user` because `adminCreateUserV2`'s payload returned only `user` (no keypair), which would have dropped the "reveal generated keypair (secret key, shown once)" flow. The schema has since added `keypair` (with the one-time secret key) to `CreateUserV2Payload`, and `adminBulkCreateUsersWithKeypairV2` superseded the deprecated `adminBulkCreateUsersV2`. Both single and bulk create therefore now run on v2 while preserving — and, for bulk, newly adding — the keypair-reveal flow.
- **TOTP activation is self-service** (`baiClient.initialize_totp` / `activate_totp` act on the session user), so a single `UserV2` fragment fed from either `adminUsersV2` (admin) or `myUserV2` (self) is sufficient — the modal keeps its colocated fragment rather than receiving plain scalars.
- Form values are mapped to the camelCase `UpdateUserV2Input` / `CreateUserV2Input` with `status` / `role` enum conversion, and project node ids are converted to UUIDs (`toLocalId`) for `group_ids`.

**Review feedback (post-rebase)**
- `UserInfoModal` / `UpdateUsersModal` bodies are split into same-file `*Contents` components that `throw` when user data is unavailable; the modal shell catches via `ErrorBoundary` and renders the error alert inside the still-open modal (same pattern as `ResourcePresetSettingModal`).
- `UserProfileSettingModalFragment` is inlined into `UserProfileSettingModal` (separate fragment file removed).

**Review feedback (agatha197 pass)**
- `BulkCreateUserFromCSVModal`: keypairs are now surfaced on partial-failure as well (one-time secret keys must never be silently dropped).
- `PurgeUsersModal`: removed the v1 `purge_user` fallback and `supports('bulk-purge-users')` gate; `adminBulkPurgeUsersV2` is unconditional.
- `UpdateUsersModal`: fixed `_.isEmpty` predicate that was erroneously stripping numeric fields (`containerUid`, `containerMainGid`).
- `UserSettingModal`, `TOTPActivateModal`, `TOTPActivateForm`, `GeneratedKeypairListModal`: added missing `'use memo'` directives.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup).


ironAiken2:
* **#7673** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7673?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7666** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7666?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7673).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

yomybaby (dismissed):
Please resolve merg conflicts

react/src/components/UserInfoModal.tsx
```diff
@@ -2,66 +2,67 @@
...
-          main_access_key @since(version: "23.09.7")
         }
       }
     `,
-    {
-      email: userEmail,
-      isNotSupportSudoSessionEnabled: false,
-      isTOTPSupported: isTOTPSupported ?? false,
-    },
+    userInfoFrgmt ?? null,
```
nowgnuesLee, line 74:
```suggestion
    userInfoFrgmt,
```

react/src/components/UserInfoModal.tsx
```diff
@@ -81,74 +82,96 @@ const UserInfoModal: React.FC<Props> = ({
...
-          </BAIFlex>
-        </Descriptions.Item>
-      </Descriptions>
+      {!user ? (
+        <BAIAlert
+          type="error"
+          showIcon
+          title={t('credential.FailedToLoadUserInfo')}
+        />
+      ) : (
```
nowgnuesLee: Rather than handling it this way, using an Error Boundary and throwing the error seems more appropriate.

react/src/components/UpdateUsersModal.tsx
```diff
@@ -78,301 +96,315 @@ const UpdateUsersModal = ({ users, ...modalProps }: UpdateUsersModalProps) => {
...
-            fallback={
-              <Form.Item label={t('credential.Projects')}>
-                <BAISelect loading />
+        {users.length === 0 ? (
+          <BAIAlert
+            type="error"
+            showIcon
+            title={t('credential.FailedToLoadUserInfo')}
+          />
+        ) : (
```
nowgnuesLee: Rather than handling it this way, using an Error Boundary and throwing the error seems more appropriate.

react/src/components/UserProfileSettingModalFragment.graphql.ts
nowgnuesLee, line 1: Why is this separated into a separate file? It doesn’t seem to align with our pattern.

nowgnuesLee (changes requested):

ironAiken2:
@yomybaby Rebased this branch onto the latest `main` via `gt sync` — the branch is now up to date with `main` (0 commits behind) and merges cleanly, so the merge conflict is resolved. PTAL when you have a chance.

react/src/components/UserInfoModal.tsx
ironAiken2: Applied — the fragment ref is now passed directly without the nullish coalescing.

react/src/components/UserInfoModal.tsx
ironAiken2: Refactored as suggested: the modal body is now split into a `UserInfoModalContents` component (same file) that throws when user data is unavailable, and the modal shell catches it with an `ErrorBoundary` whose fallback renders the error alert inside the still-open modal — same approach as `ResourcePresetSettingModal`.

react/src/components/UpdateUsersModal.tsx
ironAiken2: Same treatment here: extracted `UpdateUsersModalContents` (same file) which throws when the user list is empty; the `ErrorBoundary` in the modal shell renders the error alert as the modal body. The mutation/onOk logic stays in the parent, sharing the plural fragment via a module-level `graphql` const.

react/src/components/UserProfileSettingModalFragment.graphql.ts
ironAiken2: You're right — inlined the fragment into `UserProfileSettingModal` and deleted the separate file. It had been split out for no structural reason; nothing else imported it.

react/src/components/GeneratedKeypairListModal.tsx
```diff
@@ -101,35 +105,39 @@ const GeneratedKeypairListModal: React.FC<GeneratedKeypairListModalProps> = ({
...
               title: t('general.E-Mail'),
-              dataIndex: ['user_info', 'email'],
+              dataIndex: ['keypair', 'user', 'basicInfo', 'email'],
               key: 'email',
               sorter: (a, b) =>
-                localeCompare(a?.user_info?.email, b?.user_info?.email),
+                localeCompare(
+                  a?.keypair?.user?.basicInfo?.email,
+                  b?.keypair?.user?.basicInfo?.email,
+                ),
```
agatha197, line 119: How about using function to check sortable like other components (e.g., `isEnableSorter()`)?

agatha197 (changes requested):

react/src/components/BulkCreateUserFromCSVModal.tsx
```diff
@@ -649,9 +622,10 @@ const BulkCreateUserFromCSVModal: React.FC<BulkCreateUserFromCSVModalProps> = ({
             resolve();
             return;
           }
-          const createdCount =
-            res.adminBulkCreateUsersV2?.createdUsers?.length ?? 0;
-          const failed = res.adminBulkCreateUsersV2?.failed ?? [];
+          const createdList =
+            res.adminBulkCreateUsersWithKeypairV2?.created ?? [];
+          const createdCount = createdList.length;
+          const failed = res.adminBulkCreateUsersWithKeypairV2?.failed ?? [];
```
agatha197, line 628: **[HIGH] One-time secret keys silently lost on partial-failure bulk create** (re: lines 629–651)

In `onCompleted`, the keypair-surfacing (`setCreatedKeypairs`) lives **only** in the `else` (full-success) branch. When `failed.length > 0`, the code sets `failedRows`/`createdCount` and a warning, then returns — it never reveals the keypairs of the users that **did** succeed. Since `adminBulkCreateUsersWithKeypairV2` returns `secretKey` only at creation time, those secret keys are permanently lost (no recovery short of delete-and-recreate), and `GeneratedKeypairListModal` (gated on `createdKeypairs`) never opens.

This diverges from the sibling `UserSettingModal.tsx` (~lines 349–352), which runs the keypair-reveal **unconditionally after** the if/else, so its partial-failure batches still reveal the secrets.

**Fix:** hoist the keypair-surfacing out of the `else` so it runs for both outcomes, and render the keypair-download screen alongside (or before) the failure table.
```ts
const keypairs = _.map(createdList, (c) => c.keypair);
if (keypairs.length > 0) setCreatedKeypairs(keypairs);
else onRequestClose(true);
```

react/src/components/UserDropdownMenu.tsx
```diff
@@ -73,40 +73,33 @@ const UserDropdownMenu: React.FC<{
...
-          full_name
-          ...UserProfileSettingModalFragment
-        }
-        myClientIp @skipOnClient(if: $isNotSupportUpdateUserV2) {
-          clientIp
+  const { myUserV2: user, myClientIp } =
+    useLazyLoadQuery<UserDropdownMenuQuery>(
+      graphql`
+        query UserDropdownMenuQuery {
+          myUserV2 {
```
agatha197, line 80: **[HIGH] Dropping the `@skipOnClient` gate on `myClientIp` can blank the whole header on managers < 26.4.2**

On `main` this queried the v1 `user(email:)` field (available on every manager) and gated the newer field with `myClientIp @skipOnClient(if: $isNotSupportUpdateUserV2)` + `currentClientIp = baiClient.supports('update-user-v2') ? myClientIp?.clientIp : undefined`. This PR replaces it with a bare `myUserV2 { … }` + unconditional `myClientIp { clientIp }`.

Per `data/schema.graphql`, `myUserV2` is *"Added in 26.2.0"* and `myClientIp` is *"Added in 26.4.2"*. `UserDropdownMenu` renders in the header on every page (`WebUIHeader.tsx`) wrapped only in `BAIErrorBoundary` (visible error fallback, not the null fallback). So against a manager < 26.4.2 the query errors and the entire header (user menu, logout, notifications, theme toggle) is replaced by an error UI. The "manager >= 26.4.3 only" assumption is not enforced anywhere in the repo.

**Fix (pick one):** (a) re-gate `myClientIp`/`myUserV2` behind an explicit `isManagerVersionCompatibleWith('26.4.2')` / `@skipOnClient`; (b) wrap this query in `ErrorBoundaryWithNullFallback` (or render a minimal static menu) so an unavailable field degrades gracefully; or (c) add an explicit min-manager-version login gate so the assumption is enforced.

react/src/components/PurgeUsersModal.tsx
```diff
@@ -42,11 +36,25 @@ interface PurgeUsersFormValues {
...
+          id
+          basicInfo {
+            email
+          }
+        }
+      `,
+      usersFrgmt,
+    ),
+  );
+
```
agatha197, line 52: **[MEDIUM] Dead legacy v1 `purge_user` fallback the PR claims to remove; version comments are wrong** (re: `supportsBulkPurge` gate ~line 66, v1 mutation lines 82–94, `else` loop lines 154–211)

The PR's stated goal is to remove all v1 fallbacks ("manager >= 26.4.3 only"). But this modal still ships the v1 `purge_user` mutation and its entire `else` branch, guarded by `supports('bulk-purge-users')`. That feature is enabled at **26.4.0** (`packages/backend.ai-client/src/client.ts:894`), so on every supported manager the `else` branch is unreachable dead code — exactly the kind of fallback this refactor removed from the sibling modals.

The inline comments are also wrong: lines 68/82/155 label the boundary as `26.3.0`, but the feature gates at `26.4.0`.

**Fix:** delete the v1 `purge_user` mutation, the `supportsBulkPurge` check, and the `else` branch (154–211), keeping only `adminBulkPurgeUsersV2`; remove the now-unused `useMutationWithPromise`/`getErrorMessage` imports; drop/correct the `26.3.0` comments. (The destructive-confirmation UX via `BAIDeleteConfirmModal` is correct — no change needed there.)

react/src/components/UpdateUsersModal.tsx
```diff
@@ -78,303 +95,332 @@ const UpdateUsersModal = ({ users, ...modalProps }: UpdateUsersModalProps) => {
...
           ?.validateFields()
           .then((values) => {
             setIsPending(true);
-            const userList = _.compact(users);
 
-            // Build props with type safety and remove empty values
-            const props = _.omitBy(
+            // v2: same field-inclusion rules as the legacy path, but with
+            // camelCase keys and the UserStatusV2 enum.
+            const input = _.omitBy(
```
agatha197, line 113: **[LOW] `_.isEmpty` strips numeric fields in the bulk-update `omitBy` predicate**

`(value) => _.isNil(value) || value === '' || _.isEmpty(value)` treats any number as empty (`_.isEmpty(5) === true`, `_.isEmpty(0) === true`), so `containerUid` / `containerMainGid` (bound to `<InputNumber>`, typed `number`) are always stripped from the v2 input and can never be set via the bulk Update Users modal. `UpdateUserV2Input.containerUid: Int` is a real, settable field.

This is carried over verbatim from the legacy v1 path (pre-existing, not a regression), but the refactor touches this exact predicate.

**Fix:** `(value) => _.isNil(value) || value === '' || ((_.isArray(value) || _.isPlainObject(value)) && _.isEmpty(value))`. Confirm with the author whether fixing it is in-scope for FR-3019.

react/src/components/BulkCreateUserFromCSVModal.tsx
```diff
@@ -222,9 +220,11 @@ const BulkCreateUserFromCSVModal: React.FC<BulkCreateUserFromCSVModalProps> = ({
       mutation BulkCreateUserFromCSVModalMutation(
         $input: BulkCreateUserV2Input!
       ) {
-        adminBulkCreateUsersV2(input: $input) {
-          createdUsers {
-            id
+        adminBulkCreateUsersWithKeypairV2(input: $input) {
```
agatha197, line 223: **[LOW] Stale comment** (re: block comment at lines 209–214)

The comment still says the bulk payload "does not expose secret keys … so they are fetched separately per created email after the mutation completes." That's no longer true: `adminBulkCreateUsersWithKeypairV2` returns `created { keypair { … } }` in the payload and the per-email `keypairs(email:)` re-fetch query was deleted. The nearby comment (~lines 643–645) already states this correctly.

**Fix:** update lines 209–214 to reflect that keypairs come directly from the mutation payload; no follow-up fetch.

react/src/components/UserSettingModal.tsx
```diff
@@ -483,7 +479,7 @@ const UserSettingModal: React.FC<UserSettingModalProps> = ({
       destroyOnHidden
       onOk={handleOk}
       confirmLoading={
-        isInFlightCommitModifyUserSetting ||
+        isInFlightUpdateUserV2 ||
```
agatha197, line 549: **[LOW] Edit modal silently falls back to a create form when the target fragment is null** (re: title/branch logic ~line 469)

Unlike `UserInfoModal`/`UpdateUsersModal` (which `throw` into an inner `ErrorBoundary` when the user fragment is unavailable), `UserSettingModal` renders a **create** form when opened for edit but the looked-up node is null (`AdminUserManagement.tsx` passes `… ?? null`). In practice the ref is always non-null (the action originates from a visible row), so this is an edge case — a transient store eviction would present an empty create form. No wrong mutation fires (create/edit branch on the same `user` value).

**Fix (non-blocking):** for consistency with the ErrorBoundary pattern adopted elsewhere in this PR, wrap the edit content in an `ErrorBoundary` and `throw` when a non-null id was requested but `useFragment` returns null — or branch on the requested `userId` rather than the resolved fragment.

react/src/components/UserSettingModal.tsx
```diff
@@ -132,13 +149,13 @@ const formatBulkUsername = (
...
 interface UserSettingModalProps extends BAIModalProps {
-  userEmail?: string | null;
+  userSettingFrgmt?: UserSettingModalFragment$key | null;
   bulkCreate?: boolean;
   onRequestClose: (success: boolean) => void;
 }
 
 const UserSettingModal: React.FC<UserSettingModalProps> = ({
-  userEmail = null,
+  userSettingFrgmt = null,
```
agatha197, line 158: **[LOW] Missing `'use memo'` directive** (opt-in gap, not a removal)

Per `CLAUDE.md` / `react.instructions.md`, component bodies should declare `'use memo'` for React Compiler. These touched components omit it: `UserSettingModal`, `TOTPActivateModal` (+ exported `TOTPActivateForm`), and `GeneratedKeypairListModal`. Verified against merge-base that none had it before, so the "never remove existing `'use memo'`" rule is **not** violated — the other modals in this PR all carry it, so this is just an inconsistency.

**Fix (optional):** add `'use memo';` as the first statement in each of those component bodies.

react/src/components/AdminUserManagement.tsx
```diff
@@ -594,33 +580,47 @@ const AdminUserManagement: React.FC<AdminUserManagementProps> = () => {
...
-        }
-        bulkCreate={openBulkCreateModal}
-        loading={isPendingSettingModalOpen}
-        onRequestClose={(success) => {
-          setEmailForSettingModal(null);
-          setOpenCreateModal(false);
-          setOpenBulkCreateModal(false);
-          if (success) {
-            updateFetchKey();
+      {userIdForInfoModal && (
```
agatha197: **[LOW] PR description vs. implementation mismatch** (no code change needed)

The description says detail/edit modals "fetch the selected user on-demand via `adminUsersV2` filtered by uuid, `limit: 1`." The actual (and safer) implementation spreads `...UserInfoModalFragment` / `...UserSettingModalFragment` on the **list** query and looks the node up in the already-loaded `adminUsersV2.edges` — there is no second `useLazyLoadQuery` and no `limit: 1`. This nicely sidesteps the `first`+`offset` pagination footgun, but the description is inaccurate.

**Fix:** correct the PR description to match the fragment-on-list-query approach.

agatha197 (dismissed):
## Workflow review — FR-3019 UserV2 fragment migration

The v2 fragment migration is **well-structured in its core design**: detail/edit modals are fed from fragments spread on the list query (sidestepping the `first`+`offset` pagination footgun), `PurgeUsersModal` correctly uses the type-to-confirm `BAIDeleteConfirmModal`, and the create paths surface the one-time secret keys from the mutation payload.

There are **two blockers** to resolve before merge:

1. **[HIGH]** `BulkCreateUserFromCSVModal` silently discards the one-time secret keys of successfully-created users on a partial-failure batch (unrecoverable). Diverges from `UserSettingModal`, which reveals them.
2. **[HIGH]** `UserDropdownMenu` removes the `@skipOnClient` version gate on `myClientIp` and can replace the entire header with an error UI on managers < 26.4.2 — the "manager >= 26.4.3 only" assumption is not enforced anywhere.

One **[MEDIUM]**: `PurgeUsersModal` still ships a dead v1 `purge_user` fallback (with wrong `26.3.0` version comments) that this "remove all v1 fallbacks" refactor should have deleted.

The remaining items are low-severity nits / a PR-description fix — see inline comments.

_Generated by a multi-agent workflow review (5 dimensions → adversarial verification → synthesis); the two HIGH findings were additionally hand-verified against the checked-out branch._

react/src/components/BulkCreateUserFromCSVModal.tsx
ironAiken2: Fixed. Moved `setCreatedKeypairs` out of the `else` branch so it runs for both full-success and partial-failure cases. The `onRequestClose(true)` fallback (for when `createdList` is empty) is now in both branches. Stale comment on the per-email re-fetch also updated in the same commit.

react/src/components/UserDropdownMenu.tsx
ironAiken2: The removal is intentional. This PR is part of the manager >= 26.4.3-only series (parent PR #7666 already dropped all version-gated v1 fallbacks). `myClientIp` is gated behind `update-user-v2` (26.4.2), so it's guaranteed available on every manager this WebUI supports. The `@skipOnClient(if: $isNotSupportUpdateUserV2)` guard was only needed when we still supported < 26.4.2; dropping it is deliberate cleanup.

react/src/components/PurgeUsersModal.tsx
ironAiken2: Fixed. Removed the `supportsBulkPurge` check, the v1 `PurgeUsersModalMutation` (`purge_user`), and the entire `else` branch. The `commitBulkPurge` path is now unconditional since `adminBulkPurgeUsersV2` has been available since 26.3.0 and the PR targets >= 26.4.3. Deleted the corresponding `__generated__/PurgeUsersModalMutation.graphql.ts` and removed unused imports (`useSuspendedBackendaiClient`, `useErrorMessageResolver`, `useMutationWithPromise`, `PayloadError`).

react/src/components/UpdateUsersModal.tsx
ironAiken2: Fixed. Changed the predicate to `_.isNil(value) || value === '' || (!_.isNumber(value) && _.isEmpty(value))` so numeric fields (`containerUid`, `containerMainGid`) are preserved in the mutation input when set by the user.

react/src/components/BulkCreateUserFromCSVModal.tsx
ironAiken2: Fixed in the same commit as the keypair-loss fix. The block comment now reads: "adminBulkCreateUsersWithKeypairV2 returns each keypair's one-time secret key directly in the payload; no follow-up fetch is needed."

react/src/components/UserSettingModal.tsx
ironAiken2: Acknowledged. The `user === null` → create-form fallback is intentional here because `UserSettingModal` handles both create and edit in one component (unlike `UserInfoModal`/`UpdateUsersModal` which are view/bulk-edit only). In the edit path the ref is always non-null (the action originates from a visible list row), so this case cannot be triggered by normal UI flow. Adding an ErrorBoundary throw in this component would require splitting it into two components (create vs edit), which is a larger refactor outside this PR's scope. Leaving as-is with the comment noting the invariant.

react/src/components/UserSettingModal.tsx
ironAiken2: Fixed. Added `'use memo'` to `UserSettingModal`, `TOTPActivateModal`, `TOTPActivateForm`, and `GeneratedKeypairListModal`. None had it before this PR, so the 'never remove existing' rule isn't violated — this is purely additive.

react/src/components/AdminUserManagement.tsx
ironAiken2: Thanks for catching this. Updating the PR description to reflect the actual implementation: fragments are spread on the list query (`...UserInfoModalFragment` / `...UserSettingModalFragment` on `adminUsersV2.edges`) and the node is looked up in the already-loaded edges — there is no second `useLazyLoadQuery` and no `limit: 1` filter.

nowgnuesLee (approved):
LGTM

PR #7686, opened 2026-06-02
fix(FR-3017): version-branch myRoles entityType filter shape

### LINKED ISSUE #7665: myRoles permission filter sends RBACElementType as EnumFilter, breaking 26.4.x backends
## Background

`useCurrentUserProjectRoles` issues a `myRoles` query whose filter sends `permission.entityType` as an **EnumFilter** object:

```
myRoles(first: 100, filter: { permission: { entityType: { equals: PROJECT_ADMIN_PAGE } } })
```

This matches the **latest** schema (`PermissionNestedFilter.entityType: RBACElementTypeFilter`). However, older backends (observed on **26.4.3**) define this field as `RBACElementType` (a **direct enum**), so they reject the request with:

```
Enum "RBACElementType" cannot represent non-enum value: {equals: PROJECT_ADMIN_PAGE}
(GRAPHQL_VALIDATION_FAILED)
```

Because this is a **query validation** error (not a field/runtime error), the `@catch(to: RESULT)` on the field does **not** absorb it — the whole query is rejected and the app fails to load against 26.4.x backends.

## Temporary workaround (NOT the real fix)

While testing FR-3016 against a 26.4.3 backend, the query was changed locally to send the enum directly (`entityType: PROJECT_ADMIN_PAGE`) and `data/schema.graphql` `PermissionNestedFilter.entityType` was set to `RBACElementType` so the local Relay compile passes. This workaround is **incompatible with the latest backend** (which expects the Filter form), so it must not ship as-is.

## Proper fix (options to decide)

- (a) Branch the sent form by backend capability: send `RBACElementTypeFilter` (`{equals`}) on new cores and the direct enum on older ones (via a `baiClient.supports(...)` flag).
- (b) Have the backend accept the EnumFilter form on the affected older releases.
- (c) Define/raise the minimum supported core version so the Filter form is always valid.

## Acceptance Criteria

- Both super-admin and regular users can load the app (myRoles resolves) without `GRAPHQL_VALIDATION_FAILED` across the supported backend range (26.4.x through latest).
- No local `data/schema.graphql` divergence is required.

JIRA Issue: FR-3017

nowgnuesLee:
Resolves #7665 (FR-3017)

## Problem

`useCurrentUserProjectRoles` issued the `myRoles` query with `permission.entityType` always as the **`RBACElementTypeFilter` wrapper** (`{ equals: PROJECT_ADMIN_PAGE }`). Cores **<= 26.4.3** define this field as a **direct `RBACElementType` enum** and reject the wrapper at **query-validation** time:

```
Enum "RBACElementType" cannot represent non-enum value: {equals: PROJECT_ADMIN_PAGE}
(GRAPHQL_VALIDATION_FAILED)
```

Because it is a query-validation error (not a field error), `@catch(to: RESULT)` cannot absorb it — the whole query is rejected and the app **fails to load** against those backends.

## Fix (option a — version branch)

- **`backend.ai-client`**: add a `rbac-element-type-filter` capability flag, gated at **`26.4.4rc4`** — the exact tag where the backend switched `PermissionNestedFilter.{scope,entity}Type` to the `RBACElementTypeFilter` wrapper (backend.ai #11442).
- **`useCurrentUserProjectRoles`**: pass the `permission` filter as a **Relay variable** so a single compiled query carries either shape, and build it at runtime:
  - `>= 26.4.4rc4` → wrapper `{ entityType: { equals: PROJECT_ADMIN_PAGE } }`
  - `<= 26.4.3` / `26.4.4rc1-3` → direct enum `{ entityType: PROJECT_ADMIN_PAGE }`

The generated Relay type only models the latest (wrapper) shape, so the legacy branch is cast (`as unknown as ...`).

### Why gate on the exact rc
`26.4.4rcN` builds already carry the wrapper, but PEP440 sorts `26.4.4rcN < 26.4.4`. Gating at the stable `26.4.4` would misclassify rc4–rc6 as older and send them the direct enum, re-triggering the bug. Gating at `26.4.4rc4` classifies every release correctly:

| version | branch |
|---|---|
| <= 26.4.3 | direct enum |
| 26.4.4rc1–3 (pre-wrapper) | direct enum |
| 26.4.4rc4–rc6, 26.4.4, 26.5.x+ | wrapper |

## Verification

Automated — `bash scripts/verify.sh`: **Lint / Format / TypeScript PASS**, Relay regenerated.
(The harness's "Vite warmup paths" check also fails on a clean `main` — pre-existing, unrelated to this change.)

### TODO — manual checks (cannot be covered by verify.sh)

`verify.sh` only compiles against the latest (wrapper) schema; it cannot exercise the `<= 26.4.3` path. Confirm both branches against live backends:

**Wrapper path — backend `>= 26.4.4rc4`:**
- [ ] Super-admin loads the app; no `GRAPHQL_VALIDATION_FAILED` in console/network
- [ ] Regular (non-admin) user loads the app; `myRoles` resolves
- [ ] Network tab: `myRoles` request sends `permission.entityType = { equals: PROJECT_ADMIN_PAGE }`

**Direct-enum path — backend `<= 26.4.3`:**
- [ ] Super-admin and regular user both load the app without `GRAPHQL_VALIDATION_FAILED`
- [ ] Network tab: `myRoles` request sends `permission.entityType = PROJECT_ADMIN_PAGE` (bare enum)
- [ ] Project-admin scope still resolves (admin UI gating unchanged)

**Regression:**
- [ ] No `data/schema.graphql` divergence required (acceptance criterion from FR-3017)

> Note: after pulling this change onto a running dev session, do a **full browser refresh** — `_features` is cached on the global `backendaiclient`, so HMR alone may keep the stale capability map.

### DIFF
--- a/react/src/hooks/useCurrentUserProjectRoles.ts
+++ b/react/src/hooks/useCurrentUserProjectRoles.ts
@@ -3,7 +3,11 @@
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
 import { useSuspendedBackendaiClient } from '.';
-import { useCurrentUserProjectRolesQuery } from '../__generated__/useCurrentUserProjectRolesQuery.graphql';
+import {
+  useCurrentUserProjectRolesQuery,
+  PermissionNestedFilter,
+  RBACElementType,
+} from '../__generated__/useCurrentUserProjectRolesQuery.graphql';
 import { useCurrentProjectValue } from './useCurrentProject';
 import { graphql, useLazyLoadQuery } from 'react-relay';
 
@@ -36,12 +40,23 @@ export interface CurrentUserProjectRolesResult {
 export const useCurrentUserProjectRoles = (): CurrentUserProjectRolesResult => {
   const baiClient = useSuspendedBackendaiClient();
 
+  // <= 26.4.3 takes the bare enum; >= 26.4.4rc4 takes the wrapper. FR-3017.
+  const PROJECT_ADMIN_PAGE = 'PROJECT_ADMIN_PAGE' satisfies RBACElementType;
+  const permissionFilter: PermissionNestedFilter = {
+    // Cast confined to the one field the generated type can't model.
+    entityType: (baiClient.supports('rbac-element-type-filter')
+      ? { equals: PROJECT_ADMIN_PAGE }
+      : PROJECT_ADMIN_PAGE) as PermissionNestedFilter['entityType'],
+  };
+
   const data = useLazyLoadQuery<useCurrentUserProjectRolesQuery>(
     graphql`
-      query useCurrentUserProjectRolesQuery {
+      query useCurrentUserProjectRolesQuery(
+        $permissionFilter: PermissionNestedFilter
+      ) {
         myRolesResult: myRoles(
           first: 100
-          filter: { permission: { entityType: { equals: PROJECT_ADMIN_PAGE } } }
+          filter: { permission: $permissionFilter }
         ) @catch(to: RESULT) {
           edges {
             node {
@@ -62,7 +77,7 @@ export const useCurrentUserProjectRoles = (): CurrentUserProjectRolesResult => {
         }
       }
     `,
-    {},
+    { permissionFilter },
     {
       // store-or-network keeps the result cached across pages for the session.
       fetchPolicy: baiClient.supports('my-roles')
--- a/packages/backend.ai-client/src/client.ts
+++ b/packages/backend.ai-client/src/client.ts
@@ -570,10 +570,7 @@ export class Client {
     const v4_replacements: Record<string, string> = {
       session_name: 'sess_id',
     };
-    if (
-      this._apiVersionMajor !== null &&
-      Number(this._apiVersionMajor) < 5
-    ) {
+    if (this._apiVersionMajor !== null && Number(this._apiVersionMajor) < 5) {
       // For V3/V4 API compatibility
       Object.keys(v4_replacements).forEach((key) => {
         let index = fields.indexOf(key);
@@ -829,6 +826,13 @@ export class Client {
     }
     if (this.isManagerVersionCompatibleWith('26.4.3')) {
       this._features['model-deployment-extended-filter'] = true;
+    }
+    // entityType filter became the RBACElementTypeFilter wrapper in 26.4.4rc4
+    // (#11442); older cores take the direct enum. FR-3017.
+    // TODO(FR-3017): pinned to the rc tag since 26.4.4rcN < 26.4.4 in PEP440;
+    // simplify to '26.4.4' once rc builds are out of use.
+    if (this.isManagerVersionCompatibleWith('26.4.4rc4')) {
+      this._features['rbac-element-type-filter'] = true;
     }
   }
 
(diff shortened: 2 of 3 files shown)


nowgnuesLee:
* **#7686** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7686).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/hooks/useCurrentUserProjectRoles.ts
```diff
@@ -36,12 +36,24 @@ export interface CurrentUserProjectRolesResult {
...
 
+  // entityType is a direct `RBACElementType` enum on cores <= 26.4.3 but the
+  // `RBACElementTypeFilter` wrapper on >= 26.4.4rc4. Sent as a variable so one
+  // compiled query carries either shape; the generated type models only the
+  // wrapper, hence the cast. See `rbac-element-type-filter` flag. FR-3017.
+  const permissionFilter = (baiClient.supports('rbac-element-type-filter')
+    ? { entityType: { equals: 'PROJECT_ADMIN_PAGE' } }
+    : {
+        entityType: 'PROJECT_ADMIN_PAGE',
+      }) as unknown as useCurrentUserProjectRolesQuery['variables']['permissionFilter'];
```
agatha197: **non-blocking nit (type safety)**

`as unknown as ...` turns off type checking for the entire object, so it won't catch enum typos or a wrong field name. The only thing the generated type can't express is the single `entityType` field, so we can narrow the cast to just that field and let everything else be validated against `PermissionNestedFilter`:

```ts
import type {
  PermissionNestedFilter,
  RBACElementType,
} from '../__generated__/useCurrentUserProjectRolesQuery.graphql';

const PROJECT_ADMIN_PAGE = 'PROJECT_ADMIN_PAGE' satisfies RBACElementType;

const permissionFilter: PermissionNestedFilter = {
  // <= 26.4.3 takes the bare enum; >= 26.4.4rc4 takes the wrapper.
  // The only field the generated type can't model is this one, so the cast is confined here.
  entityType: (baiClient.supports('rbac-element-type-filter')
    ? { equals: PROJECT_ADMIN_PAGE }
    : PROJECT_ADMIN_PAGE) as PermissionNestedFilter['entityType'],
};
```

This gets all three: ① `unknown` removed, ② the whole object is validated as `PermissionNestedFilter`, ③ enum typos are caught at compile time by `satisfies` (verified with `tsc --strict`).

Note: you can't just swap `as unknown as` → `satisfies` or drop `unknown` outright — the legacy `string` branch isn't assignable to `PermissionNestedFilter`, so the cast has to be confined to the field as above.

Since this is a generated-type vs legacy-wire-format mismatch, some unsoundness is inherent — feel free to take the lighter route of just adding `satisfies RBACElementType` to the two literals while keeping the existing cast if you'd rather minimize the diff. Not a blocker either way.

agatha197 (approved):
LGTM

PR #7699, opened 2026-06-04
feat(FR-3032): add BAIAuditLogNodes common audit-log table to BUI

nowgnuesLee:
Resolves #7700 (FR-3032)

> **Stacked on** #7550 (`feat(FR-2949): add AuditLogTable component spec`). Review/merge that spec PR first.
>
> Implements the spec authored in **FR-2949** (`.specs/FR-2949-audit-log-table/spec.md`), under Epic **FR-2920**.

## Scope

Implements only the **공통 뷰 (common view)** of the audit-log table as a reusable BUI component, per the spec's `공통 뷰 — 레이아웃 / 컬럼 / 정렬·페이지네이션` sections. Follows the thin `*Nodes` precedent (`BAISchedulingHistoryNodes`): a presentational table over a Relay fragment; **filters, pagination, and per-surface query orchestration stay in the consuming surface**.

## Changes (`packages/backend.ai-ui/`)

- **`BAIAuditLogNodes`** (`components/fragments/`) — table over an `AuditLogV2 @relay(plural: true)` fragment.
  - Visible columns: **Time · Operation · Actor · Status · Description · Duration**.
  - Hidden via column toggle (`defaultHidden`): **Request ID · Action ID** (copyable `BAIId`).
  - Sortable on `createdAt / operation / status` → `AuditLogOrderField` (`ENTITY_TYPE` omitted — single-resource scope).
  - Actor reads `user.basicInfo.email` directly (backend resolves from `triggered_by`; no separate lookup, no UUID fallback).
  - Description: `BAIText` ellipsis + hover tooltip. Exposes `customizeColumns` / `disableSorter` / `onChangeOrder` + table passthrough.
- **`BAIAuditLogStatusTag`** (+ story) — semantic status badge: `SUCCESS`→success, `ERROR`→error, `RUNNING`→info+ripple, `UNKNOWN`→outline dot.
- `comp:BAIAuditLogNodes` i18n keys (en/ko); component exports; generated Relay fragment.

## Not in this PR (deferred to follow-up sub-tasks under FR-2920)

Filter bar (`BAIGraphQLPropertyFilter`: date / status / operation / actor), cursor pagination (**default page size 10**, options `[10, 50, 100]`), render-as-you-fetch tab hosting, version gating, and the Session / VFolder / Deployment entry points.

## Verification

[...]

PR #7550, opened 2026-05-21
feat(FR-2949): add AuditLogTable component spec

nowgnuesLee:
Resolves #7549 (FR-2949)

## Summary

- Drafts the spec for a reusable **`AuditLogTable`** component that surfaces per-entity audit log entries (Time, Operation, Actor, Status, Description, Duration, Request ID, Action ID) for VFolder / Session / Deployment, scoped to a single resource via `scopedAuditLogsV2` + `AuditLogFilter.entityId`.
- Defines the component contract only — props (`entityType`, `entityId`), columns (with `customizeColumns` toggling), filters (date range, status, operation as free-text, actor via UserSelect), cursor pagination via `usePaginationFragment` + `BAITable`, sorting, loading / empty / error / unsupported states, and the `useAuditLogV2Available()` readiness gate. Page-level embedding decisions are out of scope and will land in a follow-up spec under the FR-2920 Epic.
- Syncs `data/schema.graphql` to expose the unreleased audit-log surface the spec relies on: `scopedAuditLogsV2`, `AuditLogScope`, `EntityTypeScope`, `UUIDScope`, `AuditLogFilter.entityId`, plus an unrelated `ModelRevision.deploymentId` and `UpdateRuntimeVariantInput.description` semantic clarification carried by the same backend snapshot.

## Backend dependencies (blocked-by)

The component is staged behind a backend-readiness check and will only mount once both land:

- **RBAC relaxation** so non-admin callers may invoke `scopedAuditLogsV2` for resources they own.
- **`entity_id` filter support** on the user-facing v2 audit log API.

Both are tracked under Epic FR-2920.

## Files

- `.specs/FR-2949-audit-log-table/spec.md` — Korean spec body (filename stays `spec.md` per workflow).
- `.specs/FR-2949-audit-log-table/metadata.json` — Epic / Task / source-issue / phase / language mapping.
- `data/schema.graphql` — backend snapshot sync (audit-log additions + `ModelRevision.deploymentId` + minor input rewording).

## Test plan

- [ ] Spec body renders correctly and links resolve.
- [ ] All Acceptance Criteria in the spec are verifiable post-backend.
- [ ] `data/schema.graphql` diff matches the upstream backend snapshot (no hand-edits beyond the sync).
- [ ] Reporter approves the spec before any implementation sub-tasks are filed under FR-2920.

PR #7703, opened 2026-06-04
feat(FR-3033): add ScopedAuditLog preloaded audit-log view (filter + refresh)

### LINKED ISSUE #7701: Add ScopedAuditLog preloaded-query reader for scopedAuditLogsV2
Add a reusable BUI container that queries `scopedAuditLogsV2` and renders the common audit-log view by composing `BAIAuditLogNodes` (FR-3032) with filtering and refresh.

## Component

`BAIScopedAuditLogTable` (`packages/backend.ai-ui/src/components/fragments/`).

### Props

- `scope: AuditLogScope` — passed directly to the `scopedAuditLogsV2(scope: ...)` query argument.
- `entityId?: string` — AND-combined into `AuditLogFilter.entityId` (`{ equals: entityId `}) so the table is scoped to a single resource without overwriting user-applied filters (AC2: no other resource's logs leak in).
- Remaining `BAIAuditLogNodes` table props pass through.

### Behavior (mirrors the `SessionSchedulingHistoryModal` idiom)

- `useLazyLoadQuery` on `scopedAuditLogsV2(scope, filter, orderBy)`, feeding `edges.map(node)` to `BAIAuditLogNodes` (plural fragment spread on `edges.node`).
- **Filter**: `BAIGraphQLPropertyFilter` — status (enum: SUCCESS/ERROR/UNKNOWN/RUNNING → `AuditLogStatusFilter`), operation (string contains), created-at (datetime range).
- **Refresh**: `BAIFetchKeyButton` + `useFetchKey`.
- **Keep previous rows** on filter/sort/refresh via `useDeferredValue` (no blank flash).
- Default sort time DESC; sortable on time/operation/status.

## Deferred (call out explicitly; follow-up sub-tasks)

- **Actor filter (user-selector)** — needs a user-list query a non-admin (Phase 2 RBAC) may not be able to run; deferred until the picker + UUID-resolution story is settled.
- **Server-side cursor pagination** — v1 uses client-side `BAITable` pagination (page size 10, options [10, 50, 100]) over the scoped result set, exactly like the scheduling-history precedent. Server cursor pagination lands when the backend `scopedAuditLogsV2` is ready.
- **Surface-level version gating + entry points** (sub-tab / button) — the Session / VFolder / Deployment integration tasks own mounting + gating; this component assumes it is only mounted on supported backends.

## Acceptance

- `bash scripts/verify.sh` — Relay / Lint / Format / TypeScript pass (compile/type-level; backend does not serve `scopedAuditLogsV2` yet, so not runtime-verified).
- Exported from `backend.ai-ui`; a surface can drop it in with only `scope` + `entityId`.

Builds on the spec authored in FR-2949 (`.specs/FR-2949-audit-log-table/spec.md`), Epic FR-2920.

JIRA Issue: FR-3033

nowgnuesLee:
Resolves #7701 (FR-3033)

> **Stacked on** #7699 (`feat(FR-3032): add BAIAuditLogNodes common audit-log table to BUI`). Review/merge that PR first — this composes `BAIAuditLogNodes` from it.
>
> Builds on the spec authored in **FR-2949** (`.specs/FR-2949-audit-log-table/spec.md`), Epic **FR-2920**. Blocked by FR-3032.

## What

`ScopedAuditLog` (`react/src/components/ScopedAuditLog.tsx`) — reads a **preloaded** `scopedAuditLogsV2` query and renders a `BAIAuditLogNodes` (FR-3032) table with a filter bar + refresh. Mirrors the **`DeploymentSchedulingHistoryModal`** idiom (inline, no modal).

- **The consumer owns `useQueryLoader` / `loadQuery`** and builds the variables — it puts the resource scope (entity type / id, triggering user UUID) into `AuditLogScope`, starts the request (render-as-you-fetch), and passes `queryRef` + `loadQuery` (as `onReload`).
- **This view** reads `queryRef` via `usePreloadedQuery` and re-runs filter / sort / refresh through `onReload({ ...queryRef.variables, filter | orderBy }, { fetchPolicy: 'network-only' })`. The resource scope is carried forward in `queryRef.variables`, so only `filter` / `orderBy` are overridden.
- **Keep previous rows**: reads the **deferred** `queryRef` (`useDeferredValue`), so a re-fetch shows the old rows with an inline loading indicator instead of a Suspense flash.

### Props

- **`queryRef: PreloadedQuery<ScopedAuditLogQuery>`** — from the consumer's `useQueryLoader`.
- **`onReload: (variables, options?) => void`** — the consumer's `loadQuery`, called on filter/sort/refresh.
- Plus `BAIAuditLogNodes` table passthrough.

### Behavior

- **Filter**: `BAIGraphQLPropertyFilter` — status (enum, strict → `{ status: { equals } }`), operation (`contains`), created-at (range). **Refresh**: `BAIFetchKeyButton`. **Sort**: `convertToOrderBy`, default time DESC.
- **i18n**: host `auditLog.*` keys (en/ko). Exports the `ScopedAuditLogQuery` node for the consumer's `useQueryLoader`. Render inside `Suspense` + `BAIErrorBoundary`.

```tsx
const [queryRef, loadQuery] = useQueryLoader(ScopedAuditLogQuery);
// on tab click: loadQuery({ scope, orderBy: [{ field: 'CREATED_AT', direction: 'DESC' }] }, { fetchPolicy: 'store-and-network' });
<Suspense fallback={<Skeleton active />}>
  {queryRef && <ScopedAuditLog queryRef={queryRef} onReload={loadQuery} />}
</Suspense>
```

## Owned by the consuming surface (deferred, under FR-2920)

`useQueryLoader` wiring, building the `scope`, version gating, and entry points (the "Audit Log" sub-tab) live with the Session / VFolder / Deployment integration tasks.

## Verification

`bash scripts/verify.sh`: **Relay / Lint / Format / TypeScript PASS** (the lone `FAIL`, "Vite warmup paths", is the pre-existing `react/vite.config.ts` bug, failing identically on clean `main`).

Compile/type-level — the backend does not serve `scopedAuditLogsV2` yet, so the render path isn't runtime-verified.

### DIFF
--- a/react/src/components/ScopedAuditLog.tsx
+++ b/react/src/components/ScopedAuditLog.tsx
@@ -1,0 +1,197 @@
+import type {
+  AuditLogOrderBy,
+  AuditLogScope,
+  ScopedAuditLogQuery as ScopedAuditLogQueryType,
+} from '../__generated__/ScopedAuditLogQuery.graphql';
+import { convertToOrderBy } from '../helper';
+import {
+  BAIAuditLogNodes,
+  type BAIAuditLogNodesProps,
+  BAIFetchKeyButton,
+  BAIFlex,
+  BAIGraphQLPropertyFilter,
+  filterOutNullAndUndefined,
+  useFetchKey,
+} from 'backend.ai-ui';
+import * as _ from 'lodash-es';
+import { useDeferredValue } from 'react';
+import { useTranslation } from 'react-i18next';
+import {
+  graphql,
+  PreloadedQuery,
+  usePreloadedQuery,
+  UseQueryLoaderLoadQueryOptions,
+} from 'react-relay';
+
+export type { AuditLogScope };
+
+export const ScopedAuditLogQuery = graphql`
+  query ScopedAuditLogQuery(
+    $scope: AuditLogScope!
+    $filter: AuditLogFilter
+    $orderBy: [AuditLogOrderBy!]
+    $limit: Int
+    $offset: Int
+  ) {
+    scopedAuditLogsV2(
+      scope: $scope
+      filter: $filter
+      orderBy: $orderBy
+      limit: $limit
+      offset: $offset
+    ) {
+      count
+      edges {
+        node {
+          ...BAIAuditLogNodesFragment
+        }
+      }
+    }
+  }
+`;
+
+export interface ScopedAuditLogProps extends Omit<
+  BAIAuditLogNodesProps,
+  | 'auditLogFrgmt'
+  | 'loading'
+  | 'order'
+  | 'onChangeOrder'
+  | 'dataSource'
+  | 'pagination'
+> {
+  queryRef: PreloadedQuery<ScopedAuditLogQueryType>;
+  onReload: (
+    variables: ScopedAuditLogQueryType['variables'],
+    options?: UseQueryLoaderLoadQueryOptions,
+  ) => void;
+}
+
+/**
+ * ScopedAuditLog - Reads a preloaded `scopedAuditLogsV2` query and renders it as
+ * a `BAIAuditLogNodes` table with a filter bar and a refresh button. Mirrors the
+ * `DeploymentSchedulingHistoryModal` idiom: the consumer owns `useQueryLoader` /
+ * `loadQuery` (and the scope); this view reads the `queryRef` via
+ * `usePreloadedQuery` and re-runs filter / sort / refresh through `onReload`.
+ * Reading the *deferred* `queryRef` keeps the previous rows visible while the
+ * next result loads (inline loading, no Suspense-fallback flash). Render inside a
+ * `Suspense` (+ `BAIErrorBoundary`) boundary.
+ */
+const ScopedAuditLog = ({
+  queryRef,
+  onReload,
+  ...tableProps
+}: ScopedAuditLogProps) => {
+  'use memo';
+  const { t } = useTranslation();
+  const [fetchKey, updateFetchKey] = useFetchKey();
+
+  const filter = queryRef.variables.filter ?? undefined;
+  const orderBy = queryRef.variables.orderBy?.[0];
+  const order = orderBy
+    ? `${orderBy.direction === 'DESC' ? '-' : ''}${_.camelCase(orderBy.field)}`
+    : null;
+  const pageSize = queryRef.variables.limit ?? 10;
+  const offset = queryRef.variables.offset ?? 0;
+  const current = pageSize ? Math.floor(offset / pageSize) + 1 : 1;
+
+  const deferredQueryRef = useDeferredValue(queryRef);
+  const isRefetching = deferredQueryRef !== queryRef;
+
+  const data = usePreloadedQuery<ScopedAuditLogQueryType>(
+    ScopedAuditLogQuery,
+    deferredQueryRef,
+  );
+
+  return (
+    <BAIFlex direction="column" align="stretch" gap="sm">
+      <BAIFlex justify="between" wrap="wrap" gap="sm">
+        <BAIGraphQLPropertyFilter
+          value={filter}
+          onChange={(next) => {
+            onReload(
+              { ...queryRef.variables, filter: next, offset: 0 },
+              { fetchPolicy: 'network-only' },
+            );
+          }}
+          filterProperties={[
+            {
+              key: 'status',
+              propertyLabel: t('auditLog.Status'),
+              type: 'enum',
+              strictSelection: true,
+              options: [
+                { label: 'SUCCESS', value: 'SUCCESS' },
+                { label: 'ERROR', value: 'ERROR' },
+                { label: 'RUNNING', value: 'RUNNING' },
+                { label: 'UNKNOWN', value: 'UNKNOWN' },
+              ],
+            },
+            {
+              key: 'operation',
+              propertyLabel: t('auditLog.Operation'),
+              type: 'string',
+              fixedOperator: 'contains',
+            },
+            {
+              key: 'triggeredBy',
+              propertyLabel: t('auditLog.TriggeredBy'),
+              type: 'string',
+              fixedOperator: 'contains',
+            },
+            {
+              key: 'createdAt',
+              propertyLabel: t('auditLog.Time'),
+              type: 'datetime',
+              defaultOperator: 'after',
+            },
+          ]}
+        />
+        <BAIFetchKeyButton
+          value={fetchKey}
+          onChange={(key) => {
+            updateFetchKey(key);
+            onReload(queryRef.variables, { fetchPolicy: 'network-only' });
+          }}
+          loading={isRefetching}
+          autoUpdateDelay={7_000}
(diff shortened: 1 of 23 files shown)


nowgnuesLee:
* **#7809** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7735** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7735?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7729** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7729?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7720** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7720?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7703** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7703?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7703?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7699** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7699?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7550** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7550?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7703).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/ScopedAuditLog.tsx
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Re-run the query when filter / order / refresh changes. Same signature as
+   * `useQueryLoader`'s `loadQuery`, so the consumer can pass it directly. The
+   * resource scope lives in `scope` (carried forward via `queryRef.variables`),
+   * so re-fetches override only `filter` / `orderBy`.
+   */
+  onReload: (
+    variables: ScopedAuditLogQueryType['variables'],
+    options?: UseQueryLoaderLoadQueryOptions,
+  ) => void;
```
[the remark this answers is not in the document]
nowgnuesLee: This should be fine now, as I’ve changed the implementation approach.

react/src/components/ScopedAuditLog.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+export type { AuditLogScope };
+
+export const ScopedAuditLogQuery = graphql`
+  query ScopedAuditLogQuery(
+    $scope: AuditLogScope!
+    $filter: AuditLogFilter
+    $orderBy: [AuditLogOrderBy!]
+  ) {
+    scopedAuditLogsV2(scope: $scope, filter: $filter, orderBy: $orderBy) {
```
agatha197: **No pagination args on a paginated V2 connection.** `scopedAuditLogsV2` is a Strawberry V2 connection that accepts `first/last/limit/offset` (+ `before/after`) and exposes `pageInfo` and a total `count`, but this query requests none of them and reads neither `pageInfo` nor `count`. As a result only the server's default page of audit entries is fetched — there is no "load more" or page control, so older entries beyond that first page are unreachable. `BAIAuditLogNodes` paginating client-side over whatever it receives makes this misleading (the user pages within only the first server page). Suggest wiring real pagination — `limit`/`offset` for table-style or `first`/`after` for infinite scroll — and at minimum surfacing `count`. (Backend doesn't serve this field yet, so this is a static observation, not runtime-verified.)

react/src/components/ScopedAuditLog.tsx
```diff
@@ -0,0 +1,160 @@
...
+const ScopedAuditLog = ({
+  queryRef,
+  onReload,
+  ...tableProps
+}: ScopedAuditLogProps) => {
+  'use memo';
+  const { t } = useTranslation();
+  const [fetchKey, updateFetchKey] = useFetchKey();
+  const [filter, setFilter] = useState<AuditLogFilter>();
+  const [order, setOrder] = useState<string | null>('-createdAt');
```
agatha197: **Local `filter`/`order` state is decoupled from `queryRef.variables`.** Both are initialized to hardcoded defaults (`filter` undefined, `order` `'-createdAt'`) independent of the variables the consumer preloaded the query with. This assumes every consumer starts the query with exactly `{ filter: undefined, orderBy: CREATED_AT DESC }`. It happens to hold for the current sole consumer, but any consumer that preloads a different initial sort/filter would render a filter bar / sort indicator that disagrees with the actual query. Consider deriving the initial state from `queryRef.variables`, or documenting the contract explicitly.

react/src/components/ScopedAuditLog.tsx
```diff
@@ -0,0 +1,160 @@
...
+          setOrder(nextOrder);
+          onReload(
+            {
+              ...queryRef.variables,
+              orderBy: convertToOrderBy<AuditLogOrderBy>(nextOrder),
+            },
+            { fetchPolicy: 'network-only' },
+          );
+        }}
+        auditLogFrgmt={_.map(data.scopedAuditLogsV2?.edges, 'node')}
```
agatha197: Nit: `_.map(data.scopedAuditLogsV2?.edges, 'node')` can include `null` nodes (the connection's edge `node` is nullable). Consider `filterOutNullAndUndefined(...)` here, as `ConnectedKernelList` does for its kernel nodes, so `BAIAuditLogNodes` never receives null rows.

agatha197 (changes requested):
The preloaded-query idiom is clean and mirrors `DeploymentSchedulingHistoryModal` well — the deferred `queryRef` for no-flash refetch is a nice touch, and the filter/sort/scope wiring is type-correct against the schema (verified field-by-field: status→`equals`, operation/triggeredBy→`contains`, createdAt→`after/before`, orderBy field/direction, and the status enum options all match the generated input types).

A few things worth addressing before merge — mainly the missing pagination on a connection that clearly supports it (see inline). The other two are lower severity (a latent coupling in the local filter/sort state, and an edge-node null nit).

react/src/components/ScopedAuditLog.tsx
nowgnuesLee: I applied it. Thanks.

react/src/components/ScopedAuditLog.tsx
nowgnuesLee: I applied it. Thanks.

react/src/components/ScopedAuditLog.tsx
nowgnuesLee: I applied it. Thanks.

agatha197 (approved):
LGTM

PR #7724, opened 2026-06-04
style(FR-2977): reframe revision-source action wording and source icon

### LINKED ISSUE #7601: Refactor DeploymentAddRevisionModal to id-based revision source
Refactor `DeploymentAddRevisionModal` so that revision-source prefill flows through a single id-based GraphQL query and a single `useFragment` call, eliminating the dual-effect `pending*Prefill` state pump and the list-query eager-fetch that surfaced during review of FR-2957 (#7561).

## Why This Work Is Needed

Surfacing the "Duplicate as new revision" action (FR-2957) made the modal reuse the same prefill plumbing that exists for "Load current revision". On a careful look the current design has real React/Relay smells:

1. **Effect cascade with state pump** — `applyDuplicatePrefill` stashes data into `pendingDuplicatePrefill` state and switches mode; a second `useEffect` reads the pending state, applies the prefill, then clears it. Two `useEffectEvent`s + intermediate state replace what should be a single derived effect. Same anti-shape exists already for `presetTransferPrefill` and `customTransferPrefill` (introduced in FR-2862) — three instances of the same pattern in one file.
1. **Fragment ref held in `useState`** — `DeploymentRevisionHistoryTab` stores a Relay fragment ref (`DeploymentAddRevisionModal_duplicateRevision$key`) in component state and threads it down to the modal. Fragment refs are Relay store handles; on cache eviction/GC they can go stale. The Relay-idiomatic alternative is to keep just the revision id in state and let the modal's own query resolve the source.
1. **List query over-fetch** — the revision-history list query spreads the full `DeploymentAddRevisionModal_duplicateRevision` fragment on every row (cluster config, resource slots, resource opts entries, runtime config, model mount config, extra mounts, image, model definition). Every visible revision pays the full prefill cost even though most users never click duplicate.
1. **Two `useFragment` calls reading the same fragment** — the modal calls `useFragment` once for `currentRevision` and once for `duplicateRevisionFrgmt`, both on the same fragment definition. The only reason is to satisfy `applyRevisionPrefill`'s single `$data` parameter type. This is a tell that the abstraction is off.
1. **`destroyOnHidden` masking coupling** — the modal is force-remounted on every open to avoid stale state. Works, but uses mount/unmount as the synchronization key instead of a clean data flow, and pays a full Relay query refetch each time.

## Proposed Approach

Land this refactor as a base PR. FR-2957 (PR #7561, currently in draft) will be moved on top as a stacked PR; its diff will shrink to just the i18n key, two menu items, the icon, and the `onClick` that sets the parent-owned mode + source id.

### Scope (DO)

- Add `modelRevision(id: $sourceRevisionId) @include(if: $hasSourceRevision) { ...RevisionSource `} to the modal's `useLazyLoadQuery`. Share `RevisionSource` between `currentRevision` and the duplicate source via a single fragment.
- Make `mode` a controlled prop on `DeploymentAddRevisionModal` (`mode: 'preset' | 'custom'`, `onModeChange`, `sourceRevisionId?: string | null`). The parent decides mode and source in the same tick when opening for duplicate — the async-`setMode` race disappears.
- Reduce the prefill machinery to a **single** effect: `if (mode === 'custom' && source && appliedSourceIdRef.current !== source.id) { customForm.resetFields(); applyRevisionPrefill(source); appliedSourceIdRef.current = source.id; `}.
- Remove the `pendingDuplicatePrefill` state and both `useEffectEvent`s introduced for the duplicate path.
- Remove the `...DeploymentAddRevisionModal_duplicateRevision` spread from `DeploymentRevisionHistoryTabListQuery` (rows only need fields they actually render).
- Revisit `destroyOnHidden` — once mode + source are controlled, it should be removable. Validate empirically before dropping.

### Scope (DO NOT)

- **Do not** touch `presetTransferPrefill` / `customTransferPrefill`. These transfer values between two different form shapes (preset ↔ custom form), which is a structurally different problem from the source-revision prefill. Unifying those is a separate follow-up.
- **Do not** change the modal's UX (alert copy, mode segmented control, Apply button placement). Pure data-flow refactor only.

## Acceptance Criteria

- `DeploymentAddRevisionModal`:
  - One `useFragment` call site for revision source (down from two).
  - Zero `useEffectEvent` calls for the source-revision prefill path (the two added in FR-2957 are gone; the two for preset/custom transfer remain unchanged).
  - `mode` and `sourceRevisionId` are received as props from the parent.
- `DeploymentRevisionHistoryTab`:
  - No fragment ref held in `useState`. Source revision id is the only state passed to the modal.
  - List query does not spread the source-revision fragment.
- All three existing flows still pass manual regression:
  1. "Add Revision" button (no source, preset mode by default).
  1. "Load current revision" alert button (source = current revision, custom mode).
  1. Mode toggle preserves form values via the existing preset↔custom transfer path (unchanged).
- `bash scripts/verify.sh` clean (Relay/Lint/Format/TS).

## Stack Plan

```
main
 └── (base) THIS issue  — refactor data flow
      └── (top) FR-2957 — duplicate-revision menu/i18n/icon only
```

## Related

- FR-2957 (#7561) — "Duplicate as new revision" feature, will be restacked on top of this.
- FR-2862 — introduced the preset/custom transfer state pump pattern (Done). Not within scope here.

JIRA Issue: FR-2977

yomybaby:
Resolves #7601 (FR-2977)

> **Stacked on #7671 (FR-2977).** That PR adds the "Add new revision from this" entry in the revision history row menu and the revision-detail drawer. This PR is a wording + icon polish on that same entry.

## Summary

The revision-source action seeds a **new linear revision** pre-filled from an existing revision's configuration — it does **not** fork or duplicate the history (see #7671 / #7561, which deliberately avoided a branching icon). The previous copy ("Add new revision from this" / KO "이 리비전으로 추가") leaned on a "duplicate" framing that misreads the actual behavior and read inconsistently next to the in-modal **"Load current revision"** alert, whose own copy already frames loading a revision's settings as the *starting point* for a new revision.

This PR aligns the entry with that "based on this / starting point" mental model:

- **Reframe the action label** (`deployment.AddNewRevisionFromThis`) across all 21 locales:
  - EN: `Add new revision from this` → **`New revision based on this`**
  - KO: `이 리비전으로 추가` → **`이 리비전 기반으로 리비전 생성`** (noun-ending menu style, matching `리비전 보기` / `배포 삭제`; makes "creates a revision" explicit)
  - The other 19 locales are updated to the same "new revision based on this" framing, keeping each locale's established term for *revision* (es de-accented to `Revision` and ja switched to Latin-script `Revision` to match each locale's existing deployment strings).
- **Switch the action icon** to lucide **`CopyPlus`** (`<CopyPlusIcon size={14} />`) in `DeploymentRevisionHistoryTab.tsx` — both entry points (the row overflow menu item and the drawer **More (⋯)** dropdown). It reads as "create a new revision from this one's settings" (copy **+** create new), distinct from the plain `Plus` used by the generic *Add Revision* button, without the fork/branch metaphor.

## Scope

- Out of scope by design: the in-modal alert keys (`deployment.LoadCurrentRevision`, `deployment.CurrentRevisionAvailableDescription`, `deployment.CurrentRevisionConfigurationLoaded`) are **unchanged** — the alert already uses the correct "Load / starting point" vocabulary, and the new action label now reads coherently alongside it.
- Pure copy + icon change. No behavioral or data-flow change.

## Test plan

- [x] Revision History row ⋯ menu shows the new label with the CopyPlus icon, opens the Add Revision modal prefilled from that revision.
- [x] Revision-detail drawer **More (⋯)** dropdown shows the same label + icon, same prefill behavior.
- [x] In-modal "Load current revision" alert copy/behavior unchanged.
- [x] `bash scripts/verify.sh` — Relay, Lint, Format, TypeScript all PASS.

### DIFF
--- a/react/src/components/DeploymentRevisionHistoryTab.tsx
+++ b/react/src/components/DeploymentRevisionHistoryTab.tsx
@@ -14,7 +14,6 @@ import DeploymentAddRevisionModal from './DeploymentAddRevisionModal';
 import DeploymentRevisionDetailDrawer from './DeploymentRevisionDetailDrawer';
 import FolderLink from './FolderLink';
 import {
-  CopyOutlined,
   LoadingOutlined,
   MoreOutlined,
   PlayCircleOutlined,
@@ -52,6 +51,7 @@ import {
 } from 'backend.ai-ui';
 import dayjs from 'dayjs';
 import * as _ from 'lodash-es';
+import { CopyPlusIcon } from 'lucide-react';
 import {
   parseAsInteger,
   parseAsString,
@@ -469,7 +469,7 @@ const DeploymentRevisionHistoryTab: React.FC<
               {
                 key: 'duplicate',
                 title: t('deployment.AddNewRevisionFromThis'),
-                icon: <CopyOutlined />,
+                icon: <CopyPlusIcon size={token.fontSize} />,
                 showInMenu: 'always',
                 disabled: isDeploymentInStoppedCategory(deploymentStatus),
                 onClick: () => {
@@ -664,7 +664,7 @@ const DeploymentRevisionHistoryTab: React.FC<
                       {
                         key: 'duplicate',
                         label: t('deployment.AddNewRevisionFromThis'),
-                        icon: <CopyOutlined />,
+                        icon: <CopyPlusIcon size={token.fontSize} />,
                         disabled:
                           isDeploymentInStoppedCategory(deploymentStatus),
                         onClick: () => {
--- a/resources/i18n/de.json
+++ b/resources/i18n/de.json
@@ -937,7 +937,7 @@
     "ActivePoolDescription": "Replikate, die derzeit über App Proxy Datenverkehr empfangen.",
     "ActivenessStatus": "Aktivitätsstatus",
     "AddAccessToken": "Zugriffstoken hinzufügen",
-    "AddNewRevisionFromThis": "Aus dieser Revision hinzufügen",
+    "AddNewRevisionFromThis": "Neue Revision basierend auf dieser",
     "AddRevision": "Revision hinzufügen",
     "AdditionalMounts": "Zusätzliche Einhängepunkte",
     "AnotherDeploymentInProgress": "Ein weiteres Deployment-Update wird bereits durchgeführt. Bitte warten Sie, bis es abgeschlossen ist, bevor Sie es erneut versuchen.",
--- a/resources/i18n/el.json
+++ b/resources/i18n/el.json
@@ -937,7 +937,7 @@
     "ActivePoolDescription": "Αντίγραφα που λαμβάνουν τρέχουσα κίνηση μέσω App Proxy.",
     "ActivenessStatus": "Κατάσταση ενεργότητας",
     "AddAccessToken": "Προσθήκη Token Πρόσβασης",
-    "AddNewRevisionFromThis": "Προσθήκη από αυτή τη Revision",
+    "AddNewRevisionFromThis": "Νέα Revision βάσει αυτής",
     "AddRevision": "Προσθήκη Revision",
     "AdditionalMounts": "Πρόσθετες Προσαρτήσεις",
     "AnotherDeploymentInProgress": "Μια άλλη ενημέρωση ανάπτυξης βρίσκεται ήδη σε εξέλιξη. Παρακαλούμε περιμένετε να ολοκληρωθεί πριν προσπαθήσετε ξανά.",
--- a/resources/i18n/en.json
+++ b/resources/i18n/en.json
@@ -924,7 +924,7 @@
     "ActivePoolDescription": "Replicas currently receiving traffic through App Proxy.",
     "ActivenessStatus": "Activeness Status",
     "AddAccessToken": "Add Access Token",
-    "AddNewRevisionFromThis": "Add new revision from this",
+    "AddNewRevisionFromThis": "New revision based on this",
     "AddRevision": "Add Revision",
     "AdditionalMounts": "Additional Mounts",
     "AnotherDeploymentInProgress": "Another deployment update is already in progress. Please wait for it to complete before trying again.",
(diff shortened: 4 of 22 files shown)


yomybaby:
* **#7743** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7743?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7724** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7724?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7724?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7671** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7671?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7561** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7561?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7602** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7602?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7724).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/DeploymentRevisionHistoryTab.tsx
```diff
@@ -464,7 +464,7 @@ const DeploymentRevisionHistoryTab: React.FC<
               {
                 key: 'duplicate',
                 title: t('deployment.AddNewRevisionFromThis'),
-                icon: <CopyOutlined />,
+                icon: <CopyPlusIcon size={14} />,
```
agatha197: please use token value

agatha197 (changes requested):

react/src/components/DeploymentRevisionHistoryTab.tsx
yomybaby: Done — replaced the hardcoded `14` with `token.fontSize` for the CopyPlus icon at both entry points (row overflow menu + drawer More dropdown), so it tracks the theme token instead of a magic number. (ac52c1f43)

agatha197 (approved):
LGTM

PR #7753, opened 2026-06-05
feat(FR-3056): use ModelHealthCheck enable flag for opt-in health check in revisions and presets

### LINKED ISSUE #7752: Use ModelHealthCheck enable flag for opt-in health check in deployment revisions and presets
Adopt the new ModelHealthCheck `enable` flag for model deployment revision presets and the add-revision flow, so health checks become explicitly opt-in instead of being disabled by nulling the whole object.

## Why This Work is Needed

Today the frontend disables a model deployment health check by sending `healthCheck: null`. Because `initialDelay`, `maxRetries`, etc. live inside `ModelHealthCheck`, the server cannot apply its default `initialDelay` (or other defaults) when health check is turned off — there is no object to attach them to.

Backend PR lablup/backend.ai#11863 (target 26.4.4) adds an `enable` field to `ModelHealthCheck` / `ModelHealthCheckInput` (`enable: Boolean! = false`). The frontend should stop nulling the object and instead always send the object with `enable: true/false`, leaving the other fields unset when disabled, so the server can seed defaults.

## Expected Outcomes

- `data/schema.graphql` synced with backend: `enable` added to `ModelHealthCheck` (output) and `ModelHealthCheckInput` (input), version-gated to 26.4.4.
- **Deployment Revision Preset create/update**: when the health-check checkbox is checked, payload sends `healthCheck.enable = true` with configured fields; when unchecked, sends `healthCheck.enable = false` with the other fields unset (instead of `healthCheck: null`).
- **Add Revision (custom mode)**: when any health check field is filled, send `enable = true`; otherwise `enable = false`. Add an enable checkbox consistent with the revision preset UX.
- **Version skew safety**: the `enable` field is only sent to managers >= 26.4.4; on older managers (<= 26.4.3) the existing null-based behavior is preserved so requests are not rejected. Fragments read `enable @since(version: "26.4.4")`.
- `bash scripts/verify.sh` passes (Relay compile, lint, format, tsc).

## References

- Backend PR: lablup/backend.ai#11863 (BA-6242, merged, target 26.4.4)
- Originating Teams thread (devops / Frontend channel)

JIRA Issue: FR-3056

agatha197:
Resolves #7752 (FR-3056)

## Summary

Adopts the new `ModelHealthCheck.enable` flag (backend PR lablup/backend.ai#11863) so health checks become explicitly opt-in instead of being disabled by nulling the whole object. The flag is gated behind **`26.4.4rc7`** until the final 26.4.4 release ships (TODO in `client.ts` to simplify to `26.4.4`).

- **Schema** (`data/schema.graphql`): synced `enable` onto `ModelHealthCheck` (output, `Boolean!`) and `ModelHealthCheckInput` (input, `Boolean! = false`), plus other additive backend fields.
- **Version-skew safety**: `enable` is gated behind `isManagerVersionCompatibleWith('26.4.4rc7')` on the input side and `@since(version: "26.4.4rc7")` on every fragment selection, so managers `< 26.4.4rc7` keep the legacy null-based behavior and never receive the unknown field.
- **Add Revision — custom command mode** (`DeploymentAddRevisionModal.tsx`): any filled health-check field → `enable = true`; all empty → `enable = false`. The health-check inputs serve as the toggle.
- **Edit/prefill correctness**: `enableHealthCheck` (preset) and the add-revision prefill derive from `enable` rather than object presence — a source revision with `enable: false` no longer pre-fills stale fields.

### Preset model-config form (`AdminDeploymentPresetModelConfigItem.tsx`)

- **Configure Service** is now an always-visible `BAICard` (with header divider) instead of a collapsible panel: the enable-service toggle is gone and the service block is always part of the model definition. **Add Metadata** remains a collapsible section.
- The health-check detail fields and the metadata section render through `Form.Item` `dependencies` / `shouldUpdate` render-props instead of `useWatch` + manual `useToggle`/`useEffect` sync.

### Preset validation requiredness

Preset create/update validate `model_definition` against the backend's **strict** `ModelDefinition` (`common/config.py`), which is stricter than the lenient revision path:

- **`service.port`** **is required** — no backend default.
- **Health-check detail fields are required while the check is enabled.** The fields carry backend defaults, but the GraphQL input defaults them to `null` and the strict model rejects an explicit `null`, so omitting them cannot fall back to the defaults. They render only while the check is enabled, so the `required` rules apply only then.
- **Disabling the health check sends** **`healthCheck: null`** (omits the object) rather than `{ enable: false }`, because a present `health_check` object triggers strict inner-field validation that then fails on the injected nulls.
- `shell` is required too (no hard-coded frontend fallback): it is a non-null `str` in the strict model, so it falls under the same constraint as the health-check fields.

## Test plan

- [x] `pnpm relay` + `bash scripts/verify.sh` → Relay / Lint / Format / TypeScript pass.
- [x] Preset with health check **off** → request sends `healthCheck: null`, create/update succeeds.
- [x] Preset with health check **on** and fields filled → succeeds.
- [x] Preset form blocks submit when `port` is blank, or when a health-check field is blank while the check is enabled.
- [ ] Manager `< 26.4.4rc7`: flows still send the legacy shape (`null` when disabled, no `enable` key) and do not error.
- [ ] Add a revision in custom command mode with/without a health-check path → `enable` reflects the path.

## Follow-ups (separate stacked PRs / backend)

- **Backend**: let preset create/update validate `model_definition` against the lenient `ModelDefinitionDraft` so unspecified fields stay unspecified (consistent with the revision path and the GraphQL nullable contract). That would let the health-check fields be truly optional on the frontend instead of required.
- Surface the **enable option in the add-revision flow**.
- Show the health-check state in the **preset info modal**.

### Deployment Presets List

deployment-presets-list

### Preset — Configure Service card with Enable Health Check

preset-configure-service-health-check

### Deployment Detail — Add Revision button

deployment-detail-add-revision-btn

### Add Revision Modal (Preset Mode)

add-revision-preset-mode


agatha197:
* **#7753** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7753?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#7767](https://github.com/lablup/backend.ai-webui/pull/7767) <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7767?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#7775](https://github.com/lablup/backend.ai-webui/pull/7775) <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7775?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7753?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7753).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1089,15 +1104,27 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
+                  healthCheck: (() => {
+                    // The path field doubles as the enable toggle: a filled path
+                    // means the health check is on, an empty one means it is off.
+                    const enabled = !!values.commandHealthCheck;
+                    const configuredFields = {
+                      path: values.commandHealthCheck,
+                      interval: values.commandInterval ?? 10,
+                      maxRetries: values.commandMaxRetries ?? 10,
+                      maxWaitTime: values.commandMaxWaitTime ?? 15,
+                      initialDelay: values.commandInitialDelay,
```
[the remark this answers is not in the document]
agatha197: Good catch — aligned the code to the requirement. `enabled` now derives from **any** filled health-check field (path, initial delay, max retries, interval, max wait time), not just the path, so it matches FR-3056's "enable when any health-check field is filled" wording. When enabled without an explicit path, the 26.4.4 server falls back to its default health-check path (the custom runtime variant now ships one). The dedicated enable checkbox was the optional ("would be nice") part of the request; the path/field inputs serve as the toggle here, consistent with the compact command-mode form. Fixed in d44bdd77f.

nowgnuesLee (dismissed):
Please clean up useless comments.

agatha197:
@nowgnuesLee Cleaned up the comments in cae831e1c:

- Removed the decorative JSX section dividers (`{/* ── Name + Path ── */}`, `{/* Pre-start actions */}`, `{/* Health check */}`, etc.) in `AdminDeploymentPresetModelConfigItem.tsx`.
- Dropped the duplicated/over-verbose narration (e.g. the redundant model-definition JSX comment in `AdminDeploymentPresetSettingPageContent.tsx`, the box-drawing header in `ModelConfigItem`).
- Condensed the remaining comments, keeping only the load-bearing rationale (version-skew gating, strict `ModelDefinition` validation / null-vs-`{enable:false}`).

Net −22 lines. `prettier --check` passes on the touched files. PTAL 🙏

react/src/components/AdminDeploymentPresetModelConfigItem.tsx
```diff
@@ -2,291 +2,219 @@
...
+          <BAIFlex gap="md" wrap="wrap">
+            <Form.Item
+              {...restField}
+              name={[listItemName, 'service', 'port']}
+              label={t('adminDeploymentPreset.modelDef.Port')}
+              style={{ flex: 1, minWidth: 100 }}
+              rules={[{ required: true }]}
+            >
+              <InputNumber
+                min={1}
```
ironAiken2: The backend rejects `port = 1` — `PresetModelServiceConfigInput.port` is declared with `gt=1` (**exclusive**), so the minimum accepted value is **2**:

```
1 validation error for PresetModelServiceConfigInput
port
  Input should be greater than 1 [type=greater_than, input_value=1, input_type=int]
```

```suggestion
                min={2}
```

The same constraint applies to the **port field in `DeploymentAddRevisionModal.tsx`** (`min={1}` around line 1722, outside this diff): the lenient `ModelServiceConfigDraft` resolves into the strict `ModelServiceConfig` (`gt=1`) via `to_resolved()`, so `port = 1` fails there too.

More broadly, please audit every numeric form item in this PR against the backend pydantic constraints (`PresetModel*Input` in `common/dto/manager/v2/deployment_revision_preset/request.py` and `ModelHealthCheck`/`ModelServiceConfig` in `common/config.py`) and mirror them as `InputNumber` `min`/`max` so invalid values are blocked in the UI instead of surfacing as a generic backend error.

react/src/components/AdminDeploymentPresetModelConfigItem.tsx
```diff
@@ -362,13 +295,14 @@ const ModelConfigItem: React.FC<{
                       label={t(
                         'adminDeploymentPreset.modelDef.HealthCheckExpectedStatus',
                       )}
-                      rules={[{ required: true }]}
+                      tooltip={t('modelService.ExpectedStatusTooltip')}
                       style={{ flex: 1, minWidth: 160 }}
+                      rules={[{ required: true }]}
                     >
                       <InputNumber
                         min={100}
```
ironAiken2: Same exclusive-bound issue as `port`: `expected_status_code` is declared with `gt=100` (**exclusive**) in both `PresetModelHealthCheckInput` and the strict `ModelHealthCheck`, so `100` passes the UI check but is rejected by the backend.

```suggestion
                        min={101}
```

ironAiken2 (changes requested):
Form-level numeric validation does not match the backend pydantic constraints — creating a preset with `port = 1` (and `expectedStatusCode = 100`) passes the UI but fails on the backend with a generic internal error. Please mirror the backend bounds (`gt=1` → min 2, `gt=100` → min 101) in the `InputNumber` props, and audit the remaining numeric fields in this PR the same way. See inline comments.

react/src/components/AdminDeploymentPresetModelConfigItem.tsx
agatha197: Fixed in 91fb89af8. Set the service `port` `InputNumber` `min` to **2** here (it was `1`), and the same for the `commandPort` field in `DeploymentAddRevisionModal.tsx` (~line 1722), matching the backend `gt=1` exclusive bound. I also audited the remaining numeric fields against the backend pydantic constraints: `initialDelay` is `ge=0` (UI `min={0}` ✓), and `interval` / `maxRetries` / `maxWaitTime` have no server-side lower bound, so their UI mins (1 / 1 / 1, and 0 for add-revision maxRetries) are at or stricter than the backend — no change needed.

react/src/components/AdminDeploymentPresetModelConfigItem.tsx
agatha197: Fixed in 91fb89af8. Set the `expectedStatusCode` `InputNumber` `min` to **101** (was `100`), matching the backend `expected_status_code` `gt=100` exclusive bound.

agatha197:
Addressed the latest review feedback:

**@ironAiken2 — numeric bounds vs backend pydantic constraints** (91fb89af8)
- Service `port` `min`: `1` → `2` (backend `gt=1`, exclusive) in both the preset model-config form and the add-revision `commandPort` field.
- Health-check `expectedStatusCode` `min`: `100` → `101` (backend `gt=100`, exclusive).
- Audited the rest: `initialDelay` is `ge=0` (UI `min={0}` already correct); `interval` / `maxRetries` / `maxWaitTime` have no server-side lower bound, so their UI mins are at or stricter than the backend — left as-is. Inline threads resolved.

**@nowgnuesLee — clean up useless comments**
- Redundant comments in the preset forms were removed in cae831e1c.

Also, per a follow-up request, the required `mem` resource slot now defaults to **GiB** instead of MiB (b6371aaca), matching the `defaultUnit="g"` convention used by every other memory field in the app.

Verification: Relay / Lint / Format / TypeScript all pass for the changed files (the unrelated `FluentEmojiIcon.tsx` tsc error is a local worktree dependency-link artifact, not from this PR).

ironAiken2 (approved):
LGTM

PR #7751, opened 2026-06-05
feat(FR-3055): encode current project and scope into the URL routing scheme

yomybaby:
Resolves #7750 (FR-3055)

## Summary

Encode the current project and the routing scope into the URL path, moving "current project" out of ambient global state.

- Scope-aware routes: `/project/:projectName/<feature>` (user), `/project/:projectName/admin/<feature>` (project-admin), `/admin/<feature>` (super/domain-admin)
- New primitives: `buildPath()` + menuKey↔scope map (`react/src/helper/pathBuilder.ts`); `useRouteScope` / `useCurrentMenuKey` / `useActiveProjectName` / `useProjectPath` (`react/src/hooks/useRouteScope.ts`)
- `ProjectScopeLayout` keeps the URL `:projectName` and the existing `currentProjectAtom` in sync (reuses `useCurrentProject`, no parallel store); `AdminScopeLayout` is the seam for global admin
- Replace first-path-segment assumptions (`location.pathname.split('/')[1]`, `startsWith('/admin-...')`) with route `handle.scope` / `handle.menuKey` (via `useMatches`) across menu, sider, page-access guard, help button, plugin loader, deployment config
- `react/src/legacyRedirects.tsx`: Class A (static `/admin-*` → `/admin/*`) + Class B (runtime project-scoped) redirect shims so old URLs canonicalize to the new ones — both old and new URLs resolve (regression-safe)
- Migrate ~40 imperative navigations + all menu links to `buildPath`
- Header project switch swaps only the `:projectName` segment, preserving the rest of the path so in-progress UI survives (e.g. the session launcher form stays mounted)

## Verification

- `bash scripts/verify.sh`: Lint / Format / TypeScript / Vite warmup **PASS**; Relay compiles clean with **0 `__generated__` changes** (the verify.sh Relay step only fails on a sandbox watchman-priority limitation, not on the code)
- `pnpm --prefix react test`: **924 tests pass** (44 files, incl. new `pathBuilder` / `useRouteScope` / `getPathFromMenuKey` specs)
- Dev-server smoke (Playwright): `/`, `/session`, `/project/default/session`, `/admin/session`, `/project/default/admin/users` all render with no uncaught errors

## Test plan

- [ ] Login → general-menu URLs are `/project/<name>/...`; switching project keeps the same page (session launcher form preserved)
- [ ] Old URLs (`/session`, `/admin-deployments`, `/project-admin-session`, `/credential`, `/summary`) redirect to the new canonical paths
[...]

PR #7755, opened 2026-06-07
fix(FR-3057): scope-aware 404, help-doc anchors, and invalid-project fallback

### LINKED ISSUE #7754: Fix cross-scope 404, help-doc anchors, and invalid-project fallback in scope-aware routing
Follow-up to FR-3055 (scope-aware URL routing). Browser-walking every route surfaced three navigation regressions/gaps introduced by (or exposed in) the new `/admin/<feature>` and `/project/:name/<feature>` namespaces.

## Why This Work is Needed

After encoding scope into the URL, a few navigation edge cases behave worse than before: cross-scope typo URLs render a blank page instead of 404, the help button links to the docs homepage for several pages because its lookup table still uses pre-scope keys, and deep-linking an inaccessible project drops the requested feature/scope.

## Expected Outcomes

- Cross-scope / unknown-feature URLs under the scope namespaces (e.g. /admin/start, /project/x/agent) render Page404 instead of an empty content area; all valid routes still render.
- The in-app help button resolves to the correct docs anchor on Deployments and the admin/project-admin pages (scope-aware menu keys added to the lookup table).
- An invalid/inaccessible :projectName redirect preserves the requested scope + feature on the fallback project when it maps to a real route (e.g. /project/<invalid>/admin/data -> /project/<fallback>/admin/data), collapsing to the session page only otherwise.
- Verified by walking the full route set in the browser against a live cluster.

JIRA Issue: FR-3057

yomybaby:
Resolves #7754 (FR-3057)

Stacked on #7751 (FR-3055).

## Summary

Follow-up fixes for navigation regressions found by browser-walking every route after the scope-aware routing change (#7751):

- **Cross-scope URLs now 404 instead of rendering blank** (`react/src/hooks/useWebUIMenuItems.tsx`): under the `/admin/<feature>` and `/project/:name/<feature>` namespaces, a feature key valid in a *different* scope (e.g. `/admin/start`, `/project/x/agent`) matched no route and rendered an empty `<Outlet/>`. `isCurrentPageNotFound` now validates the scope+feature via `scopeFeatureToMenuKey()` and returns 404. Valid routes are unaffected.
- **Help button resolves the correct docs anchor** (`react/src/components/WEBUIHelpButton.tsx`): the lookup table was keyed by pre-scope first-segments, so `deployments` and the admin/project-admin menu keys fell back to the docs homepage. Added the scope-aware menu keys mapping to the same anchors as their legacy keys.
- **Invalid / inaccessible `:projectName` shows an explicit "not found / no access" status instead of silently switching projects** (`react/src/components/MainLayout/ProjectScopeLayout.tsx`, `WebUIHeader.tsx`): per review feedback — a project name is just a name (it may not exist or be access-restricted), so redirecting to an arbitrary "current/first" project is misleading. The layout now renders an antd `<Result status="404">` ("Project '<name>' was not found or you don't have access to it") with a "Go to <your project>" action; and the header project selector shows **no selection (placeholder)** for an invalid URL project instead of misleadingly showing the last valid project. New i18n keys translated into all 21 languages.

## Verification (live browser walk against a cluster, super-admin)

- Walked the full route set (user / admin / project-admin), all 16 old-URL redirect shims, and transitions (menu clicks, project switch on list + form, back button) — all correct; valid routes render.
- After this PR (browser-verified): cross-scope typo URLs (`/admin/start` etc.) render Page404; `/project/default/deployments` help → `model_serving/model_serving.html`; `/project/<invalid>/session` renders the "project not found / no access" notice (no redirect) with a "Go to <project>" button.
- `pnpm --prefix react test`: 917 tests pass; tsc / lint / format clean.

## Route-error states redesign (v3, minimal)

All route-level error states now share one minimal composition (`RouteErrorContent`): a monospace **path pill** with the broken segment highlighted (warm chip + wavy accent underline), a heading, a secondary hint, and a CTA. No decorative background/illustration.

- **Invalid project** (`/project/<bad>/...`): pill marks the project segment; hint points at the header selector; CTA goes to the **alphabetically-first accessible project**.
- **Unknown feature/path** (`/project/default/model-store2`, any 404): `Page404` renders the same language, marking the unresolved segment, keeping its existing copy.
- **No accessible projects**: title + administrator hint (new i18n key `AskAdminForProjectAccess`, 21 languages).
- Pill is `aria-hidden` (heading carries semantics); the broken-segment ink is contrast-tuned per theme (`#B25400` light / `#F2A045` dark).

| Invalid project — light | Invalid project — dark | Unknown page — light |
|---|---|---|
| invalid light | invalid dark | feature 404 |

### DIFF
--- a/react/src/components/MainLayout/ProjectScopeLayout.tsx
+++ b/react/src/components/MainLayout/ProjectScopeLayout.tsx
@@ -11,6 +11,7 @@ import {
 import { getRouteScopeAndKey } from '../../hooks/useRouteScope';
 import ProjectScopeErrorState from './ProjectScopeErrorState';
 import { Button } from 'antd';
+import { ArrowRightIcon } from 'lucide-react';
 import React, { useEffect, useEffectEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Outlet, useLocation, useParams } from 'react-router-dom';
@@ -108,6 +109,9 @@ const ProjectScopeLayout: React.FC = () => {
         extra={
           <Button
             type="primary"
+            size="large"
+            icon={<ArrowRightIcon size="1em" />}
+            iconPosition="end"
             onClick={() =>
               webuiNavigate(buildPath('project', 'session', ownProject))
             }
--- a/react/src/components/MainLayout/ProjectScopeLayout.tsx
+++ b/react/src/components/MainLayout/ProjectScopeLayout.tsx
@@ -8,11 +8,12 @@ import {
   useCurrentProjectValue,
   useSetCurrentProject,
 } from '../../hooks/useCurrentProject';
-import { Button, Result } from 'antd';
-import { BAIFlex } from 'backend.ai-ui';
+import { getRouteScopeAndKey } from '../../hooks/useRouteScope';
+import ProjectScopeErrorState from './ProjectScopeErrorState';
+import { Button } from 'antd';
 import React, { useEffect, useEffectEvent } from 'react';
 import { useTranslation } from 'react-i18next';
-import { Outlet, useParams } from 'react-router-dom';
+import { Outlet, useLocation, useParams } from 'react-router-dom';
 
 interface BackendAIClientGroups {
   groups?: string[];
@@ -51,6 +52,7 @@ const ProjectScopeLayout: React.FC = () => {
   const { t } = useTranslation();
   const baiClient = useSuspendedBackendaiClient();
   const { projectName: rawProjectName } = useParams<{ projectName: string }>();
+  const location = useLocation();
   const currentProject = useCurrentProjectValue();
   const setCurrentProject = useSetCurrentProject();
   const webuiNavigate = useWebUINavigate();
@@ -85,19 +87,7 @@ const ProjectScopeLayout: React.FC = () => {
     // No groups at all: the user belongs to no project. Render a terminal
     // "no accessible projects" status.
     if (groups.length === 0) {
-      return (
-        <BAIFlex
-          direction="column"
-          align="center"
-          justify="center"
-          style={{ width: '100%', height: '100%' }}
-        >
-          <Result
-            status="warning"
-            title={t('projectSelect.NoAccessibleProjects')}
-          />
-        </BAIFlex>
-      );
+      return <ProjectScopeErrorState variant="no-projects" />;
     }
 
     // Invalid / inaccessible project name while the user DOES have other
@@ -111,30 +101,21 @@ const ProjectScopeLayout: React.FC = () => {
         ? currentProject.name
         : groups[0];
     return (
-      <BAIFlex
-        direction="column"
-        align="center"
-        justify="center"
-        style={{ width: '100%', height: '100%' }}
-      >
-        <Result
-          status="404"
-          title={t('projectSelect.ProjectNotFoundOrNoAccess', {
-            project: projectName,
-          })}
-          subTitle={t('projectSelect.SwitchToAccessibleProject')}
-          extra={
-            <Button
-              type="primary"
-              onClick={() =>
-                webuiNavigate(buildPath('project', 'session', ownProject))
-              }
-            >
-              {t('projectSelect.GoToProject', { project: ownProject })}
-            </Button>
-          }
-        />
-      </BAIFlex>
+      <ProjectScopeErrorState
+        variant="not-found"
+        projectName={projectName}
+        featureKey={getRouteScopeAndKey(location.pathname).featureKey}
+        extra={
+          <Button
+            type="primary"
+            onClick={() =>
+              webuiNavigate(buildPath('project', 'session', ownProject))
+            }
+          >
+            {t('projectSelect.GoToProject', { project: ownProject })}
+          </Button>
+        }
+      />
     );
   }
 
--- a/react/src/components/MainLayout/ProjectScopeLayout.tsx
+++ b/react/src/components/MainLayout/ProjectScopeLayout.tsx
@@ -3,13 +3,13 @@
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
 import { buildPath } from '../../helper/pathBuilder';
-import { useSuspendedBackendaiClient } from '../../hooks';
+import { useSuspendedBackendaiClient, useWebUINavigate } from '../../hooks';
 import {
   useCurrentProjectValue,
   useSetCurrentProject,
 } from '../../hooks/useCurrentProject';
-import WebUINavigate from '../WebUINavigate';
-import { BAIAlert, BAIFlex } from 'backend.ai-ui';
+import { Button, Result } from 'antd';
+import { BAIFlex } from 'backend.ai-ui';
 import React, { useEffect, useEffectEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Outlet, useParams } from 'react-router-dom';
@@ -30,12 +30,14 @@ interface BackendAIClientGroups {
  *   - membership = `baiClient.groups.includes(name)`
  *
  * If the URL project name is invalid (not a member, or no resolvable id), the
- * layout redirects (replace) to a guaranteed-valid fallback project's session
- * page rather than rendering with an undefined project id:
- *   - fallback = current atom project name if the user is still a member of it,
- *     otherwise `groups[0]`.
- *   - if the user has NO groups at all, we cannot synthesize a `/project//...`
- *     path, so we render the existing "no accessible projects" guidance.
+ * layout does NOT silently switch to another project — the name is just a name
+ * and may legitimately be missing or access-restricted (stale bookmark, shared
+ * link to a project the user cannot access, renamed/deleted project). Instead it
+ * renders an explicit "not found / no access" guidance:
+ *   - if the user has NO groups at all, the "no accessible projects" guidance.
+ *   - otherwise a "project not found or no access" notice, keeping the header
+ *     project selector available and offering a button that navigates to one of
+ *     the user's own projects on an explicit click.
  *
  * Sync (effect-only, idempotent):
  *   - `setCurrentProject({ projectName, projectId })` runs in an effect keyed on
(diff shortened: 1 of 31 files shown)


yomybaby:
* **#8423** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8423?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8409** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8409?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8408** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8408?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8208** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8208?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7755** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7755?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7755?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7751** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7751?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7755).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/hooks/useWebUIMenuItems.tsx
```diff
@@ -918,6 +926,28 @@ export const useWebUIMenuItems = (props?: UseWebUIMenuItemsProps) => {
...
+      getRouteScopeAndKey(location.pathname);
+    const hasScopePrefix =
+      location.pathname.startsWith('/admin/') ||
+      location.pathname.startsWith('/project/');
+    if (
+      hasScopePrefix &&
+      pathFeatureKey &&
+      !scopeFeatureToMenuKey(pathScope, pathFeatureKey) &&
+      !matchesDynamicRoute &&
+      !plugins?.page?.some((page) => page?.url === pathFeatureKey)
```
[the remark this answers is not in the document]
yomybaby: Superseded by #8208 (FR-3279): `isCurrentPageNotFound` (including this plugin exemption) is deleted there — 404 is now owned by router catch-all routes. Under the new architecture a scoped plugin URL like `/admin/<pluginUrl>` does **not** produce a blank page: the scoped catch-all renders `UnknownRoutePage`, which returns `null` for plugin featureKeys, while `PluginLoader` activates the plugin element via the `useCurrentMenuKey()` featureKey fallback (the Lit element is a sibling of the router `<Outlet/>` in MainLayout's content area, so it renders regardless of route match). Resolving as addressed upstack.

react/src/hooks/useWebUIMenuItems.tsx
```diff
@@ -918,6 +926,28 @@ export const useWebUIMenuItems = (props?: UseWebUIMenuItemsProps) => {
...
+    ) {
+      return true;
```
ironAiken2, line 948: When an invalid path is entered under the admin namespace (e.g. `/admin/start`), the 404 page renders with the **admin theme** applied — `AutoAdminPrimaryColorProvider` keys purely on the route scope, so this also happens for users who have no admin role at all.

Since the page is a 404 (not an actual admin page), it would look better to render it with the **default theme** instead of the admin one.

react/src/pages/Page404.tsx
```diff
@@ -25,44 +33,36 @@ const Page404 = () => {
...
+  // that is the part the router could not resolve. Long paths are capped
+  // with a leading ellipsis so detail-page URLs stay readable.
+  const rawSegments = location.pathname.split('/').filter(Boolean);
+  const capped =
+    rawSegments.length > MAX_SEGMENTS
+      ? ['…', ...rawSegments.slice(-(MAX_SEGMENTS - 1))]
+      : rawSegments;
+  const segments: RouteErrorSegment[] = capped.map((text, i) => ({
+    text,
+    broken: i === capped.length - 1,
```
ironAiken2, line 46: The pill always marks the **last** segment as broken, but that is not necessarily the segment the router failed on. For example, entering `/admi/default/start` (a typo of `/admin`, and not `/project` either) marks `start` as the broken segment — while the actual unresolvable part is the first segment `admi`.

It would be better to first check the leading segment: if it is not a valid scope prefix (`project` / `admin`) or a known top-level route, mark **that** segment as broken instead of unconditionally highlighting the last one.

react/src/components/MainLayout/WebUIHeader.tsx
```diff
@@ -62,6 +67,14 @@ const WebUIHeader: React.FC<WebUIHeaderProps> = ({ onClickMenuIcon }) => {
   const location = useLocation();
   const routeScope = useRouteScope();
   const currentMenuKey = useCurrentMenuKey();
+  // When the URL carries an invalid/inaccessible `:projectName`, the atom keeps
+  // the last valid project, which would make the header selector look like that
+  // project is selected. Detect this and show the selector unselected instead.
+  const activeProjectName = useActiveProjectName();
+  const accessibleProjectNames =
+    (baiClient as unknown as { groups?: string[] })?.groups ?? [];
+  const isUrlProjectInvalid =
```
ironAiken2, line 76: This handles the invalid `:projectName` case under `/project/...`, but in the 404 case from the other comment (e.g. `/admi/default/start`, where the first segment is not a valid scope prefix) the header project selector still shows the last valid project. In that situation the selector should be empty (placeholder) as well — the URL does not correspond to any project context.

react/src/components/MainLayout/ProjectScopeLayout.tsx
```diff
@@ -79,32 +85,39 @@ const ProjectScopeLayout: React.FC = () => {
   }, [projectName]);
 
   if (!isValid) {
```
ironAiken2, line 87: For a URL like `/project/defaut/303`, the project `defaut` does not exist — but the error shown is the generic "page not found" 404 keyed on `303` (the scope-aware 404 check appears to win before this layout gets to render its invalid-project state).

When both the `:projectName` and the feature segment are invalid, the **project error should take precedence** — the project context comes first in the URL hierarchy. And same as the other comment: the header project selector should be empty in this state as well.

ironAiken2 (changes requested):
Requesting changes for four UX issues found while walking the error/404 paths:

1. Invalid paths under `/admin/*` render the 404 page with the admin theme — even for users without any admin role. The 404 should use the default theme.
2. The 404 path pill always marks the last segment as broken; for `/admi/default/start` the actual failure is the first segment (`admi`).
3. In that same 404 state, the header project selector should be empty (placeholder) — the URL has no project context.
4. `/project/defaut/303` shows the generic 404 keyed on `303` even though the project `defaut` itself does not exist — the invalid-project error should take precedence, with the selector empty there too.

react/src/hooks/useWebUIMenuItems.tsx
```diff
@@ -918,6 +926,28 @@ export const useWebUIMenuItems = (props?: UseWebUIMenuItemsProps) => {
...
+    }
+
     // Check if path is in valid paths set
     if (allValidPaths.has(currentPathKey)) return false;
```
ironAiken2, line 952: Navigating to the Settings or Logs page from the header user dropdown (`/usersettings?tab=general`, `/usersettings?tab=logs`) shows the "page not found" error, even though the route itself exists and renders.

It looks like `/usersettings` has no `menuKey` in its route handle, so `useCurrentMenuKey` falls back to the first path segment (`usersettings`) — and that key is in none of the valid-path sources: it is not a menu item, and `ROUTER_STATIC_PATHS` only collects the project-scope children, admin-scope children, and legacy redirect shims, which excludes the global no-prefix routes (`/usersettings`, `/logs`, …) declared after them. So `isCurrentPageNotFound` returns true for a perfectly valid page.

ironAiken2 (changes requested):
One more issue: the Settings / Logs pages reached from the header user dropdown (`/usersettings?tab=general|logs`) are incorrectly flagged as 404 — details in the inline comment.

react/src/components/MainLayout/ProjectScopeLayout.tsx
yomybaby: Verified fixed on the current stack top (with #8423 applied): `/project/defaut/303` now renders the invalid-project state ("Project 'defaut' was not found or you don't have access to it") with the header selector cleared — the project error takes precedence over the feature-segment 404, as requested.

react/src/hooks/useWebUIMenuItems.tsx
yomybaby: Verified fixed on the current stack top: `/usersettings?tab=general` renders the Settings & Logs page normally (breadcrumb + General/Logs tabs), no 404. The `/usersettings` route now matches inside the scoped route tree, so `useCurrentMenuKey`'s fallback no longer misclassifies it.

react/src/hooks/useWebUIMenuItems.tsx
yomybaby: Decision (agreed with @yomybaby): keeping the admin theming as-is for this stack — it keys on the URL scope and is not critical for the routing redesign. May be revisited in a follow-up if the 404-with-admin-theme look bothers users in practice.

react/src/pages/Page404.tsx
yomybaby: Addressed in #8423 (second commit) by removing the broken marking on the 404 pill entirely: a 404 cannot know which segment failed (`/admi/…` breaks on the first segment, not the last), so any single marking is a guess. All segments now render neutral; `ProjectScopeErrorState` — which does know the failing segment — keeps its accurate `broken` marking. (Chosen over marking the whole path, which reads as noise.)

react/src/components/MainLayout/WebUIHeader.tsx
yomybaby: Fixed in #8423 (second commit): the header selector now shows its placeholder on projectless router-owned 404s — i.e. the `notFound` catch-all matched without a `:projectName` param (e.g. `/admi/default/start`). Scoped 404s under a valid project URL (e.g. `/project/default/303`) keep their project context, since the URL still names a real project. Runtime-verified both cases.

agatha197 (approved):
LGTM

ironAiken2 (approved):

PR #7767, opened 2026-06-08
feat(FR-3065): persist preset form inputs to URL with secret-stripping

agatha197:
Resolves #7766 (FR-3065)

## Summary

Broadens the deployment-preset Create form's URL state-sync so a shared link restores more of the form, while keeping secret-prone values out of the URL.

Previously `syncFormToURL` excluded the whole `environ` and `modelDefinition` sections (`_.omit(..., ['environ','modelDefinition'])`), so a shared/reloaded link lost the env-var layout and the entire model definition. This replaces the wholesale omit with a single `sanitizeFormValuesForURL` helper that persists the structure/config but strips secrets:

- `environ` → variable **names** kept, **values blanked**.
- `modelDefinition` per model → kept (name, modelPath, port, shell, **startCommand**, health-check config, metadata, enable flags); **`preStartActions[].args` blanked** (action kept).
- Everything else (name, ids, resources, cluster, replica, etc.) syncs as before.

`service.startCommand` is the model launch command (service config), not a credential, so it is persisted — restoring it lets a shared link rebuild the full service definition. Only free-text that can hold credentials/tokens (`environ[].value`, `preStartActions[].args`) is blanked.

Restore is unchanged (`_.merge(initialValues, formValuesFromURL)`), so the newly-synced sections repopulate on load. Create-mode-only + 500ms debounce + `history: 'replace'` behavior preserved.

Top-level `startupCommand` / `bootstrapScript` retention is intentionally left as-is (pre-existing behavior); tightening those is a separate concern.

## Test plan

- [x] `bash scripts/verify.sh` → Relay / Lint / Format / TypeScript all pass.
- [ ] Create preset: add env vars + a model definition → reload the page → names/config (incl. startCommand) restore, env-var **values** and model **preStartActions args** come back empty.
- [ ] Confirm no secret values appear in the address bar while editing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7775, opened 2026-06-08
feat(FR-3068): add explicit Enable Health Check toggle in add-revision custom mode

### LINKED ISSUE #7773: Add explicit Enable Health Check toggle in the add-revision custom mode
Follow-up to FR-3056. In the deployment **Add Revision** modal (custom command mode, `DeploymentAddRevisionModal.tsx`), the health-check section is currently inferred purely from whether any health-check field is filled (any filled → `enable = true`; all empty → `enable = false`). There is no explicit affordance telling the user that health check is on/off, unlike the Deployment Revision Preset form which has an **Enable Health Check** checkbox.

## Why

- Review feedback (Frontend sync): the add-revision flow should expose the same explicit enable affordance as the preset form, so the user clearly opts in/out rather than relying on implicit field-presence detection.
- Backend (PR lablup/backend.ai#11863, BA-6242, 26.4.4) treats `enable` as the single meaningful flag; every other health-check field is optional and server-seeded. The UI toggle should map directly to `enable`.

## Expected outcomes

- Add an **Enable Health Check** checkbox to the add-revision custom-mode form, consistent with `AdminDeploymentPresetModelConfigItem.tsx`.
- When checked → submit `healthCheck.enable = true` (fields optional, server seeds defaults); when unchecked → `healthCheck.enable = false` with fields omitted.
- Prefill the checkbox from the source revision's `enable` on 26.4.4; preserve legacy null-based behavior on managers <= 26.4.3 (gate via `isManagerVersionCompatibleWith('26.4.4')` + `@since`).
- Health-check fields remain optional (not `required`), matching FR-3056.

## References

- Parent work: FR-3056 (PR #7753)
- Backend PR: lablup/backend.ai#11863 (BA-6242, 26.4.4)

JIRA Issue: FR-3068

agatha197:
Resolves #7773 (FR-3068)

> Stacked on #7753 (FR-3056).

## Summary

Follow-up to FR-3056. In the deployment **Add Revision** modal (`DeploymentAddRevisionModal.tsx`), health-check enable was previously inferred purely from whether any health-check field was filled. This adds an explicit **Enable Health Check** checkbox, consistent with the Deployment Revision Preset form (`AdminDeploymentPresetModelConfigItem.tsx`).

- New `commandEnableHealthCheck` form field renders a checkbox; the health-check fields (path, interval, max retries, max wait time, expected status code, initial delay) are revealed only when it is checked.
- The checkbox is shown for **every runtime variant and definition mode**, not just custom + command — matching the preset form, where a health check is configurable regardless of runtime. For non-command modes a minimal `modelDefinition` override carrying only the health check is sent.
- **Required when enabled**: whenever the checkbox is on, all health-check fields are required in the UI, on every manager version — mirroring the preset form.
- **Submit shape is version-aware**: checked → `healthCheck.enable = true` with the configured fields; unchecked → `healthCheck.enable = false` on 26.4.4+ managers, or `null` on managers ≤ 26.4.3 (preserving the legacy shape). Gated via `baiClient.supports('model-health-check-enable')`.
- **Prefill**: the checkbox derives from the source revision's `enable` (26.4.4) / health-check object presence (≤ 26.4.3).
- **Default off**: a new (non-prefilled) revision starts with the toggle off — an explicit opt-in, matching the preset form.
- Numeric bounds match the backend exclusive constraints (`port ≥ 2` for `gt=1`, `expectedStatusCode ≥ 101` for `gt=100`), same as the preset-form fix in #7753.
- Added `modelService.EnableHealthCheck` i18n key across all 22 languages.

## Test plan

- [x] `bash scripts/verify.sh` → Relay / Lint / Format / TypeScript all pass.
- [ ] On a 26.4.4 manager: add a revision with the checkbox on → all fields required, submits `enable: true` + fields; with it off → `enable: false`.
- [ ] Edit/prefill: a source revision with the health check enabled/disabled pre-checks the box accordingly.
- [ ] On a ≤ 26.4.3 manager: off → `healthCheck: null`; on → legacy object (fields required), no `enable` key.
- [ ] Toggle is visible for non-custom runtimes (e.g. vLLM) and custom + file mode, not only custom + command.


agatha197:
* **#7939** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7939?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7937** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7937?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7903** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7903?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7776** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7776?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7775** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7775?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7775?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7753** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7753?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#7767](https://github.com/lablup/backend.ai-webui/pull/7767) <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7767?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7775).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

ironAiken2 (commented):
_Superseded by the request-changes review below._

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1711,76 +1705,165 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
-                              />
-                            </Form.Item>
-                          </BAIFlex>
+                          <Form.Item
+                            name="commandPort"
+                            label={t('modelService.Port')}
+                            tooltip={t('modelService.PortTooltip')}
+                          >
+                            <InputNumber
+                              min={1}
```
ironAiken2: `ModelServiceConfig.port` is `gt=1` (exclusive) on the backend, so `1` passes the form but fails server-side. Same as the preset form fix in #7753 (a5303c9f3):
```suggestion
                              // Lenient draft resolves into strict
                              // `ModelServiceConfig.port` (`gt=1`, exclusive),
                              // so the lowest accepted port is 2.
                              min={2}
```

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1711,76 +1705,165 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
+                                        'adminDeploymentPreset.modelDef.HealthCheckExpectedStatus',
+                                      )}
+                                      tooltip={t(
+                                        'modelService.ExpectedStatusTooltip',
+                                      )}
+                                      rules={[{ required: true }]}
+                                      style={{ flex: 1, minWidth: 160 }}
+                                    >
+                                      <InputNumber
+                                        min={100}
```
ironAiken2: Backend `expected_status_code` is `gt=100` (exclusive), so the lowest accepted code is 101 — `100` would be rejected server-side. Same bound as applied to the preset form in #7753:
```suggestion
                                        // Backend `expected_status_code` is
                                        // `gt=100` (exclusive), so the lowest
                                        // accepted code is 101.
                                        min={101}
```

ironAiken2 (changes requested):
Two things:

**1. Numeric bounds (inline suggestions below)** — same backend-bounds issue that a5303c9f3 fixed on #7753: the backend pydantic constraints are exclusive (`gt`), so `port = 1` and `expectedStatusCode = 100` pass the form but get rejected server-side with a generic error. The remaining numeric fields (interval/maxRetries/maxWaitTime/initialDelay) are already at or stricter than the backend constraints, so they're fine as-is.

**2. Health check is gated behind custom runtime + command mode** — in the deployment preset form (`AdminDeploymentPresetModelConfigItem`), the health check is configurable regardless of runtime. In this add-revision flow, the Enable Health Check toggle only renders under `runtimeVariantId === 'custom'` && `customDefinitionMode === 'command'`, so for non-custom runtimes (e.g. vLLM) a revision cannot configure a health check while a preset can. Please check whether this is intentional — if the backend accepts `health_check` for any runtime, the toggle should be exposed outside custom/command mode as well (config-file mode is arguably fine since the model-definition file carries it).

react/src/components/DeploymentAddRevisionModal.tsx
agatha197: Fixed in latest commit: `min` updated to `2` (matching the backend `gt=1` exclusive constraint on `ModelServiceConfig.port`). Same bound applied as in #7753 a5303c9f3.

react/src/components/DeploymentAddRevisionModal.tsx
agatha197: Fixed in latest commit: `min` updated to `101` (matching the backend `gt=100` exclusive constraint on `expected_status_code`). Same bound applied as in #7753.

agatha197:
Addressed feedback from @ironAiken2:

**1. Numeric bounds** — Fixed `port min→2` and `expectedStatusCode min→101` to match the backend pydantic `gt=1` and `gt=100` exclusive constraints (same fix as a5303c9f3 in #7753).

**2. Health check visibility** — The Enable Health Check toggle is now shown for all runtimes (not only custom+command). The toggle and its detail fields are placed after the runtime-config section and rendered for every variant except custom+file mode (where the model-definition YAML carries the health check config). For non-custom runtimes, when the toggle is enabled, a minimal `modelDefinition` override containing only the `service.healthCheck` block is sent alongside the request so the backend can apply it over the runtime-generated defaults.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1719,106 +1721,192 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
+                      getHc('commandEnableHealthCheck') ? (
+                        <BAIFlex direction="column" align="stretch" gap="xs">
+                          <Form.Item
+                            name="commandHealthCheck"
+                            label={t(
+                              'adminDeploymentPreset.modelDef.HealthCheckPath',
+                            )}
+                            tooltip={t('modelService.HealthCheckTooltip')}
+                            rules={[{ required: true }]}
+                          >
```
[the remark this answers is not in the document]
agatha197: Intentional design decision: when the toggle is on, all health-check fields are required in the UI on every manager version, mirroring the preset form (`AdminDeploymentPresetModelConfigItem`), which requires them unconditionally. We opted for the explicit, version-independent UX over the 26.4.4 server-seed-defaults path. The submit-time comment and the PR description have been updated accordingly — the fields are no longer described as optional on 26.4.4+.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -1539,18 +1550,8 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
...
-            commandModelMount: '/models',
-            // 60s was too short for large models to finish loading before the
-            // first health check. Default to 1800s as a short-term fix until a
-            // shared backend/frontend default (feature flag) lands. See FR-3005.
-            commandInitialDelay: 1800,
-            commandMaxRetries: 10,
-            commandInterval: 10,
-            commandMaxWaitTime: 15,
+            commandEnableHealthCheck: false,
             environ: [],
```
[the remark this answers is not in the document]
agatha197: Resolved by updating the PR description: default-off is intentional. An explicit opt-in toggle that requires all fields when on reads more naturally as off-by-default, and it matches the preset form. Prefill is unaffected — the checkbox still derives from the source revision's health-check state, so editing an existing revision pre-checks the box correctly.

agatha197:
@ironAiken2 Thanks — both points addressed:

1. **Numeric bounds**: `port` min is now `2` and `expectedStatusCode` min is `101` (backend exclusive `gt=1` / `gt=100`), matching the #7753 fix (commit `bb907711b`). Inline threads resolved.
2. **Health-check gating**: the Enable Health Check toggle now renders for every runtime variant and definition mode, not only custom + command (commit `552687b9a`). For non-command modes a minimal `modelDefinition` override carrying only the health check is sent.

Also aligned the design with the preset form per Copilot's threads: all health-check fields are required when the toggle is on (every manager version), and the toggle defaults off. Re-requesting your review.

ironAiken2 (changes requested):
**Requesting changes — the runtime-variant presets load, but there's no UI to enter the runtime variables.**

While testing this branch in the add-revision modal, the runtime-variant presets query returns data successfully:

```graphql
query useRuntimeParameterSchemaPresetsQuery($filter: RuntimeVariantPresetFilter, $orderBy: [RuntimeVariantPresetOrderBy!]) {
  runtimeVariantPresetsResult: runtimeVariantPresets(filter: $filter, orderBy: $orderBy, first: 100) {
    edges { node { name displayName category targetSpec { ... } uiOption { ... } id } }
  }
}
```

But the runtime variable input controls never render — there is no field in the form to actually enter the runtime parameter values.

These presets feed `useRuntimeParameterSchema` → `RuntimeParameterFormSection` (`ParameterControl` renders the per-parameter slider/number/select/checkbox/text inputs). Since the query returns edges but the section shows nothing, it looks like the schema built from `runtimeVariantPresets` ends up empty (or the section isn't mounted on this path), so `RuntimeParameterFormSection` falls through its `if (!groups) return null` guard.

Could you check why the loaded presets aren't surfaced as input controls? As it stands, users can't set runtime variables in the add-revision flow.

agatha197:
I'll handle it in #7903

> **Requesting changes — the runtime-variant presets load, but there's no UI to enter the runtime variables.**
>
> While testing this branch in the add-revision modal, the runtime-variant presets query returns data successfully:
>
> ```graphql
> query useRuntimeParameterSchemaPresetsQuery($filter: RuntimeVariantPresetFilter, $orderBy: [RuntimeVariantPresetOrderBy!]) {
>   runtimeVariantPresetsResult: runtimeVariantPresets(filter: $filter, orderBy: $orderBy, first: 100) {
>     edges { node { name displayName category targetSpec { ... } uiOption { ... } id } }
>   }
> }
> ```
>
> But the runtime variable input controls never render — there is no field in the form to actually enter the runtime parameter values.
>
> These presets feed `useRuntimeParameterSchema` → `RuntimeParameterFormSection` (`ParameterControl` renders the per-parameter slider/number/select/checkbox/text inputs). Since the query returns edges but the section shows nothing, it looks like the schema built from `runtimeVariantPresets` ends up empty (or the section isn't mounted on this path), so `RuntimeParameterFormSection` falls through its `if (!groups) return null` guard.
>
> Could you check why the loaded presets aren't surfaced as input controls? As it stands, users can't set runtime variables in the add-revision flow.

​

ironAiken2 (approved):
LGTM

PR #7822, opened 2026-06-11
fix(FR-3091): stabilize chat TPS during streaming (live elapsed denominator + cancel stale token counts)

### LINKED ISSUE #7821: fix(FR-3091): Chat TPS display keeps rising after streaming ends due to stale async token counts
Resolves FR-3091

## Problem

The TPS (tokens per second) display in the chat UI continuously increases after streaming ends, instead of showing a stable rate value.

## Root Cause

`useTokenCount` (`react/src/hooks/useTokenizer.ts`) fires a new async token count computation on every render during streaming — one per incoming chunk. These async operations all run concurrently and are **not cancelled** when a newer computation starts.

When streaming ends, `endTime` becomes fixed. However, the queued async computations continue resolving one by one, each calling `setNum` with a progressively larger token count. Since `TPS = lastAssistantTokenCount / (endTime - startTime)`, with a fixed denominator but an increasing numerator, TPS keeps climbing until all queued computations complete.

## Fix

Add a cancellation flag (cleanup return) to the `useEffect` in `useTokenCount` so that when `input` changes, the previous async computation's result is discarded. Only the latest computation for the current input string updates state.

agatha197:
Resolves #7821 (FR-3091)

## Summary

The chat TPS indicator misbehaved in several ways. This PR fixes them and improves accuracy.

1. **TPS shot up during streaming and snapped down at the end** (the dominant, user-visible symptom: e.g. climbing past 1100 then dropping to ~115). Caused by a frozen elapsed-time denominator under the React Compiler.
2. **TPS kept creeping up *after* streaming ended.** Caused by stale, uncancelled async token-count computations resolving one by one.
3. **`think` (reasoning) content was excluded** from the token count.
4. **The numerator was only a gpt-tokenizer estimate**, not the model's actual output token count.

### (1) Frozen elapsed denominator — `ChatTokenCounter`

`TPS = tokenCount / elapsedSec`. `ChatTokenCounter` uses `'use memo'`, so the React Compiler memoizes the elapsed expression on its **tracked** deps (`startTime` / `endTime`) only. `Date.now()` read in render is an untracked impure call, so the denominator was computed once — right after `startTime` was set, when elapsed ≈ 0s — and reused. The growing numerator then drove the displayed TPS ever higher until `endTime` changed and the expression recomputed against real time.

**Fix:** keep `Date.now()` out of render. Since TPS only changes when the token count changes, sample the wall clock in an effect keyed on the token count (`measuredAt`), and divide by `endTime ?? measuredAt`. Numerator and denominator are then snapshots of the same instant, with no periodic re-renders. (An earlier revision used a 200 ms ticking state; replaced per review with this coherent token-count-driven sampling — see thread.)

### (2) Stale async token counts — `useTokenCount` (`useTokenizer.ts`)

`useTokenCount` launched a new async `encodeAsync()` per streaming chunk without cancelling prior ones; after `endTime` was fixed, queued results kept resolving and bumping the numerator. **Fix:** a `cancelled` flag in the `useEffect` cleanup so only the latest `input`'s result updates state. Regression-tested for out-of-order resolution.

### (3) Reasoning tokens — `ChatTokenCounter`

`<think>` output is extracted into `reasoning` parts; the counter now includes both `text` and `reasoning` parts in the TPS numerator and the total token count.

### (4) Model-reported usage tokens — `ChatCard` + `ChatTokenCounter`

On the client-side streaming path, the model's `outputTokens` (from the stream's finish event) is attached to the final message via `toUIMessageStreamResponse`'s `messageMetadata`. `ChatTokenCounter` prefers that exact count for the TPS numerator and **falls back to the gpt-tokenizer estimate** during streaming (usage unknown mid-stream) and on paths/models that omit usage — so behavior is never worse than before.

## TPS definition

Numerator = output tokens of the last assistant message (text + reasoning) — model-reported when available, else gpt-tokenizer estimate. Denominator = decode window `startTime` (first output token / status → `streaming`) to `endTime` (stream end), excluding prefill / TTFT, matching the vLLM / Ollama convention.

## Verification

- Manual, in-browser (dev server, streaming model): before → TPS climbed past 1100 then dropped to ~115; after → stable real rate throughout, frozen at the final value on completion.
- Unit tests: `ChatTokenCounter` (denominator sampled at token-count updates and holds steady while only time passes; reasoning tokens counted; usage `outputTokens` preferred with estimate fallback) and `useTokenizer` (out-of-order stale result discarded). All pass.
- Lint passes on the changed files.

### Test limitation (honest scope note)

The denominator unit test guards the **sampling semantics** (TPS holds while time passes, re-samples on token-count change) and keeps time out of render. It does **not** reproduce the React Compiler memoization freeze itself, since the compiler is not guaranteed to run identically under Vitest (jsdom) as in the production build. The compiler-specific freeze was validated **manually in the browser** (1100 → real-rate before/after), not by an automated test.

### DIFF
--- a/react/src/components/Chat/ChatTokenCounter.tsx
+++ b/react/src/components/Chat/ChatTokenCounter.tsx
@@ -8,7 +8,7 @@ import { Typography, Tag, Divider } from 'antd';
 import { BAIQuestionIconWithTooltip, BAIFlex } from 'backend.ai-ui';
 import { t } from 'i18next';
 import { map, last } from 'lodash-es';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 
 interface ChatTokenCounterProps {
   input: string;
@@ -29,7 +29,7 @@ const ChatTokenCounter: React.FC<ChatTokenCounterProps> = ({
   const inputTokenCount = useTokenCount(input);
   const allChatMessageString = map(messages, (message) =>
     message?.parts
-      ?.filter((part) => part.type === 'text')
+      ?.filter((part) => part.type === 'text' || part.type === 'reasoning')
       .map((part) => part.text)
       .join(''),
   ).join('');
@@ -39,16 +39,53 @@ const ChatTokenCounter: React.FC<ChatTokenCounterProps> = ({
   const lastAssistantMessageString =
     lastAssistantMessage?.role === 'assistant'
       ? lastAssistantMessage?.parts
-          ?.filter((part) => part.type === 'text')
+          ?.filter((part) => part.type === 'text' || part.type === 'reasoning')
           .map((part) => part.text)
           .join('') || ''
       : '';
 
-  const lastAssistantTokenCount = useTokenCount(lastAssistantMessageString);
+  const estimatedLastAssistantTokenCount = useTokenCount(
+    lastAssistantMessageString,
+  );
+  // Prefer the model-reported output token count (attached to the final
+  // message metadata in ChatCard) over the client-side gpt-tokenizer estimate.
+  // It is only present once streaming finishes and only on paths that report
+  // usage (client-side streamText); during streaming and on server paths that
+  // omit usage, fall back to the estimate.
+  const usageOutputTokens =
+    lastAssistantMessage?.role === 'assistant'
+      ? (
+          lastAssistantMessage?.metadata as
+            | { outputTokens?: number }
+            | undefined
+        )?.outputTokens
+      : undefined;
+  const lastAssistantTokenCount =
+    typeof usageOutputTokens === 'number' && usageOutputTokens > 0
+      ? usageOutputTokens
+      : estimatedLastAssistantTokenCount;
+
+  // The elapsed denominator needs a current timestamp, but reading `Date.now()`
+  // directly in render does not work: under the React Compiler ('use memo'),
+  // the elapsed expression is memoized on its tracked deps (startTime/endTime)
+  // only — `Date.now()` is an untracked impure call, so the denominator would
+  // freeze at ~0s from when startTime was first set, inflating TPS until
+  // streaming ends. TPS only changes when the token count changes, so rather
+  // than a ticking timer we sample the wall clock at each token-count update:
+  // numerator and denominator are then snapshots of the same instant, with no
+  // periodic re-renders in between, while time stays out of render so the
+  // compiler freeze cannot return. If the stream stalls the display holds its
+  // last value instead of decaying; the final value at endTime is unaffected.
+  const [measuredAt, setMeasuredAt] = useState<number | null>(null);
+  useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- capture the wall-clock instant the token count was observed; it cannot be derived in render without reintroducing the compiler-frozen Date.now()
+    setMeasuredAt(Date.now());
+  }, [lastAssistantTokenCount]);
+
   let tokenPerSecond = 0;
   if (lastAssistantTokenCount > 0 && startTime) {
-    // eslint-disable-next-line react-hooks/purity
-    const elapsedSec = ((endTime ?? Date.now()) - startTime) / 1000;
+    const sampledAt = endTime ?? measuredAt;
+    const elapsedSec = sampledAt !== null ? (sampledAt - startTime) / 1000 : 0;
     tokenPerSecond = elapsedSec > 0 ? lastAssistantTokenCount / elapsedSec : 0;
   }
 
--- a/react/src/hooks/useTokenizer.ts
+++ b/react/src/hooks/useTokenizer.ts
@@ -10,17 +10,30 @@ export const encodeAsync = async (str: string) => {
   return encode(str).length;
 };
 
+// Indirection object so `useTokenCount`'s async call can be spied on in tests.
+// A direct module-local call to `encodeAsync` cannot be intercepted under ESM,
+// which makes the out-of-order cancellation behavior (FR-3091) untestable.
+export const tokenCounter = { encodeAsync };
+
 export const useTokenCount = (input: string = '') => {
+  'use memo';
   const [value, setNum] = useState(0);
 
   useEffect(() => {
+    let cancelled = false;
     startTransition(() => {
-      encodeAsync(input)
-        .then(setNum)
+      tokenCounter
+        .encodeAsync(input)
+        .then((count) => {
+          if (!cancelled) setNum(count);
+        })
         .catch(() => {
-          setNum(input.length);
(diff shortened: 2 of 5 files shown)


agatha197:
* **#7822** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7822?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7822?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7822).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/hooks/useTokenizer.ts
```diff
@@ -11,16 +11,23 @@ export const encodeAsync = async (str: string) => {
...
   const [value, setNum] = useState(0);
 
   useEffect(() => {
+    let cancelled = false;
     startTransition(() => {
       encodeAsync(input)
-        .then(setNum)
+        .then((count) => {
+          if (!cancelled) setNum(count);
+        })
```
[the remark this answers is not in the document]
agatha197: Good call. Added a deterministic regression test in `useTokenizer.test.ts` (`discards a stale result that resolves after a newer input`) that exercises exactly this race: it changes the input before the first `encodeAsync` resolves, resolves the newer computation first, then resolves the stale one later with a different value, and asserts the count does not regress. To make the module-internal async call interceptable under ESM, the call now goes through a small `tokenCounter` indirection object. Verified the test fails when the cancellation guard is removed and passes with it in place.

yomybaby (approved):
-

yomybaby (dismissed):
The content inside the `think` block is not being included in the TPS calculation. Please fix it.

react/src/components/Chat/ChatTokenCounter.tsx
```diff
@@ -39,16 +39,34 @@ const ChatTokenCounter: React.FC<ChatTokenCounterProps> = ({
...
+    return () => window.clearInterval(id);
+  }, [isMeasuring]);
+
   let tokenPerSecond = 0;
   if (lastAssistantTokenCount > 0 && startTime) {
-    // eslint-disable-next-line react-hooks/purity
-    const elapsedSec = ((endTime ?? Date.now()) - startTime) / 1000;
+    const elapsedSec = ((endTime ?? now) - startTime) / 1000;
     tokenPerSecond = elapsedSec > 0 ? lastAssistantTokenCount / elapsedSec : 0;
   }
```
ironAiken2: The 200 ms interval isn't actually needed here. TPS only changes when the token count changes (i.e., when a chunk arrives) — between chunks the ticking denominator only produces a cosmetic decay, at the cost of a re-render loop (5/s) while streaming.

It also leaves the numerator and denominator sampled at different moments: on a chunk render `now` is up to 200 ms stale, and on an interval tick the token count is from the last chunk. Sampling the wall clock in the effect that reacts to the token-count update pairs them as snapshots of the same instant, while still keeping time out of render so the React Compiler freeze this PR fixes cannot come back:

```suggestion
  // While streaming (startTime set, endTime not yet), the elapsed denominator
  // needs a current timestamp — but reading `Date.now()` directly in render
  // does not work: under the React Compiler ('use memo'), the elapsed
  // expression is memoized on its tracked deps (startTime/endTime) only —
  // `Date.now()` is an untracked impure call, so the denominator freezes at
  // ~0s from when startTime was first set, inflating TPS until streaming ends.
  // TPS only changes when the token count changes, so instead of a ticking
  // timer we sample the wall clock at each token-count update: numerator and
  // denominator are then snapshots of the same moment, with no periodic
  // re-renders in between.
  const [measuredAt, setMeasuredAt] = useState<number | null>(null);
  useEffect(() => {
    setMeasuredAt(Date.now());
  }, [lastAssistantTokenCount]);

  let tokenPerSecond = 0;
  if (lastAssistantTokenCount > 0 && startTime) {
    const sampledAt = endTime ?? measuredAt;
    const elapsedSec = sampledAt !== null ? (sampledAt - startTime) / 1000 : 0;
    tokenPerSecond = elapsedSec > 0 ? lastAssistantTokenCount / elapsedSec : 0;
  }
```

Behavioral difference: if the stream stalls, the displayed TPS holds its last value instead of decaying — which is arguably more consistent ("the measurement didn't change, so the display doesn't either"). The final value at `endTime` is identical in both versions.

Note: the `live elapsed denominator` unit test asserts the wall-clock decay (`advanceTimersByTime` → 50.0), so it needs to be rewritten to the new semantics — denominator advances when the token count updates (rerender with a larger count after advancing the clock), not with time alone.

ironAiken2 (changes requested):
The 200 ms interval driving the TPS denominator should be replaced with sampling the wall clock at token-count updates — see the inline suggestion. It removes the periodic re-render loop, keeps numerator/denominator coherent, and preserves the React Compiler freeze fix. The `live elapsed denominator` unit test needs to be rewritten to the new semantics accordingly.

react/src/components/Chat/ChatTokenCounter.tsx
agatha197: Adopted — thanks, this is the better approach. Replaced the 200 ms interval with a wall-clock sample captured in an effect keyed on the token count (`measuredAt`), so the numerator and denominator are now snapshots of the same instant and there is no periodic re-render loop. Time stays out of render, so the React Compiler freeze fix is preserved. As you noted, a stalled stream now holds its last value instead of decaying, and the final value at `endTime` is unchanged.

Also rewrote the `elapsed denominator` unit test to the new semantics: it asserts TPS holds steady while only time passes and re-samples when the token count changes (a ticking timer would have decayed the held value). Done in `1a2c595cc`.

agatha197:
@yomybaby Re: the `think` block content not being included in the TPS calculation — this was fixed in `59b065ab3` (pushed after your review). `<think>` output is extracted into `reasoning` parts by `extractReasoningMiddleware({ tagName: "think" })`, and `ChatTokenCounter` now counts both `text` and `reasoning` parts in the TPS numerator and the total token count. Covered by the `reasoning part inclusion` unit tests in `ChatTokenCounter.test.tsx`.

Additionally, on the client-side streaming path the TPS numerator now prefers the model-reported `outputTokens` (usage) over the gpt-tokenizer estimate when available (`38a25750b`).

ironAiken2 (changes requested):
The reasoning token is not included.

ironAiken2 (approved):
LGTM

PR #7838, opened 2026-06-11
fix(FR-3092): scope personal session list to current user

### LINKED ISSUE #7824: User Sessions page (/session) shows other users' sessions for admin (should be user-scoped)
# Problem

On the per-user **Sessions** page (`/session`), an admin account sees sessions created by **other users** in addition to its own. The `/session` page is the individual user's page and must show **only the current user's own sessions**, regardless of role.

# Root cause (confirmed, WebUI-owned)

`react/src/pages/ComputeSessionListPage.tsx:159` builds the query with a **project scope and no user restriction**:

```
// ComputeSessionListPage.tsx:158-164
const queryVariables: ComputeSessionListPageQuery$variables = {
  scopeId: `project:${currentProject.id}`,
  offset: baiPaginationOption.offset,
  first: baiPaginationOption.first,
  filter: mergeFilterValues([statusFilter, queryParams.filter, typeFilter]),
  order: queryParams.order || '-created_at',
};
```

The same `$scopeId` is reused by all six `compute_session_nodes` calls in the query (main list at L178–184, plus the `all`/`interactive`/`inference`/`batch`/`system` tab-count subqueries at L195–234) — so **both the table rows and the per-type tab counts** are project-wide.

**Why only admins see others' sessions (verified in manager source):** `ComputeSessionPermissionContextBuilder.build_ctx_in_project_scope` (`backend.ai/src/ai/backend/manager/models/session/row.py:1504-1514`) merges (1) role-derived permissions on all sessions in the project with (2) the caller's own sessions. For a plain `user` role, `MEMBER_PERMISSIONS = frozenset()` (empty, row.py:1392) so they only see their own sessions; for `admin`/`superadmin` — and also **`monitor`** (`MONITOR_PERMISSIONS` includes `READ_ATTRIBUTE`, row.py:1387-1390) — the project context covers everyone's sessions. The WebUI asks for "everything readable in project X" and the backend answers correctly. **The bug is in what the WebUI requests; no backend change is needed.**

**Not a recent regression:** the page has been project-wide-scoped since its creation in FR-319 (commit `490bab509`, PR #2932, using the now-deprecated `project_id: $projectId` argument with identical semantics); `a7f175e0b` (FR-2028, PR #5262) later migrated it to `scopeId: "project:..."`. The project-wide visibility was originally deliberate — the page even ships an admin/monitor-only `user_email` filter (L447–452). It became a bug semantically once the dedicated admin pages landed: `/admin-session` (AdminComputeSessionListPage, FR-2316 — intentionally omits `scopeId` for SystemScope) and `/project-admin-session` (ProjectAdminSessionPage, FR-2944 — `projectSessionsV2(scope: { projectId })`). With those pages existing, `/session` is now the **personal** page and must be user-scoped.

# Recommended fix (WebUI)

**Keep the `project:` scope and merge a current-user filter** into every `compute_session_nodes` call (the backend filter spec supports `user_id` → `user_uuid`; `gql_legacy/session.py:124`):

1. `const [currentUser] = useCurrentUserInfo();` (`react/src/hooks/backendai.tsx:98`, `uuid` = `baiClient.user_uuid`) → `const currentUserFilter = \`user_id == "${currentUser.uuid}"\``.
1. Main list (L162): `filter: mergeFilterValues([statusFilter, queryParams.filter, typeFilter, currentUserFilter])`.
1. The five tab-count subqueries currently use **literal** filter strings — convert them to query variables (e.g. `$allFilter`, `$interactiveFilter`, …) computed in JS with `mergeFilterValues`, then run `pnpm relay`.
1. Remove the now-meaningless admin-only `user_email` filter property (L447–452) — on a strictly personal page every row is the current user's (admins filter by owner on the admin pages instead).

**Alternative considered and rejected:** `scopeId: \`user:${currentUser.uuid}\`` (format is backend-supported; precedent in `LegacyModelTryContentButton.tsx:407`). Rejected because `build_ctx_in_user_scope` (row.py:1517-1522) applies **no project restriction** — the page would show the user's sessions across all projects, silently decoupling it from the header project selector.

# Related affected surfaces (same root pattern — suggest separate ticket)

- `react/src/pages/DashboardPage.tsx:83` passes the same `project:` scope into `SessionCountDashboardItemFragment` / `RecentlyCreatedSessionFragment`, whose aliases are literally `myInteractive`, `myBatch`, … (`SessionCountDashboardItem.tsx:44-70`) — an admin's dashboard "my session" counts and recent-sessions list also include other users' sessions.
- (For completeness) `AdminDashboardPage.tsx:78` uses the same scope — admin page, likely intentional. `VFolderNodeListPage.tsx:166` follows the analogous project-scope pattern via the vfolder RBAC builder — possible sibling issue on the personal folders page, needs a product call.

# Open questions

- Product confirmation: should `/session` be strictly own-sessions for admins (recommended — `/admin-session` and `/project-admin-session` cover the admin use case)? Should `monitor` keep any cross-user visibility there?
- Fix DashboardPage "my sessions" in the same PR or a separate FR ticket?

# Notes

- Found during 26.4.8 release testing. Root cause verified against both WebUI and manager sources (line references as of current `main`).

JIRA Issue: FR-3092

ironAiken2:
Resolves #7824 (FR-3092)

> [!NOTE]
> CSV export currently does not support email-based filtering. As a result, when an admin exports from the user session page, the list includes sessions from all users rather than just their own. This will be addressed once BA-6480 is resolved

## Problem

The User Sessions page (`/session`) showed **other users' sessions** to admins and monitors. Backend RBAC grants admins/monitors read permission on all sessions in a project, and the page queried `compute_session_nodes` with only `scope_id: project:<id>` and no user restriction — so both the table rows and the per-type tab count badges were project-wide instead of personal.

## What changed

`react/src/pages/ComputeSessionListPage.tsx`:

- Added `const [currentUser] = useCurrentUserInfo()` and built a `user_id == "<current user uuid>"` filter.
- Merged the current-user filter into the main list query filter via `mergeFilterValues([statusFilter, queryParams.filter, typeFilter, currentUserFilter])`.
- Converted the five tab-count subqueries (`all` / `interactive` / `inference` / `batch` / `system`) from literal filter strings inside the GraphQL tag to query variables (`$allFilter`, `$interactiveFilter`, `$inferenceFilter`, `$batchFilter`, `$systemFilter`) computed in JS, preserving the exact existing status/type semantics and adding the current-user clause.
- Removed the admin/monitor-only `user_email` property from the filter dropdown — on a strictly personal page every row belongs to the current user, so the filter was meaningless.
- The `project:` scope is kept, so the page still reacts to the header project selector. Regenerated Relay artifacts via `pnpm relay`.

## Verification

```
=== Relay ===
--- Relay: PASS ---
=== Lint ===
--- Lint: PASS ---
=== Format ===
--- Format: PASS ---
=== TypeScript ===
--- TypeScript: PASS ---
=== Vite warmup paths ===
--- Vite warmup paths: PASS ---
=== ALL PASS ===
```

## Review follow-ups

- `user_id` queryfilter support on `compute_session_nodes` has been verified statically (schema/queryfilter spec) only — it should be runtime-verified against a live manager before merge.
- The page's CSV export still exports unscoped (project-wide) sessions for admins; tracked as a separate Jira follow-up.

### DIFF
--- a/react/src/pages/ComputeSessionListPage.tsx
+++ b/react/src/pages/ComputeSessionListPage.tsx
@@ -18,7 +18,7 @@ import SessionNodes, {
 import { handleRowSelectionChange } from '../helper';
 import { ExtractResultValue } from '../helper/resultTypes';
 import { useWebUINavigate } from '../hooks';
-import { useCurrentUserRole } from '../hooks/backendai';
+import { useCurrentUserInfo, useCurrentUserRole } from '../hooks/backendai';
 import { useBAIPaginationOptionStateOnSearchParam } from '../hooks/reactPaginationQueryOptions';
 import { useBAISettingUserState } from '../hooks/useBAISetting';
 import { useCSVExport } from '../hooks/useCSVExport';
@@ -44,7 +44,6 @@ import {
   BAIPropertyFilter,
   BAISelectionLabel,
   BAISessionsIcon,
-  filterOutEmpty,
   filterOutNullAndUndefined,
   INITIAL_FETCH_KEY,
   mergeFilterValues,
@@ -78,11 +77,15 @@ type SessionNode = NonNullableNodeOnEdges<ComputeSessionNodesData>;
 
 const CARD_MIN_HEIGHT = 200;
 
+const NOT_FINISHED_STATUS_FILTER =
+  'status != "TERMINATED" & status != "CANCELLED"';
+
 const ComputeSessionListPage = () => {
   'use memo';
   const currentProject = useCurrentProjectValue();
 
   const userRole = useCurrentUserRole();
+  const [currentUser] = useCurrentUserInfo();
 
   const { t } = useTranslation();
   const { token } = theme.useToken();
@@ -146,7 +149,7 @@ const ComputeSessionListPage = () => {
   const statusFilter =
     queryParams.statusCategory === 'running' ||
     queryParams.statusCategory === undefined
-      ? 'status != "TERMINATED" & status != "CANCELLED"'
+      ? NOT_FINISHED_STATUS_FILTER
       : 'status == "TERMINATED" | status == "CANCELLED"';
 
   const isNotRunningCategory = (status?: string | null) => {
@@ -155,12 +158,45 @@ const ComputeSessionListPage = () => {
 
   const [fetchKey, updateFetchKey] = useFetchKey();
 
+  // This page is strictly personal: even admins/monitors (who hold read
+  // permission on all project sessions) should only see their own sessions.
+  const currentUserFilter = `user_id == "${currentUser.uuid}"`;
+
   const queryVariables: ComputeSessionListPageQuery$variables = {
     scopeId: `project:${currentProject.id}`,
     offset: baiPaginationOption.offset,
     first: baiPaginationOption.first,
-    filter: mergeFilterValues([statusFilter, queryParams.filter, typeFilter]),
+    filter: mergeFilterValues([
+      statusFilter,
+      queryParams.filter,
+      typeFilter,
+      currentUserFilter,
+    ]),
     order: queryParams.order || '-created_at',
+    allFilter: mergeFilterValues([
+      NOT_FINISHED_STATUS_FILTER,
+      currentUserFilter,
+    ]),
+    interactiveFilter: mergeFilterValues([
+      NOT_FINISHED_STATUS_FILTER,
+      'type == "interactive"',
+      currentUserFilter,
+    ]),
+    inferenceFilter: mergeFilterValues([
+      NOT_FINISHED_STATUS_FILTER,
+      'type == "inference"',
+      currentUserFilter,
+    ]),
+    batchFilter: mergeFilterValues([
+      NOT_FINISHED_STATUS_FILTER,
+      'type == "batch"',
+      currentUserFilter,
+    ]),
+    systemFilter: mergeFilterValues([
+      NOT_FINISHED_STATUS_FILTER,
+      'type == "system"',
+      currentUserFilter,
+    ]),
   };
 
   const deferredQueryVariables = useDeferredValue(queryVariables);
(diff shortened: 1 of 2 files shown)


ironAiken2:
* **#7838** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7838?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7838?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7838).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/pages/ComputeSessionListPage.tsx
```diff
@@ -155,85 +158,123 @@ const ComputeSessionListPage = () => {
 
   const [fetchKey, updateFetchKey] = useFetchKey();
 
+  // This page is strictly personal: even admins/monitors (who hold read
+  // permission on all project sessions) should only see their own sessions.
+  const currentUserFilter = `user_id == "${currentUser.uuid}"`;
+
```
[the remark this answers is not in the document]
ironAiken2: Agreed — this is a real gap, but intentionally out of scope for this PR. The unscoped CSV export on the personal page (and the unscoped session-detail drawer) is tracked as a Jira follow-up: FR-3105. The PR description's Review follow-ups section also notes this.

react/src/pages/ComputeSessionListPage.tsx
```diff
@@ -428,7 +469,7 @@ const ComputeSessionListPage = () => {
                 ]}
               />
               <BAIPropertyFilter
-                filterProperties={filterOutEmpty([
+                filterProperties={[
```
agatha197, line 472: **Owner column — default to hidden on this page.**

Removing the `user_email` *filter* property here is right, but the table still renders an **Owner (OwnerEmail) column** that is shown by default. It is defined in the shared `SessionNodes` (`react/src/components/SessionNodes.tsx:369-374`): superadmin-only, `defaultHidden: false`. Now that every row on `/session` belongs to the current user, that column is redundant noise on the personal page.

Suggestion: make the owner column **default-hidden on `/session`**. Note `SessionNodes` is shared with `/admin-session` (`AdminComputeSessionListPage`), where the owner column *is* meaningful — so please scope this per-page (e.g. a prop on `SessionNodes`, or a `columnOverrides`/`customizeColumns` default from `ComputeSessionListPage`) rather than flipping the shared `defaultHidden` globally.

react/src/pages/ComputeSessionListPage.tsx
```diff
@@ -155,85 +158,123 @@ const ComputeSessionListPage = () => {
 
   const [fetchKey, updateFetchKey] = useFetchKey();
 
+  // This page is strictly personal: even admins/monitors (who hold read
+  // permission on all project sessions) should only see their own sessions.
+  const currentUserFilter = `user_id == "${currentUser.uuid}"`;
```
agatha197, line 163: **CSV export content — trim/scope the owner-email field too.**

The owner column carries `exportKey: 'user_email'` (`react/src/components/SessionNodes.tsx:372`), so the superadmin CSV export (`exportSettings` / `onExport` at `ComputeSessionListPage.tsx:596`) still emits a `user_email` field. Now that the page is self-scoped, that value is always the current user's own email — redundant in the export.

This compounds the already-noted limitation that the CSV export remains **project-wide** (it does not apply `currentUserFilter`), so an admin can still dump other users' sessions from the personal page. Whatever the resolution for the project-wide scope follow-up, the CSV *content* should also drop or scope the redundant owner-email field for `/session`.

agatha197 (commented):
Two follow-up suggestions on the consequences of making `/session` a strictly personal page. Both are about the owner-email surface that the new `user_id` scoping makes redundant here — neither blocks the core fix.

react/src/pages/ComputeSessionListPage.tsx
ironAiken2: Thanks for the suggestion, but I think we can keep the owner column as is. There is no real harm in showing it to the user — it simply displays their own email on the personal page. Also, the CSV export already lets users pick which fields to include, and the table column settings allow hiding it manually if anyone finds it redundant. Given that, I'd rather avoid adding per-page column override plumbing just for this. Keeping it unchanged.

react/src/pages/ComputeSessionListPage.tsx
ironAiken2: Thanks for the suggestion, but I think we can keep the owner column as is. There is no real harm in showing it to the user — it simply displays their own email on the personal page. Also, the CSV export already lets users pick which fields to include, and the table column settings allow hiding it manually if anyone finds it redundant. Given that, I'd rather avoid adding per-page column override plumbing just for this. Keeping it unchanged.

agatha197 (changes requested):
Please add user scope to CSV exporting too.

ironAiken2:
> Please add user scope to CSV exporting too.

​The server-side CSV extraction API does not currently support filtering based on user email addresses. We will address this once it becomes available.

agatha197 (approved):
LGTM

PR #7884, opened 2026-06-12
fix(FR-3131): polish deployment preset Resources form and review display

agatha197:
Resolves #7883 (FR-3131)

> **Stacked on** #7885 (FR-3129). Review/merge #7885 first.

## Problem

Found while working on FR-3129 (showing resource opts on the deployment preset review page). Separate UX defects/polish in the admin deployment preset **Resources** section:

- Entering a long custom Resource Option **key** expanded its input and squashed the value input to near-zero width.
- The MEM input defaulted to `m` (MiB) instead of the expected `g` (GiB).
- The review step rendered `undefined NaN` for resource slots when a slot row was added but left empty.

## Changes

- **Resource Options input** (`AdminDeploymentPresetSettingPageContent.tsx`): lay out each key/value row as a CSS grid with `minmax(0, 1fr)` tracks so a long custom key can't expand its input and squash the value field — the tracks stay 50/50 and long content scrolls within each. (`minWidth: 0` on the flex `Form.Item` alone didn't work because it doesn't cascade to antd's nested control wrappers.)
- **MEM default unit** (`AdminDeploymentPresetResourceFields.tsx`): pass `defaultUnit="g"` to `BAIDynamicUnitInputNumber` so memory defaults to GiB.
- **Review display** (`AdminDeploymentPresetReviewSummary.tsx`): filter out empty/partial resource-slot rows (show `-` instead of `undefined NaN`); render resource options as wrapping `name: value` tokens that stay inside the card; trim names/values at display.

### Resource Options input — long key keeps a 50/50 split (MEM also defaults to GiB)
resource-options-input

### Review step — empty resource slots show `-`, long resource options wrap inside the card
review-resources

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7887, opened 2026-06-13
fix(FR-3133): remove duplicate '#' prefix in Korean ApplyingRevision i18n string

agatha197:
Resolves #7884 (FR-3133)

## Summary

The Korean translation for `deployment.ApplyingRevision` contained `#{{revisionNumber}}` while all other 20 languages use `{{revisionNumber}}` (without `#`). The call site in `DeploymentConfigurationSection.tsx` already prepends `#` to the value before passing it to the translation function, so Korean users saw `##4` instead of `#4` when a revision was being applied.

**Root cause**: `ko.json` `ApplyingRevision` key had an extra `#` that is inconsistent with all other language files.

**Fix**: Remove the extra `#` from the Korean `ApplyingRevision` string to match the behavior of all other languages.

## Changed Files

- `resources/i18n/ko.json` — remove duplicate `#` from `deployment.ApplyingRevision`

## Verification

```
- en: "Applying revision {{revisionNumber}}." + code prepends `#` → "Applying revision #4." ✅
- ko (before): "리비전 #{{revisionNumber}}을(를) 적용 중입니다." + code prepends `#` → "리비전 ##4을(를) 적용 중입니다." ❌
- ko (after):  "리비전 {{revisionNumber}}을(를) 적용 중입니다." + code prepends `#` → "리비전 #4을(를) 적용 중입니다." ✅
```

### DIFF
--- a/resources/i18n/ko.json
+++ b/resources/i18n/ko.json
@@ -995,7 +995,7 @@
     "ApplyRevision": "리비전 적용",
     "ApplySuccess": "리비전 #{{revisionNumber}} 적용이 요청되었습니다.",
     "Applying": "적용 중",
-    "ApplyingRevision": "리비전 #{{revisionNumber}}을(를) 적용 중입니다.",
+    "ApplyingRevision": "리비전 {{revisionNumber}}을(를) 적용 중입니다.",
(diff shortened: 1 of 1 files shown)


nowgnuesLee (approved):
LGTM

PR #7906, opened 2026-06-15
fix(FR-3141): preserve image architecture when loading a previous revision in Add Revision

### LINKED ISSUE #7904: Add Revision: loading a previous revision selects wrong image architecture (aarch64 instead of original x86_64)
## Summary

When adding a revision to a model deployment and using **"load previous revision"** (이전 배포 불러오기 / 리비전 가져오기), the image field does not preserve the originally-used image. Instead it auto-selects the top-most image in the list, which is an `aarch64` image, even though the loaded revision used an `x86_64` image.

## Reproduction

1. Have a successful deployment #1 running on an x86_64 image (e.g. kimi k2.7).
1. Create revision #2.
1. Use "load previous revision" to import revision #1.
1. Modify arguments and deploy.
1. **Observed:** the selected image is a different image (top-most `aarch64`) instead of the original `x86_64` image.

## Root cause (frontend)

The architecture information is lost when prefilling the form, not a parsing failure:

- `react/src/components/DeploymentAddRevisionModal.tsx` (~L861) `applyRevisionToCustomForm` sets the form's `environments.version` to `rev.imageV2.identity.canonicalName`, which is `registry/namespace:tag` **without** the `@architecture` suffix. The fragment (~L310) only selects `imageV2.identity.canonicalName` and does not fetch `architecture`.
- In `react/src/components/ImageEnvironmentSelectFormItems.tsx`, the exact-match step compares `getImageFullName(image) === version`, where `getImageFullName` always appends `@${architecture`} (`react/src/helper/index.tsx:398`). Since the prefilled `version` has no `@arch`, exact match fails for every image.
- The fallback partial-match step strips `@arch` and matches by `name:tag` only, so both x86_64 and aarch64 candidates match; `_.find` returns the **first** entry. The image array is sorted by version desc then `localeCompare(architecture)`, and `aarch64` sorts before `x86_64`, so the aarch64 image is selected — matching the reported symptom.

## Fix direction

1. Add `architecture` to the `imageV2.identity` fragment selection in `DeploymentAddRevisionModal.tsx` (schema `ImageV2IdentityInfo.architecture: String!` already exists).
1. In `applyRevisionToCustomForm`, set `version` to `${canonicalName}@${architecture`} so the exact-match step succeeds and the original image/architecture is preserved.
1. Check the async preset-load path (~L622-633) which also fetches only `canonicalName` — same symptom likely applies there.

## Origin

Reported in the devops / Backend.AI Cloud Teams channel.

JIRA Issue: FR-3141

agatha197:
Resolves #7904 (FR-3141)

> [!NOTE]
> Test: In Dogbowl, Add Revisions with different architecture.

## Problem

When adding a revision to a model deployment and using **"load previous revision"** (이전 배포 불러오기 / 리비전 가져오기), the image field did not preserve the originally-used image. It silently switched to the top-most image in the list — an `aarch64` image — even though the loaded revision (e.g. kimi k2.7) used an `x86_64` image.

### Reproduction

1. Successful deployment #1 running on an x86_64 image.
2. Create revision #2 → load previous revision (#1) → modify args → deploy.
3. **Observed:** a different image (top-most `aarch64`) is selected instead of the original `x86_64` image.

## Root cause (frontend)

The architecture information was dropped during form prefill — not a parsing failure:

- `applyRevisionToCustomForm` (and the preset async-load path) set the form's `environments.version` to `imageV2.identity.canonicalName`, which is `registry/namespace:tag` **without** the `@architecture` suffix. The fragment/query fetched only `canonicalName`, not `architecture`.
- `ImageEnvironmentSelectFormItems`' exact-match step compares against `getImageFullName(image)` = `registry/namespace:tag@architecture`. With no `@arch` in the prefilled value, exact match failed for every image.
- The fallback partial-match step strips `@arch` and matches by `name:tag` only, so both x86_64 and aarch64 candidates matched; `_.find` returns the **first** entry. The image array is sorted by version desc then `localeCompare(architecture)`, and `aarch64` sorts before `x86_64`, so the aarch64 image got selected.

## Fix

- Fetch `identity.architecture` in the revision fragment (`DeploymentAddRevisionModal_revisionSource`) and in the preset image query (`DeploymentAddRevisionModalImageNameQuery`).
- Prefill `version` as `canonicalName@architecture` so the exact-match step picks the originally-used image. (The existing partial-match already proved `getImageFullName(image)` equals `canonicalName@architecture` for the matching image, so this is format-safe.)
- Applied to both prefill entry points: load-previous-revision and preset async load.

## Verification

`bash scripts/verify.sh` — Lint **PASS**, Format **PASS**. The Relay / TypeScript failures observed locally are **pre-existing and unrelated** to this change:

- An in-progress, uncommitted `data/schema.graphql` sync (UNRELEASED→26.4.4) on the working tree leaves 4 `__generated__` mutation files dirty and breaks `AdminDeploymentPresetSettingPage.tsx` (`ModelDefinitionInput`→`PresetModelDefinitionInput`). None of these files are part of this PR.
- `FluentEmojiIcon.tsx` reports a local `@lobehub/fluent-emoji` patched-dep type error; the file is byte-identical to `main` (environmental, green on CI).

This PR commits only the 5 files relevant to the fix.

## Test plan

- Load a previous revision that used an `x86_64` image and confirm the image field keeps `x86_64` (not the top `aarch64`).
- Repeat via a deployment preset.


agatha197:
* **#7906** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7906?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7906?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7906).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

nowgnuesLee (approved):
LGTM

PR #7927, opened 2026-06-16
feat(FR-3147): add opt-in auto-refresh interval dropdown to BAIFetchKeyButton

### LINKED ISSUE #7920: Add auto-refresh interval menu (more button) to BAIFetchKeyButton
Add a "more" button (dropdown menu) to the shared `BAIFetchKeyButton` component that lets the user opt in to automatic refresh at a chosen interval (every few seconds). Auto-refresh should only run when the user has explicitly selected an interval from the menu — it must stay off by default. The menu should let the user pick the refresh interval and turn auto-refresh on/off.

Captured while working on branch: fix/FR-3141-revision-image-arch

JIRA Issue: FR-3147

agatha197:
Resolves #7920 (FR-3147)

## Summary

Adds an **opt-in auto-refresh interval dropdown** to the shared `BAIFetchKeyButton` (BUI package). The feature is **off by default** and only runs auto-refresh once the user explicitly selects an interval — matching the issue's requirement.

The component stays **presentational/controllable**: BUI cannot import the host's `useBAISetting` (it lives in `react/src`), so the interval is exposed as a controllable value. A follow-up stacked PR wires `useBAISetting` persistence into the consumers that actually poll.

## What changed

- **The parent opts in simply by wiring `onChangeAutoUpdateDelay`** — there is **no separate boolean flag**. Providing that handler shows the ellipsis dropdown and makes `autoUpdateDelay` a controllable value; omitting it renders **exactly as before** (a single tooltip-wrapped refresh button), so all ~55 existing call sites are unchanged. (A dropdown is only meaningful when changes can be handled, so the handler's presence is the natural gate — and unlike "is `autoUpdateDelay` set", it doesn't wrongly trigger for consumers passing `autoUpdateDelay={null}`.)
- **`autoUpdateDelay?: number | null`** is the single auto-refresh interval value (ms; `null` = Off) — controllable via `onChangeAutoUpdateDelay`. No separate "value" prop.
- **`autoUpdateDelayOptions?: number[]`** — overridable preset list (default `AUTO_UPDATE_DELAY_OPTIONS` = `[5s, 10s, 15s, 30s, 60s]`).
- **UI (per the request):** a `Space.Compact` of the refresh `Button` + an ellipsis `Dropdown`. **Off** → only the refresh arrow (current look). **On** → the button shows `N초` / `Ns` and the dropdown lists the presets plus an **Off** entry, with a check mark on the active one.
- **i18n:** added `comp:BAIFetchKeyButton` keys `AutoRefresh`, `EverySeconds` (`{{count}}` placeholder), `Off` to **all 21 BUI locale files**. Host `resources/i18n` untouched (comp:* keys live only in the BUI instance).
- **Story:** added an `Interval Dropdown` story + prop docs/argTypes.

## Note on UI

The reference image mentioned in the request didn't come through, so the exact placement follows the textual spec (off → arrow only; on → `N초` label; `Space.Compact` + ellipsis `Dropdown`). Happy to adjust placement/labels if the mockup differs.

## Screenshots

From the Storybook `Interval Dropdown` story. Off by default (just the refresh arrow + caret); once an interval is picked the dropdown trigger shows the `N초` label next to the caret, and the active option is check-marked. The caret flips up while the menu is open. (Storybook uses the English BUI i18n instance — Korean renders `끄기` / `15초`.)

| Off (default) | Menu open | Interval selected (15s) |
|---|---|---|
| off | menu | selected 15s |

Reopening the menu shows the check mark on the active interval (caret pointing up):

menu with 15s checked

## Verification

`bash scripts/verify.sh` — for this change: **Relay / Format / `backend.ai-ui` Lint — PASS · `backend.ai-ui` & `react` `tsc` — 0 errors.** (The aggregate harness occasionally flags two phantom `tsc` errors in untouched files — `FluentEmojiIcon.tsx`, `AdminDeploymentPresetSettingPage.tsx` — which don't reproduce on a clean run or on `main`; a fresh-worktree incremental-cache artifact, not part of this change.)

## Stack

Bottom of a 2-PR stack:
1. **(this PR)** `BAIFetchKeyButton` gains the opt-in interval dropdown.
2. **(#7928)** Wire `useBAISetting` persistence and enable the dropdown only on the components that currently poll.

### DIFF
--- a/packages/backend.ai-ui/src/components/BAIFetchKeyButton.stories.tsx
+++ b/packages/backend.ai-ui/src/components/BAIFetchKeyButton.stories.tsx
@@ -23,10 +23,12 @@ A refresh button that manages fetch keys for data refetching with auto-update ca
 | \`loading\` | \`boolean\` | \`false\` | Loading state of the data fetch |
 | \`lastLoadTime\` | \`Date\` | \`undefined\` | Timestamp of the last successful load |
 | \`showLastLoadTime\` | \`boolean\` | \`false\` | Shows "Last updated: X ago" in tooltip |
-| \`autoUpdateDelay\` | \`number \\| null\` | \`null\` | Auto-refresh interval in milliseconds |
+| \`autoUpdateDelay\` | \`number \\| null\` | \`null\` | Auto-refresh interval in ms (controllable); \`null\` = Off |
 | \`onChange\` | \`(fetchKey: string) => void\` | - | Callback fired when fetch key updates |
 | \`hidden\` | \`boolean\` | \`false\` | Hides the button completely |
 | \`pauseWhenHidden\` | \`boolean\` | \`true\` | Pauses auto-update when button is hidden |
+| \`onChangeAutoUpdateDelay\` | \`(delayMs: number \\| null) => void\` | - | Fired when the user picks an interval; **providing it shows the dropdown** |
+| \`autoUpdateDelayOptions\` | \`number[]\` | \`[5000, 10000, 15000, 30000, 60000]\` | Interval presets (ms); "Off" is auto-prepended |
 
 For all other props, refer to [Ant Design Button](https://ant.design/components/button).
         `,
@@ -67,7 +69,8 @@ For all other props, refer to [Ant Design Button](https://ant.design/components/
     },
     autoUpdateDelay: {
       control: { type: 'number' },
-      description: 'Auto-refresh interval in milliseconds, null to disable',
+      description:
+        'Auto-refresh interval in ms (controllable); null disables (Off)',
       table: {
         type: { summary: 'number | null' },
         defaultValue: { summary: 'null' },
@@ -94,6 +97,23 @@ For all other props, refer to [Ant Design Button](https://ant.design/components/
       table: {
         type: { summary: 'boolean' },
         defaultValue: { summary: 'true' },
+      },
+    },
+    onChangeAutoUpdateDelay: {
+      action: 'onChangeAutoUpdateDelay',
+      description:
+        'Fired when the user selects an interval or "Off"; providing it shows the dropdown',
+      table: {
+        type: { summary: '(delayMs: number | null) => void' },
+      },
+    },
+    autoUpdateDelayOptions: {
+      control: { type: 'object' },
+      description:
+        'Interval presets in ms; "Off" is auto-prepended by the component',
+      table: {
+        type: { summary: 'number[]' },
+        defaultValue: { summary: '[5000, 10000, 15000, 30000, 60000]' },
       },
     },
   },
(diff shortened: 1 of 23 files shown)


agatha197:
* **#8265** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8265?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8233** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8233?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7928** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7928?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7927** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7927?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7927?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7927).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
```diff
@@ -116,24 +188,93 @@ const BAIFetchKeyButton: React.FC<BAIFetchKeyButtonProps> = ({
...
+          iconPosition="end"
+          icon={isDropdownOpen ? <CaretUpOutlined /> : <CaretDownOutlined />}
+          aria-label={t('comp:BAIFetchKeyButton.AutoRefresh')}
+        >
+          {isAutoRefreshOn
+            ? t('comp:BAIFetchKeyButton.EverySeconds', {
+                count: (selectedDelay as number) / 1000,
+              })
+            : null}
+        </Button>
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit. The dropdown trigger now spreads `{...buttonProps}`, so it inherits `type`, `danger`, `ghost`, `disabled`, etc. from the consumer — `disabled` now also prevents opening/changing the interval, and styling stays consistent with the refresh button. The trigger keeps its own `icon`/`iconPosition`/`aria-label`/`size` (declared after the spread).

packages/backend.ai-ui/src/components/BAIFetchKeyButton.stories.tsx
```diff
@@ -209,6 +228,53 @@ export const AutoUpdate: Story = {
   },
 };
 
+export const IntervalDropdown: Story = {
+  parameters: {
+    docs: {
+      description: {
+        story:
+          'Providing `onChangeAutoUpdateDelay` opts the button into the interval dropdown: an ellipsis menu lets the user pick an auto-refresh interval or turn it off. Auto-refresh is off by default; once an interval is selected it shows as a label (e.g. "15s") on the refresh button.',
+      },
```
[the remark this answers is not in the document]
agatha197: Updated the story description to match the actual UI: it now says a caret trigger next to the refresh button opens the interval menu, and the selected interval label appears on the dropdown trigger button (not the refresh button).

agatha197:
Follow-up from an adversarial self-review (multi-agent) of this PR — three minor/nit refinements applied on top of the Copilot fixes:

1. **`buttonProps` no longer duplicated onto the caret trigger.** The dropdown trigger now spreads only the visual/state props (`type`, `danger`, `ghost`, `disabled`, `size`, `className`, `style`, …) and excludes `id`, `onClick`, and `title`. Previously the blanket `{...buttonProps}` on both buttons could produce a duplicate `id` (invalid HTML), a consumer `onClick` firing on caret clicks (in addition to toggling the dropdown), and a stray `title` tooltip on the caret. Latent today (no consumer wires `onChangeAutoUpdateDelay` yet) but fixed before polling consumers land.
2. **Caret trigger now has a tooltip** (`Auto Refresh`), matching the refresh button's hover affordance; the `aria-label` is kept for assistive tech.
3. **Storybook `argTypes.autoUpdateDelay.description`** synced with the prop table ("controllable; null disables (Off)").

Backward-compat path (no `onChangeAutoUpdateDelay`) is unchanged. verify.sh: Relay/Lint/Format pass; the lone TS error is the pre-existing `FluentEmojiIcon` worktree artifact, unrelated to these files.

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
```diff
@@ -1,26 +1,70 @@
...
+   * parent decides whether to show the dropdown simply by wiring this handler
+   * (typically to a persisted setting).
+   */
+  onChangeAutoUpdateDelay?: (delayMs: number | null) => void;
+  /**
+   * Interval presets shown in the dropdown, in milliseconds. "Off" (`null`) is
+   * always prepended by the component and must NOT be included here.
+   * Defaults to {@link AUTO_UPDATE_DELAY_OPTIONS}.
+   */
+  autoUpdateDelayOptions?: number[] | readonly number[];
```
nowgnuesLee:
```suggestion
  autoUpdateDelayOptions?: readonly number[];
```

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
```diff
@@ -116,24 +188,109 @@ const BAIFetchKeyButton: React.FC<BAIFetchKeyButtonProps> = ({
...
+      icon: checkMark(selectedDelay === null),
+      onClick: () => setSelectedDelay(null),
+    },
+    ...autoUpdateDelayOptions.map((ms) => ({
+      key: String(ms),
+      label: t('comp:BAIFetchKeyButton.EverySeconds', { count: ms / 1000 }),
+      icon: checkMark(selectedDelay === ms),
+      onClick: () => setSelectedDelay(ms),
+    })),
+  ];
```
nowgnuesLee, line 307: This seems like too much separation.

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
```diff
@@ -116,24 +188,109 @@ const BAIFetchKeyButton: React.FC<BAIFetchKeyButtonProps> = ({
...
+          <Button
+            {...triggerButtonProps}
+            size={size}
+            iconPosition="end"
+            icon={isDropdownOpen ? <CaretUpOutlined /> : <CaretDownOutlined />}
+            aria-label={t('comp:BAIFetchKeyButton.AutoRefresh')}
+          >
+            {isAutoRefreshOn
+              ? t('comp:BAIFetchKeyButton.EverySeconds', {
+                  count: (selectedDelay as number) / 1000,
```
nowgnuesLee: do we need to translate this one?

nowgnuesLee (changes requested):

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
agatha197: Done — simplified to `readonly number[]`. A mutable `number[]` is assignable to `readonly number[]`, so the union was redundant; `AUTO_UPDATE_DELAY_OPTIONS` (an `as const` tuple) still satisfies the default.

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
agatha197: Agreed — removed the `id`/`onClick`/`title` destructuring and now spread `{...buttonProps}` onto the caret trigger directly, same as the refresh button. This keeps the styling/state props (`type`, `danger`, `disabled`, …) consistent across the compact group without the extra separation. (The earlier split was guarding the latent edge case of a consumer passing `id`/`onClick` to a configurable button; accepting the simpler spread as you prefer.)

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
agatha197: Yes — the seconds unit is translated. This follows the existing time-unit translation convention in the codebase (e.g. `t('time.Sec')` → 'sec' / '초' / '秒', used in SessionLauncher, DeploymentAddRevisionModal, etc.). The `EverySeconds` key is already localized per-locale (ko '{{count}}초', ja '{{count}}秒', …), so the dropdown labels render in the user's language rather than a hardcoded string.

packages/backend.ai-ui/src/locale/de.json
```diff
@@ -154,7 +154,12 @@
     "SelectDomain": "Domäne auswählen"
   },
   "comp:BAIFetchKeyButton": {
+    "AutoRefresh": "Automatische Aktualisierung",
+    "EveryHours": "{{count}}h",
+    "EveryMinutes": "{{count}}m",
+    "EverySeconds": "{{count}}s",
```
nowgnuesLee, line 160: do we really need to translate these keys?

nowgnuesLee (dismissed):

packages/backend.ai-ui/src/locale/de.json
agatha197: why not?

agatha197: https://github.com/lablup/backend.ai-webui/blob/654fcf1a3982c8c4c21848ec9c1b2d93bb602b10/resources/i18n/ko.json#L3280 There's already similar code in webui too.

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
```diff
@@ -116,24 +197,140 @@ const BAIFetchKeyButton: React.FC<BAIFetchKeyButtonProps> = ({
...
+  // Icon-only refresh button. When the interval dropdown is shown, the selected
+  // interval label ("Ns") lives on the dropdown trigger, not here.
+  const refreshButton = (
+    <Button
+      title={tooltipTitle ? undefined : t('comp:BAIFetchKeyButton.Refresh')}
+      loading={displayLoading}
+      size={size}
+      icon={<ReloadOutlined />}
+      onClick={() => {
+        onChange(new Date().toISOString());
```
yomybaby, line 237: **Manual refresh should reset the auto-refresh interval.**

Clicking this button while auto-refresh is on only calls `onChange(...)` — it doesn't touch the `useInterval` timer above (line ~195), so the next auto-triggered refresh still fires on its original schedule. Once #8233's countdown border lands on top of this, that mismatch becomes visually obvious: the border keeps animating its own cycle while the actual refresh fires at a different moment.

Suggestion: give the manual click a way to restart the interval — e.g. extend `useInterval` with a `resetKey` param (similar to `useIntervalValue`'s existing `triggerKey`) that re-creates the interval when it changes, and bump that key here alongside `onChange(...)`.

packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx
```diff
@@ -116,24 +197,140 @@ const BAIFetchKeyButton: React.FC<BAIFetchKeyButtonProps> = ({
...
+        open={isDropdownOpen}
+        onOpenChange={setIsDropdownOpen}
+        menu={{ items: menuItems }}
+      >
+        <Tooltip title={t('comp:BAIFetchKeyButton.AutoRefresh')}>
+          <Button
+            {...buttonProps}
+            size={size}
+            iconPosition="end"
+            icon={isDropdownOpen ? <CaretUpOutlined /> : <CaretDownOutlined />}
```
yomybaby, line 326: Suggestion: replace `CaretDownOutlined` / `CaretUpOutlined` with lucide's `ChevronDown` (already a dependency of both `react/package.json` and `packages/backend.ai-ui/package.json`, and used elsewhere in the codebase, e.g. `VFolderTable.tsx`, `RoleAssignmentTab.tsx`). A single `ChevronDown` rotated 180° on open (`transform: isDropdownOpen ? 'rotate(180deg)' : undefined`) is the more common expand/collapse idiom than swapping two different glyphs, and keeps this control visually consistent with lucide icons used elsewhere.

yomybaby (changes requested):
Two follow-up requests from testing the countdown-border UX built on top of this PR (visible in FR-3300 / #8233):

1. Manual refresh doesn't reset the auto-refresh timer, so it can drift out of sync with the countdown border added in #8233 (see inline comment).
2. Icon suggestion: swap the Caret icons for lucide's `ChevronDown` to match the icon set already used elsewhere in the codebase (see inline comment).

yomybaby (approved):
LGTM

Issue #7756, opened 2026-06-08
VFolder selector component supporting subpath as mount source


yomybaby:
## Motivation

When configuring mount-related options (e.g. session creation, Deployment revision), selecting a vfolder currently mounts the **root** of that vfolder (with an optional alias). The backend already supports an option to use a **subpath** of the vfolder as the mount source instead of the root.

## Required Features

Provide a vfolder selection component that supports specifying a subpath as the mount source, in addition to the existing root-mount + alias behavior.

- Allow the user to pick a vfolder and optionally specify a subpath within it to be used as the mount source.
- Surface this in the mount option UIs for session creation and Deployment revision configuration.

## Notes

- The subpath mount source option already exists on the backend; this issue covers the frontend selection component to expose it.

----

Captured while working on branch: feat/FR-3057-routing-followup-fixes

JIRA Issue: FR-3059

PR #7932, opened 2026-06-16
feat(FR-3154): add reusable BAIVFolderMountConfigInput component

nowgnuesLee:
Stacked on #7931 · Part of **FR-3059** (#7756) · Resolves **FR-3154**

## Summary
- Add `BAIVFolderMountConfigInput`, a reusable, **schema-agnostic** controlled input for configuring vfolder mounts (`packages/backend.ai-ui/src/components/fragments/`).
- Composes `BAIVFolderSelect` (multiple) to pick folders; each selected folder renders as a row below the select with:
  - a **mount destination (alias)** input — path inside the container (empty falls back to `/home/work/{name}`),
  - an optional **subpath** free-text input — which subfolder of the vfolder to mount as the source (empty = root),
  - a remove (✕) button.
- Emits a single `VFolderMountConfigValue[]` value via `value` / `onChange`, so it wraps in one `Form.Item` and the consumer maps it to any mount mutation input. No GraphQL/schema coupling.
- Inline validation: alias format (`^[a-zA-Z0-9_/.-]*$`), alias overlap between folders, and relative-subpath / no-`..` checks.
- Adds BUI i18n keys (`comp:BAIVFolderMountConfigInput`) for `en` and `ko`.

## Notes
- Standalone component only. Wiring it into session creation / Deployment revision pages is a follow-up (out of scope for this stack, per the spec).

## Verification (`scripts/verify.sh`)
- Relay / Lint / Format: **PASS**
- `backend.ai-ui` package `tsc --noEmit`: **0 errors**
- The single TypeScript failure reported by `verify.sh` is **pre-existing on `main`** (unrelated Relay input-type mismatch in `react/src/pages/AdminDeploymentPresetSettingPage.tsx`, `service.port: number | null` vs `number`) and is not introduced by this PR.

## Test plan
- [ ] Selecting multiple vfolders adds a row per folder with alias + subpath inputs
- [ ] Remove (✕) clears the row and its alias/subpath values
- [ ] Invalid alias format, overlapping mount paths, and invalid subpath show inline errors
- [ ] Empty mount destination shows the effective default path as helper text

### DIFF
--- a/packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
+++ b/packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
@@ -1,0 +1,446 @@
+import { useBAIi18n } from '../../hooks/useBAIi18n';
+import BAIFlex from '../BAIFlex';
+import BAIQuestionIconWithTooltip from '../BAIQuestionIconWithTooltip';
+import BAIText from '../BAIText';
+import BAIVFolderSelect from './BAIVFolderSelect';
+import { useControllableValue } from 'ahooks';
+import { Button, Form, Input, Skeleton, Tag, Tooltip, theme } from 'antd';
+import * as _ from 'lodash-es';
+import { XIcon } from 'lucide-react';
+import React, { Suspense } from 'react';
+
+/**
+ * A single vfolder mount configuration emitted by BAIVFolderMountConfigInput.
+ *
+ * - `vfolderId` is the vfolder's **UUID** (`row_id`), so consumers can forward
+ *   it to mount mutation inputs without further conversion.
+ * - `subpath` is the mount **source**: which subfolder inside the vfolder to
+ *   mount. Empty means the vfolder root.
+ * - `mountDestination` is the **raw alias** the user typed, stored verbatim so
+ *   the input box never transforms text mid-edit: `''` mounts at the default
+ *   `${aliasBasePath}${name}`, a relative segment like `data` resolves to
+ *   `${aliasBasePath}data`, and an absolute path like `/data` is used as-is.
+ *   Resolve it to the full container path with {@link inputToMountDestination}.
+ */
+export interface VFolderMountConfigValue {
+  vfolderId: string;
+  name?: string;
+  mountDestination?: string;
+  subpath?: string;
+}
+
+export interface BAIVFolderMountConfigInputProps {
+  value?: VFolderMountConfigValue[];
+  defaultValue?: VFolderMountConfigValue[];
+  onChange?: (value: VFolderMountConfigValue[]) => void;
+  currentProjectId?: string;
+  filter?: string;
+  disabled?: boolean;
+  /** Base path prepended to a relative alias input (mirrors VFolderTable). */
+  aliasBasePath?: string;
+  /**
+   * Names of folders that are auto-mounted (dotfile folders). Their default
+   * mount paths (`${aliasBasePath}${name}`) are added to the overlap set so a
+   * user alias colliding with an auto-mounted folder is flagged — mirrors
+   * VFolderTable's `FolderAliasOverlappingToAutoMount` check. Also shown as a
+   * read-only tag list at the bottom of the component.
+   */
+  autoMountedFolderNames?: string[];
+}
+
+// Mirrors the alias validation used by the legacy VFolderTable mount UI.
+export const vFolderAliasNameRegExp = /^[a-zA-Z0-9_/.-]*$/;
+
+const DEFAULT_ALIAS_BASE_PATH = '/home/work/';
+
+/**
+ * Convert a user-entered alias input into the resolved mount destination,
+ * following the same rule as VFolderTable's `inputToAliasPath`:
+ * - empty input        -> `${basePath}${name}`
+ * - input starting `/` -> used as-is (absolute path)
+ * - otherwise          -> `${basePath}${input}` (relative to the base path)
+ */
+export const inputToMountDestination = (
+  name: string,
+  input: string | undefined,
+  basePath: string,
+) => {
+  const trimmed = input?.trim();
+  if (!trimmed) return `${basePath}${name}`;
+  if (trimmed.startsWith('/')) return trimmed;
+  return `${basePath}${trimmed}`;
+};
+
+// Subpath must be a relative path that does not escape the vfolder.
+const isSubpathInvalid = (subpath?: string) => {
+  const trimmed = subpath?.trim();
+  if (!trimmed) return false;
+  if (trimmed.startsWith('/')) return true;
+  return trimmed.split('/').some((segment) => segment === '..');
+};
+
+export interface VFolderMountConfigStatusOptions {
+  /** Base path prepended to a relative alias input (mirrors VFolderTable). */
+  aliasBasePath?: string;
+  /** Names of auto-mounted folders, included in the overlap check. */
+  autoMountedFolderNames?: string[];
+}
+
+export interface VFolderMountConfigEntryStatus {
+  /** The resolved absolute mount path for the entry (for display). */
+  mountDestination: string;
+  /** Alias error, if any: a bad path format or a colliding mount path. */
+  aliasError?: 'invalidFormat' | 'overlapping';
+  /** Set when the subpath is absolute or escapes the vfolder via `..`. */
+  subpathError?: boolean;
+}
+
+/**
+ * Compute, per entry, its resolved mount destination and any alias/subpath
+ * error — the single source of truth behind the component's inline feedback.
+ * Exported so a consumer can gate a form on validity (see
+ * {@link isVFolderMountConfigValid}) or translate the error kinds itself.
+ */
+export const getVFolderMountConfigStatuses = (
+  value: VFolderMountConfigValue[] | undefined,
+  options?: VFolderMountConfigStatusOptions,
+): Record<string, VFolderMountConfigEntryStatus> => {
+  const basePath = options?.aliasBasePath ?? DEFAULT_ALIAS_BASE_PATH;
+  const entries = value ?? [];
+
+  const mountDestinationByVFolderId: Record<string, string> = {};
+  entries.forEach((entry) => {
+    mountDestinationByVFolderId[entry.vfolderId] = inputToMountDestination(
+      entry.name || entry.vfolderId,
+      entry.mountDestination,
+      basePath,
+    );
+  });
+  // Auto-mounted folders occupy their default mount path; include them so a
+  // user alias colliding with an auto-mounted folder counts as an overlap.
+  const autoMountDestinations = (options?.autoMountedFolderNames ?? []).map(
+    (n) => inputToMountDestination(n, '', basePath),
+  );
+  const destinationCounts = _.countBy([
+    ...Object.values(mountDestinationByVFolderId),
+    ...autoMountDestinations,
+  ]);
+
+  const statuses: Record<string, VFolderMountConfigEntryStatus> = {};
+  entries.forEach((entry) => {
+    const mountDestination = mountDestinationByVFolderId[entry.vfolderId];
+    statuses[entry.vfolderId] = {
+      mountDestination,
+      aliasError: !vFolderAliasNameRegExp.test(entry.mountDestination ?? '')
+        ? 'invalidFormat'
+        : destinationCounts[mountDestination] > 1
+          ? 'overlapping'
+          : undefined,
+      subpathError: isSubpathInvalid(entry.subpath),
+    };
+  });
+  return statuses;
+};
+
+/** True when every entry's alias and subpath are valid. */
+export const isVFolderMountConfigValid = (
+  value: VFolderMountConfigValue[] | undefined,
+  options?: VFolderMountConfigStatusOptions,
+): boolean =>
+  Object.values(getVFolderMountConfigStatuses(value, options)).every(
+    (status) => !status.aliasError && !status.subpathError,
+  );
+
+/**
+ * Reusable, schema-agnostic input for configuring vfolder mounts.
+ *
+ * Users pick vfolders with {@link BAIVFolderSelect} (in `row_id` mode, so the
+ * value is the vfolder UUID); each selected folder appears as a row below the
+ * select where its mount destination (alias) and an optional subpath can be
+ * edited. The alias input follows VFolderTable's rule (relative inputs are
+ * prefixed with `aliasBasePath`, absolute inputs are used as-is); the emitted
+ * `mountDestination` stores that raw alias verbatim, which the consumer
+ * resolves to the full path with {@link inputToMountDestination}. The component
+ * is controlled and emits a single `VFolderMountConfigValue[]` value.
+ *
+ * The inline per-row errors are advisory UX only. To gate a form on validity,
+ * wrap the component in one named `Form.Item` and call
+ * {@link isVFolderMountConfigValid} from a `rules` validator so
+ * `form.validateFields()` rejects on invalid input:
+ *
+ * ```tsx
+ * <Form.Item
+ *   name="mounts"
+ *   rules={[
+ *     {
+ *       validator: (_rule, value) =>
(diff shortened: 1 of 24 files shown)


nowgnuesLee:
* **#7933** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7933?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#7932** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7932?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7932?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#7931** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7931?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7932).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
```diff
@@ -0,0 +1,407 @@
...
+  aliasBasePath?: string;
+}
+
+// Mirrors the alias validation used by the legacy VFolderTable mount UI.
+export const vFolderAliasNameRegExp = /^[a-zA-Z0-9_/.-]*$/;
+
+const DEFAULT_ALIAS_BASE_PATH = '/home/work/';
+
+const resolveName = (entry: VFolderMountConfigValue) =>
+  entry.name || toLocalId(entry.vfolderId);
```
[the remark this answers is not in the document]
nowgnuesLee: No longer applies after the latest revision: the component now uses BAIVFolderSelect's `valuePropName="row_id"`, so `vfolderId` is already the UUID and `resolveName` is `entry.name || entry.vfolderId` — the `toLocalId`/`atob` path was removed entirely.

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
```diff
@@ -0,0 +1,407 @@
...
+              nextIds.map((id) => {
+                const existing = mountConfigs.find(
+                  (entry) => entry.vfolderId === id,
+                );
+                if (existing) {
+                  return nameByValue[id] && !existing.name
+                    ? { ...existing, name: nameByValue[id] }
+                    : existing;
+                }
+                const name = nameByValue[id] ?? toLocalId(id);
```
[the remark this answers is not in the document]
nowgnuesLee: No longer applies: in `row_id` mode `onChange` receives vfolder UUIDs directly. The new-entry fallback is now `resolvedName ?? id` (no `toLocalId`), so there is no base64 decode to throw.

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
```diff
@@ -0,0 +1,407 @@
...
+                  <XIcon
+                    size={16}
+                    style={{
+                      cursor: disabled ? 'not-allowed' : 'pointer',
+                      color: token.colorTextQuaternary,
+                      marginTop: token.marginXXS,
+                      flexShrink: 0,
+                    }}
+                    onClick={() => !disabled && removeEntry(entry.vfolderId)}
+                  />
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed: the remove control is now an antd `Button type="text"` with an `aria-label`, so it is keyboard-focusable and has an accessible name.

agatha197 (changes requested):
### Review — actionable items

This is an additive BUI primitive with no consumer wired yet, so none of these block today. Listing what should be fixed before the first consumer wires it in.

#### Correctness

- **Single source of truth for `mountDestination`.** Display/validation recompute the destination every render (`destinationByVFolderId`, ~L197-205 / L307-313), but the *emitted* value uses the stored `entry.mountDestination` written across four handlers. These can diverge. Please settle on one — emit the recomputed value, or validate the stored one.
- **`onChange` updates `name` without recomputing `mountDestination`** (~L267-270). The existing-entry branch returns `{ ...existing, name: resolvedName }` while leaving `mountDestination` at `${basePath}${uuid}`. Because `onResolvedNamesChange` is guarded by `!entry.name`, the UI self-corrects but the emitted value can still carry the UUID-based path. Recompute via `inputToMountDestination(...)` in that branch.
- **Alias inputs desync on external value re-drive** (~L133-160). `aliasInputs` re-seeds only via the `selectedIdsKey` (id-set) effect and preserves existing ids verbatim. If a controlled parent re-drives `value` with the same vfolderId set but different `mountDestination`s (reset/discard, async `setFieldsValue`), the local alias text goes stale. Reconcile against a destinations digest, re-deriving `next[id]` when the stored `mountDestination` no longer matches the local alias (mirrors VFolderTable's `aliasMap` reconcile).

#### Parity / contract

- **Missing auto-mount collision check** (`destinationCounts`, ~L193-198). Only within-selection overlap is detected. The legacy VFolderTable this component claims to mirror also rejects collisions with auto-mounted (dotfile) folders (`FolderAliasOverlappingToAutoMount`). Either accept the auto-mount folder names as a prop and include them in the overlap set, or document the limitation in the JSDoc.
- **Validation contract is advisory-only.** The overlap/alias validators are module-private and the emitted `VFolderMountConfigValue[]` exposes no validity surface, so the JSDoc "wrap in one `Form.Item`" guidance is misleading (that wrapper gets no `rules`). Please either export the validators or expose a `status` / `onValidityChange` surface, and correct the JSDoc accordingly.

#### i18n

- **19 locales were seeded with verbatim English** (`locale/{ja,th,de,…}.json`, 11 keys × 19 = 209 English values). This diverges from the full-translation precedent (FR-3032). Please run `fw:i18n` before merge, or scope this PR to `en` + `ko` and let the translation flow backfill the rest — avoid committing English into non-`en` files.

#### UI / layout

- **Remove (✕) button vertical alignment.** The row uses `align="start"`, so when the alias column shows its helper text (the resolved `/home/work/...` path) the row grows taller and the ✕ button sticks to the top instead of lining up with the inputs. Please vertically center the ✕ with the input row (e.g. align it to the input height rather than the full, helper-text-inclusive row height).

  remove button alignment

#### Nit

- **Use named gap tokens.** `gap={token.sizeXXS}` (L208, L289, L320) should be `gap="xxs"` to match the BUI convention (the file already mixes both).

nowgnuesLee:
> ### Review — actionable items
> 
> This is an additive BUI primitive with no consumer wired yet, so none of these block today. Listing what should be fixed before the first consumer wires it in.
> 
> #### Correctness
> 
> - **Single source of truth for `mountDestination`.** Display/validation recompute the destination every render (`destinationByVFolderId`, ~L197-205 / L307-313), but the *emitted* value uses the stored `entry.mountDestination` written across four handlers. These can diverge. Please settle on one — emit the recomputed value, or validate the stored one.
> - **`onChange` updates `name` without recomputing `mountDestination`** (~L267-270). The existing-entry branch returns `{ ...existing, name: resolvedName }` while leaving `mountDestination` at `${basePath}${uuid}`. Because `onResolvedNamesChange` is guarded by `!entry.name`, the UI self-corrects but the emitted value can still carry the UUID-based path. Recompute via `inputToMountDestination(...)` in that branch.
> - **Alias inputs desync on external value re-drive** (~L133-160). `aliasInputs` re-seeds only via the `selectedIdsKey` (id-set) effect and preserves existing ids verbatim. If a controlled parent re-drives `value` with the same vfolderId set but different `mountDestination`s (reset/discard, async `setFieldsValue`), the local alias text goes stale. Reconcile against a destinations digest, re-deriving `next[id]` when the stored `mountDestination` no longer matches the local alias (mirrors VFolderTable's `aliasMap` reconcile).
> 
> #### Parity / contract
> 
> - **Missing auto-mount collision check** (`destinationCounts`, ~L193-198). Only within-selection overlap is detected. The legacy VFolderTable this component claims to mirror also rejects collisions with auto-mounted (dotfile) folders (`FolderAliasOverlappingToAutoMount`). Either accept the auto-mount folder names as a prop and include them in the overlap set, or document the limitation in the JSDoc.
> - **Validation contract is advisory-only.** The overlap/alias validators are module-private and the emitted `VFolderMountConfigValue[]` exposes no validity surface, so the JSDoc "wrap in one `Form.Item`" guidance is misleading (that wrapper gets no `rules`). Please either export the validators or expose a `status` / `onValidityChange` surface, and correct the JSDoc accordingly.
> 
> #### i18n
> 
> - **19 locales were seeded with verbatim English** (`locale/{ja,th,de,…}.json`, 11 keys × 19 = 209 English values). This diverges from the full-translation precedent (FR-3032). Please run `fw:i18n` before merge, or scope this PR to `en` + `ko` and let the translation flow backfill the rest — avoid committing English into non-`en` files.
> 
> #### UI / layout
> 
> - **Remove (✕) button vertical alignment.** The row uses `align="start"`, so when the alias column shows its helper text (the resolved `/home/work/...` path) the row grows taller and the ✕ button sticks to the top instead of lining up with the inputs. Please vertically center the ✕ with the input row (e.g. align it to the input height rather than the full, helper-text-inclusive row height).
> 
>   remove button alignment
> 
> #### Nit
> 
> - **Use named gap tokens.** `gap={token.sizeXXS}` (L208, L289, L320) should be `gap="xxs"` to match the BUI convention (the file already mixes both).
> 

​I applied. Thanks!

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
```diff
@@ -0,0 +1,440 @@
...
+    aliasBasePath,
+    autoMountedFolderNames,
+  });
+
+  const columnLabel = (label: string, tooltip: string) => (
+    <BAIFlex gap="xxs" align="center" style={{ flex: 1 }}>
+      <BAIText type="secondary">{label}</BAIText>
+      <Tooltip title={tooltip}>
+        <CircleHelpIcon size={14} color={token.colorTextQuaternary} />
+      </Tooltip>
```
agatha197: This hand-rolls the help-icon-with-tooltip widget. Two suggestions:

**1. Reuse `BAIQuestionIconWithTooltip`.** It already exists in `backend.ai-ui` and is the app-wide standard for a standalone help tooltip (used in `FolderCreateModal`, `FolderCreateModalV2`, `BulkCreateUserFromCSVModal`, `OverlayNetworkSettingModal`, …). It wraps `Tooltip` + antd `QuestionCircleOutlined` with `cursor: 'help'` and `token.colorTextTertiary`. The current code uses a lucide `CircleHelpIcon` and `token.colorTextQuaternary`, so this help icon ends up visually different (different glyph + color) from every other help tooltip in the app.

```tsx
const columnLabel = (label: string, tooltip: string) => (
  <BAIFlex gap="xxs" align="center" style={{ flex: 1 }}>
    <BAIText type="secondary">{label}</BAIText>
    <BAIQuestionIconWithTooltip title={tooltip} />
  </BAIFlex>
);
```

(For per-field labels the equivalent is `Form.Item`'s `tooltip` prop, which renders the same antd icon — but since these are shared *column headers* rather than field labels, the standalone `BAIQuestionIconWithTooltip` is the right fit here.)

**2. Avoid the hardcoded `size={14}` px.** Icon sizes in this repo are derived from theme tokens (`size={token.fontSizeSM}` / `size={token.fontSize}`, e.g. `DiagnosticResultList`, `DeploymentRevisionHistoryTab`, `ModelStoreListPageV2`). Switching to `BAIQuestionIconWithTooltip` removes the magic number entirely; if a custom icon is kept for some reason, please size it via `token.fontSizeSM` instead of a literal `14`.

agatha197 (commented):
A couple of consistency nits on the `columnLabel` header helper — see the inline comment. Both are about reusing the existing shared help-icon component and dropping the hardcoded pixel size.

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
```diff
@@ -0,0 +1,440 @@
...
+                  title={t('comp:BAIVFolderMountConfigInput.RemoveFolder')}
+                >
+                  <Button
+                    type="text"
+                    size="small"
+                    disabled={disabled}
+                    aria-label={t(
+                      'comp:BAIVFolderMountConfigInput.RemoveFolder',
+                    )}
+                    icon={<XIcon size={16} color={token.colorTextQuaternary} />}
```
agatha197: `size={16}` is a hardcoded px value. The size + color here match the predecessor `react/src/components/VFolderMountFormItem.tsx`, so it's consistent with that component — but the repo convention is to size lucide icons from theme tokens (e.g. `ChatPage.tsx` uses `<TrashIcon size={token.size} />`, and `token.size` is 16 by default). Please use `size={token.size}` to keep the value while dropping the magic number.

Note this `16` is also coupled to two other spots that must stay in sync: the header spacer `width: 16` (line 300, whose comment explicitly says "aligning the header with the row's 16px remove-button column") and the button cell. Deriving all of them from `token.size` (or a shared const) removes that hidden coupling. The `color={token.colorTextQuaternary}` is fine as-is — it matches the predecessor.

agatha197 (commented):
One more nit on the remove-icon size — see inline.

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
nowgnuesLee: Done. Replaced the hand-rolled `Tooltip` + lucide `CircleHelpIcon` with the shared `BAIQuestionIconWithTooltip`, so the column help icon now matches the app-wide standard glyph (`QuestionCircleOutlined`) and color (`colorTextTertiary`). The hardcoded `size={14}` is gone, and the now-unused `CircleHelpIcon` import was removed.

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
nowgnuesLee: Done. Changed the remove icon to `size={token.size}` (keeps 16 while dropping the magic number), and keyed the header spacer to `width: token.size` as well so the two stay in sync from a single source — the hidden coupling you flagged is now derived, not duplicated. `color={token.colorTextQuaternary}` left as-is per your note.

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
```diff
@@ -0,0 +1,445 @@
...
+              {t('comp:BAIVFolderMountConfigInput.Name')}
+            </BAIText>
+            {columnLabel(
+              t('comp:BAIVFolderMountConfigInput.PathAndAlias'),
+              t('comp:BAIVFolderMountConfigInput.PathAndAliasTooltip'),
+            )}
+            {columnLabel(
+              t('comp:BAIVFolderMountConfigInput.Subpath'),
+              t('comp:BAIVFolderMountConfigInput.SubpathTooltip'),
+            )}
```
ironAiken2: I don’t think there’s any need to split this into a function.

ironAiken2 (changes requested):

packages/backend.ai-ui/src/components/fragments/BAIVFolderMountConfigInput.tsx
nowgnuesLee: Done. Inlined the two column headers directly into the return JSX and removed the `columnLabel` helper — agreed it reads better without the extra indirection for just two call sites.

ironAiken2 (approved):
LGTM

agatha197 (approved):
LGTM

PR #7977, opened 2026-06-18
fix(FR-3090): robustly clean up E2E vfolders so the test server is not left dirty

### LINKED ISSUE #7819: vfolder E2E tests do not clean up created folders (test server left dirty)
## Problem

vfolder E2E test scenarios leave the folders they create on the shared test server instead of removing them. After a test run, many `e2e-test-*` virtual folders remain in **READY** state, polluting the environment and risking flaky/duplicate-name failures on subsequent runs.

Observed leftovers (example run):

- `e2e-test-project-type-folder-*`
- `e2e-test-consecutive-1-*`, `e2e-test-consecutive-2-*`
- `e2e-test-file-create-*`
- `e2e-test-user-type-folder-*`
- `e2e-test-folder-layout-*`
- `e2e-test-folder-rw-*`

## Requirement

Every E2E test case must **completely delete all entities it creates** (here: virtual folders — move to trash AND delete forever) during teardown, so the test server is left clean regardless of whether the test passed or failed.

## Current state / gaps

- Cleanup runs in `test.afterAll` using `moveToTrashAndVerify` + `deleteForeverAndVerifyFromTrash` (`e2e/utils/test-util.ts`).
- Failures are silently swallowed (`console.log('Could not delete ${name}, it may not exist')` / `/* ignore cleanup errors */`), so when deletion fails the folder is left behind with no test failure surfaced.
- As a result, partial/failed cleanups accumulate folders over time.

## Expected

- Make teardown robust so created folders are guaranteed to be permanently removed (consider `afterEach` per-folder cleanup in addition to `afterAll`).
- Do not silently ignore cleanup failures — surface them (fail or warn loudly) so leftover entities are detected.
- Consider a global sweep that deletes orphaned `e2e-test-*` folders before/after the suite.
- Apply consistently across all `e2e/vfolder/*.spec.ts` cases (and use the same pattern for other entity-creating suites).

## Notes

- Found during 26.4.8 release E2E testing.
- Cleanup deletions on the current test backend can also return `403 NotEnoughPermission (soft-delete)` for the test account; robust teardown should account for / report this rather than swallow it.

JIRA Issue: FR-3090

ironAiken2:
Resolves #7819 (FR-3090)

## Summary

E2E vfolder tests left `e2e-test-*` / `e2e-*` orphans on the shared test server, polluting it over time. There were two root causes:

1. Cleanup hooks wrapped the delete flow in a `try/catch` that only logged on failure, so a failed deletion silently left an orphan.
2. **More importantly, cleanup could hang for the full 180s test budget.** A transiently pointer-blocked click (e.g. the Active/Trash tab briefly covered by an antd portal during cleanup) retried until the test timeout because the suite set no `actionTimeout`. When a cleanup test or `afterAll`/`afterEach` hook times out, the orphan is left behind — regardless of whether the test itself passed.

### Robust per-test cleanup
- Add a never-throwing `cleanupVFolderSafely(page, folderName, dataPath?)` helper in `e2e/utils/cleanup-util.ts`:
  1. normal trash + delete-forever flow;
  2. on failure, retry delete-forever (the folder may already be in Trash);
  3. fall back to `sweepVFolders()` to remove an Active orphan by name.

  It only warns (never throws), so a cleanup hiccup cannot fail an otherwise-passing test, but the orphan is actually removed.
- Convert the `afterAll` / `afterEach` cleanup hooks across the vfolder specs to use it.
- **In-test delete assertions are intentionally left untouched** — `vfolder-crud` (create → trash → restore → delete-forever), `vfolder-consecutive-deletion`, and the in-test deletes in `vfolder-explorer-modal` exercise the delete flow itself.

### Global teardown safety net
- `e2e/global-cleanup.teardown.ts` (wired as chromium's `teardown`) sweeps leftover `e2e-*` vfolders as the user (`/data`) and as the admin (`/admin-data`, where project folders are deletable), plus services, once after the suite — so the server is left clean even when a per-test hook is aborted by a timeout.
- `sweepVFolders` now skips rows whose move-to-trash button is **disabled** (project folders on `/data`, folders the account doesn't own on `/admin-data`) via `firstActionableVFolderName`, and **continues past a failing folder instead of breaking**, so one stuck orphan no longer strands the rest of the sweep.
- Raise the teardown timeout to 300s so it can drain a backlog of leftover folders.

### The keystone fix — bound every action
- **`playwright.config.ts`: set a global `actionTimeout: 30000`.** Every action now fails fast with a catchable error instead of consuming the whole 180s test budget. 30s is 3× the largest explicit per-action timeout in the suite (10s), so it never cuts a legitimately slow action short. This removes the 180s hang in **both** the global teardown sweep **and** the per-spec cleanup hooks (which run regardless of how the suite is invoked).
- Guard `moveToTrashAndVerify` with a bounded `toBeEnabled` check before the click so a disabled button fails fast instead of retrying to the test timeout.

## Verification (live cluster)

- The hardened sweep drained **5 leftover orphans** from a prior dirty run; a second cleanup pass converged to `removed=0` (server clean).
- A full **`--workers=1`** vfolder run is green; the previously-"failing" specs (consecutive-deletion, explorer-modal, `file-*` uploads, permissions `RO↔RW` which are `test.fixme`) all pass / skip correctly in isolation. Failures seen at higher local parallelism are **load-induced flakiness** against the shared backend (a different random subset fails each run), not regressions.

## Running the tests

> [!IMPORTANT]
> **Run vfolder e2e with a single worker — the CI standard.** `playwright.config.ts` already forces `workers: 1` + `retries: 2` whenever `CI` is set, so pre-merge GitHub Actions runs are serial and retried automatically (and the `actionTimeout` fix means a hung action can no longer stall the whole CI run). Locally, the default parallelism (~one worker per 2 cores) against the **shared** backend causes load-induced flakiness, so run with **`--workers=1`** (matching CI). If you want it faster while staying reliable, use `--workers=2 --retries=2` — the retries recover transient load-flakes and the bounded `actionTimeout` makes each retry cheap.

### DIFF
--- a/e2e/global-cleanup.teardown.ts
+++ b/e2e/global-cleanup.teardown.ts
@@ -1,0 +1,78 @@
+/**
+ * Global cleanup teardown (FR-3090).
+ *
+ * Runs once after the whole suite — wired via the `chromium` project's
+ * `teardown: 'cleanup'` in `playwright.config.ts` — regardless of whether the
+ * tests passed or failed. Per-test `afterEach` cleanup is best-effort and can be
+ * aborted when a test exhausts its 180s budget (the cleanup then stops mid-way,
+ * leaving an `e2e-*` orphan behind). This teardown is the safety net that
+ * guarantees the shared test server is left clean.
+ *
+ * It is intentionally tolerant: every sweep is best-effort and never fails the
+ * run — a cleanup hiccup must not turn a green suite red. It sweeps as the user
+ * (personal folders on /data) and as the admin (/admin-data), because different
+ * specs create folders under different accounts:
+ *   - vfolder-crud / file-* / type-selection (user) → personal folders on /data
+ *   - type-selection project (admin) → project folders, created on /project-data
+ *     but only DELETABLE from /admin-data (VFolderNodesV2 on /project-data does
+ *     not expose trash/delete). /admin-data also lists every other admin-visible
+ *     folder, so it doubles as a catch-all.
+ *
+ * The pattern is the broad `/e2e-/i` (covering `e2e-test-*`, `e2e-mod-*`, the
+ * dot-prefixed auto-mount folder, etc.). Broad matching is safe HERE — unlike in
+ * a per-test `sweepVFolders` call — because the teardown runs after every other
+ * project has finished, so no parallel test owns a live `e2e-*` folder.
+ */
+import { sweepServices, sweepVFolders } from './utils/cleanup-util';
+import { loginAsAdmin, loginAsUser } from './utils/test-util';
+import { Page, test } from '@playwright/test';
+
+// Matches every e2e vfolder naming scheme used across the specs.
+const E2E_VFOLDER_PATTERN = /e2e-/i;
+
+/** Run a best-effort sweep, swallowing any error so teardown never fails. */
+async function safeSweep(
+  label: string,
+  fn: (page: Page) => Promise<unknown>,
+  page: Page,
+): Promise<void> {
+  try {
+    await fn(page);
+  } catch (error) {
+    console.warn(`[global-cleanup] ${label} sweep failed (ignored):`, error);
+  }
+}
+
+test.describe('Global e2e cleanup', () => {
+  // The sweep may have to trash + delete-forever several leftover folders, each
+  // a multi-step navigate/filter/confirm flow, and a stuck folder now fails its
+  // action at the 30s actionTimeout before being skipped. Give the cleanup more
+  // room than the default 180s so it can drain a backlog instead of timing out
+  // mid-sweep and leaving the rest behind.
+  test.describe.configure({ timeout: 300_000 });
+
+  test('sweep leftover e2e vfolders (user)', async ({ page, request }) => {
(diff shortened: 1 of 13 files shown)


ironAiken2:
* **#7977** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/7977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=7977).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

nowgnuesLee (approved):
LGTM

nowgnuesLee (dismissed):
Please resolve the conficts

agatha197 (approved):
LGTM

PR #8062, opened 2026-06-27
feat(FR-3222): migrate User Resource Policy page to Strawberry V2 GraphQL

### LINKED ISSUE #8061: Migrate User Resource Policy page to Strawberry V2 GraphQL query
## Overview

Part of FR-3183 (Remove legacy V1 GraphQL nodes and complete Strawberry V2 migration).

Migrate the User Resource Policy management page (`react/src/components/UserResourcePolicyList.tsx` and related components) from the legacy V1 `user_resource_policies` query to the Strawberry V2 connection (`adminUserResourcePoliciesV2` / `UserResourcePolicyV2`).

## Scope

- Replace the legacy `user_resource_policies` query with the V2 connection field
- Update create/update flows to use the V2 mutations (`adminCreateUserResourcePolicyV2`, `adminUpdateUserResourcePolicyV2`)
- Migrate Relay fragments to the `UserResourcePolicyV2` type
- Follow offset (`limit`/`offset`) pagination mode for the V2 connection (see graphql-pagination rule)

## Acceptance Criteria

- No remaining references to the legacy V1 `user_resource_policies` query in the User Resource Policy page
- List, create, and update flows use Strawberry V2 (`*V2`) fields
- `scripts/verify.sh` passes (Relay, Lint, Format, TypeScript)

JIRA Issue: FR-3222

nowgnuesLee:
Resolves #8061 (FR-3222)

## Summary

Migrates the **User Resource Policy** tab on the Resource Policy page from the legacy V1 `user_resource_policies` query to the Strawberry **V2** GraphQL API, as part of FR-3183 (legacy V1 `_nodes` removal / V2 migration).

New components are added alongside the legacy ones — the legacy files are **kept on disk, now unused, as a safety backup** (per request):

- **`BAIUserResourcePolicyV2Table`** (in `backend.ai-ui`) — presentational table over a `UserResourcePolicyV2` plural fragment. The name column uses `BAINameActionCell` with per-row Edit / Delete actions.
- **`UserResourcePolicyV2.tsx`** — query orchestrator reading a preloaded `adminUserResourcePoliciesV2` query; owns filter / sort / pagination (`limit`/`offset`) / refresh, the create-edit modal, and the delete flow (`adminDeleteUserResourcePolicyV2`).
- **`UserResourcePolicyV2SettingModal.tsx`** — create/edit via `adminCreateUserResourcePolicyV2` / `adminUpdateUserResourcePolicyV2`.

### Lazy load on tab activation

`ResourcePolicyPage` owns `useQueryLoader` and loads the V2 query only when the `user` tab is active (`useEffectEvent` + `useEffect([curTabKey])`), mirroring the ScopedAuditLog / FR-3215 login-sessions pattern.

**Version branching (manager-gated).** The V2 `BAIGraphQLPropertyFilter` emits the AND/OR/NOT sub-filter combinators, which `UserResourcePolicyV2Filter` only gained in **manager 26.7.0** (schema: "Added in 26.7.0"). The page therefore gates on the `sub-filter` client feature: when supported it renders `UserResourcePolicyV2`, and on older managers (< 26.7.0) it falls back to the legacy `UserResourcePolicyList` so the tab keeps working without sending arguments the manager would reject. The legacy fallback is retained for that compatibility window rather than removed outright.

## Table components moved into `backend.ai-ui`

The presentational tables are consolidated into the shared BUI package, following the `BAIAdminUserV2Table` precedent:

| Before (host) | After (BUI) |
| --- | --- |
| `UserResourcePolicyV2Table` | `BAIUserResourcePolicyV2Table` |
| `LoginSessionTable` | `BAILoginSessionTable` |
| `LoginHistoryTable` | `BAILoginHistoryTable` |

- Switched `useTranslation` → `useBAIi18n` (react-i18next direct imports are ESLint-blocked inside BUI per FR-2986); table strings moved into `comp:BAI*Table.*` namespaces in `packages/backend.ai-ui/src/locale/*.json` (all 21 languages).
- Fragments / props renamed with the `BAI` prefix (`BAI*TableFragment`, `BAI*TableProps`); components exported from the package index.
- Host orchestrators (`UserResourcePolicyV2`, `LoginSession`, `LoginHistory`) now consume them from `backend.ai-ui`. No behavior change — query orchestration (filter, pagination, sort, row actions) stays in the host.
- Host i18n keys orphaned by the move (`loginHistory.FailReason`, `loginSession.InvalidatedAt`, `loginSession.User`) removed from all locales; keys still used by the host modal / filter bars are kept.

## Schema notes

- `max_quota_scope_size` (Int) → `maxQuotaScopeSize` (`BinarySizeInfo { value, display }`); unlimited still rendered as ∞ and submitted as `BinarySizeInput { expr: '-1' }`.
- snake_case → camelCase field renames throughout.
- Pagination uses **`limit` + `offset`** only (offset mode), per the V2 connection pagination rule.

## Newly surfaced field

- `maxConcurrentLogins` is now surfaced as both a table column (rendered as ∞ when null) and a setting-modal form field. New i18n key `MaxConcurrentLogins` added across all locales.

## Units

- `maxQuotaScopeSize` is **decimal GB end-to-end**: the modal converts with `GBToBytes` (×10^9) and the list renders with `convertToDecimalUnit(value, 'auto')` — what the admin types is what the list shows.

## Out of scope

- Legacy `UserResourcePolicyList.tsx` / `UserResourcePolicySettingModal.tsx` left in place — still used as the < 26.7.0 fallback render (see "Version branching" above).

## Needs verification (backend)

- Confirm the manager accepts `BinarySizeInput { expr: "-1" }` as the "unlimited" sentinel for `maxQuotaScopeSize` on create/update (legacy passed the Int `-1`).

## Verification

`bash scripts/verify.sh`: **Relay ✅ / Lint ✅ / Format ✅ / TypeScript ✅**.

> The `Vite warmup paths` check reports FAIL, but this is **pre-existing on `main`** and unrelated to this PR — `react/vite.config.ts` is untouched. The check's awk window over-captures string literals outside the `warmup` block (e.g. `sw.js`, `buffer`, `**/*.map`), which are not real warmup entries.

### DIFF
--- a/packages/backend.ai-client/src/client.ts
+++ b/packages/backend.ai-client/src/client.ts
@@ -933,6 +933,19 @@ export class Client {
       // TODO(FR-3139): simplify to '26.4.4' once rc builds are out of use.
       this._features['model-runtime-variant-preset-values'] = true;
     }
+    if (this.isManagerVersionCompatibleWith('26.7.0')) {
+      // Strawberry V2 filter inputs gained the AND/OR/NOT sub-filter
+      // combinators (schema: "Added in 26.7.0"), which
+      // `BAIGraphQLPropertyFilter` emits when conditions are combined. Older
+      // managers reject those arguments.
+      this._features['sub-filter'] = true;
+    }
+    if (this.isManagerVersionCompatibleWith('26.8.0')) {
+      // BinarySizeInfo replaced `value: Int!` with `expr: String!` (exact
+      // bytes as a decimal string, round-trippable into BinarySizeInput.expr)
+      // in the 26.8.0 supergraph. Older managers only serve `value`.
+      this._features['binary-size-expr'] = true;
+    }
   }
 
   /**
--- a/packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
+++ b/packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
@@ -1,13 +1,15 @@
 import { useBAIi18n } from '../hooks/useBAIi18n';
+import BAIButton from './BAIButton';
 import BAIFlex from './BAIFlex';
 import BAISelect from './BAISelect';
-import { CloseCircleOutlined } from '@ant-design/icons';
+import { CloseCircleOutlined, SearchOutlined } from '@ant-design/icons';
 import { useControllableValue } from 'ahooks';
 import {
   AutoComplete,
   Button,
   DatePicker,
   Input,
+  InputNumber,
   Space,
   Tag,
   Tooltip,
@@ -807,6 +809,35 @@ const BAIGraphQLPropertyFilter = <
               t('comp:BAIGraphQLPropertyFilter.SelectDateTime')
             }
           />
+        ) : selectedProperty?.type === 'number' ? (
+          // `number` properties get a numeric-only input so non-numeric text
+          // can never reach `Number(value)` and surface as a `NaN` condition.
+          // Guard the displayed value too: switching from a string property
+          // can carry over non-numeric `search`, whose `Number(...)` is `NaN`,
+          // which antd `InputNumber` cannot render.
+          <>
+            <InputNumber
+              value={
+                search === '' || Number.isNaN(Number(search))
+                  ? null
+                  : Number(search)
+              }
+              onChange={(val) => {
+                setIsValid(true);
+                setSearch(_.isNil(val) ? '' : String(val));
+              }}
+              onPressEnter={() => addCondition(search)}
+              style={{ minWidth: 200 }}
+              placeholder={
+                selectedProperty?.placeholder ??
+                t('comp:BAIPropertyFilter.PlaceHolder')
+              }
+            />
+            <BAIButton
+              icon={<SearchOutlined />}
+              onClick={() => addCondition(search)}
+            />
+          </>
         ) : (
           <Tooltip
             title={isValid || !isFocused ? '' : selectedProperty.rule?.message}
--- a/packages/backend.ai-ui/src/components/BAIUserResourcePolicyV2Table.tsx
+++ b/packages/backend.ai-ui/src/components/BAIUserResourcePolicyV2Table.tsx
@@ -1,0 +1,170 @@
+import {
+  BAIColumnsType,
+  BAIColumnType,
+  BAITable,
+  BAITableProps,
+  convertToDecimalUnit,
+  filterOutEmpty,
+  filterOutNullAndUndefined,
+} from '..';
+import type {
+  BAIUserResourcePolicyV2TableFragment$data,
+  BAIUserResourcePolicyV2TableFragment$key,
+} from '../__generated__/BAIUserResourcePolicyV2TableFragment.graphql';
+import { useBAIi18n } from '../hooks/useBAIi18n';
+import dayjs from 'dayjs';
+import * as _ from 'lodash-es';
+import { graphql, useFragment } from 'react-relay';
+
+export type UserResourcePolicyV2InList = NonNullable<
+  BAIUserResourcePolicyV2TableFragment$data[number]
+>;
+
+const availableUserResourcePolicySorterKeys = [
+  'name',
+  'maxVfolderCount',
+  'maxConcurrentLogins',
+  'maxSessionCountPerModelSession',
+  'maxQuotaScopeSize',
+  'maxCustomizedImageCount',
+  'createdAt',
+] as const;
+
+export const availableUserResourcePolicySorterValues = [
+  ...availableUserResourcePolicySorterKeys,
+  ...availableUserResourcePolicySorterKeys.map((key) => `-${key}` as const),
+] as const;
+
+const isEnableSorter = (key: string) => {
+  return _.includes(availableUserResourcePolicySorterKeys, key);
+};
+
+export interface BAIUserResourcePolicyV2TableProps extends Omit<
+  BAITableProps<UserResourcePolicyV2InList>,
+  'dataSource' | 'columns' | 'onChangeOrder'
+> {
+  userResourcePoliciesFrgmt: BAIUserResourcePolicyV2TableFragment$key;
+  disableSorter?: boolean;
+  customizeColumns?: (
+    baseColumns: BAIColumnsType<UserResourcePolicyV2InList>,
+  ) => BAIColumnsType<UserResourcePolicyV2InList>;
+  onChangeOrder?: (
+    order: (typeof availableUserResourcePolicySorterValues)[number] | null,
+  ) => void;
+}
+
+const BAIUserResourcePolicyV2Table = ({
+  userResourcePoliciesFrgmt,
+  disableSorter,
+  customizeColumns,
+  onChangeOrder,
+  ...tableProps
+}: BAIUserResourcePolicyV2TableProps) => {
+  'use memo';
+  const { t } = useBAIi18n();
+
+  const userResourcePolicies =
+    useFragment<BAIUserResourcePolicyV2TableFragment$key>(
+      graphql`
+        fragment BAIUserResourcePolicyV2TableFragment on UserResourcePolicyV2
+        @relay(plural: true) {
+          id
+          name
+          createdAt
+          maxVfolderCount
+          maxConcurrentLogins
+          maxSessionCountPerModelSession
+          maxQuotaScopeSize {
+            expr
+          }
+          maxCustomizedImageCount
+        }
+      `,
+      userResourcePoliciesFrgmt,
+    );
+
+  const baseColumns = _.map(
+    filterOutEmpty<BAIColumnType<UserResourcePolicyV2InList>>([
+      {
+        title: t('comp:BAIUserResourcePolicyV2Table.Name'),
+        dataIndex: 'name',
+        key: 'name',
+        fixed: 'left',
+        sorter: isEnableSorter('name'),
+      },
+      {
+        title: t('comp:BAIUserResourcePolicyV2Table.MaxVFolderCount'),
+        dataIndex: 'maxVfolderCount',
+        key: 'maxVfolderCount',
+        sorter: isEnableSorter('maxVfolderCount'),
+        render: (text) => (_.toNumber(text) === 0 ? '∞' : text),
+      },
+      {
+        title: t('comp:BAIUserResourcePolicyV2Table.MaxConcurrentLogins'),
+        dataIndex: 'maxConcurrentLogins',
+        key: 'maxConcurrentLogins',
+        sorter: isEnableSorter('maxConcurrentLogins'),
+        render: (text) => (_.isNil(text) ? '∞' : text),
+      },
+      {
+        title: t(
+          'comp:BAIUserResourcePolicyV2Table.MaxSessionCountPerModelSession',
+        ),
+        dataIndex: 'maxSessionCountPerModelSession',
+        key: 'maxSessionCountPerModelSession',
+        sorter: isEnableSorter('maxSessionCountPerModelSession'),
+      },
+      {
+        title: t('comp:BAIUserResourcePolicyV2Table.MaxQuotaScopeSize'),
+        dataIndex: 'maxQuotaScopeSize',
+        key: 'maxQuotaScopeSize',
+        sorter: isEnableSorter('maxQuotaScopeSize'),
+        render: (__, row) =>
+          row.maxQuotaScopeSize.expr === '-1'
+            ? '∞'
+            : (convertToDecimalUnit(row.maxQuotaScopeSize.expr, 'auto')
+                ?.displayValue ?? '-'),
+      },
+      {
+        title: t('comp:BAIUserResourcePolicyV2Table.MaxCustomizedImageCount'),
+        dataIndex: 'maxCustomizedImageCount',
+        key: 'maxCustomizedImageCount',
+        sorter: isEnableSorter('maxCustomizedImageCount'),
+      },
+      {
+        title: t('comp:BAIUserResourcePolicyV2Table.CreatedAt'),
+        dataIndex: 'createdAt',
+        key: 'createdAt',
+        sorter: isEnableSorter('createdAt'),
+        render: (text) => (text ? dayjs(text).format('lll') : '-'),
+      },
+    ]),
+    (column) => {
+      return disableSorter ? _.omit(column, 'sorter') : column;
+    },
+  );
+
+  const allColumns = customizeColumns
+    ? customizeColumns(baseColumns)
+    : baseColumns;
+
+  return (
+    <BAITable
+      resizable
+      rowKey="id"
+      showSorterTooltip={false}
+      dataSource={filterOutNullAndUndefined(userResourcePolicies)}
+      columns={allColumns}
+      scroll={{ x: 'max-content' }}
+      onChangeOrder={(order) => {
+        onChangeOrder?.(
+          (order as (typeof availableUserResourcePolicySorterValues)[number]) ||
+            null,
+        );
+      }}
+      {...tableProps}
+    />
+  );
+};
+
+export default BAIUserResourcePolicyV2Table;
--- a/react/src/components/UserResourcePolicyV2SettingModal.tsx
+++ b/react/src/components/UserResourcePolicyV2SettingModal.tsx
@@ -1,0 +1,314 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+import {
+  CreateUserResourcePolicyInputV2,
+  UserResourcePolicyV2SettingModalCreateMutation,
+} from '../__generated__/UserResourcePolicyV2SettingModalCreateMutation.graphql';
+import { UserResourcePolicyV2SettingModalFragment$key } from '../__generated__/UserResourcePolicyV2SettingModalFragment.graphql';
+import {
+  UpdateUserResourcePolicyInput,
+  UserResourcePolicyV2SettingModalModifyMutation,
+} from '../__generated__/UserResourcePolicyV2SettingModalModifyMutation.graphql';
+import { GBToBytes, bytesToGB } from '../helper';
+import { SIGNED_32BIT_MAX_INT } from '../helper/const-vars';
+import FormItemWithUnlimited from './FormItemWithUnlimited';
+import {
+  Form,
+  Input,
+  Alert,
+  App,
+  InputNumber,
+  theme,
+  FormInstance,
+} from 'antd';
+import { BAIModal, BAIModalProps, BAIFlex } from 'backend.ai-ui';
+import * as _ from 'lodash-es';
+import React, { useRef } from 'react';
+import { useTranslation } from 'react-i18next';
+import { graphql, useFragment, useMutation } from 'react-relay';
+
+interface UserResourcePolicyV2SettingModalProps extends Omit<
+  BAIModalProps,
+  'onOk' | 'onCancel'
+> {
+  userResourcePolicyFrgmt: UserResourcePolicyV2SettingModalFragment$key | null;
+  onOk: () => void;
+  onCancel: () => void;
+}
+
+const UserResourcePolicyV2SettingModal: React.FC<
+  UserResourcePolicyV2SettingModalProps
+> = ({ userResourcePolicyFrgmt = null, onOk, onCancel, ...baiModalProps }) => {
+  'use memo';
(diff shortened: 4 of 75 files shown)


nowgnuesLee:
* **#8150** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8150?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8062** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8062?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8062?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8062).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

packages/backend.ai-client/src/client.ts
```diff
@@ -933,6 +933,12 @@ export class Client {
       // TODO(FR-3139): simplify to '26.4.4' once rc builds are out of use.
       this._features['model-runtime-variant-preset-values'] = true;
     }
+    if (this.isManagerVersionCompatibleWith('26.4.0')) {
+      // Strawberry V2 filter inputs gained the AND/OR/NOT sub-filter
+      // combinators, which `BAIGraphQLPropertyFilter` emits when conditions
+      // are combined. Older managers reject those arguments.
+      this._features['sub-filter'] = true;
+    }
```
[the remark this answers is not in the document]
nowgnuesLee: Good catch — fixed. Gated `sub-filter` on `26.7.0` to match the schema annotation ("Added in 26.7.0") for the AND/OR/NOT combinators, so 26.4.x–26.6.x managers no longer receive arguments they'd reject.

packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
```diff
@@ -745,6 +746,22 @@ const BAIGraphQLPropertyFilter: React.FC<BAIGraphQLPropertyFilterProps> = ({
...
           />
+        ) : selectedProperty?.type === 'number' ? (
+          // `number` properties get a numeric-only input so non-numeric text
+          // can never reach `Number(value)` and surface as a `NaN` condition.
+          <InputNumber
+            value={search === '' ? null : Number(search)}
+            onChange={(val) => {
+              setIsValid(true);
+              setSearch(_.isNil(val) ? '' : String(val));
+            }}
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed. The displayed value is now guarded with `Number.isNaN` so carrying over non-numeric `search` from a string property renders as empty (`null`) instead of `NaN`.

react/src/pages/ResourcePolicyPage.tsx
```diff
@@ -55,7 +89,18 @@ const ResourcePolicyPage: React.FC<ResourcePolicyPageProps> = () => {
...
+                <UserResourcePolicyV2
+                  queryRef={userResourcePolicyQueryRef}
+                  onReload={loadUserResourcePolicyQuery}
+                />
+              ) : (
+                <Skeleton active />
+              )
+            ) : (
+              <UserResourcePolicyList />
+            )}
```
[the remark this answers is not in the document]
nowgnuesLee: Intentional — the version branching is by design. The V2 filter emits AND/OR/NOT sub-filters that `UserResourcePolicyV2Filter` only gained in manager 26.7.0, so we fall back to the legacy list on older managers to avoid sending rejected arguments. PR description updated to document this instead of the stale "no version branching" note.

packages/backend.ai-ui/src/components/BAIUserResourcePolicyV2Table.tsx
```diff
@@ -0,0 +1,179 @@
...
+        sorter: isEnableSorter('maxVfolderCount'),
+        render: (text) => (_.toNumber(text) === 0 ? '∞' : text),
+      },
+      {
+        title: t('resourcePolicy.MaxConcurrentLogins'),
+        dataIndex: 'maxConcurrentLogins',
+        key: 'maxConcurrentLogins',
+        sorter: isEnableSorter('maxConcurrentLogins'),
+        render: (text) => (_.isNil(text) ? '∞' : text),
+      },
```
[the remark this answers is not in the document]
nowgnuesLee: The scope changed deliberately — `maxConcurrentLogins` is now surfaced as a column and a form field (with a new i18n key). Updated the PR description to reflect this rather than the stale "out of scope" note.

react/src/components/UserResourcePolicyV2SettingModal.tsx
```diff
@@ -0,0 +1,313 @@
...
+            onError(error) {
+              message.error(
+                error?.message ||
+                  t('resourcePolicy.CannotUpdateResourcePolicy'),
+              );
+            },
+          });
+        }
+      })
+      .catch(() => {});
```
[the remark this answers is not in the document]
nowgnuesLee: Keeping as-is. `validateFields().then(...).catch(() => {})` is the established project convention across 10+ setting modals (e.g. KeypairResourcePolicySettingModal, ProjectResourcePolicySettingModal); the rejection is antd's form-validation failure, which surfaces inline per-field errors — not an unexpected runtime error. Changing it here would break consistency with sibling modals.

react/src/components/UserResourcePolicyV2SettingModal.tsx
```diff
@@ -0,0 +1,313 @@
...
+  };
+
+  const handleOk = () => {
+    return formRef?.current
+      ?.validateFields()
+      .then((values) => {
+        const maxQuotaScopeSizeExpr =
+          values?.max_quota_scope_size === -1
+            ? '-1'
+            : String(GBToBytes(values?.max_quota_scope_size));
```
agatha197, line 131: Please use dynamic units.

packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
```diff
@@ -745,6 +746,29 @@ const BAIGraphQLPropertyFilter: React.FC<BAIGraphQLPropertyFilterProps> = ({
...
+            }
+            onChange={(val) => {
+              setIsValid(true);
+              setSearch(_.isNil(val) ? '' : String(val));
+            }}
+            onPressEnter={() => addCondition(search)}
+            style={{ minWidth: 200 }}
+            placeholder={
+              selectedProperty?.placeholder ??
+              t('comp:BAIPropertyFilter.PlaceHolder')
```
agatha197: How about setting it just `undefined`?

react/src/components/UserResourcePolicyV2Table.tsx
```diff
@@ -0,0 +1,179 @@
...
+      },
+      {
+        title: t('resourcePolicy.MaxQuotaScopeSize'),
+        dataIndex: 'maxQuotaScopeSize',
+        key: 'maxQuotaScopeSize',
+        sorter: isEnableSorter('maxQuotaScopeSize'),
+        render: (__, row) =>
+          row.maxQuotaScopeSize.value === -1
+            ? '∞'
+            : bytesToGB(row.maxQuotaScopeSize.value),
```
agatha197: here too.

agatha197 (changes requested):

packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
nowgnuesLee: We'd prefer to keep showing a placeholder here — without it the empty number input gives no hint that a value should be typed, and the other input branches (date/autocomplete) also show a fallback placeholder. Keeping the fallback for consistency.

react/src/components/UserResourcePolicyV2SettingModal.tsx
nowgnuesLee: Applied partially — the list now displays maxQuotaScopeSize with dynamic units (`convertToBinaryUnit(value, 'auto')`, e.g. "10 GiB"). For this modal input, however, we're keeping the fixed-GB `InputNumber`: `BAIDynamicUnitInputNumber` takes string values ("10g"), which conflicts with the numeric `-1` unlimited sentinel used by `FormItemWithUnlimited` and forces an `undefined`-as-unlimited pattern that hurts readability here.

react/src/components/UserResourcePolicyV2Table.tsx
nowgnuesLee: Applied — the column now renders with dynamic units via `convertToBinaryUnit(value, 'auto')?.displayValue` (e.g. "10 GiB") instead of fixed GB.

react/src/components/UserResourcePolicyV2Table.tsx
```diff
@@ -0,0 +1,180 @@
...
+      {
+        title: t('resourcePolicy.MaxQuotaScopeSize'),
+        dataIndex: 'maxQuotaScopeSize',
+        key: 'maxQuotaScopeSize',
+        sorter: isEnableSorter('maxQuotaScopeSize'),
+        render: (__, row) =>
+          row.maxQuotaScopeSize.value === -1
+            ? '∞'
+            : convertToBinaryUnit(row.maxQuotaScopeSize.value, 'auto')
+                ?.displayValue,
```
ironAiken2: The modal reads/writes this field in **decimal GB** — the input has `suffix="GB"` and converts with `GBToBytes` / `bytesToGB` (10^9) — but the list renders it with `convertToBinaryUnit` (1024-based). So a policy saved as "10 GB" shows up here as "9.31 GiB", and the number never matches what the admin typed. Legacy rendered with `bytesToGB` (decimal), so this is also a behavior change.

Since the input side is fixed decimal GB, how about using the decimal helper here?

```suggestion
        render: (__, row) =>
          row.maxQuotaScopeSize.value === -1
            ? '∞'
            : convertToDecimalUnit(row.maxQuotaScopeSize.value, 'auto')
                ?.displayValue,
```

(with the import swapped accordingly). While you're here, `?.displayValue` renders empty on a conversion miss — a `?? '-'` fallback would be nice. :)

react/src/components/UserResourcePolicyV2SettingModal.tsx
ironAiken2, line 131: Same unit question from the input side: `GBToBytes` is decimal (×10^9), but the server models this as `BinarySizeInfo` / `BinarySizeInput`. If the manager interprets the expr as binary (GiB, ×2^30), every non-unlimited value is off by ~7%.

This pairs with the `expr: '-1'` sentinel you already listed under "Needs verification (backend)" — please confirm the unit basis with the manager team and align both this conversion and the list display (see the table comment) to whichever it is.

react/src/components/UserResourcePolicyV2SettingModal.tsx
```diff
@@ -0,0 +1,313 @@
...
+              required: true,
+              message: t('data.explorer.ValueRequired'),
+            },
+            {
+              max: 255,
+            },
+          ]}
+        >
+          <Input disabled={!!userResourcePolicy} />
+        </Form.Item>
```
ironAiken2, line 237: The legacy modal validated duplicate names client-side (`existingPolicyNames` + `resourcePolicy.ResourcePolicyNameAlreadyExists`); V2 drops that, so a duplicate now only surfaces on the server round-trip. Was that intentional? The orchestrator already holds the loaded policy list, so keeping the inline validator for immediate feedback is cheap.

react/src/components/UserResourcePolicyV2SettingModal.tsx
```diff
@@ -0,0 +1,313 @@
...
+  theme,
+  FormInstance,
+} from 'antd';
+import { BAIModal, BAIModalProps, BAIFlex } from 'backend.ai-ui';
+import * as _ from 'lodash-es';
+import React, { useRef } from 'react';
+import { useTranslation } from 'react-i18next';
+import { graphql, useFragment, useMutation } from 'react-relay';
+
+interface Props extends Omit<BAIModalProps, 'onOk' | 'onCancel'> {
```
ironAiken2: How about `UserResourcePolicyV2SettingModalProps` instead of the generic `Props`, matching `UserResourcePolicyV2TableProps` / `UserResourcePolicyV2Props` in the sibling files? :)

resources/i18n/en.json
```diff
@@ -2402,13 +2402,15 @@
     "IdleTimeoutSec": "Idle timeout (sec.)",
     "Keypair": "Keypair",
     "KeypairResourcePolicy": "Keypair Resource Policy",
+    "MaxConcurrentLogins": "Max Concurrent Login Sessions",
```
ironAiken2: Key says `Logins`, value says "Login **Sessions**" — small drift between key and copy. Fine to leave, but worth a glance for consistency with `MaxConcurrentSFTPSessions` right below. :)

ironAiken2 (changes requested):
Reviewed the V2 migration — the orchestrator / table / modal split follows the established pattern (`LoginSession`), pagination correctly uses `limit` + `offset` only, the delete flow keeps the typed-confirmation modal, and the `sub-filter` gating with the legacy fallback is well documented. 👍

Requesting changes for one user-visible correctness issue: `maxQuotaScopeSize` is entered in decimal GB in the modal but rendered with binary (1024-based) units in the list, so the displayed number never matches what the admin typed. Details inline.

Non-blocking note:
- `BAIGraphQLPropertyFilter`: the new `InputNumber` branch for `type: 'number'` looks right (and closes the NaN hole), but please double-check the existing `type: 'number'` consumers (`ReservoirArtifactDetailPage`, `SessionLauncherPage`) still render acceptably with the numeric-only input.

### ⚠ Needs your decision
- Should `maxQuotaScopeSize` be **decimal (GB, 10^9)** or **binary (GiB, 2^30)** end-to-end — and does the manager accept `BinarySizeInput { expr: "-1" }` as the unlimited sentinel?
- Gating the **entire** V2 page on `sub-filter` (26.7.0) means managers that already expose `adminUserResourcePoliciesV2` but are < 26.7.0 stay on the legacy page — is that the intended conservative window?
- Is dropping the client-side duplicate-name validation intentional (server-only enforcement now)?

react/src/components/UserResourcePolicyV2Table.tsx
nowgnuesLee: Applied — the column now renders with `convertToDecimalUnit(value, 'auto')?.displayValue ?? '-'`, so the list matches the modal's decimal-GB input (10 GB in → "10 GB" shown) and a conversion miss falls back to `-`.

react/src/components/UserResourcePolicyV2SettingModal.tsx
nowgnuesLee: The unit basis here is decimal GB by design: the manager stores `max_quota_scope_size` as a plain byte count and `BinarySizeInput.expr` is parsed as bytes — it doesn't reinterpret the value as GiB. Since the input is labeled "GB" and converts with `GBToBytes` (×10^9), what the admin types is exactly what gets stored, so there's no ~7% drift. The list display is now aligned to the same decimal basis (see the table thread).

react/src/components/UserResourcePolicyV2SettingModal.tsx
nowgnuesLee: Intentional. With V2 offset pagination the orchestrator only holds the current page (default 10 rows), so an `existingPolicyNames` validator can no longer see all policies — it would pass for a name that exists on another page, giving a false sense of safety and confusing users when the server then rejects it. We dropped the partial client-side check and rely on server-side enforcement, which surfaces through the mutation error toast.

react/src/components/UserResourcePolicyV2SettingModal.tsx
nowgnuesLee: Applied — renamed to `UserResourcePolicyV2SettingModalProps` to match the sibling components.

resources/i18n/en.json
nowgnuesLee: Applied — the copy now matches the key: "Max Concurrent Logins" in English, with the "Sessions" wording dropped across all 22 locales.

agatha197 (approved):
LGTM

packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
```diff
@@ -745,6 +746,29 @@ const BAIGraphQLPropertyFilter: React.FC<BAIGraphQLPropertyFilterProps> = ({
...
+            }
+          />
```
ironAiken2: As changes to the value of `InputNumber` are not immediately reflected in `addCondition`, it would be a good idea to display the search button alongside it.

packages/backend.ai-ui/src/components/BAIUserResourcePolicyV2Table.tsx
```diff
@@ -0,0 +1,178 @@
...
+        key: 'maxCustomizedImageCount',
+        sorter: isEnableSorter('maxCustomizedImageCount'),
+      },
+      {
+        title: 'ID',
+        dataIndex: 'id',
+        key: 'id',
+        defaultHidden: true,
+        render: (id: string) => <BAIId globalId={id} />,
+      },
```
ironAiken2: The ‘id’ field in the user resource policy appears to be the same as the ‘name’ field; we need to check whether this is because there is no actual ‘id’ and the ‘name’ is being displayed in its place. If that is the case, this column could probably be deleted.

packages/backend.ai-ui/src/components/BAIUserResourcePolicyV2Table.tsx
```diff
@@ -0,0 +1,178 @@
...
+        title: t('comp:BAIUserResourcePolicyV2Table.MaxQuotaScopeSize'),
+        dataIndex: 'maxQuotaScopeSize',
+        key: 'maxQuotaScopeSize',
+        sorter: isEnableSorter('maxQuotaScopeSize'),
+        render: (__, row) =>
+          row.maxQuotaScopeSize.value === -1
+            ? '∞'
+            : (convertToDecimalUnit(row.maxQuotaScopeSize.value, 'auto')
+                ?.displayValue ?? '-'),
+      },
```
ironAiken2, line 127: The Korean translation for ‘maxQuotaScopeSize’ is not consistent. In the creation and edit modals, it is referred to as ‘최대 폴더 크기’.

react/src/components/UserResourcePolicyV2SettingModal.tsx
```diff
@@ -0,0 +1,313 @@
+/**
```
ironAiken2, line 1: I got an error 

```
{
  "errors": [
    {
      "message": "Int cannot represent non 32-bit signed integer value: 123000000000",
      "path": [
        "adminUpdateUserResourcePolicyV2",
        "userResourcePolicy",
        "maxQuotaScopeSize",
        "value"
      ],
      "extensions": {
        "code": "DOWNSTREAM_SERVICE_ERROR",
        "serviceName": "strawberry"
      }
    }
  ],
  "data": {
    "adminUpdateUserResourcePolicyV2": null
  }
}
```

ironAiken2 (changes requested):

packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
nowgnuesLee: I applied it.

packages/backend.ai-ui/src/components/BAIUserResourcePolicyV2Table.tsx
nowgnuesLee: I applied it.

packages/backend.ai-ui/src/components/BAIUserResourcePolicyV2Table.tsx
nowgnuesLee: I deleted it.

react/src/components/UserResourcePolicyV2SettingModal.tsx
nowgnuesLee: I resolved this error.

ironAiken2 (changes requested):
please resolve the conflicts :)

ironAiken2 (approved):
LGTM

PR #8072, opened 2026-06-29
fix(FR-3179): show actual allocated resources on session detail page

ironAiken2:
Resolves FR-3179 — https://lablup.atlassian.net/browse/FR-3179

> The Jira→GitHub clone for FR-3179 did not exist at PR creation time, so the GitHub `Resolves #N (FR-3179)` reference could not be added. Replace this line with `Resolves #<N> (FR-3179)` once the webhook clones the issue.

## Summary

The session detail page previously showed only the **requested** resources. When the actually-allocated resources differ from the request — e.g. a 1.5 fGPU request rounded down to 1.0 fGPU by the quantum size — the user could not tell what the session actually holds.

The **Resource Allocation** row now shows the actually-allocated resources (`occupied_slots`) as the primary value, and surfaces any shortfall against the request inline:

- Added `occupied_slots` to `SessionDetailContentFragment` (and the list/dashboard queries that store-prefetch it).
- Each resource chip shows the allocated amount as the primary value; when it differs from the request, the requested amount is appended **in the same chip** as a muted `/ <value>` reference sharing one unit (e.g. `1 / 2 Core`). The whole number group carries an *Allocated / Requested* tooltip.
- The **Resource Allocation** label shows a `WarningOutlined` icon (in `colorWarning`) with a tooltip explaining that fewer resources were allocated than requested, whenever any slot differs.
- Before allocation (e.g. PENDING sessions where `occupied_slots` is empty), it falls back to the requested resources as the primary value with no comparison shown.
- Allocated vs. requested are compared after normalizing slot values to numbers and dropping zero-valued slots, so the `/ requested` reference and the label warning only appear on a genuine difference.
- Resource chips are separated from each other by a vertical divider (same pattern as `ImageNodeSimpleTag`).

## Component changes

- `BAIResourceNumberWithIcon` gained an optional `comparedValue` prop. When set, it renders `value / comparedValue` (the compared part muted) and wraps the whole number group in an internally-defined *Allocated / Requested* tooltip — independent of the resource icon's own description tooltip.
- `BAINumberWithUnit` gained a matching `comparedValue` prop for binary (memory) slots.
- `ResourceNumbersOfSession` gained an optional `comparedResource` prop. Only the session detail page passes it (allocated as `resource`, requested as `comparedResource`); the other 5 call sites are unchanged.

## i18n

- Host `session.AllocatedLessThanRequested` (en: `Fewer resources were allocated than requested.`, ko: `자원이 요청보다 적게 할당되었습니다.`) — the label warning tooltip.
- BUI `comp:BAIResourceNumberWithIcon.AllocatedVsRequested` (en: `Allocated / Requested`, ko: `할당 자원 / 요청 자원`) — the chip number-group tooltip.
- Both keys translated into all remaining target languages (19 additional locales, host + BUI).
- The earlier `session.RequestedResources` key (from the superseded secondary-row approach) was removed.

## Verification

`bash scripts/verify.sh` → Relay / Lint / Format / TypeScript all pass. Regenerated `__generated__` artifacts (the list/dashboard queries that store-prefetch `SessionDetailContentFragment`) are committed.

## Testing

This can be verified on the **10.82.0.130** test server using a session that was intentionally allocated fewer resources than it requested:

`/session?sessionDetail=7448ec52-c18a-4dfe-b248-5fbafe250086`

Open the session detail page and check the **Resource Allocation** row — the under-allocated slot shows the allocated amount with the requested amount as a muted `/ requested` reference, and the row label shows the warning icon.

### After

The **Resource Allocation** row now shows the actually-allocated resources.

Resource allocation after

> **Note:** the `/ requested` reference and the label warning icon only appear when the allocation differs from the request (e.g. fractional-GPU requests rounded down by the quantum size). All finished sessions available on the test backend use integer slots (1 Core / 1 GiB, no AI accelerator), so the difference case could not be reproduced for a screenshot — the captured state shows the primary allocation row, which is visually identical for integer slots.


ironAiken2:
* **#8072** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8072?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8072?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8072).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1516,50 +1531,86 @@ const UnifiedAcceleratorChip: React.FC<{ type: string }> = ({ type }) => {
...
+    ) : null;
+  // Separate each resource chip with a vertical divider (same pattern as
+  // ImageNodeSimpleTag) so the resources read as distinct items.
+  const chips = _.compact([...slotChips, acceleratorChip]);
+  return (
+    <>
+      {chips.map((chip, index) => (
+        <React.Fragment key={index}>
+          {index > 0 ? (
+            <Divider orientation="vertical" style={{ marginInline: 0 }} />
```
agatha197, line 1651: **Nit (real regression):** the vertical `Divider` is baked unconditionally into this shared component, guarded only by `index > 0`, so it now affects all five callers of `ResourceNumbersOfSession` — not just the session-detail target. Four use row flex (cosmetic at worst), but `DeploymentPresetDetailModal.tsx:200` renders these chips inside `<BAIFlex direction="column" align="stretch">`, where each vertical divider becomes a stray ~0.9em 1px line between stacked items — a real visual regression plus scope creep beyond FR-3179.

Consider gating the divider on an opt-in prop (e.g. `showDividers`) set only by `SessionDetailContent`, or rendering dividers only when `comparedResource` is present, and eyeballing the four non-detail callers.

react/src/components/SessionDetailContent.tsx
```diff
@@ -262,6 +273,50 @@ const SessionDetailContent: React.FC<{
...
+  const occupiedSlotNumbers = _.omit(
+    parseSlotsToNumbers(session?.occupied_slots),
+    unifiedSlotName,
+  ) as Record<string, number>;
+  // Slot types whose allocated amount differs from what was requested. Only
+  // meaningful once the session is actually allocated (occupied_slots present).
+  // Used solely to decide whether to flag the section label with a warning icon;
+  // the per-chip `allocated / requested` rendering is computed inside
+  // `ResourceNumbersOfSession` itself.
+  const slotTypesDifferingFromRequest = hasOccupiedSlots
```
agatha197: **Nit (copy vs. logic):** detection here uses strict not-equal (`requested !== occupied`), i.e. *any* difference, but the row-label tooltip copy is directional — `session.AllocatedLessThanRequested` (EN "Fewer resources were allocated than requested." / KO "자원이 요청보다 적게 할당되었습니다."). If allocation ever exceeds the request, the label text would be factually wrong (the neutral per-chip "Allocated / Requested" tooltip stays correct). Low risk given the round-down premise, but it's an unasserted assumption.

Either tighten detection to `occupied < requested` to match the copy, or soften the label to something direction-agnostic like "Allocated resources differ from requested."

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1516,50 +1531,86 @@ const UnifiedAcceleratorChip: React.FC<{ type: string }> = ({ type }) => {
...
+      // field on hover.
+      <UnifiedAcceleratorChip type={displayResource.acceleratorType} />
+    ) : displayResource &&
+      displayResource.accelerator &&
+      displayResource.acceleratorType &&
+      _.isNumber(displayResource.accelerator) ? (
+      <BAIResourceNumberWithIcon
+        // @ts-ignore
+        type={displayResource.acceleratorType}
+        value={_.toString(displayResource.accelerator * containerCount)}
```
agatha197: **Nit (latent contract gap, not currently reachable):** the `comparedResource` JSDoc says "for any slot whose primary value differs… the compared amount is appended," but this numeric `{accelerator, acceleratorType}` branch renders `BAIResourceNumberWithIcon` with no `comparedValue`. It's not reachable today — the only caller passing `comparedResource` (`SessionDetailContent`) uses raw slot maps that route fGPU through `slotChips` via the dotted `cuda.shares` key, so the detail page works. But the documented contract would break the moment a form-shaped caller passes `comparedResource`.

Either thread `comparedResource?.accelerator` into this branch, or narrow the JSDoc + add a one-line note that the dotted-key `slotChips` path is what covers fGPU on the detail page (so a future refactor doesn't silently break it).

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1516,50 +1531,86 @@ const UnifiedAcceleratorChip: React.FC<{ type: string }> = ({ type }) => {
...
+          : _.toNumber(raw) * containerCount + '';
+      // When a compared resource is present, compare it against this slot's
+      // primary value. A difference passes the compared value to the chip, which
+      // appends it as a muted `/ <value>` reference and wraps the whole number
+      // group in an "allocated / requested" tooltip (separate from the icon's
+      // description tooltip).
+      const comparedRaw = comparedResource
+        ? _.get(comparedResource, type)
+        : undefined;
+      const isDifferent =
```
agatha197: **Nit (cosmetic edge case):** `isDifferent` compares raw values with `_.toNumber`, but for `mem` the raw is a byte count while `BAINumberWithUnit` renders both operands via `convertToBinaryUnit(..., 'g', 2, true)`. Two byte values inside the same 0.01-GiB bucket could render "8 / 8 GiB" with the warning icon — reads as a rendering bug. Realistically near-impossible for memory (allocated as requested), so acceptable to leave; if you want to be safe, compare on the displayed/rounded value or skip the reference when the rendered strings are equal.

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1516,50 +1531,86 @@ const UnifiedAcceleratorChip: React.FC<{ type: string }> = ({ type }) => {
...
+        value={_.toString(displayResource.accelerator * containerCount)}
+      />
+    ) : null;
+  // Separate each resource chip with a vertical divider (same pattern as
+  // ImageNodeSimpleTag) so the resources read as distinct items.
+  const chips = _.compact([...slotChips, acceleratorChip]);
+  return (
+    <>
+      {chips.map((chip, index) => (
+        <React.Fragment key={index}>
```
agatha197: **Nit (key stability):** `key={index}` on the `React.Fragment` makes the inner `key={type}` inert. Since `chips = _.compact([...slotChips, acceleratorChip])` drops `'0'` slots and conditionally appends the accelerator, index keys can map a chip to a different slot's prior node across renders. Output stays correct (these are stateless presentational chips), so this is purely the loss of the previous stable per-type key. Consider carrying `type` / `'accelerator'` / `'unified'` alongside each element and using that as the Fragment key.

packages/backend.ai-ui/src/locale/zh-TW.json
```diff
@@ -306,6 +306,9 @@
   "comp:BAIResourceGroupSelect": {
     "SelectResourceGroup": "選擇資源群組"
   },
+  "comp:BAIResourceNumberWithIcon": {
+    "AllocatedVsRequested": "已分配 / 请求"
```
agatha197: **Nit (translation):** this zh-TW value uses the simplified character `请` and is byte-identical to zh-CN. The host zh-TW key was localized correctly (`請求`), so this is an isolated BUI slip. Should be `"已分配 / 請求"` (traditional `請`).

agatha197 (commented):
Reviewed via a multi-agent, adversarially-verified pass. Overall: **approve with nits** — the core goal works end-to-end. I verified the headline scenario (a 1.5 fGPU request rounded down to 1.0) genuinely renders `1 / 1.5` on the session detail page: `buildResourceWithUnifiedSlot` passes the raw slot map, the dotted `cuda.shares` key survives `_.omit` and flows through the `slotChips` loop that threads `comparedValue`. i18n coverage is thorough and real across host + BUI locales.

No blockers. The inline comments are all low-severity polish — the only one I'd fix before merge is the divider leaking into a column-layout caller. Nice, clean additive design on the `comparedValue` prop.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Fixed in ee181b62a. Gated the vertical `Divider` behind an opt-in `showDividers` prop that only `SessionDetailContent` sets, so the four non-detail callers — including the `DeploymentPresetDetailModal` column layout — no longer render stray dividers between stacked chips. The distinct-item separation intent is preserved on the detail page.

react/src/components/SessionDetailContent.tsx
ironAiken2: Fixed in ee181b62a. Tightened the section-label detection to `(occupied ?? 0) < (requested ?? 0)` so the directional `AllocatedLessThanRequested` copy only appears when allocation is genuinely less than requested. The neutral per-chip `allocated / requested` reference still fires on any difference, as you noted it should.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Fixed in ee181b62a. Narrowed the `comparedResource` JSDoc to state that comparison applies to the numeric `slotChips` only, and that the object-shaped `{ accelerator, acceleratorType }` branch is intentionally not threaded — on the detail page fGPU is compared via its dotted `cuda.shares` slot key (which routes through `slotChips`). This documents the covered path so a future refactor will not silently break it.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Acknowledged — leaving as-is per your note. For `mem` the raw byte values are identical whenever memory is allocated exactly as requested (the realistic case), so the 0.01-GiB bucket collision is not reachable in practice. Flagging here so it is a conscious decision rather than an oversight; happy to switch to a rendered-value comparison if a real case ever surfaces.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Fixed in ee181b62a. Each chip now carries a stable identity key (slot `type` / `accelerator` / `unified`) used as the `React.Fragment` key, replacing the array index. The now-inert inner `key={type}` was removed.

packages/backend.ai-ui/src/locale/zh-TW.json
ironAiken2: Fixed in ee181b62a. Corrected to the traditional character: 已分配 / 請求.

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1516,50 +1545,108 @@ const UnifiedAcceleratorChip: React.FC<{ type: string }> = ({ type }) => {
...
+        // primary value. A difference passes the compared value to the chip,
+        // which appends it as a muted `/ <value>` reference and wraps the whole
+        // number group in an "allocated / requested" tooltip (separate from the
+        // icon's description tooltip).
+        const comparedRaw = comparedResource
+          ? _.get(comparedResource, type)
+          : undefined;
+        const isDifferent =
+          !_.isUndefined(comparedRaw) &&
+          _.toNumber(comparedRaw) !== _.toNumber(value);
```
agatha197, line 1593: **① Raw comparison vs. rounded display (low).**

`isDifferent` compares the **raw** slot-map values, but the chip renders **rounded/unit-converted** values (`toDisplayValue` → `convertToBinaryUnit(...,'g',2,true)` / `formatAmount(...).toFixed()`). A difference below display precision passes this check yet renders two identical-looking numbers:
- mem `4294967296` vs `4294967295` → both display `4` → `4 / 4 GiB`
- `cuda.shares` `0.333` vs `0.334` (round_length 2) → `0.33 / 0.33`

The same raw comparison drives `slotTypesAllocatedLessThanRequested` in `SessionDetailContent`, so the warning icon can appear even when every displayed number is identical. Consider comparing on the displayed (rounded) values, or ignoring differences below display precision. Low likelihood in practice since backend memory allocation is normally exact, and it never *misses* a genuinely visible difference — only over-flags sub-precision ones.

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1516,50 +1545,108 @@ const UnifiedAcceleratorChip: React.FC<{ type: string }> = ({ type }) => {
...
+  // session detail page). Each chip carries a stable identity key (slot type /
+  // 'accelerator' / 'unified') so React reconciles by resource across renders
+  // even though `'0'` slots are dropped and the accelerator chip is
+  // conditionally appended.
+  const displayResource = resource;
+  const slotChips = _.compact(
+    _.map(
+      _.omit(displayResource, 'shmem', 'accelerator', 'acceleratorType'),
+      (value, type) => {
+        if (value === '0') {
```
agatha197: **② Exactly-zero allocated slot drops its chip (low).**

A slot allocated exactly `0` is skipped here (`value === '0'` → `null`), but `slotTypesAllocatedLessThanRequested` still flags it (`0 < requested`). Net effect: the label shows the warning icon/tooltip while the corresponding chip disappears entirely.

Example: requested `cuda.shares=1`, occupied `0` → the user sees the warning but no `0 / 1` GPU chip indicating *which* resource was denied. Consider keeping the chip (rendering `0 / 1`) when `comparedResource` has a value for that slot. Narrow — only the fully-denied case; partial under-allocation (e.g. `0.5 / 1`) renders fine, and the label tooltip still conveys the information.

react/src/components/SessionDetailContent.tsx
```diff
@@ -89,6 +90,15 @@ const buildResourceWithUnifiedSlot = (
     : slots;
 };
 
+// Parse a resource-slot JSON string into a numeric map, dropping zero-valued
+// slots so that requested vs. allocated slots compare equal when they only
+// differ by absent/zero entries.
+const parseSlotsToNumbers = (slots?: string | null) =>
```
agatha197, line 98: **③ Missing `'use memo'` on a now data-heavy body (nit).**

This PR adds non-trivial render-body computation to `SessionDetailContent` — `parseSlotsToNumbers` (JSON.parse + `_.mapValues` + `_.omitBy`), the `buildResourceWithUnifiedSlot` calls, and `slotTypesAllocatedLessThanRequested` (`_.omit`/`_.union`/`_.keys`/`.filter`) — but the component has no `'use memo'` directive, so all of it recomputes on every render. Pre-existing omission (not a regression introduced here), but since the file is being touched, adding `'use memo'` at the top of the body matches the project convention.

packages/backend.ai-ui/src/components/BAINumberWithUnit.tsx
```diff
@@ -23,12 +31,23 @@ const BAINumberWithUnit = ({
     unitType === 'binary'
       ? convertToBinaryUnit(numberUnit, 'auto', 2, true)
       : convertToDecimalUnit(numberUnit, 'auto', 2, true);
+  const convertedComparedByTargetUnit =
```
agatha197, line 35: **④ Missing `'use memo'` (nit).**

This adds a computed value (`convertToBinaryUnit`/`convertToDecimalUnit`) to the component body, but `BAINumberWithUnit` still lacks a `'use memo'` directive — unlike its sibling `BAIResourceNumberWithIcon`, which has one. Pre-existing and nit-level; worth adding while here for consistency.

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1516,50 +1545,108 @@ const UnifiedAcceleratorChip: React.FC<{ type: string }> = ({ type }) => {
...
-        (value, type) => {
-          return value === '0' ? null : (
+  'use memo';
+  // `resource` is the primary value to display; `comparedResource` (when given)
+  // is the reference to compare against (e.g. requested vs. allocated on the
+  // session detail page). Each chip carries a stable identity key (slot type /
+  // 'accelerator' / 'unified') so React reconciles by resource across renders
+  // even though `'0'` slots are dropped and the accelerator chip is
+  // conditionally appended.
+  const displayResource = resource;
```
agatha197: **⑤ Redundant alias (nit).**

`const displayResource = resource;` aliases the prop with no transformation and is then used throughout in place of `resource`. It reads as leftover scaffolding — using `resource` directly would be clearer.

agatha197 (commented):
I left some comment. please check them.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Fixed in cd9dbe09e. Two layers, matching your suggestion to compare on displayed values: (1) `BAIResourceNumberWithIcon` now drops a `comparedValue` that rounds to the same displayed number as the primary (via the same `formatAmount` used for rendering — GiB/2dp for binary slots, `round_length` otherwise), so `4 / 4 GiB` / `0.33 / 0.33` pairs can no longer render; the raw comparison in `ResourceNumbersOfSession` remains only as an obviously-equal short-circuit. (2) `slotTypesAllocatedLessThanRequested` in `SessionDetailContent` now normalizes both operands to displayed precision (`toDisplayedNumber`: binary → GiB 2dp via `convertToBinaryUnit`, else `round_length`) before the `<` check, so the label warning only fires on a visible difference.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Fixed in cd9dbe09e. The `value === '0'` skip now keeps the chip when a non-zero compared (requested) amount exists for that slot, so the fully-denied case renders `0 / 1` and the user can see which resource was denied. Zero slots without a compared amount — including all five callers that don't pass `comparedResource` — are dropped exactly as before.

react/src/components/SessionDetailContent.tsx
ironAiken2: Fixed in cd9dbe09e. Added `'use memo'` at the top of the `SessionDetailContent` body.

packages/backend.ai-ui/src/components/BAINumberWithUnit.tsx
ironAiken2: Fixed in cd9dbe09e. Added `'use memo'` to `BAINumberWithUnit`, matching its sibling `BAIResourceNumberWithIcon`.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Fixed in cd9dbe09e. Removed the `displayResource` alias — the component now uses the `resource` prop directly throughout.

agatha197 (changes requested):
Can you move icon to the right with more margin?

agatha197 (approved):
LGTM

PR #8099, opened 2026-07-01
feat(FR-3237): confirm endpoint reachability before showing offline banner

### LINKED ISSUE #8096: Confirm offline state with a lightweight endpoint GQL probe before showing the offline banner
## Problem

The network status banner (`NetworkStatusBanner.tsx`) currently shows the "You are offline" alert based solely on the browser's own offline judgment (e.g. `navigator.onLine` / browser online/offline events). The browser can report offline (or a false positive) even when the Backend.AI API endpoint is actually reachable, which can surface a misleading offline banner to the user.

## Proposed solution

When the browser judges the network to be offline, do one additional confirmation round-trip against the API endpoint before concluding "offline":

- Fire the same lightweight GraphQL query used during login to verify endpoint connectivity.
- Only if that probe also fails, treat the state as genuinely offline and show the banner.
- If the probe succeeds, the endpoint is actually connected — suppress the offline banner.

This makes the offline determination reflect real endpoint reachability rather than the browser's local network guess alone.

## Notes

- Reuse the existing login-time endpoint check GQL query rather than introducing a new probe.
- Keep the probe cheap and short-timeout so it doesn't delay the banner meaningfully when the endpoint is truly down.

—

Captured while working on branch: feature/theme-families

JIRA Issue: FR-3237

nowgnuesLee:
Resolves #8096 (FR-3237)

## Problem

`NetworkStatusBanner` decided the "You are offline" banner **solely** from `navigator.onLine` (via ahooks `useNetwork`). `navigator.onLine` is only a heuristic — it is frequently a false positive on VPNs, captive portals, and virtualized networks — so the banner could appear even while the Backend.AI API endpoint was perfectly reachable.

## Solution

When the browser reports offline, confirm reachability with a **plain `fetch` to the endpoint's `/health`** before showing the banner:

- The probe runs from a `useEffect` + `useEffectEvent` while the browser reports offline: once immediately, then every 5s so the banner reflects current reachability. No Relay / react-query layer — the result is a transient local state flag with no caching value.
- The banner shows **only** when `navigator.onLine` is offline **AND** the `/health` probe fails (network error, timeout, or non-OK response). While the probe is in flight, the banner stays hidden.
- The probe is bounded to ~5s via `AbortSignal.timeout`, so a genuinely unreachable endpoint still surfaces the banner promptly.
- The effect cleanup drops in-flight probe results and resets the flag, so each offline episode starts with the banner hidden instead of flashing a stale failure from the previous episode.
- The endpoint URL comes from the authenticated global client (`useSuspendedBackendaiClient()` → `client._config.endpoint`), joined with the same `endpoint + path` convention the client itself uses.

Also migrated the soft-timeout `Alert`'s deprecated top-level `onClose` to the antd v6 `closable` object, and added the `'use memo'` directive to the component (React Compiler convention) since the file was being edited.

## Verification

`bash scripts/verify.sh` — Relay / Lint / Format / TypeScript all **PASS**. (The unrelated "Vite warmup paths" check fails identically on a clean `main` checkout — it validates build artifacts, not this change.)

## Test plan

**How to test**: turn off Wi-Fi (so the browser reports offline), then open DevTools → Network and confirm that requests to `/health` are being fired every ~5s — that is the reachability probe running.

- [ ] With a reachable endpoint, toggle the browser to "Offline" (DevTools Network → Offline, or `navigator.onLine` false): the offline banner should **not** appear because the `/health` probe succeeds.
- [ ] With the endpoint genuinely unreachable (stop the manager / block the host) and the browser offline: the banner appears within ~5s.
- [ ] Restore reachability while still offline: the banner clears on the next probe (~5s).
- [ ] Go back online and offline again: no stale banner flash at the start of the second offline episode.
- [ ] Soft-timeout warning alert still shows and is dismissable.

### DIFF
--- a/react/src/components/NetworkStatusBanner.tsx
+++ b/react/src/components/NetworkStatusBanner.tsx
@@ -2,14 +2,18 @@
  @license
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
+import { useSuspendedBackendaiClient } from '../hooks';
 import { useDebounce, useNetwork } from 'ahooks';
 import { Alert } from 'antd';
 import { createStyles } from 'antd-style';
 import { atom, useSetAtom } from 'jotai';
-import { useEffect, useState } from 'react';
+import { useEffect, useEffectEvent, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
 const isDisplayedNetworkStatusState = atom(false);
+
+const REACHABILITY_PROBE_INTERVAL_MS = 5_000;
+const REACHABILITY_PROBE_TIMEOUT_MS = 5_000;
 
 const useStyles = createStyles(({ token, css }) => ({
   borderError: css`
@@ -24,12 +28,15 @@ const useStyles = createStyles(({ token, css }) => ({
   `,
 }));
 const NetworkStatusBanner = () => {
+  'use memo';
   const { t } = useTranslation();
   const network = useNetwork();
+  const client = useSuspendedBackendaiClient();
   const setDisplayedStatus = useSetAtom(isDisplayedNetworkStatusState);
   const { styles } = useStyles();
   const [showSoftTimeoutAlert, setShowSoftTimeoutAlert] = useState(false);
   const [dismissSoftTimeoutAlert, setDismissSoftTimeoutAlert] = useState(false);
+  const [endpointUnreachable, setEndpointUnreachable] = useState(false);
 
   useEffect(() => {
     const softHandler = () => {
@@ -62,7 +69,47 @@ const NetworkStatusBanner = () => {
     wait: 5_000,
   });
 
-  const shouldOpenOfflineAlert = !network.online;
+  const browserOffline = !network.online;
+
+  // `navigator.onLine` is only a heuristic (false positives on VPNs, captive
+  // portals, virtualized networks), so confirm with the endpoint's `/health`
+  // before concluding the user is offline.
+  const probeEndpointReachability = useEffectEvent(async () => {
+    try {
+      const resp = await fetch(`${client._config.endpoint}/health`, {
+        signal: AbortSignal.timeout(REACHABILITY_PROBE_TIMEOUT_MS),
+      });
+      return !resp.ok;
+    } catch {
+      return true;
+    }
+  });
+
+  useEffect(
+    function probeEndpointWhileBrowserOffline() {
+      if (!browserOffline) {
+        return;
+      }
+      let disposed = false;
+      const runProbe = async () => {
+        const unreachable = await probeEndpointReachability();
+        if (!disposed) {
+          setEndpointUnreachable(unreachable);
+        }
+      };
+      runProbe();
+      const intervalId = setInterval(runProbe, REACHABILITY_PROBE_INTERVAL_MS);
+      return function cleanupProbe() {
+        disposed = true;
+        clearInterval(intervalId);
+        // Reset so the next offline episode doesn't flash a stale failure.
+        setEndpointUnreachable(false);
+      };
+    },
+    [browserOffline],
+  );
+
+  const shouldOpenOfflineAlert = browserOffline && endpointUnreachable;
   const shouldOpenSoftAlert =
     !shouldOpenOfflineAlert && debouncedShowAlert && !dismissSoftTimeoutAlert;
 
@@ -85,9 +132,10 @@ const NetworkStatusBanner = () => {
           title={t('webui.NetworkSoftTimeout')}
           className={styles.borderWarning}
           banner
-          closable
-          onClose={() => {
-            setDismissSoftTimeoutAlert(true);
+          closable={{
+            onClose: () => {
+              setDismissSoftTimeoutAlert(true);
+            },
           }}
         />
       )}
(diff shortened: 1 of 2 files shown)


nowgnuesLee:
* **#8099** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8099?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8099?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8099).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/NetworkStatusBanner.tsx
```diff
@@ -62,7 +75,48 @@ const NetworkStatusBanner = () => {
...
+    },
+    // Only probe while the browser reports offline; re-probe periodically so
+    // the banner reflects the current endpoint reachability, not a stale
+    // one-shot result.
+    enabled: browserOffline,
+    // A disabled query does not poll, so this only runs while `browserOffline`.
+    refetchInterval: 5_000,
+    retry: false,
+    staleTime: 0,
+  });
```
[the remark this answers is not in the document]
nowgnuesLee: Good catch — this applied to the earlier `getServerVersion` probe, which went through `_wrapWithPromise` and its soft-timeout timer. Resolved by 45380bc30: the probe is now a plain `fetch` to the endpoint's `/health`, which bypasses `_wrapWithPromise` entirely, so the probe path can no longer schedule (or leak) `backend-ai-network-soft-time-out` timers.

react/src/components/NetworkStatusBanner.tsx
```diff
@@ -62,7 +69,47 @@ const NetworkStatusBanner = () => {
...
+  const probeEndpointReachability = useEffectEvent(async () => {
+    try {
+      const resp = await fetch(`${client._config.endpoint}/health`, {
+        signal: AbortSignal.timeout(REACHABILITY_PROBE_TIMEOUT_MS),
+      });
+      return !resp.ok;
+    } catch {
+      return true;
+    }
+  });
```
agatha197, line 86: **Add a unit test for the reachability probe.**

`scripts/verify.sh` only runs Relay/lint/format/tsc, so the probe's branches (OK → reachable, non-OK → unreachable, network/CORS throw → unreachable, timeout → unreachable) are never exercised. Lifting the fetch-and-classify into a pure, **exported** `isEndpointUnreachable(endpoint)` helper lets a small Vitest spec cover them with no component render / Relay setup:

```tsx
// react/src/components/NetworkStatusBanner.test.tsx
import { isEndpointUnreachable } from './NetworkStatusBanner';
import { afterEach, describe, expect, it, vi } from 'vitest';

describe('isEndpointUnreachable', () => {
  afterEach(() => {
    vi.restoreAllMocks();
  });

  it('probes `${endpoint}/health` and treats an OK response as reachable', async () => {
    const fetchMock = vi
      .spyOn(globalThis, 'fetch')
      .mockResolvedValue(new Response(null, { status: 200 }));
    await expect(isEndpointUnreachable('https://api.test')).resolves.toBe(false);
    expect(fetchMock).toHaveBeenCalledWith(
      'https://api.test/health',
      expect.objectContaining({ signal: expect.any(AbortSignal) }),
    );
  });

  it('treats a non-OK response as unreachable', async () => {
    vi.spyOn(globalThis, 'fetch').mockResolvedValue(new Response(null, { status: 503 }));
    await expect(isEndpointUnreachable('https://api.test')).resolves.toBe(true);
  });

  it('treats a network/CORS error as unreachable', async () => {
    vi.spyOn(globalThis, 'fetch').mockRejectedValue(new TypeError('Failed to fetch'));
    await expect(isEndpointUnreachable('https://api.test')).resolves.toBe(true);
  });

  it('treats a timeout as unreachable', async () => {
    vi.spyOn(globalThis, 'fetch').mockRejectedValue(
      new DOMException('The operation timed out.', 'TimeoutError'),
    );
    await expect(isEndpointUnreachable('https://api.test')).resolves.toBe(true);
  });
});
```

agatha197 (commented):
One optional follow-up on test coverage (non-blocking).

react/src/components/NetworkStatusBanner.tsx
nowgnuesLee: Thanks — added component-level tests instead in 61b70f5 (`react/src/components/NetworkStatusBanner.test.tsx`).

I went with a component render test rather than a pure exported `isEndpointUnreachable` helper because the classification logic (`!resp.ok` on success, `catch → unreachable`) is trivial — a helper spec would mostly be a change-detector, and it can't cover what this PR actually changes: the **gating** (offline + reachable ⇒ banner suppressed, the `navigator.onLine` false-positive case) and the **reset on reconnect** (no stale flash). The component test covers:

- online ⇒ no banner, no probe
- offline + non-OK (503) ⇒ banner, and asserts the `${endpoint}/health` + `AbortSignal` probe call
- offline + thrown (network/CORS/timeout) ⇒ banner
- offline + reachable (200) ⇒ banner suppressed (the whole point)
- reconnect ⇒ banner cleared

Only 3 mocks needed (`useSuspendedBackendaiClient`, ahooks `useNetwork`, `fetch`), so no Relay setup. All green + `scripts/verify.sh` passes.

agatha197 (approved):
LGTM

PR #8101, opened 2026-07-02
feat(FR-3228): add idle-checker reclamation status column to session list

### LINKED ISSUE #8077: Show idle-checker reclamation status & utilization as a session list column
## Problem

In the user console, a session can be reclaimed/deleted by the idle resource-reclamation policy (utilization idle checker). Today, to monitor whether their own session is at risk, a user must open the session detail and hover over the idle-status section to read the per-resource utilization vs. reclamation-threshold values. This information is hard to discover — during user onboarding/training at a customer site, many users reported they could not find it.

## Request

Surface the idle-checker reclamation status (and the underlying average-utilization / threshold values) directly in the session list so users can see at a glance whether a session is close to being reclaimed.

## Proposed Design (converged from discussion)

- Add an **optional column** to the session list. The session list column settings currently have no utilization / idle-checker column to enable.
- Use a **single column** showing the **reclamation status as text** (primary signal) rather than two separate columns — avoid a dedicated standalone "reclamation status" column plus a separate utilization column.
- Add a **traffic-light style indicator (red / yellow / green)** representing the **overall** reclamation risk, placed next to the idle-checker info. Do **not** color-code the individual CPU/GPU numbers.
  - The color must be a single overall status, not per-resource. Reason: the threshold check operator (`thresholds_check_operator`, AND vs OR) changes how per-resource values map to risk. With AND, a resource below its threshold can still be safe if another resource exceeds; with OR, a single under-threshold resource puts the session at risk. Per-resource coloring would therefore be ambiguous.
- Place an **info icon** next to the status to clearly invite the user to hover.
- On hover, show a **tooltip with detailed info**: per-resource average utilization vs. threshold (e.g. `CPU: 123.1 / 1 | GPU: 5.7 / 1`) plus an explanation of what each status color means.
- Label the values clearly as an average, e.g. **"average utilization over the recent 6 hours / reclamation threshold (%)"**.

## Color semantics to document (in tooltip / help)

The current implementation already color-codes utilization by percentage. The status colors should follow the existing reclamation thresholds, roughly:

- **Red** — utilization is very close to / below the reclamation threshold; session is a reclamation target and may be deleted soon.
- **Yellow** — utilization is low but not yet at the reclamation threshold; safe for now but at risk if it drops further.
- **Green** — utilization is comfortably above the threshold; not at risk for the time being.

## Open considerations

- The hover tooltip will need to hold a fair amount of text (per-resource values + color legend); the layout must handle multiple resource values without becoming overly long.
- Final UI to be reviewed with the team before implementation.
- This should likely be cherry-picked / patched into the 26.4.3 release branch for the customer; the exact patch criteria are still to be decided.

## Reference (current color logic)

The existing logic already derives red/orange/green from utilization vs. threshold (single-threshold case mirrors the requested behavior). The new work is to expose that status, with a tooltip and an explanatory legend, as an opt-in session-list column.

—

Captured while working on branch: fr-3221

JIRA Issue: FR-3228

nowgnuesLee:
Resolves #8077 (FR-3228)

## Summary

Adds an **opt-in "Reclamation Status" column** to the compute-session list so users can see at a glance whether a session is at risk of being reclaimed (deleted) by the utilization idle checker. Previously this was only discoverable by opening the session detail and hovering the idle-status section — during customer onboarding many users reported they could not find it.

## What's included

- **`SessionReclamationStatusCell`** — the opt-in column cell (`defaultHidden: true`, toggled via the existing column-settings UI): a traffic-light badge (red / yellow / green) + status text ("At risk" / "Warning" / "Safe") + an info icon.
- **`SessionReclamationStatusPopover`** — a shared Relay fragment component opened from the info icon, containing:
  - an **operator-aware condition sentence** ("Resources are reclaimed when **all** / **any** of the following conditions remain at the at-risk level", per `thresholds_check_operator`),
  - **one traffic-light row per resource** with its own status label and `(current % | threshold %)` values,
  - a **legend** explaining how each level is determined (safe / warning / at risk).
  - It declares its own `idle_checks` selection, parses it itself, and renders nothing while the checker has no measurement data (e.g. grace period).
- **Session detail drawer reuses the same badge cell** — the utilization idle-check item in `SessionIdleChecks` renders `SessionReclamationStatusCell` (badge + status text + info icon) via a fragment spread, replacing the old dark "utilization / threshold" tooltip and its title text, so the drawer looks identical to the list column.
- **i18n**: new `session.ReclamationStatus*` keys in all 22 locales. Korean uses the customer-preferred wording **"자원 수거 기준값"**.

## Review feedback round (2026-07-07)

- The popover condition sentence now says "usage-based conditions" (all locales).
- The grace-period tag label's **Korean translation only** changed to **"최초 회수 판단까지"** (`session.GracePeriod`; other languages keep "Grace Period"). Note: the Korean heading in `IdleCheckDescriptionModal` shares this key and changes with it.
- A customer-requested explanation tooltip (ⓘ icon) was added next to the grace-period tag (`session.GracePeriodTooltip`, all locales): time remaining until the first reclamation decision, usage collected during the period, decisions continue on the recent-usage average afterwards.
- The post-grace transient text "Grace expired. Checking..." (`session.TimeoutExceeded`, removed) was replaced with **"Checking reclamation status"** (`session.ReclamationStatusChecking`, all locales).

## Design decisions

- **Single overall status** honoring `thresholds_check_operator`: `or` → worst-color wins (any under-threshold resource endangers the session); `and` → best-color wins (one comfortable resource keeps it safe). Individual numbers in the popover are not color-coded — each row gets its own badge instead.
- Per-resource colors reuse the existing `getUtilizationCheckerColor` thresholds, which now returns `{ color, boundary }` so classification and any displayed cutoff can never diverge.
- **Legacy support removed for this release branch**: the pre-24.12 `compute_session` lazy-load fallback and the `@since` gates on `idle_checks` are dropped (the 26.4.3 target manager always provides `ComputeSessionNode.idle_checks`). ⚠️ **If this work is forward-ported to `main`, the `@since` gates and fallback must be restored** since main supports older managers.

### Session list — opt-in column with popover
session list reclamation popover

### Session detail drawer — same shared popover
(taken before the 2026-07-07 feedback round; the drawer now shows the same badge cell as the list)
session detail reclamation popover

## Verification

Relay ✅ · TypeScript ✅ · ESLint ✅ · Prettier ✅ (run individually via `corepack pnpm@10.16.0` — this branch pins pnpm `^10.16.0`).

## Test server

http://10.82.1.99:8090

## Notes

- **Draft** — gated on the team UI review requested in the issue.
- The base `26.4.3` branch received a separate `chore:` commit syncing `pnpm-lock.yaml` with its own `package.json` (pre-existing drift), kept out of this PR's diff.
- Follow-up candidate (separate ticket): the multi-resource branch of `getUtilizationCheckerColor` still has the historical `_.every`/`_.some` always-true-callback semantics (last/first resource wins); the new column computes its overall status correctly via `getOverallReclamation` instead.

## Base

Targets the **`26.4.3` release maintenance branch** (not `main`), per the customer-patch requirement in the issue.


yomybaby (approved):
LGTM

PR #8063, opened 2026-06-27
feat: add multi-theme-family support with custom accent color

inureyes:
## Summary

Adds the named theme-family feature from the `backend.ai-go` / `backend.ai-dol` reference projects: the user picks a visual identity (orange / stained / glass / reverie / bliss), each with light / dark / system, plus an end-user custom accent color. The selector lives in the existing User Settings page next to the light/dark control.

The webui already had the light/dark/system axis and per-user color override on Ant Design's token system; this adds the missing family axis. Families are expressed as Ant Design `ThemeConfig` pairs (not raw CSS variables), so antd components are themed through the existing token pipeline and antd's algorithm derives the shades the reference projects hand-author.

## What changed

- `resources/theme.json` gains `families` + `defaultFamily`. The current default is synthesized as the `default` family, so a `theme.json` without `families` still works (backward compatible). Schema updated in `resources/theme.schema.json`.
- `useThemeFamily` (new) owns family + accent state in localStorage, resolves the active family, applies the accent through antd's token algorithm, and is the sole writer of the `data-theme-family` body attribute. `useCustomThemeConfig` now delegates to it, so the single `ConfigProvider`, `ReverseThemeProvider`, and `WebUISider` become family-aware with no change of their own.
- Signature header effects (glass `backdrop-filter` blur, stained gradient) live in `resources/theme-families.css`, keyed on `data-theme-family`, with `@supports` and `prefers-reduced-transparency` fallbacks. `index.html` sets the attribute before first paint to avoid FOUC.
- `WebUIHeader` resolves header text/icon contrast per family via a `headerScheme` lever (`dark` / `light` / `auto`) and a conditional `ReverseThemeProvider`. Omitting `headerScheme` preserves the historical behavior, so the default family is visually unchanged.
- Family selector and accent color picker (`ThemeAccentColorPicker`) added to the User Settings page.
- `colorSanitizer` (new) provides dependency-free hex validation and WCAG contrast helpers for the accent input.

## Persistence and backend

Persistence is localStorage-only today. The WebUI has no server-side per-user preferences store (`BackendAISettingsStore` is localStorage-only, `/config/*` is global/superadmin, `/user-config/*` is session-mounted dotfiles, and `modify_user` has no free-form field). All theme state is routed through one hook so a future server-backed store is a drop-in swap. `docs/theme-family-backend-sync-spec.md` proposes the Manager-side API for cross-device sync.

## Verification

[...]

PR #8104, opened 2026-07-02
feat(FR-3239): refine user theme selector — product-owned built-in families and design-spec simplifications

### LINKED ISSUE #8103: Provide a theme selector for users
Provide a theme selector on the User Settings page so users can pick a visual theme for the WebUI.

- **Theme** selector (theme families such as Stained, Glass, Reverie, Bliss) and a **Primary Color** picker in User Settings
- Built-in families ship in a product-owned `resources/theme-families.json`, separate from the operator-owned `theme.json` (operator `families` entries can still extend/override built-ins)
- The header keeps the reversed-scheme design per the design spec
- Selection persists in localStorage for now; cross-device sync is specced separately in `docs/theme-family-backend-sync-spec.md`

JIRA Issue: FR-3239

ironAiken2:
Resolves #8103 (FR-3239)

> Stacked on #8063 (`feature/theme-families`).

Refines the theme-family feature per design-spec review.

## Summary of Changes

1. **ReverseThemeProvider simplified** — removed the `enabled` prop; the header always renders reversed-scheme children (restores the original behavior)
2. **`headerScheme` removed** — dropped the per-family header-scheme concept from theme.json / schema / types / hooks / tests / CSS comments
3. **`colorSanitizer` removed** — deleted the sanitize/contrast-warning logic and its i18n keys (the accent only ever comes from ColorPicker output)
4. **Single-file family catalog** — all named families live in `theme.json`'s `families` block (no separate `theme-families.json`); the `default` family is always synthesized from the top-level `light`/`dark`
5. **Layered catalog structure** — extracted the pure `resolveThemeFamilyCatalog` merge function (synthesized default → theme.json families → remote (future))
6. **`allowThemeMode` gating** — new config.toml flag; the Theme / Primary Color items appear in User Settings only when enabled
7. **Hook API reshape** — consolidated into `useCustomThemeConfig` (applied theme / family selection) + `useDefaultTheme` (branding edit: draft/reset)
8. **Branding draft seeding bug fix** — draft/reset source corrected from the family-resolved config to the raw `theme.json`
9. **Primary color setting internalized** — moved to the `custom_primary_color` user setting, owned by `ThemeAccentColorPicker`
10. **Preview-mode isolation** — the preview window ignores the family selection / primary color, and the corresponding User Settings items are hidden there
11. **i18n** — theme/primary-color copy fixes (ko/en) plus the 4 theme keys translated into the 19 remaining languages
12. **Primary color override limited to `colorPrimary`** *(post-review)* — the user-selected accent no longer overrides `colorLink` / `Layout.headerBg`; those stay family-owned, and Ant Design derives the rest of the palette from `colorPrimary`
13. **Per-scheme primary color + reset support** *(post-review)* — `custom_primary_color` is now `{ light?, dark? }` with separate light/dark pickers in `ThemeAccentColorPicker` (same structure as the Branding page's primary color; an unset scheme shows the active family's color, and clearing both removes the setting). User Settings reset (per-item and group) now covers the theme items: **Theme** clears the stored selection back to the `default` family, and **Primary Color** returns to the `theme.json` defaults — `setActiveThemeFamily(undefined)` clears the selection, so the changed badge works correctly for both items

## Simplifications

- **`ReverseThemeProvider`**: removed the `enabled` prop — the header always renders reversed-scheme children per the design spec. `WebUIHeader` is restored to its original always-wrapped form (`color: token.colorBgBase`).
- **`headerScheme` removed** from `theme.json`, `theme.schema.json`, `ThemeFamilyConfig`, `useThemeFamily`, tests, and CSS comments — header contrast handling is fixed by design, not per-family configuration.
- **`helper/colorSanitizer.ts` removed** (with tests): the accent color only ever comes from antd `ColorPicker.toHexString()`, so the sanitize/contrast-warning layers were unnecessary. The accent low-contrast warning UI and its `userSettings.AccentLowContrastWarning` i18n key were dropped with it.

## Single-file family catalog in `theme.json`

- All named families ship in **`theme.json`'s `families` block** — `stained` / `glass` / `reverie` / `bliss` — keeping theme configuration in one operator-facing file (no separate `theme-families.json`).
- The **`default` family is always synthesized from the top-level `light`/`dark`** — the same document the Branding page edits — and a `default` key inside `families` is ignored so a stale copy can never shadow branding edits in preview mode. Pre-family theme.json files (no `families` block) keep working unchanged as a single-entry catalog (family selection is simply unavailable).
- The `defaultFamily` operator key was dropped; `default` is always the initial selection.
- Catalog merge extracted to a pure `resolveThemeFamilyCatalog(rawConfig, remoteFamilies?)` — the planned admin-defined-families backend store plugs in as the next layer later (see `docs/theme-family-backend-sync-spec.md`).
- Malformed `families` entries (missing `light`/`dark`) are dropped instead of breaking the catalog; theme loading is never blocked.

## i18n

- Copy changes (ko/en): `DescThemeFamily` → "테마를 설정합니다." / "Set the theme.", `ThemeAccentColor` → "주요 색상" / "Primary Color", `DescThemeAccentColor` → "테마의 주요 색상을 설정합니다." / "Set the theme's primary color."
- Filled the 4 `userSettings` theme keys (`ThemeFamily`, `DescThemeFamily`, `ThemeAccentColor`, `DescThemeAccentColor`) into the 19 remaining languages, reusing each language's existing `theme.PrimaryColor` / `DescThemeMode` wording for consistency.

## Operator gating: `allowThemeMode`

- New `[general] allowThemeMode` boolean in `config.toml` (default `false`), wired through the standard `loginConfig.ts` -> `baiClient._config` path.
- Only when enabled does the User Settings page offer the **Theme** (family) selector and the **Primary Color** picker. The **Theme Mode** (light/dark/system) select stays available regardless; the family selector additionally requires more than one family in the catalog.

## Theme-config hook API reshape

Consolidated the `useCustomThemeConfig` / `useThemeFamily` / `useUserCustomThemeConfig` trio into two intent-named surfaces:

- **`useCustomThemeConfig()` → `themeConfig` / `activeThemeFamily` / `setActiveThemeFamily` / `themeFamilies`** — `themeConfig` is the site-applied config (default theme with the active family and custom primary color resolved); the other fields drive the User Settings family selector and its reset behavior. The hook keeps its long-standing name and import path, so all theme providers (`DefaultProviders`, `ReverseThemeProvider`, `ThemeSecondary/AdminProvider`, `WebUISider`, login/about/menu) only change their destructuring.
- **`useDefaultTheme()` → `defaultTheme` / `setDefaultTheme`** (+ path-based `updateDefaultTheme` / `getDefaultThemeValue` / `resetDefaultTheme(paths?)`) — the *editable* default-theme document (localStorage draft of `theme.json`) backing the Branding page and its setting items.

This fixes a real bug: the Branding draft/reset previously read the **family-resolved** config, so editing branding while e.g. Stained was active seeded the draft with purple family tokens. `useDefaultTheme` is now seeded from the raw `theme.json`.

The custom primary color also moved out of the `useThemeFamily` API: it is an ordinary `custom_primary_color` user setting owned by `ThemeAccentColorPicker`; `useCustomThemeConfig` reads it internally to override `colorPrimary` per scheme.

**Branding preview mode**: the preview window must show the edited default theme as-is, so `useCustomThemeConfig` ignores the family selection and custom primary color there, and the Theme / Primary Color items are hidden in User Settings while previewing.

## Screenshots

Each built-in theme family applied to the Start page (light mode):

| `default` | `stained` |
|---|---|
| theme-default | theme-stained |

| `glass` | `reverie` |
|---|---|
| theme-glass | theme-reverie |

| `bliss` | |
|---|---|
| theme-bliss | |

## Verification

- `bash scripts/verify.sh`: Relay / Lint / Format / TypeScript **PASS** (`Vite warmup paths` FAIL pre-exists on the parent branch, unrelated to these changes).
- Unit tests updated: `families.default` is ignored in favor of the top-level `light`/`dark` (incl. a stale draft copy in preview mode), synthesized-default coverage, pre-family theme.json backward compat (single-entry catalog), invalid-entry filtering, and per-scheme accent coverage. The full react vitest suite passes.
- Family data (`stained`/`glass`/`reverie`/`bliss`) carried over 1:1 from the previous split file; the `default` family comes from the top-level `light`/`dark` (no duplicated `families.default` entry).


ironAiken2:
* **#8104** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8104?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8104?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8104).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/hooks/useDefaultTheme.ts
```diff
@@ -0,0 +1,86 @@
...
+  );
+
+  // Initialize the draft from theme.json on first load.
+  // Note: useBAISettingUserState returns null (not undefined) when
+  // localStorage has no value.
+  const initializeDefaultTheme = useEffectEvent(() => {
+    if (_.isNil(defaultTheme) && !_.isNil(rawThemeConfig)) {
+      setDefaultTheme(_.cloneDeep(rawThemeConfig));
+    } else if (_.isNil(rawThemeConfig)) {
+      setDefaultTheme(undefined);
```
agatha197: **Potential data-loss bug in the init effect**

The `else if` branch here unconditionally resets `defaultTheme` to `undefined` whenever `rawThemeConfig` hasn't loaded yet — regardless of whether `defaultTheme` already holds a persisted value:

```ts
const initializeDefaultTheme = useEffectEvent(() => {
  if (_.isNil(defaultTheme) && !_.isNil(rawThemeConfig)) {
    setDefaultTheme(_.cloneDeep(rawThemeConfig));
  } else if (_.isNil(rawThemeConfig)) {
    setDefaultTheme(undefined); // wipes an existing draft
  }
});
```

`rawThemeConfig` starts as `undefined` on every mount until the async `theme.json` fetch resolves, so this branch fires on essentially every page load/refresh — wiping any already-persisted Branding draft (`custom_theme_config`) before it gets reseeded from the raw theme once the fetch completes. An admin who has customized branding could see their draft reset on a refresh that happens to race the fetch.

Note: this exact logic (with `defaultTheme`/`rawThemeConfig` renamed from `userCustomThemeConfig`/`themeConfig`) already existed in the deleted `useUserCustomThemeConfig.ts`, so it's not a regression introduced by this PR — but since this PR rewrites the hook anyway, it's a good opportunity to fix it.

**Suggested fix** — drop the `else if` branch; the "seed once" intent in the doc comment is fully covered by the first branch:

```ts
const initializeDefaultTheme = useEffectEvent(() => {
  if (_.isNil(defaultTheme) && !_.isNil(rawThemeConfig)) {
    setDefaultTheme(_.cloneDeep(rawThemeConfig));
  }
});
```

`rawThemeConfig` only ever transitions `undefined -> defined` once per session (via the `custom-theme-loaded` event) and never reverts, so the `else if` branch only ever ran during the initial async-loading window anyway. Removing it doesn't change first-load-with-no-draft behavior — it just stops clobbering an existing draft.

config.toml.sample
```diff
@@ -35,6 +35,7 @@ directoryBasedUsage = false             # If true, display the amount of usage p
 isDirectorySizeVisible = true           # If false, directory size in folder explorer will show `-`. default value is set to true.
 maxCountForPreopenPorts = 10            # The maximum allowed number of preopen ports. If you set this option to 0, the feature of preopen ports is disabled.
 allowCustomResourceAllocation = true    # If true, display the custom allocation on the session launcher.
+allowThemeMode = false                  # If true, users can change the theme mode and the theme's primary color in user settings.
```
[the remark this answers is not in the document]
ironAiken2: Fixed in 3dbab5c4b — the comment now describes what the flag actually gates (the theme family selector and the primary color picker) and explicitly notes that the light/dark theme mode selector remains available regardless of this flag.

react/src/hooks/useDefaultTheme.ts
ironAiken2: Fixed in 3dbab5c4b — dropped the `else if` branch as suggested. The seed-once intent is fully covered by the first branch, and the persisted branding draft is no longer wiped while `theme.json` is still loading.

agatha197 (changes requested):
Please unify theme.json with theme-families as a one.

react/src/hooks/useCustomThemeConfig.tsx
```diff
@@ -0,0 +1,207 @@
...
+  rawConfig: CustomThemeConfig | undefined,
+  remoteFamilies?: Record<string, ThemeFamilyConfig>,
+): ThemeFamilyCatalog =>
+  rawConfig
+    ? {
+        [DEFAULT_THEME_FAMILY]: {
+          light: rawConfig.light,
+          dark: rawConfig.dark,
+        },
+        ...pickValidThemeFamilies(rawConfig.families),
```
agatha197: 🔴 **must-fix (regression): branding color/font edits no longer reach the applied theme or the preview.**

The synthesized `default` is spread first, then `...pickValidThemeFamilies(rawConfig.families)` on top. Since this PR moves the built-in families into `theme.json`'s `families` block, the shipped `resources/theme.json` now contains a **`families.default` entry** (`['default','stained','glass','reverie','bliss']`). That entry is valid, so it passes `pickValidThemeFamilies` and **overrides the synthesized default.** As a result `activePair = families['default']` and `themeConfig.light/dark` come from `families.default`, not from the top-level `light`/`dark`.

But the admin **Branding editor** still reads/writes **only the top-level `light`/`dark`**: `ThemeColorPicker` → `updateDefaultTheme('light.…')`, `FontFamilySettingItem` → `light.token.fontFamily`, and `useDefaultTheme` get/reset all operate on the top level. Edits land in `draft.light`, but resolution reads `draft.families.default.light` (a static copy the editor never touches), so the edits are **lost**.

Repro: on the Branding page change light `colorPrimary` `#FF7A00`→`#00AA00`, click Preview → the preview still shows the old `#FF7A00` from `families.default`. Same for font-family and every light/dark token (logo still works, preserved via `...rawConfig`).

At base (cb3bb94) `families` had no `default` entry, so the synthesized default survived and the branding editor worked; this PR adds `families.default`, which under the unchanged spread order silently breaks it — a regression.

Fix options (pick per design intent): ① drop the `default` key from `theme.json`'s `families` (leave it synthesized); ② adjust the merge order so the synthesized default always wins; ③ make the Branding editor target `families.default`.

react/src/hooks/useCustomThemeConfig.tsx
```diff
@@ -0,0 +1,207 @@
...
+ *   `theme.json` `families` (including its own `default` entry)
+ *   -> admin-defined (`remoteFamilies`, future backend store).
+ * A `default` entry synthesized from the top-level `light`/`dark` is spread
+ * first, so it only survives when `families` does not define its own
+ * `default` — this keeps a pre-family theme.json (no `families` block)
+ * working as a single-entry catalog.
+ */
+export const resolveThemeFamilyCatalog = (
+  rawConfig: CustomThemeConfig | undefined,
+  remoteFamilies?: Record<string, ThemeFamilyConfig>,
```
agatha197: nit (YAGNI/scope): no caller in the tree passes `remoteFamilies` — the only call site, `useCustomThemeConfig`, invokes it with a single argument. It's meant for a future backend store, but until that lands it widens this pure function's public signature and the doc comment describes a merge order the code can't exercise. Suggest adding it when the backend-sync store actually arrives (the doc already points to a separate spec).

react/src/components/ThemeAccentColorPicker.tsx
```diff
@@ -2,47 +2,96 @@
...
+  const familyPair = themeFamilies[activeThemeFamily];
+  const familyLight =
+    familyPair?.light?.token?.colorPrimary ??
+    theme.getDesignToken({ algorithm: theme.defaultAlgorithm }).colorPrimary;
+  const familyDark =
+    familyPair?.dark?.token?.colorPrimary ??
+    theme.getDesignToken({ algorithm: theme.darkAlgorithm }).colorPrimary;
+
+  const setSchemeAccent = (scheme: 'light' | 'dark', color?: string) => {
+    const next = _.omitBy({ ...accent, [scheme]: color }, _.isUndefined);
```
agatha197, line 34: nit (test gap): `setSchemeAccent`'s per-scheme clear logic — clearing one scheme drops just that key, clearing the last scheme deletes the whole setting (`undefined`, so the changed badge / theme falls back) — has no test on this write path. `useCustomThemeConfig.test.tsx` only covers the read side (light-only / both-scheme), and there's no `ThemeAccentColorPicker` test. The code itself is correct (`omitBy` + `isEmpty` handling is right); adding a regression case is recommended.

react/src/components/ThemeAccentColorPicker.tsx
```diff
@@ -2,47 +2,96 @@
...
+    >
+      <Row gutter={[16, 4]}>
+        <Col xl={6} lg={24}>
+          <BAIFlex
+            gap="sm"
+            style={{ color: token.colorTextTertiary }}
+            wrap="wrap"
+          >
+            {t('userSettings.LightMode')}:
+            <ColorPicker
```
agatha197: nit (reuse): this light/dark two-column `ColorPicker` layout (Row/Col + labels + `disabledAlpha allowClear`) is nearly identical to the sibling `BrandingSettingItems/ThemeColorPicker` this PR also touches. They differ only in the write target (here: `custom_primary_color` scheme override; there: `useDefaultTheme` paths). Extracting a shared presentational component (`<LightDarkColorPicker>`) and injecting the persistence strategy would reduce future drift.

resources/theme.schema.json
```diff
@@ -17,7 +17,7 @@
     },
     "families": {
       "type": "object",
-      "description": "Additional named theme families beyond the built-in 'default' family (synthesized from the top-level light/dark). Keyed by family id (e.g. 'stained', 'glass').",
+      "description": "Selectable theme families, keyed by family id (e.g. 'default', 'stained', 'glass'). The 'default' entry is what users see before choosing a family; when it is absent (or this block is omitted entirely, as in pre-family theme.json files), the top-level light/dark is shown instead and family selection is unavailable.",
```
agatha197: nit (inaccurate doc): "the 'default' entry ... when it is absent ... family selection is unavailable" — `family selection is unavailable` is only true when the **entire `families` block is omitted**. `resolveThemeFamilyCatalog` always spreads a synthesized `default` first, so defining e.g. `stained`+`glass` without a `default` entry still yields a catalog of length 3, and `UserSettingsPage` gates the selector on `Object.keys(families).length > 1`, so the selector IS shown. Suggest scoping the clause to "when the block itself is omitted."

resources/i18n/mn.json
```diff
@@ -3386,6 +3388,8 @@
     "SessionTerminationDialog": "Session дуусахыг цуцлах боломжгүй. Та хуралдааныг зогсоохыг хүсч байна уу?",
     "ShellEnvironments": "Shell Environments",
     "Theme": "Сэдэв",
+    "ThemeAccentColor": "Үндсэн өнгөний",
```
agatha197: nit (i18n, low confidence): the Mongolian `ThemeAccentColor` value `"Үндсэн өнгөний"` ends in the genitive suffix -ий (~'of the color'). A setting-row title should be nominative `"Үндсэн өнгө"` (the sibling `DescThemeAccentColor` correctly uses the accusative `өнгийг`). Cosmetic only — it still renders.

agatha197 (commented):
## Review: FR-3239 user theme selector refactor

Multi-agent review (5 dimensions in parallel → adversarial verification). Leaving the 6 confirmed findings inline.

- 🔴 **1 must-fix**: `families.default` overrides the synthesized default, so **branding color/font edits and preview are nullified** (`useCustomThemeConfig.tsx`) — a regression. See inline.
- 🟡 5 nit/low: `remoteFamilies` YAGNI, `setSchemeAccent` test gap, ColorPicker layout duplication, inaccurate schema description, mn.json inflection.

**Verified OK**: the 4 theme i18n keys exist in all 21 languages, the removed `AccentLowContrastWarning` is cleaned up with no orphans, and the `allowThemeMode` gating / `useDefaultTheme` raw seeding / `setActiveThemeFamily(undefined)` clear paths are logically correct.

Happy to approve once the must-fix is resolved.

react/src/components/ReverseThemeProvider.tsx
```diff
@@ -13,7 +13,7 @@ interface ReverseThemeProviderProps extends ConfigProviderProps {
 const ReverseThemeProvider: React.FC<ReverseThemeProviderProps> = ({
   ...props
 }) => {
-  const themeConfig = useCustomThemeConfig();
+  const { themeConfig } = useCustomThemeConfig();
```
agatha197, line 17: This component body was touched in this PR but has no `'use memo'` directive. Please add `'use memo';` as the first statement (consistent with `ThemeAccentColorPicker`).

react/src/components/ThemeSecondaryProvider.tsx
```diff
@@ -16,7 +16,7 @@ import React, { useContext } from 'react';
 const ThemeSecondaryProvider: React.FC<ConfigProviderProps> = ({
   ...props
 }) => {
-  const themeConfig = useCustomThemeConfig();
+  const { themeConfig } = useCustomThemeConfig();
```
agatha197, line 20: Please add `'use memo'` — the body was touched here without the directive.

react/src/components/ThemeAdminProvider.tsx
```diff
@@ -14,7 +14,7 @@ import * as _ from 'lodash-es';
 import React, { useContext } from 'react';
 
 const ThemeAdminProvider: React.FC<ConfigProviderProps> = ({ ...props }) => {
-  const themeConfig = useCustomThemeConfig();
+  const { themeConfig } = useCustomThemeConfig();
```
agatha197, line 18: Please add `'use memo'` — the body was touched here without the directive.

react/src/hooks/useCustomThemeConfig.tsx
ironAiken2: Fixed in cad2c15f8 — `resolveThemeFamilyCatalog` now synthesizes the `default` family from the top-level `light`/`dark` and wraps the merged `families`/`remoteFamilies` layers in `_.omit(..., DEFAULT_THEME_FAMILY)`, so a `families.default` entry can never shadow the branding document. Additionally, bcf672f8f consolidated the built-in families back into `theme.json` and the shipped `families` block no longer contains a `default` key at all (only stained/glass/reverie/bliss). Branding color/font edits and preview now resolve from the top-level document again.

resources/theme.schema.json
ironAiken2: Fixed in bcf672f8f — the schema description now scopes the clause as suggested: the `default` family is documented as always synthesized from the top-level `light`/`dark` (a `default` key here is ignored), and "family selection is unavailable" now applies only "when this block is omitted entirely".

react/src/hooks/useCustomThemeConfig.tsx
ironAiken2: Fixed in 2ff76dbfa — dropped the unused `remoteFamilies` parameter; `resolveThemeFamilyCatalog` now takes only `rawConfig` and the doc comment describes just the shipped merge (synthesized `default` + `theme.json` `families`). The parameter can be reintroduced when the backend-sync store actually lands.

react/src/components/ThemeAccentColorPicker.tsx
ironAiken2: Fixed in 2ff76dbfa — added `ThemeAccentColorPicker.test.tsx` covering the write path: merging a new scheme color into an existing accent, clearing one scheme keeps the other key, clearing the last overridden scheme drops the whole setting (`undefined`), plus the family-color fallback when no accent is set.

react/src/components/ThemeAccentColorPicker.tsx
ironAiken2: Fixed in 2ff76dbfa — extracted the shared presentational `LightDarkColorPicker` (Row/Col + scheme labels + `ColorPicker` defaults) and injected the persistence strategy per scheme via props. Both `ThemeAccentColorPicker` (accent user setting) and `BrandingSettingItems/ThemeColorPicker` (`useDefaultTheme` paths) now render through it.

resources/i18n/mn.json
ironAiken2: Fixed in 2ff76dbfa — changed the Mongolian `ThemeAccentColor` title to nominative `"Үндсэн өнгө"` (the genitive `өнгөний` was a leftover); `DescThemeAccentColor` keeps its accusative form.

react/src/components/ReverseThemeProvider.tsx
ironAiken2: Fixed in 2ff76dbfa — added `'use memo';` as the first statement of the component body.

react/src/components/ThemeSecondaryProvider.tsx
ironAiken2: Fixed in 2ff76dbfa — added `'use memo';` as the first statement of the component body.

react/src/components/ThemeAdminProvider.tsx
ironAiken2: Fixed in 2ff76dbfa — added `'use memo';` as the first statement of the component body.

agatha197 (approved):
LGTM

yomybaby (approved):
LGTM

PR #8119, opened 2026-07-02
feat(FR-3243): add SFTP resource group settings from the Resource Group page

### LINKED ISSUE #8113: Add SFTP scaling group settings to Storage Proxy page
## Summary

Add the ability to configure per-storage-proxy **SFTP scaling groups** directly from the Storage Proxy list page, so superadmins no longer need to drop to the manager etcd CLU.

## Background

An "SFTP scaling group" designates, per storage proxy, which scaling group(s) handle SFTP (file up/download) sessions. It is stored in etcd, per proxy:

- etcd key: `volumes/proxies/{proxy}/sftp_scaling_groups`
- value: comma-joined scaling-group names, e.g. `"group-1,group-2"`

Today this is only settable via the manager CLI (`backend.ai mgr etcd set-storage-sftp-scaling-group <proxy> "g1,g2"`). There is no GraphQL mutation and the StorageVolume GraphQL type does not expose the field.

## Requirements

1. On the Storage Proxy list page, add an overflow (`⋮` more) dropdown button next to the existing refresh button, containing an **'SFTP Scaling Group Settings'** menu item (mirrors the Diagnostics page ⋮ menu / Export CSV pattern).
1. Clicking the menu item opens a modal with an antd Form containing two Form.Items:
   - a **storage proxy select** (choose the proxy to configure)
   - a **scaling group select** (`BAIResourceGroupSelect`, multiple/tags mode — a proxy may map to several scaling groups)
1. On submit, persist the selection to etcd at `volumes/proxies/{proxy}/sftp_scaling_groups` as a comma-joined string.
1. When a proxy is selected, prefill the currently-configured scaling groups so the modal shows existing state.
1. Superadmin-only (the underlying `/config/set` etcd endpoint requires superadmin); gate the menu item visibility accordingly.

## Implementation notes (frontend-buildable)

- No GraphQL mutation / no clean client wrapper exists. The webui `Setting.set/get` wrappers force-prefix `config/` and cannot reach the `volumes/...` subtree.
- Correct write: raw signed request `POST /config/set` with `{ key: 'volumes/proxies/{proxy}/sftp_scaling_groups', value: groups.join(',') `}. Read/prefill: `POST /config/get` (same raw key). Empty selection -> `POST /config/delete` (or set empty string).
- Verified against backend: the `/config/set` handler writes via the same config-scoped etcd the CLI uses, so the raw `volumes/...` key targets the exact same location.
- Proxy list source: derive distinct `proxy` values from the existing `storage_volume_list` query.

## i18n

Add keys for the menu item, modal title, form labels, and success/error messages across all supported locales.

JIRA Issue: FR-3243

agatha197:
Resolves #8113 (FR-3243)

## Summary

Lets superadmins manage, from the **Resource Group** management page, which **storage proxies** a resource group (scaling group) is allowed to use for **SFTP** file transfer. This is the per-proxy etcd allow-list `volumes/proxies/{proxy}/sftp_scaling_groups` (comma-joined); an unset key means "all groups". Every write reads the proxy's current list first and merges, so other groups already on a proxy are never clobbered. Two complementary entry points:

### 1. Bulk — "Update Resource Groups" modal (from the Resource Group list)
- Checkbox multi-select resource-group rows (superadmin only) → a `N selected` label + a gear **Settings** button appears (same pattern as `ProjectStoragePermissionTable`).
- The gear opens a generic **Update Resource Groups** modal: an info alert lists the selected groups, followed by the settings fields (currently just **SFTP Storage Proxies**; the modal is structured so more bulk-editable settings can be added as sibling fields later).
- The proxy field is prefilled with the union of proxies already serving any selected group. On submit, **every selected group ends up on every chosen proxy** (idempotent union-add computed from the full per-proxy map — a proxy already serving one selected group no longer skips the others). Deselecting a proxy removes the selected groups from it, behind a typed confirm.

### 2. Per group (from the Resource Group create/edit modal)
- Superadmin-only **SFTP Storage Proxies** multi-select, **prefilled** with the proxies that currently serve this group.
- On save, this group's membership is **synced** to the selected proxies: added to newly-selected proxies, removed from deselected ones — only *this* group is touched, every other group on each proxy is preserved.

## Implementation

- Shared `useSFTPResourceGroups` hook centralizes the etcd read-merge-write via `applyGroupProxyDelta` (add groups to / remove groups from a set of proxies at once). When a proxy's list becomes empty it **deletes** the etcd key instead of writing `''` — the manager's comma-parser turns `''` into `['']` (a bogus empty-named group), whereas a deleted key reads back as the documented "all groups" default.
- No GraphQL mutation exists for this etcd key, and `Setting.set/get` force-prefix `config/`, so reads/writes use raw superadmin signed requests to `/config/{get,set,delete}`.
- Bulk modal is `UpdateResourceGroupsModal` (generic title + info alert + per-field labels/tooltips); per-group field lives in `ResourceGroupSettingModal`.
- New reusable `BAIStorageProxySelect` in `backend.ai-ui` (+ Storybook), deriving distinct proxy names from `storage_volume_list` (with a `TODO(needs-backend)` for a dedicated field).
- Copy is permission-framed ("groups a proxy is allowed to use for SFTP"): the `sftp_scaling_groups` setting is a per-proxy allow-list and is advisory on the manager side (exposed via `vfolder.list_hosts()`, not gated on compute-session scheduling — that stays a separate `allowed_session_types` concern).
- New `storageProxy` i18n keys + `resourceGroup.UpdateResourceGroups`, across all 21 locales.

## Screenshots

**Single-group edit** — `Modify Resource Group` modal with the SFTP Storage Proxies field:

single update modal

**Bulk update** — `Update Resource Groups` modal (proxy prefilled from the selected groups):

bulk update modal

**Result** — submitting the bulk modal applies the proxy to **all** selected groups (here `test6` / `test2` gained `local` even though `test4` already had it):

bulk update result

## Verification

```
Relay / Lint / Format / Vite warmup: PASS
```

Manually verified end-to-end against a live backend (superadmin):
- **Single update** — set a proxy on one group via the edit modal; the row reflects it.
- **Bulk update** — selected 3 groups (one already had the proxy), submitted the prefilled modal; **all** selected groups received the proxy.
- **Remove** — cleared the proxy in the bulk modal; the typed confirm fires and the groups drop it.

**Checklist:**

- [ ] Documentation
- [x] Minimum required manager version — none; uses existing `/config/{get,set,delete}` REST (superadmin), no schema change
- [x] Minimum requirements to check during review — both flows preserve other groups on a proxy (merge / delta); the per-group sync only removes *this* group from deselected proxies; empty list deletes the key rather than writing `''`
- [x] Test case(s) to demonstrate the difference of before/after — manual single + bulk + remove flows verified (screenshots above)

### DIFF
--- a/react/src/components/ResourceGroupSettingModal.tsx
+++ b/react/src/components/ResourceGroupSettingModal.tsx
@@ -7,7 +7,13 @@ import { ResourceGroupSettingModalCreateMutation } from '../__generated__/Resour
 import { ResourceGroupSettingModalFragment$key } from '../__generated__/ResourceGroupSettingModalFragment.graphql';
 import { ResourceGroupSettingModalUpdateMutation } from '../__generated__/ResourceGroupSettingModalUpdateMutation.graphql';
 import { newLineToBrElement } from '../helper';
-import { useCurrentDomainValue } from '../hooks';
+import { useCurrentDomainValue, useSuspendedBackendaiClient } from '../hooks';
+import {
+  computeProxyDelta,
+  proxiesServingGroups,
+  useSFTPProxyResourceGroupsQuery,
+  useSFTPResourceGroups,
+} from '../hooks/useSFTPResourceGroups';
 import { ScalingGroupOpts } from './ResourceGroupList';
 import { QuestionCircleOutlined } from '@ant-design/icons';
 import {
@@ -31,9 +37,11 @@ import {
   BAICard,
   BAIFlex,
   BAIDomainSelect,
+  BAIStorageProxySelect,
+  useBAILogger,
 } from 'backend.ai-ui';
 import * as _ from 'lodash-es';
-import { Suspense, useMemo, useRef } from 'react';
+import React, { Suspense, useDeferredValue, useMemo, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import { graphql, useFragment, useMutation } from 'react-relay';
 
@@ -49,6 +57,7 @@ type FormInputType = {
   scheduler: string;
   pendingTimeout: number;
   retriesToSkip: number;
+  sftpProxies: string[];
 };
 
 interface ResourceGroupCreateModalProps extends ModalProps {
@@ -61,10 +70,14 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
   onRequestClose,
   ...modalProps
 }) => {
+  'use memo';
   const { t } = useTranslation();
   const { token } = theme.useToken();
   const { message } = App.useApp();
+  const { logger } = useBAILogger();
+  const baiClient = useSuspendedBackendaiClient();
   const currentDomain = useCurrentDomainValue();
+  const { applyGroupProxyDelta } = useSFTPResourceGroups();
   const formRef = useRef<FormInstance>(null);
 
   const resourceGroup = useFragment(
@@ -82,6 +95,23 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
     `,
     resourceGroupFrgmt,
   );
+
+  // Superadmin-only SFTP proxy membership for this group, read from etcd without
+  // suspending: gate on a deferred `open` so enabling runs in a transition. The
+  // field below skeletons until the map resolves, then mounts with its prefill
+  // as the Form.Item `initialValue` — the rest of the form renders immediately.
+  const deferredOpen = useDeferredValue(modalProps.open);
+  const { data: proxyResourceGroups, isFetching: isSftpMapFetching } =
+    useSFTPProxyResourceGroupsQuery({
+      enabled: baiClient.is_superadmin && !!deferredOpen,
+    });
+  const currentSftpProxies = proxiesServingGroups(
+    proxyResourceGroups,
+    resourceGroup?.name ? [resourceGroup.name] : [],
+  );
+  const isSftpMapLoading =
+    proxyResourceGroups === undefined || isSftpMapFetching;
+
   const [commitUpdateResourceGroup, isInFlightCommitUpdateResourceGroup] =
     useMutation<ResourceGroupSettingModalUpdateMutation>(graphql`
       mutation ResourceGroupSettingModalUpdateMutation(
@@ -192,6 +222,40 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
               wsproxy_api_token: values.wsProxyAPIToken,
             };
 
+            // After the group is saved, sync its SFTP proxy membership to the
+            // selected proxies (add to newly-selected proxies, remove from
+            // deselected ones, preserving other groups on each proxy).
+            // Deselection here removes without a separate confirm — unlike the
+            // bulk modal — because Update is an explicit, per-group edit.
+            // Superadmin-only.
+            const applySftpProxies = (groupName: string): Promise<void> => {
+              if (!baiClient.is_superadmin) {
+                return Promise.resolve();
+              }
+              const { added, removed } = computeProxyDelta(
+                currentSftpProxies,
+                values.sftpProxies ?? [],
+              );
+              return applyGroupProxyDelta([groupName], added, removed).then(
+                (results) => {
+                  const failed = results.filter((r) => r.status === 'rejected');
+                  if (failed.length > 0) {
+                    failed.forEach((r) => {
+                      if (r.status === 'rejected') {
+                        logger.error(
+                          'Failed to update SFTP storage proxies:',
+                          r.reason,
+                        );
+                      }
+                    });
+                    message.error(
+                      t('storageProxy.FailedToUpdateSFTPResourceGroups'),
+                    );
+                  }
+                },
+              );
+            };
+
             if (resourceGroup) {
               commitUpdateResourceGroup({
                 variables: {
@@ -214,7 +278,9 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
                     return;
                   }
                   message.success(t('resourceGroup.ResourceGroupModified'));
-                  onRequestClose?.(true);
+                  applySftpProxies(resourceGroup.name).finally(() =>
+                    onRequestClose?.(true),
+                  );
                 },
                 onError: (err) => {
                   message.error(err.message);
@@ -267,7 +333,9 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
                     }
 
                     message.success(t('resourceGroup.ResourceGroupCreated'));
-                    onRequestClose?.(true);
+                    applySftpProxies(values.name).finally(() =>
+                      onRequestClose?.(true),
+                    );
                   },
                   onError: (err) => {
                     message.error(err.message);
@@ -357,6 +425,28 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
           >
             <Input.Password placeholder={t('resourceGroup.EnterAPIToken')} />
           </Form.Item>
+          {baiClient.is_superadmin ? (
+            isSftpMapLoading ? (
+              <Form.Item label={t('storageProxy.SFTPStorageProxies')}>
+                <Select loading />
+              </Form.Item>
+            ) : (
+              // Wrap the Form.Item (not the select — that would break Form's
+              // value/onChange binding) in Suspense: BAIStorageProxySelect's
+              // proxy-options query suspends. The prefill is registered as the
+              // Form.Item `initialValue`, computed from the now-resolved map.
+              <Suspense fallback={<Select loading />}>
+                <Form.Item
+                  label={t('storageProxy.SFTPStorageProxies')}
+                  name="sftpProxies"
+                  initialValue={currentSftpProxies}
+                  tooltip={t('storageProxy.SFTPStorageProxiesDescription')}
+                >
+                  <BAIStorageProxySelect mode="multiple" allowClear />
+                </Form.Item>
+              </Suspense>
+            )
+          ) : null}
           <Row>
             <Col span={12}>
               <Form.Item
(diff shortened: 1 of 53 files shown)


agatha197:
* **#8119** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8119?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8119?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8119).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/SFTPScalingGroupSettingModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+            ? value
+                .split(',')
+                .map((group) => group.trim())
+                .filter(Boolean)
+            : [];
+        form.setFieldValue('scaling_groups', groups);
+      })
+      .catch(() => {
+        form.setFieldValue('scaling_groups', []);
+      });
```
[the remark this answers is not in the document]
agatha197: Fixed: `handleProxyChange` now checks `form.getFieldValue('proxy') !== proxy` before applying the response, so a stale (slower) response for a previously-selected proxy can no longer overwrite the currently-selected proxy's scaling groups.

react/src/components/SFTPScalingGroupSettingModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+          })
+          .catch(() => {
+            message.error(t('storageProxy.FailedToUpdateSFTPResourceGroups'));
+          })
+          .finally(() => {
+            setIsSubmitting(false);
+          });
+      })
+      .catch(() => {});
+  };
```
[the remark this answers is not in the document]
agatha197: Fixed: the empty catch now logs via `useBAILogger` (`logger.error('SFTP scaling group form validation failed:', error)`) instead of silently swallowing the validation rejection.

react/src/components/SFTPScalingGroupSettingModal.tsx
```diff
@@ -0,0 +1,182 @@
...
+            items {
+              id
+              proxy
+            }
+          }
+        }
+      `,
+      { limit: 100, offset: 0 },
+      { fetchPolicy: 'store-and-network' },
+    );
```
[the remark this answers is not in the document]
agatha197: Fixed, and also extracted the proxy select into a reusable `BAIStorageProxySelect` component in backend.ai-ui. It now fetches with a limit of 1000 volumes (was 100) — in practice no deployment approaches that many storage volumes, so every distinct proxy is guaranteed to surface.

agatha197:
Follow-up review pass (independent, on Sonnet) surfaced one real issue Copilot's round missed: in `handleProxyChange`'s `/config/get` `.catch()`, any genuine request failure (network/permission/5xx) was silently treated the same as "key doesn't exist yet" and reset the resource-groups field to empty with no indication to the user — a subsequent Save would then fire `/config/delete` and wipe a real, previously-configured setting. Fixed: the catch now logs via `useBAILogger` and shows `message.error(...)` so the failure is visible before the user can accidentally submit over it. Also added the missing `'use memo'` directive to `StorageProxyList` per the React Compiler convention, since this PR substantially expands its state/logic.

agatha197:
Ran a quality pass (reuse / simplification / efficiency / altitude, 4 independent reviewers) over the diff. Applied:
- Hoisted `etcdKey` out of the component body (was redefined every render for no reason).
- Deduped the repeated `form.getFieldValue('proxy') !== proxy` staleness check in `handleProxyChange` into one local `isStale()` helper shared by the `.then`/`.catch` branches.
- `BAIStorageProxySelect.tsx`: replaced the hand-written `_.filter(..., (proxy): proxy is string => !!proxy)` type-guard with `_.compact`, matching the established `_.compact(_.map(...))` idiom already used elsewhere in this package (e.g. `BAIProjectBulkEditModal.tsx`).
- Added a `TODO(needs-backend, FR-3243)` marker on the volume→proxy derivation workaround, so a future dedicated proxy-list field is a one-line swap.
- Added a per-proxy in-session cache (`useRef`) in the modal so toggling back and forth between already-visited proxies doesn't re-issue a `/config/get` request.

Deliberately **not** applied, flagged as a follow-up instead: two independent review angles (reuse + altitude) both pointed out that the raw `useBaiSignedRequestWithPromise` + hardcoded `/config/get|set|delete` calls in the modal reimplement the pattern the existing `Setting`/`Registry` classes in `src/lib/backend.ai-client-node.ts` already establish for other etcd-backed settings modals (`OverlayNetworkSettingModal.tsx`, `SchedulerSettingModal.tsx`, `ConfigurationsSettingList.tsx`), and suggested adding typed methods to the existing `StorageProxy` client class instead. This is architecturally sound, but touches a different workspace package (`packages/backend.ai-client`) with different TS strictness settings that this PR's diff and verification never covered — treating it as out of scope for this PR rather than editing an unverified file. Worth a small follow-up PR.

react/src/components/SFTPScalingGroupSettingModal.tsx
```diff
@@ -0,0 +1,169 @@
...
+                body: { key: etcdKey(proxy), value: groups.join(',') },
+              });
+        return request
+          .then(() => {
+            message.success(t('storageProxy.SFTPResourceGroupsUpdated'));
+            onRequestClose(true);
+          })
+          .catch(() => {
+            message.error(t('storageProxy.FailedToUpdateSFTPResourceGroups'));
+          })
```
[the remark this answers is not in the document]
agatha197: Fixed: the submit-failure catch now calls `logger.error('Failed to update SFTP resource groups:', error)` before showing the user-facing message, matching the fetch and validation handlers in this file.

react/src/hooks/useSFTPScalingGroups.tsx
```diff
@@ -0,0 +1,173 @@
...
+  const queryClient = useQueryClient();
+
+  const getGroupsForProxy = (proxy: string): Promise<string[]> =>
+    baiSignedRequest({
+      method: 'POST',
+      url: '/config/get',
+      body: { key: etcdKey(proxy), prefix: false },
+    }).then((res: { result?: string | null }) => parseGroups(res?.result));
+
+  const setGroupsForProxy = (proxy: string, groups: string[]) =>
```
nowgnuesLee: Let's mark this as returning a Promise here too.
(Personally, I think it'd be cleaner if `baiSignedRequest` were typed as `Promise<any>`.)

nowgnuesLee (changes requested):

react/src/hooks/useSFTPScalingGroups.tsx
agatha197: Added an explicit `: Promise<unknown>` return type to `setGroupsForProxy`, matching `getGroupsForProxy`'s `Promise<string[]>` annotation. `unknown` (rather than `any`) since nothing consumes the `/config/set` response.

On the aside: I left `useBaiSignedRequestWithPromise`'s own return typing untouched here — it's a shared helper used across the codebase, so retyping it to `Promise<any>` felt out of scope for this PR. Happy to file a follow-up if you'd like it tightened project-wide.

agatha197: Follow-up: per the suggestion in the JIRA notification thread, `baiSignedRequestWithPromise` and its `useBaiSignedRequestWithPromise` hook are now generically typed as `<ResponseType = any,>(...) => Promise<ResponseType>`. The SFTP hook now threads the response type through the generic — dropping the `as` cast in `getProxyResourceGroups` and the inline `res` annotation in `getGroupsForProxy`.

This also surfaced and fixed a latent issue in `InviteFolderSettingModal`: its `list_invitees` queryFn returned `null` when the modal was closed (previously masked by the `any` return, now a hard type error). Changed to return `[]`, which is behavior-equivalent (the closed modal renders empty either way).

react/src/components/ResourceGroupSettingModal.tsx
```diff
@@ -357,6 +425,32 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
           >
             <Input.Password placeholder={t('resourceGroup.EnterAPIToken')} />
           </Form.Item>
+          {baiClient.is_superadmin ? (
+            isSftpMapLoading ? (
+              <Form.Item label={t('storageProxy.SFTPStorageProxies')}>
+                <Skeleton.Input active block />
```
nowgnuesLee:
```suggestion
                <Select loading />
```

react/src/components/ResourceGroupSettingModal.tsx
```diff
@@ -357,6 +425,32 @@ const ResourceGroupSettingModal: React.FC<ResourceGroupCreateModalProps> = ({
...
+              </Form.Item>
+            ) : (
+              // Wrap the Form.Item (not the select — that would break Form's
+              // value/onChange binding) in Suspense: BAIStorageProxySelect's
+              // proxy-options query suspends. The prefill is registered as the
+              // Form.Item `initialValue`, computed from the now-resolved map.
+              <Suspense fallback={<Skeleton.Input active block />}>
```
nowgnuesLee:
```suggestion
              <Suspense fallback={<Select loading />}>
```

nowgnuesLee (changes requested):

react/src/components/ResourceGroupSettingModal.tsx
agatha197: Applied — replaced the `Skeleton.Input` loading placeholder with `<Select loading />` so the loading state matches the actual select field. Fixed in latest commit.

react/src/components/ResourceGroupSettingModal.tsx
agatha197: Applied — the Suspense fallback now uses `<Select loading />` to match the resolved `BAIStorageProxySelect`. Fixed in latest commit.

nowgnuesLee (approved):
LGTM

ironAiken2 (approved):
LGTM

PR #8148, opened 2026-07-03
feat(FR-3256): add RuntimeVariantPreset admin CRUD tab

### LINKED ISSUE #8147: Add RuntimeVariantPreset admin CRUD tab to admin-deployments page
## Background

The manager backend (since 26.4.2, `required` flag since 26.4.4) already implements full CRUD for `RuntimeVariantPreset` via GraphQL (Strawberry V2):

- Query: `runtimeVariantPresets(filter, orderBy, limit/offset or cursor pagination)` -> `RuntimeVariantPresetConnection`
- Query: `runtimeVariantPreset(id: UUID!)`
- Mutation: `adminCreateRuntimeVariantPreset(input: CreateRuntimeVariantPresetInput!)` — superadmin only
- Mutation: `adminUpdateRuntimeVariantPreset(input: UpdateRuntimeVariantPresetInput!)` — superadmin only
- Mutation: `adminDeleteRuntimeVariantPreset(id: UUID!)` — superadmin only

`RuntimeVariantPreset` fields: `id`, `runtimeVariantId`, `name`, `description`, `rank`, `targetSpec { presetTarget (ENV/ARGS), valueType (STR/INT/FLOAT/BOOL/FLAG), defaultValue, key `}, `required` (26.4.4+), `category`, `displayName`, `uiOption`, `createdAt`, `updatedAt`.

The WebUI currently only **reads** these presets (`useRuntimeParameterSchema` hook, used to render deployment revision forms). There is no admin UI to create/edit/delete `RuntimeVariantPreset` records — today this requires direct API/DB access.

## Scope

Add a new tab to the existing admin Deployments page (`react/src/pages/AdminDeploymentListPage.tsx`, which already hosts tabs like Deployments / Model Store Management / Prometheus Preset / Deployment Presets via `BAICard` `tabList`) called **"Runtime Variant Presets"**.

1. **List page** — `BAITable` bound to the `runtimeVariantPresets` V2 connection, `limit`/`offset` pagination, filter by name / runtime variant, sortable by name/rank/createdAt.
1. **Create/Edit modal** (not a separate page) with fields: runtime variant (select), name, description, preset target (ENV/ARGS), value type (STR/INT/FLOAT/BOOL/FLAG), default value, key, required (toggle, gated behind manager 26.4.4 compatibility).
1. **Delete** via `BAIConfirmModalWithInput` (typed confirmation — permanent deletion, not `Popconfirm`, per project convention).
1. **Feature-gate** the whole tab behind manager version compatibility (>= 26.4.2), matching the existing pattern for other feature-gated tabs (e.g. `isDeploymentPresetSupported`) in `AdminDeploymentListPage`.
1. Full i18n across en/ko/ja/th under a new `adminRuntimeVariantPreset.*` namespace.

## Out of scope

`category`, `displayName`, `uiOption` are read-only on the type and not exposed on `Create`/`UpdateRuntimeVariantPresetInput` — editing them is out of scope here (possible backend follow-up).

JIRA Issue: FR-3256

agatha197:
Stacked on #8336 (FR-3256).

## Summary

Adds the **runtime-variant-presets** tab to the deployment-admin page:

- `AdminRuntimeVariantPreset` — the tab view, wired into `AdminDeploymentPage` as a hoisted `preloadedQuery` tab consistent with the other tabs.
- `BAIRuntimeVariantPresetTable` — the fragment table (owns `<BAITable>`, named per the FR-3256 table/role convention; renamed from `…Nodes`).
- `BAIRuntimeVariantPresetSettingModal` — create/edit modal. Its Suspense fallback uses `BAISelect`.

## Test plan
- [x] ESLint clean


agatha197:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8148?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#8148** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8336** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8336?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8258** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8148).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/pages/AdminDeploymentListPage.tsx
```diff
@@ -543,6 +544,10 @@ const AdminDeploymentListPage: React.FC = () => {
       key: 'deployment-presets',
       label: t('adminDeploymentPreset.TabTitle'),
     },
+    {
+      key: 'runtime-variant-presets',
+      label: t('adminRuntimeVariantPreset.TabTitle'),
+    },
```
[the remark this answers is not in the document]
agatha197: Intentional: the manager-version gate for this tab was deliberately removed at the maintainer's request — the supported manager baseline for this build already includes the `runtimeVariantPresets` query (available since 26.4.2), so a separate `baiClient.supports()` gate is redundant here. The field-level `required` toggle is still gated behind the existing `runtime-variant-preset-required` flag where a genuine version difference exists.

react/src/pages/AdminDeploymentListPage.tsx
```diff
@@ -578,6 +583,11 @@ const AdminDeploymentListPage: React.FC = () => {
             <AdminDeploymentPresetListPage />
           </BAIErrorBoundary>
         )}
+        {currentTab === 'runtime-variant-presets' && (
+          <BAIErrorBoundary>
+            <AdminRuntimeVariantPresetListPage />
+          </BAIErrorBoundary>
+        )}
```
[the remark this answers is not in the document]
agatha197: Same as above — the tab-content render is intentionally ungated. The version gate was removed by design; the supported manager baseline guarantees the `runtimeVariantPresets` query exists.

react/src/pages/AdminRuntimeVariantPresetListPage.tsx
```diff
@@ -0,0 +1,319 @@
...
+          limit: $limit
+          offset: $offset
+        ) {
+          count
+          edges {
+            node {
+              id
+              name
+              ...AdminRuntimeVariantPresetNodesFragment
+            }
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit — removed the explicit `id`/`name` selections on the `runtimeVariantPresets.edges.node` since the orchestrator only forwards the fragment ref to the child component and never reads those fields directly. The node now selects only `...AdminRuntimeVariantPresetNodesFragment`.

react/src/components/AdminRuntimeVariantPresetNodes.tsx
```diff
@@ -0,0 +1,298 @@
...
+              {
+                key: 'delete',
+                title: t('button.Delete'),
+                icon: <DeleteFilled />,
+                type: 'danger' as const,
+                onClick: () => onDelete?.(preset),
+              },
+            ]}
+          />
+        ),
```
nowgnuesLee: Please let the consumer implement this part.

react/src/components/AdminRuntimeVariantPresetNodes.tsx
nowgnuesLee, line 1: Let's move this component to bui.

react/src/components/RuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,370 @@
...
+        targetSpec {
+          presetTarget
+          valueType
+          defaultValue
+          key
+        }
+        required @since(version: "26.4.4rc9")
+      }
+    `,
+    presetFrgmt ?? null,
```
nowgnuesLee:
```suggestion
    presetFrgmt,
```

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,370 @@
...
+
+  const [commitUpdate, isInFlightUpdate] =
+    useMutation<RuntimeVariantPresetSettingModalUpdateMutation>(graphql`
+      mutation RuntimeVariantPresetSettingModalUpdateMutation(
+        $input: UpdateRuntimeVariantPresetInput!
+      ) {
+        adminUpdateRuntimeVariantPreset(input: $input) {
+          preset {
+            id
+          }
```
nowgnuesLee, line 130: In the update mutation, please include all fields so the node updates without needing a refetch.

react/src/components/RuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,370 @@
...
+          label={t('adminRuntimeVariantPreset.PresetTarget')}
+          name="presetTarget"
+          rules={[
+            {
+              required: true,
+              message: t('adminRuntimeVariantPreset.PresetTargetRequired'),
+            },
+          ]}
+        >
+          <Select
```
nowgnuesLee: Let's use BAISelect.

react/src/components/RuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,370 @@
...
+          label={t('adminRuntimeVariantPreset.ValueType')}
+          name="valueType"
+          rules={[
+            {
+              required: true,
+              message: t('adminRuntimeVariantPreset.ValueTypeRequired'),
+            },
+          ]}
+        >
+          <Select
```
nowgnuesLee: Let's use BAISelect.

react/src/components/AdminRuntimeVariantPreset.tsx
nowgnuesLee, line 1: Please use usePreloadedQuery instead of useLazyLoadQuery.

react/src/pages/AdminRuntimeVariantPresetListPage.tsx
```diff
@@ -0,0 +1,302 @@
...
+  const baiClient = useSuspendedBackendaiClient();
+
+  const [columnOverrides, setColumnOverrides] = useBAISettingUserState(
+    'table_column_overrides.AdminRuntimeVariantPresetListPage',
+  );
+
+  const [deletingPreset, setDeletingPreset] =
+    useState<RuntimeVariantPresetNodeInList | null>(null);
+  const [editingPresetKey, setEditingPresetKey] =
+    useState<RuntimeVariantPresetSettingModalFragment$key | null>(null);
```
nowgnuesLee: You probably don't need this key type

react/src/pages/AdminRuntimeVariantPresetListPage.tsx
```diff
@@ -0,0 +1,302 @@
...
+    setTablePaginationOption,
+  } = useBAIPaginationOptionStateOnSearchParam({
+    current: 1,
+    pageSize: 10,
+  });
+
+  const [queryParams, setQueryParams] = useQueryStates(
+    {
+      order: parseAsString.withDefault('rank'),
+      filter: parseAsJson<GraphQLFilter>((value) => value as GraphQLFilter),
```
nowgnuesLee: Please use RuntimeVariantPresetFilter type.

react/src/pages/AdminRuntimeVariantPresetListPage.tsx
```diff
@@ -0,0 +1,302 @@
...
+  };
+
+  const deferredQueryVariables = useDeferredValue(queryVariables);
+  const deferredFetchKey = useDeferredValue(fetchKey);
+
+  // `RuntimeVariantPreset` only exposes `runtimeVariantId` (no nested
+  // `runtimeVariant` field), so the query below also fetches `runtimeVariants`
+  // (unfiltered, small curated catalog e.g. vllm/sglang/nim/...) to resolve
+  // names for the table's "Runtime Variant" column.
+  const queryRef = useLazyLoadQuery<AdminRuntimeVariantPresetListPageQuery>(
```
nowgnuesLee: where did u spread RuntimeVariantPresetSettingModal fragment?

nowgnuesLee (changes requested):

react/src/components/AdminRuntimeVariantPresetNodes.tsx
agatha197: Fixed — row actions (edit/delete) are no longer hardcoded in the table component. The table moved to `backend.ai-ui` as `BAIRuntimeVariantPresetNodes`, which now exposes `customizeColumns` only; the host page (`AdminRuntimeVariantPresetListPage`) appends the actions column itself via `customizeColumns={(base) => [...base, actionsColumn]}`.

react/src/components/AdminRuntimeVariantPresetNodes.tsx
agatha197: Done — moved to `packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetNodes.tsx`, following the `BAIRouteNodes` conventions (`useBAIi18n()` with `comp:BAIRuntimeVariantPresetNodes.*` keys, `useConnectedBAIClient()`, exported from `fragments/index.ts`). `react/src/components/AdminRuntimeVariantPresetNodes.tsx` was deleted.

react/src/components/RuntimeVariantPresetSettingModal.tsx
agatha197: Applied, thanks.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
agatha197: Applied to both the create and update mutations — both now select `id runtimeVariantId name description rank targetSpec { presetTarget valueType defaultValue key } required createdAt updatedAt` so the Relay store updates the node in place without a refetch.

react/src/components/RuntimeVariantPresetSettingModal.tsx
agatha197: Applied — switched to `BAISelect`.

react/src/components/RuntimeVariantPresetSettingModal.tsx
agatha197: Applied — switched to `BAISelect`.

react/src/components/AdminRuntimeVariantPreset.tsx
agatha197: Applied — refactored to render-as-you-fetch: the page now splits into an outer component (`useQueryLoader` + a reload effect keyed on filter/order/pagination/fetchKey) and an inner content component that reads via `usePreloadedQuery`, following the `ScopedAuditLog`/`DeploymentSchedulingHistoryModal` pattern. Previous rows stay visible during refetch via `useDeferredValue(queryRef)`.

react/src/pages/AdminRuntimeVariantPresetListPage.tsx
agatha197: Agreed — simplified. The editing state now holds the host preset node type directly (the query's own node type, which already carries the fragment spreads) instead of importing the fragment $key type separately.

react/src/pages/AdminRuntimeVariantPresetListPage.tsx
agatha197: Applied — `filter` is now typed as `RuntimeVariantPresetFilter` and the redundant cast in `queryVariables` was dropped.

react/src/pages/AdminRuntimeVariantPresetListPage.tsx
agatha197: Good catch — that's exactly why the bui move needed care: a `backend.ai-ui` fragment cannot spread a host (`react/src`) fragment (relay.config.js has `react` extend `backend.ai-ui` as its base project, not the other way around), so `BAIRuntimeVariantPresetNodesFragment` cannot include `...RuntimeVariantPresetSettingModalFragment`. Instead, the host list query now spreads both fragments directly on the same node (`id ...BAIRuntimeVariantPresetNodesFragment ...RuntimeVariantPresetSettingModalFragment`), and the host keeps its own map of full nodes keyed by `id` to look up the modal fragment ref when a row's edit action is clicked.

react/src/components/AdminRuntimeVariantPresetNodes.tsx
```diff
@@ -0,0 +1,249 @@
...
+        sorter: isEnableSorter('name'),
+        fixed: 'left',
+      },
+      {
+        key: 'description',
+        title: t('comp:BAIRuntimeVariantPresetNodes.Description'),
+        dataIndex: 'description',
+        defaultHidden: true,
+        render: (desc: string | null) => desc || '-',
+      },
```
nowgnuesLee, line 130: Please apply `sorter: isEnableSorter('key')` to the other columns as well.

nowgnuesLee (changes requested):

react/src/components/AdminRuntimeVariantPresetNodes.tsx
agatha197: Applied — added `sorter: isEnableSorter('<key>')` to every remaining column (description, runtimeVariantId, runtimeVariant, presetTarget, valueType, defaultValue, key, required, updatedAt) so all columns follow the same explicit pattern as name/rank/createdAt, consistently deferring to `availablePresetSorterKeys` for which ones actually enable a server-side sort.

react/src/components/AdminRuntimeVariantPresetTable.tsx
```diff
@@ -0,0 +1,263 @@
...
+      },
+      {
+        key: 'runtimeVariant',
+        title: t('adminRuntimeVariantPreset.RuntimeVariant'),
+        sorter: isEnableSorter('runtimeVariant'),
+        render: (__, record) => {
+          const name = runtimeVariantNameMap?.[record.runtimeVariantId];
+          return name ? name : <BAIId uuid={record.runtimeVariantId} />;
+        },
+      },
```
nowgnuesLee: I think this should be handled on the caller side via customizeColumns.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetTable.tsx
nowgnuesLee, line 1: Please move this component to bui.

nowgnuesLee (changes requested):

packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx
```diff
@@ -811,7 +811,7 @@ const BAIGraphQLPropertyFilter = <
           <Tooltip
             title={isValid || !isFocused ? '' : selectedProperty.rule?.message}
             open={!isValid && isFocused}
-            color={token.colorError}
+            color="red"
```
agatha197, line 846: This is to always show the text color as white with red background.

ironAiken2 (changes requested):
please resolve the conflicts :)

react/src/components/AdminRuntimeVariantPreset.tsx
```diff
@@ -0,0 +1,109 @@
...
+  // settings, so `AdminRuntimeVariantPresetContent` only ever has to call
+  // `onReload(...)`.
+  const reload = (
+    variables: AdminRuntimeVariantPresetContentQueryType['variables'],
+    options?: UseQueryLoaderLoadQueryOptions,
+  ) => {
+    const nextLimit = variables.limit ?? 10;
+    const nextOffset = variables.offset ?? 0;
+    setQueryParams({
+      filter: (variables.filter as RuntimeVariantPresetFilter) ?? null,
```
nowgnuesLee:
```suggestion
      filter: variables.filter ?? null,
```

nowgnuesLee (changes requested):

react/src/components/AdminRuntimeVariantPreset.tsx
agatha197: Fixed — dropped the redundant `RuntimeVariantPresetFilter` cast. `variables.filter` is already `RuntimeVariantPresetFilter | null | undefined`, so `variables.filter ?? null` type-checks without it.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,430 @@
...
+        }
+      >
+        <Suspense
+          fallback={
+            <Form.Item
+              label={t(
+                'comp:BAIRuntimeVariantPresetSettingModal.RuntimeVariant',
+              )}
+            >
+              <Skeleton.Input active block />
```
nowgnuesLee:
```suggestion
              <Select loading />
```

react/src/components/AdminDeploymentList.tsx
```diff
@@ -0,0 +1,511 @@
...
+  const webUINavigate = useWebUINavigate();
+
+  const [editingDeploymentId, setEditingDeploymentId] = useState<string | null>(
+    null,
+  );
+  const [deletingDeploymentId, setDeletingDeploymentId] = useState<
+    string | null
+  >(null);
+  const [drawerRevisionFrgmt, setDrawerRevisionFrgmt] =
+    useState<DeploymentRevisionDetail_revision$key | null>(null);
```
nowgnuesLee: You don't need to use fragment$key.

react/src/components/AdminDeploymentList.tsx
```diff
@@ -0,0 +1,511 @@
...
+                          });
+                        }}
+                        fallback={
+                          <Typography.Text type="secondary">-</Typography.Text>
+                        }
+                      />
+                    ),
+                  };
+                }
+                if (col.key === 'name') return next;
```
nowgnuesLee: Why not just return in the if statement above?

nowgnuesLee (changes requested):

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
agatha197: Applied the suggestion — swapped the Suspense fallback to `<Select loading />` so it matches the loaded `BAIRuntimeVariantSelect` shape instead of a generic skeleton bar.

react/src/components/AdminDeploymentList.tsx
agatha197: Fixed — now storing `drawerRevisionId: string | null` and deriving the revision fragment fresh from `deploymentNodes` on each render, matching the `editingDeployment`/`deletingDeployment` pattern already in this file. Dropped the `DeploymentRevisionDetail_revision$key` import.

react/src/components/AdminDeploymentList.tsx
agatha197: Simplified — each branch now returns directly (with `defaultHidden` computed once up front and included in the non-name branches), instead of building `next` and re-checking `col.key === 'name'` afterward.

react/src/components/AdminRuntimeVariantPreset.tsx
```diff
@@ -0,0 +1,313 @@
...
+      <BAIUnmountAfterClose>
+        <BAIRuntimeVariantPresetSettingModal
+          open={isCreating || !!editingPreset}
+          presetFrgmt={editingPreset}
+          onRequestClose={(success) => {
+            setIsCreating(false);
+            setEditingPreset(null);
+            if (success) {
+              updateFetchKey();
+            }
```
nowgnuesLee, line 311: We need to refetch differently depending on whether it's an update or a create.

nowgnuesLee (changes requested):

react/src/components/AdminRuntimeVariantPreset.tsx
agatha197: Thanks for the catch. We decided to keep a plain refetch on modal success rather than branch on create vs. update.

Both clean ways to skip the refetch on update have downsides here: inferring `wasCreating` in the consumer is state it shouldn't need to track, and a Relay connection directive (`@prependNode`) would require drilling the connection `__id` into the modal and would insert the new row out of sort order (this is an offset/limit + sorted/filtered list) with a stale `count`.

Since the list is small and a `network-only` refetch re-sorts and re-counts correctly, always refetching on success is the simplest correct behavior — updates still get in-place field updates via Relay's by-id store merge, so the extra refetch is cheap. Happy to revisit if you feel strongly.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetTable.tsx
```diff
@@ -0,0 +1,250 @@
...
+        name @required(action: NONE)
+        description
+        rank
+        targetSpec {
+          presetTarget
+          valueType
+          defaultValue
+          key
+        }
+        required @since(version: "26.4.4")
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit — changed the `required` field's directive to `@since(version: "26.4.4rc9")` to match the `runtime-variant-preset-required` capability flag (which turns on at rc9), consistent with `useRuntimeParameterSchema.ts`. Relay artifacts regenerated.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,429 @@
...
+        name
+        description
+        rank
+        targetSpec {
+          presetTarget
+          valueType
+          defaultValue
+          key
+        }
+        required @since(version: "26.4.4")
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit — the `required` field's `@since` is now `26.4.4rc9` in all three fragments/mutations here, matching the capability cutoff so the field isn't stripped on rc managers. Relay artifacts regenerated.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,429 @@
...
+            <Switch />
+          </Form.Item>
+        ) : null}
+        {preset ? (
+          <Form.Item
+            label={t('comp:BAIRuntimeVariantPresetSettingModal.Rank')}
+            name="rank"
+            tooltip={t('comp:BAIRuntimeVariantPresetSettingModal.RankTooltip')}
+          >
+            <InputNumber min={0} style={{ width: '100%' }} />
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit — the rank input now uses `precision={0}` so only integers can be entered, matching the GraphQL `Int` type.

react/src/components/AdminRuntimeVariantPreset.tsx
```diff
@@ -0,0 +1,324 @@
...
+  const uuidRule = {
+    message: t('general.InvalidUUID'),
+    validate: (value: string) => isValidUUID(value.toLowerCase()),
+  };
+
+  return (
+    <BAIFlex direction="column" align="stretch" gap={'sm'}>
+      <BAIFlex justify="between" wrap="wrap" gap={'sm'}>
+        <BAIFlex gap={'sm'} align="start" wrap="wrap" style={{ flexShrink: 1 }}>
+          <BAIGraphQLPropertyFilter
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit — bound the filter to `BAIGraphQLPropertyFilter<RuntimeVariantPresetFilter>` and dropped the `as GraphQLFilter` cast, so the concrete schema type flows through `value`/`onChange` (matching `AdminDeploymentPreset`).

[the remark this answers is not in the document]
agatha197: Good catch — this is a valid version-compatibility concern. `RuntimeVariantPresetFilter.AND/OR/NOT` only exist from manager 26.7.0, but the tab targets the 26.4 baseline, so a multi-condition filter would fail GraphQL validation on 26.4–26.6.

Leaving this open for a product decision: either gate multi-condition entry behind a capability check (e.g. `baiClient.supports('sub-filter')`) / emit a flat filter on legacy managers, or raise the tab's baseline. It needs a design call rather than a mechanical fix, so I'm not auto-applying it — flagging for follow-up.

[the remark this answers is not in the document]
agatha197: Implemented in the latest commit — multi-condition filtering is now gated behind `baiClient.supports('sub-filter')`. On managers without the capability the property filter uses `singleCondition`, so it emits a flat filter (no `AND`/`OR`/`NOT`) that validates on the 26.4 baseline; on newer managers multi-condition is allowed. Matches the existing `sub-filter` gating pattern (e.g. `ResourcePolicyPage`).

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetTable.tsx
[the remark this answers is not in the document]
agatha197: Follow-up: per maintainer decision, we're aligning `@since` to the version the schema itself documents — the `required` field is annotated `"""Added in 26.4.4."""` — so this now uses `@since(version: "26.4.4")` rather than the `rc9` capability-cutoff. Trade-off noted: on `26.4.4rcX` managers the field is version-gated out (PEP440 `rc < final`), matching the schema's stated "since" rather than the earliest rc it shipped in.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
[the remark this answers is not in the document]
agatha197: Follow-up: per maintainer decision, we're aligning `@since` to the version the schema itself documents — the `required` field is annotated `"""Added in 26.4.4."""` — so this now uses `@since(version: "26.4.4")` rather than the `rc9` capability-cutoff. Trade-off noted: on `26.4.4rcX` managers the field is version-gated out (PEP440 `rc < final`), matching the schema's stated "since" rather than the earliest rc it shipped in.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx
```diff
@@ -0,0 +1,429 @@
...
+      >
+        <Suspense
+          fallback={
+            <Form.Item
+              label={t(
+                'comp:BAIRuntimeVariantPresetSettingModal.RuntimeVariant',
+              )}
+            >
+              <BAISelect loading />
+            </Form.Item>
```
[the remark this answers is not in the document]
agatha197: Fixed. The Suspense fallback now registers the `runtimeVariantId` field with the same `name` and required rule (and a disabled loading select), so submitting during the transient loading window is blocked by validation instead of sending a create mutation with `runtimeVariantId` undefined.

packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetTable.tsx
```diff
@@ -0,0 +1,250 @@
...
+  };
+
+  const baseColumns = _.map(
+    filterOutEmpty<BAIColumnType<RuntimeVariantPresetNodeInList>>([
+      {
+        key: 'name',
+        title: t('comp:BAIRuntimeVariantPresetTable.Name'),
+        dataIndex: 'name',
+        sorter: isEnableSorter('name'),
+        fixed: 'left',
```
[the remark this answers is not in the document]
agatha197: Fixed. The Name column is now marked `required: true`, which `BAITable` treats as always-visible, so it can no longer be hidden through table settings and the per-row edit/delete actions it hosts always remain reachable.

react/src/components/AdminRuntimeVariantPreset.tsx
```diff
@@ -0,0 +1,331 @@
...
+                  }
+                  message.success(t('adminRuntimeVariantPreset.PresetDeleted'));
+                  setDeletingPreset(null);
+                  onReload(queryRef.variables, { fetchPolicy: 'network-only' });
+                  resolve();
+                },
+                onError: (error) => {
+                  logger.error(error);
+                  message.error(error?.message || t('general.ErrorOccurred'));
+                  reject();
```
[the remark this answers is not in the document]
agatha197: Fixed. The delete handler is now a plain callback that calls `commitDeletePreset` directly with its `onCompleted` / `onError` callbacks, instead of wrapping it in a `Promise` that rejected. Since `BAIDeleteConfirmModal` invokes `onOk` without awaiting it, the previous rejection was unhandled; errors are still surfaced via `message.error`.

PR #8156, opened 2026-07-06
fix(FR-3214): kill the whole dev process group on shutdown so leaked watch orphans don't exhaust file descriptors

### LINKED ISSUE #8047: Dev server fails with too-many-watched-files when running multiple repos concurrently
## Problem

When running several repositories' dev servers at once, the dev server fails with "too many watched files" (EMFILE). It also dies on macOS, where FSEvents was supposed to make watching cheap.

## Root cause (measured via `lsof` on a live macOS dev stack)

It is **open file descriptors, not "too many directories watched."** A single live `vite` dev process holds ~**5,900 open regular files**, of which ~**3,800 are immutable pnpm-store dependency `.d.ts`**. These are held open by **vite-plugin-checker**'s TypeScript watch program. On this machine its `tsc` watch worker uses **per-file kqueue (1 open FD per watched file)** because macOS FSEvents is failing in this environment (plain recursive `fs.watch` throws EMFILE; watchman also crashes).

Why it then EMFILEs:

- Dev launched from a terminal inherits the login-shell `ulimit -n` (1,048,576) and survives a single repo.
- Dev launched from a GUI / IDE task inherits launchd soft `maxfiles` = **256**, so 5,900 >> 256 fails instantly.
- Running **N repos at once** multiplies ~5,900 x N toward `kern.maxfiles` (368,640).

This is **not a Vite regression** — the retired craco stack's fork-ts-checker watched the same dependency `.d.ts` via the same TS-watch-program architecture. Watching dependency `.d.ts` is inherent to running any in-process TS type-checker.

## Fix (implemented, proven on the real stack)

Stop the checker's TS watch program from **watching** the pnpm store. It still **reads** every dependency `.d.ts` to type-check (the exclude only drops the file watcher), so type-checking stays complete and source files keep their normal fast per-file watch.

`vite-plugin-checker` has no inline watchOptions API, and the exclude spec must be an **absolute** store path (TS anchors `**/`-globs under the project basePath, so they can never match the store realpath, which lives outside the repo). That path is machine-specific, so it cannot live in the committed tsconfig.

Implementation in `react/vite.config.ts` (dev/serve only): resolve the pnpm store root at startup and write a gitignored `react/tsconfig.checker.json` that extends the base tsconfig and adds `watchOptions.excludeDirectories` / `excludeFiles` for `<storeRoot>/**`, then point `vite-plugin-checker` at it. Falls back to the plain base tsconfig for npm/yarn flat installs or if store resolution fails (no-op, original behavior preserved).

## Measured result (real `pnpm dev` boot, lsof on the vite pid)

| metric | before | after |
|---|---|---|

|—|—|—|
|---|---|---|

| vite open regular files | 5,884 | 1,404 (-76%) |
|---|---|---|

| files held open under pnpm store | 3,825 | 2 |
|---|---|---|

| type errors reported | 0 | 0 (identical) |
|---|---|---|

| live source-edit detection | 2s | 2s (preserved) |
|---|---|---|

This removes the per-repo FD multiplier that makes concurrent multi-repo dev fall over. HMR (Vite's separate chokidar watcher) and type-check correctness are unaffected.

## Notes

- 1,404 FDs is still above the launchd soft limit of 256, so a single GUI-launched repo may still need a raised FD limit — but the N-times multiplier (the actual multi-repo symptom) is gone. Raising the launcher's soft limit is optional belt-and-suspenders, not the primary fix.

—

JIRA Issue: FR-3214

nowgnuesLee:
Resolves #8047 (FR-3214)

## Problem

Running several repos' dev servers at once fails with `EMFILE: too many open files`. A major driver turned out to be **leaked orphan watch processes accumulating across dev restarts**: on a live machine, `ps` showed relay watch chains from dev sessions started **3 and 8 days earlier** still alive with `PPID 1` — `nodemon → pnpm run relay --watch → relay-compiler cli.js → native relay --watch binary` — each holding its full set of file-watcher FDs forever.

### Root cause

`dev.mjs` forwarded SIGINT/SIGTERM to **one PID only** (`child.kill(sig)` on the `concurrently` process), and `concurrently --kill-others` likewise signals only its **direct children**. The relay watch tree is unusually deep (`concurrently → pnpm → nodemon → pnpm → node cli.js → native relay binary`), and per-PID signal relay breaks at the nodemon hop — killing the direct child orphans everything below it.

The leak fires on every shutdown path that doesn't broadcast to the terminal's foreground process group:

- IDE/agent task stop (signals the top PID only) — the common case when dev runs under VS Code / code-server / an agent session
- `concurrently --kill-others` after one child dies
- graceful-shutdown chains that hang (a deep child swallowing SIGINT while its parents wait), which leaks even under a real terminal Ctrl+C
- double Ctrl+C interrupting a cleanup in progress

## Fix

`scripts/dev.mjs` now owns the whole tree's lifecycle:

1. **Spawn `concurrently` as a detached process-group leader** (`detached: true`, non-Windows). Every descendant — however deep — inherits the group.
2. **Signal the group, not the PID.** SIGINT/SIGTERM/SIGHUP handlers forward via `process.kill(-child.pid, sig)`, so every member receives the signal directly, with no reliance on intermediate layers relaying it. (SIGHUP is now forwarded explicitly because a detached group no longer receives the terminal's hangup.)
3. **Unconditional kill guarantee.** Two backstops ensure nothing outlives dev.mjs:
   - On any forwarded signal, a 5s timer escalates the whole group to **SIGKILL** if graceful shutdown hangs.
   - On `concurrently` exit (covers the `--kill-others` path, where dev.mjs itself never received a signal), dev.mjs probes the group (`kill(-pgid, 0)`), SIGTERMs survivors, waits up to 5s, then **SIGKILLs** the remainder before exiting.

Windows falls back to direct `child.kill` (negative-PID kill throws there); macOS/Linux — where the leak was observed — get the full group semantics.

This PR previously took a different approach (excluding the pnpm store from vite-plugin-checker's TS watch program via a generated `tsconfig.checker.json`). That reduced per-process FDs but left the actual leak — orphaned watch trees — unfixed; it has been reverted in favor of this root-cause fix.

## Measurements (live, this machine)

**Scenario 1 — single-PID SIGTERM to dev.mjs (the IDE-stop leak path):**

| | before fix | after fix |
|---|---|---|
| group members while running | 11 (tsc, vite, esbuild, portless, nodemon, pnpm×3, relay cli, native relay) | 11 |
| survivors 8s after `kill -TERM <dev.mjs pid>` | relay chain leaked (observed as day-old `PPID 1` zombies) | **0** |

**Scenario 2 — `kill -9` one child (tsc) to trigger `concurrently --kill-others`:**

- All 11 group members gone within 10s, dev.mjs exited with the child's status, log shows the normal `--kill-others` SIGTERM cascade. **0 survivors.**

## Verification

- `bash scripts/verify.sh` — `=== ALL PASS ===` (Relay / Lint / Format / TypeScript).
- `node --check scripts/dev.mjs` passes.
- Both live shutdown scenarios above measured on a real dev boot of this branch (`ps ax -o pid,ppid,pgid` group-membership check before/after).

### DIFF
--- a/scripts/dev.mjs
+++ b/scripts/dev.mjs
@@ -68,8 +68,62 @@ const args = [
   `${portlessSpec}-- pnpm --prefix ./react run start`,
 ];
 
-const child = spawn('concurrently', args, { stdio: 'inherit', env, shell: false });
-const forward = (sig) => child.kill(sig);
-process.on('SIGINT', forward);
-process.on('SIGTERM', forward);
-child.on('exit', (code, signal) => process.exit(signal ? 1 : (code ?? 0)));
+// Spawn concurrently as the leader of its own process group (`detached`).
+// The watch tree below it is deep (concurrently → pnpm → nodemon → pnpm →
+// relay binary), and per-PID signal forwarding breaks at the nodemon hop:
+// killing only `child.pid` orphans the relay watch chain, which then keeps
+// thousands of file-watcher FDs open forever (FR-3214). Signaling the
+// negative PID (= the whole group) reaches every descendant directly, no
+// matter how deep the chain is or whether intermediate layers forward.
+const detached = process.platform !== 'win32';
+const child = spawn('concurrently', args, { stdio: 'inherit', env, shell: false, detached });
+
+// Signal every process in the group (negative PID = the whole group), falling
+// back to a direct kill when the group is already reaped. Win32 has no process
+// groups, so there the whole fix degrades to the plain per-PID kill.
+const signalTree = detached
+  ? (sig) => {
+      try {
+        process.kill(-child.pid, sig);
+      } catch {
+        child.kill(sig); // group already reaped — best-effort direct kill (no-throw)
+      }
+    }
+  : (sig) => child.kill(sig);
+const treeAlive = () => {
+  if (!detached) return false;
+  try {
+    process.kill(-child.pid, 0); // probe: throws ESRCH when no member survives
+    return true;
+  } catch {
+    return false;
+  }
+};
+
+// Guarantee the group dies even when graceful shutdown hangs (e.g. the relay
+// binary swallowing SIGINT while its parents wait on it): from the first
+// shutdown signal, poll for survivors and escalate to SIGKILL once the grace
+// period runs out. Nothing in this group may outlive dev.mjs.
+const SIGKILL_GRACE_MS = 5000;
+let escalation = null;
+const escalate = async () => {
+  const deadline = Date.now() + SIGKILL_GRACE_MS;
+  while (treeAlive() && Date.now() < deadline) {
+    await new Promise((resolve) => setTimeout(resolve, 100));
+  }
+  if (treeAlive()) signalTree('SIGKILL');
+};
+const shutdown = (sig) => {
+  signalTree(sig); // repeated Ctrl+C re-signals; the escalation only starts once
+  return (escalation ??= escalate());
+};
+process.on('SIGINT', () => shutdown('SIGINT'));
+process.on('SIGTERM', () => shutdown('SIGTERM'));
+process.on('SIGHUP', () => shutdown('SIGHUP')); // terminal close no longer HUPs the detached group
+
+child.on('exit', async (code, signal) => {
+  // concurrently is gone; anything still alive in its group is an orphan
+  // (leaked grandchild mid-shutdown). Sweep it before exiting.
+  if (treeAlive()) await shutdown('SIGTERM');
+  process.exit(signal ? 1 : (code ?? 0));
+});


nowgnuesLee:
* **#8407** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8407?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8395** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8395?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8394** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8394?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8357** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8357?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8156** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8156?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8156?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8156).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

agatha197 (approved):
### LGTM 

Solid fix that targets the real root cause — dev.mjs not owning the whole watch tree. Making `concurrently` a detached process-group leader and signaling `-pgid` with a SIGKILL escalation cleanly covers the `--kill-others`, SIGHUP, and Windows-degrade paths. `node --check` passes, and the remaining points are all edge cases outside real-world usage, so this is good to merge.

Before merging, please just **document these two residual risks** in the PR description / a comment:

**1\. If dev.mjs is SIGKILLed or crashes, the group is orphaned (not a regression).**
The whole fix assumes dev.mjs receives a catchable signal (TERM/INT/HUP) and lives long enough to run the sweep. Because the group is now detached, it's decoupled from dev.mjs's death and the terminal hangup — so a supervisor that sends SIGKILL (uncatchable) to dev.mjs skips the sweep and reproduces exactly this leak. Most IDEs/agents send SIGTERM, so real usage is covered, but this is the one hole left open.

**2\. The** **`setsid`** **escape assumption is implicit.**
The fix relies entirely on every descendant staying in the `concurrently` group. If anything in the chain calls `setsid()` / spawns its own `detached` group, `-pgid` can't reach it and it still leaks. The current toolchain (concurrently/pnpm/nodemon/relay/vite/tsc) doesn't do this, so it holds today, but it's a hidden assumption that could silently reintroduce the leak on a future tooling change — worth a one-line comment.

**FYI (no change needed):**

- The exit handler signals `-child.pid` after the leader has already been reaped, so there's a theoretical PID/PGID reuse race, but the window is microtask-sized and negligible.
- The missing `child.on('error')` handler (when the `concurrently` binary is absent) is a pre-existing gap, not a regression.
- Verified and fine: `escalation ??=` single-flight, no 5s wait on normal shutdown, and SIGKILL releasing FDs immediately.

yomybaby (approved):
LGTM

PR #8191, opened 2026-07-07
feat(FR-3271): add detailed permissions tab with per-scope-type tables and grant-state tags

### LINKED ISSUE #8179: Add "Detailed Permissions" tab: per-scope-type tables with grant-state tags (FR-3264)
Parent task: FR-3264 (GitHub clone #8165). Spec: `.specs/FR-3264-role-detail-drawer-detailed-permissions/spec.md`.

Graphite stack base (PR 1 of 3). This PR delivers the at-a-glance **read view**. The scope-level edit modal (single) lands in the next stacked PR, and multi-scope bulk edit in the one after.

## Goal (FR-1, FR-2, FR-3, FR-4, FR-7)

Merge the standalone **Scopes** and **Permissions** tabs of `RoleDetailDrawer` into a single **"Detailed Permissions"** tab (Korean 세부 권한) built from **one BAICard section per assigned scope type**, each row showing a scope instance plus color-coded grant-state tags per selectable entity type.

## Files

- NEW `react/src/components/RolePermissionAssignmentTab.tsx` — merged tab: derives the distinct assigned scope types from the role scopes (reuse the `scopes(first: 100)` fragment precedent from `CreatePermissionModal`), renders one `BAICard` per type stacked in a `BAIFlex direction="column"` (layout ref: `ProjectFolderPermissionPanel.tsx`), empty state (`rbac.NoScopesToDisplay`) when the role has no scopes.
- NEW per-scope-type section (inline or `RoleScopePermissionSection.tsx`) — issues its OWN queries so filter/refresh on one card does not refetch others: (a) `adminRole(id).scopes(filter: {entityType}, limit, offset)` for rows, (b) `adminPermissions(filter: {roleId, scopeType}, limit, offset)` to compute tag state, (c) `rbacPermissionMatrix` for the selectable entity list + full operation set. Card-scoped `BAIFetchKeyButton` refresh in `extra`; content-scoped `BAIGraphQLPropertyFilter` in the body (per `.claude/rules/use-bai-card.md`). Pagination is **offset-based** (`limit`+`offset` only — per `.claude/rules/graphql-pagination.md`) and kept in **local React state (NOT nuqs/URL)**.
- MODIFY `react/src/components/RoleDetailDrawerContent.tsx` — tab list becomes [Detailed Permissions (default), Assignments]; remove the Scopes tab; remove the `s*`/`p*` nuqs keys and the tab-change reset for them (keep `a*` for Assignments). Keep the `Descriptions` metadata block EXACTLY as-is (FR-7). Wrap the new tab in `<Suspense fallback={<Skeleton active/>}>`.
- MODIFY `react/src/components/RoleDetailDrawer.tsx` — drop the `RoleScopeTabFragment` spread + its scope\* variables from `RoleDetailDrawerQuery` (the new tab self-fetches). Keep the role, assignment, and (for now) permission fragments.
- DELETE `react/src/components/RoleScopeTab.tsx` — fully absorbed (scopes query + `resolveScopeName` name resolution move into the new section). Keep `RolePermissionTab.tsx` / `CreatePermissionModal.tsx` in place for now so permission CRUD is not lost before its replacement lands in PR 2 (they are removed there).
- i18n: add `rbac.DetailedPermissions`, `rbac.FullyAllowed`, `rbac.PartiallyAllowed`, `rbac.NotAllowed`, `rbac.NoScopesToDisplay` to `resources/i18n/en.json` + `ko.json` (Korean 세부 권한 etc.); reuse `rbac.types.*`, `rbac.ScopeId`, `rbac.ScopeRawId`.

## Grant-state tag rule (FR-4)

For each (scopeId, entityType): compare the role`s granted operations against the FULL operation set the matrix lists for (scopeType, entityType) — including `GRANT_\*` delegate ops. All granted → green (fully); some → gold (partial); none → gray (not allowed, still shown, never hidden). Tags are **display-only** (no click-to-edit). Entity types absent from the matrix render no tag.

## Acceptance criteria

- Opening a role drawer shows exactly the new tab + Assignments; Detailed Permissions is default active; Assignments renders identically to today.
- A role with PROJECT + USER scopes shows a Project card and a User card and no other; a role with no scopes shows the empty state.
- Filtering/refreshing one card does not reload another; each section query declares only `(limit, offset)`.
- Tag colors match fully/partial/none per (scope, entity); the metadata `Descriptions` block is unchanged; `autoAssign` still gated by `baiClient.supports("role-auto-assign")`.
- Keep the `rbac-filter-wrapper` support branch for filter value wrapping.

## Verification

`bash scripts/verify.sh` ends with `=== ALL PASS ===` (Relay compile + lint + format + tsc). Run `pnpm run relay` after adding graphql tags.

JIRA Issue: FR-3271

ironAiken2:
Resolves #8179 (FR-3271)

## Summary

Introduces the new **"Detailed Permissions"** tab in `RoleDetailDrawer`, merging the former Scopes and Permissions views into per-scope-type cards. This is PR 1 of the FR-3264 stack — the at-a-glance **view** (grant-state tags). The scope-level permission edit modal follows in PR 2/3.

### What changed

- **New** `react/src/components/RolePermissionDetailTab.tsx` — a thin orchestrator. Reads the role via a single `roleNodeFrgmt` (`id` + `totalScopes: scopes(first: 1) { count }` for the tab-level empty state) and renders one card per scope type that **`rbacPermissionMatrix` reports** — scope types are not hardcoded client-side, so a type added on the server needs no code change here. Each card gets its type's configurable entity × operation set derived from the matrix.
- **New** `react/src/components/ScopedRolePermissionCard.tsx` — one scope-type card. Issues its own query from the first render (so per-card search/refresh never refetches the others, FR-3):
  - `adminRole.scopes(filter: { scopeType: { equals: <type> } }, limit, offset)` — **server-filtered by scope type and server-paginated** via the `EntityFilter.scopeType` filter added in **manager 26.8.0**, with `useBAIPaginationOptionState` driving `limit`/`offset` (offset mode only, per `.claude/rules/graphql-pagination.md`).
  - `adminRole.permissions(filter: { scopeType: ... }, limit: 100)` — the role's grants for tag-state computation (the role is implicit via the connection; no `roleId` filter).
  - Scope-id search via **`BAIGraphQLPropertyFilter<EntityFilter>`** (`scopeId`, default `iContains`, 26.8.0) paired with `BAIFetchKeyButton` in the card body (`UserFolderPermissionPanelV2` pattern). The **resolved scope name is not searchable** — `EntityFilter` exposes no name field (see Notes).
  - Renders nothing when the role has no scope of its type (`count === 0` and no active filter) — FR-2's "only assigned scope types" holds per card.
  - Columns: scope name via `BAINameActionCell` (the FR-6 edit action attaches here in PR 2), raw scope id via `BAIId`, and **display-only entity-type `Tag`s** color-coded by grant state — green (all matrix operations granted), gold (some), gray (none — still shown, never hidden).
- **New** `react/src/helper/rbacGrantState.ts` (+ unit test) — pure `computeRBACGrantState()` comparing the role's granted operations against the full matrix operation set for a `(scopeType, entityType)` pair → `full` / `partial` / `none`.
- **`data/schema.graphql` synced to manager 26.8.0** — `EntityFilter` gains `scopeType` / `scopeId` filter fields ("Added in 26.8.0"), which is what makes server-side scope-type filtering possible (the previously escalated backend gap is resolved). The schema bump also refreshes unrelated generated artifacts.
- **Removed** the standalone **Scopes** and **Permissions** tabs and their components: `RoleScopeTab.tsx`, `RolePermissionTab.tsx`, and `CreatePermissionModal.tsx` are deleted (superseded by the merged tab; permission editing returns as the edit modal in PR 2). `RoleFormModal` now sources `RBACElementType` from its own generated types.
- **Refactored** `react/src/components/RoleAssignmentTab.tsx` — fragment moved from `on Query` (`adminRoleAssignments(filter: { roleId })`) to **`on Role` using the `Role.users` connection**; node-based refetch; `roleId`/`projectScopeId` derived from its own fragment; `a*` nuqs keys → local React state (**nothing inside the drawer writes to the URL** — only the page-level `roleDetail` param remains); typed `RoleAssignmentFilter`.
- **Refactored** `react/src/components/RoleDetailDrawer.tsx` — the drawer issues no query of its own. It renders from the selected list row's fragment ref (`roleFrgmt`) via a `@refetchable` `RoleDetailDrawerFragment` spread on the page's list nodes, so opening is fetch-free; the header refresh button refetches just that role (`RoleDetailDrawerRefetchQuery`) inside a transition. `RoleAssignmentTab`'s `scopes(first: 1)` is aliased to `firstScope` to avoid an argument conflict with `RoleNodesFragment`'s `scopes(first: 3)` now that both compose into one query. The `roleDetail` URL param still restores the drawer on reload when the row is on the current page.
- **Naming** — every prop carrying a Role-node-derived fragment is uniformly `roleNodeFrgmt` (`RoleDetailDrawerContent`, `RoleAssignmentTab`, `RolePermissionDetailTab`, `ScopedRolePermissionCard`, `RoleFormModal`).
- **No legacy filter branching** — the `baiClient.supports('rbac-filter-wrapper')` guard is dropped in the new components; 26.4.x LTS always supports the wrapped operator form.
- **i18n** — adds `rbac.DetailedPermissions`, `rbac.FullyAllowed`, `rbac.PartiallyAllowed`, `rbac.NotAllowed`; rewords `rbac.NoScopesToDisplay` to convey "no scopes are configured for this role"; **removes** the now-unused `rbac.SearchScopes` — all 21 locales.
- Spec documents for the stack live in `.specs/FR-3264-role-detail-drawer-detailed-permissions/` (to be refreshed once the stack settles).

### Notes / caveats

- **Requires manager ≥ 26.8.0** for the `EntityFilter.scopeType` / `scopeId` filters. On older managers the scopes connection rejects/ignores these fields — the tab is not version-gated (the whole RBAC detail surface targets new managers).
- Scope search matches the **raw scope id** only (`iContains`); the resolved display name (project name / email / …) is not server-searchable. If name search is needed, it requires a backend `EntityFilter` name field (follow-up).
- Tag-state computation fetches up to 100 permission rows per card (exact paging against the visible rows' scope ids via `PermissionFilter.scopeId: { in: [...] }`, 26.8.0, is deferred as a follow-up); a known bound for very large roles.
- One query per matrix scope type is issued in parallel when the tab opens; cards for unassigned types self-hide after their (empty) response.

## Verification

`bash scripts/verify.sh` → Relay, Lint, Format, TypeScript all pass. Unit tests `react/src/helper/rbacGrantState.test.ts` pass. Server-filter behavior (`scopeType` / `scopeId iContains`, server pagination) needs a live check against a 26.8.0 manager.

## Stacked on

This is the **base PR** of the FR-3264 stack (`main → FR-3271 → FR-3272 → FR-3273`). PR 2 (FR-3272) adds the single-scope permission edit modal wired into `ScopedRolePermissionCard`; PR 3 (FR-3273) adds multi-scope bulk edit (row selection + 'Keep as is' semantics).


ironAiken2:
* **#8268** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8268?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8264** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8264?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8194** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8192** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8192?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8191** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8191?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8191?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8191).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

resources/i18n/de.json
```diff
@@ -2195,6 +2195,7 @@
     "ScopeRawId": "Geltungsbereich-ID",
     "ScopeType": "Bereichstyp",
     "ScopeTypeAndId": "Bereichstyp / Ziel",
+    "SearchScopes": "Geltungsbereiche suchen",
```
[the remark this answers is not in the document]
ironAiken2: Moved this PR's five sibling keys (rbac.DetailedPermissions, FullyAllowed, PartiallyAllowed, NotAllowed, NoScopesToDisplay) into this PR — all 21 locales now carry them in this diff, alongside rbac.SearchScopes.

react/src/components/RolePermissionDetailTab.tsx
```diff
@@ -0,0 +1,116 @@
...
+        }
+      `,
+      {},
+      { fetchPolicy: 'store-and-network' },
+    );
+
+  // One card candidate per matrix scope type, carrying its configurable
+  // entity × operation set. The scope types come from the server as-is —
+  // nothing is hardcoded client-side.
+  const formattedPermissionMartixs = _.uniqBy(
```
agatha197, line 72: Could we render the per-scope-type cards in a fixed **domain → project → user** order (system first if it ever appears)?

Right now `_.uniqBy(rbacPermissionMatrix, …)` preserves whatever order `rbacPermissionMatrix` returns, so the card order simply follows the server response. I'd like the tab to always show the cards in domain → project → user order regardless of how the matrix comes back. Sorting the deduped list against a canonical scope-type order before mapping would do it (unknown/new types can fall to the end).

react/src/components/RolePermissionDetailTab.tsx
```diff
@@ -0,0 +1,116 @@
...
+          operations: _.uniq(
+            entity.actions.map((action) => action.requiredPermission),
+          ),
+        }),
+      ),
+  }));
+
+  if (role.totalScopes?.count === 0) {
+    return (
+      <BAICard>
```
agatha197: This `BAICard` has no `tabList`, so per `.claude/rules/use-bai-card.md` it should pass `styles={{ body: { paddingTop: 0 } }}` for header/body flush consistency — the sibling `ScopedRolePermissionCard` already does. Minor (no title here, so little visual impact) but it's a rule deviation in new code.

react/src/components/ScopedRolePermissionCard.tsx
```diff
@@ -0,0 +1,392 @@
...
+      ...scopeIdFilter,
+      scopeType: { equals: scopeType },
+    } as ScopedRolePermissionCardQuery['variables']['scopeFilter'],
+    scopeLimit: baiPaginationOption.limit,
+    scopeOffset: baiPaginationOption.offset,
+    // The role itself is implicit via the `adminRole.permissions` connection.
+    permissionFilter: {
+      scopeType: { equals: scopeType },
+    } as ScopedRolePermissionCardQuery['variables']['permissionFilter'],
+    permissionLimit: PERMISSION_FETCH_LIMIT,
```
agatha197, line 163: `permissionFilter` only constrains `scopeType`, and this fetches up to `PERMISSION_FETCH_LIMIT` (100) permission rows, from which every visible scope row's grant-state tag is computed. If a role holds more than 100 `(scopeId × entityType × operation)` rows for one scope type, scope rows on later table pages get an empty grant set and are mis-tagged as `none`/`partial` (gray/orange) even when fully granted. It fails safe (never over-grants), but the tags are wrong under scale.

Also, the comment at lines 41–42 says *"no per-scopeId permission filter exists to page it exactly against the visible rows"* — that's no longer true on the 26.8.0 schema this PR ships: `PermissionFilter.scopeId: StringFilter` exists (with `in: [String!]`). Suggest either paging permissions to the current page's scope ids (`scopeId: { in: [...] }`, a second data-dependent fetch once the scopes resolve) to drop the cap, or at minimum correcting the comment. Tracking the real paging as a follow-up is fine.

react/src/components/RoleDetailDrawer.tsx
```diff
@@ -180,15 +127,12 @@ const RoleDetailDrawerInner: React.FC<RoleDetailDrawerInnerProps> = ({
         )}
       </BAIFlex>
       <RoleDetailDrawerContent
-        roleDetailFrgmt={data.adminRole}
-        scopeQueryRef={data}
-        assignmentQueryRef={data}
-        permissionQueryRef={data}
+        roleNodeFrgmt={adminRole}
         onDataChange={onDataChange}
```
agatha197: `onDataChange` is threaded from `RoleDetailDrawerInner` down through `RoleDetailDrawerContent` into `RoleAssignmentTab.onAssignmentChange`, but the outer `RoleDetailDrawer` never passes `onDataChange` into `RoleDetailDrawerInner` (and `RoleDetailDrawerProps` doesn't declare it), so it's always `undefined`. The UI still refreshes via `RoleAssignmentTab`'s own refetch + Relay store normalization, so this is dead plumbing rather than a bug. Suggest dropping the `onDataChange` / `onAssignmentChange` props (or wiring them through if a parent refresh is actually intended).

react/src/components/RoleDetailDrawerContent.tsx
```diff
@@ -2,36 +2,24 @@
...
   onDataChange?: () => void;
 }
 
 const RoleDetailDrawerContent: React.FC<RoleDetailDrawerContentProps> = ({
-  roleDetailFrgmt,
-  scopeQueryRef,
-  assignmentQueryRef,
-  permissionQueryRef,
+  roleNodeFrgmt,
   onTabReset: _onTabReset,
```
agatha197: `onTabReset` (destructured as `_onTabReset`) is unused — no caller passes it and it was already dead before this PR. Since this refactor already removes the tab-reset machinery, the `onTabReset?: () => void;` field in `RoleDetailDrawerContentProps` and this destructure can go too.

agatha197 (changes requested):
Reviewed the Detailed Permissions tab. Solid refactor overall — fragment architecture, single `$id` query, pagination single-mode compliance, i18n key parity, and no dangling references to the deleted components all look good. No blockers. One primary request (fixed scope-type card ordering) plus a medium correctness caveat and a few cleanups, left inline.

react/src/components/RolePermissionDetailTab.tsx
ironAiken2: We'd like to keep this server-driven: the tab intentionally renders cards in the order `rbacPermissionMatrix` returns them, so the server owns the canonical scope-type ordering and a newly added type lands in the right position without a client change. If the matrix ordering needs adjusting, we'd prefer fixing it server-side over hardcoding a client-side sort.

react/src/components/RolePermissionDetailTab.tsx
ironAiken2: Fixed in 05fd3188d — added `styles={{ body: { paddingTop: 0 } }}` to the empty-state card.

react/src/components/ScopedRolePermissionCard.tsx
ironAiken2: Good catch on the stale comment — corrected in 05fd3188d. `PermissionFilter.scopeId: { in: [...] }` does exist on 26.8.0; paging exactly against the visible rows' scope ids needs a second data-dependent fetch after the scopes resolve, so this PR keeps the capped fetch and we'll track exact paging as a follow-up.

react/src/components/RoleDetailDrawer.tsx
ironAiken2: Removed in 05fd3188d — dropped the dead `onDataChange` plumbing (`RoleDetailDrawerInner` / `RoleDetailDrawerContent`) along with `RoleAssignmentTab.onAssignmentChange`; the tab's own refetch + Relay store normalization already cover the refresh.

react/src/components/RoleDetailDrawerContent.tsx
ironAiken2: Removed in 05fd3188d together with the rest of the dead plumbing.

react/src/components/RoleDetailDrawer.tsx
```diff
@@ -3,7 +3,6 @@
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
 import { RoleDetailDrawerQuery } from '../__generated__/RoleDetailDrawerQuery.graphql';
-import { useSuspendedBackendaiClient } from '../hooks';
 import RoleDetailDrawerContent from './RoleDetailDrawerContent';
 import RoleFormModal from './RoleFormModal';
 import { Alert, Drawer, Skeleton, Tooltip, Typography, theme } from 'antd';
```
nowgnuesLee: <p class='graphite__hidden'><i>[Re: line +23]</i></p>
```gt_suggestion
  roleId: string || null;
```
I think null would be better than undefined when the value is absent.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8191?utm_source=unchanged-line-comment">Graphite</a>.</p>

react/src/components/RoleDetailDrawer.tsx
```diff
@@ -62,97 +61,43 @@ const RoleDetailDrawer: React.FC<RoleDetailDrawerProps> = ({
 interface RoleDetailDrawerInnerProps {
```
nowgnuesLee: <p class='graphite__hidden'><i>[Re: lines +52 to +56]</i></p>
Instead of separating them, please use a deferred value and drawer.loading.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8191?utm_source=unchanged-line-comment">Graphite</a>.</p>

react/src/components/RoleDetailDrawer.tsx
```diff
@@ -62,97 +61,43 @@ const RoleDetailDrawer: React.FC<RoleDetailDrawerProps> = ({
...
     },
   );
 
-  if (!data.adminRole) {
+  if (!adminRole) {
     return (
-      <Alert type="warning" showIcon message={t('general.ErrorOccurred')} />
+      <Alert type="warning" showIcon description={t('general.ErrorOccurred')} />
     );
   }
```
nowgnuesLee: Instead of this approach, the `@required` directive seems like a better fit.

react/src/helper/rbacGrantState.ts
nowgnuesLee, line 1: I don't think this function contains logic that warrants splitting it out.

react/src/helper/rbacGrantState.ts
```diff
@@ -0,0 +1,55 @@
...
+  if (uniqueFullOperations.size === 0) {
+    return 'none';
+  }
+
+  let grantedWithinMatrix = 0;
+  for (const operation of uniqueFullOperations) {
+    if (grantedOperations.has(operation)) {
+      grantedWithinMatrix += 1;
+    }
+  }
```
nowgnuesLee: Since it's a Set, the intersection method should be enough.

nowgnuesLee (dismissed):

react/src/helper/rbacGrantState.ts
ironAiken2: We'd like to keep the split: the helper exists to pin the tri-state edge behavior with unit tests (`rbacGrantState.test.ts`, 8 cases) — notably that stale grants outside the matrix don't count toward `partial`, and that an empty configurable set resolves to `none`. The body itself did get simpler though — the manual counting loop is now `Set.prototype.intersection` per your other comment.

react/src/components/RoleDetailDrawer.tsx
ironAiken2: Moot in the latest revision — the `roleId` prop is gone entirely. The drawer now renders from the selected list node's fragment ref (`roleFrgmt?: RoleDetailDrawerFragment$key | null`), so there is no id prop left to type.

react/src/components/RoleDetailDrawer.tsx
ironAiken2: Reworked in the latest revision: the `Inner` split is removed. The drawer renders from the list query's fragment data via `useRefetchableFragment`, so opening issues no fetch at all — there's nothing to defer or show `loading` for. The remaining `Suspense` only guards the tabs' own queries, and the refresh button refetches just this role inside a transition.

react/src/components/RoleDetailDrawer.tsx
ironAiken2: Gone in the latest revision — the `adminRole` query and its manual `!adminRole` guard were removed with the switch to the fragment-driven drawer. The remaining `role &&` gate is the closed-drawer state (null fragment ref), not error handling.

react/src/helper/rbacGrantState.ts
ironAiken2: Applied — the counting loop is now `uniqueFullOperations.intersection(grantedOperations).size`.

yomybaby (approved):

agatha197 (dismissed):
please resolve merge conflicts

PR #8193, opened 2026-07-07
fix: normalize vfolder host usage percentage fraction from manager >= 26.3

### LINKED ISSUE #8188: StorageSelect host usage status always shows "Adequate" with manager >= 26.3 (percentage is a 0-1 fraction)
## Summary

`StorageSelect` classifies host usage with `< 70` / `< 90` thresholds on `volume_info[<host>].usage.percentage`, assuming a 0–100 scale:

```tsx
// react/src/components/StorageSelect.tsx
const usagePercent = vhostInfo?.volume_info?.[host]?.usage?.percentage;
const usageLabel =
  usagePercent === undefined
    ? t('data.usage.Unknown')
    : usagePercent < 70
      ? t('data.usage.Adequate')
      : usagePercent < 90
        ? t('data.usage.Caution')
        : t('data.usage.Insufficient');
```

Since manager 26.3.0, `list_hosts` returns `percentage` as a **0–1 fraction** (the `* 100` was dropped in the BA-4820 refactor — see lablup/backend.ai#12632). A fraction is always `< 70`, so on managers ≥ 26.3 every storage host renders **"Adequate" / green regardless of actual usage**, and the `autoSelectType === 'usage'` sort operates on near-zero values.

## Expected resolution

The root cause is the manager regression (lablup/backend.ai#12632); once it is fixed, no WebUI change is strictly required.

Tracking here so WebUI can decide whether to:

1. wait for the manager fix (issue closes when lablup/backend.ai#12632 ships), or
2. add a defensive normalization (`percentage <= 1 ? percentage * 100 : percentage`) so the indicator works across manager versions already in the field (26.3–26.x).

Note: option 2 misclassifies a genuinely ≤1%-full host from a fixed manager as fraction-scaled, which is harmless (≤1% is "Adequate" either way).

## Affected versions

Any WebUI release paired with Backend.AI manager ≥ 26.3.

rapsealk:
Resolves #8188

## Symptom

With Backend.AI manager >= 26.3, every storage host in `StorageSelect` renders **"Adequate" / green regardless of actual usage**, and the `autoSelectType === 'usage'` auto-select sorts near-zero values (so "pick the least-used host" is meaningless). The same badge in the folder list's Host column (`VFolderNodesV2`) is affected.

## Root cause

Manager >= 26.3 returns `volume_info[<host>].usage.percentage` from `GET /folders/_/hosts` as a **0–1 fraction** instead of 0–100 — the `* 100` was dropped in lablup/backend.ai#9570. Tracked upstream in lablup/backend.ai#12632 (a manager-side fix is being opened in parallel). WebUI thresholds the value with `< 70` / `< 90` on a 0–100 scale, so a fraction is always "Adequate".

## Why fix it client-side too

Even after the manager fix ships, managers 26.3–26.x already deployed in the field keep returning fractions. This PR adds a defensive normalization (`percentage <= 1 ? percentage * 100 : percentage`) so the usage indicator and the usage-based auto-select work against **both fixed and unfixed managers**. The helper carries a `WORKAROUND` comment referencing lablup/backend.ai#12632 and should be removed once pre-fix managers are out of support.

**Known caveat (from the issue):** a genuinely <= 1%-full host reported by a fixed manager is misread as fraction-scaled. This only affects nearly-empty hosts, which is the harmless end of the scale for both the status buckets and the least-used sort.

## Changes

- `react/src/helper/index.tsx` — new shared `normalizeVFolderHostUsagePercent` helper with the WORKAROUND comment.
- `react/src/components/StorageSelect.tsx` — normalize `usagePercent` (drives the Adequate/Caution/Insufficient label and `StorageUsageBadge` color) and replace the string-path `_.minBy(..., 'volume_info.usage.percentage')` with a callback over the normalized value so the usage auto-select sorts on the same scale.
- `react/src/components/VFolderNodesV2.tsx` — normalize `usagePercent` in `VFolderHostCell` (same label + badge pattern).
- `react/src/helper/index.test.tsx` — unit tests for the helper (fraction scaling, 0/1 boundaries, 0–100 pass-through, `undefined`).

## Other `usage.percentage` consumers audited (no change needed)

- `QuotaPerStorageVolumePanelCard` / `QuotaPerStorageVolumeDashboardItem` / `HostQuotaTrigger` / `HostQuotaModalContent` — read only `capabilities` from `list_hosts()`; quota numbers come from GraphQL `quota_scope` bytes.
- `StorageHostResourcePanel` — computes percent from `used_bytes / capacity_bytes` (GraphQL `StorageVolume.usage` JSON), not from the REST percentage.
- `UsageProgress` — computes percent from quota-scope bytes.
- `diagnostics/rules/storageProxyRules.ts` — computes percentage from raw bytes itself.
- `useCurrentProject` / `useProjectResourceGroups` / BUI `BAIClientProvider` types — declare `usage.percentage` in types but only consume `sftp_scaling_groups`.
- `StorageUsageBadge` (backend.ai-ui) — thresholds its `percent` prop, but both call sites are the two components fixed above; normalization happens before the prop is passed, keeping the BUI package untouched.

## Verification

- `bash scripts/verify.sh`: Relay PASS, Lint PASS, Format PASS, TypeScript PASS. ("Vite warmup paths" FAIL is pre-existing on a clean `origin/main` checkout — reproduced via `git stash`; unrelated to this change.)
- `pnpm run test` in `react/`: 44 files, 915 tests passed (includes the 5 new helper tests).

Closes #8188

### DIFF
--- a/react/src/components/StorageSelect.tsx
+++ b/react/src/components/StorageSelect.tsx
@@ -2,6 +2,7 @@
  @license
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
+import { normalizeVFolderHostUsagePercent } from '../helper';
 import { useSuspendedBackendaiClient } from '../hooks';
 import { useSuspenseTanQuery } from '../hooks/reactQueryAlias';
 import useControllableState_deprecated from '../hooks/useControllableState';
@@ -88,7 +89,10 @@ const StorageSelect: React.FC<Props> = ({
           host,
           volume_info: vhostInfo?.volume_info?.[host],
         })),
-        'volume_info.usage.percentage',
+        // Normalize so fraction-scaled (manager >= 26.3) and 0-100 scaled
+        // percentages sort on the same scale. See `normalizeVFolderHostUsagePercent`.
+        (item) =>
+          normalizeVFolderHostUsagePercent(item.volume_info?.usage?.percentage),
       )?.host;
       nextHost = lowestUsageHost || nextHost;
     }
@@ -122,7 +126,9 @@ const StorageSelect: React.FC<Props> = ({
       }
       optionLabelProp={showUsageStatus ? 'label' : 'value'}
       options={_.map(vhostInfo?.allowed, (host) => {
-        const usagePercent = vhostInfo?.volume_info?.[host]?.usage?.percentage;
+        const usagePercent = normalizeVFolderHostUsagePercent(
+          vhostInfo?.volume_info?.[host]?.usage?.percentage,
+        );
         const usageLabel =
           usagePercent === undefined
             ? t('data.usage.Unknown')
--- a/react/src/components/VFolderNodesV2.tsx
+++ b/react/src/components/VFolderNodesV2.tsx
@@ -8,6 +8,7 @@ import {
   VFolderNodesV2Fragment$key,
 } from '../__generated__/VFolderNodesV2Fragment.graphql';
 import { VFolderNodesV2RestoreMutation } from '../__generated__/VFolderNodesV2RestoreMutation.graphql';
+import { normalizeVFolderHostUsagePercent } from '../helper';
 import { useSuspendedBackendaiClient, useWebUINavigate } from '../hooks';
 import { useCurrentUserInfo } from '../hooks/backendai';
 import { useSuspenseTanQuery, useTanQuery } from '../hooks/reactQueryAlias';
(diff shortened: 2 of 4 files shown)


rapsealk:
Closing without merge — superseded by the upstream manager fix lablup/backend.ai#12634, which restores the 0–100 percentage scale at the source. The defensive client-side normalization is no longer planned (#8188 closed as not planned). Branch kept for reference in case field deployments of manager 26.3–26.x without the fix make it needed after all.

PR #8228, opened 2026-07-09
chore(FR-3296): add e2e coverage for the top-bar project switcher

### LINKED ISSUE #8222: Add e2e coverage for the top-bar project switcher
Covers the switcher's visibility and dropdown contents, that switching projects scopes the page's data to the newly active project, and that the selection persists across navigation.

`e2e/project/project-switcher.spec.ts`. Verified: 8 passed (6 tests + 2 cleanup).

JIRA Issue: FR-3296

lizable:
Resolves #8222 (FR-3296)

Adds e2e coverage for the top-bar project switcher. All assertions compare the UI against the logged-in account's real project membership read from the client (`current_group` / `groups`) — no project name, folder name, or count is hardcoded, so the spec is portable across accounts and clusters.

## Covered E2E Scenarios

| # | Scenario | What it verifies |
|---|---|---|
| 1 | Admin sees the active project in the top bar | Switcher renders the "Project" label, the account's actual current project name, and the dropdown affordance |
| 2 | Regular user sees the active project | Same rendering contract under the user role, against that account's own membership |
| 3 | Admin opens the dropdown | Every project the client reports membership in is listed; the active one carries `aria-selected`; the system-level "model-store" catalog entry (admin-only, not real group membership) is verified structurally via the rendered dropdown when present; the search affordance exists |
| 4 | User opens the dropdown | The user's own memberships are listed with the active one selected; with a single membership and no model-store entry, no group headers render |
| 5 | Switching projects re-scopes the Data page (`@requires-seeded-data`) | After switching, the set of visible folder names actually changes (polled), the URL does not change (client-side switch), and switching back restores the original set — skips with an auditable reason when the account has fewer than 2 projects |
| 6 | Selection persists across navigation | A switched project survives a full route change within the same session; the original selection is restored afterward so no account-scoped state leaks between runs |

Tests run serially — the switcher selection is account-scoped shared state.

## How to verify

1. Copy `e2e/envs/.env.playwright.sample` to `e2e/envs/.env.playwright` and set `E2E_WEBUI_ENDPOINT` / `E2E_WEBSERVER_ENDPOINT`, `E2E_ADMIN_EMAIL` / `E2E_ADMIN_PASSWORD`, and `E2E_USER_EMAIL` / `E2E_USER_PASSWORD`.
2. Preconditions: the admin account should belong to **2+ projects** to exercise the switching scenarios (5 and 6) — with fewer they skip with an auditable `@requires-seeded-data` reason rather than fail. No specific project or folder names are required (everything is read from the account's own state). Read-only — the tests restore the original project selection.
3. Run (one invocation at a time — concurrent e2e runs against the same shared cluster degrade hydration):
   ```
   pnpm exec playwright test e2e/project/project-switcher.spec.ts
   ```
4. Expected: **8 passed** (6 spec tests + 2 global-cleanup teardown), ~3 min.

Note: the initial version of this spec hardcoded cluster-specific data (project names, a pre-existing folder name, folder counts); the second commit replaced that wholesale with dynamic reads from `client.current_group` / `client.groups`.

### DIFF
--- a/e2e/project/project-switcher.spec.ts
+++ b/e2e/project/project-switcher.spec.ts
@@ -1,6 +1,22 @@
 // spec: e2e/.agent-output/test-plan-project-switcher.md
 // scenarios: 1 (Project Switcher Display), 2 (Project Switcher Dropdown Contents),
 //            3 (Switching Project Scopes Data), 4 (Project Selection Persistence)
+//
+// Project names, project counts, and folder names are all account/cluster-
+// specific data -- an earlier version of this file hardcoded them (e.g. the
+// admin's active project as "RB테스트", a specific pre-existing folder
+// "bai-quota-sanity-2", and an assertion that no "bai-test-*" project was
+// present), which silently breaks on any account or cluster where that exact
+// data doesn't hold. Every test below instead reads the logged-in account's
+// actual project membership from the client (`current_group` / `groups`,
+// the same source components read via `useCurrentProjectValue()` /
+// `baiClient.current_group` -- see feature-gate-util.ts) and asserts the UI
+// matches that real data, rather than a guessed fixed value. The data-scope
+// test (3) goes further: instead of asserting a specific expected folder
+// count/name per project, it compares the *set* of visible folder names
+// before and after switching, which proves re-scoping happened without
+// needing to know or create any specific folder.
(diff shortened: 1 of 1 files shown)


yomybaby:
Superseded by #9730 (FR-3972), rewritten for the Astryx UI.

PR #8258, opened 2026-07-13
refactor(FR-3256): hoist AdminDeploymentPage tabs to preloaded queries

agatha197:
## Summary

Base of the FR-3256 deployment-admin stack. Builds `AdminDeploymentPage` with four tabs — **deployments, model store, prometheus presets, deployment presets** — each loading via a hoisted `preloadedQuery` owned by the page, so switching tabs preserves each tab's already-fetched rows (only a full page reload resets them). Includes the tab-naming refactor and the supporting client / UI / i18n changes.

Stack (bottom→top): **#8258 (this)** → #8336 (sort-convention cleanup) → #8148 (RuntimeVariantPreset CRUD tab). The RuntimeVariantPreset tab that used to sit at the bottom is now the top.

## Test plan
- [x] ESLint clean


agatha197:
* **#8148** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8148?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8336** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8336?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8258** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8258?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8258).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

nowgnuesLee (changes requested):
Please resolve the copilot's reviews

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+    };
+    // Reset to defaults first, then apply the target tab's snapshot, so no
+    // stale `filter`/`order` from the previous tab bleeds through.
+    setQueryParams(null);
+    setQueryParams({
+      tab: key as TabKey,
+      filter: stored.queryParams.filter,
+      order: stored.queryParams.order,
+    });
+    setTablePaginationOption(stored.tablePaginationOption ?? { current: 1 });
```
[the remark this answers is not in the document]
agatha197: Fixed in the latest commit. The tab-switch fallback now resets both `current` and `pageSize` to the page defaults (`{ current: 1, pageSize: 10 }`), so a tab opened for the first time no longer inherits the previous tab's page size from the shared pagination state.

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+  };
+
+  const ensureDeploymentsLoaded = useEffectEvent(() => {
+    if (currentTab === 'deployments' && !deploymentQueryRef) {
+      const sort = parseDeploymentOrder(queryParams.order);
+      loadDeploymentQuery(
+        {
+          filter:
+            (queryParams.filter as DeploymentFilter | null) ??
+            DEPLOYMENT_RUNNING_FILTER,
```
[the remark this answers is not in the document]
agatha197: The running/finished scope is owned by the deployments tab's `filter`: the child `AdminDeployment` merges the status condition into `filter` whenever the user toggles running/finished, and `DEPLOYMENT_RUNNING_FILTER` is only the first-load default applied when there is no persisted filter at all. So a normal in-app flow always carries an explicit status condition; the "user filter without a status key" case only arises from hand-crafted/legacy URLs. Reconciling the radio with such partial URLs would belong to the status-model, not this preloaded-query refactor — noting it as a possible follow-up.

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+  // deliberately avoided: with one shared key set, an inactive tab can never
+  // leak its `filter`/`order` into another tab (the previous per-tab
+  // `useQueryStates` design leaked because every mounted tab read the same
+  // unprefixed `filter`/`order` keys simultaneously).
+  const [queryParams, setQueryParams] = useQueryStates(
+    {
+      tab: tabParser,
+      order: parseAsString,
+      filter: parseAsJson<unknown>((value) => value),
+    },
```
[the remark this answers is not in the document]
agatha197: The previous standalone page encoded the toggle in a shareable `statusCategory` query key; the new tabbed page models status inside the deployments `filter` instead and does not hydrate that legacy key. Translating old `?statusCategory=` bookmarks is out of scope for this refactor and is noted as a possible follow-up.

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+
+  // Every tab's query ref is owned here (not inside the tab component) so the
+  // ref — and its already-fetched rows — survive the tab unmounting on a tab
+  // switch; only a full page reload resets it.
+
+  // --- Deployments tab ---
+  const [deploymentQueryRef, loadDeploymentQuery] =
+    useQueryLoader<AdminDeploymentQueryType>(AdminDeploymentQuery);
+  const [deploymentColumnOverrides, setDeploymentColumnOverrides] =
+    useBAISettingUserState('table_column_overrides.AdminDeployment');
```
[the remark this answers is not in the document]
agatha197: Acknowledged. This `table_column_overrides.*` key changed because the component was renamed/rewritten as part of this refactor (the standalone list-page component became a hoisted preloaded-query tab). We are intentionally accepting a one-time reset of the per-user column layout preference here rather than pinning the legacy storage key, since the table was substantially reworked. If preserving the old preferences matters, we can map the legacy keys in a follow-up.

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+        { fetchPolicy: 'store-and-network' },
+      );
+    }
+  });
+
+  // --- Model store management tab ---
+  const [modelCardQueryRef, loadModelCardQuery] =
+    useQueryLoader<AdminModelCardTableQueryType>(AdminModelCardTableQuery);
+  const [modelCardColumnOverrides, setModelCardColumnOverrides] =
+    useBAISettingUserState('table_column_overrides.AdminModelCardTable');
```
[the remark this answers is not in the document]
agatha197: Acknowledged. This `table_column_overrides.*` key changed because the component was renamed/rewritten as part of this refactor (the standalone list-page component became a hoisted preloaded-query tab). We are intentionally accepting a one-time reset of the per-user column layout preference here rather than pinning the legacy storage key, since the table was substantially reworked. If preserving the old preferences matters, we can map the legacy keys in a follow-up.

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+        { fetchPolicy: 'store-and-network' },
+      );
+    }
+  });
+
+  // --- Prometheus preset tab ---
+  const [prometheusQueryRef, loadPrometheusQuery] =
+    useQueryLoader<AdminPrometheusPresetQueryType>(AdminPrometheusPresetQuery);
+  const [prometheusColumnOverrides, setPrometheusColumnOverrides] =
+    useBAISettingUserState('table_column_overrides.AdminPrometheusPreset');
```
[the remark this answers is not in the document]
agatha197: Acknowledged. This `table_column_overrides.*` key changed because the component was renamed/rewritten as part of this refactor (the standalone list-page component became a hoisted preloaded-query tab). We are intentionally accepting a one-time reset of the per-user column layout preference here rather than pinning the legacy storage key, since the table was substantially reworked. If preserving the old preferences matters, we can map the legacy keys in a follow-up.

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,476 @@
...
+        { fetchPolicy: 'store-and-network' },
+      );
+    }
+  });
+
+  // --- Deployment presets tab ---
+  const [deploymentPresetQueryRef, loadDeploymentPresetQuery] =
+    useQueryLoader<AdminDeploymentPresetQueryType>(AdminDeploymentPresetQuery);
+  const [deploymentPresetColumnOverrides, setDeploymentPresetColumnOverrides] =
+    useBAISettingUserState('table_column_overrides.AdminDeploymentPreset');
```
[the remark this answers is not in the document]
agatha197: Acknowledged. This `table_column_overrides.*` key changed because the component was renamed/rewritten as part of this refactor (the standalone list-page component became a hoisted preloaded-query tab). We are intentionally accepting a one-time reset of the per-user column layout preference here rather than pinning the legacy storage key, since the table was substantially reworked. If preserving the old preferences matters, we can map the legacy keys in a follow-up.

react/src/helper/index.tsx
```diff
@@ -987,3 +987,30 @@ export const convertToOrderBy = <
...
+ */
+export const convertOrderByToString = (
+  orderBy:
+    | ReadonlyArray<{ field?: string; direction?: string }>
+    | null
+    | undefined,
+): string | null => {
+  const first = orderBy?.[0];
+  if (!first?.field) return null;
+  return `${first.direction === 'DESC' ? '-' : ''}${_.camelCase(first.field)}`;
```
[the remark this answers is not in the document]
agatha197: Added unit tests for `convertOrderByToString` in `react/src/helper/index.test.tsx`, covering empty/`null`/`undefined` input, ASC/DESC, SCREAMING_SNAKE_CASE enum fields (e.g. `CREATED_AT` → `createdAt` / `-createdAt`), and multi-entry input (first entry only).

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,482 @@
...
+  // across tab switches (this page never unmounts), so returning to a tab shows
+  // the same rows without refetching.
+  useEffect(
+    function loadActiveTabOnActivation() {
+      ensureDeploymentsLoaded();
+      ensureModelCardsLoaded();
+      ensurePrometheusPresetsLoaded();
+      ensureDeploymentPresetsLoaded();
+    },
+    [currentTab, currentProject.id],
```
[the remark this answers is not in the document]
agatha197: By design. Filter/sort/paging changes flow through `onReload` (the child calls it, which re-runs `loadQuery` and swaps the query ref), so the data does react to those changes. The activation effect intentionally performs only the initial lazy load per tab and is guarded by the ref so it never refetches on re-activation. Not reacting to raw URL-state edits (e.g. browser back/forward) while staying on the same tab is a known limitation of the hoisted-preloaded-query pattern — consistent with `AdminComputeSessionListPage` — since a full URL→refetch sync would defeat the ref persistence this page is built around. Noted as a possible follow-up.

react/src/components/AdminDeployment.tsx
```diff
@@ -73,114 +120,37 @@ const AdminDeploymentListPageContent: React.FC = () => {
...
-      ]).withDefault('running'),
-    },
-    { history: 'replace' },
-  );
-
-  const [columnOverrides, setColumnOverrides] = useBAISettingUserState(
-    'table_column_overrides.AdminDeploymentListPage',
-  );
-
   const [fetchKey, updateFetchKey] = useFetchKey();
```
nowgnuesLee: we don't need to use `useFetchKey`

react/src/components/AdminDeployment.tsx
```diff
@@ -460,7 +470,10 @@ const AdminDeploymentListPageContent: React.FC = () => {
         deploymentFrgmt={editingDeployment ?? null}
         onRequestClose={(success) => {
           setEditingDeploymentId(null);
-          if (success) updateFetchKey();
+          if (success) {
+            updateFetchKey();
+            onReload(queryRef.variables, { fetchPolicy: 'network-only' });
+          }
```
nowgnuesLee, line 470: Please handle the reload separately for create and update.

react/src/components/AdminDeploymentPreset.tsx
```diff
@@ -46,86 +86,28 @@ const AdminDeploymentPresetListPage: React.FC = () => {
   const baiClient = useSuspendedBackendaiClient();
   const webuiNavigate = useWebUINavigate();
 
-  const [columnOverrides, setColumnOverrides] = useBAISettingUserState(
-    'table_column_overrides.AdminDeploymentPresetListPage',
-  );
+  const [fetchKey, updateFetchKey] = useFetchKey();
```
nowgnuesLee: we don't need to use `useFetchKey`

react/src/components/AdminModelCard.tsx
nowgnuesLee, line 1: This doesn't return only a table, so putting "table" in the file name isn't appropriate.

react/src/components/AdminModelCardTable.tsx
```diff
@@ -89,115 +154,40 @@ const AdminModelCardListPage: React.FC = () => {
...
-      filter: parseAsJson<ModelCardV2Filter>(
-        (value) => value as ModelCardV2Filter,
-      ),
-    },
-    {
-      history: 'replace',
-    },
-  );
 
   const [fetchKey, updateFetchKey] = useFetchKey();
```
nowgnuesLee: we don't need to use `useFetchKey`

react/src/components/AdminPrometheusPreset.tsx
```diff
@@ -2,143 +2,129 @@
...
-  const {
-    baiPaginationOption,
-    tablePaginationOption,
-    setTablePaginationOption,
-  } = useBAIPaginationOptionStateOnSearchParam({
-    current: 1,
-    pageSize: 10,
-  });
-
   const [fetchKey, updateFetchKey] = useFetchKey();
```
nowgnuesLee: we don't need to use `useFetchKey`

nowgnuesLee (changes requested):

react/src/pages/AdminDeploymentPage.tsx
nowgnuesLee: I think we can handle the reload with a conditional in a single function.

nowgnuesLee (changes requested):

react/src/components/AdminModelCard.tsx
agatha197: It returns BAITable.

react/src/components/AdminDeployment.tsx
agatha197: Removed `useFetchKey` here. `BAIFetchKeyButton` doesn't actually read its `value` prop — it self-generates the key on each refresh (`onChange(new Date().toISOString())`) and re-anchors its countdown via an internal `cycleKey` — and the refetch is driven directly by `onReload(queryRef.variables, { fetchPolicy: 'network-only' })`. So the hook was redundant: `onChange` now just calls `onReload`, and I made `BAIFetchKeyButton.value` optional so it can be omitted.

react/src/components/AdminDeploymentPreset.tsx
agatha197: Removed `useFetchKey` here. `BAIFetchKeyButton` doesn't actually read its `value` prop — it self-generates the key on each refresh (`onChange(new Date().toISOString())`) and re-anchors its countdown via an internal `cycleKey` — and the refetch is driven directly by `onReload(queryRef.variables, { fetchPolicy: 'network-only' })`. So the hook was redundant: `onChange` now just calls `onReload`, and I made `BAIFetchKeyButton.value` optional so it can be omitted.

react/src/components/AdminModelCardTable.tsx
agatha197: Removed `useFetchKey` here. `BAIFetchKeyButton` doesn't actually read its `value` prop — it self-generates the key on each refresh (`onChange(new Date().toISOString())`) and re-anchors its countdown via an internal `cycleKey` — and the refetch is driven directly by `onReload(queryRef.variables, { fetchPolicy: 'network-only' })`. So the hook was redundant: `onChange` now just calls `onReload`, and I made `BAIFetchKeyButton.value` optional so it can be omitted.

react/src/components/AdminPrometheusPreset.tsx
agatha197: Removed `useFetchKey` here. `BAIFetchKeyButton` doesn't actually read its `value` prop — it self-generates the key on each refresh (`onChange(new Date().toISOString())`) and re-anchors its countdown via an internal `cycleKey` — and the refetch is driven directly by `onReload(queryRef.variables, { fetchPolicy: 'network-only' })`. So the hook was redundant: `onChange` now just calls `onReload`, and I made `BAIFetchKeyButton.value` optional so it can be omitted.

react/src/components/AdminDeployment.tsx
agatha197: Kept a single always-refetch path here. Distinguishing create vs update isn't strictly necessary for the reload. If it ever were, the cleaner approach would be to pass the created/updated node to `onRequestClose` so the caller can act on that specific node — rather than threading a create/update flag down. Leaving it as one refetch for now.

react/src/components/AdminModelCard.tsx
agatha197: This component renders `<BAITable>` directly, unlike the sibling tabs which delegate their table to a fragment child (e.g. `AdminDeployment` → `BAIModelDeploymentNodes`). So it *is* the table-owning component, and the `...Table` name follows the convention we've applied elsewhere in this stack (a component that owns/renders a `BAITable` is named `...Table`). Keeping the name.

react/src/pages/AdminDeploymentPage.tsx
agatha197: Done — restructured to render-as-you-fetch. `loadTab(tab, params, pagination)` is now a single guarded function; `onTabChange` calls it directly in the click handler (so the fetch starts before the re-render, not in an effect reacting to the tab change). The `useEffect` is kept only for entries that don't go through `onTabChange` — initial mount, full reload, and direct `?tab=...` URLs — and is keyed on `currentProject.id` (not `currentTab`) so the project-scoped model-store tab still reloads on project change without double-loading on a plain tab switch. The five `ensure*Loaded` were collapsed into the one `loadTab` conditional.

react/src/components/AdminModelCard.tsx
nowgnuesLee, line 1: This component doesn’t return just a table, so using the `...Table` suffix goes against our naming convention. `AdminModelCard` should be enough.

react/src/helper/index.tsx
```diff
@@ -987,3 +987,30 @@ export const convertToOrderBy = <
...
+};
```
nowgnuesLee: The function name doesn't make it clear that it only converts the first entry. Consider changing the signature to accept a single value rather than an array.

react/src/pages/AdminDeploymentPage.tsx
```diff
@@ -0,0 +1,499 @@
...
+  // deliberately avoided: with one shared key set, an inactive tab can never
+  // leak its `filter`/`order` into another tab (the previous per-tab
+  // `useQueryStates` design leaked because every mounted tab read the same
+  // unprefixed `filter`/`order` keys simultaneously).
+  const [queryParams, setQueryParams] = useQueryStates(
+    {
+      tab: tabParser,
+      order: parseAsString,
+      filter: parseAsJson<unknown>((value) => value),
+    },
```
nowgnuesLee, line 125: how about using `useTabQuerySnapshot`

nowgnuesLee (changes requested):

react/src/components/AdminModelCard.tsx
agatha197: This component renders the `<BAITable>` directly — `AdminModelCardTable.tsx:401` (`<BAITable<ModelCardNode> dataSource={...} columns={columns} />`) — rather than delegating it to a fragment child. So the `...Table` suffix names the component that actually owns and renders the model-card table; the filter / refresh / create controls and the setting/delete modals are the standard chrome around it. Keeping `AdminModelCardTable`.

react/src/pages/AdminDeploymentPage.tsx
agatha197: We intentionally moved off `useTabQuerySnapshot` here. It implements the tab-URL "whole query-string swap" pattern (A-ii), which is safe only when each tab is a self-contained child that unmounts while inactive. This page hoists every tab's preloaded `useQueryLoader` state to the parent, so all tabs are mounted at once — and under that setup `useTabQuerySnapshot`'s whole-query swap let one tab's `filter`/`order` leak into another (the cross-tab filter-persistence bug this refactor was fixing). The current single `useQueryStates({ tab, filter, order })` + `queryMapRef` snapshot (A-i, the `AdminComputeSessionListPage` idiom) keeps each tab's state isolated, so `useTabQuerySnapshot` would reintroduce that leak here.

react/src/helper/index.tsx
agatha197: Renamed `convertOrderByToString` → `firstOrderByToString` so the name makes the first-entry-only behavior explicit. Kept the array parameter (rather than switching to a single value): every caller has `variables.orderBy` (the array) in hand, and it mirrors `convertToOrderBy`'s array return, so an array signature avoids adding `?.[0]` at ~9 call sites. The single-column-sort assumption is why only the first entry is used.

react/src/components/AdminDeployment.tsx
```diff
@@ -460,7 +465,9 @@ const AdminDeploymentListPageContent: React.FC = () => {
         deploymentFrgmt={editingDeployment ?? null}
         onRequestClose={(success) => {
           setEditingDeploymentId(null);
-          if (success) updateFetchKey();
+          if (success) {
+            onReload(queryRef.variables, { fetchPolicy: 'network-only' });
+          }
         }}
```
nowgnuesLee, line 471: This is an edit, not a create, so there's no need to refetch.

Instead, please enumerate the updated fields in DeploymentSettingModal's update mutation — Relay will then update the store from the mutation response.

```suggestion
          if (success) {
            onReload(queryRef.variables, { fetchPolicy: 'network-only' });
          }
        }}
```

react/src/components/AdminModelCard.tsx
```diff
@@ -427,36 +417,44 @@ const AdminModelCardListPage: React.FC = () => {
...
+          isModelStoreProject={group?.type === 'MODEL_STORE'}
+          modelStoreProject={groups?.[0] ?? null}
           onRequestClose={(success) => {
             setIsSettingModalOpen(false);
             setEditingModelCardId(null);
             if (success) {
-              updateFetchKey();
+              onReload(queryRef.variables, { fetchPolicy: 'network-only' });
             }
           }}
```
nowgnuesLee, line 459: This is an edit, not a create, so there's no need to refetch.

react/src/components/AdminModelCard.tsx
agatha197: Done — renamed `AdminModelCardTable` → `AdminModelCard` (component, its `AdminModelCardQuery` / delete mutations, props, the file, the `table_column_overrides.*` setting key, and the `AdminDeploymentPage` importer; Relay artifacts regenerated). Pushed across the stack.

nowgnuesLee (changes requested):

PR #8283, opened 2026-07-14
feat: increase service launcher health check initial delay default to 600s

nowgnuesLee:
## Summary
- Increase the default health-check `initial_delay` for the service launcher from **60s to 600s**.
- Applied consistently across all default paths so create and edit flows stay in sync:
  - `ServiceLauncherPageContent.tsx` — form initial values (create/edit) and submit-time `??` fallbacks (5 sites)
  - `generateModelDefinitionYaml.ts` — generator fallback and doc example
  - `parseModelDefinitionYaml.ts` — parser fallback when the YAML omits `initial_delay`
  - `parseModelDefinitionYaml.test.ts` — default-value test expectation updated
- Existing services with an explicitly stored `initial_delay` keep their value; only the default for unset values changes.

## Test plan
- [x] `jest` — `parseModelDefinitionYaml` suite passes (9/9)
- [ ] Create a new service in the service launcher and confirm the Initial Delay field defaults to 600.
- [ ] Edit an existing service whose model-definition.yaml omits `initial_delay` and confirm the field shows 600 (existing explicit values remain unchanged).

### DIFF
--- a/react/src/components/ServiceLauncherPageContent.tsx
+++ b/react/src/components/ServiceLauncherPageContent.tsx
@@ -683,7 +683,7 @@ const ServiceLauncherPageContent: React.FC<ServiceLauncherPageContentProps> = ({
           port: values.commandPort ?? 8000,
           healthCheckPath: values.commandHealthCheck ?? '/health',
           modelMountDestination: values.commandModelMount ?? '/models',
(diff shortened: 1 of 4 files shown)


yomybaby (approved):
LGTM

PR #8285, opened 2026-07-14
fix(FR-3326): reset Prometheus query preset modal on reopen

### LINKED ISSUE #8282: Prometheus query preset create/edit modal state not reset on open/close
On the admin Prometheus query preset page, the create/edit modal's state is not reflected immediately when the modal is opened or closed — it retains stale values from the previous open (form not reset / unmount-after-close not applied).

Additionally, audit other recently added modals for the same problem class and fix any that share it.

Notes:

- Same bug class as FR-3094 (Deployment create/edit modal form not reset on open).
- Likely fix: wrap the modal with `BAIUnmountAfterClose` and/or key the form state on an id-based open state, per the `react-modal-drawer` conventions.

----

Captured while working on branch: feat/FR-3279-router-owned-route-errors

JIRA Issue: FR-3326

agatha197:
Resolves #8282 (FR-3326)

## Summary

The admin Prometheus query preset create/edit modal retained stale values from a previous open (e.g. opening the Create modal after editing a preset still showed the prior values).

Root cause: `PrometheusQueryPresetEditorModal` was kept **always-mounted** by `PrometheusPresetTab` (to avoid suspending the surrounding list when its categories query loaded). Its form instance therefore survived close, and Ant Design applies `initialValues` only once per instance lifetime — so later opens with different `initialValues` were ignored. `destroyOnHidden` / `preserve={false}` unmount the `<Form>` body but not the persistent instance, so they did not fix it (same failure mode as FR-3094).

## Approach

Fully unmount the modal on close so a fresh `Form.useForm()` instance re-applies `initialValues` on every open, and let the category picker fetch its own data behind a **local Suspense boundary** instead of preloading it in the parent:

- **`PrometheusQueryPresetEditorModal`** is wrapped in **`BAIUnmountAfterClose`** by `PrometheusPresetTab`, so it fully unmounts on close — a **fresh `Form.useForm()` instance re-applies `initialValues` on every open**, and stale state can no longer carry across reopens. This is the structural staleness fix (deterministic client-side behavior, not dependent on close-animation timing).
- **`PrometheusCategorySelect`** (new) fetches the categories query itself with `useLazyLoadQuery` and **owns its own `Suspense` boundary**. Opening the editor is now a plain state update — no `fetchQuery` / `useQueryLoader` / `usePreloadedQuery` orchestration — and the surrounding preset list never suspends while categories load (the field shows a loading Select).

## Review feedback addressed

- @nowgnuesLee — "No need to manually await `fetchQuery` here — the action already handles the transition." → removed all preloading; opening the editor is a plain state update.
- @nowgnuesLee — "Create a Select component that fetches the query itself, and wrap the select in Suspense." → added `PrometheusCategorySelect`, which self-fetches inside its own Suspense boundary.

## Changes (4 files)

- **`PrometheusCategorySelect.tsx`** (new): self-fetching category picker wrapped in its own Suspense boundary; forwards the `Form.Item` control props through to the inner `BAISelect`, and shows a loading Select as the fallback.
- **`PrometheusQueryPresetEditorModal.tsx`**: dropped the exported categories query, the `categoriesQueryRef` prop, and `usePreloadedQuery`; the Category field now renders `<PrometheusCategorySelect>`. The form staleness fix (fresh `Form.useForm()` per open via unmount) is unchanged.
- **`PrometheusPresetTab.tsx`**: removed the `fetchQuery` + `useQueryLoader` preload dance; `openEditor` is now a plain state update, the Create/Edit triggers use `onClick`, and the modal is always wrapped in `BAIUnmountAfterClose`.
- **`PrometheusQueryPresetNodes.tsx`**: the row Edit action reverts to a synchronous `onClick`.

## How to test

**Test against a production bundle, not `pnpm dev`.** The staleness reproduces reliably only in a real build; dev-server HMR / React Fast Refresh remounts the form instance on edit and masks it.

```bash
# on this branch, from repo root
pnpm install
pnpm run build            # → build/web/
npx serve -s build/web    # or any static file server pointed at build/web
```

Then, logged in as an admin:

1. Open the Deployments admin page → **Prometheus Preset** tab (shown only when the manager supports Prometheus presets).
2. Click a preset's **Edit** action or **Create** — the modal opens immediately; the Category field shows a brief loading state on the first open while its categories load.
3. Edit a field (e.g. the query template), then **Cancel** / close.
4. Reopen — edit a different preset, or click **Create**.
   - **Before this fix:** the modal still shows the previous open's values.
   - **After this fix:** the modal shows the reopened target's own `initialValues` (Create → empty; Edit → that preset's values).

## Verification

- `pnpm relay`, `tsc` (no `react/src` errors), ESLint (0 on changed files), Prettier — all pass on the changed files.
- Note: `scripts/verify.sh` reports pre-existing, unrelated failures present on `main` (a `gpt-tokenizer` `.d.ts` parse issue and React-Compiler lint findings in untouched `packages/backend.ai-ui` components).

### DIFF
--- a/react/src/components/PrometheusCategorySelect.tsx
+++ b/react/src/components/PrometheusCategorySelect.tsx
@@ -1,0 +1,56 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+import { PrometheusCategorySelectQuery } from '../__generated__/PrometheusCategorySelectQuery.graphql';
+import { BAISelect, BAISelectProps } from 'backend.ai-ui';
+import * as _ from 'lodash-es';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { graphql, useLazyLoadQuery } from 'react-relay';
+
+/**
+ * Category picker for the Prometheus query preset editor. Fetches the category
+ * list itself via `useLazyLoadQuery`, so it suspends until the list resolves —
+ * the caller owns the `<Suspense>` boundary (see `PrometheusQueryPresetEditorModal`)
+ * so it can scope the fallback to just this field and keep the rest of the form
+ * interactive. Owns the category-picker defaults (clearable, empty-state copy);
+ * control props injected by the enclosing `Form.Item` flow through to the
+ * underlying select, and callers can still override any default via props.
+ */
+const PrometheusCategorySelect: React.FC<BAISelectProps> = (props) => {
+  'use memo';
+  const { t } = useTranslation();
+  const { prometheusQueryPresetCategories } =
+    useLazyLoadQuery<PrometheusCategorySelectQuery>(
+      graphql`
+        query PrometheusCategorySelectQuery {
+          prometheusQueryPresetCategories {
+            id
+            name
+          }
+        }
+      `,
+      {},
+    );
+
+  // The list query may legitimately return an empty array in dev environments
+  // where no categories are seeded; the form must still submit fine with
+  // categoryId left null.
+  const options = _.map(prometheusQueryPresetCategories ?? [], (category) => ({
+    label: category.name,
+    value: category.id,
+  }));
+
+  return (
+    <BAISelect
+      allowClear
+      placeholder={t('prometheusQueryPreset.NoCategory')}
+      notFoundContent={t('prometheusQueryPreset.NoCategory')}
+      options={options}
+      {...props}
+    />
+  );
+};
+
+export default PrometheusCategorySelect;
--- a/react/src/components/PrometheusPresetTab.tsx
+++ b/react/src/components/PrometheusPresetTab.tsx
@@ -8,7 +8,6 @@ import {
   PrometheusPresetTabQuery$variables,
   QueryDefinitionFilter,
 } from '../__generated__/PrometheusPresetTabQuery.graphql';
-import { PrometheusQueryPresetEditorModalFragment$key } from '../__generated__/PrometheusQueryPresetEditorModalFragment.graphql';
 import { convertToOrderBy } from '../helper';
 import { useBAIPaginationOptionStateOnSearchParam } from '../hooks/reactPaginationQueryOptions';
 import { useBAISettingUserState } from '../hooks/useBAISetting';
@@ -16,13 +15,16 @@ import AutoUpdateFetchKeyButton, {
   LONG_AUTO_UPDATE_DELAY_OPTIONS,
 } from './AutoUpdateFetchKeyButton';
 import PrometheusQueryPresetEditorModal from './PrometheusQueryPresetEditorModal';
-import PrometheusQueryPresetNodes from './PrometheusQueryPresetNodes';
+import PrometheusQueryPresetNodes, {
+  PrometheusQueryPresetNodeInList,
+} from './PrometheusQueryPresetNodes';
 import { App } from 'antd';
 import {
   BAIButton,
   BAIDeleteConfirmModal,
   BAIFlex,
   BAIGraphQLPropertyFilter,
+  BAIUnmountAfterClose,
   INITIAL_FETCH_KEY,
   toLocalId,
   useFetchKey,
@@ -63,12 +65,20 @@ const PrometheusPresetTab: React.FC = () => {
   const [fetchKey, updateFetchKey] = useFetchKey();
   const [isOpenEditorModal, setIsOpenEditorModal] = useState(false);
   const [editingPreset, setEditingPreset] =
-    useState<PrometheusQueryPresetEditorModalFragment$key | null>(null);
+    useState<PrometheusQueryPresetNodeInList | null>(null);
   const [deletingPreset, setDeletingPreset] =
     useState<DeletingPresetTarget | null>(null);
   const [columnOverrides, setColumnOverrides] = useBAISettingUserState(
     'table_column_overrides.PrometheusPresetTab',
   );
+
+  // The editor's category picker (`PrometheusCategorySelect`) fetches its own
+  // query inside its own Suspense boundary, so opening is a plain state update —
+  // no preloading, and the surrounding preset list never suspends.
+  const openEditor = (preset?: PrometheusQueryPresetNodeInList | null) => {
+    setEditingPreset(preset ?? null);
+    setIsOpenEditorModal(true);
+  };
 
   const [commitDeleteMutation, isInflightDelete] =
     useMutation<PrometheusPresetTabDeleteMutation>(graphql`
@@ -164,7 +174,7 @@ const PrometheusPresetTab: React.FC = () => {
           <BAIButton
             type="primary"
             icon={<PlusIcon />}
-            onClick={() => setIsOpenEditorModal(true)}
+            onClick={() => openEditor()}
           >
             {t('prometheusQueryPreset.CreatePreset')}
           </BAIButton>
@@ -191,25 +201,24 @@ const PrometheusPresetTab: React.FC = () => {
           columnOverrides,
           onColumnOverridesChange: setColumnOverrides,
         }}
-        onEditPreset={(preset) => {
-          setIsOpenEditorModal(true);
-          setEditingPreset(preset);
-        }}
+        onEditPreset={(preset) => openEditor(preset)}
         onDeletePreset={(preset) => {
           setDeletingPreset({ id: preset.id, name: preset.name });
         }}
       />
-      <PrometheusQueryPresetEditorModal
-        open={isOpenEditorModal}
-        presetFrgmt={editingPreset}
-        onRequestClose={(success) => {
-          setIsOpenEditorModal(false);
-          setEditingPreset(null);
-          if (success) {
-            updateFetchKey();
-          }
-        }}
-      />
+      <BAIUnmountAfterClose>
+        <PrometheusQueryPresetEditorModal
+          open={isOpenEditorModal}
+          presetFrgmt={editingPreset}
+          onRequestClose={(success) => {
+            setIsOpenEditorModal(false);
+            setEditingPreset(null);
+            if (success) {
+              updateFetchKey();
+            }
+          }}
+        />
+      </BAIUnmountAfterClose>
       <BAIDeleteConfirmModal
         open={!!deletingPreset}
         title={t('prometheusQueryPreset.PermanentlyDeletePreset', {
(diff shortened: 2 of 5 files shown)


agatha197:
* **#8286** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8286?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8285** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8285?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8285?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8285).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/PrometheusPresetTab.tsx
```diff
@@ -70,6 +81,29 @@ const PrometheusPresetTab: React.FC = () => {
...
+  // store, then `loadQuery` resolves from it instantly and the modal's
+  // `usePreloadedQuery` never suspends the surrounding list.
+  const openEditor = async (
+    preset?: PrometheusQueryPresetEditorModalFragment$key | null,
+  ) => {
+    await fetchQuery(
+      relayEnv,
+      PrometheusQueryPresetCategoriesQuery,
+      {},
+    ).toPromise();
```
nowgnuesLee: No need to manually await fetchQuery here — the action already handles the transition.

react/src/components/PrometheusQueryPresetEditorModal.tsx
```diff
@@ -110,27 +125,13 @@ const PrometheusQueryPresetEditorModal: React.FC<
...
-      {},
-      {
-        fetchPolicy:
-          deferredOpen && baiModalProps.open
-            ? 'store-and-network'
-            : 'store-only',
-      },
+    usePreloadedQuery<PrometheusQueryPresetEditorModalCategoriesQuery>(
+      PrometheusQueryPresetCategoriesQuery,
+      categoriesQueryRef,
```
nowgnuesLee: How about creating a Select component that fetches the query itself? And wrap the select in Suspense

nowgnuesLee (changes requested):

react/src/components/PrometheusPresetTab.tsx
agatha197: Done — removed the manual `fetchQuery` + `useQueryLoader` preload entirely. Opening the editor is now a plain state update; the category data is fetched lazily by the new `PrometheusCategorySelect` behind its own Suspense boundary, so nothing needs to be preloaded on the action. (de26fff7a)

react/src/components/PrometheusQueryPresetEditorModal.tsx
agatha197: Done — added `PrometheusCategorySelect`, which runs `useLazyLoadQuery` itself and owns its own `<Suspense>` boundary (with a loading Select fallback). The modal now just renders `<PrometheusCategorySelect>` inside the Category `Form.Item`, so opening the editor never suspends the surrounding preset list.

react/src/components/PrometheusCategorySelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * props injected by the enclosing `Form.Item` flow through to the inner select.
+ */
+const PrometheusCategorySelect: React.FC<BAISelectProps> = (props) => {
+  'use memo';
+  return (
+    <Suspense fallback={<BAISelect {...props} loading disabled />}>
+      <PrometheusCategorySelectInner {...props} />
+    </Suspense>
+  );
+};
```
nowgnuesLee: let the caller wrap it in suspense instead.

react/src/components/PrometheusPresetTab.tsx
```diff
@@ -70,6 +71,16 @@ const PrometheusPresetTab: React.FC = () => {
     'table_column_overrides.PrometheusPresetTab',
   );
 
+  // The editor's category picker (`PrometheusCategorySelect`) fetches its own
+  // query inside its own Suspense boundary, so opening is a plain state update —
+  // no preloading, and the surrounding preset list never suspends.
+  const openEditor = (
+    preset?: PrometheusQueryPresetEditorModalFragment$key | null,
```
nowgnuesLee: Please don't use the fragment key this way — it's already included in the node type of PrometheusPresetTabQuery.

react/src/components/PrometheusPresetTab.tsx
```diff
@@ -70,6 +71,16 @@ const PrometheusPresetTab: React.FC = () => {
     'table_column_overrides.PrometheusPresetTab',
```
nowgnuesLee, line 72: <p class='graphite__hidden'><i>[Re: line +67]</i></p>
Please don't use the fragment key this way — it's already included in the node type of PrometheusPresetTabQuery.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8285?utm_source=unchanged-line-comment">Graphite</a>.</p>

react/src/components/PrometheusPresetTab.tsx
```diff
@@ -191,25 +202,24 @@ const PrometheusPresetTab: React.FC = () => {
...
+      <BAIUnmountAfterClose>
+        <PrometheusQueryPresetEditorModal
+          open={isOpenEditorModal}
+          presetFrgmt={editingPreset}
+          onRequestClose={(success) => {
+            setIsOpenEditorModal(false);
+            setEditingPreset(null);
+            if (success) {
+              updateFetchKey();
+            }
```
nowgnuesLee, line 218: We need to refetch differently depending on whether it's an update or a create.

nowgnuesLee (changes requested):

react/src/components/PrometheusCategorySelect.tsx
agatha197: Done in 86ed37147 — moved the `<Suspense>` boundary out of `PrometheusCategorySelect`. The component now just self-fetches via `useLazyLoadQuery` and suspends; the caller (`PrometheusQueryPresetEditorModal`) owns the boundary, wrapping the Category `Form.Item` so the fallback is scoped to that one field while the rest of the form stays interactive. The `Form.Item` sits inside the boundary so its injected `value`/`onChange` still reach the select.

react/src/components/PrometheusPresetTab.tsx
agatha197: Done in 86ed37147 — `PrometheusPresetTab` no longer imports `PrometheusQueryPresetEditorModalFragment$key`. `editingPreset` state and the `openEditor` param are now typed as `PrometheusQueryPresetNodeInList` (the node type flowing from `PrometheusPresetTabQuery`, which already carries the editor fragment ref through the nested `...PrometheusQueryPresetEditorModalFragment` spread). `PrometheusQueryPresetNodes.onEditPreset` was updated to hand back that same node type.

react/src/components/PrometheusPresetTab.tsx
agatha197: Done in 86ed37147 (same fix as the related thread) — dropped the `PrometheusQueryPresetEditorModalFragment$key` import and typed `editingPreset`/`openEditor` off `PrometheusQueryPresetNodeInList`, the node type from `PrometheusPresetTabQuery` that already includes the editor fragment ref.

react/src/components/PrometheusPresetTab.tsx
agatha197: This is intentional — we deliberately don't branch the refetch on update vs create. Both paths call `updateFetchKey()`, which triggers a single `network-only` refetch of the current page while preserving the user's active filter/sort/pagination. Keeping create identical to update avoids resetting filters or jumping pages on create, and the full refetch already reflects the new/edited row. Happy to revisit if you'd prefer create to e.g. jump to the first page.

ironAiken2 (approved):
LGTM

react/src/components/PrometheusQueryPresetEditorModal.tsx
```diff
@@ -359,17 +319,28 @@ const PrometheusQueryPresetEditorModal: React.FC<
...
-        <Form.Item
-          label={t('prometheusQueryPreset.Category')}
-          name="categoryId"
+        <Suspense
+          fallback={
+            <Form.Item label={t('prometheusQueryPreset.Category')}>
+              <BAISelect
+                loading
+                disabled
+                placeholder={t('prometheusQueryPreset.NoCategory')}
```
nowgnuesLee:
```suggestion
```

react/src/components/PrometheusQueryPresetEditorModal.tsx
```diff
@@ -359,17 +319,28 @@ const PrometheusQueryPresetEditorModal: React.FC<
...
-          />
-        </Form.Item>
+          <Form.Item
+            label={t('prometheusQueryPreset.Category')}
+            name="categoryId"
+          >
+            <PrometheusCategorySelect
+              allowClear
+              placeholder={t('prometheusQueryPreset.NoCategory')}
+              notFoundContent={t('prometheusQueryPreset.NoCategory')}
```
nowgnuesLee: Let's move these options to `PrometheuseCategorySelect`.

nowgnuesLee (changes requested):

react/src/components/PrometheusQueryPresetEditorModal.tsx
agatha197: Done — simplified the Suspense fallback to a plain `<BAISelect loading />`, dropping the redundant `disabled`/`placeholder` now that the field's defaults live in `PrometheusCategorySelect`.

react/src/components/PrometheusQueryPresetEditorModal.tsx
agatha197: Done — moved `allowClear`, `placeholder`, and `notFoundContent` into `PrometheusCategorySelect` as its defaults (still overridable via props, applied before `{...props}`). The call site is now just `<PrometheusCategorySelect />`.

nowgnuesLee (approved):
LGTM

PR #8302, opened 2026-07-15
feat(FR-3334): add common bulk-error modal and preserve form state on partial bulk failure

### LINKED ISSUE #8300: Add common bulk-error modal component and preserve form state on partial bulk failure
## Background

FR-3273 introduced multi-scope bulk permission edit (`RoleScopePermissionEditModal`). Bulk mutations can partially fail, and per-request errors need proper surfacing. This applies to bulk operations in general, so the error display should be a shared component.

### 1. Common bulk-error modal component

Create a reusable modal component that displays bulk request errors in a **table view** — one row per failed request.

- Each bulk operation has a different response shape, so the component must accept the table column definitions via props (e.g., a `columns`-like prop) instead of hardcoding them.

### 2. Preserve form state on partial failure (separate from the modal)

When some of the bulk requests fail, the existing form must not be reset:

- Keep the form items for the fields that had problems so the user can fix and retry them.
- Fields without problems remain displayed as-is.

JIRA Issue: FR-3334

ironAiken2:
Resolves #8300 (FR-3334)

## Summary

FR-3273's multi-scope bulk permission edit could partially fail and only surfaced errors as a comma-joined notification while unconditionally closing the modal, discarding the user's edits. This PR adds shared bulk-error components and makes the edit modal preserve its form state so failed requests can be fixed and retried.

### 1\. New shared component: `BAIBulkErrorModal` (`packages/backend.ai-ui`)

Displays the per-request errors of any bulk operation in a table — one row per failed request.

```ts
interface BAIBulkErrorModalProps<RecordType = AnyObject>
  extends Omit<BAIModalProps, 'children' | 'onOk' | 'onCancel' | 'footer' | 'type'> {
  columns: BAIColumnsType<RecordType>;   // caller-defined — every bulk op has its own response shape
  dataSource: RecordType[];              // one record per failed request; each carries a unique `key`
  alertDescription?: ReactNode;          // optional guidance; rendered in a BAIAlert above the table
  onRequestClose: () => void;            // dismissal callback (no footer; header X / mask / Esc)
}
```

- Props extend `BAIModalProps` per the component-props-extension rule; built on `BAIModal` + `BAITable` (row keys resolved from each record's `key` field, antd default).
- Client-side pagination: at most 10 rows per page (`hideOnSinglePage` — the pager appears only when there are more than 10 failures).
- Title defaults to a localized "Action execution failed" (KO: "실행 오류") prefixed with a `TriangleAlert` icon in `token.colorError` (icon only — title text keeps the default color); a caller-provided `title` is rendered verbatim instead.
- `alertDescription` (`ReactNode`, no default) is rendered as the body of a `BAIAlert` above the table under a fixed localized alert title "Error Occurred" (KO: "문제가 발생했습니다.").
- Uses `useBAIi18n`; new locale keys `comp:BAIBulkErrorModal.ActionExecutionFailed` / `ErrorOccurred` added to all 21 BUI locale files.
- Includes unit tests (custom-columns row rendering, default/overridden title, `onRequestClose`) and Storybook stories.

### 2\. New shared component: `BAICheckbox` (`packages/backend.ai-ui`)

antd `Checkbox` reads only `isFormItemInput` from `FormItemInputContext` — unlike Input-likes it never opts into the surrounding form item's validation status, so a field error leaves it visually unchanged. `BAICheckbox` reads `status` and `name` from the same context and:

- paints the error state itself (`colorError` border, `colorErrorHover` fill when checked).

Painting is all it does — it never mutates form state. Clearing the field error on change is the call site's decision (some flows must keep the mark), done from the caller's `onChange` via `form.setFields([{ name, errors: [] }])`.

Outside a `Form.Item` the context is empty and it behaves exactly like a plain antd `Checkbox`. Includes a Storybook story.

### 3\. Form-state preservation in `RoleScopePermissionEditModal`

- **On partial failure the modal no longer closes.** The grid/form keeps the user's edits so the failed cells can be retried as-is; an antd `message.error` toast gives the at-a-glance failure cue and `BAIBulkErrorModal` opens on top with one row per failed request; the alert inside it appends a secondary-text "(Success: n, Failed: m)" request count (new `rbac.PermissionsPartialFailureCounts` key in all 21 languages). The failure table renders **Target** as a `ScopeType|name` double tag and **Permission** as an `EntityType|operation` double tag (`BAIDoubleTag`, first tag blue / second default), plus the server's **Error Message**.
- **Failed cells are flagged in the grid.** Every grid cell is a named form field in both modes (single-scope: a `Form.Item` per cell registered at mount, pre-checked with the scope's granted state; bulk: `BAIBulkEditFormItem` registering on edit). Failed cell keys are written back as empty-string field errors (`form.setFields` with `errors: ['']`), which sets the item's error status — rendered by `BAICheckbox` — without printing a message under the cell. Toggling a failed cell clears its mark (the edit modal's own `onChange` — not `BAICheckbox` — clears the field error).
- **Successful requests are not re-submitted.** Successfully applied cells (including new permission ids from the add mutation) are layered onto the fragment-derived initial state via an `appliedCellOverrides` map, so the per-scope diff on the next save only produces work for the cells that actually failed. Single-scope and multi-scope bulk mode both reconcile through `applyBulkPermissionCells` (already-satisfied scopes fall out of the diff).
- **On full success** the behavior is unchanged (success message + close + parent refetch).
- **Cancel after a partial failure** now reports `success=true` to the parent so it refetches — some changes did reach the backend.
- New host i18n keys under `rbac` (`Permission`, `ErrorMessage`, `PermissionsPartialFailureDescription` — "Some permission changes failed. Please check the items.") added to all 21 language files.

### Note

- `react/src/__generated__/ScopedAuditLogQuery.graphql.ts` was regenerated: pre-existing drift on `main` (schema gained `AuditLogFilter.actedAs` / `UUIDFilter` without a committed `pnpm relay` run) that `scripts/verify.sh` flags. No query text changed in this PR.

## How to test

The test server **`10.82.0.151`** is configured so that bulk permission requests **always fail** — use it to reproduce the partial-failure UX:

1. Point the WebUI at `10.82.0.151`, log in, and open RBAC management → a role's scoped permission card.
2. Edit permissions for a single scope (or select multiple scopes → bulk edit) and save.
3. The requests fail → the edit modal stays open with your edits intact, the failed cells are outlined in red, and the bulk-error modal lists one row per failed request with the double-tag columns.
4. Toggle a failed cell → its red mark clears immediately. Save again → only the still-failing cells are re-submitted; cancel instead → the parent still refetches (partial changes reached the backend).

## Verification

```
bash scripts/verify.sh → === ALL PASS === (Relay / Lint / Format / TypeScript / Vite warmup)
vitest packages/backend.ai-ui BAIBulkErrorModal.test.tsx → 4 passed
```

The final amend (`BAICheckbox` extraction + message copy update) passed pre-commit lint/format; the full harness runs in CI.


ironAiken2:
* **#8363** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8363?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8360** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8360?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8302** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8302?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8302?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8302).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/RoleScopePermissionEditModal.tsx
```diff
@@ -509,61 +565,141 @@ const RoleScopePermissionEditModal: React.FC<
...
+      ) => {
+        let cells = appliedUpdates.get(appliedScopeId);
+        if (!cells) {
+          cells = new Map<string, string | null>();
+          appliedUpdates.set(appliedScopeId, cells);
+        }
+        cells.set(cellKey, permissionId);
+      };
+      const reasonMessage = (reason: unknown) =>
+        reason instanceof Error ? reason.message : String(reason);
```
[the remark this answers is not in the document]
ironAiken2: Fixed in the latest push — `reasonMessage` now unwraps `PayloadError[]` rejections from `useMutationWithPromise`, extracting each element's `message` (joined with newlines) before falling back to `String(reason)`.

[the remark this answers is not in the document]
ironAiken2: Already handled in the current revision — `reasonMessage` (now shared at `react/src/helper/mutationError.ts`) unwraps each `PayloadError`'s `message` and joins them; `String()` is only the last-resort fallback for non-array, non-Error reasons.

react/src/components/RoleScopePermissionEditModal.tsx
```diff
@@ -594,57 +718,60 @@ const RoleScopePermissionEditModal: React.FC<
           // switched into edit mode — the checkbox starts checked.
           initialValue={true}
         >
-          <Checkbox />
+          <BAICheckbox />
```
agatha197: It is not mandatory, but how about surfacing the failure message as a tooltip on the errored cell?

Image

react/src/components/RoleScopePermissionEditModal.tsx
```diff
@@ -594,57 +718,60 @@ const RoleScopePermissionEditModal: React.FC<
...
-        onChange={(event) => toggleCell(key, event.target.checked)}
-      />
+      <Form.Item
+        name={key}
+        valuePropName="checked"
+        // Pre-checked with the scope's currently-granted state.
+        initialValue={singleInitialKeys.has(key)}
+        style={{ marginBottom: 0 }}
+      >
+        <BAICheckbox />
```
agatha197: this one too.

packages/backend.ai-ui/src/components/BAIBulkErrorModal.tsx
```diff
@@ -0,0 +1,109 @@
...
+            {t('comp:BAIBulkErrorModal.ActionExecutionFailed')}
+          </BAIFlex>
+        )
+      }
+      okText={t('general.button.Close')}
+      onOk={() => onRequestClose()}
+      onCancel={() => onRequestClose()}
+      // Informational modal — a lone Close button; a Cancel twin would be
+      // meaningless here.
+      footer={(_originNode, { OkBtn }) => <OkBtn />}
```
agatha197: How about setting it null? It already has close button at the top.

agatha197 (changes requested):

packages/backend.ai-ui/src/components/BAICheckbox.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { FormItemInputContext } from 'antd/es/form/context';
+import React, { use } from 'react';
+
+const useStyles = createStyles(({ css, token }) => ({
+  errorCheckbox: css`
+    .ant-checkbox {
+      border-color: ${token.colorError} !important;
+    }
+    .ant-checkbox-checked {
+      background-color: ${token['red-4']} !important;
```
agatha197:
```suggestion
      background-color: ${token.colorErrorHover} !important;
```

agatha197: Considering theme changes, I think it would be better to maintain the colorError series.

agatha197 (dismissed):

react/src/components/RoleScopePermissionEditModal.tsx
ironAiken2: The error message may be too long and wordy, and it does not seem appropriate to display it as a tooltip. :)

packages/backend.ai-ui/src/components/BAIBulkErrorModal.tsx
ironAiken2: Applied in c02c19a — the footer is now `null` (okText/onOk dropped); dismissal goes through the header X / mask / Esc via `onCancel`. Updated the unit test, Storybook docs, and JSDoc accordingly.

packages/backend.ai-ui/src/components/BAICheckbox.tsx
ironAiken2: Applied in c02c19a — switched `token['red-4']` to `token.colorErrorHover` so the checked error fill follows the colorError series across theme changes.

react/src/components/RoleScopePermissionEditModal.tsx
```diff
@@ -347,54 +384,71 @@ const RoleScopePermissionEditModal: React.FC<
...
-  // edit mode hold the checkbox boolean.
-  const [bulkForm] = Form.useForm<Record<string, boolean | undefined>>();
+  // Single-scope edit: the one scope's currently-granted cells, used to
+  // pre-check the grid via each cell's Form.Item `initialValue`.
+  const singleInitialKeys = initialKeysOf(singleScopeId);
+  // Every grid cell is a boolean field on this form, keyed by its cell key.
+  // Single-scope: every rendered cell registers at mount, pre-checked with
+  // the scope's granted state. Bulk: only cells switched into edit mode
+  // register (`BAIBulkEditFormItem`); 'Keep as is' cells stay `undefined`.
+  const [form] = Form.useForm<Record<string, boolean | undefined>>();
```
nowgnuesLee: Use useRef for the form here unless useForm is specifically needed.

nowgnuesLee (changes requested):

react/src/components/RoleScopePermissionEditModal.tsx
ironAiken2: Applied in fb3a0d6b1 — replaced `Form.useForm` with `useRef<FormInstance<Record<string, boolean | undefined>>>` and `<Form ref={formRef}>`; all usages are post-mount imperative calls (`setFields`, `getFieldsValue`), so `useForm` wasn't specifically needed.

nowgnuesLee (approved):
LGTM

PR #8329, opened 2026-07-20
fix(FR-3340): render Chat recovery form immediately and show model fetch as button loading

### LINKED ISSUE #8321: Chat: model-list fetch shows a blank ~30s skeleton before the recovery form appears
## Background

The Playground Chat card loads available models by fetching the endpoint's OpenAI-compatible `/models` (`react/src/components/Chat/ChatCard.tsx`). FR-3212 bounded a previously-infinite hang with `AbortSignal.timeout(30000)` so an unresponsive endpoint eventually falls into the CustomModelForm recovery UX ("LLM models not found. Verify the base path and token compatibility with OpenAI, then press 'Refresh Model Information'").

## Problem

While the `/models` request is in flight, the model area sits behind a Suspense boundary that renders only a bare skeleton. When the endpoint is slow or unresponsive, the user stares at that skeleton for up to the full **30 seconds** before any actionable UI (the base-path / token form + "Refresh Model Information") appears. Nothing signals that the app is waiting on the endpoint, so it reads as "the page is just very slow / frozen" rather than "we're trying to reach your model server."

This is more visible now that the Chat endpoint selector can surface a deployment whose replica traffic status is ACTIVE but whose model server is not actually answering `/v1/models` (traffic-router readiness is not OpenAI-endpoint readiness) — e.g. a cold-starting or mock endpoint.

## Steps to reproduce

1. Open Playground -> Chat with an endpoint whose `/v1/models` does not respond (a mock/e2e endpoint, or one still cold-starting).
1. Observe the Chat card.

**Expected:** quick, contextual feedback that the model server is being contacted, with the manual base-path / token fallback reachable immediately.

**Actual:** a blank skeleton for up to 30s, then the "LLM models not found" recovery form.

## Root cause

- `ChatCard.tsx` - the model-list query is Suspense-driven and aborts after `AbortSignal.timeout(30000)`; the fallback during that window is a bare skeleton with no status text and no manual escape hatch.

## Proposed directions (open)

- Render the endpoint header + a "Connecting to the model server... (up to 30s)" state immediately instead of a blank skeleton, so context and the manual base-path / token form are reachable right away.
- Or make the wait two-stage: an optimistic short attempt (~5s) that surfaces "still connecting / set base path manually" while retrying in the background.
- Consider shortening or making the 30s bound configurable, balanced against genuine slow-but-healthy cold starts (the reason FR-3212 chose 30s).

## Notes

- Follow-up to FR-3212 (added the 30s timeout). Orthogonal to FR-3332: the endpoint filter decides which deployment is selectable but cannot guarantee the model server answers `/v1/models`, so the graceful-loading UX must live in the Chat card.

JIRA Issue: FR-3340

agatha197:
Resolves #8321 (FR-3340)

## Problem

The Playground Chat card loads available models by fetching the endpoint's OpenAI-compatible `/models` (`ChatCard.tsx`). Because that fetch used a **suspending** `useSuspenseTanQuery`, the whole `ChatCard` sat behind the parent's bare `<Card loading>` skeleton for up to the **30s** abort timeout added in FR-3212 — no header, no status, and no way to reach the manual base-path/token recovery form until the fetch settled. A slow or unresponsive endpoint therefore read as "the page is frozen" rather than "we're contacting your model server".

## Change

Make the `/models` fetch **non-suspending** so the card renders immediately, and express the wait as a **plain loading state** rather than a separate alert box:

- `useModels`: `useSuspenseTanQuery` → `useTanQuery`, exposing `isLoadingModels`. `models` is normalized to `[]` while loading so consumers keep a stable `ChatModel[]` shape. The 30s `AbortSignal.timeout`, error codes, `queryKey`, and `select` are unchanged.
- `PureChatCard` no longer suspends on `/models`, so the **card header (endpoint / agent / sync selectors) renders instantly**. `ChatInput` is disabled while loading and re-enables on success.
- The `CustomModelForm` recovery UI (base-path / token inputs + **Refresh Model Information**) is now **reachable immediately during the wait**. While the fetch is in flight, the wait is surfaced as an ordinary loading state — the **Refresh button spins** and the **base-path / token inputs are disabled** — rather than a warning-styled status box. The existing **"LLM models not found" warning is shown only once the fetch has settled** (`!isConnecting`), so a slow-but-healthy endpoint no longer momentarily reads as a failure. On a genuine failure the existing warning + recovery flow is unchanged.
- `EndpointTokenSelect` now forwards `disabled` to its plain-`<Input>` fallbacks (no endpoint id / no tokens), so the token field is actually disabled during the wait — not only the `<Select>` variant.

### Notes for reviewers

- Per review feedback, the earlier warning-styled **"Connecting to the model server…" alert** was **dropped** in favor of the button-loading + disabled-inputs treatment (no flashing/morphing status box). Its `chatui.ConnectingToModelServer` i18n key and the 500ms debounce were removed with it; **no new i18n keys are added**.
- The `<Suspense fallback={<Card loading/>}>` boundaries in `ChatPage.tsx` and `ModelCardChat.tsx` are intentionally **kept** — they still cover the sibling Relay `useLazyLoadQuery<ChatCardQuery>` (endpoint) suspension inside `PureChatCard`. They just no longer catch the `/models` fetch. Please don't remove them in a later cleanup.
- Scope is loading-UX only: the antd `<Card>` was **not** migrated to `BAICard` (heavily-customized full-height chat container) and the 30s timeout value was **not** changed.

## Verification

- `bash scripts/verify.sh`: **ALL PASS** (Relay ✅, Lint ✅, Format ✅, TypeScript ✅, Vite warmup ✅). The one terminology finding is a pre-existing warn-only item unrelated to this change.
- Acceptance: header renders immediately; the manual base-path/token form is reachable during the wait; the Refresh button shows loading and the inputs are disabled while `/models` is in flight; the "not found" warning only appears after the fetch settles; success and failure paths preserved.

### DIFF
--- a/react/src/components/Chat/CustomModelForm.tsx
+++ b/react/src/components/Chat/CustomModelForm.tsx
@@ -36,6 +36,7 @@ const CustomModelForm: React.FC<CustomModelFormProps> = ({
   hasNoDesiredReplicas,
   onSubmit,
 }) => {
+  'use memo';
   const { t } = useTranslation();
   const { token: themeToken } = theme.useToken();
   const formRef = useRef<FormInstance>(null);
@@ -71,26 +72,30 @@ const CustomModelForm: React.FC<CustomModelFormProps> = ({
         }}
       >
         {hasNoDesiredReplicas ? (
-          <div style={{ marginBottom: themeToken.size }}>
-            <Alert
-              type="warning"
-              showIcon
-              title={t('chatui.NoDesiredReplicas')}
-            />
-          </div>
+          <Alert
+            type="warning"
+            showIcon
+            title={t('chatui.NoDesiredReplicas')}
+            style={{ marginBottom: themeToken.size }}
+          ></Alert>
         ) : null}
-        <div style={{ marginBottom: themeToken.size }}>
-          <Alert type="warning" showIcon title={t('chatui.CannotFindModel')} />
-        </div>
+        <Alert
+          type="warning"
+          showIcon
+          title={t('chatui.CannotFindModel')}
+          style={{ marginBottom: themeToken.size }}
+        ></Alert>
         <Form.Item label={t('modelService.BasePath')} name="basePath">
           <Input
             placeholder="v1"
             prefix={shrinkControlSize ? undefined : endpointUrl}
+            disabled={loading}
           />
         </Form.Item>
         <Form.Item label={t('modelService.Token')} name="token">
           <EndpointTokenSelect
             loading={loading}
+            disabled={loading}
             endpointId={endpointId}
             style={{
               width: shrinkControlSize ? '100%' : '200px',
--- a/react/src/components/Chat/ChatCard.tsx
+++ b/react/src/components/Chat/ChatCard.tsx
@@ -3,7 +3,7 @@
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
 import { ChatCardQuery } from '../../__generated__/ChatCardQuery.graphql';
-import { useSuspenseTanQuery } from '../../hooks/reactQueryAlias';
+import { useTanQuery } from '../../hooks/reactQueryAlias';
 import { useAIAgent } from '../../hooks/useAIAgent';
 import PureChatHeader from './ChatHeader';
 import PureChatInput from './ChatInput';
@@ -122,7 +122,7 @@ function useModels(
     }
   };
 
-  const { data: modelsResult } = useSuspenseTanQuery<{
+  const { data: modelsResult, isLoading: isLoadingModels } = useTanQuery<{
     data: Array<ChatModel>;
     error?: number;
   }>({
@@ -183,13 +183,19 @@ function useModels(
       ? provider.modelId
       : (modelsResult?.data?.[0]?.id ?? 'custom');
 
-  const modelsError =
-    modelsResult.error && getModelsErrorMessage(modelsResult.error);
+  const modelsError = modelsResult?.error
+    ? getModelsErrorMessage(modelsResult.error)
+    : undefined;
 
   return {
-    models: modelsResult?.data,
+    // useTanQuery leaves `data` undefined until the query settles; normalize to
+    // an empty array so consumers (ChatHeader, the CustomModelForm gate) keep a
+    // stable `ChatModel[]` shape. `isLoadingModels` distinguishes the in-flight
+    // window from a genuinely empty result.
+    models: modelsResult?.data ?? [],
     modelId,
     modelsError,
+    isLoadingModels,
   } as const;
 }
 
@@ -276,7 +282,7 @@ const PureChatCard: React.FC<ChatCardProps> = ({
     chat.provider.basePath,
     agentEndpointUrl || endpoint?.url,
   );
-  const { models, modelId, modelsError } = useModels(
+  const { models, modelId, modelsError, isLoadingModels } = useModels(
     chat.provider,
     fetchKey,
     baseURL,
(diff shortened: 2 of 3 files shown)


agatha197:
* **#8329** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8329?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8329?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8329).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/src/components/Chat/CustomModelForm.tsx
```diff
@@ -23,6 +23,7 @@ type CustomModelFormProps = {
   token?: string;
   endpointId?: string | null;
   loading: boolean;
+  isConnecting?: boolean;
```
ironAiken2: Instead of introducing a separate `isConnecting` prop, this can be handled with the existing `loading` prop. Please reuse `loading` to drive the connecting state here rather than adding a new prop.

react/src/components/Chat/CustomModelForm.tsx
```diff
@@ -80,7 +83,24 @@ const CustomModelForm: React.FC<CustomModelFormProps> = ({
...
+            // While the endpoint's /models fetch is still in flight, surface a
+            // neutral "connecting" status instead of the "not found" warning so
+            // a slow-but-healthy endpoint does not read as a failure. The
+            // base-path / token inputs below stay reachable throughout the wait.
+            <Alert
+              type="info"
+              showIcon
+              icon={<LoadingOutlined />}
+              title={t('chatui.ConnectingToModelServer')}
+            />
```
ironAiken2: Showing an alert for a fetch that can finish quickly doesn't look great — on a fast network this will likely appear as a blink/flash. Also, the 'Refresh Model Information' button below already displays a loading indicator, so a separate connecting alert doesn't seem necessary here.

ironAiken2 (changes requested):
Please see the inline comments — reuse the existing `loading` prop instead of adding `isConnecting`, and reconsider the connecting alert since the refresh button already conveys the loading state.

react/src/components/Chat/CustomModelForm.tsx
agatha197: I kept `isConnecting` as a separate prop rather than folding it into `loading`, because the two track different states.

`loading` here is `isPendingUpdate` — the `useTransition` state around the user-triggered *Refresh Model Information* submit — so it is only `true` after the user clicks Refresh. The connecting status needs to reflect the initial `/models` discovery (`isLoadingModels`), which is already in flight on first render while `isPendingUpdate` is still `false`. If the connecting UI were driven by `loading`, it would never appear during that initial fetch — which is exactly the window this PR targets.

So the value has to come from the query's loading state, not the transition. Happy to rename the prop if the naming is the concern, but they can't share a source.

react/src/components/Chat/CustomModelForm.tsx
agatha197: Good call on the flash — addressed in the latest push:

- The connecting status is now debounced by 500ms (`useDebounce`), so a fast `/models` fetch never flashes it; it only appears when the wait actually exceeds ~500ms.
- It now uses the same `warning` type as the settled "not found" alert, so the box morphs in place (spinner + "Connecting…" → warning icon + "not found") instead of popping blue → yellow.

On the Refresh button already conveying loading: that spinner is `loading={isPendingUpdate}`, i.e. the `useTransition` state of a *user-initiated* Refresh submit. During the initial `/models` discovery the user hasn't clicked anything, so `isPendingUpdate` is `false` and the button shows no spinner. Without an inline status there, the form would render the "LLM models not found" warning while the fetch is still in flight — the false negative this PR is fixing. So the status covers a window the button doesn't.

ironAiken2 (approved):
LGTM

PR #8357, opened 2026-07-23
feat(FR-3214): exclude pnpm store from the dev type-check watcher to cut per-process watch FDs

nowgnuesLee:
Related to #8047 (FR-3214) — the issue itself is resolved by #8156 (process-group kill); this PR is the complementary footprint reduction, split out per review discussion.

> **Stacked on** #8156

## Problem

Even with orphan leaks fixed in #8156, each *live* dev server still holds far more file descriptors than it needs. Measured on a live dev stack: a single `vite` dev process holds **~5,894 open file descriptors**, of which **~4,481 are regular files inside the pnpm store** (immutable dependency `.d.ts` type declarations, ~3,840 of them).

The holder is **vite-plugin-checker**'s TypeScript watch program (`ts.createWatchProgram`), which installs a per-file watcher on every file in the program — including every dependency `.d.ts` in the store. On macOS each per-file watch costs 1 open FD, so:

- GUI/IDE-launched dev inherits launchd's soft `maxfiles` = **256** → instant EMFILE.
- Terminal-launched dev survives one repo but running N repos multiplies ~5,900 × N toward the kernel limit.

## Fix

Tell the checker's TS watch program **not to watch the pnpm store** via `watchOptions.excludeDirectories` / `excludeFiles`.

Two constraints shape the implementation:

1. **The exclude spec must be an absolute store path.** TS anchors `**/`-globs under the project `basePath` (`combinePaths()`), so they can never match the store realpath which lives *outside* the repo (`~/Library/pnpm/store/v11/...`). Measured: `**/node_modules/**`, `**/store/**` etc. all had **zero** effect. Since the absolute path is machine-specific, it cannot live in the committed `tsconfig.json`.
2. **vite-plugin-checker (0.9.3) has no inline watchOptions API** — it calls `ts.createWatchCompilerHost` without a `watchOptionsToExtend` arg, so the only injection point is the tsconfig file it reads.

Therefore `react/vite.config.ts` now generates a **gitignored `react/tsconfig.checker.json`** at dev startup — `{ extends: './tsconfig.json', watchOptions: { excludeDirectories/excludeFiles: ['<absolute store root>/**'] } }` — and points `checker({ typescript: { tsconfigPath } })` at it. The store root reuses the existing `pnpmStorePath` (from `pnpm store path`).

## Why excluding the store from the watch is safe

- **Type-checking stays complete.** `watchOptions` excludes only drop the *file watcher*, not the *read*. TS still reads every dependency `.d.ts` to build the program — verified: error count identical before/after (real-stack boot: "Found 0 errors" both ways; in an error-injection harness, 296 → 296).
- **CI / pre-commit / `scripts/verify.sh` are untouched.** Those run one-shot `tsc --noEmit` (no `--watch`), which ignores `watchOptions` entirely — and this PR does not modify the committed `react/tsconfig.json` at all. The exclude lives only in the gitignored derived file that the dev-mode checker reads. Every gate that actually enforces type safety is unaffected.
- **`vite build` is untouched.** The derived-config helper is only *called* when `command === 'serve'`; a build is handed the base tsconfig directly and performs no file I/O at all (a one-shot build installs no file watchers anyway).
- **HMR and workspace packages keep full watch.** HMR is Vite's own separate watcher, unrelated to the checker's watchOptions. `backend.ai-ui` / `backend.ai-client` resolve via tsconfig `paths` to `../packages/*/src` — project-internal source outside the store — so they keep normal per-file watch and instant re-check.
- **Graceful fallback.** If the store path can't be resolved (npm/yarn flat installs) or the derived file can't be written, the checker falls back to the plain `tsconfig.json` — original behavior, no exclude.

The only theoretical gap: a dependency `.d.ts` changing **while dev is running, with zero source edits** won't auto-trigger a re-check. In practice `pnpm install/add/update` triggers Vite's dep re-optimization / dev restart (which re-reads the store), and using a new package requires a source edit — which is watched and re-triggers the checker. The store is content-addressed and immutable during normal dev, and CI re-checks everything regardless.

## Measurements (real stack, `lsof` on the live vite pid)

| config | vite open REG files | files held open under pnpm store | type errors | live edit detection |
|---|---|---|---|---|
| before (no watchOptions) | **5,884** | 3,825 | baseline | YES (~2s) |
| after (this PR) | **1,404** (−76%) | **2** | **identical** | **YES (~2s)** |

Also verified this is **not a Vite regression** by actually booting the last pure-craco commit (`00e8431a1`) in a worktree and measuring the same `tsc --watch` program: craco-era = **6,870 FDs** (6,055 in-store) vs current Vite = 5,894 — the Vite stack was already slightly *lighter*; the store-watch cost is inherent to running an in-process TS type-checker, and this PR removes it.

## Changes since the first review

Four follow-up commits, all in `resolveCheckerTsconfigPath` / its call site:

1. **Stale derived config is now removed on the degrade path** (review request). If an earlier run wrote `tsconfig.checker.json` and a later run resolves no store root, the file is unlinked instead of left behind with a stale exclude spec.
2. **The side effects are scoped to `serve`.** The helper is evaluated while the `plugins` array is built, so it previously ran for *every* command — meaning the new cleanup made `vite build` delete `tsconfig.checker.json` out from under a dev server running in another terminal. The call site now guards on `command === 'serve'` and hands `build` the base tsconfig directly.
3. **The config is published atomically** (raised by the Graphite bot). `writeFileSync` opens with `O_TRUNC`, leaving the file zero-length for ~4.9 us p50 / 11.2 us p99; under contention 54,589 of 81,795 concurrent reads saw an empty file. It now writes a pid-unique temp and `rename(2)`s it into place — same stress test after: 0 empty reads. Content was never corrupted (all writers produce byte-identical output in a single `write(2)`); the failure mode was the truncate window, not interleaving.
4. **Cleanup helper extracted** (`removeQuietly`) and two over-long comment blocks trimmed.

**On `excludeFiles` vs `excludeDirectories`** (review question): they are not redundant, and `excludeFiles` is the load-bearing one. TypeScript dispatches the two lists by watcher kind — `watchFile()` is tested only against `excludeFiles`, `watchDirectory()` only against `excludeDirectories` — so a directory-subtree glob cannot suppress the per-file watchers this PR targets. Measured on `react/tsconfig.json`, same program, only `watchOptions` varying:

| variant | total FDs | store `.d.ts` |
|---|---|---|
| no excludes | 6,028 | 3,903 |
| `excludeDirectories` only | 6,028 | 3,903 |
| `excludeFiles` only | 1,485 | **0** |
| both (what ships) | 1,485 | **0** |

`excludeDirectories` contributes nothing measurable here; it is kept only because this is one machine / one TS version / macOS and it costs nothing to leave.

## Verification

- **Re-measured on this stacked branch** (same method both runs: boot `pnpm run dev`, wait for the checker's "Found 0 errors", then `lsof` the vite pid): parent branch without the exclude = **6,056** open REG files (**4,566** under the pnpm store) vs this branch = **87** open REG files (**2** under the store). `tsconfig.checker.json` generated with the correct absolute store glob and confirmed gitignored; checker reports "Found 0 errors" in both runs.
- `bash scripts/verify.sh` — `=== ALL PASS ===` (Relay / Lint / Format / TypeScript), re-run on this stacked branch.
- Real-stack boot test (from the original round of this change): `tsconfig.checker.json` generated correctly, confirmed gitignored, checker reports "Found 0 errors", FD table above.
- Injected a TS2322 into `react/src/index.tsx` via atomic save while dev was running → checker reported it within ~2s (source watch intact); reverting cleared it.

### DIFF
--- a/react/vite.config.ts
+++ b/react/vite.config.ts
@@ -1,7 +1,15 @@
 import react from '@vitejs/plugin-react';
 import compression from 'compression';
 import { execSync } from 'node:child_process';
-import { createReadStream, existsSync, readFileSync, statSync } from 'node:fs';
+import {
+  createReadStream,
+  existsSync,
+  readFileSync,
+  renameSync,
+  rmSync,
+  statSync,
+  writeFileSync,
+} from 'node:fs';
 import { createRequire } from 'node:module';
 import {
   basename,
@@ -13,7 +21,7 @@ import {
   resolve,
 } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { defineConfig, loadEnv, type Plugin } from 'vite';
+import { defineConfig, loadEnv, normalizePath, type Plugin } from 'vite';
 import checker from 'vite-plugin-checker';
 import { nodePolyfills } from 'vite-plugin-node-polyfills';
 import { VitePWA } from 'vite-plugin-pwa';
@@ -24,6 +32,18 @@ import { devReviewOverlayPlugin } from './vite-plugins/reviewOverlay';
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const projectRoot = resolve(__dirname, '..');
 const require = createRequire(import.meta.url);
+
+const baseTsconfigPath = resolve(__dirname, 'tsconfig.json');
+const checkerTsconfigPath = resolve(__dirname, 'tsconfig.checker.json');
+
+// Deletes a path we are abandoning anyway; never throws.
+function removeQuietly(path: string): void {
+  try {
+    rmSync(path, { force: true });
+  } catch {
+    // Cleanup is best-effort.
+  }
+}
 
 // With `enableGlobalVirtualStore: true` (pnpm-workspace.yaml), pnpm hard-links
 // every dependency into a single global store outside this repo, with the
@@ -68,6 +88,53 @@ function resolvePnpmStorePath(): string | undefined {
       err instanceof Error ? err.message : err,
     );
     return undefined;
+  }
+}
+
+// Excludes the pnpm store from the checker's `tsc` watch program (FR-3214).
+// The worker opens one FD per file in the program, including thousands of
+// immutable store `.d.ts`, so concurrent dev servers hit EMFILE. `watchOptions`
+// excludes drop the watcher but not the read, so type-checking stays complete.
+// vite-plugin-checker exposes no inline watchOptions, leaving the tsconfig it
+// reads as the only injection point, and the exclude must be absolute: TS
+// anchors `**/`-globs under the project basePath, which can never match a store
+// realpath outside the repo. That path is machine-specific, hence a gitignored
+// derived config rather than an entry in the committed tsconfig.
+function resolveCheckerTsconfigPath(storeRoot: string | undefined): string {
+  if (!storeRoot) {
+    // Drop a checker config left by an earlier run; the base tsconfig wins here.
+    removeQuietly(checkerTsconfigPath);
+    return baseTsconfigPath;
+  }
+  // TS watchOptions globs use forward slashes, even on Windows.
+  const storeGlob = `${normalizePath(storeRoot)}/**`;
+  // Publish atomically: `writeFileSync` truncates first, so a concurrent dev
+  // server can read an empty file, whereas `rename(2)` is never observed midway.
+  const tmpPath = `${checkerTsconfigPath}.${process.pid}.tmp`;
+  try {
+    writeFileSync(
+      tmpPath,
+      JSON.stringify(
+        {
+          extends: './tsconfig.json',
+          watchOptions: {
+            excludeDirectories: [storeGlob],
+            excludeFiles: [storeGlob],
+          },
+        },
+        null,
+        2,
+      ) + '\n',
+    );
+    renameSync(tmpPath, checkerTsconfigPath);
+    return checkerTsconfigPath;
+  } catch (err) {
+    removeQuietly(tmpPath);
+    console.warn(
+      '[vite] could not write tsconfig.checker.json; the type-check worker will watch pnpm-store .d.ts files and may hit EMFILE under multi-repo dev:',
+      err instanceof Error ? err.message : err,
+    );
+    return baseTsconfigPath;
   }
 }
 
@@ -921,12 +988,20 @@ export default defineConfig(({ command, mode }) => {
       // terminal and as a browser overlay during `vite dev`. Vite itself
       // only strips types via esbuild, so without this plugin type
       // errors are visible only in the IDE or via `scripts/verify.sh`.
-      // `tsconfigPath` pins the checker to the consumer's tsconfig
-      // (which has the `paths` workaround for pnpm-global-virtual-store
-      // type resolution — see FR-2925).
+      // `tsconfigPath` pins the checker to the consumer's tsconfig (which has
+      // the `paths` workaround for pnpm-global-virtual-store type resolution —
+      // see FR-2925). During dev it is the derived `tsconfig.checker.json`
+      // that excludes the pnpm store from the watch program (FR-3214) — see
+      // `resolveCheckerTsconfigPath` above; for `vite build` it is the plain
+      // base tsconfig.
       checker({
         typescript: {
-          tsconfigPath: resolve(__dirname, 'tsconfig.json'),
+          // Only `serve` owns tsconfig.checker.json; `vite build` reads the base
+          // config and must not write or delete a file a live dev server uses.
+          tsconfigPath:
+            command === 'serve'
+              ? resolveCheckerTsconfigPath(pnpmStorePath)
+              : baseTsconfigPath,
         },
       }),
 
--- a/.gitignore
+++ b/.gitignore
@@ -167,6 +167,11 @@ data/merged_schema.graphql
 # webui release download temp directories (safe to delete)
 scripts/temp-releases/
 
+# derived type-check tsconfig generated at dev startup (FR-3214): excludes the
+# machine-specific pnpm store path from vite-plugin-checker's TS watch program
+react/tsconfig.checker.json
+react/tsconfig.checker.json.*.tmp
+
 # ignore playwright mcp test files
 .playwright-mcp/
 


nowgnuesLee:
* **#8407** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8407?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8395** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8395?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8394** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8394?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8357** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8156** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8156?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8357).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

react/vite.config.ts
```diff
@@ -71,6 +77,64 @@ function resolvePnpmStorePath(): string | undefined {
...
+// (`~/Library/pnpm/store/...`). Since that path is machine-specific it can't
+// live in the committed tsconfig — we derive it at dev startup and write a
+// gitignored `tsconfig.checker.json` that extends the base config.
+//
+// Returns the path the checker should read: the derived config when a store
+// root is known, otherwise the base `tsconfig.json` (npm/yarn flat installs, or
+// store-resolution / write failure → no-op, original behavior preserved).
+function resolveCheckerTsconfigPath(storeRoot: string | undefined): string {
+  const baseTsconfig = resolve(__dirname, 'tsconfig.json');
+  if (!storeRoot) {
```
agatha197, line 104: **Requested change — clean up a stale derived config on the degrade path.** If an earlier `serve` wrote `tsconfig.checker.json` and a later run resolves no store root (pnpm store lookup fails, or a flat npm/yarn install), we return the base tsconfig but leave the previously-generated file on disk. It's gitignored and unreferenced so it's harmless, but the checker silently keeps reading a stale exclude config across runs. Suggest unlinking it here before returning:
```ts
if (!storeRoot) {
  try { rmSync(resolve(__dirname, 'tsconfig.checker.json'), { force: true }); } catch {}
  return baseTsconfig;
}
```

react/vite.config.ts
```diff
@@ -71,6 +77,64 @@ function resolvePnpmStorePath(): string | undefined {
...
+    return baseTsconfig;
+  }
+  const checkerTsconfig = resolve(__dirname, 'tsconfig.checker.json');
+  // TS watchOptions globs use forward slashes, even on Windows.
+  const storeGlob = `${normalizePath(storeRoot)}/**`;
+  try {
+    writeFileSync(
```
agatha197, line 115: **Nit (optional) — side-effecting file write during config evaluation.** `writeFileSync` runs as a side effect of evaluating the `defineConfig` factory (the `plugins` array), so it fires on every config load/reload rather than as part of the dev-server lifecycle. It works, but writing a file while resolving config is a little surprising. Consider moving the generation into a small plugin `configResolved`/`buildStart` hook so the write is tied to an explicit lifecycle step. Non-blocking.

react/vite.config.ts
```diff
@@ -71,6 +77,64 @@ function resolvePnpmStorePath(): string | undefined {
...
+      checkerTsconfig,
+      JSON.stringify(
+        {
+          extends: './tsconfig.json',
+          watchOptions: {
+            excludeDirectories: [storeGlob],
+            excludeFiles: [storeGlob],
```
agatha197, line 122: **Nit (optional).** `excludeFiles` is matched against files, whereas `<storeRoot>/**` is a directory-subtree glob already covered by `excludeDirectories` above — so this line is redundant belt-and-suspenders. It's harmless (and your measurements confirm the exclude works), just noting it in case you'd rather keep `excludeFiles` for an actual file pattern or drop it.

agatha197 (changes requested):
Mechanism looks correct and I verified the injection locally (generated an equivalent `tsconfig.checker.json`; `tsc --showConfig` carries `watchOptions` with the absolute store glob, inherits the base `paths` workaround, and the program file set is unchanged — so type-checking stays complete and only the watchers are dropped). The FD reduction itself is macOS/kqueue-specific and I can't reproduce it on Linux, so that rests on your live `lsof` measurements, which look thorough.

Requesting one change (the stale-file cleanup) plus two optional nits inline. None are blocking correctness — happy to see it merge once the cleanup is addressed or consciously waived.

yomybaby (approved):
LGTM

react/vite.config.ts
nowgnuesLee: Done — addressed in `8aa3bac`.

`resolveCheckerTsconfigPath` now unlinks the derived config before falling back, so a file written by an earlier run can't outlive the store root that justified it:

```ts
const checkerTsconfig = resolve(__dirname, 'tsconfig.checker.json');
if (!storeRoot) {
  // Drop a checker config left by an earlier run; the base tsconfig wins here.
  try {
    rmSync(checkerTsconfig, { force: true });
  } catch {
    // Cleanup is best-effort.
  }
  return baseTsconfig;
}
```

I hoisted the `checkerTsconfig` declaration above the guard so both paths share it. `force: true` keeps the common case (nothing to delete) silent, and the `try`/`catch` keeps a cleanup failure from taking down the dev server — the fallback still returns the base tsconfig either way.

`scripts/verify.sh` passes (Relay / Lint / Format / TypeScript / terminology).

On the two nits: leaving both as-is in this PR rather than expanding its scope. The `configResolved`/`buildStart` move is a fair point and I'd rather do it as a follow-up where the write ordering against the checker plugin's own init can be verified on a live dev run; ~~the redundant `excludeFiles` I'm keeping as belt-and-suspenders~~. Happy to take either if you feel strongly.

---

**Correction (edited):** the struck sentence above is wrong — I accepted the premise of nit 3 without checking it. `excludeDirectories` and `excludeFiles` are dispatched separately by watcher kind in TS, so `excludeFiles` is not redundant; it is the line that actually drops the per-file store watchers this PR is about. Details and source references in [the nit 3 thread](https://github.com/lablup/backend.ai-webui/pull/8357#discussion_r3655056944).

Thanks for the thorough verification pass — especially confirming the program file set was unchanged.

react/vite.config.ts
nowgnuesLee: Follow-up on this one — your framing turned out to be load-bearing, not just stylistic.

Because `resolveCheckerTsconfigPath()` is evaluated while the `plugins` array is constructed, it runs for *every* command, not just `serve`. `pnpmStorePath` is `undefined` unless `command === 'serve'`, so the cleanup I just added took the `!storeRoot` branch during `vite build` and deleted `tsconfig.checker.json` — including out from under a dev server running in another terminal, whose `tsc` watch program was started on that exact file.

Fixed in `ec756bd` by scoping the side effects to dev at the call site:

```ts
tsconfigPath:
  command === 'serve'
    ? resolveCheckerTsconfigPath(pnpmStorePath)
    : baseTsconfigPath,
```

So `build` now reads the base config and performs no file I/O at all. I also hoisted both tsconfig paths to module scope to drop the duplicated `resolve(__dirname, ...)`.

I'm still leaving the `configResolved`/`buildStart` move itself for a follow-up: `vite-plugin-checker` takes `tsconfigPath` as a constructor argument and spawns its `tsc` worker from its own hook, so writing the file in a later hook introduces a cross-plugin ordering dependency I'd want to confirm on a live dev run rather than reason about. Happy to file it as a separate issue if you'd prefer it tracked.

Leaving this thread open for you.

nowgnuesLee: Closing this one out — after the `ec756bd` fix I don't think the hook move buys anything, and I'd rather not pay its cost.

The concrete hazard in "config evaluation is not the dev lifecycle" was that the function ran for commands that have nothing to do with the type-check watcher. That's now gone: the call is guarded on `command === 'serve'`, so `build` never reaches it and performs no file I/O at all. What's left is a write that happens only when a dev server is being configured — which is exactly the lifecycle step it belongs to. Moving it into `configResolved`/`buildStart` would relocate the same write to a slightly later point in the same command, for the same single dev process.

The residual is that Vite re-evaluates the config when `vite.config.ts` or a `.env*` file changes and restarts the server. That rewrite is idempotent — same store root, same JSON — and it lands before the checker's worker restarts with it, so it's a no-op in practice rather than something a hook would protect against.

Against that, the move has a real cost: `vite-plugin-checker` takes `tsconfigPath` as a constructor argument and spawns its `tsc` worker from its own hook, so writing the file in a later hook creates a cross-plugin ordering dependency — the file has to exist before that worker starts, and neither plugin declares that relationship. Trading a guaranteed-ordered write for one that depends on hook ordering between two plugins is a downgrade, and verifying it would mean exercising every path that restarts the worker.

So: keeping the write where it is. Thanks for pushing on it — the framing is what surfaced the `build`-deletes-the-file problem, which was the part that actually mattered.

Also trimmed the oversized header comment above the function in `6b272b3b` while in here (25 lines to 9).

react/vite.config.ts
nowgnuesLee: I went to check this against the installed TypeScript (5.5.4) before dropping the line, and I think the premise doesn't hold — `excludeDirectories` can't cover the file watchers, so these two aren't belt-and-suspenders.

`createWatchFactory` builds both watch entry points through the same helper, but dispatches the exclude list on the *watcher kind* (`lib/typescript.js:123192`):

```js
return {
  watchFile:      createExcludeHandlingAddWatch("watchFile"),
  watchDirectory: createExcludeHandlingAddWatch("watchDirectory")
};
function createExcludeHandlingAddWatch(key) {
  return (file, cb, flags, options, ...) =>
    !matchesExclude(
      file,
      key === "watchFile" ? options?.excludeFiles : options?.excludeDirectories,
      ...
    ) ? factory[key].call(...) : excludeWatcherFactory(...);
}
```

So `watchFile()` is tested **only** against `excludeFiles`, and `watchDirectory()` **only** against `excludeDirectories`. A directory-subtree glob in `excludeDirectories` suppresses directory watchers whose own path matches; it never reaches the per-file watchers for files inside that subtree.

That matters here because the FD cost this PR targets *is* the per-file layer — the ~3,800 store `.d.ts` each holding one kqueue FD are all `watchFile()` calls. `excludeFiles` is the line that drops them. `excludeDirectories` is doing real but separate work: the directory watchers TS installs for failed-lookup locations and `node_modules` resolution.

Two different watcher layers, one line each. If I dropped `excludeFiles` the per-file watchers would come back and the FD reduction would go with them.

Two caveats on how far I took this:

- This is from reading the 5.5.4 source, not from re-running `lsof` with the line removed — I haven't measured the negative case.
- `isIgnoredByWatchOptions` (`:8060`) *does* union both lists, which is probably where the "they overlap" intuition comes from. That one sits on the directory-watcher callback path, filtering events for files under an already-watched directory, so it doesn't affect which watchers get installed.

Correcting myself, since it's on this PR and actionable in the wrong direction: in [my earlier reply](https://github.com/lablup/backend.ai-webui/pull/8357#discussion_r3654958311) I accepted your framing and called `excludeFiles` redundant belt-and-suspenders. That was wrong, and someone acting on it would silently undo FR-3214. I've annotated that comment to point here.

Leaving this thread open rather than resolving it — I'm contradicting your read, so it should be yours to close. Happy to run the `lsof` negative case if you want the measurement before deciding.

react/vite.config.ts
```diff
@@ -71,6 +81,52 @@ function resolvePnpmStorePath(): string | undefined {
...
+          },
+        },
+        null,
+        2,
+      ) + '\n',
+    );
+    return checkerTsconfigPath;
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in `5c641cc8`, though the mechanism is not quite the one described — I reproduced it before changing anything, and the measurements move two of the three claims.

**Setup.** Same 229-byte payload the config actually writes, 4 concurrent writers in a tight loop, 1 reader, 3s.

```
writeFileSync (before):  reads=81,795  zero=54,589  shortRead=0  parseErr=54,589
temp + rename (after):   reads=58,833  zero=0       shortRead=0  parseErr=0
```

And the window itself, timed around `openSync(path,'w')` → `writeSync` over 20,000 iterations:

```
empty window (truncate -> write done):  p50 4.9us   p99 11.2us   max 354us
```

**On "file corruption if writes interleave":** not supported. `resolvePnpmStorePath()` shells out to `pnpm store path`, so every writer on a machine produces byte-identical content, and 229 bytes is a single `write(2)`. Across 81,795 reads under contention, **zero** were partial — readers saw 229 bytes or 0 bytes, never a mixed file. Interleaved bytes aren't the failure mode.

**On "reading a partially-written JSON file":** right conclusion, wrong mechanism. Not partial — *empty*. `writeFileSync` defaults to flag `'w'` (`O_WRONLY|O_CREAT|O_TRUNC`), so the file is zero-length between open and write. That's the ~5us window above, and it's what all 54,589 parse failures were.

**On TOCTOU:** correct, and it's also why I didn't take the suggested patch. Read-existing-then-skip is itself a check-then-act with a gap, and it doesn't narrow the window at all when the content differs. (It also compares only `excludeDirectories[0]` and ignores `excludeFiles`, which would make it skip a rewrite that was actually needed.)

**What I did instead** — write to a pid-unique temp, then `rename(2)`, which is atomic, so a reader sees either the complete old file or the complete new one and the empty state stops existing:

```ts
const tmpPath = `${checkerTsconfigPath}.${process.pid}.tmp`;
writeFileSync(tmpPath, /* ... */);
renameSync(tmpPath, checkerTsconfigPath);
```

Cost is one extra syscall once per dev-server start (~74us vs ~36us). Temp is unlinked on the failure path and covered by `.gitignore`.

**Scope, for whoever reads this later:** this never affected the case FR-3214 is about. Separate checkouts each write their own `react/tsconfig.checker.json`, so there's no shared path to contend on. It required two `vite serve` processes in the *same* checkout, and the blast radius was a dev server's type-checking failing until restart — no effect on build output. Worth fixing because it's three lines, not because it was likely.

react/vite.config.ts
nowgnuesLee: Measured it. You were right that one of the two lines is redundant — it's just the other one, and I need to correct something I wrote above.

**Method.** `tsc -p <variant> --noEmit --watch` against `react/tsconfig.json`, same program every run, only `watchOptions` varying. Waited for `Watching for file changes`, +3s, then `lsof -p <pid>`.

| variant | total FDs | store `.d.ts` |
|---|---:|---:|
| A — no excludes | 6,028 | 3,903 |
| B — `excludeDirectories` only | 6,028 | 3,903 |
| C — both (what ships) | 1,485 | **0** |
| D — `excludeFiles` only | 1,485 | **0** |

And a live `pnpm dev` on this branch, for the process that actually consumes the derived config:

```
vite (hosts the checker worker)   total=1,524   store=2   store .d.ts=0
tsc --watch (from scripts/dev.mjs) total=512    store=481 store .d.ts=428
```

Baseline A (6,028 / 3,903) lines up with the ~5,900 / ~3,800 in the PR description, so the harness is measuring the right thing.

**Confirms the main point:** `excludeFiles` is load-bearing. D alone produces the entire reduction; dropping it would have put all 3,903 watchers back.

**Correction to my previous comment:** I claimed `excludeDirectories` was "doing real but separate work — the directory watchers TS installs for failed-lookup locations and `node_modules` resolution." The measurement doesn't support that. B is identical to A down to the individual FD, and D is identical to C. In this configuration `excludeDirectories` contributes exactly nothing. I asserted that from reading the dispatch logic rather than measuring it, and I shouldn't have.

So your read that the pair is redundant was correct — the redundant line is `excludeDirectories`, not `excludeFiles`.

**On what to do about it:** I'd keep both, and I don't feel strongly. The case for keeping is only that this is one machine, one TS version, macOS/kqueue — I haven't checked whether `excludeDirectories` earns its place under inotify or a different `watchDirectory` strategy, and it costs nothing to leave. The case for dropping is that it demonstrably does nothing here and the file is clearer with one line than two. Your call; say the word and I'll drop it.

Leaving the thread open for you either way.

agatha197 (approved):
LGTM

PR #8384, opened 2026-07-25
docs(FR-3372): add relay-mutation-store-updates skill for refetch-after-mutation audit

nowgnuesLee:
Relates to FR-3372 (sub-task of FR-3170).

> The GitHub clone of FR-3372 had not appeared when this PR was opened, so there is no `Resolves #N` line yet — it can be added once the Jira webhook lands.

## Problem

Setting modals take a nullable `*Frgmt` prop, branch `create` vs `update` inside, and report both outcomes to the caller with the same `onRequestClose(true)`. The caller then does:

```tsx
onRequestClose={(success) => {
  if (success) updateFetchKey();   // full list requery
  setOpen(false);
}}
```

A pure field edit therefore requeries the entire list, which is precisely the behaviour FR-3170 was filed against: with this in place, Relay's normalized cache buys us nothing.

The tempting fix — pass `false` after a successful update so the caller skips the refetch — is wrong, and `UserSettingModal` already does it. The prop is declared `(success: boolean)` and `false` already means *"cancelled"*, so this makes cancel and success indistinguishable to the caller; anything the caller later wants to do on success stops firing after updates.

The create/update distinction never needed to travel through the callback at all. **Callers already hold it**: `AdminUserManagement` renders separate edit and create instances, and every single-instance caller has `editingX` in the handler's closure. So the decision belongs at the call site, and `success` stays honest.

The convention was not accidental. The `react-modal-drawer` skill documented `if (result) updateFetchKey(); // success path` as *the* close-callback pattern, so both humans and agents have been reproducing it from the docs.

## Analysis

Verified against the current tree — measurements, not estimates.

**Relay's write-side APIs are essentially unused.** Across 109 `useMutation` call sites: 1 `updater:`, 1 `@appendEdge`, 0 `optimisticResponse`, 0 `ConnectionHandler`, 0 `commitLocalUpdate`. Relay is being driven as a fetch client.

**The backend already returns what we need.** In `data/schema.graphql`, 41 of 50 update-class payloads (`Modify*` / `Update*`) return the entity. Only 9 return bare `ok`/`msg`, and 5 of those already have node-returning successors (`ModifyResourcePreset` → `UpdateResourcePresetPayload`, `ModifyScalingGroup` → `UpdateResourceGroupPayload`, and the three resource-policy pairs). The gap is on the client: we are not asking for the node.

**Modal-level counts.** Of 23 update mutations inside modal components, 13 select a node payload and 10 select only `ok`/`msg` — the latter being exactly the legacy `Modify*` set above. No modal distinguishes create from update in its close callback; every one passes the same `true`. The sole exception, `ContainerRegistryEditorModal`, already threads `onOk('create' | 'modify')` and *still* refetches on both.

**The tightest cluster** is the `*List` + `*SettingModal` family — `ResourceGroupList`, `ResourcePresetList`, `UserResourcePolicyList`, `ProjectResourcePolicyList`, `KeypairResourcePolicyList`, `AdminUserCredentialList`, `PrometheusPresetTab` — seven call sites with an identical shape.

### Two concrete bugs surfaced

1. ~~**`AutoScalingRuleEditorModalLegacy.tsx:136`** selects `rule { metric_name, … }` **without `id`**~~ — **withdrawn on review.** The Relay compiler adds `id` to the network operation automatically on node-identifiable types (visible in the generated `*Mutation.graphql.ts` operation text), so the payload already merges. Explicit `id` in the source is a readability convention, not a fix; the spec marks item B rejected.

2. **`UserSettingModal.tsx`** *sends* `groupIds` (`:448`) and its fragment *reads* `projects { edges { node { id } } }` (`:213`), but the payload never selects `projects` — and the update path deliberately skips the refetch via `onRequestClose(false)` (`:468`). Changing a user's project membership succeeds server-side and leaves the UI stale. This is the partial-coverage trap: no refetch *and* no store update.

## Change

Analysis-only issue, so this PR ships the guardrail and the remediation plan rather than the refactor.

- **New `relay-mutation-store-updates` skill.** The decision table — including the exception Copilot review sharpened: an update to a field the list query **filters or sorts on** still needs the refetch, because a store patch cannot evict or reorder an already-fetched connection edge — the partial-coverage warning, the legacy→modern mutation table, the `updater:` recipe for `ok`/`msg`-only payloads (modelled on `AgentSettingModal`, the one place doing this today), the call-site pattern for keeping `success` honest, and the cases where a refetch genuinely *is* correct — with the requirement that it carry a comment saying why. Names `react/eslint.config.js`'s `no-restricted-syntax` block as the landing spot for the eventual lint rule, so the skill reads as the interim fix it is.
- **New `.specs/FR-3372-refetch-after-mutation/spec.md`** — the remediation spec, enumerating what is wrong by category (A–F) with the affected files, plus requirements, the fix approach, and a five-step work breakdown.
- **`react-modal-drawer`** — corrected in place and pointed at the new skill: `onRequestClose(success)` reports mutation success truthfully, and the caller decides the refetch. (`react-async-actions` and `react-suspense-fetching` were also corrected here originally, but FR-3489 #8655 pruned those skills from the repo, so their hunks left this diff in rebase.)
- **`relay-patterns`** — had no mutation guidance at all. Added a short section plus a review-checklist line.
- **`AGENTS.md`** — registered the skill under On-Demand Skills.

No production code changes; no behaviour change.

## Follow-up

Detailed in the spec; remediation belongs in separate PRs under FR-3170:

1. The `UserSettingModal` payload bug above — small, independently shippable. (The `AutoScalingRule` `id` bug was withdrawn; see above.)
2. Remove the `onRequestClose(false)` workaround in `UserSettingModal`, moving the decision to `AdminUserManagement`.
3. The 13 modals whose payloads already return a node — drop the update-path refetch at the call site, verifying field coverage per the skill's diff procedure **and** that no changed field appears in the list query's `filter`/`orderBy` first.
4. The 10 legacy `ok`/`msg` modals — migrate to the node-returning successor (5 exist) or add an `updater:` (3).
5. A lint rule for `if (success) updateFetchKey()`, per FR-3170's "structural guarantee" requirement.

## Verification

- `prettier --check` now passes on every markdown file this PR touches — the review-fix commit ran `--write` on all four, so the formatting churn previously deferred for `react-modal-drawer` / `relay-patterns` landed there.
- No code changed in this PR; `bash scripts/verify.sh` → `=== ALL PASS ===` was still run on the stacked branch above (which includes this diff).
- Every claim above was checked against the tree: schema payload shapes and modal selection sets were counted by script, and the two bugs were confirmed by reading the files directly.

https://claude.ai/code/session_01JWaAm6sfEVj5TVd5oQgvte


nowgnuesLee:
* **#8401** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8401?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8384** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8384).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

agatha197 (changes requested):
please resolve conflicts and check copilot's reviews

.claude/skills/relay-mutation-store-updates/SKILL.md
```diff
@@ -0,0 +1,252 @@
...
+## 1. The decision table
+
+Ask **what changed**, not **did it succeed**.
+
+| What the mutation changed                                            | What to do                                                                            |
+| -------------------------------------------------------------------- | ------------------------------------------------------------------------------------- |
+| **Fields of an entity already in the store** (update)                | Select the changed fields on the returned node. Relay merges by `id`. **No refetch.** |
+| Same, but the payload returns only `ok`/`msg` (legacy)               | Keep the refetch and comment why (§4). Nothing to merge.                              |
+| **List membership** — a row added or removed (create/delete)         | Refetch the list (§6). Don't patch the connection.                                    |
+| Server-derived fields you did not send (computed status, timestamps) | Select those too; refetch only if the server cannot return them (§7).                 |
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in fa7690c99. The decision table gained a dedicated row for updates whose changed field is in the list's `filter`/`orderBy` arguments (refetch — the patch cannot evict or reorder the edge), §7 now lists it as the first legitimate update-refetch case with the canonical status-toggle examples, and the checklist requires diffing changed fields against the list query's filter/orderBy before dropping any refetch. The headline rule and the same passages in `relay-patterns` were qualified accordingly. This matches what the stacked implementation PR #8401 verified per call site.

.claude/skills/react-modal-drawer/SKILL.md
```diff
@@ -90,16 +90,26 @@ interface FolderCreateModalProps extends BAIModalProps {
...
   }}
 />
 ```
 
+**The argument means "the list changed", not "it succeeded".** Refetching on
+every success throws away Relay's normalized cache: an *update* mutation that
+returns its changed fields has already patched the store, so `updateFetchKey()`
+there is a redundant round-trip. This matters most in modals that handle both
+create and update behind one nullable fragment prop — those two paths must not
+share a single `onRequestClose(true)` exit.
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in fa7690c99. The paragraph now states the contract the spec defines: the argument means "the mutation succeeded", passed truthfully, and the caller decides whether to refetch from the create/update context it already holds — keeping the refetch when the edited field is a filter/sort key. The "list changed" framing is gone.

.specs/FR-3372-refetch-after-mutation/spec.md
```diff
@@ -0,0 +1,213 @@
...
+| `RBACManagementPage.tsx:181`              | 소프트 비활성인데 payload가 `status` 미포함                |
+| `FairShareList.tsx:680`                   | `afterUpdate` 콜백 경유, 필드 전용 수정                    |
+| `UserSettingModal.tsx:902`                | REST TOTP 제거 후 무관한 access-key 목록 재조회            |
+
+#### E-3. 가장 싼 수정 — payload가 **이미** 변경 필드를 반환하는데도 refetch
+
+둘 다 파일을 직접 열어 확인했습니다. mutation selection을 손댈 필요 없이 refetch 호출만 지우면 됩니다.
+
+- **`RBACManagementPage.tsx:202`** — `adminUpdateRole`이 `{ id, status }`를 반환(`:145`)해 Relay가 이미 패치하는데 `updateFetchKey()`를 호출
+- **`ReservoirArtifactDetailPage.tsx:329`** — `cancelImportArtifact`가 `artifactRevision { id status }`를 반환하는데 페이지 전체 재조회
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in fa7690c99. E-3 is rewritten: both refetches are load-bearing because the changed field is the list's own server-side filter (`RBACManagementPage` filters `status: { in: [queryParams.status] }` and the activate action only shows under the DELETED tab; `ReservoirArtifactDetailPage`'s alerts come from the `status: PULLING` connection). Work item 1 is marked "nothing to remove", and the E-2 rows for the same call sites were corrected too. The stacked PR #8401 reached the same conclusion during implementation and left all three refetches untouched.

.claude/skills/react-suspense-fetching/SKILL.md
```diff
@@ -94,10 +94,13 @@ import { useFetchKey, INITIAL_FETCH_KEY } from 'backend.ai-ui';
 
 const [fetchKey, updateFetchKey] = useFetchKey();
 
-// in a mutation:
+// in a mutation that changes list membership (create / delete):
 onCompleted: () => { updateFetchKey(); message.success(...); }
 ```
 
+For an **update** mutation, don't bump the fetch key — return the changed
+fields so Relay patches the store instead. See `relay-mutation-store-updates`.
```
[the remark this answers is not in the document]
nowgnuesLee: `react-suspense-fetching` was pruned from the repo by FR-3489 (#8655), so the flagged text left this diff in rebase. The same correction — a store patch cannot evict or reorder an edge when the changed field is in the active filter/sort — was applied in fa7690c99 to the skills that remain here (`relay-mutation-store-updates`, `relay-patterns`, `react-modal-drawer`).

AGENTS.md
```diff
@@ -99,6 +99,8 @@ read `package.json` / `pnpm-workspace.yaml` / `ls` rather than expecting a list
 - **Documentation**: `docs-writing-guide` skill (fw plugin; user manual structure, terminology, multilingual rules)
 - **Astryx UI fixes**: `astryx-fix` skill (assignee gate before starting, measure-before-you-fix, theme-defaults-first procedure, known traps, verification bar)
 - **Astryx UI bug reporting**: `astryx-bug-report` skill (capture-only intake for visual / behavioral defects and `discussion` items — "is this intended?" / "propose X instead" — filed under epic FR-3491 as Bugs and Tasks respectively, duplicate + relates scan). Use it when the ask is "record this", `astryx-fix` when it is "fix this".
+- **Relay**: `relay-patterns` skill (fragment architecture, naming conventions, query optimization)
```
agatha197, line 102: `relay-patterns` does not exist — #8655 deleted it, and it is already absent at this branch's merge-base. Registering it here points agents at a skill that cannot load.

.claude/skills/relay-mutation-store-updates/SKILL.md
```diff
@@ -0,0 +1,252 @@
...
+- [ ] Prefer spreading the consumer's fragment over hand-listing fields
+- [ ] Schema checked (`data/schema.graphql`) before concluding a refetch is required
+- [ ] Legacy `ok`/`msg`-only mutation → refetch kept, with a comment saying why
+- [ ] `success` is passed truthfully; the refetch decision lives at the call site
+- [ ] Create/delete refetch the list — no `@appendEdge`/`@deleteRecord` added
+- [ ] A refetch after an **update** has a comment naming why the store can't be patched
+
+## Related
+
+- `react-async-actions` — §6 refetch hierarchy; this skill is its detailed form
```
agatha197, line 248: All four entries in this list (`react-async-actions`, `relay-patterns`, `react-modal-drawer`, `react-suspense-fetching`) were deleted by #8655 and are absent at this merge-base. §1 also grounds this skill's rationale in `react-async-actions` §6, which no longer exists.

.claude/skills/relay-mutation-store-updates/SKILL.md
```diff
@@ -0,0 +1,252 @@
...
+- Choosing the selection set of a `mutation` payload
+- "The list doesn't refresh after saving" / "why is it fetching twice?"
+
+## 1. The decision table
+
+Ask **what changed**, not **did it succeed**.
+
+| What the mutation changed                                            | What to do                                                                            |
+| -------------------------------------------------------------------- | ------------------------------------------------------------------------------------- |
+| **Fields of an entity already in the store** (update)                | Select the changed fields on the returned node. Relay merges by `id`. **No refetch.** |
```
agatha197, line 37: The decision table has no row for "the changed field is in the list query's filter/orderBy → keep the refetch"; the words filter/orderBy appear nowhere in the 252 lines, and §7's exception list omits it too. #8401 reports that this exact check invalidated 28 of 33 candidate call sites, so it is the rule this skill most needs to state.

.claude/skills/relay-mutation-store-updates/SKILL.md
```diff
@@ -0,0 +1,252 @@
...
+When the caller renders **separate instances** for create and edit — as
+`AdminUserManagement` does with `userSettingFrgmt={selectedUser}` and
+`={null}` — the branch collapses: the edit instance simply never refetches.
+
+```tsx
+<UserSettingModal
+  userSettingFrgmt={selectedUserForSettingModal}
+  onRequestClose={() => {
+    setSelectedUserForSettingModal(null);
+    // no refetch: the update mutation returns the node and Relay patches it
```
agatha197, line 188: This worked example prescribes dropping the refetch in `AdminUserManagement`, which #8401 concluded is a bug: the list filters on `status` unconditionally and this modal edits `status`, so the row must be evicted rather than patched. #8401 keeps the refetch — following this example would reintroduce the stale ACTIVE-tab row.

.claude/skills/relay-mutation-store-updates/SKILL.md
```diff
@@ -0,0 +1,252 @@
...
+    id
+    ...YourListRow_user # ← the fragment your list row already renders
+  }
+}
+```
+
+If the consumer has no reusable fragment, hand-list every field the UI reads —
+including `id`. **`id` is mandatory**: Relay keys normalized records by node id,
+so a payload without it is written to a detached record and merged nowhere.
+`AutoScalingRuleEditorModalLegacy.tsx` gets this wrong — it returns all eight
```
agatha197, line 74: This is the `AutoScalingRuleEditorModalLegacy` finding the PR description says was withdrawn (the compiler auto-adds `id` on node-identifiable types), but the text still asserts it as a bug, and `spec.md` §B still lists it un-struck.

agatha197 (changes requested):
Checked every cross-reference and factual claim in the new skill against the tree at this branch's merge-base (ffad514d). The analysis behind this holds up; the findings below are all in the artifact that ships, which is the part agents will actually load.

AGENTS.md
nowgnuesLee: Fixed in #8401 (f15a90cd2), where this PR's commits now live after the stack was folded — the dead `relay-patterns` bullet is removed from AGENTS.md; only `relay-mutation-store-updates` is registered.

.claude/skills/relay-mutation-store-updates/SKILL.md
nowgnuesLee: Fixed in #8401 (f15a90cd2) — the Related list now points only at targets that exist (the FR-3372 spec and the epic), and the §1 grounding in `react-async-actions` §6 is dropped.

.claude/skills/relay-mutation-store-updates/SKILL.md
nowgnuesLee: Addressed in #8401 — f15a90cd2 added the filter/orderBy row to the decision table, §7 and the checklist; ad51feb81 then inverted that rule (patch in place, manual refresh re-evaluates), and the table/§7/checklist state the inversion with its date.

.claude/skills/relay-mutation-store-updates/SKILL.md
nowgnuesLee: Fixed in #8401 (f15a90cd2, regrounded in c6bee0dcc) — the `AdminUserManagement` example no longer prescribes dropping that refetch: it now explains the edit path keeps it because the status-toggle payload returns only the id, so there is nothing to patch.

.claude/skills/relay-mutation-store-updates/SKILL.md
nowgnuesLee: Fixed in #8401 (f15a90cd2) — the withdrawn missing-`id` finding is corrected in the skill (§2 states the compiler auto-selects `id` on node-identifiable types) and struck in spec.md §B and the D-1 row.

nowgnuesLee:
Folded into #8401 — the stack was collapsed, so the relay-mutation-store-updates skill, the FR-3372 spec and the AGENTS.md registration now ship there (with the review fixes applied: f15a90cd2, ad51feb81, c6bee0dcc). All feedback on this PR is answered above and carried into #8401.

PR #8415, opened 2026-07-28
refactor(FR-3256): align deployment sort to the shared convention

agatha197:
> ♻️ **Replaces #8336** — auto-closed by GitHub after an accidental force-push to a rewritten branch; code identical, prior discussion on #8336.

## Summary

Aligns the deployment table's sort with the shared convention every other table uses. `BAIModelDeploymentNodes` sort keys move from the server-enum shape (`NAME`, `CREATED_AT`, …) to camelCase, so callers bridge to the `DeploymentOrderField` enum via the shared `convertToOrderBy` / `convertFirstOrderByToString` helpers. The deployment-specific `parseDeploymentOrder` (and the parent's local `deploymentOrderToString`) are removed; `DeploymentListPage` / `ProjectAdminDeploymentsPage` / `AdminDeployment` migrate to the shared helpers.

### ⚠️ Breaking change (intentional)
- Removes `parseDeploymentOrder` from the published `backend.ai-ui` entry point; external consumers switch to `convertToOrderBy`.
- Internal sort keys change from the server-enum shape to camelCase. Existing `?order=` URL bookmarks stay compatible (`convertToOrderBy` normalizes via `snakeCase().toUpperCase()`).

## Test plan
- [x] `pnpm relay` compiles cleanly
- [x] `tsc --noEmit` passes (react + backend.ai-ui)
- [x] ESLint clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8414, opened 2026-07-28
refactor(FR-3256): hoist AdminDeploymentPage tabs to preloaded queries

agatha197:
> ♻️ **Replaces #8258** — that PR was auto-closed by GitHub when its branch was accidentally force-pushed to a rewritten (orphan) history and can no longer be reopened. The code here is byte-identical; prior review discussion lives on #8258.

## Summary

Base of the FR-3256 deployment-admin stack. Builds `AdminDeploymentPage` with four tabs — **deployments, model store, prometheus presets, deployment presets** — each loading via a hoisted `preloadedQuery` owned by the page, so switching tabs preserves each tab's already-fetched rows (render-as-you-fetch; only a full page reload resets them). Includes the tab-naming refactor (`…Nodes` → `…Table` for BAITable-owning components, `AdminModelCard`) and the supporting client / UI / i18n changes.

## Test plan
- [x] `pnpm relay` compiles cleanly
- [x] `tsc --noEmit` passes (react + backend.ai-ui)
- [x] ESLint clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #8416, opened 2026-07-28
feat(FR-3256): add RuntimeVariantPreset admin CRUD tab

agatha197:
> ♻️ **Replaces #8148** — auto-closed by GitHub after an accidental force-push to a rewritten branch; code identical, prior discussion on #8148.

## Summary

Adds the **runtime-variant-presets** tab to the deployment-admin page:
- `AdminRuntimeVariantPreset` — the tab view, wired into `AdminDeploymentPage` as a hoisted `preloadedQuery` tab.
- `BAIRuntimeVariantPresetTable` — the fragment table (owns `<BAITable>`, named per the FR-3256 table/role convention; renamed from `…Nodes`).
- `BAIRuntimeVariantPresetSettingModal` — create/edit modal.

## Test plan
- [x] ESLint clean


agatha197:
Stacked on #8415 (FR-3256).

agatha197:
* **#8416** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8416?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8416?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8415** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8415?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#8414** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8414?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8416).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

ironAiken2 (approved):
LGTM

react/src/components/AdminRuntimeVariantPreset.tsx
```diff
@@ -0,0 +1,319 @@
...
+      <BAIUnmountAfterClose>
+        <BAIRuntimeVariantPresetSettingModal
+          open={isCreating || !!editingPreset}
+          presetFrgmt={editingPreset}
+          onRequestClose={(success) => {
+            setIsCreating(false);
+            setEditingPreset(null);
+            if (success) {
+              onReload(queryRef.variables, { fetchPolicy: 'network-only' });
+            }
```
nowgnuesLee:
```suggestion
            if (success && editingPreset === null) {
              onReload(queryRef.variables, { fetchPolicy: 'network-only' });
            }
```

react/src/hooks/useRuntimeParameterSchema.ts
```diff
@@ -138,14 +138,14 @@ export function useRuntimeParameterSchema(
...
-              # 26.4.4rc manager (rc < final in PEP440), even though the field
-              # already exists there from rc9 onward.
-              required @since(version: "26.4.4rc9")
+              # Version-gated per the schema: required was added in 26.4.4
+              # (see data/schema.graphql), so it is stripped from the request
+              # on older managers to keep the presets query valid on legacy
+              # backends. We follow the schema version, not the capability
+              # flag's rc pin (runtime-variant-preset-required in client.ts,
+              # at 26.4.4rc9); FR-3056 will simplify that flag to the final
+              # tag. Pre-final 26.4.4rc managers are intentionally not covered.
```
nowgnuesLee: are these comments really necessary?

nowgnuesLee (changes requested):

react/src/hooks/useRuntimeParameterSchema.ts
agatha197: Removed — only `required @since(version: "26.4.4")` remains now. 26.4.4rc managers aren't a consideration here, so there was nothing left in that comment worth keeping.

react/src/components/AdminRuntimeVariantPreset.tsx
agatha197: You're right that Relay normalizes the update mutation's payload into the store — the mutation returns the full preset keyed by `id`, so the edited row's fields do re-render without a refetch.

The reason the refetch is still there is that this list is resolved entirely server-side. `AdminRuntimeVariantPresetQuery` passes `filter`, `orderBy`, `limit`/`offset` and reads `count`, and it isn't a `@connection` — so Relay updates the record's *fields*, but not the list's membership, order, or `count`.

`UpdateRuntimeVariantPresetInput` can change `name` and `rank`. Both are sortable columns, and `name` is also one of the two filterable properties. So skipping the refetch on edit would leave the table stale in ordinary cases:

- Sorted by name: renaming `alpha` → `zulu` updates the cell but keeps the row sitting above `beta`.
- Name filter active: an edited row that no longer matches stays visible, and `count` stays wrong.
- Either of the above can also mean the row now belongs on a different page.

A conditional refetch (only when `name`/`rank` actually changed) would be more precise, but the page only holds `id` + fragment refs for `editingPreset`, so it would require widening `onRequestClose` to report which fields changed. Given the refetch is a single page of ≤10 rows, and the delete / filter / sort / paging paths already reload with `network-only`, I've kept it as-is for consistency.

Leaving this thread open in case you'd still prefer the narrower version — happy to switch.

nowgnuesLee:
```suggestion
            if (success && editingPreset === null) {
              onReload(queryRef.variables, { fetchPolicy: 'network-only' });
            }
```

nowgnuesLee (changes requested):

react/src/components/AdminRuntimeVariantPreset.tsx
agatha197: Applied in 4f6f6d9f5 — you're right, and my earlier reply missed the relevant context.

I'd argued the refetch was needed because `name` and `rank` are sortable (and `name` filterable), so an edit can move the row. That's true mechanically, but it's the behavior the team already agreed *not* to have: the parent branch's `64d81a43f` applied exactly this gate to `AdminDeployment`, `AdminModelCard` and `AdminPrometheusPreset` (and `UserResourcePolicyV2` already followed it), on the rationale that an edited row should stay in place so the user can see the edit took effect, and only get reordered/filtered out on the next explicit refresh — mirroring session Terminating → Terminated visibility. This tab was the last holdout, so it now matches.

One precondition worth noting, since `64d81a43f` had to widen `DeploymentSettingModalUpdateMutation` for it: no widening was needed here. `BAIRuntimeVariantPresetSettingModalUpdateMutation` already selects every field `BAIRuntimeVariantPresetTableFragment` renders, so Relay merges the edited record by id and the row re-renders without a refetch. The only field the mutation omits is `runtimeVariant { name }`, which is keyed off `runtimeVariantId` — not something `UpdateRuntimeVariantPresetInput` can change.

`bash scripts/verify.sh` → `=== ALL PASS ===`.

nowgnuesLee (approved):
LGTM

PR #8421, opened 2026-07-28
fix(FR-3386): keep dashboard visible during interval refetch and isolate card suspensions

### LINKED ISSUE #8417: Dashboard page stuck in Suspense fallback after page refresh
## Symptom

When the `/dashboard` page is refreshed (full page reload), the entire page falls into a Suspense fallback state and **never escapes it** — the loading state persists indefinitely instead of rendering the dashboard content.

## Expected behavior

After a refresh, the dashboard should show its Suspense fallback briefly while queries load, then render normally.

## Notes for investigation

- The whole page suspends at once, which suggests a page-level Suspense boundary is involved rather than per-card boundaries.
- Check the dashboard's query orchestration (`useLazyLoadQuery` / fetchKey / `useDeferredValue` usage) for a condition where a suspended query never resolves or re-suspends in a loop on initial mount.

JIRA Issue: FR-3386

ironAiken2:
Resolves #8417 (FR-3386)

## Problem

Refreshing `/dashboard` rendered the board briefly, then the **entire page collapsed into the route-level Suspense `<Skeleton/>` at the first 15-second interval refetch and never escaped**. Reproduced deterministically on `main`: board appears ~20s after reload, stays until ~30s, disappears at ~35s (first interval tick) and never returns (observed 55s+).

Two causes:

1. **Interval refetch suspended the page query.** `DashboardPage` fed the raw `fetchKey` into `useLazyLoadQuery` with `fetchPolicy: 'network-only'`. Every 15s tick created a brand-new operation with no store data, so the whole page fell back to the route-level Suspense. The rest of the codebase already uses the `useDeferredValue(fetchKey)` idiom for exactly this reason (e.g. `AdminComputeSessionListPage`, `DeploymentListPage`, `StorageProxyList`) — the dashboard was the only interval-refetching page without it.
2. **Two cards could hold the whole page hostage.** `MyResource` and `MyResourceWithinResourceGroup` suspend on their own REST query (`useResourceLimitAndRemaining` → `POST /resource/check-presets`, the heaviest manager call on this page) but were wrapped only in `BAIBoardItemErrorBoundary` — which catches errors, not suspension — unlike every other card which has its own `<Suspense>`. On a cold load (refresh), one slow `check-presets` kept the entire page in the route-level skeleton.

## Changes

- Add `const deferredFetchKey = useDeferredValue(fetchKey)` and use it for the page query's `fetchKey`/`fetchPolicy` and all card `fetchKey` props, so interval refetches happen in the background while the committed board stays on screen. `isRefetching` now also covers the `fetchKey !== deferredFetchKey` window.
- Wrap `MyResource` and `MyResourceWithinResourceGroup` in their own `<Suspense fallback={<Skeleton/>}>` (same pattern as the other board cards), so a slow `check-presets` skeletons only those two cards instead of the whole page.

## Verification

Behavioral verification on a live dev server (Playwright, superadmin, backend 26.8.0-alpha):

| Scenario | main | this PR |
|---|---|---|
| Refresh, healthy backend, observed 75s (3+ interval ticks) | board disappears at 35s, never returns | board visible from 20s through 75s, no fallback |
| Refresh with `POST /resource/check-presets` held forever | whole page stuck in skeleton 105s+, no recovery even after unblocking | board renders at 15s; only the two resource cards show skeletons; recovers fully once the request completes |

Repro recipe for reviewers: DevTools → Network → block `check-presets` → refresh `/dashboard`.

### DIFF
--- a/react/src/pages/DashboardPage.tsx
+++ b/react/src/pages/DashboardPage.tsx
@@ -32,7 +32,7 @@ import {
   useInterval,
 } from 'backend.ai-ui';
 import * as _ from 'lodash-es';
-import { Suspense, useTransition } from 'react';
+import { Suspense, useDeferredValue, useTransition } from 'react';
 import { useTranslation } from 'react-i18next';
 import { graphql, useLazyLoadQuery } from 'react-relay';
 
@@ -48,8 +48,15 @@ const DashboardPage: React.FC = () => {
   const buildProjectPath = useProjectPath();
 
   const [fetchKey, updateFetchKey] = useFetchKey();
+  // Deferring the fetchKey keeps the committed board on screen while the
+  // interval-driven `network-only` refetch suspends in the background.
+  // Feeding the raw fetchKey into useLazyLoadQuery made every 15s tick
+  // re-suspend the whole page into the route-level fallback (FR-3386).
+  const deferredFetchKey = useDeferredValue(fetchKey);
   const [isPendingIntervalRefetch, startIntervalRefetchTransition] =
     useTransition();
+  const isRefetching =
+    isPendingIntervalRefetch || fetchKey !== deferredFetchKey;
 
   const [localStorageBoardItems, setLocalStorageBoardItems] =
     useBAISettingUserState('dashboard_board_items');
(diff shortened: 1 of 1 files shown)


ironAiken2:
* **#8421** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8421?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8421?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8421).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

yomybaby (approved):
LGTM

PR #8432, opened 2026-07-29
test: e2e auto-heal 2026-07-28

lablup-octodog:
## Summary

Automated e2e auto-heal produced by the **daily backend.ai-webui digest** pipeline (run 2026-07-28).

Two spec files were triaged as **HEAL** (test-side drift, no app regression) and fixed. Every healed test was re-run and verified passing before this PR was opened.

## Healed tests (verified re-pass)

- `e2e/auto-scaling-rule-preset/preset-crud.spec.ts`
  - _Superadmin cannot save an edited preset without a Name_
  - _Superadmin cannot save an edited preset without a Metric Name_
  - **Cause:** #8285 fixed the Prometheus preset edit modal to correctly re-apply `initialValues` on reopen. The tests had relied on the previous (buggy) empty-form behavior to trigger required-field validation. Fix: explicitly clear the Name / Metric Name field before Save.

- `e2e/chat/chat-sync.spec.ts`
  - _Sending a message from one synced pane triggers send in all other synced panes_
  - _User can re-enable sync after disabling it and text propagation resumes_ (was already green; re-verified)
  - **Cause:** #8329 made ChatCard's model fetch non-suspending with an `isLoadingModels` flag that disables `ChatInput` during load, exposing a send-after-endpoint-switch race. Fix: wait for the pane's own `/v1/models` response + model header before sending.

## Exhausted / unhealed

None — all four in-scope tests pass reliably on re-run.

## Notes

- Produced automatically; no Jira issue required for auto-heal PRs.
- Report doc: `/home/ubuntu/e2e-digest-reports/report-2026-07-28.md`
- cc reviewers: drift-causing PR author(s) requested below.

🤖 Generated with the daily digest auto-heal pipeline.

---

## Follow-up fix: `chat-sync.spec.ts` flakiness (added after review)

Verification of this PR surfaced a problem the original auto-heal missed. Both
describes in `chat-sync.spec.ts` run in `serial` mode, so the failure of
`Turning off sync isolates input to the individual pane` — a test this PR did not
touch — **skipped the two healed tests entirely**. A file-level run never
executed them, so the "verified passing" claim above only held when the tests
were run individually.

That test was genuinely flaky (2/5 passes in isolation), and the cause is a real
test bug rather than an app regression:

- The tests clicked a pane's sync toggle and then immediately typed, using
  `expect(input).toHaveValue('')` as the gate.
- That gate is a **no-op whenever the input is already empty**, so it returned
  instantly without proving the toggle had landed.
- `ChatHeader` wraps `onChangeSync` in `startTransition`, so the pane keeps its
  previous `sync` value for a beat after the click. Typing inside that window
  still mirrors the text into the shared `synchronizedMessage` atom, and the
  text propagates to the other pane — exactly the failure observed.

Fix: a `setSync` helper that clicks the toggle and waits for the `SyncSwitch`
icon to report the new state, used everywhere sync is toggled. The fixed 500ms
`waitForTimeout` in the multi-endpoint send test was also dropped, since the
deterministic waits around it already cover that settle.

**Result:** `chat-sync.spec.ts` now passes **7/7 with `--retries=0`**, and the
previously-flaky test passes 5/5 on repeat (was 2/5).

### Known remaining issue (out of scope, pre-existing on `main`)

`preset-crud.spec.ts` has flaky tests that are unrelated to this PR and are not
addressed here — `Page Load > pagination controls`, `Edit > update the Metric
Name`, `Edit > update the Time Window`. They pass when re-run and fail
intermittently under parallel load, which points at test isolation between
presets rather than an app defect. Worth a separate heal pass.

### DIFF
--- a/e2e/auto-scaling-rule-preset/preset-crud.spec.ts
+++ b/e2e/auto-scaling-rule-preset/preset-crud.spec.ts
@@ -805,9 +805,17 @@ test.describe(
       const modal = page.getByRole('dialog');
       await expect(modal).toBeVisible();
 
-      // The edit modal form does not auto-populate from initialValues on re-open
-      // (persistent Form instance). The Name field is already empty — just save
-      // immediately to trigger the required-field validation error.
+      // FR-3383/PR #8285 fixed the edit modal so it correctly re-applies
+      // initialValues on reopen — the Name field is now pre-populated with
+      // the preset's current name. Explicitly clear it to exercise the
+      // required-field validation.
+      const nameInput = modal.getByRole('textbox', {
+        name: 'Name',
+        exact: true,
+      });
+      await expect(nameInput).toHaveValue(presetName);
+      await nameInput.fill('');
+
       // Click "Save"
       await modal.getByRole('button', { name: 'Save' }).click();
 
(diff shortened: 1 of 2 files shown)


agatha197:
* **#8432** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8432?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8432?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8432).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

agatha197 (approved):
LGTM

PR #8420, opened 2026-07-28
refactor(FR-3332): migrate Chat deployment lookups off the legacy endpoint queries

agatha197:
> **Stacked on** #8298 (FR-3332). Merge that first.

## What

Follow-up to the `myDeployments` migration in #8298. The Chat deployment-selection surface still read the legacy Graphene `endpoint` queries and carried `endpoint` terminology in its JS/TS identifiers. This PR moves the remaining single-deployment lookups onto the Strawberry `deployment(id:)` field and renames the identifiers to match. **No user-visible behavior change.**

### Queries migrated to the Deployments API

| Document | Before | After |
| --- | --- | --- |
| `ChatCardQuery` | `endpoint(endpoint_id: UUID!) { endpoint_id url replicas }` | `deployment(id: ID!) { id networkAccess { endpointUrl } replicaState { desiredReplicaCount } }` |
| `DeploymentSelectValueQuery` | `endpoint(endpoint_id: UUID!) { name endpoint_id url }` | `deployment(id: ID!) { id metadata { name } networkAccess { endpointUrl } }` |
| `ChatHeader_Endpoint` fragment | `on Endpoint { endpoint_id name }` | `ChatHeader_Deployment on ModelDeployment { id metadata { name } }` |

`deployment(id:)` takes the global Relay ID while the chat provider state, the `deploymentId` URL param, and the select's value are all local UUIDs, so both call sites convert with `toGlobalId('ModelDeployment', …)` on the way in and `toLocalId` on the way out — the same pattern `DeploymentTokenSelect` already uses. The empty-string variable still pairs with the `store-only` fetch policy to skip the request when no deployment is selected.

With this, the whole Chat deployment surface (`myDeployments` list, single-deployment lookup, access tokens) sits on the Strawberry Deployments API; no Graphene `endpoint*` query remains in `react/src/components/Chat/`.

### Renamed (app-level identifiers)
- `onChangeEndpoint` → `onChangeDeployment`, `endpointFrgmt` → `deploymentFrgmt`, `selectedEndpoint` → `selectedDeployment`
- `defaultEndpointId` → `defaultDeploymentId`, `useDefaultEndpointId` → `useDefaultDeploymentId`, `endpointName` → `deploymentName`
- camelCase `endpointId` → `deploymentId` (props/vars/local query variable)
- local result var `endpoint` → `deployment`; `Endpoint` type alias → `SelectedDeployment`
[...]

PR #8462, opened 2026-07-29
fix(FR-3397): explain the deployment state when chat has no endpoint URL

### LINKED ISSUE #8440: Chat: replica/routing state alert is unreachable when the deployment has no endpoint URL
## Problem

On the Chat page, a deployment with no endpoint URL falls through to the generic error **"Endpoint URL is not valid."** (`error.InvalidBaseURL`) instead of explaining the deployment's actual state. The replica warning added in FR-3156 never appears, because it is gated behind the very condition that is false in this case.

## Root cause

The `chatui.NoDesiredReplicas` alert lives **inside** `CustomModelForm` (`react/src/components/Chat/CustomModelForm.tsx:74-81`), and `ChatCard` renders that form only when `baseURL` is truthy:

```
// react/src/components/Chat/ChatCard.tsx:564
{baseURL && (endpoint || agentEndpointUrl) && _.isEmpty(models) && (
  <CustomModelForm ... hasNoDesiredReplicas={hasNoDesiredReplicas} />
)}
```

`baseURL` is `undefined` whenever the endpoint URL is falsy:

```
// react/src/components/Chat/ChatCard.tsx:206-218
return endpointUrl ? new URL(basePath ?? '', endpointUrl).toString() : undefined;
```

So when the endpoint has no URL, the whole form is skipped — taking the replica alert with it — and the `!baseURL` branch renders the generic error instead:

```
// react/src/components/Chat/ChatCard.tsx:596
{!baseURL ? <Alert title={t('error.InvalidBaseURL')} type="error" ... /> : null}
```

## Two distinct gaps

**1. The `NoDesiredReplicas` alert is effectively unreachable.** A deployment whose desired replica count is 0 normally has no endpoint URL either, so the alert is suppressed in exactly the situation it was written for (FR-3156).

**2. There is no messaging for "desired replicas >= 1 but not routed yet".** `hasNoDesiredReplicas` is `endpoint?.replicas === 0` (`ChatCard.tsx:260`), which only covers desired replicas. A deployment that is PENDING / DEGRADED, or has no revision yet, has desired >= 1 and no active replica — that state has no message at all and lands on `error.InvalidBaseURL`.

The generic wording is misleading in both cases: the URL is not malformed, it does not exist yet.

## Reproduction

1. Create a deployment without any revision.
1. Open the Chat page and select that deployment.
1. Observed: the intro alert plus **"Endpoint URL is not valid."**
1. Expected: an explanation of why chat is unavailable (no revision / no active replica / desired replicas is 0).

Observed backend state for such a deployment:

```
name: "aaa"   url: ""   replicas: 1   status: DEGRADED   lifecycle: PENDING
```

Note `replicas` is 1, not 0 — so even if the form did render, the current `NoDesiredReplicas` condition would not fire.

## Proposed change

- Move the replica / routing-state alert **out of** the `baseURL &&` gate so it renders whenever the endpoint is known, with or without a URL.
- Distinguish the unavailable states and give each its own message: desired replicas is 0, no revision yet, and provisioned but no active replica (PENDING / DEGRADED).
- Reserve `error.InvalidBaseURL` for an actually malformed URL.

## Notes

- Reproduced on webui v26.8.0-rc.0 against manager 26.8.0rc1.
- `react/src/components/Chat/ChatCard.tsx` and `CustomModelForm.tsx` are identical between v26.8.0-rc.0 and `main` — this is not a release regression.
- Follow-up to FR-3156, which introduced the `NoDesiredReplicas` alert.

JIRA Issue: FR-3397

agatha197:
Resolves #8440 (FR-3397)

Stacked on #8509 (`test/FR-3332-chat-e2e-deployment-mocks`) → `main`.

_#8420 (`refactor/FR-3332-chat-deployment-naming`) and #8298 (`fr-3332`), previously below #8509 in this stack, have since merged to `main`, so #8509 now targets `main` directly._

## Problem

On the Chat page, a deployment with no endpoint URL fell through to the generic `error.InvalidBaseURL` — *"Endpoint URL is not valid."* — instead of saying why chat was unavailable. The URL is not malformed; it does not exist yet.

The `chatui.NoDesiredReplicas` warning added in FR-3156 lived inside `CustomModelForm`, and `ChatCard` renders that form only when `baseURL` is truthy. A deployment that cannot serve normally has no URL either, so the warning was suppressed in exactly the situation it was written for. There was also no message at all for "desired replicas ≥ 1 but nothing routed yet" (PENDING / DEGRADED, or no revision).

## Change

`ChatCard` now derives a single unavailability reason outside any URL gate and renders one alert for it:

| State | Message |
|---|---|
| Desired replica count is 0 | `chatui.NoDesiredReplicas` |
| No revision | `deployment.NoCurrentRevisionDeployed` |
| No `RUNNING` + traffic-`ACTIVE` replica | `deployment.NoRunningReplicas` |
| No endpoint URL issued yet | `chatui.NoEndpointUrlIssued` *(new key)* |
| URL supplied but unparseable | `error.InvalidBaseURL` |

Checks run most-specific first, so exactly one availability alert renders.

Placement follows what each message already looked like rather than unifying it:

- `chatui.NoDesiredReplicas` keeps the panel it had as `CustomModelForm`'s own alert (FR-3156) — above the form, on the same `colorBgContainer` panel, with the form's `paddingInline` so the two alerts line up when both are on screen. No bottom padding, so the form panel's own top padding is the single gap between them.
- Every other reason renders as a plain card-level alert in the spot `error.InvalidBaseURL` already occupied.

`CustomModelForm` only loses the replica alert it could no longer reach; its "cannot find model" warning is untouched.

### Two decisions worth reviewing

**"No revision" reads `revisionHistory.count`, not the revision pointers.** `currentRevisionId` / `deployingRevisionId` stay `null` for the entire PENDING stage — the manager only writes `deploying_revision_id` when it flips the lifecycle to `DEPLOYING`, and `current_revision_id` is assigned later still by the rollout handlers. Keying "no revision" off those pointers would report *"this deployment has no revision"* for the issue's own reproduction case, which does have one. `revisionHistory.count` is existence, not rollout state.

**"Active replica" is `RUNNING` **and** traffic-`ACTIVE`** — the manager's definition of a replica that is serving requests, and the same predicate the Chat deployment selector filters on (FR-3332). Deployment-level `status` is a monotonic lifecycle axis and cannot stand in for it. Note this is stricter than `DeploymentDetailPage`, which counts `RUNNING` only.

### Scope notes

- **`chatui.NoEndpointUrlIssued` is a state the issue did not enumerate.** Without it, a deployment that is otherwise healthy but has no URL yet still lands on the misleading `error.InvalidBaseURL`. Rare and transient, but it is the one remaining hole in the mapping.
- **`error.InvalidBaseURL` behaviour is unchanged when no deployment is resolved** (AI-agent binding, nothing selected, failed lookup). Narrowing that path is out of scope here.
- **No new version gate.** `revisionHistory` and `replicas(filter: { status, trafficStatus })` are 25.19.0 baseline — the same baseline `deployment(id:)` already assumes on the parent branch. (`ReplicaFilter.healthStatus` is the only 26.8.0 field, and it is not used. The `deployment-replica-nested-filter` client flag gates `DeploymentFilter.replicas` on the *list* query, which is a different field.)

## Not addressed here

The `e2e/chat` mocks still return the legacy `endpoint` shape (`endpoint_id` / `url` / `replicas`, keyed on `vars.endpointId`) while `ChatCardQuery` selects `deployment(id: $deploymentId)`. That drift arrives with #8420, which changed the query and touched no e2e file — it is not introduced by this PR, and updating the mocks belongs with the parent rather than in a child that would conflict with it. `e2e/chat/chat.spec.ts:670` also still carries a `.fixme` on the invalid-URL test; it should be re-run against the new alert once the mocks match, not un-`fixme`'d on reasoning alone.

## Verification

E2E was not run — the `e2e/chat` mocks are stale on the parent branch (above), and the suite needs a live cluster.


react/src/components/Chat/ChatCard.tsx
```diff
@@ -219,6 +220,76 @@ function createBaseURL(
...
+const CHAT_UNAVAILABLE_REASON_MESSAGE_KEYS: Record<
+  ChatUnavailableReason,
+  string
+> = {
+  'no-desired-replicas': 'chatui.NoDesiredReplicas',
+  'no-revision': 'deployment.NoCurrentRevisionDeployed',
+  'no-active-replica': 'deployment.NoRunningReplicas',
+  'no-endpoint-url': 'chatui.NoEndpointUrlIssued',
+  'invalid-base-url': 'error.InvalidBaseURL',
+};
```
yomybaby: This does not need to be separated separately; if you write t('deployment.NoCurrentRevisionDeployed') where needed, you can infer just by looking at the key in the same way, and it is also easy to check the text according to the key. Please remove CHAT_UNAVAILABLE_REASON_MESSAGE_KEY.

yomybaby (changes requested):
Image

Image

The background color of the above Alert is one is white and one is gray. If this is something that cannot be helped due to the view hierarchy, please add bottom padding when it's white to make it look good.

react/src/components/Chat/ChatCard.tsx
agatha197: Removed `CHAT_UNAVAILABLE_REASON_MESSAGE_KEYS` and replaced it with a `switch`-based `getChatUnavailableReasonMessage` helper — each `t()` call now sits next to the reason case it belongs to.

agatha197:
Addressed feedback from @yomybaby:

- **Message key indirection**: replaced `CHAT_UNAVAILABLE_REASON_MESSAGE_KEYS` with a `switch`-based `getChatUnavailableReasonMessage(t, reason)` helper — the `t()` call for each reason is now inline in its own `case`.
- **Alert background/padding**: the mismatch is inherent to the view hierarchy — the `no-desired-replicas` alert sits in its own white (`colorBgContainer`) wrapper meant to align with `CustomModelForm`'s panel below it, while the other reasons render as a plain card-level `Alert`. Added `paddingBlockEnd` to the white wrapper, applied only when `CustomModelForm` does *not* follow directly (tracked via a new `showCustomModelForm` flag), so the white background gets a proper bottom gap instead of running flush into the next element, without doubling the gap in the case where the form does follow.

agatha197:
Verified the padding fix by rendering all three states via mocked deployments:

**`no-desired-replicas` alert alone** (no `CustomModelForm` following) — bottom padding now matches the top, so the white background doesn't run flush into the next element:

no-desired-replicas-alone

**`no-desired-replicas` alert with `CustomModelForm` following** — no extra bottom padding added here, so the gap between them isn't doubled:

no-desired-replicas-with-form

**Fallback alert for another reason** (`no-active-replica`, untouched code path) — same visual result as the first case, confirming the inconsistency from the original screenshots is resolved:

fallback-no-active-replica

agatha197:
Closing this without merging.

The core fix (surfacing the correct unavailability reason instead of the generic "Endpoint URL is not valid" message) works, but the remaining review feedback on the `no-desired-replicas` alert's placement/background surfaced a deeper design question: this alert was originally coupled 1:1 with `CustomModelForm` (FR-3156) and always rendered together with it. This PR decoupled the reason detection from the form's own render condition, which introduced a new "alert alone, no form" state that doesn't fit either of the two existing visual treatments (the white form-matching panel, or the plain card-level alert) cleanly.

That's a genuine UX design call (unify the styling vs. keep the two treatments and special-case the padding) rather than a straightforward bug fix, so we're not pursuing it further right now. Closing this PR and the linked issue.

Issue #8453, opened 2026-07-29
Remove header project selector on super-admin pages via an explicit project prop contract


yomybaby:
## Problem Statement

Super admins working on the admin pages (`/admin-session`, `/admin-deployments`, `/admin-data`) still see the header's current-project selector, and its value silently affects what those pages show and — worse — where they create things. A super admin browsing all folders can create a folder, launch a FileBrowser/SFTP session, or deploy a model into whatever project happens to be selected in the header, without noticing. Meanwhile "this session/folder belongs to a different project" alerts fire based on a selector that is conceptually irrelevant here: admin pages operate above project scope.

## Solution

Remove the project selector from the header on the three super-admin routes, and stop admin-reachable components from reading the ambient current project. Components adopt an explicit **`project: { id, name } | null` required prop contract**: the page decides the project context. On admin pages (no ambient project) creation modals embed their own required project selector, session-launching buttons are disabled with an explanatory tooltip, and project-mismatch alerts simply do not render. Display/permission flows derive the project from the resource being viewed instead of from the header.

## User Stories

1. As a super admin, I want the header project selector hidden on `/admin-session`, `/admin-deployments`, and `/admin-data`, so that I am not misled into thinking these views are scoped by it.
1. As a super admin on the admin Data page, I want the folder-creation modal to ask me explicitly which project the folder belongs to, so that I never create a folder in a project I did not notice.
1. As a super admin creating a deployment from the admin Deployments page (or via the deploy action on a model folder), I want to pick the target project inside the modal, so that the deployment lands where I intend.
1. As a super admin adding a revision to a deployment, I want the model-folder picker, resource-group options, and resource limits scoped to the deployment's own project, so that the form always matches the deployment I am editing.
1. As a super admin opening a session detail from the admin session list, I do not want a "not in your project" warning, so that irrelevant project context stops adding noise.
1. As a super admin opening the folder explorer from the admin Data page, I want permissions evaluated against the folder's own project, so that what I can do does not depend on an invisible header selection.
1. As a super admin viewing a deployment detail, I want the Add-revision button available regardless of any header project selection, so that I can operate on any deployment.
1. As a super admin on the admin Data page, I want FileBrowser/SFTP buttons disabled with a clear reason ("cannot launch sessions from the admin menu"), so that I am not able to silently start a session in an arbitrary project.
1. As a super admin on the pending-sessions tab, I want to choose among ALL resource groups (not just those of some project), so that I can inspect any scheduler queue.
[...]

PR #8465, opened 2026-07-29
feat(FR-3410): explicit project for deployment creation, model-folder deploy, and model-card creation

### LINKED ISSUE #8457: Explicit project for deployment creation, model-folder deploy, and model-card creation
## Parent

FR-3407.

## What to build

Apply the project prop contract to the remaining creation flows around deployments: the deployment-creation modal embeds a required project selector when `project` is `null`; the model-folder deploy flow derives its project from the folder's own ownership (an in-modal selector only for user-owned folders) and never from the ambient project; model-card creation targets only the resolved model-store project — the silent ambient fallback is deleted, and creation is blocked with a clear state when the model-store project cannot be resolved.

## Acceptance criteria

- [ ] Deployment-creation modal: `null` → required project selector; the chosen project is used in the create mutation; general pages pass ambient explicitly
- [ ] Model-folder deploy: project-owned folder → project derived from ownership; user-owned folder → in-modal selector; ambient never used
- [ ] Model-card creation never falls back to the ambient project; unresolved model-store project blocks creation visibly
- [ ] Contract vitest tests for both modals
- [ ] Verification harness passes

## Blocked by

- The prop-contract/folder-creation ticket (contract + ADR + test pattern).

JIRA Issue: FR-3410

yomybaby:
Resolves #8457 (FR-3410)

> **Stacked PR** — this PR stacks on #8463 (FR-3408, `feat/FR-3408-project-prop-contract-folder-create`) and must be merged after it. The `docs/adr` / `projectContext.ts` context in the diff comes from that base.
>
> **#8494 (FR-3423) is stacked on top of this branch** and still carries the earlier design of `VFolderDeployModal`; it needs a rebase onto the current tip. See "Follow-ups" below.

Second application of the explicit project prop contract (ADR-0001, epic FR-3407 / #8453): the remaining creation flows around deployments stop reading the ambient current project.

## The design changed during review

The ticket's acceptance criteria described an in-modal project selector as the `null` behaviour for both deployment modals. Review concluded that `null` is not a state either modal can legitimately reach:

**A deployment is always created inside one project, and creation is offered only from the project-scoped user menu.** The admin menus edit deployments; they never create them. So instead of modelling "no project" and adding selector UI plus a runtime error path for it, both modals now take a **non-null project** and the surfaces that lack one simply do not offer the action. ADR-0001 has been updated with this as explicit guidance ("prefer no null state over the Modal tier when every reachable surface is project-scoped"), and the compliance checklist now starts by asking whether `null` is reachable at all.

### `DeploymentSettingModal` — project contract encoded in the props type

`DeploymentSettingModalProps` is now a union:

- **Edit-only call sites** pass a non-null `deploymentFrgmt` and no `project` at all (`project?: never`).
- **Project-scoped call sites**, which may open either mode, must pass a non-null `project`.

Removed as a consequence: the in-modal `ProjectSelectForAdminPage`, the `projectId` form field, the `selectedProject` state, the `afterOpenChange` reset, and the `message.error(t('general.ErrorOccurred'))` "missing project" branch. Edit-only call sites now render the modal under a non-null fragment guard, which is what their `open={!!editingDeployment}` already expressed.

The create branch keeps one annotated non-null assertion on `project`: `deploymentFrgmt` is an opaque Relay fragment key rather than a unit type, so TypeScript cannot use it as a discriminant *inside* the component even though it enforces the pairing at every call site.

### `VFolderDeployModal` — required project, preloaded query

- Takes a required non-null `project` and a `vfolderId`; the create mutation targets exactly that project. No in-modal selector, no ownership derivation.
- The folder's own owning project is deliberately **not** consulted. A mismatch between the page's project and a project-type folder's owner is a reporting concern — a possible future non-blocking alert — not a targeting one. It is not implemented here.
- Converted from `useLazyLoadQuery` + `useDeferredValue(open)` to a **preloaded query** (`useQueryLoader` in the opener's Deploy handler), following `DeploymentSchedulingHistoryModal` / `DeploymentBasicInfoCard`. This fixes the reported close-time bug at the root: the query no longer derives variables during render from state that the close path clears, so it can no longer fire a request for an empty global id. It also starts the fetch in the click event instead of after the modal mounts.
- The `loading={deferredOpen !== open}` skeleton it used to render was already unreachable — `BAIUnmountAfterClose` mounts the modal already-open and `useDeferredValue` returns the current value on the initial render — so it was dropped rather than ported.
- Auto-deploy no longer requires an ownership-derived project: single preset + single resource group is enough again.

### Call sites

- `VFolderNodes` / `VFolderNodesV2` gain `project: ProjectContextOrNull` and a `canDeploy` gate on the row action. A page with no project context has no valid deploy target, so `AdminVFolderNodeListPage` (`project={null}`) stops offering "Deploy as service".
- `VFolderNodeListPage`, `ProjectAdminDataPage`, `DeploymentListPage` pass their page-level project; `DeploymentListPage` renders the setting modal only when that project resolves.
- Edit-only call sites (`AdminDeployment`, `ProjectAdminDeploymentsPage`, `DeploymentBasicInfoCard`) pass no `project` at all.

### Removed: three unreachable create-mode modals

`VFolderNodes`, `VFolderNodesV2` and `ModelCardDrawer` each mounted a `DeploymentSettingModal` whose `isCreateDeploymentOpen` was never set to `true`. This is dead on `origin/main` too — not a regression introduced here — so the modals, their `useToggle` state, and `ModelCardDrawer`'s `project` prop (which existed only to feed one of them) are gone.

### `AdminModelCardSettingModal` — ambient fallback deleted

- `modelStoreProjectId: modelStoreProject?.id ?? currentProject.id!` → `modelStoreProject.id` only. When the model-store project cannot be resolved, creation stays visibly blocked by the existing `modelStore.ProjectNotFound` alert + disabled OK button.
- The "change your current project to model-store first" Popconfirm before folder creation is gone: `FolderCreateModalV2` now targets the model-store project explicitly, so no ambient switch is needed (the `isModelStoreProject` prop is removed).
- The now-dead `group(id: $currentProjectId)` query field and the ambient read feeding it are removed from `AdminModelCard.tsx` and `AdminDeploymentPage.tsx` (the `model-store-management` case).

### Suspense / transition

`VFolderDeployModal`'s local Suspense boundary is kept, and kept **mounted unconditionally** with the modal switched in underneath it. The Deploy action runs inside `startTransition` (`BAINameActionCell`), which only protects the render that state update causes — the preloaded presets query. The modal holds further suspending sources that start *after* it commits, outside any transition: `useProjectResourceGroups` (`useSuspenseTanQuery`), `BAIProjectResourceGroupSelect` via the same hook, and `BAIAvailablePresetSelect`, whose value query re-runs with `skip: false` as soon as the modal preselects the first preset. Removing the boundary blanks the folder list; keeping it mounted means the initial transition still never shows its fallback.

### Tests / i18n / docs

- Contract vitest tests for both modals, rewritten for the final shape: no selector rendered, resource-group options scoped to the passed project, mutation variables carrying exactly that project.
- `VFolderDeployModal.test.tsx` documents a relay-test-utils gotcha: `loadQuery` reaches the mock network *before* the operation is registered as pending, so `queueOperationResolver` never sees it — the tests resolve the started request explicitly.
- **No net i18n change.** The three keys added mid-review (`deployment.TargetProjectRequired`, `deployment.FolderOwnershipUnresolved`, `…Description`) were all removed again with the UI that used them, so `resources/i18n/` is unchanged against the base. `data.folders.TargetProjectRequired` is a different key and belongs to `FolderCreateModalV2`.
- **No net Relay artifact change for `VFolderDeployModalQuery`** — the query is variable-less again, as on the base. Only `AdminModelCardQuery.graphql.ts` changes, from the removed `group(id:)` field.
- ADR-0001 updated: the FR-3410 entry now describes what was actually built, and the "no null state" guidance plus the revised compliance checklist are recorded.

## Known gaps

- The auto-deploy path (single preset + single resource group → mutate and navigate without modal chrome) and its close behaviour still have no regression test.
- The project/folder-owner mismatch alert described above is deliberately not implemented.

## Follow-ups

- **#8494 (FR-3423) needs a rebase.** It is stacked on this branch and still contains the pre-review `VFolderDeployModal` (ownership derivation + in-modal selector). Its `noDeployTooltip` prop and this PR's `canDeploy` gate overlap and should be reconciled into one mechanism — `canDeploy` is a `TODO(FR-3423)` in the source. `canDeploy` has to exist here regardless: the stack merges bottom-up, and without it `/admin-data` would ship a deploy action that does nothing.

## Verification

```
scripts/verify.sh: Relay PASS / Lint PASS / Format PASS / TypeScript PASS /
Vite warmup PASS / Terminology PASS → ALL PASS
react vitest suite: 61 files / 1082 tests passed
```


react/src/components/VFolderDeployModal.tsx
```diff
@@ -95,24 +111,46 @@ const VFolderDeployModal: React.FC<VFolderDeployModalProps> = ({
...
 
+  // Project-owned folder → the folder's own project is the deploy target.
+  // User-owned folder (or unresolved node) → no derived project; the user
+  // must pick one with the in-modal selector below.
+  const ownershipProject: ProjectContextOrNull =
+    vfolder_node?.ownership_type === 'group' &&
+    vfolder_node.group &&
+    vfolder_node.group_name
+      ? { id: vfolder_node.group, name: vfolder_node.group_name }
+      : null;
```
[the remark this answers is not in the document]
yomybaby: Fixed in `0ad2f3c0b`. Ownership is now a three-way value — `user` / `group` / `unresolved` — instead of "group with both fields, else user". Only `ownership_type === 'user'` renders the project selector; an unresolved folder (unknown ownership, or a group folder missing `group`/`group_name`) shows an error alert and disables Deploy rather than offering a picker. Applies to both sites you flagged.

react/src/components/DeploymentSettingModal.tsx
```diff
@@ -59,8 +78,15 @@ const DeploymentSettingModal: React.FC<DeploymentSettingModalProps> = ({
...
-  const { id: projectId, name: projectName } = useCurrentProjectValue();
   const currentDomain = useCurrentDomainValue();
+  // ADR-0001: the target project comes exclusively from the `project` prop.
+  // When it is `null`, the user picks the target project with the in-modal
+  // selector below; the chosen value is tracked here so the resource-group
+  // options follow the chosen project, never the ambient one.
+  const [selectedProject, setSelectedProject] = useState<ProjectContext | null>(
+    null,
+  );
+  const effectiveProject = project ?? selectedProject;
```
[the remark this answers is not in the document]
yomybaby: Fixed in `0ad2f3c0b` — `selectedProject` is reset in `afterOpenChange` when the modal opens, mirroring `FolderCreateModalV2`.

react/src/components/AdminModelCardSettingModal.tsx
```diff
@@ -559,9 +520,14 @@ const AdminModelCardSettingModal: React.FC<AdminModelCardSettingModalProps> = ({
...
       </BAIModal>
+      {/* New model folders are created in the model-store project — the same
+          project that backs the VFolder selector — never in the ambient
+          current project. The button that opens this modal renders only when
+          the model-store project is resolved, so `toProjectContext` yields a
+          non-null project here in practice. */}
       <FolderCreateModalV2
         open={isOpenCreateFolderModal}
-        project={toProjectContext(currentProject)}
+        project={toProjectContext(modelStoreProject ?? {})}
```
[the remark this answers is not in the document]
yomybaby: Fixed in `0ad2f3c0b`. The modal now derives `modelStoreProjectContext = toProjectContext(modelStoreProject)` once and gates the form, the OK button, the ProjectNotFound alert, and the create mutation on that single value — so the gate can no longer disagree with what `FolderCreateModalV2` receives when `name` is null.

react/src/components/VFolderDeployModal.tsx
```diff
@@ -95,24 +111,54 @@ const VFolderDeployModal: React.FC<VFolderDeployModalProps> = ({
...
+            ownership_type
+            group
+            group_name
+          }
         }
       `,
-      {},
+      { vfolderGlobalId: toGlobalId('VirtualFolderNode', vfolderId ?? '') },
       {
         fetchPolicy: deferredOpen ? 'store-and-network' : 'store-only',
```
agatha197: **High** — 여기 `fetchPolicy` 가드가 `deferredOpen`만 보고 현재 `open`을 확인하지 않습니다. 취소/닫기 시 `open`은 즉시 false가 되지만 `deferredOpen`은 한 렌더 뒤처져 stale true를 유지하는데, 그 렌더에서 `vfolderId`도 이미 undefined라 `toGlobalId('VirtualFolderNode','')`(빈 ID)로 실제 네트워크 요청이 나갑니다. 형제 컴포넌트 `ModelCardDrawer.tsx`는 `deferredOpen && open && modelCardId`로 이미 이 문제를 막아뒀습니다 — 동일하게 `&& open && vfolderId`를 추가해 주세요.

react/src/components/VFolderDeployModal.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+  );
+  return { environment, onClose };
+};
+
+const submitDeploy = async (user: ReturnType<typeof userEvent.setup>) => {
+  await user.click(screen.getByTestId('mock-resource-group-select'));
+  await user.click(screen.getByRole('button', { name: 'modelStore.Deploy' }));
+};
+
+describe('VFolderDeployModal project derivation contract (ADR-0001)', () => {
```
agatha197: **Medium** — `ownership_type: 'group'`인데 `group`/`group_name`이 null인 경우(불완전한 폴더 데이터) 새로 추가된 `FolderOwnershipUnresolved` Alert 및 버튼 disable 로직을 검증하는 테스트가 없습니다. 이 PR이 새로 도입한 사용자 대면 에러 상태이니 케이스 추가를 권장합니다.

react/src/components/DeploymentSettingModal.tsx
```diff
@@ -154,15 +180,17 @@ const DeploymentSettingModal: React.FC<DeploymentSettingModalProps> = ({
             },
           });
         } else {
-          if (!projectId) {
-            message.error(t('general.ErrorOccurred'));
+          // The required rule on the in-modal project selector (null mode)
+          // guarantees a chosen project before validation passes; this guard
+          // only narrows the type.
+          if (!effectiveProject) {
```
agatha197: **Minor** — `!effectiveProject`일 때 조용히 `return`합니다(과거엔 `message.error(t('general.ErrorOccurred'))` 호출). 폼 필수 검증으로 사실상 도달 불가능하다는 주석은 맞지만, 방어적으로 에러 토스트를 유지해도 손해는 없어 보입니다.

agatha197 (changes requested):
**🔴 High — `VFolderDeployModal` 종료 시 잘못된 `vfolder_node` 쿼리 재요청 가능성**

`open`과 `vfolderId`가 같은 state(`deployFallbackVfolderId`)에서 동시에 파생되는데, 취소/배포 실패/배포 성공 세 경로 모두 두 값을 같은 렌더에서 동시에 지웁니다. `useDeferredValue(open)`는 값이 바뀐 뒤 한 렌더는 이전(stale) 값을 유지하므로, 그 렌더에서 `open=false`, `deferredOpen=true(stale)`, `vfolderId=undefined`가 됩니다. 이때 `fetchPolicy`는 `deferredOpen`만 보고 `'store-and-network'`를 선택하고, `vfolderGlobalId`는 `toGlobalId('VirtualFolderNode', '')`(빈 ID)로 계산되어 **모달을 닫을 때마다 캐시된 적 없는 빈 ID로 실제 네트워크 요청이 나갑니다.** `store-and-network`는 캐시 미스에서 suspend하므로, 감싸는 `<Suspense fallback={null}>`이 트리거되어 닫힘 애니메이션이 끊기거나 auto-deploy 실패 경로의 `afterClose` 호출이 이 무의미한 요청이 끝날 때까지 지연될 수 있습니다.

같은 PR에서 손대지 않은 형제 컴포넌트 `ModelCardDrawer.tsx`는 이미 이 문제를 막아뒀습니다:
```tsx
fetchPolicy:
  deferredOpen && open && modelCardId
    ? 'store-and-network'
    : 'store-only',
// "Skip the network round-trip until the drawer has actually committed
// to opening and a model-card UUID is known."
```
`VFolderDeployModal`도 동일하게 `deferredOpen && open && vfolderId`로 가드하는 걸 제안합니다.

**🟡 Medium — 새 `isOwnershipUnresolved` 에러 상태가 테스트되지 않음**

`ownership_type: 'group'`인데 `group`/`group_name`이 null인 경우(불완전한 폴더 데이터) 새로 추가된 `FolderOwnershipUnresolved` Alert와 버튼 disable 로직을 검증하는 테스트가 없습니다. 이 PR이 새로 도입한 사용자 대면 에러 UI이니 최소 1개 케이스 추가를 권장합니다.

**🟢 Minor — 무언(silent) 실패 가능성**

`!effectiveProject`일 때 조용히 `return`합니다(과거엔 `message.error(t('general.ErrorOccurred'))` 호출). 폼 필수 검증으로 사실상 도달 불가능하다는 주석은 맞지만, 방어적으로 에러 토스트를 유지해도 손해는 없어 보입니다. 향후 리팩터링으로 `selectedProject`/Form 내부 필드가 어긋나면 "버튼 눌러도 무반응"이 될 수 있습니다.

**🟢 Minor — PR 설명과 실제 diff 불일치**

본문은 새 i18n 키로 `deployment.TargetProjectRequired` 하나만 언급하지만 실제로는 `FolderOwnershipUnresolved`/`FolderOwnershipUnresolvedDescription`까지 3개가 21개 로케일에 추가됐습니다. 또 "removed from `AdminModelCardListPage`"라고 되어 있는데 실제 변경 파일은 `AdminDeploymentPage.tsx`(model-store-management 케이스)와 `AdminModelCard.tsx`입니다.

react/src/components/VFolderDeployModal.tsx
yomybaby: Fixed at the root rather than by widening the guard.

`VFolderDeployModal` now uses a **preloaded query**: the opener calls `loadQuery` in the Deploy click handler (`useQueryLoader`) and the modal reads it with `usePreloadedQuery`, following the `DeploymentSchedulingHistoryModal` / `DeploymentBasicInfoCard` pattern. There is no `open`-derived `fetchPolicy` left, and no render-time variable derivation — so the empty-global-id request cannot happen by construction rather than being guarded against.

The query also no longer needs `$vfolderGlobalId` at all: the target project comes from a required `project` prop instead of the folder’s ownership, so the operation is variable-less again and `VFolderDeployModalQuery.graphql.ts` is unchanged against the base.

Two related findings while doing this:

- The `loading={deferredOpen !== open}` skeleton this component rendered was already unreachable — `BAIUnmountAfterClose` mounts the modal already-open, and `useDeferredValue` returns the current value on the initial render. Dropped rather than ported.
- The local `<Suspense>` boundary has to stay, and stay mounted unconditionally. `startTransition` only protects the render its state update causes; `useProjectResourceGroups` (`useSuspenseTanQuery`), `BAIProjectResourceGroupSelect`, and `BAIAvailablePresetSelect`’s value query all suspend *after* the modal commits, outside any transition. Removing the boundary blanks the folder list.

react/src/components/VFolderDeployModal.test.tsx
yomybaby: Obsolete — the state you asked to cover no longer exists.

Review concluded that deriving the deploy target from the folder’s ownership was the wrong model: a deployment is always created inside one project, and every surface that can open this modal is already project-scoped. `VFolderDeployModal` now takes a required non-null `project` from the page, so `vfolder_node { ownership_type group group_name }`, the `ownership` derivation, the `isOwnershipUnresolved` alert and its two i18n keys (all 21 locales) are all gone.

A mismatch between the page’s project and a project-type folder’s owner is a reporting concern — a possible future non-blocking alert — not a targeting one, and is deliberately not implemented here.

The rewritten contract tests cover the replacement contract: no selector rendered, resource-group options scoped to the passed project, and the mutation carrying exactly that project.

react/src/components/DeploymentSettingModal.tsx
yomybaby: Agreed that the silent `return` was wrong, but the fix is the type system rather than restoring the toast.

Creation is offered only from the project-scoped user menu — the admin menus edit deployments, they never create them — so "no project in create mode" is not a state that can legitimately be reached, and it should not be representable. `DeploymentSettingModalProps` is now a union:

- edit-only call sites pass a non-null `deploymentFrgmt` and no `project` at all (`project?: never`);
- project-scoped call sites, which may open either mode, must pass a non-null `project`.

So the silent branch is gone along with the in-modal selector, the `projectId` form field, the `selectedProject` state, the `afterOpenChange` reset and the `deployment.TargetProjectRequired` key. Edit-only call sites render the modal under a non-null fragment guard, which is what their `open={!!editingDeployment}` already expressed.

One annotated non-null assertion remains in the create branch: `deploymentFrgmt` is an opaque Relay fragment key rather than a unit type, so TypeScript cannot use it as a discriminant *inside* the component even though it enforces the pairing at every call site.

Also caught while doing this: `VFolderNodes`, `VFolderNodesV2` and `ModelCardDrawer` each mounted a `DeploymentSettingModal` whose `isCreateDeploymentOpen` was never set to `true` — dead on `origin/main` too — so those three unreachable modals are removed.

react/src/components/VFolderDeployModal.tsx
```diff
@@ -309,6 +352,33 @@ const VFolderDeployModal: React.FC<VFolderDeployModalProps> = ({
...
+            // the OK button stays disabled until a project is chosen.
+            required
+          >
+            {/* The Suspense boundary swallows Form.Item's injected props, so
+                the selection is tracked in component state — same manual-
+                wiring pattern as FolderCreateModalV2. */}
+            <Suspense fallback={<Skeleton.Input active block />}>
+              <ProjectSelectForAdminPage
+                data-testid="vfolder-deploy-project-select"
+                domain={currentDomain}
```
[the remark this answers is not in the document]
yomybaby: Obsolete — `ProjectSelectForAdminPage` is no longer rendered by either deployment modal.

Both `VFolderDeployModal` and `DeploymentSettingModal` now take the target project from the page as a required non-null prop, so there is no in-modal project selector left and the membership-filter concern does not apply here. A surface without a project context simply does not offer the action (`canDeploy` on the folder row).

yomybaby:
Re-review requested — all four open threads are addressed and resolved, but the design moved rather than being patched, so the diff is worth a fresh read rather than a delta read.

**The design changed.** The ticket's acceptance criteria called for an in-modal project selector as the `null` behaviour of both deployment modals. Review concluded `null` isn't a state either modal can legitimately reach: a deployment is always created inside one project, and creation is offered only from the project-scoped user menu — the admin menus edit deployments, never create them. Both modals now take a **non-null project**, and the surface that lacks one stops offering the action. ADR-0001 records this as guidance and the compliance checklist now begins by asking whether `null` is reachable at all.

Per finding:

- **High (fetchPolicy / empty-id query)** — fixed at the root. `VFolderDeployModal` uses a preloaded query (`useQueryLoader` in the Deploy handler → `usePreloadedQuery`), so there is no `open`-derived fetch policy and no render-time variable derivation. The query is variable-less again, so its Relay artifact is unchanged against the base.
- **Medium (`isOwnershipUnresolved` untested)** — the state is gone with the ownership derivation; tests were rewritten for the replacement contract.
- **Minor (silent failure)** — fixed with a props union rather than by restoring the toast, so "create without a project" is no longer representable.
- **Minor (PR description)** — the body is rewritten against the final diff.

**Worth your attention:**

1. `canDeploy` on `VFolderNodes` / `VFolderNodesV2` overlaps with #8494 (FR-3423), which is stacked on this branch and takes the opposite position — it deliberately does *not* infer the admin context from `project` being `null`, and disables the action with a tooltip instead of hiding it. `canDeploy` still has to exist here, because the stack merges bottom-up and without it `/admin-data` would ship a deploy action that silently does nothing. #8494 needs a rebase onto this tip regardless; the two mechanisms should be reconciled into one there.
2. Three unreachable `DeploymentSettingModal` instances (in `VFolderNodes`, `VFolderNodesV2`, `ModelCardDrawer`) are removed. Their `isCreateDeploymentOpen` was never set to `true`, on `origin/main` too. If that was a stub someone intended to wire up rather than dead code, say so and I'll restore it.
3. The create branch keeps one annotated non-null assertion on `project`. `deploymentFrgmt` is an opaque Relay fragment key, not a unit type, so TypeScript can't use it as a discriminant *inside* the component even though the union enforces the pairing at every call site. Open to a better encoding.

**Known gaps, stated rather than hidden:** the auto-deploy path (single preset + single resource group → mutate and navigate with no modal chrome) and its close behaviour still have no regression test, and the project/folder-owner mismatch alert is deliberately not implemented.

Verification: `scripts/verify.sh` → `=== ALL PASS ===`; `react` vitest 61 files / 1082 tests passed.

agatha197 (approved):
LGTM

PR #8467, opened 2026-07-29
feat(FR-3411): scope Add revision and resource allocation to the deployment's own project

### LINKED ISSUE #8460: Scope Add revision and resource allocation to the deployment's own project
## Parent

FR-3407.

## What to build

Scope the Add-revision flow to the deployment's own project. The resource-allocation form fragment takes a required non-null `project` and scopes accessible resource groups, presets, and resource limits to it (the parent owns any selector; the fragment never reads ambient state). The Add-revision modal derives the deployment's project (id and name) from the deployment itself and feeds it to the form fragment and the model-folder picker, so adding a revision works identically regardless of any header selection. General call sites (session launcher and friends) pass the ambient project explicitly with no behavior change.

## Acceptance criteria

- [ ] Resource-allocation form fragment requires a non-null project and scopes resource-group options, presets, and limits to it
- [ ] Add-revision modal derives the deployment's project; the model-folder picker and resource form follow it
- [ ] Works on the admin deployment detail regardless of header project selection
- [ ] Session launcher and other general call sites pass ambient explicitly — no behavior change
- [ ] Contract vitest tests; verification harness passes

## Blocked by

- The prop-contract/folder-creation ticket.

JIRA Issue: FR-3411

yomybaby:
Resolves #8460 (FR-3411)

> Stacks on #8465 (FR-3410), which stacks on #8463 (FR-3408). Third layer of the ADR-0001 explicit-project-prop-contract chain (epic FR-3407 / #8453).

## What this does

Scopes the Add-revision flow to the deployment's own project, and converts the shared resource-allocation form fragment to the explicit project prop contract.

### `ResourceAllocationFormItems` — form-fragment tier (required non-null `project`)

- Takes a REQUIRED `project: ProjectContext` prop; the internal `useCurrentProjectValue` read and the null-project throw are deleted. Per ADR-0001 form-fragment tier, it never embeds a project selector — the parent owns any selector and passes the chosen/derived value down.
- The prop drives all three project-dependent pieces:
  - the `accessible_scaling_groups(project_id: $projectID)` query variable,
  - `BAIProjectResourceGroupSelect`'s `projectName`,
  - `useResourceLimitAndRemaining`'s `currentProjectName` (REST preset check `group` param).
- **Indirect ambient read fixed too**: `useResourceLimitAndRemaining` used the ambient project's derived resource-group atom to decide whether `scaling_group` may accompany the preset check. It now accepts an optional `accessibleResourceGroupNames` override; `ResourceAllocationFormItems` passes the names from its own per-project `accessible_scaling_groups` query, so the check is fully keyed off the passed project. Other consumers keep the ambient fallback unchanged.

### `DeploymentAddRevisionModal` — derive-from-resource tier

- Derives the deployment's project — **id and name** — from the deployment itself: the fragment now selects `metadata { projectId, projectV2 @since(26.4.3) { basicInfo { name } } }` (same shape `DeploymentBasicInfoCard` already uses).
- The derived project scopes:
  - the model-folder picker (`BAIVFolderSelect currentProjectId`) in BOTH Preset and Custom forms (the ambient read is gone),
  - the resource-allocation form (`project` prop),
  - in-modal model-folder creation (`FolderCreateModalV2 project`).
- Adding a revision works identically regardless of any header project selection.
- **Defensive null path**: if the project cannot be resolved (missing metadata or a pre-26.4.3 manager without `projectV2`), a visible warning alert (`deployment.CannotResolveDeploymentProject`) is shown, the submit button and folder picker are disabled, and the resource form is not rendered — never a silent fallback to ambient.

### General call sites

- `SessionLauncherPage` — the only other render site of `ResourceAllocationFormItems` — narrows ambient at page level with `toProjectContext(useCurrentProjectValue())` and passes `project` explicitly. The previous in-component throw semantics for an unresolvable ambient project are preserved at the page, so behavior is byte-identical.
- `LaunchMultipleSessionsModal` does not render the component (it only imports the form-value type and already reads ambient at its own level for `useResourceLimitAndRemaining`) — no change needed.

### Tests (contract vitest, FR-3408/3410 pattern)

- `ResourceAllocationFormItems.contract.test.tsx`: with a deliberately different mocked ambient project (decoy), asserts the RG query variables, the RG select `projectName`, and the preset-check params (`group` + `scaling_group`) all carry exactly the passed project.
- `DeploymentAddRevisionModal.test.tsx`: asserts the folder picker, resource form, and folder-creation modal all receive the deployment's derived project (never the ambient decoy), and that an unresolvable project disables submission visibly.

### i18n / docs

- New key `deployment.CannotResolveDeploymentProject` in all 21 locales.
- ADR-0001 applications paragraph updated with the FR-3411 conversions.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Vite warmup / Terminology)
- react vitest suite: **62 files / 1063 tests passed** (includes the 5 new contract tests)


react/src/components/DeploymentAddRevisionModal.tsx
agatha197: It is possible to create a new folder, but it would be better to make it disabled when deploymentProject is null like `BAIVFolderSelector` for UI consistency.

react/src/components/DeploymentAddRevisionModal.tsx
```diff
@@ -2023,7 +2052,7 @@ const DeploymentAddRevisionModal: React.FC<DeploymentAddRevisionModalProps> = ({
       )}
       <FolderCreateModalV2
```
agatha197, line 2161: Clicking it opens `FolderCreateModalV2` with project=null, which falls back to its own in-modal project selector — letting the user pick an arbitrary project to create the folder in. onRequestClose then unconditionally calls setFieldValue('modelFolderId', ...), silently populating the disabled select field with the ID of a folder created in that arbitrary project. The form as a whole remains non-submittable, but the user is left with the false impression that a folder has been "selected."

agatha197 (changes requested):

react/src/components/DeploymentAddRevisionModal.tsx
yomybaby: Fixed in `eddc63a4b`. Both "create model folder" buttons (preset mode and custom mode) now carry the same `disabled={!deploymentProject}` gate as the adjacent `BAIVFolderSelect`, so the two controls are consistent.

react/src/components/DeploymentAddRevisionModal.tsx
yomybaby: Fixed in `eddc63a4b`. Root cause is the same as the sibling thread, so the disabled button closes this path: with no resolved `deploymentProject` the modal can no longer be opened, and therefore can never fall back to its `project={null}` tier (the in-modal project selector).

I also added a defensive guard at the render site — `open={isModelFolderCreateModalOpen && !!deploymentProject}` — so even if the flag were set some other way, `onRequestClose` can never write back the id of a folder created in a project this revision cannot mount.

agatha197 (approved):
LGTM

agatha197 (dismissed):
Please resolve conflicts

react/src/hooks/useResourceLimitAndRemaining.tsx
```diff
@@ -154,9 +165,11 @@ export const useResourceLimitAndRemaining = ({
         group: currentProjectName,
       };
 
+      const validResourceGroupNames =
+        accessibleResourceGroupNames ?? _.map(nonSftpResourceGroups, 'name');
```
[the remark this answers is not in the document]
yomybaby: Confirmed, and not fixed here — deliberately.

`useResourceGroupsForCurrentProject()` is still called unconditionally, so the ambient scaling-group/vfolder-host requests fire even when `accessibleResourceGroupNames` is supplied. Hooks can't be called conditionally, so the only real fix is the one you describe: split the ambient fallback out into a wrapper hook and have the explicit-project path call the pure one. That touches every caller of `useResourceLimitAndRemaining`, most of which are outside this stack, so doing it inside FR-3411 would widen the PR well past its subject.

What this PR does guarantee is the targeting contract: the values the form *uses* come from the passed project, never from the ambient one. The remaining ambient read only costs a redundant fetch and a suspend. That's recorded on the prop itself (`accessibleResourceGroupNames` in `useResourceLimitAndRemaining.tsx`) as a sanctioned interim state, in the same shape as the other ADR-0001 interim readers.

Resolving as a tracked follow-up rather than silently.

PR #8480, opened 2026-07-30
test(FR-3332): rewrite chat e2e mocks for the Deployments API

agatha197:
Follow-up to #8298 / #8420 (FR-3332). No dedicated issue — this repairs the e2e coverage for the migration those PRs landed, and adds the coverage #8462's alert was missing; happy to file a Jira ticket if you'd rather it be tracked separately.

Stacked on #8462 (`fix/FR-3397-chat-deployment-state-alert`) → #8420 (`refactor/FR-3332-chat-deployment-naming`) → #8298 (`fr-3332`) → `main`. Rebased onto #8462's latest revision (the one that stopped suppressing `chatui.CannotFindModel`).

## Why

Every test in `e2e/chat/` fails against a live manager on top of the FR-3332 stack. The specs' mocks still describe the legacy Graphene shapes the Chat surface no longer queries. Two breakages, both silent:

- **Same operation name, wrong shape.** `ChatPageQuery` and `ChatCardQuery` kept their names, so the interceptor still matched them — but they returned `endpoint_list { items { endpoint_id } }` and `endpoint { url }` while the migrated queries select `myDeployments` and `deployment(id:)`. Relay received payloads missing the fields it asked for, which made the `@catch` result not-ok and dropped the page into its error boundary.
- **Renamed operations are not an error.** `EndpointSelectQuery` / `EndpointSelectValueQuery` no longer exist (now `DeploymentSelect*`). `graphql-interceptor` forwards unmatched operations to the real backend, so the specs asserted mock data (`mock-endpoint`) against whatever deployments the target actually had.

`DeploymentTokenSelectQuery` was never mocked at all, so it sent the fake mock UUID to the manager and got back `badly formed hexadecimal UUID string`.

Neither FR-3332 commit touched `e2e/chat/`, and `origin/main`'s `ChatPageQuery` does select `endpoint_list` — so the mocks were correct before the migration and only these queries moved.

## What changed

| | Before | After |
| --- | --- | --- |
| `ChatPageQuery` | `endpoint_list { items { endpoint_id } }` | `myDeployments { edges { node { id } } }` |
| `ChatCardQuery` | `endpoint { endpoint_id url replicas name }` | `deployment { id networkAccess { endpointUrl } replicaState { … } revisionHistory { count } activeReplicas { count } metadata { name } }` |
| `EndpointSelectQuery` | `endpoint_list { total_count items }` | `DeploymentSelectQuery` → `myDeployments { count edges { node … } }` |
| `EndpointSelectValueQuery` | `endpoint { endpoint_id name url }` | `DeploymentSelectValueQuery` → `deployment { id metadata { name } networkAccess { endpointUrl } }` |
| `DeploymentTokenSelectQuery` | *(unmocked)* | empty `accessTokens` connection |

- Global Relay IDs are built the way `toGlobalId` does (`base64("ModelDeployment:<uuid>")`), since Chat holds local UUIDs and converts on the way into `deployment(id:)`. `fromMockDeploymentGlobalId` reverses it so a factory can tell deployment A from B.
- One shared `chatGraphQLMocks()` map replaces the four restated inline copies; a spec spreads it and overrides a single operation. Reply strings and deployment names are exported constants, since restated literals are exactly what let these assertions drift.
- Terminology follows the migration: `mock-endpoint` → `mock-deployment`, `setupChatPageWithTwoEndpoints` → `setupChatPageWithTwoDeployments`.
- Removed `chatPageQueryTwoEndpointsMockResponse` — dead, no caller.

## Coverage for #8462's unavailability alert

`ChatCardQuery` now selects `replicaState`, `revisionHistory` and an `activeReplicas` alias, which ChatCard turns into a single unavailability reason. Only the no-endpoint-URL branch had a test, and it was skipped — so this adds a table-driven case per reason:

| Condition | Reported message |
| --- | --- |
| desired replica count 0 | `chatui.NoDesiredReplicas` |
| no revision | `deployment.NoCurrentRevisionDeployed` |
| no active replica | `deployment.NoRunningReplicas` |
| no endpoint URL issued | `chatui.NoEndpointUrlIssued` |
| URL supplied, unparseable | `error.InvalidBaseURL` |

Each case asserts its own message **and** that no other reason's message appears, so a regression that reports the wrong cause fails instead of passing on a substring. A separate precedence test pins the most-specific-first ordering: a deployment scaled to zero also has no revision, no active replica and no URL, and must still report the desired-replica reason.

Two more behavioural points:

- **The previously skipped null-URL test is un-skipped.** Its FIXME blamed Relay's `store-or-network` policy for serving a cached URL; the real cause was the mock describing the legacy `endpoint { url }` shape, which the migrated query never reads.
- **Both alerts can coexist.** One test covers the change in #8462 that stopped suppressing `chatui.CannotFindModel` while a reason is shown: a deployment with a reachable URL, an empty model list and no desired replicas reports both.

The table's `composer` column records **observed** behaviour, not an aspiration: the composer is disabled only when no usable base URL resolved. A deployment that is reachable but cannot answer (scaled to zero, no revision, no active replica) still accepts typing, so the alert is the only thing telling the user a send will not get a reply. Flagging it in case that is worth changing — this PR only pins the current contract.

## Flakiness fixes

Both were pre-existing races that only surfaced once the suite got far enough to hit them:

- **`waitForChatReady` gates every setup path on an *enabled* composer.** Visibility is not readiness: the composer renders disabled until the pane resolves a deployment URL and fetches its model list, and `press('Enter')` on a disabled composer silently does nothing — the failure then surfaces as a missing reply rather than as "the page was not ready".
- **`clickSyncToggle` waits for the sync toggle's lucide icon to flip** (`toggle-right` ⇄ `toggle-left`). The specs previously waited for the pane's input to become empty, which is vacuously true *before* the click lands, so the next step raced the state update. The genuine "input is cleared" assertions (where the input did hold text) are kept.

## Verification

- `bash scripts/verify.sh` → **=== ALL PASS ===** (Relay / Lint / Format / TypeScript / terminology).
- `e2e/chat` against a **26.7.0** manager (`10.122.10.215:8090`), `--workers=1` — the configuration CI uses (`workers: process.env.CI ? 1 : undefined`): **28 passed, 5 skipped, 0 failed**. Baseline before this PR on the same target: 0 chat tests passing.
- Default local parallelism (4 workers) on that shared manager: 3 of 4 runs fully green. The remaining failures are app-boot timeouts on the shared 26.7.0 box (>120 s to reach an interactive pane under four concurrent logins, vs ~15 s single-worker), not assertion failures. `CHAT_READY_TIMEOUT_MS` is a readiness gate rather than a behavioural assertion, so it is set generously (120 s, inside Playwright's 180 s per-test timeout).
- The 5 remaining skips are pre-existing `test.fixme`s unrelated to this change (streaming-stop timing, history-label auto-update, history navigation, history deletion, first-load model-fetch error suppression).

### DIFF
--- a/e2e/chat/chat.spec.ts
+++ b/e2e/chat/chat.spec.ts
@@ -7,11 +7,9 @@ import {
   setupChatPage,
   setupChatPageWithTwoDeployments,
   chatGraphQLMocks,
-  setupChatPageWithUnavailableDeployment,
   makeSseResponse,
   modelsApiMockResponse,
   waitForChatReady,
-  CHAT_READY_TIMEOUT_MS,
   MOCK_MODEL_ID,
   MOCK_DEPLOYMENT_NAME,
   MOCK_DEPLOYMENT_NAME_B,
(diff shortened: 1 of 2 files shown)


agatha197:
This PR was auto-closed as "merged" by GitHub as a side effect of restacking (the PR's declared base still pointed at #8462's branch at the moment #8462 was rebased on top of this branch's new head, which made GitHub see this PR's head as already contained in its base). No code from this branch reached `main`. Continued as #8509, now correctly based on #8420 with #8462 stacked on top of it.

PR #8482, opened 2026-07-30
style(FR-3420): cap the docs site layout width on ultrawide monitors

### LINKED ISSUE #8481: docs site: cap the docs layout width so ultrawide monitors don't stretch the shell
## Problem

On ultrawide displays (21:9, e.g. 2560x1080 / 3440x1440) the generated documentation site's three-column shell (`.doc-page`: sider + article + right-rail TOC) stretches to the full viewport width. The sider pins to the far-left edge and the TOC to the far-right edge, while the 820px article column stays centered — leaving ~600px of dead space on each side **inside** the layout. The topbar brand and the version banner likewise cling to the screen corners.

## Proposal

Follow the pattern used by reference doc sites (e.g. https://seed-design.io/components/action-button, which caps its shell at 1552px and lets the leftover width become symmetric page margin):

- Add a `--bai-layout-max` shell cap token and apply `max-width` + `margin-inline: auto` to `.doc-page`.
- Keep the sticky topbar and version banner full-bleed (background/border still span the viewport) but inset their **content** to the same cap via `padding-inline: max(<gutter>, calc((100% - var(--bai-layout-max)) / 2))`.

## Acceptance criteria

- Below the cap (<=1536px) the layout is byte-for-byte visually unchanged — no regression for 1440p/laptop viewports.
- At >=1920px the shell is centered; sider, article and TOC read as one block.
- Topbar brand aligns with the sider's left edge; the action cluster aligns with the TOC's right edge.
- No horizontal overflow at any width; mobile drawer and dark mode unaffected.

Affects `packages/backend.ai-docs-toolkit/src/styles-web.ts` (shared by the website build and the HTML preview).

JIRA Issue: FR-3420

yomybaby:
Resolves #8481 (FR-3420)

The docs site's three-column shell (`.doc-page`: sider + article + right-rail TOC) stretched to the full viewport width. On a 21:9 monitor that pinned the sider to the far-left edge and the TOC to the far-right edge while the 820px article stayed centered — roughly 600px of dead space **on each side, inside the layout**, so the three columns stopped reading as one block.

Fixing that exposed follow-on problems in the rail, all addressed here. Five commits, one file of source (`packages/backend.ai-docs-toolkit/src/styles-web.ts`, shared by `build:web` / `serve:web` and `preview:html`) plus its tests.

All captures below are rendered from a real build of this branch, against a real build of the merge-base (3f1510a0f) — not injected CSS.

2560 light before/after

2560 dark before/after

rail detail before/after

other viewports before/after

---

## 1 · Cap and centre the shell

Same pattern as reference doc sites — [seed-design.io](https://seed-design.io/components/action-button) caps its shell at 1552px and lets the leftover width become symmetric page margin. Measured there: the grid resolves to `504px 240px 1112px 200px 504px` at 2560px, a fixed 1552px shell inside growing gutters.

`--bai-layout-max: 1536px` — 280 (sider) + 240 (TOC) + 1016 for the article column, so the 820px prose keeps ~98px of slack per side, the same column-to-prose ratio as the reference. `.doc-page` gets `max-width` + `margin-inline: auto`.

The topbar and version banner stay **full-bleed** — background and bottom border still span the viewport, since a sticky bar that stops short of the edges reads as a floating card — and only their content is inset. No HTML change was needed.

| viewport | before | after |
|---|---|---|
| 1440 | page `1440@0` · sider `280@0` · main `820@330` · toc `240@1200` | **unchanged** (columns) |
| 1920 | page `1920@0` · sider `280@0` · main `820@570` · toc `240@1680` | page `1536@192` · sider `280@192` · main `820@570` · toc `240@1488` |
| 2560 | page `2560@0` · sider `280@0` · main `820@890` · toc `240@2320` | page `1536@512` · sider `280@512` · main `820@890` · toc `240@1808` |
| 3440 | page `3440@0` · sider `280@0` · main `820@1330` · toc `240@3200` | page `1536@952` · sider `280@952` · main `820@1330` · toc `240@2248` |

The article column does not move at any width — only the sider and TOC come inward, so the reading position is unchanged and only the dead space closes.

## 2 · Hold the rail grid across the cap

The first pass wrote the bar inset as `max(20px, centering-gutter)`. That form drops the gutter below the cap and drops the **inset** above it, so the brand sat at `shell+20px` on a laptop and snapped to `shell+0px` past 1536px — the logo and the menu beneath it stopped sharing a vertical grid, and the grid moved as the window was resized.

The sider already had a stable grid: its VERSION label and every group icon sit at exactly `shell+24px`, at every width. Adopted that rather than inventing a second line, reached by adding the inset *on top of* the gutter:

```css
--bai-rail-inset: 24px;
padding-inline: calc(max(0px, (100% - var(--bai-layout-max)) / 2) + var(--bai-rail-inset));
```

Below the cap the gutter term is 0 and the inset alone applies; above it the gutter grows while the inset holds content on the grid.

| viewport | logo | nav / group icon | |
|---|---|---|---|
| ≤ 1536 | +20 | +24 | 4px off |
| ≥ 1920 | +0 | +24 | 24px off |
| **every width, after** | **+24** | **+24** | aligned |

Guide line drawn at `shell + 24px`. Nested nav items (`1. Quickstart` …) stay indented as child hierarchy.

| 1440px (below the cap) | 2560px (above it) |
|---|---|
| grid 1440 | grid 2560 |

This moves the brand 4px right below the cap (+20 → +24). That 4px was the pre-existing misalignment, so **sub-cap rendering is not pixel-identical to `main`** — the column geometry is untouched, but the logo now sits on the grid.

## 3 · Flatten the in-flow rail

Centering the shell also detached the sider from the viewport edge, exposing its background as a band with no boundary on its left — the tint started mid-air. Measured, `--bai-bg-sider` is only **3/255** from `--bai-bg` (`#FCFCFC` vs `#FFFFFF`; `#111114` vs `#0E0E10` in dark): too weak to read as a deliberate rail, too strong to read as nothing. Full-bleed hid the ambiguity against the screen edge.

So the rail drops its tint and the existing `border-right` (paired with `.doc-toc`'s `border-left`) carries the column separation — one continuous surface, the treatment Fumadocs / Tailwind / React docs use.

```css
.doc-sidebar { background: var(--bai-bg); }
```

The **≤880px drawer deliberately keeps `--bai-bg-sider`**: there the sider is a fixed overlay floating above the article with its own shadow, so it does want a surface distinct from the content beneath it. Verified it still resolves to `#FCFCFC` over a `#FFFFFF` body.

mobile drawer keeps its own surface

**Alternative considered and deferred.** Commit to the tint instead, and give the shell a real outer edge on a recessed canvas. It reads better as a designed system (the step grows to 5–8 levels), but it needs a new `--bai-bg-canvas` token — reusing `--bai-bg-muted` / `--bai-bg-subtle` **inverts the ramp in dark mode**, where both are *lighter* than `--bai-bg`, making the empty gutters float above the document — and it changes the background tone at every viewport, not just wide ones. That is a site-wide visual decision, so it is not riding along here. A rounded-card variant was also built and rejected: the corners and shadow only exist at scroll-top, since the document never ends.

## 4 · Inset the sider's version divider

The version selector was separated from the nav by a `border-bottom` on the row itself. A border sits *outside* the padding box, so it always spanned the sider edge to edge — a hard full-width line cutting the rail in two. The right rail does the equivalent job (separating the outline from the Get-help block) with `.doc-toc__divider`, which is inset by the TOC's own padding and reads noticeably softer.

Matched that treatment: drop the border, draw the rule as an `::after` inset by the row's own 24px horizontal padding.

divider before / after / reference

| | extent | left end | right end |
|---|---|---|---|
| before | 280px — the full sider | sider edge (+0) | sider edge |
| after | **231px** | **536px = the +24 rail grid** | version popup's right edge |
| TOC divider (reference) | 195px in a 240px rail | +23 | −22 |

Its left end now lands on the same +24px grid as the VERSION label, the group icons and the topbar brand. Same 1px height and `--bai-border-soft` token as the TOC divider, so only the extent changed — not the weight or colour.

A pseudo-element rather than a real `<div>` like the TOC's: the version row lives outside `.doc-sidebar__scroll`, so this needs no markup change and cannot disturb the scrollport's flex sizing. It carries no `z-index`, so the version popup (`z-index: 110`, opaque) still paints over it.

### Uniform rhythm around it

The same block also read top-heavy, because its three vertical gaps were all different:

| gap | before | after |
|---|---|---|
| sider top → switcher pill | 14px | **14px** |
| pill bottom → separator | 10px | **14px** |
| separator → first nav item | 16px | **14px** |

rhythm before/after

Measured box to box, matching `.doc-toc__divider`'s symmetric 18px margins. Each gap now comes from exactly **one** declaration, which is the part worth keeping: the separator-to-first-item gap used to be a *sum* of `.doc-sidebar__scroll`'s `padding-top` (10px) and `.doc-sidebar-intro`'s `margin-top` (6px), so changing either silently shifted the rhythm. The intro's top margin is zeroed and the scrollport owns that gap outright; both keep their horizontal values, which are load-bearing for the +24px rail grid.

An optically-compensated variant (14/14/**6**, subtracting the nav item's 8px inner padding so the whitespace either side of the rule reads equal *at rest*) was built and rejected — it puts INTRODUCTION's hover pill 6px from the rule, and the separator stops looking like it separates anything. 12/12/12 also reads fine but compresses the block for no gain.

## Verification

- **Surface + grid + overflow swept at 390 / 880 / 1024 / 1440 / 1536 / 1537 / 1920 / 2560 / 3440.** Sider background equals the body background at every width; logo, banner icon, VERSION label and both group icons land on +24 at every width where the sider is visible; `documentElement.scrollWidth === innerWidth` throughout. 1536→1537 is included specifically to cover the cap crossover.
- **Dark mode** verified for every change; **mobile drawer** verified to retain its own surface and `position: fixed`.
- `pnpm --filter backend.ai-docs-toolkit test` — **303 pass / 0 fail / 1 skipped** (pre-existing). Eight new tests in `src/styles-web.test.ts` cover the cap token, the centered `.doc-page`, the `--bai-rail-inset` token, the shape of the bars' `padding-inline`, the flat rail vs. the drawer's retained surface, the TOC's mirrored border, the version row's inset separator, and the uniform 14px rhythm (including the intro's zeroed top margin, so the gap cannot go back to being a sum).
- The tests assert on **comment-stripped** CSS. The stylesheet's comments name the very tokens under test, so matching raw output would let prose satisfy an assertion.
- The new assertions were **mutation-checked**: restoring the rail tint, dropping the TOC border, reintroducing `max(20px, gutter)`, and resetting the version divider to a full-bleed `0 / 0` inset each fail the suite. The `padding-inline` test checks the declaration's *shape* — the only `max()` permitted is the one flooring the gutter at `0px` — so the grid-shifting form cannot come back silently.

### DIFF
--- a/packages/backend.ai-docs-toolkit/src/styles-web.test.ts
+++ b/packages/backend.ai-docs-toolkit/src/styles-web.test.ts
@@ -6,6 +6,9 @@
  * Coverage:
  *   - The ultrawide shell cap: --bai-layout-max is defined and .doc-page is
  *     capped and centered by it.
+ *   - The flat rail: the in-flow sider shares the page surface (its old
+ *     3/255 tint became an unbounded band once the shell was centered),
+ *     while the mobile drawer overlay keeps a surface of its own.
  *   - The rail grid: the two full-bleed sticky bars (topbar, version banner)
  *     inset their content by centering-gutter + --bai-rail-inset, so the
  *     brand and the banner icon stay on the sider's grid line at every
@@ -17,6 +20,15 @@ import assert from "node:assert/strict";
 
 import { generateWebsiteStyles } from "./styles-web.js";
 
+/**
+ * The stylesheet is heavily commented, and those comments name the very
+ * tokens these tests assert on. Matching against raw CSS would let a
+ * comment satisfy an assertion — so every test works on stripped output.
+ */
+function styles(): string {
+  return generateWebsiteStyles().replace(/\/\*[\s\S]*?\*\//g, "");
+}
+
 /** Extract the body of the first rule whose selector matches exactly. */
 function ruleBody(css: string, selector: string): string {
   const idx = css.indexOf(`\n${selector} {`);
@@ -26,8 +38,47 @@ function ruleBody(css: string, selector: string): string {
   return css.slice(start + 1, end);
 }
 
+test("generateWebsiteStyles — the in-flow rail is flat, the drawer keeps its surface", () => {
+  const css = styles();
+
+  // The capped shell detaches the rail from the viewport edge, so a tint
+  // that is only 3/255 from --bai-bg reads as an unbounded band. The rail
+  // must therefore share the page surface and lean on its border instead.
+  const sider = ruleBody(css, ".doc-sidebar");
+  assert.match(
+    sider,
+    /background:\s*var\(--bai-bg\);/,
+    "the in-flow rail must use the page surface, not a near-invisible tint",
+  );
+  assert.ok(
+    !/--bai-bg-sider/.test(sider),
+    "the in-flow rail must not reintroduce --bai-bg-sider",
+  );
+  assert.match(
+    sider,
+    /border-right:\s*1px solid var\(--bai-border\);/,
+    "with the tint gone, the border is the only thing separating rail from article",
+  );
+
+  // ...but the ≤880px drawer is a fixed overlay above the article, so it
+  // does need a surface of its own. Guard that the flattening above was
+  // not applied to it wholesale.
+  assert.match(
+    css,
+    /body\.bai-drawer-open \.doc-sidebar \{[^}]*background:\s*var\(--bai-bg-sider\);/s,
+    "the mobile drawer overlay should keep its own surface",
+  );
+});
+
+test("generateWebsiteStyles — the TOC mirrors the rail's separator", () => {
+  // The two hairlines are now the whole separation story, so they have to
+  // stay symmetric; losing one leaves the shell visibly lopsided.
+  const toc = ruleBody(styles(), ".doc-toc");
+  assert.match(toc, /border-left:\s*1px solid var\(--bai-border\);/);
+});
+
 test("generateWebsiteStyles — defines the ultrawide shell cap token", () => {
-  const css = generateWebsiteStyles();
+  const css = styles();
   assert.match(
     css,
     /--bai-layout-max:\s*1536px;/,
@@ -36,7 +87,7 @@ test("generateWebsiteStyles — defines the ultrawide shell cap token", () => {
 });
 
 test("generateWebsiteStyles — .doc-page is capped and centered by the shell token", () => {
-  const body = ruleBody(generateWebsiteStyles(), ".doc-page");
+  const body = ruleBody(styles(), ".doc-page");
   assert.match(body, /max-width:\s*var\(--bai-layout-max\);/);
   assert.match(
     body,
@@ -46,7 +97,7 @@ test("generateWebsiteStyles — .doc-page is capped and centered by the shell to
 });
 
 test("generateWebsiteStyles — defines the rail grid inset", () => {
-  const css = generateWebsiteStyles();
+  const css = styles();
   assert.match(
     css,
     /--bai-rail-inset:\s*24px;/,
@@ -55,7 +106,7 @@ test("generateWebsiteStyles — defines the rail grid inset", () => {
 });
 
 test("generateWebsiteStyles — full-bleed bars hold the rail grid at every width", () => {
-  const css = generateWebsiteStyles();
+  const css = styles();
   // The bars stay full-bleed (background + border span the viewport) and only
   // their inline padding grows. The padding must be centering-gutter PLUS the
   // rail inset, so the brand sits at shell+inset on both sides of the cap.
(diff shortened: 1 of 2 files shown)


agatha197 (approved):
LGTM

PR #8483, opened 2026-07-30
fix(FR-3406): keep the RBAC role detail drawer working on managers below 26.8.0

### LINKED ISSUE #8452: RBAC page: opening the detail drawer from the list throws an error and breaks the screen
On the RBAC page, clicking an entry in the list to open its detail drawer raises an error and the screen breaks (content no longer renders).

**Steps to reproduce**

1. Open the RBAC page.
1. Click a row in the list to open the detail drawer.
1. An error is shown and the screen becomes unusable.

**Expected**

- The detail drawer opens and renders the selected entry. If the underlying request fails, the failure is contained to the drawer and surfaced as an error message rather than breaking the page.

**Environment**

- Manager version: `26.4.8`

**Notes**

- The exact error message / stack trace is not captured yet; adding it (browser console or the on-screen error) would pin down whether this is a GraphQL field mismatch or a render-time crash.
- Possibly analogous: FR-3045 (Model Store preset detail drawer broke on Manager 26.4.3 because preset fields were renamed/restructured) — a drawer breaking due to a WebUI/manager schema mismatch. Worth checking whether the RBAC drawer query requests fields this manager version does not serve.
- A GraphDev workspace link related to this investigation is attached as a web link.

JIRA Issue: FR-3406

nowgnuesLee:
Resolves #8452 (FR-3406)

## Problem

On the RBAC page, clicking a role in the list opened the detail drawer and **took the whole page down** — the list disappeared behind an error screen.

`ScopedRolePermissionCard` (introduced in FR-3271, #8191) filters the role's scopes server-side and always passed `scopeFilter: { scopeType: { equals: <type> }, ...scopeIdFilter }`. Both of those `EntityFilter` fields are **`Added in 26.8.0`**:

```graphql
input EntityFilter {
  entityType: RBACElementTypeFilter = null
  entityId: StringFilter = null
  """Added in 26.8.0. Filter by the type of scope the entity is registered in."""
  scopeType: RBACElementTypeFilter = null
  """Added in 26.8.0. Filter by the id of scope the entity is registered in."""
  scopeId: StringFilter = null
  ...
}
```

Against the reported Manager **26.4.8** `EntityFilter` has neither field, so the query fails variable coercion and `useLazyLoadQuery` throws — same class of schema-mismatch bug as FR-3045/FR-3031. `@since` cannot express this gate: the fields never appear in the query document; they arrive as a runtime **variable value**.

## Fix

**Version-branch the drawer view itself** (instead of patching a per-card filter fallback into the new view):

- **≥ 26.8.0** (`baiClient.supports('role-mapped-scope-filter')` — the existing 26.8.0 capability; no new flag added): the merged **Detailed Permissions** card view, completely unchanged.
- **< 26.8.0**: the **legacy separate Scopes / Permissions tabs** that FR-3271 removed, restored as explicitly named `Legacy*` components. Their queries only use pre-26.8.0 schema surface (`Role.scopes` 26.4.2, `adminPermissions`, `adminCreate/Update/DeletePermission`), so they work on every manager the RBAC page supports.

The branch point is a single `supportsDetailedPermissions` check in `RoleDetailDrawerContent`; the Assignments tab is shared by both views (`Role.users` is 26.3.0).

### Restored components (with modernization)

| Legacy (removed in FR-3271) | Restored as |
|---|---|
| `RoleScopeTab` | `LegacyRoleScopeTab` |
| `RolePermissionTab` | `LegacyRolePermissionTab` |
| `CreatePermissionModal` | `LegacyCreatePermissionModal` |

The tabs are **not** verbatim restores — they are rewritten to the current data-fetching conventions (same pattern as `ScopedRolePermissionCard`):

- `useBAIPaginationOptionState` + `useLazyLoadQuery` with `limit` / `offset` query variables — filtering, ordering, and pagination all run **server-side**; no client-side narrowing anywhere.
- `useDeferredValue` on variables/fetchKey so page/filter/refresh updates render inline instead of re-suspending the tab; `useFetchKey` + `BAIFetchKeyButton` for refresh and post-mutation refetch.
- The old `useRefetchableFragment` + manual `doRefetch` orchestration and the drawer-level query are gone — each tab issues its own query (`LegacyRolePermissionTabQuery` also feeds the modal's role-scope fragment, so the new-manager path pays zero extra cost).
- Legacy nuqs URL state converted to local state, matching the FR-3271 convention that nothing inside the drawer writes to the URL.

### Also in this diff

- **i18n**: restores the ten `rbac.*` keys the legacy view uses (`CreatePermission`, `EditPermission`, `RemovePermission`, `EntityType`, `Operation`, …) across all 21 locales, with their original translations.
- The earlier revisions of this PR added a `rbac-entity-scope-filter` capability and a `SCOPE_FETCH_LIMIT` client-side fallback inside `ScopedRolePermissionCard`; both are **fully reverted** — `ScopedRolePermissionCard` and `client.ts` now match `main`.

## Verification

**Automated** — `bash scripts/verify.sh`: `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Terminology). `react` unit tests: **1048 passed** (57 files).

### :warning: Needs a manual pass on a < 26.8.0 manager

The issue names `10.122.12.154` (Manager 26.4.8), which is unreachable from my environment, so the legacy path is verified by construction (pre-26.8.0 schema surface only), not by reproduction. Please confirm:

**Manager 26.4.8 (legacy view):**
- [ ] RBAC page → click a role → the detail drawer opens and renders; the list behind it stays intact
- [ ] The drawer shows **Scopes / Permissions / Assignments** tabs (not the card view)
- [ ] Scopes tab: filter by scope type, sort, and paginate — all server-side
- [ ] Permissions tab: list renders with resolved scope names; create / edit / delete a permission round-trips
- [ ] Assignments tab still loads, filters, sorts, and pages

**Manager ≥ 26.8.0 (unchanged view):**
- [ ] The drawer shows the merged Detailed Permissions card view exactly as on `main`


react/src/components/ScopedRolePermissionCard.tsx
```diff
@@ -56,6 +57,13 @@ import { graphql, useFragment, useLazyLoadQuery } from 'react-relay';
  */
 const PERMISSION_FETCH_LIMIT = 500;
 
+/**
+ * Upper bound of scope rows fetched per card when the server cannot filter
+ * `Role.scopes` by scope type (managers without `EntityFilter.scopeType`,
+ * < 26.8.0). Scopes beyond it are not shown on those managers.
+ */
+const SCOPE_FETCH_LIMIT = 500;
```
agatha197: `SCOPE_FETCH_LIMIT` is applied to an **unfiltered** fetch (`scopeFilter: null`) that every `ScopedRolePermissionCard` instance shares — it's a budget across *all* scope types combined, not 500 rows per type. If a role's total scope count (summed across every type) exceeds 500, whichever types don't land inside that leading 500-row window will compute `scopeCount === 0` here and their card will be hidden entirely (`if (!scopeIdFilter && scopeCount === 0) return null;` below), even though the role does hold scopes of that type. The comment above reads like a per-card cap ("Scopes beyond it are not shown"), which understates this — please call out the shared-budget behavior explicitly, and ideally detect truncation (`edges.length === SCOPE_FETCH_LIMIT`) so it doesn't silently render as zero scopes.

agatha197 (changes requested):
The fix correctly avoids sending `scopeType`/`scopeId` (only valid `>= 26.8.0`) when the manager doesn't support them — that resolves the reported crash.

However, `SCOPE_FETCH_LIMIT` is a **shared budget across all scope types**, not a per-card cap, and I think that needs to be addressed before merge (see inline comment).

Since `scopeFilter: null` is sent on the unsupported path, the query returns up to 500 scope rows for the role **across every scope type combined**, and each `ScopedRolePermissionCard` instance filters that same unfiltered window down to its own `scopeType` client-side. For a role whose total scope count (summed across all types) exceeds 500, whichever types don't fall inside that leading 500-row window will compute `scopeCount === 0` for their card — and since `if (!scopeIdFilter && scopeCount === 0) return null;`, the card for that type disappears entirely, even though the role genuinely holds scopes of that type. This isn't just "rows beyond the limit aren't shown" (as the comment on `SCOPE_FETCH_LIMIT` currently states) — it's that one type's rows can be crowded out of the window by another type's rows, silently hiding an entire card with no indication anything was truncated.

Could you either:
- scope-fetch more defensively (e.g. detect truncation via `edges.length === SCOPE_FETCH_LIMIT` and surface it, rather than silently rendering as if the role has zero scopes of that type), or
- at least update the doc comment to describe the shared-budget behavior accurately so this known limitation isn't understated for future readers.

Separately (non-blocking, style-only): since `scopeFilter` is already optional with a `null` default in the generated variables type, the fallback branch could omit the `scopeFilter` key entirely instead of setting it to `null` explicitly — same behavior either way, just flagging in case you prefer the terser form. Not required.

react/src/components/ScopedRolePermissionCard.tsx
nowgnuesLee: Good catch on the shared-budget framing — you're right that the previous wording read as a per-card row cap, and that the real failure mode is worse than "rows beyond the limit aren't shown": a type crowded out of the leading 500-row window resolves to `scopeCount === 0` and its card disappears entirely.

I went with the doc-comment option rather than truncation detection. Reasoning: this path only runs on managers below 26.8.0, and it additionally requires a single role to hold more than 500 scopes summed across every type. Both conditions together make it rare enough that I'd rather not add a detection branch plus user-facing copy (new i18n keys) that we can't exercise against a real dataset. Happy to revisit if a role that large shows up in practice.

Updated in 76b40ef1b — the comment now states that the fetch is unfiltered, that the bound is a budget shared by every scope type rather than a per-type cap, that exceeding it hides the whole card for types outside the leading window, and that this is accepted rather than detected.

Also took the non-blocking style note: leaving `scopeFilter: null` explicit in the fallback branch. Both branches of that spread then declare the same three keys, which keeps them symmetrical and makes it obvious at a glance that the fallback sends no filter.

Verified on the manager from the issue (10.122.12.8, 26.4.8): the old shape returns `Field "scopeType" is not defined by type "EntityFilter"`, and the fallback shape returns scopes and permissions normally.

ironAiken2 (approved):
LGTM

agatha197 (approved):
LGTM

PR #8484, opened 2026-07-30
fix(FR-3405): resolve the property filter's selected property from current props

### LINKED ISSUE #8451: Storage host project folder permissions: clearing the domain selector permanently empties the table and cannot be undone
## Problem

In **Storage Host Info → Project Folder Permissions → Domains**, clearing the domain selector permanently breaks the panel:

1. On open, the domain defaults to the current domain and the permission table shows its row.
1. Clicking the selector's clear (X) empties the selection — the table correctly falls back to **"Please select a domain."**
1. The selector then **re-displays the previous domain**, but the table stays empty.
1. Picking that same domain from the dropdown does **nothing** — the table never comes back.

End state: the selector shows a domain name while the table permanently shows the empty placeholder. Only re-opening the drawer recovers it.

## Reproduction

Reproduced end-to-end against a live manager (100% of runs). Panel state and the GraphQL traffic:

```
[initial]          rows=1  "Please select a domain."=false
[after clear +1s]  rows=0  "Please select a domain."=true
[after clear +4s]  rows=0  "Please select a domain."=true
[after reselect]   rows=0  "Please select a domain."=true
```

`ProjectFolderPermissionPanelQuery` calls fired during that sequence:

```
{"domainName":"default","skipDomain":false}   <- initial
{"domainName":null,"skipDomain":true}          <- after clear
                                               <- after re-selecting "default": NO query at all
```

No query is issued on re-selection, so `onChange` never fires and `selectedDomainName` stays undefined.

## Root cause

`react/src/components/ProjectFolderPermissionPanel.tsx` binds the selector directly to component state and passes `undefined` when empty:

```
const [selectedDomainName, setSelectedDomainName] = useState<string | undefined>(currentDomain);
...
<BAIDomainSelect
  value={selectedDomainName}
  onChange={(value) => setSelectedDomainName((value as string | undefined) || undefined)}
  allowClear
/>
```

antd `Select` treats `value={undefined`} as **uncontrolled** and falls back to its own internal state. The sequence is:

1. Clear → antd fires `onChange(undefined)` → parent state becomes `undefined`.
1. Re-render passes `value={undefined`} → antd stops honouring the prop and reverts to its internal value, which is still the previously selected domain. That is why the selector visibly re-fills itself while the parent state is empty.
1. Selecting that same option is a no-op from antd's point of view — its internal value already equals it — so `onChange` is never fired, the parent state is never restored, and the query stays skipped (`skipDomain: true`).

The panel's query is gated on that state:

```
const queryVariables = {
  domainName: selectedDomainName ?? null,
  skipDomain: !selectedDomainName,
};
```

so the table renders `DomainStoragePermissionTable`'s `storageHost.permission.NoDomainSelected` placeholder forever.

## This is already documented in the component being used

The selector is supplied through `BAIGraphQLPropertyFilter`'s `renderInput`, whose docblock in `packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx` explicitly warns about this:

> Pass `value={null`} to the control (antd's controlled-empty value, since `value={undefined`} …)

The call site does not follow it. It also ignores the `onAddCondition` callback that `renderInput` provides, wiring the control to its own state instead — worth reviewing as part of the fix, since the filter's condition state and the panel's state are currently two unsynchronised sources of truth.

## Proposed change

Pass an explicit empty value so the control stays controlled:

```
value={selectedDomainName ?? null}
```

and normalise the `onChange` handler so the cleared state is stored consistently. Then verify that re-selecting the same domain restores the table, and consider whether the selector should go through `onAddCondition` as the filter intends.

Observed on webui v26.8.0-rc.0 (`react/src/components/ProjectFolderPermissionPanel.tsx` is unchanged on `main`).

JIRA Issue: FR-3405

agatha197:
Resolves #8451 (FR-3405)

## Problem

In **Storage Host Info → Project Folder Permissions → Domains**, clearing the domain selector permanently breaks the panel: the table empties, the selector re-fills itself with the previous domain, and re-picking that domain does nothing. Only re-opening the drawer recovers.

## Root cause

`ProjectFolderPermissionPanel` used `BAIGraphQLPropertyFilter` against its `renderInput` contract. The contract expects a **stateless** control (`value={null}`) that commits picks through `onAddCondition`; the panel instead bound the control to its own state and ignored the callback:

```tsx
renderInput: () => (
  <BAIDomainSelect value={selectedDomainName} onChange={...} allowClear />
)
```

`BAIGraphQLPropertyFilter` seeds the selected property object into state once (`useState(filterProperties[0])`) and never re-syncs it with `filterProperties`, so it keeps rendering the **first render's** `renderInput` closure. That frozen closure always supplies the mount-time `selectedDomainName`:

1. Clear → `onChange(undefined)` fires (state setters are stable) → panel state empties → the query is skipped → the table shows the "select a domain" placeholder.
2. The stale closure re-renders the Select with the **original** domain, so the selector visibly refills itself while panel state is empty.
3. Re-picking that domain is a no-op for antd — its controlled value already equals it — so `onChange` never fires, panel state is never restored, and the query stays skipped.

Instrumented trace on an unfixed build, replaying the exact sequence:

```
[render]   selected="default"  deferred={"domainName":"default","skipDomain":false}
### clear
[onChange] undefined
[render]   selected=undefined  deferred={"domainName":null,"skipDomain":true}
### re-select "default"
(no onChange, no render at all)
rows: 0
```

## Fix

Fix the call site: stop using `BAIGraphQLPropertyFilter` here. This panel never used it as a filter — a single property with `fixedOperator`, `onAddCondition` never called, so no condition tag could ever appear. The visible UI was only the `Space.Compact` pair the filter happened to render: [property select "Name"] + [domain select].

Render that pair directly, keeping the UI identical:

```tsx
<Space.Compact>
  <BAISelect
    options={[{ label: t('storageHost.permission.Name'), value: 'domainName' }]}
    value="domainName"
  />
  <BAIDomainSelect
    value={selectedDomainName ?? null}
    onChange={...}
    allowClear
  />
</Space.Compact>
```

- With no `renderInput` indirection, the domain select re-renders from the panel's current state every render — the frozen-closure path no longer exists.
- `value={selectedDomainName ?? null}` keeps the antd Select controlled when empty. `value={undefined}` flips antd to uncontrolled, which would independently reproduce the "selector refills itself / re-select is a no-op" symptom.
- The fetch-key button spinner and the projects table's loading state are additionally tied to in-flight deferred query variables, so a domain change shows as loading on both tables instead of only the domain table.

`BAIGraphQLPropertyFilter` itself is untouched. An earlier revision of this PR fixed the shared component instead (tracking the selected property by key and resolving the object from current `filterProperties` each render). It was reverted in favor of the call-site change: this panel's control is a plain domain picker driving local query state — not a GraphQL filter — so the real defect was using `renderInput` against its stateless contract. Call sites that follow the contract (e.g. `UserFolderPermissionPanelV2`, migrated in FR-3282) are unaffected by the frozen-object behavior, because their controls are stateless and `onAddCondition` is supplied at call time.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup paths, Terminology)
- The clear → re-select scenario re-run against a live manager for this revision: pending (the earlier revision's live-manager trace above covered the shared-component fix, not this call-site fix).

### DIFF
--- a/react/src/components/ProjectFolderPermissionPanel.tsx
+++ b/react/src/components/ProjectFolderPermissionPanel.tsx
@@ -9,14 +9,14 @@ import { useCurrentDomainValue } from '../hooks';
 import DomainStoragePermissionTable from './DomainStoragePermissionTable';
 import ProjectStoragePermissionTable from './ProjectStoragePermissionTable';
 import { CheckCircleOutlined } from '@ant-design/icons';
-import { Typography, theme } from 'antd';
+import { Space, Typography, theme } from 'antd';
 import {
   BAIAlert,
   BAICard,
   BAIDomainSelect,
   BAIFetchKeyButton,
   BAIFlex,
-  BAIGraphQLPropertyFilter,
+  BAISelect,
   useFetchKey,
 } from 'backend.ai-ui';
 import React, { useDeferredValue, useState } from 'react';
@@ -114,32 +114,41 @@ const ProjectFolderPermissionPanel: React.FC<
       >
         <BAIFlex direction="column" align="stretch" gap="xs">
           <BAIFlex align="center" justify="between" gap="md" wrap="wrap">
-            <BAIGraphQLPropertyFilter
-              filterProperties={[
-                {
-                  key: 'domainName',
-                  propertyLabel: t('storageHost.permission.Name'),
-                  type: 'uuid',
-                  fixedOperator: 'equals',
-                  renderInput: () => (
-                    <BAIDomainSelect
-                      value={selectedDomainName}
-                      onChange={(value) =>
-                        setSelectedDomainName(
-                          (value as string | undefined) || undefined,
-                        )
-                      }
-                      allowClear
-                      style={{ minWidth: 200 }}
-                    />
-                  ),
-                },
-              ]}
-            />
+            {/* The domain picker drives this panel's own query state rather
+                than a GraphQL filter, so it is a plain compact pair (label
+                select + domain select) instead of BAIGraphQLPropertyFilter,
+                whose renderInput contract expects stateless controls committing
+                via onAddCondition (FR-3405). */}
+            <Space.Compact>
+              <BAISelect
+                popupMatchSelectWidth={false}
+                options={[
+                  {
+                    label: t('storageHost.permission.Name'),
+                    value: 'domainName',
+                  },
+                ]}
+                value="domainName"
+                style={{ minWidth: 150 }}
+              />
+              <BAIDomainSelect
+                value={selectedDomainName ?? null}
+                onChange={(value) =>
+                  setSelectedDomainName(
+                    (value as string | undefined) || undefined,
+                  )
+                }
+                allowClear
+                style={{ minWidth: 200 }}
+              />
+            </Space.Compact>
             <BAIFetchKeyButton
               value={domainFetchKey}
               onChange={updateDomainFetchKey}
-              loading={deferredFetchKey !== domainFetchKey}
+              loading={
+                deferredFetchKey !== domainFetchKey ||
+                deferredQueryVariables !== queryVariables
+              }
             />
           </BAIFlex>
           <DomainStoragePermissionTable
@@ -181,6 +190,10 @@ const ProjectFolderPermissionPanel: React.FC<
           storageVolumeFrgmt={storageVolume}
           domainFrgmt={domain}
           permissionFrgmt={vfolder_host_permissions}
+          loading={
+            deferredFetchKey !== domainFetchKey ||
+            deferredQueryVariables !== queryVariables
+          }
         />
       </BAICard>
     </BAIFlex>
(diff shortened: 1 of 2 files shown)


ironAiken2:
* **#8484** <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8484?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lablup/backend.ai-webui/8484?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/lablup?ref=merge-queue-instructions-in-stack-comment&prId=8484).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

ironAiken2 (approved):
LGTM

PR #8493, opened 2026-07-30
fix(FR-3424): allow user assignment on system RBAC roles for pre-26.8.0 managers

### LINKED ISSUE #8492: Re-enable user assignment on system-generated RBAC roles for 26.4.x LTS managers
## Background

FR-3320 replaced the assignment table on the RBAC role detail drawer's Assignments tab with a warning alert (`rbac.SystemRoleNoAssignments`) for SYSTEM-source roles, on the assumption that system-generated roles are managed by the backend.

However, on **26.4.x LTS managers** this leaves no way to assign a project admin at all:

- Granting project admin requires assigning the user to the project's system-generated admin role.
- The one-click project-admin assignment button on the project page exists only on newer managers and is unavailable on 26.4.x due to the version gap.
- With the WebUI also blocking direct assignment on system roles, the workflow is completely dead-ended on 26.4.x.

## What to do

- Re-enable user assignment/unassignment on SYSTEM-source roles in the role detail drawer's Assignments tab, so admins on 26.4.x LTS can assign users to a project's system-generated admin role.
- Decide whether to lift the restriction unconditionally or gate it by manager version/capability (block only when the manager provides the one-click project-admin flow). Version-gating is preferred if the capability can be detected reliably.
- Keep (or adjust) the informational notice that the role itself is system-managed, while allowing the assignment actions.

## Notes

- Related: FR-3320 (introduced the block), FR-3317 (project admin setting modal).
- Target: 26.4.x LTS branch support.

JIRA Issue: FR-3424

ironAiken2:
Resolves #8492 (FR-3424)

## Background

FR-3320 replaced the assignment table on the RBAC role detail drawer's Assignments tab with a warning alert for **all** SYSTEM-source roles, on the assumption that the one-click project-admin setting on the project page covers the assignment flow. That flow is gated by `baiClient.supports('role-mapped-scope-filter')` (manager >= 26.8.0), so on 26.4.x LTS managers there was **no way to grant project admin at all** — the project-page button is hidden and the drawer blocks direct assignment.

## Changes

- `RoleAssignmentTab.tsx`: always render the assignment table; compute a read-only mode instead of hiding the table behind an alert. Read-only applies only when **all** of the following hold:
  - the role is SYSTEM-source,
  - it is registered to a PROJECT scope,
  - its name contains `admin` (case-insensitive),
  - the manager supports the project-page one-click flow (`role-mapped-scope-filter`, >= 26.8.0).
- In read-only mode the table stays visible (filter / sort / pagination / refresh all work) but the assign button, row checkboxes, and revoke actions are hidden, with the existing `rbac.SystemRoleNoAssignments` alert shown as a banner above the table.
- All other system roles — and every role on pre-26.8.0 managers — keep the fully editable assignment table, restoring the pre-FR-3320 assignment path on 26.4.x LTS.
- `RoleDetailDrawerContent.tsx`: drop the alert-vs-table branch; the tab now owns the decision.
- Fragment gains `name` / `source` fields (Relay artifacts regenerated).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Terminology)

## Stack

- Stacked on #8483 (FR-3406) — RBAC role detail drawer fix for managers below 26.8.0


react/src/components/RoleAssignmentTab.tsx
```diff
@@ -209,6 +225,14 @@ const RoleAssignmentTab: React.FC<RoleAssignmentTabProps> = ({
 
   return (
     <>
+      {isReadOnly && (
+        <BAIAlert
+          type="warning"
+          showIcon
+          description={t('rbac.SystemRoleNoAssignments')}
+          style={{ marginBottom: 12 }}
```
agatha197: please use token value

react/src/components/RoleAssignmentTab.tsx
```diff
@@ -137,6 +142,17 @@ const RoleAssignmentTab: React.FC<RoleAssignmentTabProps> = ({
...
 
+  // System-generated project admin roles are managed through the project
+  // page's one-click admin setting, which requires manager >= 26.8.0
+  // (role-mapped-scope-filter). Show their assignments read-only there; on
+  // older managers direct assignment here is the only way to grant project
+  // admin, so keep the actions available (FR-3424).
+  const isReadOnly =
+    data.source === 'SYSTEM' &&
+    !!projectScopeId &&
+    !!data.name?.toLowerCase().includes('admin') &&
```
agatha197, line 152: Is this the only way to check if someone is an administrator? Are there other possible methods, like using scopes?

agatha197 (changes requested):

react/src/components/RoleAssignmentTab.tsx
ironAiken2: Yes, at present, the only way to determine whether a user has project admin permissions is to check the ‘System Source Role’ and see if there are any projects linked via ‘Scope’, and whether the role name contains the word ‘admin’.

react/src/components/RoleAssignmentTab.tsx
```diff
@@ -208,14 +223,15 @@ const RoleAssignmentTab: React.FC<RoleAssignmentTabProps> = ({
...
-      <BAIFlex
-        justify="between"
-        align="start"
-        gap="sm"
-        wrap="wrap"
-        style={{ marginBottom: 12 }}
-      >
+    <BAIFlex align="stretch" direction="column" gap="sm">
+      {isReadOnly && (
+        <Alert
```
agatha197, line 228: Please use `BAIAlert`

agatha197 (approved):
LGTM

react/src/components/RoleAssignmentTab.tsx
ironAiken2: As BAIAlert does not support ‘narrow size’, I used Alert instead.

nowgnuesLee (approved):
LGTM

PR #8494, opened 2026-07-30
feat(FR-3423): disable the deploy row action on the administrator folder page

### LINKED ISSUE #8489: Disable the deploy row action on the administrator folder page
## Parent

FR-3407 — Remove header project selector on super-admin pages via an explicit project prop contract.

## What to build

On the administrator folder page, the "Deploy as service" row action must be disabled with a tooltip explaining that deployments are created from within the project, not from the admin menu. Elsewhere (the user folder page, the model store) it keeps working exactly as today.

The admin folder page is an oversight surface: it lists folders across every project so an administrator can see what exists. Creating a model service from it is out of scope for that purpose, and a deployment is inherently project-scoped — it belongs to a project, consumes that project's resources, and appears only in that project's deployment list. An administrator acting on a folder they merely happen to be able to see would create an endpoint in a project they may not belong to and therefore cannot observe or clean up afterwards.

This also removes an inconsistency introduced earlier in the epic: session-launching actions (file browser, SFTP) are already disabled on the admin folder page with an explanatory tooltip, while deploying — which creates a longer-lived workload than either — was still allowed.

The component must not infer the admin context for itself. Following the pattern already established for the file browser and SFTP buttons, the page states the constraint and passes the reason down; the row-action component only renders what it is told.

## Acceptance criteria

- [ ] On the administrator folder page the deploy row action is visibly disabled and its tooltip explains that deployments are created from the project
- [ ] The user folder page and the model store keep the action fully functional, unchanged
- [ ] The reason is supplied by the page, not derived inside the row-action component (no pathname check, no inferring from a null project)
- [ ] Copy is properly written and present in all locale files
- [ ] Tests cover both states
- [ ] Verification harness passes

## Blocked by

- FR-3410 (introduced the deploy action's project derivation and the pass-through prop this builds on).

JIRA Issue: FR-3423

yomybaby:
Resolves #8489 (FR-3423)

## Summary

- Deployments are project-scoped, but the administrator folder page (`AdminVFolderNodeListPage`, `/admin/data`) is an oversight surface listing folders across every project. Deploying a model from there would create an endpoint in a project the admin may not belong to and cannot afterwards see or clean up.
- The "Deploy as service" row action is now **disabled with an explanatory tooltip** on that page instead of hidden, mirroring the file-browser/SFTP disabled-with-tooltip treatment already applied there in this epic (FR-3412).
- `VFolderNodes` gains an optional `noDeployTooltip?: string` prop (forwarded to the row action's `disabled` / `disabledReason`). The component never infers the admin context itself — no pathname check, no inferring from `project` being `null` (which the user-facing data page can also pass legitimately). Only `AdminVFolderNodeListPage` passes the prop; the user Data page and the model store are unaffected.
- `VFolderNodesV2` gets the identical `noDeployTooltip` prop for API parity, since it defines the same `start-service` action — but no current caller passes it (`ProjectAdminDataPage`, its only consumer, is always project-scoped and out of this epic's scope).

## Copy

New i18n key `data.folders.CannotDeployFromAdminMenu`, added to all 21 locale files in correct alphabetical position:

> Deployments are created from within the folder's project, not the admin menu.

Tells the admin *where* to go (the folder's own project) rather than reading as a bare permission error, matching the tone of the neighbouring `data.CannotLaunchSessionInAdminMenu` string from the same epic.

## Tests

New `react/src/components/VFolderNodes.test.tsx` — a Relay-mocked render of `VFolderNodes` covering both states of the `noDeployTooltip` prop:
- **set** → the deploy action renders disabled, hovering shows the tooltip text, and clicking does not fire the handler.
- **absent** → the deploy action renders enabled and clicking it fires the handler (asserted via a mocked `VFolderDeployModal` receiving `open=true` with the correct vfolder id).

No existing test exercised the admin folder page's actions, so nothing else needed reconciling.

## Verification

- `cd react && npx vitest run` → 1282/1283 passed; the one failure (`DeploymentSettingModal.test.tsx`) is a pre-existing, documented load-dependent flake unrelated to this change — confirmed passing in isolation.
- `npx prettier --check 'resources/i18n/*.json'` → clean.


react/src/components/VFolderNodes.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+      noDeployTooltip={noDeployTooltip}
+    />
+  );
+};
+
+// `BAINameActionCell` keeps row actions as visible buttons (with their
+// disabled-reason tooltip) only when its `ResizeObserver` reports enough
+// container width; the overflow "More" menu it falls back to otherwise
+// does not carry the tooltip at all. jsdom's real layout always reports 0,
+// and the shared `resizeObserver.mock.js` stub never invokes its callback,
```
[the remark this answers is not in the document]
yomybaby: Fixed in `1526aab6c`, both parts.

- `BAINameActionCell`'s overflow menu item now wraps its label in the same tooltip the icon button uses when the action is disabled and carries a `disabledReason`. antd only sets `pointer-events: none` on `a` descendants of a disabled item (`menu/style/index.js`), so a `span` still receives hover.
- The test's `ResizeObserver` width is now per-test, and a new case renders at zero width — asserting the action appears only in the More menu and that its reason is still reachable there. It fails with exactly "Unable to find … CannotDeployFromAdminMenu" if the label wrapping is removed.

packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx
```diff
@@ -272,7 +272,17 @@ const BAINameActionCell: React.FC<BAINameActionCellProps> = ({
...
+    // antd does not fire hover events on disabled items, so the tooltip wraps
+    // the label instead of the item.
+    label:
+      action.disabled && action.disabledReason ? (
+        <Tooltip title={action.disabledReason} placement="left">
+          <span>{action.title}</span>
+        </Tooltip>
+      ) : (
+        action.title
+      ),
```
nowgnuesLee: Could you add a story covering this case?

resources/i18n/en.json
```diff
@@ -842,6 +842,7 @@
       "AdminDataPageAlert": "Only project-owned folders can be created on the Admin Data Page.",
       "CannotDeleteFolder": "Cannot delete folder mounted in one or more sessions. Please terminate the session first.",
       "CannotDeletePipelineFolder": "You can't delete pipeline folders.",
+      "CannotDeployFromAdminMenu": "Deployments are created from a project's Data page; the admin menu does not support it.",
```
nowgnuesLee: Might read a bit more naturally in active voice here.

nowgnuesLee (dismissed):

packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx
yomybaby: Added — `DisabledActionInOverflowMenu` in `BAINameActionCell.stories.tsx`, with the same width slider as `ResponsiveOverflow` so the visible icon button and the collapsed menu row can be compared directly.

One thing the story makes visible that is worth flagging: the two states no longer explain themselves the same way. Astryx types `DropdownMenuItemData.label` as a plain `string`, and `DropdownMenuItem`'s `description` slot is only reachable through the compound render path — which `items` disables outright. So where the antd version wrapped the menu label in a `Tooltip`, the reason is now folded into the label text instead. Same information, different shape; the story is where that difference is documented.

resources/i18n/en.json
yomybaby: Agreed — reworded:

> Create deployments from a project's Data page — the admin menu does not support it.

The other locales keep the same meaning, so only `en` changed.

PR #8498, opened 2026-07-30
fix(FR-3425): stop the scheduling-history expand mode from filtering the sub-step table

### LINKED ISSUE #8497: Scheduling history: expanding an all-success row in "Expand errors only" mode shows an empty sub-step table
## Summary

In the scheduling history modals, a row whose sub-steps are **all** `SUCCESS` still renders an expand (`+`) icon while the "Expand errors only" view mode is active. Expanding it shows a sub-step table with headers but no rows — only antd's "No data" placeholder. The UI advertises expandable content and then delivers nothing.

Introduced by FR-3008 (#7650), which added the three-mode expand toggle. Affects all three scheduling history modals (session / route / deployment).

## Steps to reproduce

1. Open a session that completed successfully, and open the **Session Scheduling History** modal.
1. Leave the expand mode at its default, **Expand errors only** (`errors-only`).
1. Click the `+` icon on any row whose result is `SUCCESS` (e.g. `promote-to-terminated`).
1. The expanded area shows the sub-step table header (Sub-step / Result / Error code / Message / Started at / Ended at) and a "No data" placeholder.

Expected: expanding a row shows its sub-steps. Observed: the sub-step table is empty.

## Root cause

Two places disagree about which sub-step list is authoritative.

**1. The expand affordance is computed from the unfiltered sub-steps.**

`packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryTable.tsx:67-68`

```
rowExpandable: (record: SchedulingHistoryNodeInList) =>
  !_.isEmpty(dataSource.find((h) => h.id === record.id)?.subSteps),
```

**2. The expanded content is rendered from the filtered sub-steps.**

`BAISchedulingHistoryTable.tsx:76` passes the current view mode down:

```
errorsOnly={mode === 'errors-only'}
```

and `packages/backend.ai-ui/src/components/fragments/BAISubStepNodes.tsx:184-186` drops every successful sub-step:

```
const dataSource = filterOutNullAndUndefined(subSteps)
  .filter((subStep) => !errorsOnly || subStep.result !== 'SUCCESS')
  .reverse();
```

So for a row whose sub-steps are all `SUCCESS`, `rowExpandable` is `true` (raw list is non-empty) but the rendered list is empty.

The default mode is `errors-only` (`packages/backend.ai-ui/src/hooks/useSchedulingHistoryExpandable.tsx:28-29`) and it is persisted per user via `useBAISettingUserState` (`react/src/components/SessionSchedulingHistoryModal.tsx:42`), so this is the out-of-the-box state, not an edge case.

Note that `errors-only` does not **auto**-expand success rows — `computeExpandedRowKeysForMode` only opens rows whose result is not `SUCCESS`. The empty table therefore appears only on a **manual** expand, which makes it worse: the user explicitly asked to see the sub-steps and got nothing.

The same `rowExpandable` / `errorsOnly` pair exists at lines 67-76 of both sibling components, so all three are affected:

- `packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryTable.tsx`
- `packages/backend.ai-ui/src/components/fragments/BAIRouteSchedulingHistoryTable.tsx`
- `packages/backend.ai-ui/src/components/fragments/BAIDeploymentSchedulingHistoryTable.tsx`

## Suggested fix

Three options; they are mutually exclusive.

1. **Make `rowExpandable` agree with the filter** — hide the expand icon when the errors-only filter would leave nothing to show.
1. **Scope `errorsOnly` to auto-expansion only** — use the mode to decide which rows open by default, but always render the full sub-step list once a row is expanded. **(recommended)**
1. **Render an explicit empty state** — keep the filter but replace "No data" with something like "No failed sub-steps".

Option 2 is recommended: the intent of `errors-only` is "automatically open the rows that failed", not "hide successful sub-steps". A row the user deliberately expanded should never come up empty, in any mode.

## Environment

- Found while reviewing PR #8488 (FR-3395). Unrelated to that change — FR-3395 only touches expanded-row background colors.
- Reproduced in dark theme; the filtering logic is theme-independent.

JIRA Issue: FR-3425

nowgnuesLee:
Resolves #8497 (FR-3425)

## Summary

In the scheduling history modals, expanding a row whose sub-steps all succeeded showed the sub-step table header and antd's "No data" placeholder instead of the sub-steps.

Found while reviewing #8488 (FR-3395). Unrelated to that change — FR-3395 only touches expanded-row background colors.

## Root cause

The "expand errors only" view mode was doing two jobs.

Its actual job is deciding **which rows start open**: `useSchedulingHistoryExpandable` computes `expandedRowKeys` from the mode, opening non-success rows and leaving successful ones collapsed.

But FR-3008 (#7650) also wired it to a **content filter** on the nested table:

```tsx
// BAISchedulingHistoryTable.tsx
errorsOnly={mode === 'errors-only'}
// BAISubStepNodes.tsx
.filter((subStep) => !errorsOnly || subStep.result !== 'SUCCESS')
```

The expand affordance was never part of that second job — `rowExpandable` offers an icon whenever a row has sub-steps, regardless of mode. So for a row whose sub-steps all succeeded the two disagreed: the icon said "there is something here", and the filter then removed all of it.

Since `errors-only` is the default mode and is persisted per user via `useBAISettingUserState`, this was the out-of-the-box state.

## Fix

Stop passing the mode into `errorsOnly`. The mode goes back to deciding only which rows start open; expanding a row always shows its full sub-step list.

Three lines across the three table components (session / route / deployment) — nothing else changes:

```diff
-  const { mode, expandedRowKeys, onExpandedRowsChange, expandColumnTitle } =
+  const { expandedRowKeys, onExpandedRowsChange, expandColumnTitle } =
...
-            errorsOnly={mode === 'errors-only'}
```

`rowExpandable`, the hook, and the GraphQL fragments are untouched — every row with sub-steps stays expandable in all three modes, exactly as before.

`BAISubStepNodes`'s `errorsOnly` prop goes with it. Nothing passed it once the mode stopped driving it — it existed only to serve that wiring.

That retires half of what FR-3008 built: alongside auto-expanding non-success rows, it also filtered the nested table to failing sub-steps. The filter applied per row regardless of *why* the row was open, which is what emptied a manually-expanded all-success row. Showing only failing sub-steps inside a failed row is still a reasonable idea, but it needs to tell auto-expanded rows from manually-expanded ones — a separate change, not a prop nobody calls.

The hook's `mode` doc is corrected too; it still told callers to use the mode to filter the nested sub-step table, which would have guided the next caller straight back into this bug.

### Approach note

An earlier revision of this branch fixed the disagreement from the other side, by making `rowExpandable` mode-aware so an all-success row offered no expand icon in `errors-only`. That was rejected on review: the mode means "which rows are open by default" and nothing more, so it must not remove the ability to open a row. The commits implementing that approach are still in the branch history and are fully reverted by the final one.

## Test changes

Three e2e tests were switching to "Collapse all" before expanding, because the default mode used to empty the nested table. That workaround is gone, so those tests now exercise the default `errors-only` mode directly — which is precisely the scenario that used to fail. The mock data set is all `SUCCESS` with sub-steps only on the `schedule-sessions` row, so no row auto-expands and the row is reached by a manual expand.

## Verification

- `bash scripts/verify.sh` → **ALL PASS** (Relay, Lint, Format, TypeScript, Vite warmup paths, Terminology).
- `pnpm --filter backend.ai-ui exec vitest run` → 367 passed, 1 skipped.
- `pnpm exec playwright test e2e/session/session-scheduling-history-modal.spec.ts --grep "sub-step"` against a live cluster → **4 passed** (the three edited tests plus the full open/expand/refresh/close flow).

The remaining tests in that spec were run earlier in the session and pass; two of them are intermittently flaky in `beforeEach` while waiting for the session-list table to render, before the scheduling-history modal is opened. That flakiness is unrelated to this change and predates it.

https://claude.ai/code/session_01WW8efBNZVc5Tp9BLB5RZsV


agatha197 (approved):
LGTM

PR #8499, opened 2026-07-30
fix(FR-3396): stretch SSH/SFTP connection modal so values render on one line

### LINKED ISSUE #8439: SSH/SFTP connection modal: Connection Info values render one character per line
In the **SSH / SFTP 연결** modal, the values under the **연결 정보** (Connection Info) section are squeezed into an extremely narrow column and wrap one character per line, while the modal itself is wide and most of its horizontal space stays empty.

**Observed (dark theme)**

- `사용자` renders as `w` / `o` / `r` / `k` stacked vertically.
- `호스트` renders as `1` / `2` / `7` / `.` / `0` ... one character per line.
- `포트` renders the same way, one digit per line.
- The label column is fine; only the value column collapses. The section is centered with large empty margins on both sides, so the modal width is not being used.

**Expected**

- Each Connection Info value renders on a single line, using the modal's available width.

**Environment**

- Manager version: `26.4.8`

**Notes**

- Possibly related but distinct: FR-3359 (same modal showing `127.0.0.1`/`undefined` instead of the real AppProxy TCP endpoint) is about the **values** being wrong; this issue is about the **layout** of those values.

JIRA Issue: FR-3396

agatha197:
Resolves #8439 (FR-3396)

## Summary
- In the SSH/SFTP connection modal, the "Connection Info" section's outer `BAIFlex` used the default `align="center"`, so its block-level children (including the non-bordered `Descriptions` showing user/host/port) sized to fit-content on the cross axis instead of stretching to the modal's actual width.
- Combined with antd's fixed table layout for `Descriptions` and the content column's `min-width: 1em` CSS rule, the value column collapsed to ~14px, wrapping each value one character per line (e.g. `work` → `w`/`o`/`r`/`k`, `127.0.0.1` → one digit per line).
- Fix: add `align="stretch"` to the outer `BAIFlex`, matching the sibling `VNCConnectionInfoModal` / `XRDPConnectionInfoModal`, which already use `align="stretch"` and don't exhibit this bug.

## Verification
- Reproduced live against `10.122.12.8:8090` with UI language set to Korean (사용자/호스트/포트 labels), confirmed via screenshot before/after.

## Test plan
- [x] Open a running session's SSH/SFTP connection modal with Korean UI language selected
- [x] Confirm 사용자/호스트/포트 values render on a single line, using the modal's available width
- [x] `scripts/verify.sh` passes

image

image

### DIFF
--- a/react/src/components/ComputeSessionNodeItems/SFTPConnectionInfoModal.tsx
+++ b/react/src/components/ComputeSessionNodeItems/SFTPConnectionInfoModal.tsx
@@ -117,7 +117,12 @@ const SFTPConnectionInfoModal: React.FC<SFTPConnectionInfoModalProps> = ({
       okText={t('session.appLauncher.DownloadSSHKey')}
       onOk={readAndDownloadSSHKey}
     >
-      <BAIFlex className={styles.description} direction="column" gap="md">
+      <BAIFlex
+        className={styles.description}
+        direction="column"
+        align="stretch"
+        gap="md"
+      >
         <Alert
           showIcon
           type="info"


nowgnuesLee (approved):
LGTM

PR #8505, opened 2026-07-31
fix(FR-3428): rewrite URL project segment in SwitchToProjectButton

### LINKED ISSUE #8504: Clicking "Switch to project" on a cross-project deployment does not switch the header project (URL keeps the old project)
## Symptom

On `DeploymentDetailPage`, when a deployment belongs to a different project, a warning alert appears with a **Switch to '{project}'** button (`SwitchToProjectButton`). Clicking it leaves the app in an inconsistent, half-switched state:

- The URL stays at `/project/<old-project>/deployments/<id>` — the `:projectName` segment is not rewritten.
- Since FR-3055, the URL owns the current project on `/project/:projectName/*` routes, so sider links keep pointing at the old project, a page reload snaps the project back to the old one, and the header selector / page data disagree with the URL (e.g. the deployment list shows the new project's rows under the old project's URL).

## Root cause

`SwitchToProjectButton` predates FR-3055 and still calls `setCurrentProject` (atom write) directly. `WebUIHeader.applyProjectChange` was migrated to rewrite the URL segment and let `ProjectScopeLayout` converge the atom, but the alert button was not.

## Fix

On `project` / `projectAdmin` scope, make `SwitchToProjectButton` rewrite the `:projectName` URL segment via `rewriteProjectNameInPath` (same rule as the header) and let `ProjectScopeLayout` converge `currentProjectAtom`; keep the direct atom write only outside project-scoped URLs.

Verified on a live backend: after the fix, clicking the button updates URL + header selector + sider links + client state consistently, the mismatch alert disappears, and the switch survives navigation and reload.

JIRA Issue: FR-3428

yomybaby:
Resolves #8504 (FR-3428)

## Summary

On `DeploymentDetailPage`, the cross-project warning alert's **Switch to '{project}'** button left the app in a half-switched state: it wrote `currentProjectAtom` directly, but since FR-3055 the URL owns the current project on `/project/:projectName/*` routes. The URL kept the old project, so sider links still pointed at the old project, a reload snapped the selection back, and the header selector / page data disagreed with the URL.

The same hazard existed for *any* component calling `useSetCurrentProject` directly, so this PR both fixes the button and structurally prevents the bug class.

## Changes

1. **`useSwitchProject` (new, `react/src/hooks/useRouteScope.ts`)** — the canonical way to change the current project from a component:
   - `project` / `projectAdmin` scope: navigate to the same page with only the `:projectName` segment rewritten (`rewriteProjectNameInPath`, query string preserved, defensive `buildPath` fallback), letting `ProjectScopeLayout` converge the atom to the URL.
   - global `admin` scope (no `:projectName` segment in the URL): update the atom directly — navigation-only is impossible here because the URL does not carry a project.
2. **All component callers migrated to the hook**: `SwitchToProjectButton`, `WebUIHeader.applyProjectChange` (its duplicated inline scope rule is removed), `ImportArtifactRevisionToFolderModal`, `AdminModelCardSettingModal`.
3. **ESLint gate** (`react/eslint.config.js`): `no-restricted-imports` now bans importing `useSetCurrentProject` anywhere except the sync layer, so new direct callers fail lint with a pointer to `useSwitchProject`. Allowlist:
   - `hooks/useCurrentProject.tsx`, `hooks/useRouteScope.ts` — setter home + hook implementation
   - `MainLayout/ProjectScopeLayout.tsx` — the URL→atom sync layer
   - `MainLayout/WebUIHeader.tsx` — the admin-mode-exit confirm flow, which must set the atom and navigate to a `goBackPath`-derived target in one step

## Verification

- ESLint negative test: a probe file importing `useSetCurrentProject` from a component fails with the new message; all allowlisted files pass.
- Live repro on a real backend (dev server + Playwright), deployment in project `default` opened under `/project/a한국어가능_cde/deployments/<id>`:
  - **Alert button**: click → URL becomes `/project/default/deployments/<id>`, header selector / sider links / `backendaiclient.current_group` all consistent, mismatch alert disappears, state survives navigation and reload.
  - **Header selector (hook-migrated path)**: switching back to `a한국어가능_cde` on the same page rewrites the URL in place, keeps the deployment detail mounted, and the mismatch alert correctly reappears.

### DIFF
--- a/react/eslint.config.js
+++ b/react/eslint.config.js
@@ -4,6 +4,27 @@ import relayPlugin from 'eslint-plugin-relay';
 import globals from 'globals';
 import jsoncParser from 'jsonc-eslint-parser';
 
+// Shared base for `no-restricted-imports` so the per-file allowlist blocks
+// below can re-declare the rule (flat-config rule configs replace, not merge)
+// without duplicating these entries.
+const restrictedImportPatterns = [
+  { group: ['backend.ai-ui/*', '!backend.ai-ui/dist'] },
+  { group: ['@lobehub/fluent-emoji'] },
+];
+const restrictedImportPaths = [
+  {
+    name: 'antd-style',
+    importNames: ['useThemeMode'],
+    message: "Use 'src/hooks/useThemeMode' instead.",
+  },
+  {
+    name: 'react-router-dom',
+    importNames: ['useNavigate', 'Navigate'],
+    message:
+      "Use 'useWebUINavigate' from 'src/hooks' or '<WebUINavigate>' from 'src/components/WebUINavigate' instead.",
+  },
+];
+
 export default [
   ...base,
   ...react,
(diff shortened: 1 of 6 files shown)


agatha197 (approved):
LGTM

PR #8506, opened 2026-07-31
fix(FR-3404): resolve project resource-group warning state mismatches on the Fair Share page

### LINKED ISSUE #8450: Fair Share settings: scheduler-not-applied warning missing on Project level until visiting User and returning
On the **Fair Share 설정** page, when the selected resource group does not have the Fair Share scheduler applied, a warning banner should be shown:

> 선택한 자원 그룹(upload)에 Fair Share 스케줄러가 적용되어 있지 않습니다. 스케줄러를 사용하시려면 자원 그룹 설정을 변경해 주세요.

The banner does not render when the **프로젝트** (Project) level is first opened. It only appears after navigating to the **사용자** (User) level and coming back to **프로젝트**.

**Steps to reproduce**

1. Open Fair Share 설정 and pick a resource group that has no Fair Share scheduler applied (observed with `upload`).
1. Drill down through 도메인 (`default`) to the **프로젝트** level.
   - Expected: the warning banner is shown. Actual: only the informational (blue) banner is shown; the warning (orange) banner is missing.
1. Navigate to the **사용자** level, then go back to **프로젝트**.
   - The warning banner now appears.

**Expected**

- The warning banner renders consistently at every breadcrumb level whenever the selected resource group has no Fair Share scheduler applied, including on first entry to the Project level.

**Environment**

- Manager version: `26.4.8`

**Notes**

- Screenshots attached: Project level without the warning (bug) and with the warning (expected, after returning from User).

JIRA Issue: FR-3404

nowgnuesLee:
Resolves #8450 (FR-3404)

Going **Project → User → back to Project** on the Fair Share settings page made the
⚠ "project not allowed in resource group" icon appear on **every** project row, even
though the domain allows the resource group. The symptom was intermittent, which is why
the two screenshots on the issue disagree.

## Cause

`UserResourceGroupAlert` read the step params from the URL itself via `useQueryStates`,
while its parent gated rendering on `currentStep === 'user'` — derived from
`deferredStepQueryParams`, so it lags by a render.

On the way back `project` is cleared immediately but the parent still believes it is on
the User step, so the child refetches with `projectId: ""` under `network-only`. The
server rejects the invalid UUID but still returns data:

```json
{ "errors": [ ... ], "data": { "domain": null, "group": null } }
```

`domain` is nulled too, even though `domain(name:)` does not take `projectId` — both
fields resolve through the same downstream call. Relay writes that `null` into the shared
root record `domain(name: "default")`.

`ProjectResourceGroupWarningIcon` has no `fetchPolicy`, so it uses Relay's default
`store-or-network`: the fields are present (as `null`), so it reads them **without
refetching**. `domain?.scaling_groups ?? []` becomes `[]`, the resource group is not in
the empty list, and the warning renders.

Only `domain` is cross-contaminated — the poisoning query wrote to `group(id: "", ...)`,
a different store key from each row's own `group(id: <rowProjectId>, ...)`.

## Fix

Pass the step values down as props so the gate and the query variables come from the same
deferred source. `currentStep` is `'user'` only when `resourceGroup`, `domain` and
`project` are all non-empty, so an empty `projectId` request becomes structurally
impossible.

The other call site (`FairShareWeightSettingModal`) already computes these values in
`INITIAL_FORM_VALUES` and now passes them explicitly.

## Verification

Manual, against `10.122.12.8` (`upload` resource group, `default` domain):

| | before | after |
|---|---|---|
| First entry to Project | icons hidden | icons hidden |
| After Project → User → Project | **icons shown (false positive)** | icons hidden |
| Requests with `projectId: ""` | 1 per round trip, 2 GraphQL errors each | **0** |

Ran the round trip 4 consecutive times after the fix — icons stayed hidden every time and
no request went out with an empty `projectId`.

## How to test

**Backend test server:** `10.122.12.8:8090` — a manager with Fair Share scheduling enabled
(the page is gated on `baiClient.supports('fair-share-scheduling')`, so a manager without it
redirects to `/error`). Credentials are the team's standard superadmin test account; not
repeated here since this repo is public — see `e2e/envs/.env.playwright`.

```bash
VITE_DEFAULT_API_ENDPOINT=http://10.122.12.8:8090 pnpm run dev
```

**Why this server reproduces it** — the domain allows a resource group that its projects do
not list:

```
domain("default").scaling_groups      = ["default", "upload"]
group("model-store").scaling_groups   = []
```

Any server with that shape works; on one where every project lists the resource group, the
warning has nothing to disagree about.

1. Admin → Scheduler → **Fair Share Setting**
2. Resource Group `upload` → Domain `default`. You are now on the **Project** step.
   Direct URL: `/admin/scheduler?resourceGroup=upload&domain=default`
3. Confirm there is **no** ⚠ icon next to `model-store` / `test` / `default`
4. Click a project name (e.g. `default`) to drill into the **User** step
5. Click **Project** in the breadcrumb Steps to come back
6. **Expected:** the icons stay hidden. Before this PR they appeared on every row.
7. Repeat 4–5 a few times — the old behaviour was a race, so one clean pass could be luck.

To confirm the cause is gone rather than just the symptom, watch the network panel: no
`UserResourceGroupAlertQuery` should go out with `"projectId": ""`. That request used to
return 2 GraphQL errors plus `data.domain = null`, which is what poisoned the shared store.

There are no e2e specs for this page yet, so the checks above are manual.

## Notes

- The Jira description for FR-3404 describes the *scheduler warning banner* rather than
  this warning icon. Per the reporter, the icon is the behaviour that was actually being
  reported, so the issue text likely needs updating.
- An earlier commit on this branch also made the weight-setting modal's
  `ProjectResourceGroupAlert` apply the icon's project-**or-domain** allow rule (the modal
  currently checks the project only, so it can contradict the row icon on one screen).
  That change was **reverted** to keep this PR scoped to the reported bug — the rule
  alignment is a product decision that belongs with follow-up 4 below.
- Follow-ups found while investigating, deliberately left out:
  1. `ProjectResourceGroupWarningIcon` uses `store-or-network` while its sibling
     `DomainResourceGroupWarningIcon` uses `store-and-network`. Aligning them would add a
     self-healing safety net.
  2. The icon issues one query **per row** (`ProjectFairShareTable.tsx:124`) — an N+1 that
     would disappear if `scaling_groups` were hoisted into the parent query. Hoisting into
     `FairShareListQuery` would also make the user-step alert appear in the same commit as
     the step transition instead of popping in afterwards.
  3. Separately, the scheduler warning banner is hidden whenever the table's pagination
     `offset` exceeds the resource-group count, because `selectedResourceGroupNode` is
     searched for inside a page slice that shares `limit`/`offset` with the current step's
     table. Reproducible with `?resourceGroup=upload&domain=default&current=2`. Not
     touched here.
  4. The resource-group allow rule is hand-written in several places
     (`ProjectResourceGroupAlert`, `ProjectResourceGroupWarningIcon`,
     `UserResourceGroupAlert`, plus the domain-only variant in the two `Domain*`
     components) and they already disagree: the modal's project alert checks the project
     only, while the row icon checks project-or-domain. The copies also run four different
     cache policies for the same question — each an axis that can silently disagree on
     screen. A shared `isResourceGroupAllowed()` predicate (or a hook owning the single
     query) would collapse them, and is the right place to settle the project-vs-domain
     rule question above.
  5. A sixth answer to "is this resource group allowed for this project" lives outside the
     directory: `packages/backend.ai-ui/src/hooks/useProjectResourceGroups.ts:76-137` resolves
     it over REST (`/scaling-groups?group=`) and additionally filters out SFTP-only groups,
     which the FairShare components do not. An SFTP-only resource group can therefore read as
     "not allowed" in `BAIProjectResourceGroupSelect` but "allowed" here. Worth reconciling
     alongside item 4 rather than separately.

https://claude.ai/code/session_01Vb6EvZz9GSzSrwDMB5z2Bh

### DIFF
--- a/react/src/__generated__/ProjectResourceGroupAlertQuery.graphql.ts
+++ b/react/src/__generated__/ProjectResourceGroupAlertQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<8f3c4e13ae6e0bdefc12a3b889284d66>>
+ * @generated SignedSource<<bf2117517ccb59e3046323508a1ffa7c>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -14,6 +14,9 @@ export type ProjectResourceGroupAlertQuery$variables = {
   projectId: string;
 };
 export type ProjectResourceGroupAlertQuery$data = {
+  readonly domain: {
+    readonly scaling_groups: ReadonlyArray<string | null | undefined> | null | undefined;
+  } | null | undefined;
   readonly group: {
     readonly scaling_groups: ReadonlyArray<string | null | undefined> | null | undefined;
   } | null | undefined;
@@ -37,6 +40,15 @@ v1 = {
 v2 = [
   {
     "alias": null,
+    "args": null,
+    "kind": "ScalarField",
+    "name": "scaling_groups",
+    "storageKey": null
+  }
+],
+v3 = [
+  {
+    "alias": null,
     "args": [
       {
         "kind": "Variable",
@@ -53,15 +65,23 @@ v2 = [
     "kind": "LinkedField",
     "name": "group",
     "plural": false,
-    "selections": [
+    "selections": (v2/*: any*/),
+    "storageKey": null
+  },
+  {
+    "alias": null,
+    "args": [
       {
-        "alias": null,
-        "args": null,
-        "kind": "ScalarField",
-        "name": "scaling_groups",
-        "storageKey": null
+        "kind": "Variable",
+        "name": "name",
+        "variableName": "domainName"
       }
     ],
+    "concreteType": "Domain",
+    "kind": "LinkedField",
+    "name": "domain",
+    "plural": false,
+    "selections": (v2/*: any*/),
     "storageKey": null
   }
 ];
@@ -74,7 +94,7 @@ return {
     "kind": "Fragment",
     "metadata": null,
     "name": "ProjectResourceGroupAlertQuery",
-    "selections": (v2/*: any*/),
+    "selections": (v3/*: any*/),
     "type": "Query",
     "abstractKey": null
   },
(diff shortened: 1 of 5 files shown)


ironAiken2 (approved):
LGTM

PR #8509, opened 2026-07-31
test(FR-3332): rewrite chat e2e mocks for the Deployments API

agatha197:
Replaces #8480, which GitHub auto-closed as merged when this branch was rebased under #8462 during restacking (no code reached `main` — verified `test/FR-3332-chat-e2e-deployment-mocks` is not an ancestor of `main`; the branch was also auto-deleted on merge and has been recreated here).

Follow-up to #8298 / #8420 (FR-3332). No dedicated issue — this repairs the e2e coverage for the migration those PRs landed.

Stacked on `refactor/FR-3332-chat-deployment-naming` (#8420) → `fr-3332` (#8298) → `main`. #8462 (FR-3397) now stacks on top of this PR instead of below it.

## Why

Every test in `e2e/chat/` fails against a live manager on top of the FR-3332 stack. The specs' mocks still described the legacy Graphene shapes the Chat surface no longer queries. Two breakages, both silent:

- **Same operation name, wrong shape.** `ChatPageQuery` and `ChatCardQuery` kept their names, so the interceptor still matched them — but they returned `endpoint_list { items { endpoint_id } }` and `endpoint { url }` while the migrated queries select `myDeployments` and `deployment(id:)`. Relay received payloads missing the fields it asked for, which made the `@catch` result not-ok and dropped the page into its error boundary.
- **Renamed operations are not an error.** `EndpointSelectQuery` / `EndpointSelectValueQuery` no longer exist (now `DeploymentSelect*`). `graphql-interceptor` forwards unmatched operations to the real backend, so the specs asserted mock data (`mock-endpoint`) against whatever deployments the target actually had.

`DeploymentTokenSelectQuery` was never mocked at all, so it sent the fake mock UUID to the manager and got back `badly formed hexadecimal UUID string`.

## What changed

| | Before | After |
| --- | --- | --- |
| `ChatPageQuery` | `endpoint_list { items { endpoint_id } }` | `myDeployments { edges { node { id } } }` |
| `ChatCardQuery` | `endpoint { endpoint_id url replicas name }` | `deployment { id networkAccess { endpointUrl } metadata { name } }` |
| `EndpointSelectQuery` | `endpoint_list { total_count items }` | `DeploymentSelectQuery` → `myDeployments { count edges { node … } }` |
| `EndpointSelectValueQuery` | `endpoint { endpoint_id name url }` | `DeploymentSelectValueQuery` → `deployment { id metadata { name } networkAccess { endpointUrl } }` |
| `DeploymentTokenSelectQuery` | *(unmocked)* | empty `accessTokens` connection |

- Global Relay IDs are built the way `toGlobalId` does (`base64("ModelDeployment:<uuid>")`), since Chat holds local UUIDs and converts on the way into `deployment(id:)`. `fromMockDeploymentGlobalId` reverses it so a factory can tell deployment A from B.
- One shared `chatGraphQLMocks()` map replaces the four restated inline copies; a spec spreads it and overrides a single operation. Reply strings and deployment names are exported constants, since restated literals are exactly what let these assertions drift.
- Terminology follows the migration: `mock-endpoint` → `mock-deployment`, `setupChatPageWithTwoEndpoints` → `setupChatPageWithTwoDeployments`.
- Removed `chatPageQueryTwoEndpointsMockResponse` — dead, no caller.

## Deferred to #8462

The original version of this PR (before the restack) also added e2e coverage for #8462's unavailability alert. Since this PR now sits *below* #8462 in the stack, ChatCard doesn't have that alert logic yet at this point — those test cases (and the `setupChatPageWithUnavailableDeployment` / `DeploymentServingOverrides` / `chatCardQueryNullUrlMockResponse` mock helpers that only existed to support them) were removed here and are expected to land in #8462 instead.

## Flakiness fixes

- **`waitForChatReady` gates every setup path on an *enabled* composer.** Visibility is not readiness: the composer renders disabled until the pane resolves a deployment URL and fetches its model list, and `press('Enter')` on a disabled composer silently does nothing — the failure then surfaces as a missing reply rather than as "the page was not ready".
- The sync-toggle wait race this PR originally fixed independently (`clickSyncToggle`) had already landed on `main` via a different PR as `setSync`; the rebase kept `setSync` and dropped the duplicate reimplementation.

## Verification

- `bash scripts/verify.sh` → **=== ALL PASS ===** (Relay / Lint / Format / TypeScript / terminology).

### DIFF
--- a/e2e/chat/chat-sync.spec.ts
+++ b/e2e/chat/chat-sync.spec.ts
@@ -5,6 +5,7 @@ import {
   setupChatPageWithTwoDeployments,
   MOCK_DEPLOYMENT_NAME,
   MOCK_DEPLOYMENT_NAME_B,
+  MOCK_DEPLOYMENT_URL_B,
   MOCK_REPLY_A,
   MOCK_REPLY_B,
   MOCK_MODEL_ID_B,
@@ -215,7 +216,7 @@ test.describe(
       // endpoint B) before clicking, so we don't miss a fast mocked response.
       const modelsBResponsePromise = page.waitForResponse(
         (response) =>
-          response.url().includes('mock-chat-endpoint-b') &&
+          response.url().includes(MOCK_DEPLOYMENT_URL_B) &&
           response.url().includes('/v1/models'),
       );
       await page
--- a/e2e/chat/chat-sync.spec.ts
+++ b/e2e/chat/chat-sync.spec.ts
@@ -2,7 +2,12 @@
 // Sections: 5 (Sync Input Propagation), 6 (Sync Toggle Off — Input Isolation)
 import {
   setupChatPage,
-  setupChatPageWithTwoEndpoints,
+  setupChatPageWithTwoDeployments,
+  MOCK_DEPLOYMENT_NAME,
+  MOCK_DEPLOYMENT_NAME_B,
+  MOCK_REPLY_A,
+  MOCK_REPLY_B,
+  MOCK_MODEL_ID_B,
 } from './mocking/chat-mock-data';
 import { test, expect, Page } from '@playwright/test';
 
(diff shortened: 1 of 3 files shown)


agatha197:
close this pr. octodog will handle it.

PR #8516, opened 2026-08-03
refactor(FR-3430): split FairShareList into per-step self-contained components

### LINKED ISSUE #8515: Refactor FairShareList: split the single root query into per-step self-contained components
## Problem

`react/src/components/FairShareItems/FairShareList.tsx` fetches all four fair-share connections (`adminResourceGroups`, `rgDomainFairShares`, `rgProjectFairShares`, `rgUserFairShares`) in a single root `useLazyLoadQuery`. Because of `@skip` variables, only the current step's connection is actually fetched — the single-query structure is pure overhead:

1. **Step-switching complexity**: 3 `@skip` flags, per-step order/filter variable switches (`FairShareOrderVariables` / `FairShareFilterVariables`), and `limit`/`offset` shared across 4 connections.
1. **Always-on `adminResourceGroups` fetch + page-window bug**: the resource group list is refetched on every pagination/filter change of lower steps, and the selected resource group is located via `_.find` on the **current page** of the unfiltered list. If it falls outside the page window, the scheduler warning Alert and the modal's `resourceGroupFrgmt` silently disappear.
1. **Inconsistent URL params**: `resourceGroup`/`domain` are stored by name but `project` by id, requiring a separate `projectV2` lookup just to display the project name.

## Fix

1. **Split into per-step self-contained components** — each step (ResourceGroup/Domain/Project/User) owns its own `useLazyLoadQuery`, pagination, filter, and order state. `FairShareList` keeps only the step URL state (`useQueryStates`), the Steps UI, and step-transition callbacks. Remove the `@skip` flags and variable switches.
1. **Selected resource group via single lookup** — replace the paged-list `_.find` with a name-filtered single query to supply the scheduler Alert and modal fragment.
1. **Name-only URL params** — store the project **name** in the URL. The user step resolves name → id via `rgProjectFairShares` with the `project.name` nested filter (added in 26.2.0), then queries `rgUserFairShares`. Remove the `projectV2` lookup. Deep-link/refresh/click-through all share one code path.
1. **Move row-selection state and modals into each step component** — eliminates the `as Array<...>` casting switches; step transitions reset selection naturally via unmount.
1. **Refresh button** — pass `fetchKey` down as a prop or move `AutoUpdateFetchKeyButton` into each step component, whichever is simpler.

Note: verify whether the 26.2.0 nested filter needs an `@since` gate relative to the fair share UI's minimum manager version.

JIRA Issue: FR-3430

ironAiken2:
Resolves #8515 (FR-3430)

## Problem

`FairShareList.tsx` fetched all four fair-share connections (`adminResourceGroups`, `rgDomainFairShares`, `rgProjectFairShares`, `rgUserFairShares`) in a single root `useLazyLoadQuery`. Because of the `@skip` variables, only the current step's connection was actually fetched — the single-query structure was pure overhead:

1. **Step-switching complexity** — 3 `@skip` flags, per-step order/filter variable switches (`FairShareOrderVariables` / `FairShareFilterVariables`), and `limit`/`offset` shared across 4 connections.
2. **Always-on `adminResourceGroups` fetch + page-window bug** — the resource group list was refetched on every pagination/filter change of lower steps, and the selected resource group was located via `_.find` on the *current page* of the unfiltered list. If it fell outside the page window, the scheduler warning Alert and the modal's `resourceGroupFrgmt` silently disappeared.

## Changes

1. **Per-step self-contained components** — new `ResourceGroupFairShareStep` / `DomainFairShareStep` / `ProjectFairShareStep` / `UserFairShareStep`, each owning its own `useLazyLoadQuery`, pagination, filter, order, row selection, and modals (`FairShareWeightSettingModal` / `UsageBucketModal`). `FairShareList` keeps only the step URL state, the Steps UI, and step-transition callbacks. All `@skip` flags, variable switches, and `as Array<...>` casting switches are gone; step transitions reset selection naturally via unmount.
2. **Scheduler warning via single lookup** — new `ResourceGroupSchedulerTypeAlert` queries the selected resource group by name (`filter: { name: { equals } }, limit: 1`), fixing the page-window miss. The weight-setting modal's resource group fragment is likewise fetched per step by name.
3. **URL params keep their existing format** — `resourceGroup`/`domain` by name, `project` by id, so pre-existing URLs and browser history stay valid. The project name in the Steps title is resolved by a small `projectV2` lookup (`ProjectStepNameText`) instead of being part of the root query. The unused `user` URL param is removed.
4. Step transitions now also reset the page number (`current`), which the previous forward-navigation handlers forgot (stale page carried into the next step).
5. **Raise Relay `gcReleaseBufferSize` from the default 10 to 20** (`RelayEnvironment.ts`) — re-entering a step could intermittently crash into the `ErrorBoundary` fallback (`Relay: Missing @required value at path ...`) with zero network requests. Leaving a step releases its queries into the store's GC release buffer, and the FairShare screens churn released `(query, variables)` combos fast — notably one `*ResourceGroupWarningIconQuery` per table row — so a step's records were garbage-collected between visits while the react-relay `QueryResource` cache hit on remount skipped the refetch. A larger buffer keeps recently released step queries retained across step round trips. This supersedes the per-entry fetchKey approach explored in #8565 (closed); see that PR for the full root-cause analysis.

## Behavioral notes

- `UserNotAllowedInResourceGroup` alert (`UserResourceGroupAlert`) now renders inside the user step (below the Steps panel) instead of above the info alert.
- `SchedulerPage`'s `ErrorBoundary` intentionally has no `resetKeys`: once an error is caught, the fallback stays until the user explicitly resets via the "go back to first step" button. This is by design and unchanged here.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Vite warmup / Terminology)


react/src/components/FairShareItems/DomainFairShareStep.tsx
```diff
@@ -0,0 +1,267 @@
...
+  resourceGroupName,
+  loading,
+  onClickDomainName,
+}) => {
+  'use memo';
+
+  const { t } = useTranslation();
+  const { token } = theme.useToken();
+
+  const [selectedRows, setSelectedRows] = useState<Array<DomainFairShare>>([]);
```
agatha197, line 46: The selection state (`selectedRows`/`selectedSingleRow`), modal open state, the toolbar JSX below (`BAIGraphQLPropertyFilter` + selection action buttons + `AutoUpdateFetchKeyButton`), and the `FairShareWeightSettingModal`/`UsageBucketModal` block at the bottom are all repeated almost verbatim (150+ lines) in `ProjectFairShareStep.tsx` and `UserFairShareStep.tsx`. `ResourceGroupFairShareStep.tsx` follows the same pattern minus the selection/modal parts.

Extracting this into a custom hook (e.g. a `useFairShareStepSelectionState` that bundles selection + modal state) or a shared toolbar component would reduce the maintenance burden of having to fix the same thing in 4 places.

react/src/components/FairShareItems/ProjectFairShareStep.tsx
```diff
@@ -0,0 +1,274 @@
...
+import { parseAsJson, parseAsStringLiteral, useQueryStates } from 'nuqs';
+import { useDeferredValue, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { graphql, useLazyLoadQuery } from 'react-relay';
+
+interface ProjectFairShareStepProps {
+  resourceGroupName: string;
+  domainName: string;
+  loading?: boolean;
+  onClickProjectName?: (projectName: string) => void;
```
agatha197: `ProjectFairShareTable`'s actual signature is `onClickProjectName?: (projectId: string) => void`, and it passes `record?.projectId` (the project is also identified by id in the URL). Only this file names the parameter `projectName`, which doesn't match what's actually passed. `FairShareList.tsx` correctly names the parameter `projectId` when consuming this callback, so it'd be good to align the type here to `projectId: string` too.

react/src/components/FairShareItems/ResourceGroupSchedulerTypeAlert.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  const { t } = useTranslation();
+
+  const { resourceGroups } =
+    useLazyLoadQuery<ResourceGroupSchedulerTypeAlertQuery>(
+      graphql`
+        query ResourceGroupSchedulerTypeAlertQuery(
+          $resourceGroupName: String!
+        ) {
+          resourceGroups: adminResourceGroups(
```
agatha197: When entering the Domain/Project/User steps, this query (`name` + `scheduler.type`) and the `adminResourceGroups(filter:{name:{equals:...}})` query each step component runs itself (for the `FairShareWeightSettingModal_ResourceGroupFragment`) both fire with the same filter, duplicating the lookup. Before this refactor these were combined into a single root query. There seems to be room to merge the two (fetch once at a higher level and pass down via fragment, or add the fields this component needs to the step's own fragment).

react/src/RelayEnvironment.ts
```diff
@@ -163,7 +163,9 @@ function fetchForSubscribe(
 function createRelayEnvironment() {
   return new Environment({
     network: Network.create(fetchFn, fetchForSubscribe),
-    store: new Store(new RecordSource()),
+    store: new Store(new RecordSource(), {
+      gcReleaseBufferSize: 20,
```
agatha197, line 168: This value fixes a crash on FairShare step re-entry, but `RelayEnvironment` is shared app-wide, so this doubles the Relay store's GC retention for every screen, not just FairShare. The PR description explains the rationale well, but it'd help to leave a one-line comment in the code too (e.g. `// FR-3430: retains step queries released during FairShare step navigation`) so whoever adjusts this value next knows why it's 20.

agatha197 (changes requested):
The refactoring direction (splitting into self-contained step components, fixing the page-window bug) is good. Please address the items below before merging. (`__generated__/**` was excluded from review.)

### Code quality
- Selection state, modal state, pagination, toolbar JSX, and modal JSX are nearly duplicated across the 4 step components (`ResourceGroupFairShareStep`/`DomainFairShareStep`/`ProjectFairShareStep`/`UserFairShareStep`). See the inline comment.
- `ProjectFairShareStep`'s `onClickProjectName` callback parameter name doesn't match the actual value (a project id). See the inline comment.

### Potential issues / risks
- Entering the Domain/Project/User steps duplicates the `adminResourceGroups` fetch (once for the scheduler warning, once for the modal fragment). See the inline comment.
- The `gcReleaseBufferSize` change affects the app-wide Relay store but has no explanatory comment in the code. See the inline comment.

### Test coverage
- There are no unit/e2e tests for FairShare, before or after this refactor. Given this PR fixes regression-prone behavior (the page-window bug, resetting pagination on step transitions), I'd recommend adding at least a minimal e2e scenario (navigate steps → verify filter/pagination persists correctly).

react/src/components/FairShareItems/DomainFairShareStep.tsx
ironAiken2: Extracted in 7c2c697ea: `useFairShareStepSelectionState` (bundles the multi/single row selection and modal open state, generic over the row type with a `keyField`) and `FairShareStepToolbar` (`BAIGraphQLPropertyFilter` + selection action buttons + `AutoUpdateFetchKeyButton`; the selection block is an optional prop, so `ResourceGroupFairShareStep` reuses the toolbar without it). The weight-setting/usage modal JSX stays per-step because its fragment props are entity-typed, but all its state wiring now comes from the shared hook.

react/src/components/FairShareItems/ProjectFairShareStep.tsx
ironAiken2: Renamed the callback parameter/type to `projectId: string` in 7c2c697ea, matching `ProjectFairShareTable` and the `FairShareList` consumer.

react/src/components/FairShareItems/ResourceGroupSchedulerTypeAlert.tsx
ironAiken2: Merged in 7c2c697ea: `ResourceGroupSchedulerTypeAlert` is now a fragment consumer (`ResourceGroupSchedulerTypeAlertFragment`) spread on the resource-group node each step already fetches for `FairShareWeightSettingModal_ResourceGroupFragment`, so the `adminResourceGroups` lookup fires once per step instead of twice. One visible consequence: the alert now renders inside the step content (above the toolbar) rather than above the Steps bar — consistent with how the user step renders `UserResourceGroupAlert`.

react/src/RelayEnvironment.ts
ironAiken2: Added in 62962c11e: `// FR-3430: retains step queries released during FairShare step navigation (default 10)`.

react/src/components/FairShareItems/FairShareList.tsx
```diff
@@ -578,340 +271,88 @@ const FairShareList: React.FC = () => {
...
 
+const ProjectStepNameText: React.FC<{ projectId: string }> = ({
+  projectId,
+}) => {
+  'use memo';
+
+  const { project } = useLazyLoadQuery<FairShareListProjectNameQuery>(
+    graphql`
+      query FairShareListProjectNameQuery($projectId: UUID!) {
+        project: projectV2(projectId: $projectId) {
```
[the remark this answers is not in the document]
ironAiken2: Keeping the project URL param id-based is a deliberate scope decision (commit 624332b83; PR description §3): pre-existing URLs and browser history stay valid, and the name-only scheme would additionally require the 26.2.0 nested-filter `@since` gate the issue itself flags as an open question. The `projectV2` lookup that remains is a single small query for the step title only — the page-window bug and step-switching complexity the issue targets are fixed regardless. I'll note the deviation on #8515 so the acceptance criteria stay aligned; name-only URLs can land as a follow-up once the minimum-manager-version gate is settled.

react/src/RelayEnvironment.ts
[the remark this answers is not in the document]
ironAiken2: This trade-off was made deliberately after fully building the deterministic parent-owned fetchKey approach in #8565 (closed — see its root-cause analysis): the per-entry key bookkeeping added meaningful complexity for a crash the larger buffer mitigates in practice. You're right that 20 is not a hard guarantee at page size 50; the durable fix is the planned follow-up of folding the per-row `*ResourceGroupWarningIconQuery` into the step queries (blocked on backend exposing the scaling-group association), which removes the release-buffer churn at its source. The rationale is now documented next to the value (62962c11e).

ironAiken2:
Addressed the review feedback in 62962c11e + 7c2c697ea (all inline threads replied & resolved):

- Extracted `useFairShareStepSelectionState` + `FairShareStepToolbar` to collapse the 4-way duplication
- Aligned `ProjectFairShareStep`'s callback param to `projectId`
- Merged the duplicate `adminResourceGroups` lookup — the scheduler alert is now a fragment on the step's existing resource-group node
- Documented the `gcReleaseBufferSize` rationale in code

`scripts/verify.sh` → `=== ALL PASS ===`.

Re: test coverage — agreed FairShare has no coverage today; a minimal e2e scenario (step navigation → filter/pagination persistence) is planned as a follow-up rather than in this refactor PR, to keep the diff reviewable.

react/src/components/FairShareItems/useFairShareStepSelectionState.tsx
nowgnuesLee, line 1: Is this hook actually needed? I don't really see why we need it.

react/src/components/FairShareItems/FairShareList.tsx
```diff
@@ -445,18 +175,13 @@ const FairShareList: React.FC = () => {
...
-            }}
-          >
-            {deferredStepQueryParams.project && `(${selectedProjectName})`}
-          </BAIText>
+          {deferredStepQueryParams.project && (
+            <Suspense fallback={null}>
+              <ProjectStepNameText
+                projectId={deferredStepQueryParams.project}
+              />
+            </Suspense>
```
nowgnuesLee: Don't use null fallback suspense.

react/src/components/FairShareItems/ResourceGroupFairShareStep.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  const [queryParams, setQueryParams] = useQueryStates(
+    {
+      order: parseAsStringLiteral(availableResourceGroupSorterValues),
+      filter: parseAsJson<any>((value) => value),
+    },
+    {
+      history: 'replace',
+    },
+  );
```
nowgnuesLee, line 52: The parent already handles this URL state — is there a reason each component needs to manage it too?

nowgnuesLee (dismissed):

react/src/components/FairShareItems/useFairShareStepSelectionState.tsx
ironAiken2: Removed in 12d47b235 — you're right. No other list page extracts selection state into a hook; the codebase convention (AdminComputeSessionListPage, VFolderNodeListPage, …) is plain `useState` in the component plus the shared `handleRowSelectionChange` helper. Each step now declares its selection/modal state inline and calls the helper directly. `FairShareStepToolbar` stays, since the toolbar JSX was the bulk of the duplication the earlier review asked to deduplicate.

react/src/components/FairShareItems/FairShareList.tsx
ironAiken2: Removed the boundary entirely in 12d47b235 — `ProjectStepNameText` now renders without its own Suspense. Step transitions happen in a `useDeferredValue` background render, so the name query now suspends *into* the transition and React keeps the previous UI until it resolves: the name is guaranteed to be present the moment the step commits. (With the old `fallback={null}` boundary, the transition could commit before the name query returned, so the title popped in after an empty gap.) On a direct deep link the suspension is caught by `SchedulerPage`'s Skeleton boundary, and since the name query loads in parallel with the step query the difference is negligible.

react/src/components/FairShareItems/ResourceGroupFairShareStep.tsx
ironAiken2: The steps are the actual owners of this state — `filter`/`order`/pagination are read and written only inside the step components, and the parent never reads them. The duplicate-looking `useQueryStates({ order, filter, current })` in `FairShareList` exists solely to null those keys on step transitions: all steps share the same URL key names, and the departing step can't clear them itself because it unmounts as part of that same transition. So the division is: parent = reset-on-transition orchestration only, child = actual control (same division as the project's tab-url-state Pattern A-ii, where self-contained children keep their own URL state and the host only cleans up on switch).

react/src/components/FairShareItems/ResourceGroupSchedulerTypeAlert.tsx
```diff
@@ -0,0 +1,53 @@
...
+import type { CSSProperties } from 'react';
+import { useTranslation } from 'react-i18next';
+import { graphql, useFragment } from 'react-relay';
+
+// `style` is the only pass-through key any call site would use — the antd-era
+// `extends AlertProps` surface is gone with antd; see the matching note in
+// DomainResourceGroupAlert.tsx.
+interface ResourceGroupSchedulerTypeAlertProps {
+  resourceGroupFrgmt?: ResourceGroupSchedulerTypeAlertFragment$key | null;
+  style?: CSSProperties;
```
agatha197, line 16: this is dead code.

agatha197 (changes requested):
Is there any reason to maintain pagination with 3 step? 
If I click 1, the pagination step is maintained. But if I click 2, the pagination step goes to 10. 
Image

ironAiken2:
> Is there any reason to maintain pagination with 3 step? If I click 1, the pagination step is maintained. But if I click 2, the pagination step goes to 10.

@agatha197 Good catch — that inconsistency was a bug, not intended. Each step owns its pagination via `useBAIPaginationOptionStateOnSearchParam`, which stores **both** `current` and `pageSize` in the URL, but the step-transition reset in `FairShareList` only cleared `order`/`filter`/`current` and left `pageSize` behind. So a page size chosen on one step leaked into the next step's table.

Fixed in c39e98b8e: `pageSize` is now cleared together with the rest of the per-step query state. The reset runs on both `Stepper`'s `onStepClick` and the row-click drill-down handlers, so every step transition (forward and backward) now starts from the step's own pagination defaults.

agatha197 (changes requested):
I know it is unrelated with this PR, but `Fair Share Factor` sorting doesn't work in every tab. Can you remove it from available sort items?

ironAiken2:
> I know it is unrelated with this PR, but `Fair Share Factor` sorting doesn't work in every tab. Can you remove it from available sort items?

I will work on this in a separate PR. @agatha197

agatha197 (approved):
LGTM

PR #8517, opened 2026-08-03
test: e2e auto-heal 2026-08-02 (manual run)

nowgnuesLee:
Automated e2e heal pass generated by the daily webui digest job (window: last 3 days). The digest run had **15 failures / 449 passed**; this PR heals the test-side issues and surfaces the app-side ones for follow-up.

## Fixed (test staleness — verified passing on the nightly backend)
- **`e2e/rbac/rbac-role-detail.spec.ts:264`** — under a pre-26.8.0 manager the role detail drawer renders three tabs (Scopes / Permissions / Role Assignments) with **Scopes** active by default (drift from #8483 FR-3406 / #8493 FR-3424). Now clicks the *Permissions* tab before asserting it is active. ✅ passes.
- **`e2e/serving/deployment-lifecycle.spec.ts:451`** — the "Add Revision" Preset Mode fields only render when a deployment preset exists; the test now self-provisions one via `ensureDeploymentPreset` + a `deployment-preset` feature gate (same convention as the sibling revision-attach test). ✅ passes 2×.

## Robustness + env-gated skips
- **`e2e/session/session-creation.spec.ts`** (4 tests) — resource-preset combobox selection made robust against live-polling re-renders (keyboard `Enter` instead of clicking a detaching option). The four launch tests are marked **`test.fixme`** because the nightly backend currently has **no installable session image** for the `default` project / resource group, so the required *Environments* field can never be satisfied and the Launch button stays disabled. This is an environment gap, documented inline — reviewers who prefer to keep the robustness fix but drop the `fixme` markers can do so once an image is installed.

## NOT fixed here — real app/backend regression (needs follow-up)
7 RBAC tests (`rbac-role-detail.spec.ts:379,496,611,737,823,917` and `rbac-role-crud.spec.ts:114`) are blocked by an actual regression, so their tests were **left untouched** rather than weakened:

> `RoleFormModal`'s `DomainScopeIdSelect` sends the domain **name** (`"default"`) as `scopeId` for DOMAIN-scoped roles, but the nightly manager (`26.8.0-alpha.0.8450`) now requires a **UUID** for `scope_type 'domain'`. Role creation fails server-side with `Invalid scope specified. (scope_id must be a valid UUID for scope_type 'domain', got 'default')`, so the Create Role modal stays open.

Fix requires app/backend work (either the manager accepts the domain name, or the webui resolves/passes a real domain UUID — which today isn't exposed via GraphQL). Recommend filing an FR.

## Other
- `e2e/environment/environment.spec.ts:440` failed transiently in the digest run; re-ran clean 2× with no code change needed (not included in this PR).

### DIFF
--- a/e2e/rbac/rbac-role-detail.spec.ts
+++ b/e2e/rbac/rbac-role-detail.spec.ts
@@ -290,9 +290,14 @@ test.describe(
       await expect(drawerPanel).toBeVisible({ timeout: 10000 });
 
       // 4. Verify two tabs are visible: "Permissions" and "Role Assignments".
-      // The former separate "Scopes" tab was merged into "Permissions" (now
-      // "Detailed Permissions" internally) — see the docblock on
-      // RolePermissionDetailTab.tsx.
+      // Managers below 26.8.0 (the nightly manager under test) don't support
+      // the merged "Detailed Permissions" view, so the drawer actually shows
+      // three tabs here — "Scopes", "Permissions", "Role Assignments" — with
+      // "Scopes" active by default. See the docblock on
+      // RoleDetailDrawerContent.tsx / RolePermissionDetailTab.tsx for the
+      // version-gated split. Assert the "Permissions" tab becomes active on
+      // click instead of assuming it's the default, so this passes on both
+      // legacy and merged-view managers.
       await expect(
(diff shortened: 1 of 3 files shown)


agatha197 (approved):
LGTM

PR #8533, opened 2026-08-03
chore(FR-3394): chart a wayfinder map for the recent-history resource restore bug

nowgnuesLee:
Relates to [FR-3394](https://lablup.atlassian.net/browse/FR-3394). Does **not** resolve it — this is charting only.

FR-3394 carries two reports. The second one (environment variable values disappearing) was investigated and ruled out as intended behaviour: `sanitizeSensitiveEnv` (`react/src/components/EnvVarFormList.tsx:261-295`) blanks a value when the variable *name* substring-matches a sensitive pattern, and `TIKTOKEN_CACHE_DIR` hits `/TOKEN/i`. The slash in the value was a red herring. The Jira issue is deliberately left untouched; that report is recorded under **Out of scope** on the map.

This PR adds a `.wayfinder/` decision map for the remaining report: the accelerator (`fgpu`) selection not surviving a "Start session from recent history" restore.

## Why a map

The failing code path is already located, but three things are genuinely undecided, and one of them needs a live cluster:

- four different sites in `ResourceAllocationFormItems.tsx` can rewrite `resource.accelerator` after a restore, and which one actually fires is unknown without a reproduction on a cluster with a fractional GPU slot;
- it is not settled whether restored values *should* outrank image-derived defaults — silently preserving a stale accelerator can produce a session that fails to launch, which is a worse failure than a visibly reset field;
- it is not established whether `cpu` / `mem` / `shmem` are affected too, which decides whether an accelerator-specific fix is even the right shape.

## Contents

- `map.md` — destination, notes, and a "what is already known" section capturing the restore path and the four overwrite sites with file:line references, so a later session does not re-derive them.
- `tickets/` — four tickets. Frontier (unblocked, unclaimed) is *Instrument the restore path* and *Decide whether restored resource values outrank image-derived defaults*; the latter is deliberately left unblocked so it can run in parallel.

No product code changes, so `scripts/verify.sh` is not applicable. Relative links between map and tickets and the `blocked_by` slugs were checked to resolve.

https://claude.ai/code/session_01EdM5kdViXpYxdJ2vCdXvhq

[FR-3394]: https://lablup.atlassian.net/browse/FR-3394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nowgnuesLee:
Opened this by mistake — closing.

Issue #8538, opened 2026-08-03
26.8 user manual refresh


yomybaby:
Close the documentation gap between the user manual's content baseline and the 26.8.0 release.

## Baseline

The last content-bearing docs commit is #8056 (FR-3219 user manual review pass, 2026-06-29). Everything in packages/backend.ai-webui-docs/ after that date is tooling and terminology-gate work, not content.

## Gap

105 commits touch react/ or packages/backend.ai-ui/src/ since that baseline. 103 of them shipped in 26.8; 2 landed after the release cut and belong to 26.9.

Triage result: _72 of the 105 require a manual change, across 17 of the 26 chapters._ Six of them are removals — docs to delete, not add.

## Inclusion criterion

Any user-visible change: new or changed surface, UI label and terminology renames, fixes that changed observable behaviour, and removals. Excluded: dev-only tooling, internal refactors, perf, and test-only work.

The label-rename class matters disproportionately — FR-3331 alone switched edit-form submit buttons to "Save", silently invalidating documented button names across several chapters.

## Approach

One sub-task and one PR per manual chapter. Chapter file sets are disjoint, so the PRs carry no merge conflicts and can land in any order. Each PR body carries an attribution table mapping changed section to the FR, the original PR, and its author, and every driving author is added as a reviewer.

Prose is written in all four languages (en/ko/ja/th). Screenshots are _not_ captured in these PRs — screenshot needs are recorded as TODO markers and aggregated into a manifest for a single follow-up capture pass.

## Out of scope

- Screenshot capture (a separate effort, once a server is available)
- The batched back-port to the 26.8 release branch, which refreshes docs-archive/26.8 and the live /26.8/ site via FR-3242. Must exclude the 2 post-cut commits.
- The 96 pre-existing screenshot TODO markers already in the manual

## Planning artifact

Charted as a wayfinder map under .scratch/26-8-docs/ (map.md, coverage-matrix.md, issues/), which records the standing decisions and the full 105-row triage with its exclusion reasons.

JIRA Issue: FR-3445

Issue #8355, opened 2026-07-23
BAIFetchKeyButton auto-refresh countdown border desyncs from the paused interval while a refresh is in flight


agatha197:
## Summary

While auto-refresh is on, `BAIFetchKeyButton` pauses the shared refresh interval whenever the consumer passes a `loading` flag. `ContainerLogModal` passes `loading={isPending || isRefetching`}, and the interval is paused via `useInterval(triggerRefresh, loading ? null : selectedDelay, ...)`, so the next refresh only fires `selectedDelay` ms after the in-flight request **finishes**. However, the animated countdown border (`BAICountdownBorder`) keeps running its CSS animation independently: it is reset (`resetKey={cycleKey`}) at request **start** and is neither paused nor re-anchored when `loading` returns to `false`.

## Effect

The border completes its clockwise fill `selectedDelay` ms after the request **starts**, while the actual next refresh happens `selectedDelay` ms after the request **finishes**. The visible countdown therefore lies about when the next refresh occurs. In `ContainerLogModal` the `get_logs()` query has a **15s timeout** and the default interval is **10s**, so a slow request can let the border complete a full cycle (and start looping again) before the current request even ends.

## Root cause

- `BAIFetchKeyButton` pauses the interval on `loading` (delay -> `null`) but does not pause the countdown border.
- `cycleKey` — the shared `resetKey` for both the interval and the border — is bumped only inside `triggerRefresh` (request start), never on the `loading` true->false edge.
- `BAICountdownBorder` runs `animation-iteration-count: infinite` on its SVG rect with no way to pause, so it keeps filling and looping while the request is in flight.

## Expected

The shared button should pause **and** re-anchor the countdown border together with the interval: freeze the border while `loading` (paused, exactly like the interval), then restart it from empty when `loading` returns to `false` so the newly exposed countdown fills over the same window as the freshly-restarted interval.

## Proposed fix

- Add a pause/freeze capability to `BAICountdownBorder` (e.g. `animation-play-state: paused`).
- In `BAIFetchKeyButton`, pass `paused={loading`} and re-anchor the border (bump `cycleKey`) on the `loading` true->false transition so the border and the freshly-restarted interval share the same anchor instant.
- Add/extend a Storybook story (and/or unit test) covering the "refresh in flight longer than the interval" case.

## Affected files

- `packages/backend.ai-ui/src/components/BAIFetchKeyButton.tsx`
[...]

PR #8560, opened 2026-08-03
docs(FR-3461): update SFTP to Container chapter for 26.8

### LINKED ISSUE #8554: Update the SFTP to Container chapter for 26.8
Update `src/{en,ko,ja,th}/sftp_to_container/sftp_to_container.md` for 26.8. 1 driving row.

## Driving change

- FR-3359 — the SSH/SFTP connection dialog no longer shows a bogus `127.0.0.1` address when the proxy is unavailable. It now surfaces an explicit error instead: `ProxyNotReady`, `ProxyDirectTCPNotSupported`, or `InvalidRedirectURL`.

If this chapter tells the reader to expect or work around the `127.0.0.1` value, that instruction is now wrong and must be replaced with what the new errors mean and what to do about each.

## Related

The Troubleshooting chapter's "SFTP disconnection" entry covers the same change and is handled in FR-3453. Keep the two consistent — this chapter explains the connection flow, that one handles the failure FAQ. Do not duplicate the error table in both.

## Screenshot candidate

FR-3396 changed the SSH/SFTP connection modal's width so its values no longer wrap. That was excluded from the prose triage as cosmetic, but it makes the existing screenshot of this modal stale — it is recorded in the screenshot manifest as a recapture candidate.

## Constraints

- All four languages; no version-gated prose; screenshots as TODO markers only; docs-checks.yml green

Reviewer: ironAiken2 (sole driving author).

JIRA Issue: FR-3461

yomybaby:
Resolves #8554 (FR-3461)

## Summary

Updates the SFTP to Container chapter for 26.8 in all four languages (en/ko/ja/th) and recaptures the SSH/SFTP connection modal.

Part of the 26.8 user manual refresh (epic #8538 / FR-3445).

## Attribution

| Section | Driven by | Author |
|---|---|---|
| SSH/SFTP dialog no longer shows a bogus `127.0.0.1`; explicit errors instead | FR-3359 #8355 | @ironAiken2 |
| `SSH_SFTP_connection.png` recapture (dialog width) | FR-3396 #8438 | @ironAiken2 |

## What changed and why

The chapter told readers to expect `work@127.0.0.1` from the WebUI dialog. FR-3359 replaced that with explicit errors, so the instruction was wrong. The dialog paragraph now describes the real fields (User / Host / Port, the ready-made `sftp` / `scp` / `rsync` examples, the **Download SSH Key** button) and a new **Connection Errors** section (`<a id="connection-errors">`) documents the three failures by their user-visible strings.

**A nuance the issue did not capture:** `127.0.0.1` is still legitimate in the desktop app. `useBackendAIAppLauncher.tsx` takes the `tcpHost = '127.0.0.1'` branch when `globalThis.isElectron`; the FR-3359 error paths are browser-only. So the docs say the host is local *only in the desktop app* and is the App Proxy address in a browser — rather than claiming it is never local.

Errors are keyed by their **user-visible i18n strings**, not the internal `ProxyNotReady` / `ProxyDirectTCPNotSupported` / `InvalidRedirectURL` identifiers, which never appear in the UI. Note `ProxyNotReady` covers four distinct code paths, so the prose gives a wait-and-retry step rather than a single cause.

Coordinated with FR-3453 (Troubleshooting): that chapter links here via `#connection-errors` and added an `<a id="sftp-disconnection">` anchor for the reverse link. Those cross-references resolve once both land on main.

## A pre-existing defect this uncovered

On `main`, `ja/images/SSH_SFTP_connection.png` and `th/images/SSH_SFTP_connection.png` were **byte-identical copies of the English image** — never captured in their own locale, despite SCREENSHOT-GUIDELINES.md requiring per-locale captures for any image showing WebUI chrome. All four are now genuine per-locale captures, which is why the ko/ja heights differ (localized prose reflows).

## Screenshot verification

Preflight: the old image was 1354x1198, modal-only scope. The new one is 1600x1644 (an 800x821 CSS modal at 2x) — same modal-only scope, genuinely larger because FR-3396 widened the dialog. No page chrome included.

Locale correctness was enforced, not assumed: each locale boots from scratch (language persisted to localStorage, full reload) rather than using in-place `switchLanguage()`, which leaves `<Trans>` strings inside `'use memo'` components stale. Every locale asserts the localized `session.SFTPConnection` / `session.ConnectionInformation` strings are present **and** that the English forms have count 0.

The FR-3396 fix is confirmed by measurement, not by eye: each `<pre>` is `height: 46` = one 22px line plus padding. A wrapped block would be 68.

## Worth a follow-up FR

**The example commands are unwrapped but horizontally clipped.** Measured `scrollWidth` 906-1344 against `clientWidth` 750, so the commands run past the right edge and are cut mid-token (`-P 311`, `UserKnownHostsFile=/c`). In the live UI the block scrolls, so users can reach the rest; in a static screenshot they cannot. The prose above the image says the dialog shows ready-made commands, which a reader cannot verify from the picture. No framing choice fixes this — it needs a wider modal or wrapped code block.

## Verification

All three `docs-checks.yml` steps green: `check:terminology-md`, `check:nav-titles` (112 entries), `check-terminology-i18n.mjs --strict` (0 blocking / 0 warn).

**Server state:** the capture server had zero vfolders and zero sessions, so a folder and an SFTP session had to be created. Both were destroyed and hard-purged afterwards (`delete_from_trash_bin`, not left in trash), verified empty by a second pass. The read-only capture config was used throughout, so no teardown swept anyone else's resources.

## Noted, not fixed

Thai `Host` renders as `เจ้าภาพ` (a party host) and `Port` as `ท่าเรือ` (a harbour). Both are mistranslations. Terminology precedence says the live label wins, so the Thai page names the fields as `เจ้าภาพ (Host)` / `ท่าเรือ (Port)` on first mention so readers can find them on screen, then uses the correct `โฮสต์` / `พอร์ต` in running prose. Fixing the labels themselves is an i18n change.

## Review revision (2026-08-05)

Reviewed the ko chapter with 홍성철 and applied the confirmed edits (commit c97cabbd6):

- **Connection-info paragraph trimmed (all 4 langs)** — dropped the desktop-app/browser host-origin explanation (verified factually correct against the WebUI `useBackendAIAppLauncher` code and backend.ai `scaling_group` config, but judged too internal for the manual) and the placeholder-address sentence.
- **Connection Errors section tightened (all 4 langs)** — removed the client-package fallback sentence, the dialog-guarantee closing paragraph, and the troubleshooting cross-reference; the section now ends with the three error-message bullets.
- **VSCode / SSH config paragraphs (all 4 langs)** — dropped the dialog-open caveat and the local-address clause; simplified the transcript note to "replace the host and port with the values shown in your own dialog".
- **ko only** — full-chapter prose pass (heavy pre-modifiers, connective-ending commas, overused "~할 수 있습니다", purpose/passive calques, doubled particles). Code blocks verified byte-identical; structure/anchors/labels unchanged; finalize verification passed with no corrections.


ironAiken2 (approved):
LGTM

PR #8565, opened 2026-08-04
fix(FR-3430): force a fresh fetch on FairShare step remount with per-mount fetchKey

### LINKED ISSUE #8515: Refactor FairShareList: split the single root query into per-step self-contained components

ironAiken2:
Resolves #8515 (FR-3430)

> Stacked on #8516 (`refactor/FR-3430-fairshare-step-queries`).

## Problem

Revisiting a FairShare step intermittently crashed into the "invalid parameter" `ErrorBoundary` fallback (`Relay: Missing @required value at path 'domainFairShares' ...`) with **zero network requests**, even though every server response in the session was clean. A full page reload always cleared it.

## Root cause

Three layered client-side behaviors combine (verified live in Chrome with an instrumented Relay environment):

1. **Relay store GC** — leaving a step releases its queries into the store's GC release buffer (default **10** slots). The FairShare screens churn released `(query, variables)` combos fast — notably one `*ResourceGroupWarningIconQuery` **per table row** — so a step's query gets evicted and its records garbage-collected after a normal round trip. GC runs as a deferred incremental pass, which is why the crash is intermittent: returning before the pass completes re-retains the data unharmed.
2. **QueryResource cache hit skips the fetch** — every step mounts with the same shared `INITIAL_FETCH_KEY` constant, so a remount reproduces the exact react-relay cache identifier (`fetchPolicy + variables + fetchKey`). `QueryResource.prepare()` then reuses the previous mount's entry and **never issues a network request** (`fetchPolicy` is only consulted on a cache miss — `network-only` would not help; see facebook/relay#3502).
3. **`@required(action: THROW)` fires on GC'd data** — `RelayReader._handleRequiredFieldValue` uses a loose `value == null` check, so records deleted by GC (read as `undefined`) throw exactly like an explicit server `null`.

## Solution

`FairShareList` owns **one entry fetchKey per step** and regenerates the target step's key whenever the step URL params actually move — covering clicks, browser back/forward, and `BAIBackButton` uniformly. Step components substitute that key for `INITIAL_FETCH_KEY`, so a re-entered step always misses the QueryResource cache: `store-and-network` renders surviving cached data instantly (with a background refresh) and suspends until the network response when the data was GC'd.

The keys deliberately live in the **committed parent**, not in a `useState` inside each step: a suspended step transition discards and restarts background renders, so a per-mount initializer in a step would mint a new key per render attempt and duplicate the request (an earlier revision of this fix did exactly that and produced 3-4 identical queries per transition). `ResourceGroupSchedulerTypeAlert` shares the domain-entry key since a GC'd read silently hides the alert instead of crashing. The shared `useFetchKey` hook is unchanged.

## Verification

- Root-cause A/B repro (per-mount-key revision, automated step round trips incl. GC idle gaps): before — crash within 2 cycles, dangling `rgProjectFairShares` refs in the store, 0 requests at crash time, 0 error/null responses all session; after — 0 errors across 3+ cycles with every re-entry issuing a fresh request.
- ⚠️ The final parent-owned-key revision passed `verify.sh` but its live request-count measurement is **pending** — the backend test server became unreachable mid-verification. Will re-run: expected behavior is exactly 1 step-query request per step entry (no per-attempt duplication).

## Follow-ups (out of scope)

- Fold per-row `*ResourceGroupWarningIconQuery` data into the step queries to cut release-buffer churn (needs backend: `DomainV2`/`ProjectV2` do not expose scaling-group association).
- Consider raising `gcReleaseBufferSize` in `RelayEnvironment.ts` app-wide.

### DIFF
--- a/react/src/components/FairShareItems/DomainFairShareStep.tsx
+++ b/react/src/components/FairShareItems/DomainFairShareStep.tsx
@@ -36,12 +36,14 @@ import { graphql, useLazyLoadQuery } from 'react-relay';
 
 interface DomainFairShareStepProps {
   resourceGroupName: string;
+  initialFetchKey: string;
   loading?: boolean;
   onClickDomainName?: (domainName: string) => void;
 }
 
 const DomainFairShareStep: React.FC<DomainFairShareStepProps> = ({
   resourceGroupName,
+  initialFetchKey,
   loading,
   onClickDomainName,
 }) => {
@@ -88,7 +90,9 @@ const DomainFairShareStep: React.FC<DomainFairShareStepProps> = ({
   };
   const deferredQueryVariables = useDeferredValue(queryVariables);
   const [fetchKey, updateFetchKey] = useFetchKey();
-  const deferredFetchKey = useDeferredValue(fetchKey);
+  const effectiveFetchKey =
+    fetchKey === INITIAL_FETCH_KEY ? initialFetchKey : fetchKey;
+  const deferredFetchKey = useDeferredValue(effectiveFetchKey);
 
   const { resourceGroups, domainFairShares } =
     useLazyLoadQuery<DomainFairShareStepQuery>(
@@ -132,7 +136,7 @@ const DomainFairShareStep: React.FC<DomainFairShareStepProps> = ({
       {
         fetchKey: deferredFetchKey,
         fetchPolicy:
-          deferredFetchKey === INITIAL_FETCH_KEY
+          deferredFetchKey === initialFetchKey
             ? 'store-and-network'
             : 'network-only',
       },
@@ -187,7 +191,7 @@ const DomainFairShareStep: React.FC<DomainFairShareStepProps> = ({
           <AutoUpdateFetchKeyButton
             settingId="fair-share-list"
             autoUpdateDelayOptions={LONG_AUTO_UPDATE_DELAY_OPTIONS}
-            loading={fetchKey !== deferredFetchKey}
+            loading={effectiveFetchKey !== deferredFetchKey}
             value=""
             onChange={() => {
               updateFetchKey();
@@ -202,7 +206,7 @@ const DomainFairShareStep: React.FC<DomainFairShareStepProps> = ({
         loading={
           loading ||
           queryVariables !== deferredQueryVariables ||
-          fetchKey !== deferredFetchKey
+          effectiveFetchKey !== deferredFetchKey
         }
         selectedRows={selectedRows}
         onRowSelect={(selectedRowKeys, currentPageItems) => {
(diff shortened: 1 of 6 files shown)


ironAiken2:
Closing in favor of a simpler mitigation: raising `gcReleaseBufferSize` from the default 10 to 20 in `RelayEnvironment.ts` (now part of #8516, commit 6c3a72bc1). The root-cause analysis here stays valid — FairShare step screens churn released queries fast enough to evict a step's records between visits — but a larger release buffer keeps the step queries retained across round trips, making the per-entry fetchKey machinery unnecessary.

PR #8562, opened 2026-08-03
docs(FR-3457): update Session Page chapter for 26.8

yomybaby:
Resolves #8550 (FR-3457)

> **Review update (2026-08-04):** the "Allocated vs requested resources" section (FR-3179) and its mocked `session_detail_resource_allocation.png` screenshot were **dropped from this PR** per review — the session detail panel is owned by the `sessions_all` chapter, so that content did not belong here. The review also reworded the personal-page scope note to reference the project selector in the top bar. Both changes were applied to all four languages.

## Summary

Updates the Session Page chapter for 26.8 in all four languages (en/ko/ja/th) and recaptures two screenshots.

Part of the 26.8 user manual refresh (epic #8538 / FR-3445).

## Attribution

| Section | Driven by | Author |
|---|---|---|
| Corrected the personal Sessions page address | FR-3055 #7751 | @yomybaby |
| ~~Allocated vs requested resources with mismatch warning~~ — dropped in review; belongs to the session detail panel in `sessions_all` | FR-3179 #8072 | @ironAiken2 |
| "My Total Resource Usage" rename; limit/usage toggle removed | FR-3283 #8215 | @yomybaby |

Reviewers are the driving authors minus the PR author (@yomybaby opened this PR).

## The chapter stated something verifiably false

`session_page.md:38` read *"The personal Sessions page (`/session`) ..."*. FR-3055 changed the URL scheme, so that path no longer exists — `/session` is now only a `ProjectScopedRedirect` stub (`routes.tsx:991`).

Corrected in all four languages to `/project/<project name>/session`, derived from source: `pathBuilder.ts` `buildPath()` owns the layout, `routes.tsx:236` declares the parent `project/:projectName`, and the segment is the project **name**, not a UUID. Confirmed live — logging into the capture server lands on `/project/default/start`.

During review the same note was further clarified: the page shows only the sessions of the project selected with the **project selector** in the top bar (wording aligned with the header chapter and the `SwitchToAccessibleProject` i18n label).

Scope discipline: only the demonstrably wrong sentence was fixed. The manual is **not** gaining a routing/URLs chapter this release; that is recorded as a separate follow-up candidate.

## Screenshots — two captured

| Image | Scope | Status |
|---|---|---|
| `sessions_page.png` | full viewport | ✅ recaptured, live data |
| `panel_settings.png` | dropdown element | ✅ recaptured, live data |

[...]

PR #8556, opened 2026-08-03
docs(FR-3458): update Login chapter for 26.8

yomybaby:
Resolves #8551 (FR-3458)

## Summary

Updates the Login chapter for the 26.8 release in all four languages (en/ko/ja/th), and captures the new Unauthorized Access page screenshot.

Part of the 26.8 user manual refresh (epic #8538 / FR-3445).

## Scope trim (63a14dbc2)

Per author review, the chapter was simplified after the initial capture:

- The **Login page language** section (browser-language detection explanation, resolution-order list, and note) was **removed entirely** in all four languages — the FR-2981 behavior is intentionally left undocumented at chapter level.
- In **Sidebar menus**, the intro paragraph was dropped and **Menus available to your role** now comes before **Resizing the sidebar**.
- The project-admin scope phrase now reads "the currently selected project" across languages.
- ko only: a double-particle fix (`Manager로의` → `Manager로`).

All four languages remain structurally identical (14 headings / 12 anchors / 8 images / 24 admonition fences per language).

## Attribution

| Section | Driven by | Author |
|---|---|---|
| Forbidden page; sidebar filtered by URL-scoped role | FR-3383 #8408 | @yomybaby |
| UI label audit | FR-3376 #8403 | @yomybaby |

(The FR-2981 login-language row was dropped with the scope trim above.)

## Reviewer note

**The reviewer-conflict fallback applies here.** Every driving row is authored by @yomybaby, who also opened this PR — GitHub will not accept a self-approval, so the PR would be unapprovable as assigned. Falling back to the highest-volume non-self contributor in the 26.8 window. Please reassign if someone closer to FR-3383 would rather take it.

## Screenshot

`unauthorized_access_page.png`, new, captured in all four locales at 2880x1800.

Captured as `user@lablup.com` (not admin) against `/admin/users` — the whole `/admin` subtree carries `handle: { access: 'admin' }` in `routes.tsx`. Three hard assertions confirm the forbidden page actually rendered before capture, rather than assuming.

[...]

PR #8569, opened 2026-08-04
docs(FR-3453): update Troubleshooting chapter for 26.8

### LINKED ISSUE #8546: Update the Troubleshooting chapter for 26.8
Update `src/{en,ko,ja,th}/trouble_shooting/trouble_shooting.md` for 26.8. 6 driving rows.

_This chapter mostly SHRINKS._ Two existing FAQ entries describe problems the product now handles itself — they should be rewritten or removed, not extended.

## FAQ entries whose premise is now obsolete

- _"Indicated resources do not match with actual allocation"_ — FR-3179 makes the session detail page show allocated vs requested resources with an explicit mismatch warning. The entry's premise is now handled in-product; shrink it or repoint it at the new warning.
- _"SFTP disconnection"_ — FR-3359 replaced the bogus `127.0.0.1` SSH/SFTP dialog with explicit errors. Re-read the entry against the new messages: `ProxyNotReady`, `ProxyDirectTCPNotSupported`, `InvalidRedirectURL`.

## Other driving changes

- FR-3237 — endpoint reachability is confirmed before the offline banner is shown, so spurious offline reports should no longer appear
- FR-3383 — access-denied routes now render a Forbidden page
- FR-3057 / FR-3279 — scope-aware 404 with a guided fallback for invalid-project URLs

## Constraints

- All four languages; no version-gated prose; screenshots as TODO markers only; docs-checks.yml green

Reviewers: ironAiken2, nowgnuesLee (yomybaby excluded as PR author).

JIRA Issue: FR-3453

yomybaby:
Resolves #8546 (FR-3453)

> ## ⚠️ The screenshot was captured with a mocked session
>
> **Why:** the capture server's `get_resource_slots()` returns `{"cpu":"count","mem":"bytes","cuda.device":"count"}` — **no fractional-GPU slot** — so an allocated-vs-requested mismatch cannot physically occur there. The FR-3179 author hit the same wall (#8072).
>
> **What was mocked:** two GraphQL operations at `**/admin/gql`, `ComputeSessionListPageQuery` and `SessionDetailContentQuery`, via the project's `setupGraphQLMocks` (client-side interception; nothing created server-side). Session `pytorch-train-01` with `cuda.shares` requested `0.5` / occupied `0`; CPU and memory granted exactly as requested.
>
> **Not mocked:** the resource-slot metadata — the `fGPU` unit and NVIDIA icon come from the server's own `resources/device_metadata.json`, which genuinely carries `cuda.shares` with `display_unit: "fGPU"` and `round_length: 1` (hence `0.0` rather than `0`). Everything else is real UI.

## Summary

Updates the Troubleshooting chapter for 26.8 in all four languages (en/ko/ja/th). **This chapter mostly shrinks** — two FAQ entries describe problems the product now handles itself.

Part of the 26.8 user manual refresh (epic #8538 / FR-3445).

## Attribution

| Section | Driven by | Author |
|---|---|---|
| "Indicated resources do not match with actual allocation" rewritten | FR-3179 #8072 | @ironAiken2 |
| "SFTP disconnection" re-scoped to post-connection failures | FR-3359 #8355 | @ironAiken2 |
| Offline banner no longer fires spuriously | FR-3237 #8099 | @nowgnuesLee |
| Route-error pages (Forbidden, scope-aware 404) | FR-3383 / FR-3057 / FR-3279 | @yomybaby |

Reviewers are the driving authors minus the PR author (@yomybaby opened this PR).

## Cross-chapter coordination

Three chapters in this refresh touch the same failure paths; they were coordinated rather than left to duplicate each other:

- **`#connection-errors`** — the SFTP chapter (#8560) documents the three launch errors in full, keyed by their user-visible strings. This chapter **links** there twice instead of restating them, and scopes its own entry to *post-connection* failures.
- **`<a id="sftp-disconnection">`** — added here in all four languages at the SFTP chapter's request, because heading-derived anchors are localized and its reverse link would otherwise break in ko/ja/th.
- **Forbidden page** — the Login chapter (#8556) owns the full description and the `#pages-you-cannot-open` anchor; this chapter carries a one-line bullet linking there, deliberately not a second hop to `#menus-available-to-your-role`.

These links resolve once the chapters land on main; they are cross-branch by construction.

## Not one FAQ entry per driving row

FR-3057, FR-3279 and FR-3383 all produce the same user experience — an in-app route-error page with an address pill and a "go to your first available page" button. They are documented as **one** entry with four message bullets. Splitting them would have grown a chapter that is supposed to shrink.

## Screenshot framing

The marker asked for the tooltip to be visible, which the equivalent image on #8562 deliberately excludes: at a 1440px viewport the 800px drawer starts at x=640, and antd anchors the tooltip so it overhangs the drawer's left edge by ~31px, dragging the dimmed session list behind it into frame.

Solved by capturing at an **800x900 viewport** so the drawer fills the window — the overhang then lands on the drawer's own background and the crop contains nothing but the detail panel. `md` is still true at 800 (breakpoint 768), so the two-column Descriptions layout is preserved. Crop bounds are computed per locale from the tooltip and cell borders.

Two accepted cosmetic consequences: the tooltip partly covers the Environments row above (inherent to a hover tooltip), and it sits flush against the left frame edge because antd clamps it to the viewport.

## A prose correction, made in two chapters

The worked example `1 / 2 Core` was replaced with `0.0 / 0.5 fGPU` here **and** on #8562. CPU is always granted exactly as requested in Backend.AI, so a CPU shortfall is not a reachable product state — the old example depicted something that cannot happen, and on #8562 it directly contradicted the screenshot beneath it. The fGPU round-down named two paragraphs later is the real cause.

## Verification

All three `docs-checks.yml` steps green: `check:terminology-md`, `check:nav-titles` (112 entries), `check-terminology-i18n.mjs --strict` (0 blocking / 0 warn). Each locale booted from scratch with `expect(getByText('Resource Allocation',{exact:true})).toHaveCount(0)` asserted per non-en locale. Nothing created server-side.

**No orphaned images:** the chapter referenced zero images before this change, so nothing lost its last reference.

## Label drift fixed

- ja: `**メンテナンス**` → `補修(メンテナンス)` (`webui.menu.Maintenance`)
- th: untranslated `**Maintenance**` → `การบำรุงรักษา`
- ja: `計算セッション` → `コンピュートセッション`; th: `เซสชันการคอมพิวเตอร์` / `เซสชันการประมวลผล` → `เซสชันการคำนวณ` (TERMINOLOGY `compute-session`)

Both Maintenance occurrences were fixed per file, including one outside the driving rows — leaving a single file naming the same page two different ways would have been worse than the slightly wider diff.

## Noted, not fixed

`補修(メンテナンス)` reads awkwardly in prose. If that label is itself considered wrong it needs an i18n FR, not a docs-side deviation from the live string.

## Review revision (2026-08-04)

Reviewed the ko chapter with 홍성철 and applied the confirmed edits (commit c99049ccb):

- **Ghost cross-references fixed (all 4 langs)**: the two `#connection-errors` links and the `#pages-you-cannot-open` sentence added in this PR pointed at sections that do not exist anywhere in the manual. The former now link to the [SSH/SFTP Connection to a Compute Session](#) chapter (`#ssh-sftp-container`); the latter sentence was removed. The HTML preview now reports zero broken links.
- **Offline banner entry simplified (all 4 langs)**: dropped the internal detection-logic explanation (browser-status + server probe) and the yellow slow-response banner paragraph.
- **Resource-allocation mismatch entry reverted (all 4 langs)**: restored the previous (main) content per review decision; the `session_detail_resource_allocation_warning.png` screenshots became unreferenced and were removed in all four languages.
- **ko only**: full-chapter prose pass (heavy pre-modifiers, connective-ending commas, repeated purpose phrasing, overused "따라서"); heading/anchor/label structure verified unchanged.


ironAiken2 (approved):
LGTM

PR #8574, opened 2026-08-04
fix(FR-3462): resolve partial default image references before launching a session

### LINKED ISSUE #8573: defaultImportEnvironment ignores the documented tag-less and arch-less formats in Start from URL launches
## Summary

`config.toml.sample` documents three supported formats for `defaultImportEnvironment` (full path with architecture, without architecture, without tag), but only the **full** form actually works. Every "Start from URL" launch path passes the configured string straight to session creation without ever resolving the partial forms, so a tag-less value such as `cr.backend.ai/stable/python` fails at launch time.

## Documented vs. actual

`config.toml.sample` (lines 6-12), written under `defaultImportEnvironment` by FR-1930:

```
# Supported formats:
#   - Full path with architecture: "registry/namespace:tag@arch"  → exact image
#   - Without architecture: "registry/namespace:tag"              → first available architecture
#   - Without tag: "registry/namespace"                           → latest version
```

FR-1930 implemented that matching in `ImageEnvironmentSelectFormItems.tsx` (`:273` exact → `:289` arch-less → `:313` tag-less → `images[0]`), which is reached only through the **session launcher form**. The import features never render that form.

## Failure chain

`ImportHuggingFaceModelForm.tsx:185`, `ImportRepoForm.tsx:308` and `ImportNotebookForm.tsx:83` all build the launcher value directly:

```
environments: { version: baiClient._config.default_import_environment }
```

and hand it to `startSessionWithDefault()`, which does no resolution:

```
// useStartSession.tsx:161
const imageFullName = values.environments.manual || values.environments.version;
const [kernelName, architecture] = imageFullName ? imageFullName.split('@') : ['', ''];
```

So for `cr.backend.ai/stable/python`:

1. `kernelName = "cr.backend.ai/stable/python"`, `architecture = undefined`
1. `createIfNotExists()` (`packages/backend.ai-client/src/client.ts:1355`) defaults architecture to `x86_64` and sends `lang` verbatim
1. Manager `ImageRef.parse_image_tag()` (`src/ai/backend/common/docker.py:506`) applies `tag = "latest"`
1. `cr.backend.ai/stable/python:latest` is not registered → session creation fails with an image-not-found error

## Note the asymmetry with `defaultSessionEnvironment`

`useStartSession.tsx:141` runs `default_session_environment` through `normalizeEnvironmentFormat()` (`:409`), which splits on the presence of `:` and routes a tag-less value to `environments.environment` (name-only → the select resolves the latest version). `default_import_environment` never goes through that function. The two keys therefore behave differently for identical input, despite the sample config documenting the same formats for both.

## Affected call sites

| Call site | Session type | Tag-less today |
|---|---|---|

|—|—|—|
|---|---|---|

| `ImportHuggingFaceModelForm.tsx:185` (experimental HF tab, FR-3390) | batch | fails |
|---|---|---|

| `ImportRepoForm.tsx:308` (GitHub / GitLab) | batch | fails |
|---|---|---|

| `ImportNotebookForm.tsx:83` (Notebook, immediate run) | interactive | fails |
|---|---|---|

| `ImportNotebookForm.tsx:170` ("Start with options") | — | works (goes through the launcher form) |
|---|---|---|

`e2e/utils/classes/session/SessionAPIHelper.ts:72` uses the same `split('@')` shortcut and shares the defect.

## Proposal

Resolve the configured import image against the registered image list before launching, honouring the three documented formats. `react/src/hooks/useDefaultImagesWithFallback.ts` is the established precedent for this: it resolves `defaultFileBrowserImage` / `systemSSHImage` against an images query, cached in a jotai atom and fetched via `fetchQuery` (non-suspending).

- Add a matching resolver for the import image and use it at the three import call sites.
- Keep a fully-qualified value a pass-through so current deployments see no behaviour change.
- Cover the three formats with unit tests.

## Out of scope

Validating that the configured image actually exists and surfacing it on the Diagnostics page is tracked separately in FR-3393.

JIRA Issue: FR-3462

yomybaby:
Resolves #8573 (FR-3462)

## Problem

`config.toml.sample` documents three formats for `defaultImportEnvironment`, but only the fully-qualified one worked:

```
# Supported formats:
#   - Full path with architecture: "registry/namespace:tag@arch"  → exact image
#   - Without architecture: "registry/namespace:tag"              → first available architecture
#   - Without tag: "registry/namespace"                           → latest version
```

FR-1930 implemented that matching in `ImageEnvironmentSelectFormItems` (exact → arch-less → tag-less → `images[0]`), which only runs when the **session launcher form** is rendered. The Start from URL import features build a launcher value and call `startSession()` directly, so they never reach it:

- `ImportHuggingFaceModelForm.tsx` (experimental HF tab, FR-3390) — batch
- `ImportRepoForm.tsx` (GitHub / GitLab) — batch
- `ImportNotebookForm.tsx` (Notebook, immediate run) — interactive

With `defaultImportEnvironment = "cr.backend.ai/stable/python"`, `startSession()` did `imageFullName.split('@')` and sent `lang` verbatim. The manager's `ImageRef.parse_image_tag()` then applies `tag = "latest"`, and `cr.backend.ai/stable/python:latest` is not registered — the launch fails.

## Fix

Resolve the reference **inside `startSession()`**, not at each call site. Every caller then behaves like the launcher form, and no import form needed to change.

This also fixes a latent gap: `startSession()` read only `environments.manual || environments.version` and ignored `environments.environment` — the name-only shape `normalizeEnvironmentFormat()` produces for a tag-less `defaultSessionEnvironment`. Reaching `startSession()` with that shape previously yielded an empty `kernelName`.

- `resolveImageStringFromImages()` (`react/src/helper/index.tsx`) — pure resolver implementing the three documented formats against an image list.
- `useImageReferenceResolver()` (`react/src/hooks/useDefaultImagesWithFallback.ts`) — wraps it with the images query, following the existing `useDefaultFileBrowserImageWithFallback` / `useDefaultSystemSSHImageWithFallback` pattern in the same file. Reuses `IMAGES_QUERY`, so **no new generated Relay artifacts**.
- `compareImageVersions()` moved out of `ImageEnvironmentSelectFormItems` into the helper, so the resolver and the launcher form sort versions with one shared definition instead of two copies that could drift.

### Behaviour

- A fully-qualified `registry/namespace:tag@arch` value short-circuits with **no image query** — existing deployments see no change and no extra request.
- Anything unresolvable is passed through unchanged, so the server keeps reporting the misconfiguration exactly as it does today.
- The arch-less and tag-less branches sort candidates before picking, so the choice does not depend on the order the server returned images in.

## Tests

14 new cases in `react/src/helper/index.test.tsx` covering all three formats, the deprecated `name` fallback, registry-with-port, prefix-collision safety, and empty/unmatched input. `react/src/helper/index.test.tsx`: 125 passed.

## Verification

> Note for reproducing locally in a background/agent shell: `relay-compiler` aborts when Watchman is started at a lower-than-normal nice value. Running with `/usr/local/bin` off `PATH` makes it fall back to a filesystem walk. This is environmental and unrelated to the change.

## Not covered here

- Validating that the configured image actually exists, surfaced on the Diagnostics page — tracked in FR-3393.
- `e2e/utils/classes/session/SessionAPIHelper.ts` uses the same `split('@')` shortcut inside `page.evaluate()`, where this hook cannot reach. Left as-is since e2e config always sets a full reference.

## What still needs a human check

I could not reach a live backend from this environment (the usual test servers were down), so the failing and fixed launches were **not** exercised end to end. Worth confirming against a cluster:

1. Set `defaultImportEnvironment = "cr.backend.ai/stable/python"` (no tag) and launch Start from URL → GitHub. It should now start with the latest registered `stable/python` image.
2. Set a full `...:3.13-ubuntu24.04-amd64@x86_64` value and confirm no behaviour change and no extra images query on launch.
3. Set a bogus `cr.backend.ai/missing/image` and confirm the server-side error still surfaces as before.


react/src/hooks/useDefaultImagesWithFallback.ts
```diff
@@ -111,6 +115,60 @@ export const useDefaultFileBrowserImageWithFallback = () => {
...
+      const response = await fetchQuery<useDefaultImagesWithFallbackQuery>(
+        relayEnv,
+        IMAGES_QUERY,
+        {
+          installed: true,
+        },
+        {
+          fetchPolicy: 'store-or-network',
+        },
+      ).toPromise();
```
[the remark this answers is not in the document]
yomybaby: Good catch — fixed in 5ff6b69. The resolver now builds its query variables the same way `ImageEnvironmentSelectFormItems` does:

```ts
const imageQueryVariables = baiClient?._config?.showNonInstalledImages
  ? {}
  : { installed: true };
```

Keeping the hardcoded `installed: true` would have reintroduced exactly the divergence this PR exists to remove — the two launch paths resolving the same config value to different images.

react/src/hooks/useStartSession.tsx
```diff
@@ -157,9 +159,17 @@ export const useStartSession = () => {
...
+      values.environments.manual ||
+      values.environments.version ||
+      values.environments.environment;
+    // Config values such as `defaultImportEnvironment` may omit the tag and/or
+    // the architecture. Resolve them against the registered image list here so
+    // every caller of `startSession` behaves like the launcher form (FR-3462).
+    const imageFullName = await resolveImageReference(requestedImage);
     const [kernelName, architecture] = imageFullName
       ? imageFullName.split('@')
       : ['', ''];
```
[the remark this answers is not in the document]
yomybaby: Not applying the suggestion — it would be a regression. `createIfNotExists()` declares `architecture: string = 'x86_64'` (packages/backend.ai-client/src/client.ts:1365), and a JS default parameter only fires on `undefined`. Coercing to `''` suppresses that default and sends a blank architecture to the manager.

So the `undefined` is load-bearing, not an oversight: a reference without `@arch` is meant to fall through to the client's `x86_64` default. That path predates this PR — the current sample config value `cr.backend.ai/multiarch/python:3.13-ubuntu24.04` has no `@` either.

You were right that the contract was unclear though, so 5ff6b69 makes it explicit rather than leaving it implied:

```ts
// CreateSessionInfo
// Left `undefined` when the image reference carries no `@arch` suffix, so
// that `createIfNotExists()` applies its own `x86_64` default. Passing an
// empty string instead would suppress that default and send a blank
// architecture to the manager.
architecture: string | undefined;

const [kernelName = '', architecture] = imageFullName?.split('@') ?? [];
```

The old `['', '']` fallback was the misleading part — it implied both were always strings. That's gone now.

react/src/helper/index.tsx
```diff
@@ -488,6 +488,115 @@ export const localeCompare = (a?: string | null, b?: string | null) => {
...
+ * @returns the full image name (`registry/namespace:tag@arch`), or `undefined`
+ *   when no registered image matches
+ */
+export const resolveImageStringFromImages = (
+  imageString: string | undefined | null,
+  images: ReadonlyArray<PartialImageRef | null | undefined> | undefined | null,
+): string | undefined => {
+  if (!imageString) return undefined;
+
+  const candidates = (images ?? []).filter(
```
agatha197, line 563: **Should Fix:** The candidate set here includes private images, but `ImageEnvironmentSelectFormItems` filters them out via `isPrivateImage` before matching. A config-driven import can therefore resolve to a private image the launcher form would never offer (e.g. a `stable/python` config value landing on someone's private customized build). Either apply the same private-image filter here (or in the caller before this function runs), or update the JSDoc above — right now it claims parity with the launcher form ("This mirrors the matching `ImageEnvironmentSelectFormItems` performs") that doesn't actually hold for this case.

react/src/hooks/useStartSession.tsx
```diff
@@ -157,12 +163,20 @@ export const useStartSession = () => {
...
-    const imageFullName =
-      values.environments.manual || values.environments.version;
-    const [kernelName, architecture] = imageFullName
-      ? imageFullName.split('@')
-      : ['', ''];
+    // If manual image is selected, use it as kernelName. `environment` holds a
+    // name-only reference (see `normalizeEnvironmentFormat`), which callers
+    // that never render the launcher form rely on us to resolve.
+    const requestedImage =
+      values.environments.manual ||
```
agatha197, line 174: **Should Fix:** `values.environments.manual` is the `allowManualImageNameForSession` escape hatch — the launcher form sets it precisely when the typed value did *not* match the image list. Routing it through `resolveImageReference` here means (a) an extra images query on every manual launch, and (b) if the typed value happens to look like a partial reference that *does* match something, we silently rewrite what the user explicitly typed. Suggest excluding `manual` from resolution:
```suggestion
    const imageFullName =
      values.environments.manual ||
      (await resolveImageReference(
        values.environments.version || values.environments.environment,
      ));
```

react/src/helper/index.test.tsx
```diff
@@ -1053,3 +1055,127 @@ describe('subNumberWithUnits', () => {
...
+    ).toBe('cr.backend.ai/stable/python:3.13-ubuntu24.04@aarch64');
+  });
+
+  it('should pick the latest version when the reference has no tag', () => {
+    expect(
+      resolveImageStringFromImages('cr.backend.ai/stable/python', images),
+    ).toBe('cr.backend.ai/stable/python:3.13-ubuntu24.04@aarch64');
+  });
+
+  it('should not match an environment that merely shares a name prefix', () => {
```
agatha197, line 1174: **Should Fix:** This test catches a `startsWith`-style bug in one direction (`stable/pyth` shouldn't match `stable/python`), but the fixtures don't actually assert the more dangerous direction: a longer namespace (`stable/python-ff`, present in the fixture above) leaking into a `stable/python` lookup. The "latest version" test above it would still pass even with a broken namespace filter, since `python-ff`'s only entry is `x86_64`, which loses to `python@aarch64` on the arch tiebreak regardless. Suggest adding a case where a broken filter would produce a *different, wrong* answer — e.g. give `stable/python-ff` a newer tag than `stable/python` and assert the lookup still picks `stable/python`'s version.

react/src/helper/index.tsx
```diff
@@ -488,6 +488,115 @@ export const localeCompare = (a?: string | null, b?: string | null) => {
...
+    const matched = _.first(
+      [...sameTag].sort((a, b) =>
+        localeCompare(a?.architecture, b?.architecture),
+      ),
+    );
+    return matched ? getImageFullName(matched) : undefined;
+  }
+
+  // 3. Neither tag nor architecture: take the latest version
+  if (!hasTag) {
```
agatha197, line 598: **Nice to have:** For a reference like `registry/namespace@arch` (arch but no tag), `hasArch` is parsed above but not consulted here — we fall into the "no tag" branch and can return an image with a *different* architecture than requested, silently. This is an undocumented format (not one of the three in `config.toml.sample`) and the launcher form has the same gap, so not a regression — but a silent architecture swap on a launch seems worth guarding. Either honor `architecture` in this branch when present, or return `undefined` for `hasArch && !hasTag` so the caller/server surfaces it as unresolved instead.

react/src/helper/index.tsx
```diff
@@ -488,6 +488,115 @@ export const localeCompare = (a?: string | null, b?: string | null) => {
...
+ * - `registry/namespace` → the latest version, first available architecture
+ *
+ * This mirrors the matching `ImageEnvironmentSelectFormItems` performs for the
+ * session launcher form, so launch paths that bypass that form (the Start from
+ * URL import features) resolve the same config value to the same image.
+ *
+ * @returns the full image name (`registry/namespace:tag@arch`), or `undefined`
+ *   when no registered image matches
+ */
+export const resolveImageStringFromImages = (
```
agatha197: **Nice to have:** `resolveImageStringFromImages` reads a little mechanical, and what it returns isn't quite an "image string" so much as a resolved full name. How about `resolveImageFullName(imageString, images)`? Would also line up nicely with the neighboring `getImageFullName` / `removeArchitectureFromImageFullName` :)

react/src/helper/index.tsx
```diff
@@ -856,11 +965,7 @@ export function isIpIncludedInList(ip: string, entries: string[]): boolean {
 }
 
 type SSEHandlerKeys =
```
agatha197, line 990: **Nice to have:** This reformat of `SSEHandlerKeys` looks unrelated to FR-3462 — likely picked up by `format-fix` on an unrelated Prettier version bump. Harmless, but worth dropping from this diff to keep the PR scoped to the image-resolution fix.

react/src/hooks/useDefaultImagesWithFallback.ts
```diff
@@ -111,6 +115,66 @@ export const useDefaultFileBrowserImageWithFallback = () => {
...
+        imageString
+      );
+    } catch {
+      // Keep the original value so the launch surfaces the server-side error
+      // instead of silently doing nothing.
+      return imageString;
+    }
+  };
+
+  return { resolveImageReference };
```
agatha197: **Nice to have:** The two sibling hooks in this file (`useDefaultFileBrowserImageWithFallback`, `useDefaultSystemSSHImageWithFallback`) return their value directly; this one returns `{ resolveImageReference }`. Either return the function bare, or rename to match what it returns — e.g. `useResolveImageReference()` returning the function itself reads a bit more directly than unwrapping a one-key object at each call site.

react/src/hooks/useDefaultImagesWithFallback.ts
```diff
@@ -111,6 +115,66 @@ export const useDefaultFileBrowserImageWithFallback = () => {
...
+        {
+          fetchPolicy: 'store-or-network',
+        },
+      ).toPromise();
+
+      return (
+        resolveImageStringFromImages(imageString, response?.images) ??
+        imageString
+      );
+    } catch {
```
agatha197: **Nice to have:** Two small ones:
- This hook sits between `useDefaultFileBrowserImageWithFallback` and `useDefaultSystemSSHImageWithFallback`, splitting up the two siblings that share a similar shape. Moving it above or below the pair would keep the file easier to scan.
- This `catch {}` swallows any network failure silently. A `useBAILogger` debug line here would help — right now a failed images query surfaces later as a confusing server-side launch error with no local trace of why resolution didn't happen.

react/src/helper/index.test.tsx
```diff
@@ -1053,3 +1055,127 @@ describe('subNumberWithUnits', () => {
...
+          registry: 'cr.backend.ai',
+          name: 'stable/python',
+          tag: '3.13-ubuntu24.04',
+          architecture: 'x86_64',
+        },
+      ]),
+    ).toBe('cr.backend.ai/stable/python:3.13-ubuntu24.04@x86_64');
+  });
+
+  it('should handle a registry with a port', () => {
```
agatha197, line 1260: **Nice to have:** This only covers the tag-less branch (`registry/namespace`). Port handling is trickiest for the `hasTag` detection (colon in the registry vs. colon before the tag), so `'127.0.0.1:5000/stable/python:3.13-ubuntu24.04'` (arch-less, with port) is probably the more valuable case to add.

Separately, `compareImageVersions` (describe block above) has no non-numeric case. `Number('ngc')` is `NaN`, and `NaN || 0` evaluates to `0`, so any non-numeric leading segment currently sorts as version `0` and can never win "latest". That's inherited launcher behavior, not introduced by this PR, but a test documenting it would help: `expect(compareImageVersions('2.1', 'ngc')).toBe(1)`.

config.toml.sample
```diff
@@ -3,7 +3,7 @@ apiEndpoint = "[Default API Endpoint. If blank, user input field will be shown.]
 apiEndpointText = "[Placeholder text instead of API endpoint input field.]"
 defaultSessionEnvironment = "[Default session kernel. If blank, alphabetically first kernel will be default.]"
 defaultImportEnvironment = "[Default kernel to use import features. If blank, cr.backend.ai/multiarch/python:3.9-ubuntu20.04 will be used.]"
-# Supported formats:
+# Supported formats (applies to both defaultSessionEnvironment and defaultImportEnvironment):
 #   - Full path with architecture: "registry/namespace:tag@arch" (e.g., "cr.backend.ai/multiarch/python:3.9-ubuntu20.04@x86_64")
 #     → Selects the exact image matching all components
 #   - Without architecture: "registry/namespace:tag" (e.g., "cr.backend.ai/multiarch/python:3.9-ubuntu20.04")
```
agatha197, line 9: **Nice to have:** Since this PR adds a real sort for "first available architecture" (previously undefined ordering), might be worth clarifying here that "first" now means alphabetically first (`aarch64` before `x86_64`), not server response order — that's the whole point of the sort added in this PR :)

agatha197 (commented):
Follow-up on the earlier review pass. The `showNonInstalledImages` divergence and the `architecture: ''` → `undefined` fix in `5ff6b69` address two issues raised earlier — thanks for the quick turnaround. Ran `pnpm vitest run src/helper` (129 passed) and `prettier --check` on the touched files locally against `5ff6b69`, both clean.

Remaining items below, split **Should Fix** / **Nice to Have**. Nothing here blocks merge on its own; still **Approve with suggestions** overall.

agatha197 (changes requested):
https://github.com/lablup/backend.ai-webui/pull/8574#pullrequestreview-4850674175 please check comment

react/src/helper/index.tsx
yomybaby: Fixed in cd2b9d9. `isPrivateImage` moved from `ImageEnvironmentSelectFormItems` into the helper module (single shared implementation), and `resolveImageFullName` now drops private images from its candidate set, so the JSDoc's parity claim holds. Added a test covering both directions: a private build can't win the latest-version pick and can't be resolved by its exact tag either.

react/src/hooks/useStartSession.tsx
yomybaby: Applied as suggested in cd2b9d9 — `manual` now bypasses resolution entirely and is used verbatim, so no extra images query on manual launches and no silent rewriting of what the user typed. The comment above the call site explains the escape-hatch semantics.

react/src/helper/index.test.tsx
yomybaby: Good point — the fixture couldn't distinguish a broken filter from a working one. In cd2b9d9 `stable/python-ff` now carries `9.99-ubuntu24.04` (newer than every `stable/python` tag) and a dedicated test asserts a `stable/python` lookup still returns python's own 3.13 while `stable/python-ff` returns 9.99 — a prefix-style filter now produces a different, wrong answer.

react/src/helper/index.tsx
yomybaby: Fixed in cd2b9d9 by honoring the architecture: the no-tag branch now restricts candidates to the requested arch when `@arch` is present, and returns `undefined` (→ unresolved, surfaced by the server) when that arch has no match, instead of silently swapping architectures. The JSDoc documents `registry/namespace@arch` as an accepted-but-undocumented format, and tests cover both the match and no-match cases.

react/src/helper/index.tsx
yomybaby: Renamed to `resolveImageFullName` in cd2b9d9 — agreed it lines up better with `getImageFullName` / `removeArchitectureFromImageFullName`.

react/src/helper/index.tsx
yomybaby: Reverted in cd2b9d9 — the `SSEHandlerKeys` hunk is back to main's formatting, keeping the diff scoped to the image-resolution fix.

react/src/hooks/useDefaultImagesWithFallback.ts
yomybaby: Done in cd2b9d9: renamed to `useResolveImageReference` and it now returns the function bare, so the call site reads `const resolveImageReference = useResolveImageReference();` with no one-key object unwrapping.

react/src/hooks/useDefaultImagesWithFallback.ts
yomybaby: Both applied in cd2b9d9: the hook now sits below the `useDefaultFileBrowserImageWithFallback` / `useDefaultSystemSSHImageWithFallback` pair so the two siblings stay adjacent, and the `catch` logs a `logger.debug` line (via `useBAILogger`) with the image string and error before falling back to the original value — a failed images query now leaves a local trace.

react/src/helper/index.test.tsx
yomybaby: Added both in cd2b9d9: the registry-with-port test now also covers `127.0.0.1:5000/stable/python:3.13-ubuntu24.04` (tag present, arch absent — the case where the registry colon could be mistaken for the tag separator), and `compareImageVersions` got a test documenting the inherited NaN→0 behavior (`compareImageVersions('2.1', 'ngc') === 1`), labeled as documented-not-endorsed.

config.toml.sample
yomybaby: Clarified in cd2b9d9 — `config.toml.sample` now states that "first available architecture" means alphabetically first (aarch64 before x86_64), not server response order.

agatha197 (changes requested):
please resolve merge conflicts

yomybaby:
@agatha197 Both outstanding requests are addressed:

**1. "please check comment" ([review #4850674175](https://github.com/lablup/backend.ai-webui/pull/8574#pullrequestreview-4850674175))** — all 9 points from that pass landed in `cd2b9d9` and each thread has an inline reply:

| Item | Resolution |
| --- | --- |
| Private images in the candidate set (Should Fix) | `isPrivateImage` moved into the helper module and applied in `resolveImageFullName`, so the JSDoc's parity claim with the launcher form now holds |
| `manual` routed through resolution (Should Fix) | `manual` bypasses resolution entirely and is used verbatim — no extra images query, no silent rewriting |
| Namespace-leak test not actually discriminating (Should Fix) | `stable/python-ff` now carries `9.99-ubuntu24.04`; a prefix-style filter produces a different, wrong answer |
| Silent architecture swap for `registry/namespace@arch` (Nice to have) | No-tag branch honors the requested arch and returns `undefined` when it has no match |
| `resolveImageStringFromImages` naming (Nice to have) | Renamed to `resolveImageFullName` |
| Unrelated `SSEHandlerKeys` reformat (Nice to have) | Reverted to main's formatting |
| Hook shape / naming (Nice to have) | `useResolveImageReference`, returns the function bare |
| Hook placement + silent `catch` (Nice to have) | Moved below the sibling pair; `catch` now logs via `useBAILogger` |
| Test coverage: port-with-tag, non-numeric versions (Nice to have) | Added `127.0.0.1:5000/stable/python:3.13-ubuntu24.04` and a test documenting the inherited `NaN → 0` behavior |

**2. "please resolve merge conflicts"** — resolved in `9a5bb93`. The only conflict was the import list at the top of `react/src/helper/index.test.tsx`, where this branch's `compareImageVersions` / `resolveImageFullName` collided with main's `convertFirstOrderByToString`; all three are kept. `react/src/helper/index.tsx` auto-merged cleanly and all four exports survived. Merged rather than rebased so the existing review threads stay anchored.

Verified locally against the merge result: `pnpm vitest run src/helper` → 374 passed (16 files), `tsc --noEmit -p react/tsconfig.json` → clean. `prettier --check` still flags `convertFirstOrderByToString` in both helper files, but that hunk comes from main untouched by this branch — same Prettier-version divergence you flagged on `SSEHandlerKeys`, so I left it alone rather than pulling an unrelated reformat into this diff.

PR is now `MERGEABLE`; ready for another look.

agatha197 (approved):
LGTM

PR #8582, opened 2026-08-04
fix(FR-3465): closed notifications reappear when a new notification is shown

### LINKED ISSUE #8581: Closed notifications reappear when a new notification is shown (antd 6.5 regression)
## Symptom

A notification that has already been closed (e.g. a **session notification that auto-closes after the session status changes to TERMINATED/CANCELLED**) reappears in the bottom-right toast area whenever any other notification is shown or updated.

## Root cause

Two pieces interact:

1. `useSetBAINotification().closeNotification(key)` only calls `app.notification.destroy(key)` and relies on antd firing the notice's `onClose` callback to flip `open: false` in `notificationListState` (the Jotai atom).
1. Since **FR-3330 (#8295)** bumped antd `6.3.5 → 6.5.0`, antd bundles **@rc-component/notification 2.x**, where the imperative `close(key)` was changed to silently remove the notice **without invoking `config.onClose`**:

```js
// @rc-component/notification 2.0.7 — Notifications.js
close: key => {
  setConfigList(list => list.filter(item => item.key !== key));
},
```

(In 1.2.0 / antd 6.3.x it was `close: key => { onNoticeClose(key); `}, which called `config.onClose?.()`.)

So after `closeNotification(key)` the atom entry keeps `open: true`. The reactive opener in `useBAINotificationEffect` (`processNotifications`) re-calls `app.notification.open()` for **every** `open === true` entry on **every** atom change — so the closed notification zombie-reopens as soon as any other notification appears. Exactly the reported behavior.

The duration-based auto close and the user clicking X still fire `onClose` in rc 2.x, which is why only imperative closes (`closeNotification` callers: session status auto-close in `BAIComputeSessionNodeNotificationItem`, VFolder items, `FolderExplorerModal`, `FileUploadManager`, notification click-through, …) are affected.

`clearNotification(key)` has the sibling leak: the key is never removed from `_activeNotificationKeys` because the `onClose` that used to do it no longer fires.

## Fix

Stop relying on antd's `onClose` for imperative closes — update the state explicitly:

- `closeNotification(key)`: remove the key from `_activeNotificationKeys`, call `app.notification.destroy(key)`, and set the entry's `open: false` in the atom.
- `clearNotification(key)`: remove the key from `_activeNotificationKeys` and destroy directly (instead of delegating to `closeNotification`, which would no-op on the already-removed entry).

Verified with unit tests (jsdom + real antd `App`): before the fix, `closeNotification` leaves `open: true` and the closed notice is re-opened by an unrelated state change; after the fix all paths (imperative close, duration auto close, clear) converge.

## Reproduction

1. Create a session → session notification appears.
1. Terminate the session → notification auto-closes ~3s after TERMINATED.
1. Trigger any other notification (e.g. create a folder).
1. The closed session notification reappears next to the new one.

JIRA Issue: FR-3465

yomybaby:
Resolves #8581 (FR-3465)

## Symptom

A notification that has already been closed — e.g. a **session notification that auto-closes ~3s after the session status changes to TERMINATED/CANCELLED** — reappears in the bottom-right toast area whenever any other notification is shown or updated.

## Root cause

`useSetBAINotification().closeNotification(key)` only called `app.notification.destroy(key)` and relied on antd firing the notice's `onClose` callback to flip `open: false` in `notificationListState`.

Since FR-3330 (#8295) bumped antd `6.3.5 → 6.5.0`, antd bundles **@rc-component/notification 2.x**, where the imperative `close(key)` no longer invokes `config.onClose`:

```js
// @rc-component/notification 2.0.7 — Notifications.js
close: key => {
  setConfigList(list => list.filter(item => item.key !== key));
},
```

(In 1.2.0 / antd 6.3.x this was `close: key => { onNoticeClose(key); }`, which called `config.onClose?.()`. The duration-based auto close and the X button still fire `onClose` in 2.x — only imperative closes are affected.)

So after `closeNotification(key)` the atom entry stayed `open: true`, and the reactive opener in `useBAINotificationEffect` re-calls `app.notification.open()` for every `open === true` entry on every state change — the closed notification zombie-reopens as soon as any other notification appears.

## Changes

- `closeNotification(key)` now removes the key from `_activeNotificationKeys`, destroys the antd notice, **and marks the state entry `open: false`** itself instead of relying on antd's `onClose`.
- `clearNotification(key)` likewise cleans `_activeNotificationKeys` and destroys directly (delegating to `closeNotification` would no-op on the just-removed entry).
- Adds regression tests (`react/src/hooks/__tests__/useBAINotification.test.tsx`) rendering the real antd `App` + `useBAINotificationEffect`:
  - imperative close flips `open` to `false` (fails on the previous code)
  - a closed notification is not re-opened by an unrelated state change (fails on the previous code)
  - duration-based auto close still syncs state via `onClose`
  - `clearNotification` removes the entry and it is not resurrected

## Verification

- `npx vitest run src/hooks/__tests__/useBAINotification.test.tsx` → 4 passed (the two regression cases fail against the previous implementation)

### DIFF
--- a/react/src/hooks/__tests__/useBAINotification.test.tsx
+++ b/react/src/hooks/__tests__/useBAINotification.test.tsx
@@ -1,0 +1,187 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+import {
+  useBAINotificationEffect,
+  useBAINotificationState,
+} from '../useBAINotification';
+import { act, render, screen, waitFor } from '@testing-library/react';
+import { App } from 'antd';
+import React from 'react';
+
+vi.mock('..', () => ({
+  useWebUINavigate: () => vi.fn(),
+}));
+
+vi.mock('../useBAISetting', () => ({
+  useBAISettingUserState: () => [false, vi.fn()],
+}));
+
+vi.mock('../../helper', () => ({
+  listenToBackgroundTask: vi.fn(),
+}));
+
+vi.mock('react-i18next', () => ({
+  useTranslation: () => ({ t: (k: string) => k }),
+}));
+
+vi.mock('../../components/BAINodeNotificationItem', () => ({
+  default: () => <div>node-item</div>,
+}));
+vi.mock('../../components/BAIMultiStepNotificationItem', () => ({
+  default: () => <div>multistep-item</div>,
+}));
+vi.mock('../../components/BAIGeneralNotificationItem', () => ({
+  default: ({ notification }: any) => (
+    <div data-testid={`general-item-${notification.key}`}>
+      {String(notification.description ?? '')}
+    </div>
+  ),
+}));
+
+const Effects: React.FC = () => {
+  useBAINotificationEffect();
+  return null;
+};
+
+let latestState: ReturnType<typeof useBAINotificationState>[0] | null = null;
+let latestSetters: ReturnType<typeof useBAINotificationState>[1] | null = null;
+
+const StateProbe: React.FC = () => {
+  const [notifications, setters] = useBAINotificationState();
+  React.useEffect(() => {
+    latestState = notifications;
+    latestSetters = setters;
+  });
+  return null;
+};
+
+const Harness: React.FC = () => (
+  <App>
+    <Effects />
+    <StateProbe />
+  </App>
+);
+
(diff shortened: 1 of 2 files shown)


agatha197 (approved):
LGTM

PR #8588, opened 2026-08-04
feat(FR-3387): pilot useKeyedSnapshot render-as-you-fetch tab state on the users page

### LINKED ISSUE #8418: Wire useTabQuerySnapshot with preloaded queries for render-as-you-fetch tab switching
## Problem

`useTabQuerySnapshot` (tab-url-state pattern A-ii) restores a tab s raw query string via `navigate({ search })`, but never hands that string back to the caller.

Hosts that own preloaded queries (`useQueryLoader`) need the target tab s parsed filter/order/pagination **synchronously inside the tab-change click handler** in order to start the fetch (render-as-you-fetch). Since `navigate()` only schedules the location update, the host cannot read the new search until a later render. The host is then forced to either load with the departing tab s variables, or defer the load to an effect and lose render-as-you-fetch.

## Proposed direction

- Add an action/callback parameter so the hook passes the restored query string and target tab key back at click time: `onTabChange(key)` calls `navigate({ search })` and then `action?.(key, search)`.
- Document the precondition that the host must not write the snapshotted keys through nuqs setters. nuqs flushes asynchronously, re-reads `location.search` at flush time, then merges only its own keys — so a write queued by the departing tab lands after the tab-change `navigate` and leaks that tab s filter/order into the new tab s URL.
- Decide whether hosts adopting this path should drop nuqs for the snapshotted keys entirely (single writer via navigate), and weigh that cost against staying on pattern A-i.

## Notes

- Wrapping the load in `startTransition` is not a substitute. With the tab flip urgent and the load deferred, the null-queryRef placeholder and the Suspense fallback are the same skeleton, so it only adds a render pass with no visual gain.
- `AdminDeploymentPage` currently stays on pattern A-i for exactly this reason, and is the concrete case that motivated this issue.

—

Captured while working on branch: fr-3256-tab-naming

JIRA Issue: FR-3387

nowgnuesLee:
Resolves #8418 (FR-3387)

## Summary

Pilot of a page-level keyed-snapshot tab-state architecture on the admin users page (Users / Credentials tabs). The page becomes the single owner of URL state and query fetching; the tab components become preloaded-query consumers that own no query-string state.

### Hooks (`react/src/hooks/index.tsx`)

- **`useKeyedSnapshot(sourceKey, value)`** — per-key snapshots of a caller-defined value, returned as `[currentKey, setKey, peekSnapshot]`. The caller decides the snapshot shape (here `{ queryParams, tablePaginationOption }`, seeded from the query string on the first render). `peekSnapshot(key)` is a pure read of the stored value (`undefined` if never visited); `setKey(nextKey)` snapshots the departing key and switches — so a tab click can restore state and start its query inside the same event handler (render-as-you-fetch). The pair started as one `setAfterSnapshot(nextKey)` that switched *and* returned, but a `set*` that callers depend on for its return value contradicts the setter convention, so it was split along the setter/getter seam (`peek` over `get` to avoid the false friend with `useSyncExternalStore`'s no-arg, current-state `getSnapshot`). The first argument also re-syncs the key when it changes, which is how the tab UI follows browser back/forward.
- **`useBrowserPopstateEffect(onSettled)`** — runs a callback after a browser back/forward, **once nuqs' query state matches the address bar**. nuqs applies `popstate` inside a `startTransition`, so for at least one render every `useQueryStates` value still describes the page the user just left; the hook arms a ref on `popstate` and holds the callback until `useOptimisticSearchParams()` and `window.location.search` normalize equal, then disarms. That guard is what lets callers read their nuqs state inside the callback instead of re-parsing `location.search` themselves. `pushState`/`replaceState` emit no `popstate`, so the page's own URL writes never arm it.

### `AdminUsersPage`

- One page-level `useQueryStates` group (plus the existing pagination hook) mirrors the active tab's state — **push** on tab switch, **replace** within a tab, matching what `useTabQuerySnapshot` did before.
- Two `useQueryLoader` refs; a tab click restores its snapshot and loads immediately.
- Mount / direct URL entry / reload load once; back/forward reload through `useBrowserPopstateEffect`.

### Children

`AdminUserManagement` and `AdminUserCredentialList` drop every piece of URL state — nuqs groups, pagination-on-search-param, and `fetchKey`. They receive `queryRef` + `onReload(variables, options)` (the `LoginSession` / `LoginHistory` contract): displayed values derive from `queryRef.variables`, pending UX from a deferred queryRef, and refresh is a network-only reload of the current variables.

The credential list's `?action=add` opener (an effect that opened the keypair-creation modal from the query string, then cleared the param) is **removed, not relocated**. Nothing in the repo produces that URL — it had no reachable consumer — so the modal is now opened only by the Add button on screen. Flagging it explicitly since it was a deep link an external doc could in principle have linked to; if one does, it should come back as a page-level read rather than child-owned query state.

### Follow-ups folded in after the first review

- **Back/forward now reloads.** The tab flipped but the query ref kept the departed tab's rows. The reload runs through `useBrowserPopstateEffect`, so by the time it builds its snapshot the page's nuqs state already describes the entry the user navigated to — no second URL parser, and no callback firing against stale params.
- **TOTP capability resolved before the first load.** `isTOTPSupported` was `undefined` on the first render, so the one-shot load sent `isNotSupportTotp: true` and never re-issued. The page now reads it through a new suspending `useSuspendedTOTPSupported` sharing the same TanQuery key; `useTOTPSupported` is untouched for its `useLazyLoadQuery` callers.
- **`@skipOnClient` no longer defeats the Relay store.** The directive strips those fields from the request text while the compiled operation still expected them, so the users query was permanently "missing" in the store and every `store-or-network` load hit the network — a skeleton on every visit to the tab. Pairing it with the standard `@skip(if:)` puts a `Condition` node in the normalization AST, so the store can satisfy the query. Also benefits `ForceTOTPCheckerQuery`, `ProjectAdminUsersPageQuery` and `UserDropdownMenuQuery`, which spread the same fragments.
- **Page-driven loads are `store-and-network`.** `loadQuery` defaults to `store-or-network`, which — with the store now able to satisfy these queries — would serve any variable combination the user has already seen without revalidating it (back to page 1, a tab round-trip, a user created on the other tab). `DEFAULT_LOAD_OPTIONS` sets `store-and-network` for the tab-entry path and for child-driven variable changes; the children still override with `network-only` for a manual refresh or after their own mutation. This costs no skeleton — with a complete cached record `usePreloadedQuery` reads synchronously and does not suspend.
- `convertFirstOrderByToString` from main (FR-3256) is reused instead of the branch's own inverse helper, resolved during the merge.

`useTabQuerySnapshot` and the other 8 pages are untouched; rollout is a follow-up pending this pilot's review.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Terminology)
- `vitest run src/hooks` → 26 files, 349 tests passing, including `renderHook` cases for both hooks:
  - `useKeyedSnapshot.test.tsx` — seeds the initial key from the first rendered value / snapshots the departing key and peeks the target's / peeks the live value for the current key / `undefined` for a never-visited key without switching / holds the new key while `sourceKey` still reports the departing one / re-syncs when `sourceKey` changes.
  - `useBrowserPopstateEffect.test.tsx` — silent while nuqs has not caught up with the address bar / fires exactly once on the render it catches up (and does not replay) / silent for the page's own `setQueryParams` writes.
- **Manual QA against a live cluster**, driven through Chrome DevTools on the reported sequence — users(INACTIVE) → credentials → users → filter → Back ×2 → Forward ×2:

  | Step | URL | Rendered filter |
  |---|---|---|
  | filter applied | `…&status=INACTIVE&filter={…zzz}` | `email ⊃ zzz` |
  | Back ×1 | `?tab=credentials&pageSize=20` | — |
  | Back ×2 | `?status=INACTIVE` | none |
  | Forward ×1 | `?tab=credentials&pageSize=20` | — |
  | Forward ×2 | `…&status=INACTIVE&filter={…zzz}` | `email ⊃ zzz` |

  URL and rendered state agree at every step, per-tab state survives tab round-trips, and a filter change is mirrored to the URL.

- Still worth a human pass: revisiting a tab should paint from cache with no skeleton while it revalidates, and the cold-start deep link where the TOTP capability query has not warmed yet.

https://claude.ai/code/session_01Bqm1FB6FRr9H7C1yCD6CTK

### DIFF
--- a/react/src/components/AdminUserCredentialList.tsx
+++ b/react/src/components/AdminUserCredentialList.tsx
@@ -88,6 +88,8 @@ type Keypair = NonNullable<
   >['items'][number]
 >;
 
+export const CREDENTIAL_LIST_DEFAULT_PAGE_SIZE = 20;
+
 interface AdminUserCredentialListProps {
   queryRef: PreloadedQuery<AdminUserCredentialListQueryType>;
   onReload: (
@@ -124,7 +126,7 @@ const AdminUserCredentialList: React.FC<AdminUserCredentialListProps> = ({
   // funnels changes back through `onReload`.
   const variables = queryRef.variables;
   const activeType = variables.is_active === false ? 'inactive' : 'active';
-  const pageSize = variables.limit ?? 20;
+  const pageSize = variables.limit ?? CREDENTIAL_LIST_DEFAULT_PAGE_SIZE;
   const current = Math.floor((variables.offset ?? 0) / pageSize) + 1;
 
   const deferredQueryRef = useDeferredValue(queryRef);
--- a/react/src/components/AdminUserCredentialList.tsx
+++ b/react/src/components/AdminUserCredentialList.tsx
@@ -5,13 +5,11 @@
 import { AdminUserCredentialListDeleteMutation } from '../__generated__/AdminUserCredentialListDeleteMutation.graphql';
 import { AdminUserCredentialListModifyMutation } from '../__generated__/AdminUserCredentialListModifyMutation.graphql';
 import {
-  AdminUserCredentialListQuery,
+  AdminUserCredentialListQuery as AdminUserCredentialListQueryType,
   AdminUserCredentialListQuery$data,
-  AdminUserCredentialListQuery$variables,
 } from '../__generated__/AdminUserCredentialListQuery.graphql';
 import { KeypairSettingModalFragment$key } from '../__generated__/KeypairSettingModalFragment.graphql';
 import { App } from '../app-shim';
-import { useBAIPaginationOptionStateOnSearchParam } from '../hooks/reactPaginationQueryOptions';
 import { theme } from '../theme-shim';
 import BAIRadioGroup from './BAIRadioGroup';
 import KeypairInfoModal from './KeypairInfoModal';
@@ -30,9 +28,7 @@ import {
   BAIDeleteConfirmModal,
   BAISelectionLabel,
   useBAILogger,
-  useUpdatableState,
   BAIText,
-  INITIAL_FETCH_KEY,
   PRIMARY_TAG_VARIANT,
 } from 'backend.ai-ui';
 import dayjs from 'dayjs';
@@ -46,10 +42,53 @@ import {
   SquarePenIcon,
   UndoIcon,
 } from 'lucide-react';
-import { parseAsString, useQueryState, useQueryStates } from 'nuqs';
-import { useDeferredValue, useEffect, useState, useTransition } from 'react';
+import { useDeferredValue, useState, useTransition } from 'react';
 import { useTranslation } from 'react-i18next';
-import { graphql, useLazyLoadQuery, useMutation } from 'react-relay';
+import {
+  graphql,
+  useMutation,
+  usePreloadedQuery,
+  type PreloadedQuery,
+  type UseQueryLoaderLoadQueryOptions,
+} from 'react-relay';
+
+export const AdminUserCredentialListQuery = graphql`
+  query AdminUserCredentialListQuery(
+    $limit: Int!
+    $offset: Int!
+    $filter: String
+    $order: String
+    $domain_name: String
+    $email: String
+    $is_active: Boolean
+  ) {
+    keypair_list(
+      limit: $limit
+      offset: $offset
+      filter: $filter
+      order: $order
+      domain_name: $domain_name
+      email: $email
+      is_active: $is_active
+    ) {
+      items {
+        id
+        user_id
+        access_key
+        is_admin
+        resource_policy
+        created_at
+        rate_limit
+        num_queries
+        concurrency_used @since(version: "24.09.0")
+
+        ...KeypairSettingModalFragment
+        ...KeypairInfoModalFragment
+      }
+      total_count
+    }
+  }
+`;
 
 type Keypair = NonNullable<
   NonNullable<
@@ -57,23 +96,25 @@ type Keypair = NonNullable<
   >['items'][number]
 >;
 
-const AdminUserCredentialList: React.FC = () => {
+export const CREDENTIAL_LIST_DEFAULT_PAGE_SIZE = 20;
+
+interface AdminUserCredentialListProps {
+  queryRef: PreloadedQuery<AdminUserCredentialListQueryType>;
+  onReload: (
+    variables: AdminUserCredentialListQueryType['variables'],
+    options?: UseQueryLoaderLoadQueryOptions,
+  ) => void;
+}
+
+const AdminUserCredentialList: React.FC<AdminUserCredentialListProps> = ({
+  queryRef,
+  onReload,
+}) => {
+  'use memo';
   const { t } = useTranslation();
   const { token } = theme.useToken();
   const { message, modal } = App.useApp();
   const { logger } = useBAILogger();
-
-  const [action, setAction] = useQueryState('action', parseAsString);
-
-  const [fetchKey, updateFetchKey] = useUpdatableState('first');
-  const [queryParams, setQueryParams] = useQueryStates(
-    {
-      activeType: parseAsString.withDefault('active'),
-      order: parseAsString,
-      filter: parseAsString,
-    },
-    { history: 'replace' },
-  );
 
   const [keypairSettingModalFrgmt, setKeypairSettingModalFrgmt] =
     useState<KeypairSettingModalFragment$key | null>(null);
@@ -88,71 +129,20 @@ const AdminUserCredentialList: React.FC = () => {
   const [selectedKeypairs, setSelectedKeypairs] = useState<Keypair[]>([]);
   const [isBulkUpdating, setIsBulkUpdating] = useState(false);
 
-  const {
-    baiPaginationOption,
-    tablePaginationOption,
-    setTablePaginationOption,
-  } = useBAIPaginationOptionStateOnSearchParam({
-    current: 1,
-    pageSize: 20,
-  });
+  // The page owns URL state and the fetch; this component derives every
+  // filter/order/pagination control from the last requested variables and
+  // funnels changes back through `onReload`.
+  const variables = queryRef.variables;
+  const activeType = variables.is_active === false ? 'inactive' : 'active';
+  const pageSize = variables.limit ?? CREDENTIAL_LIST_DEFAULT_PAGE_SIZE;
+  const current = Math.floor((variables.offset ?? 0) / pageSize) + 1;
 
-  const queryVariables: AdminUserCredentialListQuery$variables = {
-    limit: baiPaginationOption.limit,
-    offset: baiPaginationOption.offset,
-    is_active: queryParams.activeType === 'active',
-    filter: queryParams.filter,
-    order: queryParams.order,
-  };
-  const deferredQueryVariables = useDeferredValue(queryVariables);
-  const deferredFetchKey = useDeferredValue(fetchKey);
+  const deferredQueryRef = useDeferredValue(queryRef);
+  const isPending = deferredQueryRef !== queryRef;
 
-  const { keypair_list } = useLazyLoadQuery<AdminUserCredentialListQuery>(
-    graphql`
-      query AdminUserCredentialListQuery(
-        $limit: Int!
-        $offset: Int!
-        $filter: String
-        $order: String
-        $domain_name: String
-        $email: String
-        $is_active: Boolean
-      ) {
-        keypair_list(
-          limit: $limit
-          offset: $offset
-          filter: $filter
-          order: $order
-          domain_name: $domain_name
-          email: $email
-          is_active: $is_active
-        ) {
-          items {
-            id
-            user_id
-            access_key
-            is_admin
-            resource_policy
-            created_at
-            rate_limit
-            num_queries
-            concurrency_used @since(version: "24.09.0")
-
-            ...KeypairSettingModalFragment
-            ...KeypairInfoModalFragment
-          }
-          total_count
-        }
-      }
-    `,
-    deferredQueryVariables,
-    {
-      fetchKey: deferredFetchKey,
-      fetchPolicy:
-        deferredFetchKey === INITIAL_FETCH_KEY
-          ? 'store-and-network'
-          : 'network-only',
-    },
+  const { keypair_list } = usePreloadedQuery<AdminUserCredentialListQueryType>(
+    AdminUserCredentialListQuery,
+    deferredQueryRef,
   );
 
   const [commitModifyKeypair] =
@@ -248,31 +238,22 @@ const AdminUserCredentialList: React.FC = () => {
           );
         }
         setSelectedKeypairs([]);
-        updateFetchKey();
+        onReload(variables, { fetchPolicy: 'network-only' });
       },
     });
   };
-
-  useEffect(() => {
-    if (action === 'add') {
-      // eslint-disable-next-line react-hooks/set-state-in-effect
-      setOpenUserKeypairSettingModal(true);
-
-      setAction(null);
-    }
-  }, [action, setAction]);
 
   return (
     <BAIFlex direction="column" align="stretch" gap="sm">
       <BAIFlex justify="between" align="start" gap="xs" wrap="wrap">
         <BAIFlex gap={'sm'} align="start">
           <BAIRadioGroup
-            value={queryParams.activeType}
+            value={activeType}
             onChange={(value) => {
-              setQueryParams({ activeType: value.target.value });
-              setTablePaginationOption({
-                current: 1,
-                pageSize: tablePaginationOption.pageSize,
+              onReload({
+                ...variables,
+                is_active: value.target.value === 'active',
+                offset: 0,
               });
               setSelectedKeypairs([]);
             }}
@@ -323,9 +304,13 @@ const AdminUserCredentialList: React.FC = () => {
                 type: 'string',
               },
             ]}
-            value={queryParams.filter ?? undefined}
+            value={variables.filter ?? undefined}
             onChange={(value) => {
-              setQueryParams({ filter: value ?? null });
+              onReload({
+                ...variables,
+                filter: value ?? null,
+                offset: 0,
+              });
               setSelectedKeypairs([]);
             }}
           />
@@ -337,7 +322,7 @@ const AdminUserCredentialList: React.FC = () => {
                 count={selectedKeypairs.length}
                 onClearSelection={() => setSelectedKeypairs([])}
               />
-              {queryParams.activeType === 'active' ? (
+              {activeType === 'active' ? (
                 <Tooltip content={t('credential.Deactivate')}>
                   <BAIButton
                     icon={<BanIcon style={{ color: token.colorError }} />}
@@ -358,9 +343,9 @@ const AdminUserCredentialList: React.FC = () => {
           )}
           <Tooltip content={t('button.Refresh')}>
             <BAIButton
-              loading={deferredFetchKey !== fetchKey}
+              loading={isPending}
               onClick={() => {
-                updateFetchKey();
+                onReload(variables, { fetchPolicy: 'network-only' });
               }}
               icon={<RotateCw size="1em" />}
             />
@@ -378,7 +363,7 @@ const AdminUserCredentialList: React.FC = () => {
       </BAIFlex>
       <BAITableAstryx<Keypair>
         rowKey={'id'}
-        loading={deferredQueryVariables !== queryVariables}
+        loading={isPending}
         dataSource={filterOutNullAndUndefined(keypair_list?.items)}
         rowSelection={{
           type: 'checkbox',
@@ -432,7 +417,7 @@ const AdminUserCredentialList: React.FC = () => {
                     });
                   },
                 },
-                ...(queryParams.activeType === 'inactive'
+                ...(activeType === 'inactive'
                   ? [
                       {
                         key: 'activate',
@@ -467,7 +452,9 @@ const AdminUserCredentialList: React.FC = () => {
                                         kp.access_key !== record.access_key,
                                     ),
                                   );
-                                  updateFetchKey();
+                                  onReload(variables, {
+                                    fetchPolicy: 'network-only',
+                                  });
                                   resolve();
                                 },
                                 onError: (error) => {
@@ -482,7 +469,7 @@ const AdminUserCredentialList: React.FC = () => {
                       },
                     ]
                   : []),
-                ...(queryParams.activeType === 'active'
+                ...(activeType === 'active'
                   ? [
                       {
                         key: 'deactivate',
@@ -521,7 +508,9 @@ const AdminUserCredentialList: React.FC = () => {
                                         kp.access_key !== record.access_key,
                                     ),
                                   );
-                                  updateFetchKey();
+                                  onReload(variables, {
+                                    fetchPolicy: 'network-only',
+                                  });
                                   resolve();
                                 },
                                 onError: (error) => {
@@ -631,22 +620,26 @@ const AdminUserCredentialList: React.FC = () => {
           },
         ])}
         pagination={{
-          pageSize: tablePaginationOption.pageSize,
+          pageSize,
           total: keypair_list?.total_count || 0,
-          current: tablePaginationOption.current,
+          current,
           // TODO: need to set more options to export CSV in current page's data
-          onChange(current, pageSize) {
-            if (_.isNumber(current) && _.isNumber(pageSize)) {
-              setTablePaginationOption({
-                current,
-                pageSize,
+          onChange(nextCurrent, nextPageSize) {
+            if (_.isNumber(nextCurrent) && _.isNumber(nextPageSize)) {
+              onReload({
+                ...variables,
+                limit: nextPageSize,
+                offset: (nextCurrent - 1) * nextPageSize,
               });
               setSelectedKeypairs([]);
             }
           },
         }}
         onChangeOrder={(nextOrder) => {
-          setQueryParams({ order: nextOrder ?? null });
+          onReload({
+            ...variables,
+            order: nextOrder ?? null,
+          });
           setSelectedKeypairs([]);
         }}
       />
@@ -670,7 +663,7 @@ const AdminUserCredentialList: React.FC = () => {
           setKeypairSettingModalFrgmt(null);
           setOpenUserKeypairSettingModal(false);
           if (success) {
-            updateFetchKey();
+            onReload(variables, { fetchPolicy: 'network-only' });
           }
         }}
       />
@@ -709,7 +702,7 @@ const AdminUserCredentialList: React.FC = () => {
                   ),
                 );
                 setDeletingKeypair(null);
-                updateFetchKey();
+                onReload(variables, { fetchPolicy: 'network-only' });
               },
               onError: (error) => {
                 message.error(error?.message);
--- a/react/src/components/AdminUserCredentialList.tsx
+++ b/react/src/components/AdminUserCredentialList.tsx
@@ -5,12 +5,10 @@
 import { AdminUserCredentialListDeleteMutation } from '../__generated__/AdminUserCredentialListDeleteMutation.graphql';
 import { AdminUserCredentialListModifyMutation } from '../__generated__/AdminUserCredentialListModifyMutation.graphql';
 import {
-  AdminUserCredentialListQuery,
+  AdminUserCredentialListQuery as AdminUserCredentialListQueryType,
   AdminUserCredentialListQuery$data,
-  AdminUserCredentialListQuery$variables,
 } from '../__generated__/AdminUserCredentialListQuery.graphql';
 import { KeypairSettingModalFragment$key } from '../__generated__/KeypairSettingModalFragment.graphql';
-import { useBAIPaginationOptionStateOnSearchParam } from '../hooks/reactPaginationQueryOptions';
 import BAIRadioGroup from './BAIRadioGroup';
 import KeypairInfoModal from './KeypairInfoModal';
 import KeypairSettingModal from './KeypairSettingModal';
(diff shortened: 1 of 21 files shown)


react/src/pages/AdminUsersPage.tsx
```diff
@@ -2,25 +2,244 @@
...
+    };
+    if (currentTab === 'users') {
+      loadUsersQuery(usersVariablesOf(snapshot));
+    } else {
+      loadCredentialsQuery(credentialsVariablesOf(snapshot));
+    }
+  });
+  useEffect(() => {
+    loadInitialTabFromUrl();
+  }, []);
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in 254f3b9de. Added `useBrowserNavigatedQueryEffect`: a `popstate` listener arms a ref, and an effect runs the callback on the next commit when the watched values actually changed. `AdminUsersPage` wires it to rebuild the snapshot from the popped URL and load that tab’s query.

No de-duplication guard is needed against the click-time preload: `history.pushState` / `replaceState` never emit `popstate` (verified in nuqs’s patched history and react-router’s navigate path), so neither the tab-click nor the child `onReload` path can arm the effect.

react/src/pages/AdminUsersPage.tsx
```diff
@@ -2,25 +2,244 @@
...
+        ..._.omit(filter ?? {}, 'status'),
+        status:
+          params.status === 'ACTIVE'
+            ? { equals: 'ACTIVE' }
+            : { notEquals: 'ACTIVE' },
+      },
+      orderBy: convertToOrderBy<Required<UserV2OrderBy>>(order),
+      limit: pagination.pageSize,
+      offset: (pagination.current - 1) * pagination.pageSize,
+      isNotSupportTotp: !isTOTPSupported,
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in 254f3b9de by making the capability resolved *before* the page renders, so the mount-only load stays correct: `AdminUsersPage` now reads it through a new `useSuspendedTOTPSupported`, which suspends on the same TanQuery key (`isManagerSupportingTOTP`) the app shell already warms. `useTOTPSupported` is untouched — its `useLazyLoadQuery` callers re-fetch on their own when the variable settles, and they must stay non-blocking.

The same review pass turned up a second, related defect on those fields: `@skipOnClient` strips them from the request text, but Relay’s compiled operation still expected them, so the store recorded them as permanently missing and every `store-or-network` load went to the network — the users tab showed a skeleton on *every* visit. Pairing it with the standard `@skip(if: $isNotSupportTotp)` puts a `Condition` node in the normalization AST, so the store is complete and cache hits work.

react/src/hooks/index.tsx
```diff
@@ -68,6 +68,57 @@ export const useTabQuerySnapshot = <T extends string>(
...
+ *
+ * `setAfterSnapshot(nextKey)` snapshots the current key's value, switches the
+ * key, and synchronously returns the value stored for `nextKey` (`undefined`
+ * if never visited) — so the caller can start a preloaded query and mirror
+ * the URL inside the same event handler (render-as-you-fetch).
+ */
+export const useKeyedSnapshot = <K extends string, V>(
+  sourceKey: K,
+  value: V,
+): [K, (nextKey: K) => V | undefined] => {
```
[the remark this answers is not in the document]
nowgnuesLee: Added in 254f3b9de: `react/src/hooks/useKeyedSnapshot.test.tsx` and `react/src/hooks/useBrowserNavigatedQueryEffect.test.tsx`, six `renderHook` cases in total.

`useKeyedSnapshot` — seeds the initial key from the first rendered value; snapshots the departing key and returns the target key’s value; returns `undefined` for a never-visited key; re-syncs the current key when the source key changes (and holds still when it does not, which is the no-bounce case you asked for).

`useBrowserNavigatedQueryEffect` — fires once after a `popstate` that changed the values and consumes the navigation; stays silent when the app itself changes them.

agatha197 (changes requested):
페이지가 URL·fetch의 단일 소유자가 되고 탭 컴포넌트는 preloaded query 소비자로 내려가는 방향 자체는 아주 좋습니다. `AdminUserManagement` / `AdminUserCredentialList`에서 nuqs 그룹·`useFetchKey`·중복 pagination 훅이 전부 사라지고 `queryRef.variables` 단일 출처로 정리된 게 이 PR의 핵심 가치라고 봅니다. `@skipOnClient`가 정규화 AST에서 필드를 지우지 못해 store가 영구 miss 상태였다는 발견은 이 PR 범위를 넘어 `ForceTOTPCheckerQuery` / `ProjectAdminUsersPageQuery` / `UserDropdownMenuQuery`까지 같이 살리는 좋은 캐치입니다.

머지 전에 정리하고 갔으면 하는 것들입니다.

---

## 1. `useBrowserPopstateEffect`에 테스트가 없습니다 (블로킹)

Copilot의 커버리지 지적에 대한 답글에서 `useKeyedSnapshot.test.tsx`와 **`useBrowserNavigatedQueryEffect.test.tsx` 두 개를 추가했다("6 renderHook cases")**고 하셨고 PR 본문도 같은 주장("new `renderHook` cases for both hooks ... fires after popstate / silent on the app's own writes")을 하고 있는데, 실제 변경 파일에는 `react/src/hooks/useKeyedSnapshot.test.tsx` 하나뿐입니다. 훅이 `useBrowserPopstateEffect`로 이름과 구현이 바뀌면서 테스트가 같이 사라진 것으로 보입니다.

하필 이 훅이 이 PR에서 가장 조립이 까다로운 부분입니다 — `popstate`로 ref를 arm 하고, `useOptimisticSearchParams`와 `window.location.search`를 normalize 비교해서 nuqs가 따라잡은 시점에만 콜백을 흘려보내는 상태 기계입니다. 최소한 (a) popstate 후 nuqs가 아직 못 따라잡은 렌더에서는 조용하고, (b) 따라잡은 렌더에서 정확히 한 번 발화하며, (c) 페이지 자신의 `setQueryParams` 쓰기에는 발화하지 않는다 — 이 세 케이스는 복구해 주세요.

`useKeyedSnapshot.test.tsx` 쪽도 한 케이스만 더 추가해 주시면 좋겠습니다: `setAfterSnapshot(next)` 직후 `sourceKey`가 **아직 이전 키인 채로** 리렌더되는 경우. 현재 테스트는 항상 `setAfterSnapshot` 다음 리렌더에서 `sourceKey`와 `value`를 새 탭 것으로 같이 바꿔주기 때문에, 훅 주석이 명시적으로 방어한다고 말하는 그 상황이 실제로는 한 번도 실행되지 않습니다.

## 2. PR 본문이 실제 코드와 어긋나 있습니다

Copilot 대응 커밋에서 설계가 한 번 더 바뀌었는데 본문이 이전 회차 기준으로 남아 있습니다. CLAUDE.md 규약상 접근 방식이 바뀌면 설명도 같이 갱신되어야 합니다.

- 본문: `useBrowserNavigationEffect(onNavigated)` → 실제 export: `useBrowserPopstateEffect(onSettled)`
- 본문: "`pushState`/`replaceState` never emit `popstate` ... **no de-duplication guard is needed**" → 실제 코드(`react/src/hooks/index.tsx:146-162`)에는 `isSettlePendingRef` arm/disarm 가드와 nuqs-settle 비교가 들어 있습니다. 지금 본문은 그 가드가 왜 있는지를 오히려 부정하는 설명입니다.
- 본문: "The reload builds its snapshot from `location.search`, because on a navigation it runs before nuqs has re-parsed the URL" → 실제로는 `loadTabFromQueryParams()`가 nuqs의 `queryParams`/`currentSnapshot`을 그대로 쓰고, 대신 훅이 nuqs가 따라잡을 때까지 콜백을 붙잡는 방식으로 뒤집혔습니다.
- 본문: 자격증명 탭의 `?action=add` opener가 "now local state" → 아래 3번 참고. 실제로는 로컬 상태로 옮겨진 게 아니라 제거됐습니다.

## 3. `?action=add` 딥링크가 대체 없이 사라집니다

`AdminUserCredentialList`에서 `action` 쿼리 파라미터를 읽어 keypair 생성 모달을 여는 `useEffect`가 삭제됐고, 이를 대신하는 로컬 상태 경로는 없습니다(모달은 이제 화면 안의 Add 버튼으로만 열립니다).

레포 전체를 훑어보면 `action=add`를 만들어내는 곳이 하나도 없어서 실질적으로 dead code일 가능성이 높고, 그렇다면 지우는 것에 찬성합니다. 다만 이건 외부에서 붙여넣을 수 있었던 딥링크의 제거라 "로컬 상태로 옮겼다"가 아니라 "소비자가 없어 제거했다"로 본문에 명시해 주세요. 혹시 문서·온보딩 메일·다른 제품에서 이 URL을 쓰고 있다면 페이지 레벨에서 한 번 읽고 소비하는 형태로 살려두는 편이 안전합니다.

## 4. `loadQuery`의 기본 `store-or-network`로 인한 신선도 하락 (설계 확인 요청)

기존 두 컴포넌트는 최초 로드에서 `store-and-network`였습니다. 지금은 `loadTabQuery` / 탭 전환 / 자식의 필터·정렬·페이지 변경이 전부 `loadQuery(variables)` 기본값, 즉 `store-or-network`로 나갑니다. `@skip` 수정으로 store가 실제로 이 쿼리를 만족시킬 수 있게 된 지금은, 이미 본 적 있는 변수 조합(예: 1페이지로 되돌아오기, 탭 왕복)은 **네트워크를 아예 타지 않습니다.** 수동 새로고침 전까지 임의로 오래된 목록을 보게 됩니다.

"매 방문마다 스켈레톤"을 없애는 것이 목적이라면 `store-and-network`가 두 마리를 다 잡습니다 — 캐시된 데이터로 즉시 그려서 스켈레톤도 없고, 뒤에서 갱신도 됩니다(`usePreloadedQuery`는 suspend 하지 않습니다). 지금처럼 `store-or-network`를 유지하는 게 의도라면 그 이유를 본문에 한 줄 남겨주세요.

## 5. 콜드 스타트에서 페이지 전체가 스켈레톤이 됩니다

`useSuspendedTOTPSupported()`가 `AdminUsersPage:93`에서 호출되면서 이제 **페이지 자체가** suspend 합니다. TanQuery 캐시가 아직 안 데워진 직접 진입/새로고침에서는 라우트 레벨 fallback(`routes.tsx`의 `<BAISkeleton rows={4} />`)이 잡히고, 카드와 탭 리스트까지 통째로 사라집니다. 이전에는 카드·탭이 즉시 그려지고 본문만 `BAISkeletonAstryx`였습니다.

`isNotSupportTotp`를 초기 로드 전에 확정해야 한다는 문제 인식은 맞습니다. 다만 헤더가 유지되도록, 이 suspend 지점을 탭 카드 **안쪽**의 경계 아래로 내리거나(변수 조립을 하는 얇은 내부 컴포넌트로 분리) 앱 셸에서 이 TanQuery 키를 미리 prefetch 하는 쪽을 고려해 주세요. 본문에서도 "cold-start deep link는 사람 확인 필요"로 남겨두신 항목입니다.

## 6. 사소한 것들

- **`AdminUserManagement.tsx:476`** — `BAIFetchKeyButton`에 `value=""`를 넘기고 있습니다. 이 prop은 optional이고, 해당 docblock이 정확히 이 경우("Consumers that drive refetching directly (e.g. a preloaded `loadQuery(..., 'network-only')` in `onChange`)")를 두고 **생략하라**고 적어두었습니다. 그냥 빼주세요.
- **`hooks/index.tsx:79-85`** — `useKeyedSnapshot` 주석의 "right after `setAfterSnapshot` the caller's async URL mirror still reports the departing key"는 nuqs 2.9 기준으로 사실이 아닙니다. `useQueryStates`의 setter는 emitter를 통해 내부 state를 **같은 배치에서 동기적으로** 갱신하고, throttle 되는 것은 URL 쓰기뿐입니다. 훅의 change-triggered 동기화는 어느 쪽이든 옳게 동작하지만, 이 주석은 나중에 읽는 사람을 반대 방향으로 보냅니다.
- **`hooks/index.tsx:148-162`** — `isSettlePendingRef`는 모든 `popstate`에서 arm 되지만 disarm은 "nuqs params가 새 객체로 다시 렌더되고 + 비교가 일치할 때"만 일어납니다. 쿼리스트링이 바뀌지 않는 popstate(해시 이동 등)면 플래그가 arm 된 채로 남고, 이후 페이지 자신의 무관한 `setQueryParams` 쓰기에서 뒤늦게 `onSettled`가 한 번 발화합니다. 지금 이 페이지에서 도달 가능성은 낮지만, 훅으로 공용화하는 이상 arm 할 때 `window.location.search`를 같이 기록해두고 비교하는 정도의 저렴한 가드는 넣어둘 만합니다.

---

파일럿의 뼈대에는 이견 없습니다. 1·2·3만 정리되면 나머지(4·5)는 답변만 듣고 넘어가도 될 것 같습니다.

react/src/hooks/index.tsx
```diff
@@ -68,6 +70,101 @@ export const useTabQuerySnapshot = <T extends string>(
...
+/**
+ * Runs `onSettled` after a browser back/forward, once nuqs' query state
+ * matches the address bar — so callers can read their nuqs state inside the
+ * callback without comparing it against `location.search` themselves.
+ *
+ * The wait is the point: nuqs applies `popstate` inside a `startTransition`,
+ * so for at least one render after the event every `useQueryStates` value
+ * still describes the page the user just left.
+ */
+export const useBrowserPopstateEffect = (onSettled: () => void) => {
```
agatha197, line 145: **1. Missing tests (blocking).** This hook ships with zero coverage: the `useBrowserNavigatedQueryEffect.test.tsx` promised in the reply to Copilot's coverage comment (and in the PR body) is not in the diff — it was dropped when the hook was renamed.
Please restore the three cases it claimed: silent while nuqs has not caught up, fires exactly once when it has, and silent on the page's own `setQueryParams` writes. (`useKeyedSnapshot.test.tsx` is also missing the one case its docblock defends: a rerender where `sourceKey` is still the departing key after `setAfterSnapshot`.)

react/src/pages/AdminUsersPage.tsx
```diff
@@ -2,27 +2,254 @@
...
+
+  // First visit: mount, direct URL entry, reload.
+  const loadTabOnMount = useEffectEvent(loadTabFromQueryParams);
+  useEffect(() => {
+    loadTabOnMount();
+  }, []);
+
+  // Back/forward: the hook holds the callback until `queryParams` and
+  // `tablePaginationOption` describe the URL the user navigated to.
+  useBrowserPopstateEffect(loadTabFromQueryParams);
```
agatha197, line 262: **2. The PR description no longer matches the code.** It still describes the previous iteration — `useBrowserNavigationEffect`, "no de-duplication guard is needed", and "the reload builds its snapshot from `location.search`".
The shipped hook is `useBrowserPopstateEffect`, it *does* carry an arm/disarm guard, and it builds the snapshot from nuqs state while holding the callback until nuqs settles. CLAUDE.md asks the description to track a changed approach — please refresh it.

react/src/components/AdminUserCredentialList.tsx
```diff
@@ -248,31 +238,22 @@ const AdminUserCredentialList: React.FC = () => {
...
         }
         setSelectedKeypairs([]);
-        updateFetchKey();
+        onReload(variables, { fetchPolicy: 'network-only' });
       },
     });
   };
 
-  useEffect(() => {
-    if (action === 'add') {
```
agatha197, line 257: **3. `?action=add` is removed, not relocated.** The PR body says this opener is "now local state", but the effect is deleted with no replacement, so the deep link simply stops working.
Nothing in the repo produces that URL, so dropping it looks right — please just describe it as an unused deep link that was removed, rather than as moved to local state.

react/src/pages/AdminUsersPage.tsx
```diff
@@ -2,27 +2,254 @@
...
+    setQueryParams(
+      { tab: 'credentials', ...snapshot.queryParams },
+      { history: 'replace' },
+    );
+    setTablePaginationOption(snapshot.tablePaginationOption);
+  };
+
+  const loadTabQuery = (tab: TabKey, snapshot: TabSnapshot) => {
+    if (tab === 'users') {
+      loadUsersQuery(usersVariablesOf(snapshot));
```
agatha197: **4. `loadQuery` defaults to `store-or-network` here** (downgraded from my earlier review — not blocking, and only this tab-entry path): a variable combo seen before never revalidates. The `@skip` fix makes the cached record structurally complete, which is correct; the concern is row freshness, not field shape.
Concrete case: `adminCreateUserV2` also returns a keypair (`UserSettingModal.tsx:594`), so creating a user and then switching to Credentials can render a `keypair_list` that is missing it until a manual refresh — the old code re-validated because the tab remounted with `store-and-network`. Passing `store-and-network` here costs no skeleton: with a complete store `usePreloadedQuery` reads synchronously and does not suspend.

react/src/pages/AdminUsersPage.tsx
```diff
@@ -2,27 +2,254 @@
...
+        ? USER_LIST_DEFAULT_PAGE_SIZE
+        : CREDENTIAL_LIST_DEFAULT_PAGE_SIZE,
+  },
+});
 
 const AdminUsersPage: React.FC = () => {
   'use memo';
   const { t } = useTranslation();
-  const { currentTab, onTabChange } = useTabQuerySnapshot(tabParser);
+  const isTOTPSupported = useSuspendedTOTPSupported();
```
agatha197, line 93: **5. Page-level suspend hides the tabs on a cold start.** This makes the whole page suspend, so before the TanQuery cache is warm the route fallback (`<BAISkeleton rows={4} />` in `routes.tsx`) replaces the card *and* the tab list; previously only the card body skeletoned.
Consider moving the suspending read below the tab card (a thin inner component that assembles the variables), or prefetching the `isManagerSupportingTOTP` key in the app shell.

react/src/components/AdminUserManagement.tsx
```diff
@@ -478,9 +472,11 @@ const AdminUserManagement: React.FC<AdminUserManagementProps> = () => {
             </BAIFlex>
           )}
           <BAIFetchKeyButton
-            loading={deferredFetchKey !== fetchKey}
-            value={fetchKey}
-            onChange={updateFetchKey}
+            loading={isPending}
+            value=""
```
agatha197: **6a. Nit — drop this prop.** `BAIFetchKeyButton.value` is optional, and its docblock names exactly this case ("consumers that drive refetching directly (e.g. a preloaded `loadQuery(..., 'network-only')` in `onChange`) can omit it entirely").

react/src/hooks/index.tsx
```diff
@@ -68,6 +70,101 @@ export const useTabQuerySnapshot = <T extends string>(
...
+ * Per-key value snapshots with an uncontrolled current key. The caller
+ * defines the snapshot shape `V` (e.g. `{ queryParams,
+ * tablePaginationOption }`) and passes its live value every render; the
+ * first-render value — typically parsed from the query string — seeds the
+ * initial key's snapshot.
+ *
+ * `sourceKey` seeds the key state and re-syncs it whenever the argument's
+ * value changes (e.g. the caller derives it from the URL and the user
+ * navigates back/forward). The sync is change-triggered, not
+ * difference-triggered: right after `setAfterSnapshot` the caller's async URL
```
agatha197: **6b. Nit — this claim is not true for nuqs 2.9.** `useQueryStates`' setter updates its internal state synchronously in the same batch (only the URL write is throttled), so `queryParams` already describes the new key on the very next render.
The hook's change-triggered sync is correct either way, but the comment sends a future reader in the opposite direction.

react/src/hooks/index.tsx
```diff
@@ -68,6 +70,101 @@ export const useTabQuerySnapshot = <T extends string>(
...
+export const useBrowserPopstateEffect = (onSettled: () => void) => {
+  'use memo';
+  const handleSettled = useEffectEvent(onSettled);
+  // nuqs' view of the URL; a fresh object per popstate re-runs the effect below.
+  const nuqsSearchParams = useOptimisticSearchParams();
+  // Distinguishes back/forward from the page's own `setQueryParams` writes.
+  const isSettlePendingRef = useRef(false);
+
+  useEventListener('popstate', () => {
+    isSettlePendingRef.current = true;
```
agatha197, line 154: **6c. Nit — the flag is armed unconditionally but only disarmed on a match.** A `popstate` that does not change the query string leaves it armed (the effect never re-runs, since nuqs emits no new object), and the next unrelated `setQueryParams` write then fires `onSettled` late.
Cheap guard: record `window.location.search` when arming and bail out if the navigation did not actually change it.

agatha197 (changes requested):
Line-anchored English breakdown of my previous review, as requested. Same findings 1–6, with **#4 downgraded** after discussion: the `@skip` fix does make the cached record complete, so the remaining concern is row freshness on the tab-entry path only, not a blocker.

Blocking: **1** (missing tests), **2** (stale PR description), **3** (`?action=add` described as relocated when it is removed). **4** and **5** are design confirmations; **6a–6c** are nits.

react/src/hooks/index.tsx
nowgnuesLee: Fixed in 209ff5892 (tightened in 7f66764b5). `react/src/hooks/useBrowserPopstateEffect.test.tsx` now carries the three cases you asked for: silent while nuqs has not caught up with the address bar, fires exactly once once its params match (and does not replay), and silent for the page's own `setQueryParams` writes. It mocks the adapter's `useOptimisticSearchParams` so nuqs' view of the URL can lag `window.location` on purpose.

I did not restore the deleted predecessor verbatim — the hook is a different state machine now (it withholds the callback until nuqs settles, rather than firing once per popstate), so the interesting edges moved.

`useKeyedSnapshot.test.tsx` gains the missing case: a rerender whose `sourceKey` still reports the departing key after `setAfterSnapshot` must not override the key. Every existing case moved `sourceKey` and `value` together, so the change-triggered sync was never exercised.

`vitest run src/hooks` → 26 files / 349 tests passing.

react/src/pages/AdminUsersPage.tsx
nowgnuesLee: Description refreshed to match the shipped code:

- `useBrowserNavigationEffect(onNavigated)` → `useBrowserPopstateEffect(onSettled)`.
- "no de-duplication guard is needed" is replaced by what the guard actually does — arm on `popstate`, hold the callback until `useOptimisticSearchParams()` and `window.location.search` normalize equal, then disarm. (`pushState`/`replaceState` emitting no `popstate` is still there, but as the reason the page's own writes never arm it, not as a reason no guard exists.)
- "builds its snapshot from `location.search`" → builds it from nuqs state, which is exactly what the wait buys.
- The Verification section now lists the real test files and case names.

While there: the changelog for the `?action=add` line moved too — see the thread on `AdminUserCredentialList.tsx`.

react/src/components/AdminUserCredentialList.tsx
nowgnuesLee: Reworded, and thanks for the sweep — I re-checked and confirmed the same: nothing in the repo (app, e2e, docs) produces `?action=add`.

The description now says the opener was **removed as an unused deep link**, not moved to local state, and adds the caveat that if an external doc or onboarding mail turns out to link it, it should come back as a page-level read rather than child-owned query state — which is the shape this PR is arguing for anyway.

react/src/pages/AdminUsersPage.tsx
nowgnuesLee: Not changing this — it is the intended shape. The TOTP capability is a page-level query, so the page waiting on its Suspense boundary is the correct behaviour, not a regression. Splitting it into an inner component just to keep the tab list painted would be hiding a page-level dependency behind the header.

Leaving the route-level fallback as the single loading state for the page.

agatha197 (changes requested):
please resolve conflicts

react/src/components/AdminUserManagement.tsx
nowgnuesLee: Fixed in 0a72509ec — dropped the prop.

react/src/hooks/index.tsx
nowgnuesLee: You're right, and I verified it in the installed 2.9.2: `useQueryStates`' setter calls `emitter.emit(urlKey, …)` synchronously and the subscribed handler runs `setInternalState` in the same call (`nuqs/dist/index.js:539-556`); only the URL write goes through the throttle/flush queue.

Fixed in 0a72509ec. The guarantee stands without that premise, so the docblock now just states it — a rerender whose `sourceKey` still reports the departing key cannot undo `setAfterSnapshot` — and I swept the same "async URL mirror" wording out of the test comment that repeated it.

react/src/hooks/index.tsx
nowgnuesLee: Declining this one — the premise does not hold for how the hook calls nuqs.

`useOptimisticSearchParams()` is called with no `watchKeys`, and nuqs skips the equality check entirely in that case (`nuqs/dist/key-isolation-CDukC9n4.js:4-13`):

```js
function applyChange(newValue, keys, copy) {
  return (oldValue) => {
    if (!(keys.length === 0 ? true : keys.some(...))) return oldValue;  // keys=[] -> !true -> skipped
    return filterSearchParams(newValue, keys, copy);                    // keys=[] -> newValue as-is
  };
}
```

With `keys.length === 0` the ternary yields `true`, so the early return is never taken and `setSearchParams` always receives a **fresh** `URLSearchParams` built from `location.search`. A new identity re-renders, the effect re-runs, and the flag is consumed on that commit — including for a `popstate` that did not change the query string. So it cannot stay armed and leak into a later `setQueryParams` write.

What is left over is one `onSettled` on a query-string-preserving `popstate`, i.e. a redundant `loadTabQuery` with the same variables — a revalidation, now that these loads are `store-and-network`. And it is unreachable on this page: the only hash-preserving navigations in the app are on the session pages, the `href="#"` anchors all `preventDefault`, and the only `push` here is the tab switch, which always changes the query string.

Happy to add the guard if the hook picks up a caller on a hash-navigating page — at that point it would be filtering navigations, which the ref deliberately does not do today (it exists to attribute a nuqs update to a navigation, not to decide which navigations count).

react/src/pages/AdminUsersPage.tsx
nowgnuesLee: Fixed in bb04d363e — the page's loads now pass `fetchPolicy: 'store-and-network'`.

Applied to the child-driven variable changes too, not just the tab-entry path: before this PR both children were `store-and-network` on their first load, so a filter/sort/page change back to a seen combination used to revalidate as well. The children's explicit `network-only` (manual refresh, post-mutation) still wins via spread.

yomybaby (approved):
LGTM

PR #8590, opened 2026-08-04
fix: restore backend.ai-ui types lost to a stale path and condition order

### LINKED ISSUE #8589: fix: backend.ai-ui ships empty type declarations, breaking downstream consumers
## Summary

Every `backend.ai-ui` release from **26.4.8** onward publishes `dist/backend.ai-ui.d.ts` containing only `export {}`, and every `dist/locale/*.d.ts` likewise. The real declarations are shipped (under `dist/src/`) but nothing points at them, so any TypeScript consumer importing from `backend.ai-ui` gets an empty module.

## Impact

Downstream, `import { BAIConfigProvider, BAITable } from 'backend.ai-ui'` fails with `TS2305: Module '"backend.ai-ui"' has no exported member`, and the five `dist/locale/*` default imports fail with `TS1192: has no default export`. Because `BAITable` then resolves to nothing, every wrapper around it goes untyped and every column callback becomes an implicit `any`.

In `backend.ai-fasttrack` (which pins `26.4.0-canary-1a56f8d31-20260327`, predating the regression), bumping to `26.8.0` produces **57 type errors against a baseline of 0** — 8 from the missing exports and 49 cascading implicit-`any` errors. All 57 come from these six files alone.

## Bisect

| Version | `dist/backend.ai-ui.d.ts` | `dist/locale/en_US.d.ts` |
|---|---|---|
| 26.4.7 | `export * from './packages/backend.ai-ui/src/index'` | `import _default from '../packages/backend.ai-ui/src/locale/en_US'; export default _default` |
| **26.4.8** | `export {}` | `export {}` |
| 26.4.9 – 26.4.10 | `export {}` | `export {}` |
| 26.7.0 – 26.7.3 | `export {}` | `export {}` |
| 26.8.0, 26.8.0-rc.0, all 26.8 canaries (through `67c18bcd9-20260804`) | `export {}` | `export {}` |

So the regression landed between `26.4.7` and `26.4.8` and has shipped in every release since, including the current stable `26.8.0`.

## Two contributing defects

**1. The declared `types` entry has never existed.** `package.json` declares:

```json
"types": "./dist/index.d.ts",
"exports": { ".": { "import": "./dist/backend.ai-ui.js", "types": "./dist/index.d.ts" } }
```

`dist/index.d.ts` is not present in the tarball of *any* published version, including 26.4.7. Consumers only ever resolved types via TypeScript's fallback to the `.js` sibling `dist/backend.ai-ui.d.ts` — which is why 26.4.7 worked and 26.4.8 did not. The declared path should either be emitted or corrected to the file that is actually shipped.

**2. `insertTypesEntry` now emits an empty entry.** The emitted declaration tree moved from `dist/packages/backend.ai-ui/src/` to `dist/src/` (visible in the tarball layout), but `vite-plugin-dts`'s `insertTypesEntry: true` in `packages/backend.ai-ui/vite.config.ts` writes an entry that re-exports nothing instead of `export * from './src/index'`. The barrel it should point at, `dist/src/index.d.ts`, is present and correct.

The same emit path strips the locale declarations: `dist/src/locale/en_US.d.ts` is correct, while the `dist/locale/en_US.d.ts` that the `./dist/locale/*` export condition resolves to is `export {}`.

Possibly related: 72 files under `dist/src/` import React types via `'../../../packages/backend.ai-ui/node_modules/@types/react'`, a relative path that escapes the published package. TypeScript happens to tolerate it under pnpm's layout, but it is not a valid published path and looks like the same `entryRoot`/`preserveSymlinks` resolution issue.

## Reproducing

```bash
npm pack backend.ai-ui@26.8.0
tar -xzOf backend.ai-ui-26.8.0.tgz package/dist/backend.ai-ui.d.ts   # => export {}
tar -tzf  backend.ai-ui-26.8.0.tgz | grep 'dist/index.d.ts'          # => no match
tar -xzOf backend.ai-ui-26.8.0.tgz package/dist/src/index.d.ts       # => the real barrel
```

## Verification that this is the whole cause

Restoring the six re-exports by hand in an installed `26.8.0` — the entry barrel plus the five locale declarations a consumer imports — takes the downstream error count from **57 to 0**, with no source changes on the consumer side:

```
# dist/backend.ai-ui.d.ts
export * from './src/index'

# dist/locale/<locale>.d.ts
import _default from '../src/locale/<locale>'
export default _default
```

`backend.ai-fasttrack` is landing this as a temporary `pnpm patch` so it can adopt the `BAIFetchKeyButton` auto-refresh interval API (which requires ≥26.8.0), and will drop the patch as soon as a release ships correct declarations.

rapsealk:
Resolves #8589

> **On the reordering:** the swap of `types` and `import` looks like churn but is the half that makes the fix work. Repointing `types` without it leaves the downstream error count unchanged at 57 — measured, see the table below. Export conditions match in declaration order, so with `import` first the `types` condition is never reached.

## Problem

Every `backend.ai-ui` release since **26.4.8** — including the current stable `26.8.0` — resolves to an empty module for external TypeScript consumers. `BAIConfigProvider`, `BAITable` and friends come back as `TS2305: has no exported member`, the locale default imports fail with `TS1192`, and everything typed through them degrades to implicit `any`.

Measured downstream in `backend.ai-fasttrack`: bumping from its pinned `26.4.0-canary` (which predates the regression) to `26.8.0` produces **57 type errors against a baseline of 0** — 8 direct, 49 cascading implicit-`any`.

## Root cause

Two things in the manifest, neither of which pointed at a real file:

1. **`types` names a file the build never produces.** It has always been `./dist/index.d.ts`; the declarations land under `./dist/src/`. This was wrong in every version ever published, including working ones.

2. **`import` was ordered before `types` in each export condition.** Conditions match in declaration order, so TypeScript matched `import` and then looked for the `.d.ts` beside it — `dist/backend.ai-ui.d.ts`. The `types` condition was never consulted, which is exactly why defect 1 went unnoticed for so long.

So resolution always went through the adjacent-file fallback. That happened to work until 26.4.8, when the build layout moved and `dist/backend.ai-ui.d.ts` became an `export {}` stub. For comparison, 26.4.7 shipped it as `export * from './packages/backend.ai-ui/src/index'`.

## Fix

Point `types` at the declarations the build actually produces, and order it ahead of `import` so the condition is reached at all. Manifest-only — no build changes.

Neither half works alone: with the stale path the `types` condition names a nonexistent file, and with `import` first the condition is skipped regardless of what it names.

## Why CI and local dev never caught this

`react/tsconfig.json` aliases the package straight to source:

```jsonc
"backend.ai-ui": ["../packages/backend.ai-ui/src/index.ts"],
"backend.ai-ui/dist/locale/*": ["../packages/backend.ai-ui/src/locale/*"],
```

The in-repo app never consumes the published declarations, so no amount of in-repo type checking can see this class of breakage. Only external consumers are affected — which is why it survived three months and eight releases.

## Verification

Built the package, packed it with `pnpm pack`, and installed the tarball into `backend.ai-fasttrack` at the version that currently fails:

| Package under test | FastTrack `tsc --noEmit` |
| --- | --- |
| published `26.8.0` | **57 errors** |
| published `26.8.0` + a 6-file local `pnpm patch` filling in the stubs | 0 errors |
| `types` repointed to `dist/src/`, **conditions left in original order** | **57 errors** |
| **this branch (repointed + reordered)** | **0 errors** |

Row 3 is the isolation run for the reordering question above.

In-repo: `pnpm lint` clean, `pnpm test` 367 passed / 1 skipped, `tsc --noEmit` in `react/` 0 errors (unchanged — it aliases to source).

## Not included

Two adjacent problems left out to keep this to the mandatory fix:

- `insertTypesEntry: true` still writes the `export {}` stubs at `dist/backend.ai-ui.d.ts` and `dist/locale/*.d.ts`. With the conditions fixed they are unreachable dead files rather than a trap, so removing them is cleanup, not a fix.
- 72 published declaration files import React via `../../../packages/backend.ai-ui/node_modules/@types/react`, a path that escapes the package. The `paths` mapping added in FR-2925 for antd's JSX checking reaches declaration emit, and `vite-plugin-dts` rewrites each mapped import to its resolved location. TypeScript tolerates it under pnpm's layout today, so it is latent fragility rather than a live break — worth its own issue. (Note for whoever picks it up: `entryRoot: 'src'` is not the answer; it makes vite-plugin-dts drop the declaration output entirely.)

`backend.ai-fasttrack` is carrying the `pnpm patch` as a temporary workaround and will drop it once a release ships with this fix.

- [x] Test case(s) to demonstrate the difference of before/after — see the table above
- [ ] Documentation

### DIFF
--- a/packages/backend.ai-ui/package.json
+++ b/packages/backend.ai-ui/package.json
@@ -18,16 +18,16 @@
   ],
   "main": "./dist/backend.ai-ui.js",
   "module": "./dist/backend.ai-ui.js",
-  "types": "./dist/index.d.ts",
+  "types": "./dist/src/index.d.ts",
   "sideEffects": false,
   "exports": {
     ".": {
-      "import": "./dist/backend.ai-ui.js",
-      "types": "./dist/index.d.ts"
+      "types": "./dist/src/index.d.ts",
+      "import": "./dist/backend.ai-ui.js"
     },
     "./dist/locale/*": {
-      "import": "./dist/locale/*.js",
-      "types": "./dist/locale/*.d.ts"
+      "types": "./dist/src/locale/*.d.ts",
+      "import": "./dist/locale/*.js"
     }
   },
   "files": [
--- a/packages/backend.ai-ui/package.json
+++ b/packages/backend.ai-ui/package.json
@@ -18,16 +18,16 @@
   ],
   "main": "./dist/backend.ai-ui.js",
   "module": "./dist/backend.ai-ui.js",
-  "types": "./dist/src/index.d.ts",
+  "types": "./dist/index.d.ts",
   "sideEffects": false,
   "exports": {
     ".": {
-      "types": "./dist/src/index.d.ts",
-      "import": "./dist/backend.ai-ui.js"
+      "import": "./dist/backend.ai-ui.js",
+      "types": "./dist/index.d.ts"
     },
     "./dist/locale/*": {
-      "types": "./dist/src/locale/*.d.ts",
-      "import": "./dist/locale/*.js"
+      "import": "./dist/locale/*.js",
+      "types": "./dist/locale/*.d.ts"
     }
   },
   "files": [
--- a/packages/backend.ai-ui/src/components/BAIDynamicStepInputNumber.tsx
+++ b/packages/backend.ai-ui/src/components/BAIDynamicStepInputNumber.tsx
@@ -1,5 +1,5 @@
-import useControllableState_deprecated from '../../../../react/src/hooks/useControllableState';
 import { useUpdatableState } from '../hooks';
+import { useControllableValue } from 'ahooks';
 import { InputNumber, InputNumberProps } from 'antd';
 import * as _ from 'lodash-es';
 import React, { useEffect } from 'react';
@@ -22,12 +22,9 @@ const BAIDynamicStepInputNumber: React.FC<BAIDynamicStepInputNumberProps> = ({
   // onChange,
   ...inputNumberProps
 }) => {
-  const [value, setValue] = useControllableState_deprecated<number>(
-    inputNumberProps,
-    {
-      defaultValue: dynamicSteps[0],
-    },
-  );
+  const [value, setValue] = useControllableValue<number>(inputNumberProps, {
+    defaultValue: dynamicSteps[0],
+  });
 
   // FIXME: this is a workaround to fix the issue that the value is not updated when the value is controlled
   const [key, updateKey] = useUpdatableState('first');
--- a/packages/backend.ai-ui/vite.config.ts
+++ b/packages/backend.ai-ui/vite.config.ts
@@ -75,6 +75,8 @@ export default defineConfig(({ mode }) => {
         insertTypesEntry: true,
         compilerOptions: {
           preserveSymlinks: false,
+          rootDir: 'src',
+          paths: {}
         },
       }),
       svgr(),


nowgnuesLee (approved):
LGTM

PR #8592, opened 2026-08-04
fix(FR-3469): warn that Open to Public cannot be changed at deployment creation

### LINKED ISSUE #8591: Warn that Open to Public cannot be changed when creating a deployment
In the deployment setting modal (`DeploymentSettingModal.tsx`), the **Resource Group** field shows a warning in both create and edit mode telling the user the value cannot be changed after creation.

**Open to Public** carries the same backend constraint (the manager rejects `openToPublic` changes after creation), but the warning was only surfaced as a tooltip on the disabled checkbox in **edit** mode. When creating a deployment there was no hint at all that the setting is permanent, so users can pick the wrong visibility and only discover it is locked afterwards.

### Expected

Open to Public shows the same inline warning as Resource Group, in both create and edit mode.

### Notes

- The string `deployment.OpenToPublicCannotBeChanged` already exists in every locale, so no new translation is needed.
- Follow-up to FR-3003, which introduced the disabled + tooltip behavior for edit mode.

JIRA Issue: FR-3469

agatha197:
Resolves #8591 (FR-3469)

## Problem

In the deployment setting modal, the **Resource Group** field tells the user in
both create and edit mode that its value is fixed at creation.

**Open to Public** carries the same backend constraint, but the explanation only
ever reached **edit** mode: it rode on `CheckboxInput`'s `disabledMessage`, which
surfaces as a tooltip on hover of an already-disabled control. While *creating* a
deployment nothing is disabled, so nothing hinted that the choice is permanent.

## Changes

- Move the note into the field's `extra` slot — the same place Resource Group's
  identical note sits — so it renders in **both** create and edit mode. The
  constraint is field-level metadata rather than an explanation of a disabled
  control, which is why `disabledMessage` was the wrong carrier for it.
- With no tooltip left to host, the nested `noStyle` `BAIFormItem` collapses into
  a single `BAIFormItem name="openToPublic"`, and the `PublicCheckbox` bridge
  loses its `disabledMessage` pass-through.
- `disabled={!!deployment}` is unchanged — edit-mode behavior is the same.
- Update the `TODO(needs-backend)` marker to name the `disabled` prop + `extra`
  note as the pair to remove once the manager supports updating this setting.

### i18n

No new keys. `deployment.OpenToPublicCannotBeChanged` is reworded to be
mode-neutral, mirroring `ResourceGroupCannotBeChanged` ("The resource group
selected at creation cannot be changed."), and the sentence it duplicates is
dropped from `deployment.OpenToPublicTooltip`. Also repairs a missing comma that
left `fi.json` unparseable.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript
/ Astryx theme build / Terminology).

Checked live against a running dev server in all four states. Read binding
(edit mode shows the stored value as `[checked] [disabled]`) and write binding
(create-mode checkbox toggles) both confirmed after flattening the form item.

| | Create | Edit |
|---|---|---|
| **Light** |  |  |
| **Dark** |  |  |

Follow-up to FR-3003, which introduced the disabled + tooltip behavior for edit mode.

## Note on the rebase

This branch was written against the pre-Astryx modal and had gone into conflict
with `main`. It is rebased onto current `main` and reimplemented against the
Astryx version of the file (`BAIFormItem` + Astryx `CheckboxInput` rather than
antd `Form.Item` + `Typography.Text type="warning"` + `Tooltip`), squashed into a
single commit. The `Typography.Text type="warning"` tint is intentionally not
reintroduced: `main` already dropped it for Resource Group's identical note
(Astryx `Text` has no warning color, and a warning-tinted box would over-signal a
static informational note), and this field now matches that treatment.


ironAiken2 (approved):
LGTM

PR #8596, opened 2026-08-05
fix: externalize peer subpaths so backend.ai-ui stops bundling react-dom

### LINKED ISSUE #8595: backend.ai-ui bundles the react-dom/client renderer, crashing consumers on any other React patch
## Summary

`backend.ai-ui`'s published bundle inlines a full copy of the **`react-dom/client` renderer** instead of externalizing it. `react-dom` is declared in `peerDependencies`, and the bare `react-dom` specifier *is* external — but the `react-dom/client` subpath is not, so the renderer is compiled into `dist/backend.ai-ui.js`.

Any consumer whose `react` is not byte-identical to the `react-dom` version used at publish time crashes at import time.

## Evidence

`dist/backend.ai-ui.js.map` lists the inlined sources:

```
.../react-dom/19.2.7/.../node_modules/react-dom/cjs/react-dom-client.production.js
.../react-dom/19.2.7/.../node_modules/react-dom/cjs/react-dom-client.development.js
.../react-dom/19.2.7/.../node_modules/react-dom/client.js
```

Only the bare specifier stayed external:

```js
import $0, { unstable_batchedUpdates as V6, createPortal as lF } from "react-dom";
```

And the inlined renderer carries react-dom's own version guard, with `19.2.7` baked in:

```js
var l3 = e.version;               // e = the *external* react
if (l3 !== "19.2.7") throw Error(/* Incompatible React versions */);
```

## Symptom

A consumer on `react@19.2.6` (one patch behind) gets a hard white screen on every page that imports the package, with a single console error:

```
Error: Incompatible React versions: The "react" and "react-dom" packages must have the exact same version. Instead got:
  - react:      19.2.6
  - react-dom:  19.2.7
    at .../deps/backend__ai-ui.js
```

This is not a warning — the throw happens while the module initializes, so `BAIConfigProvider` never mounts and the app renders nothing.

## Reproduction

1. App with `react@19.2.6` / `react-dom@19.2.6` (both satisfying `^19.2.6`).
2. `pnpm add backend.ai-ui@26.8.0-canary-fe8c2d5c4-20260805`.
3. Import anything from the package (e.g. `BAIConfigProvider`) and load the page.

Observed in `backend.ai-fasttrack` after bumping to that canary (lablup/backend.ai-fasttrack#5226): the whole SPA white-screens in dev and in a production build.

## Why pinning is not a real fix

Setting the consumer's `react` to exactly `19.2.7` clears the guard, but the page then runs **two `react-dom` renderer instances** — the bundled one and the consumer's own. That is its own bug class (portals, context and event delegation crossing renderer boundaries), and it silently re-breaks on the next React patch release, since the guard compares exact strings.

## Suggested fix

Externalize the subpaths, not just the bare specifier, in the library build:

```ts
// vite.config.ts — rollupOptions
external: [/^react(\/.*)?$/, /^react-dom(\/.*)?$/, /* …other peers… */]
```

A `publint` / `are-the-types-wrong` check in the package's CI would catch this class of packaging regression, the same way it would have caught the `types` condition problem fixed in #8590.

## Impact

Every external consumer of `backend.ai-ui` is affected unless their `react` happens to be exactly the version used at publish time. Please confirm whether the released `26.8.0` (not just this canary) is built the same way.

rapsealk:
Resolves #8595

## Problem

`packages/backend.ai-ui/vite.config.ts` listed externals as plain strings:

```ts
external: ['react', 'react-dom', 'react-router-dom', 'relay-runtime', 'antd', 'antd-style', 'graphql'],
```

Rollup matches those **exactly**, so `react-dom` was external but `react-dom/client` was not — and the whole client renderer got compiled into `dist/backend.ai-ui.js`. `dist/backend.ai-ui.js.map` named the sources:

```
.../react-dom/19.2.7/.../node_modules/react-dom/cjs/react-dom-client.production.js
.../react-dom/19.2.7/.../node_modules/react-dom/cjs/react-dom-client.development.js
.../react-dom/19.2.7/.../node_modules/react-dom/client.js
```

The inlined renderer carries react-dom's own version guard, with the build-time version baked in as a string literal:

```js
var l3 = e.version;               // e = the *external* react
if (l3 !== "19.2.7") throw Error(/* Incompatible React versions */);
```

So a consumer whose `react` is not byte-identical to ours throws while the module initializes. `BAIConfigProvider` never mounts and every page renders blank — dev server and production build alike, one console error and nothing else:

```
Error: Incompatible React versions: The "react" and "react-dom" packages must have the exact same version. Instead got:
  - react:      19.2.6
  - react-dom:  19.2.7
```

That is one patch of drift. `backend.ai-fasttrack` hit it on `react@19.2.6` after adopting a 26.8 canary (lablup/backend.ai-fasttrack#5226) and is currently carrying a `pnpm` override pinning React to exactly 19.2.7 to stay usable (lablup/backend.ai-fasttrack#5232).

Four declared peers were also missing from the list entirely and were bundled whole: **`react-relay` (243 KB), `@tanstack/react-query` (81 KB), `ahooks`, `@ant-design/icons` (77 KB)**. Duplicating those is the same hazard in slower motion — a second copy of a library whose state lives behind a React context silently stops seeing the host's provider.

## Change

`external` stays an array — the entries just become patterns instead of exact strings, derived from `peerDependencies`:

```ts
const peerDependencyPatterns = Object.keys(peerDependencies).map(
  // `.` is the only regex metacharacter an npm package name can hold.
  (name) => new RegExp(`^${name.replaceAll('.', '\\.')}(/.*)?$`),
);
```

(The escape is not decorative: `backend.ai-client` is a plausible future peer in this repo, and unescaped its `.` would also match `backend-ai-client`.)

Declaring a peer is now what makes it external, so the list cannot drift out of sync with the contract again. `i18next` / `react-i18next` are `dependencies` rather than peers, so they stay bundled and BUI's i18n instance stays isolated from the host's — the behavior the old comment described, now enforced by where the package is declared rather than by remembering to leave it off a list.

### Nothing from the old list is dropped

The deleted literal list is a strict subset of `peerDependencies`, so every entry it had — `antd` included — is still external:

| old list entry | `react` | `react-dom` | `react-router-dom` | `relay-runtime` | `antd` | `antd-style` | `graphql` |
|---|---|---|---|---|---|---|---|
| still external | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |

Confirmed on the emitted bundles rather than on the config alone — the external imports actually present in `dist/backend.ai-ui.js`:

- **published today:** `antd`, `antd-style`, `react`, `react-dom`, `react-router-dom`, `relay-runtime`
- **this branch:** the same six, plus `@ant-design/icons`, `@tanstack/react-query`, `ahooks`, `react-relay`, `react/jsx-runtime`, `react/compiler-runtime`, `antd/es/form/context`

Dropped from externals: none. (`graphql` appears in neither, before or after — it is a declared peer but nothing in the package imports it at runtime.)

## Effect on the bundle

| | before | after |
|---|---|---|
| `dist/backend.ai-ui.js` | 2.97 MB | **1.67 MB** |
| inlined `node_modules` code | 3.02 MB across 45 packages | **0.71 MB across 24** |
| `react-dom` renderer inlined | yes (1.5 MB) | **no** |
| version guard in bundle | yes | **no** |

Newly external: `@ant-design/icons`, `@tanstack/react-query`, `ahooks`, `react-relay`, `react/jsx-runtime`, `react/compiler-runtime`, `antd/es/form/context`. All are declared peers (or subpaths of one), and both consumers I checked — this repo's `react/` app and `backend.ai-fasttrack` — already have every one of them installed.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup paths, Terminology).
- `pnpm test` in `packages/backend.ai-ui` → 369 passed, 1 skipped.
- `pnpm run build:react-only` → builds clean, so the host app resolves every newly external import.
- **End-to-end against the failing case:** built this branch's `dist` into `backend.ai-fasttrack` running `react@19.2.6` — the exact combination that white-screens today. The app loads and renders normally (tables, `BAIFetchKeyButton` with its countdown border and localized labels), with **zero** console errors and the React DevTools banner logging once instead of twice — i.e. one renderer instance in the page rather than two.

## Suggested follow-up

A `publint` (and `are-the-types-wrong`) check in this package's CI would catch this whole class of packaging regression automatically — this is the second one in a month, after the `types` condition problem fixed in #8590. Happy to open that separately if you want it.

### DIFF
--- a/packages/backend.ai-ui/vite.config.ts
+++ b/packages/backend.ai-ui/vite.config.ts
@@ -7,7 +7,18 @@ import dts from 'vite-plugin-dts';
 import relay from 'vite-plugin-relay-lite';
 import svgr from 'vite-plugin-svgr';
 
+import { peerDependencies } from './package.json';
+
 const __dirname = dirname(fileURLToPath(import.meta.url));
+
+// Rollup matches `external` strings exactly, so bare names left subpaths like
+// `react-dom/client` bundled — a second renderer that blank-screens consumers
+// on any other React patch (#8595). i18next / react-i18next are dependencies
+// rather than peers, so they stay bundled and keep BUI's i18n isolated.
+const peerDependencyPatterns = Object.keys(peerDependencies).map(
+  // `.` is the only regex metacharacter an npm package name can hold.
+  (name) => new RegExp(`^${name.replaceAll('.', '\\.')}(/.*)?$`),
+);
 
 export default defineConfig(({ mode }) => {
   const isDevMode = mode === 'development';
@@ -36,16 +47,7 @@ export default defineConfig(({ mode }) => {
         formats: ['es'],
       },
       rollupOptions: {
-        external: [
-          'react',
-          'react-dom',
-          'react-router-dom',
-          'relay-runtime',
-          'antd',
-          'antd-style',
-          'graphql',
-          // i18next and react-i18next are bundled to isolate BUI's i18n instance from the host app
-        ],
+        external: peerDependencyPatterns,
       },
       sourcemap: true,
       outDir: 'dist',


rapsealk:
## Pre-mortem review

Assume this merged, shipped in a release, and something broke. Here are the ways that could happen, each one checked rather than reasoned about. Evidence for every claim is at the bottom.

The diff is three lines of logic, but its blast radius is not the diff — it is **every consumer that was silently relying on a bundled copy of a peer**. That is where I looked.

### 1. A consumer doesn't have a newly-external peer installed → their build breaks

`@ant-design/icons`, `@tanstack/react-query`, `ahooks` and `react-relay` were bundled whole and are now external. A consumer without them installed gets an unresolved import.

There is no `peerDependenciesMeta`, so all 11 peers are already **mandatory** by declaration — this makes the manifest true rather than changing the contract. Both consumers I can inspect (this repo's `react/` app, and `backend.ai-fasttrack`) have all 11 installed.

**Verdict: low, and contractual.** Worth one line in the release notes for anyone who was getting away with not installing a declared peer.

### 2. A consumer *below* the declared peer range now really runs its own older copy

This is the one with teeth. Bundling was insulating consumers from their own version drift; now the declared ranges bite for real.

`backend.ai-fasttrack` today: `@tanstack/react-query` 5.100.10 against `^5.101.2`, `@ant-design/icons` 6.2.3 against `^6.3.2`. (Also `antd` 5.29.3 against `^6.5.0` — a full major behind, but `antd` was already external, so that one predates this PR.)

I built that consumer against this branch's `dist` to see whether it actually holds: `tsc --noEmit` 0 errors against a baseline of 0, and a **clean production build — 9,249 modules, no unresolved import and no missing-export warning**. The app then renders and interacts normally.

**Verdict: acceptable, with a caveat I can't discharge by building.** A static build cannot catch runtime-only API drift (a call into something added in 5.101 would still throw at runtime). Consumers below the ranges should bump; the release notes should say the ranges are now enforced in practice.

### 3. antd subpaths now resolve to the host's antd

`src/` imports ~40 distinct `antd/...` subpaths — `antd/es/form/context`, `antd/es/table`, `antd/es/upload`, 22 `antd/es/locale/*`. All of those were bundled from BUI's own antd 6; now they come from the host.

`BAICheckbox` imports `FormItemInputContext` from `antd/es/form/context`; that export exists in the consumer's antd 5.29.3, so it resolves.

Worth noting this direction is a **fix, not just a risk**: `BAICheckbox` was previously reading a context object from BUI's private antd copy, which no host `<Form>` ever populates. It can now see the host's form state.

### 4. `react/jsx-runtime` and `react/compiler-runtime` newly external

Both are present in react 19.2.6 (I checked the files). `react/compiler-runtime` does not exist before React 19.1, so a consumer on 19.0 would break — but the peer range is `^19.2.7`, so that consumer is already out of contract.

### 5. Types regress the way #8590 did

Checked against the real consumer, since that is where the last one surfaced: `tsc --noEmit` on `backend.ai-fasttrack` against this branch's `dist` reports 0 errors, matching its baseline. Repo-side, all five `typecheck` jobs are green.

### 6. The i18n isolation quietly disappears with the comment that guarded it

The old literal list carried a comment explaining that i18next is deliberately bundled. Deleting the list could have deleted the behavior.

It didn't, and not by luck: `i18next` / `react-i18next` are `dependencies`, not peers, so the rule excludes them by construction. Confirmed in the built bundle — i18next (81 KB) and react-i18next (28 KB) are still inlined.

### 7. Storybook shares this config and breaks

`.storybook/main.ts` has no `viteFinal` / `mergeConfig` touching `rollupOptions`, so the lib externals don't reach the Storybook build.

**Unverified:** I did not run `build-storybook`. If someone wants belt-and-braces before merge, that's the one command I'd add.

### 8. "CI would catch it next time"

It would not, and this is the finding I'd act on regardless of this PR.

`vitest.yml`'s path filter is `packages/backend.ai-ui/src/**`, `package.json`, `vitest.config.ts` — **`vite.config.ts` is not in it**, and no PR job runs `pnpm build` for the package. So the 12 checks on this PR are typecheck ×5, CodeQL/Analyze ×4, and triage/label/assign/CLA: **a change that rewrites the published artifact ships with no build or test gate.** That is also why the react-dom bundling reached npm in the first place.

## What I ran locally, since CI doesn't

| check | result |
|---|---|
| `bash scripts/verify.sh` | `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup, Terminology) |
| `pnpm test` (backend.ai-ui) | 369 passed, 1 skipped |
| `pnpm run build:react-only` | clean — host app resolves every newly external import |
| consumer `tsc --noEmit` (FastTrack, react 19.2.6) | 0 errors, baseline 0 |
| consumer `pnpm build` (FastTrack) | clean, 9,249 modules |
| consumer runtime (FastTrack, react 19.2.6) | renders; 0 console errors; one React DevTools banner instead of two |
| consumer build output | 73.2 MB → 71.0 MB (incl. sourcemaps) |

Both `external` shapes — the array of patterns here and a predicate function — produce a **byte-identical** `dist/backend.ai-ui.js`, so the shape is style, not behavior.

## Verdict

**Good to merge.** The change is minimal, it makes the manifest honest, and the failure modes above are either contractual or already true today. I'd attach two conditions and one follow-up:

1. **Release note:** peer dependencies are now genuinely required and their ranges are now enforced in practice. Consumers below range should bump — concretely, `backend.ai-fasttrack` is behind on `@tanstack/react-query` and `@ant-design/icons`.
2. **Publish a canary first** and let `backend.ai-fasttrack` run on it before a stable release. That repo is currently pinning React to exactly 19.2.7 (lablup/backend.ai-fasttrack#5232) to work around the bug this fixes, and removing that pin is the natural smoke test.
3. **Follow-up (separate PR):** add `packages/backend.ai-ui/vite.config.ts` to `vitest.yml`'s path filter, add a package `build` job to PR CI, and add `publint` / `are-the-types-wrong`. Two packaging regressions have now reached npm in a month (#8590, this one) through the same hole.

The one thing I have *not* proven is runtime behavior of BUI components against host copies of `@tanstack/react-query` and `ahooks` that are older than the declared range — a build cannot show that, and condition 2 is what would.

nowgnuesLee (approved):
LGTM

PR #8630, opened 2026-08-10
feat(docs-toolkit): Accept a per-language `title` in book.config.yaml

### LINKED ISSUE #8629: docs-toolkit: Support a per-language `title` in book.config.yaml
`book.config.yaml` keys `navigation` by language, but `title` is a single global string. Multi-language projects therefore cannot give the document a localized title — the source language's title is stamped onto every page of every translated build.

## Impact, measured

Backend.AI FastTrack ships an English and a Korean PDF from one `book.config.yaml`. In the Korean 26.8.0rc1 build:

- The cover reads **"Backend.AI FastTrack Documentation"** in English, while the date (`8월 2026`) and language (`한국어`) directly beneath it are localized.
- **258 of 260 pages** carry that English string as the running header, above fully Korean body text and Korean footers — the footer *is* localized, because it derives from the document structure.
- Every Korean page's `<title>` reads `Backend.AI FastTrack Documentation - 한국어`.

The same value also feeds the sidebar header and the logo `alt` text, so there is no way to reach any of them per-locale.

## Cause

`loadBookConfig` (`packages/backend.ai-docs-toolkit/src/book-config.ts`) reads the title as a plain string:

```ts
const titleMultiline = (raw.title ?? "").trimEnd();
const title = normalizeTitle(titleMultiline);
```

while navigation in the same file is already per-language:

```ts
for (const [lang, value] of Object.entries(rawNav)) { … }
```

`title` then flows into `metadata.title`, used by the web `<title>` composition, the sidebar, and the PDF header stamp (`pdf-renderer.ts`, `drawText(title, …)` on every page except the cover).

## Proposed

Accept the same shape `navigation` already uses, keeping the plain string working:

```yaml
# current — still valid
title: |
  Backend.AI WebUI
  User Guide

# proposed — per-language, falling back to `default` then to the first entry
title:
  default: "Backend.AI WebUI User Guide"
  ko: "Backend.AI WebUI 사용자 가이드"
  ja: "Backend.AI WebUI ユーザーガイド"
```

Resolution would mirror `branding.subLabel`, which already does exactly this (`subLabelMap[metadata.lang] ?? subLabelMap.default ?? ""`) — so the precedent and the code shape both exist in the toolkit today. `titleMultiline` would resolve per-language the same way, preserving the block-scalar cover rendering.

Backward compatible: a string `title` keeps its current meaning.

## Workaround in the meantime

Reduce the title to the brand name so no source-language words appear in translated builds, and express "guide"/"docs" through `branding.subLabel`, which *is* per-language. That is what FastTrack has done (lablup/backend.ai-fasttrack#5259) — it works, but it costs the English build the descriptive part of its title, which a per-language `title` would not.

Happy to send a PR for the `book-config.ts` change plus tests if the proposed shape looks right.

rapsealk:
Closes #8629.

`book.config.yaml` keys `navigation` by language but treats `title` as a single global string, so a multi-language project stamps the source language's title onto every page of every translated build.

Measured in Backend.AI FastTrack's Korean 26.8.0rc1 PDF: an English cover, and an English running header on **258 of 260** otherwise-Korean pages — above Korean body text and Korean footers, since the footer derives from document structure. Every Korean web page's `<title>` read `… Documentation - 한국어` too.

## Change

Accept the same shape `navigation` already uses:

```yaml
title:
  default: "Backend.AI WebUI User Guide"
  ko: "Backend.AI WebUI 사용자 가이드"
```

Resolution mirrors `branding.subLabel` in `config.ts` — exact language, then `default`, then the first entry — so both the precedent and the code shape already exist in the toolkit.

## Backward compatibility

A string `title` normalizes to a single `default` entry and resolves to the same value for every language, so existing projects see no change. `NormalizedBookConfig.title` / `.titleMultiline` keep their current meaning as the default pair for language-unaware callers; `titleByLang`, `titleMultilineByLang` and `resolveBookTitle(config, lang)` are additive.

Block scalars keep working per language, so a translated cover can still break across lines where the author intended:

```yaml
title:
  default: |
    Backend.AI WebUI
    User Guide
  ko: |
    Backend.AI WebUI
    사용자 가이드
```

## Wiring

Resolved at the two points where the language is known:

- `generate-pdf.ts` — inside the per-language loop: cover, running header, and the `{title}` filename placeholder.
- `website-generator.ts` — flat-mode per-language loop: `<title>` and the sidebar header.

Deliberately left on the **default** title: the root brand assets (`writeRootBrandAssets`) and the `og:site_name` fallback, since one artifact serves every language. Happy to extend if you'd rather those follow the per-language value in versioned mode too.

## Tests

Six cases added to `book-config.test.ts`, covering the compatibility path as much as the new one:

```
ok - title — a plain string resolves identically for every language
ok - title — a block scalar keeps its line breaks in titleMultiline
ok - title — a per-language map resolves each language to its own title
ok - title — a map without `default` falls back to its first entry
ok - title — per-language block scalars keep their line breaks
ok - title — a non-string map entry is dropped, not rendered
```

Full package suite: **309 passed, 0 failed, 1 skipped**. `tsc --noEmit` clean.

The "no `default` key" and "language absent from the map" cases are covered explicitly because the failure mode there is a silently-`undefined` title rather than a loud error.

### DIFF
--- a/packages/backend.ai-docs-toolkit/src/book-config.test.ts
+++ b/packages/backend.ai-docs-toolkit/src/book-config.test.ts
@@ -4,7 +4,11 @@ import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
 
-import { loadBookConfig, normalizeTitle } from "./book-config.js";
+import {
+  loadBookConfig,
+  normalizeTitle,
+  resolveBookTitle,
+} from "./book-config.js";
 
 function withSrcDir(yaml: string, fn: (dir: string) => void): void {
   const dir = fs.mkdtempSync(path.join(os.tmpdir(), "book-config-"));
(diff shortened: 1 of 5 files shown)


yomybaby (approved):
LGTM

PR #8640, opened 2026-08-11
fix(FR-3478): surface the manager's actual error when start-service fails

ironAiken2:
Resolves [FR-3478](https://lablup.atlassian.net/browse/FR-3478)

> The GitHub clone of FR-3478 has not been created by the Jira webhook yet; this reference will be updated to `Resolves #N (FR-3478)` once it appears.

## Problem

Every failure of `POST /session/{id}/start-service` — 500, 404, 401, timeout, network — was reported to the user as **"Proxy configurator is not responding."**, which points at the AppProxy coordinator even though it is not contacted at that point in the flow. At a customer site this mislabelled 35 backend timeouts over 7 days as a proxy problem (BA-7258).

Root cause: the Backend.AI JS client's `_wrapWithPromise` rejects with a **plain object** (`{ statusCode, statusText, title, msg, error_code, description, traceback }`), never an `Error`. The catch in `_resolveV2ProxyUri` used `err instanceof Error ? err.message : 'Proxy configurator is not responding.'`, so the fallback string won for every failure — and `AppLaunchError.originalError` was typed `Error`, so the manager's payload was dropped entirely at wrap time.

## Changes

- **`_resolveV2ProxyUri` catch**: resolve the manager's own text via the new exported `resolveStartServiceErrorMessage` instead of `instanceof Error`. The client's `msg` field is always `undefined` (`_wrapWithPromise` copies it off the fetch `Response`, which never carries one) and its `message` is a debug string (`server responded failure: 503 Service Unavailable - …`), so the resolver takes the clean text from `description` / `response.msg` / `response.title` and delegates to `useErrorMessageResolver.getErrorMessage`, which appends the `error_code`. The user sees e.g. `Session has no scaling group assigned (backendai_generic_unavailable)`.
- **v1 conf path**: "Proxy configurator is not responding." is kept only where it is accurate (the v1 `PUT {proxyURL}/conf` returning `undefined`), now wired to the previously unused i18n key `session.launcher.ProxyConfiguratorNotResponding`.
- **`AppLaunchError`**: `originalError` widened to `unknown` so the plain-object payload is preserved.
- **Failure notification**: `extraDescription` (extracted as `buildAppLaunchFailureExtraDescription`) shows the original rejection payload (including `traceback` / `error_code`) instead of the wrapper's uninformative stack, with `requestParameters` redacted — it echoes the raw start-service request body including `login_session_token`.
- **`_open_wsproxy` catch**: the same `err instanceof Error ? … : …` shape removed here too, so a non-`Error` rejection keeps its payload in `originalError`.
- **i18n**: new key `session.appLauncher.FailedToStartAppService` added to all 22 languages for the start-service fallback message.

## Acceptance criteria (from FR-3478)

- [x] A 500 from start-service shows the manager's title/error_code, not the proxy text — the message is exactly the manager's `msg`/`title` plus `(error_code)`, without the client's debug prefix.
- [x] A 404 (session gone) shows a session-not-found message — existing `isSessionNotFoundError` branch kept (FR-2586), now also preserving the original payload.
- [x] "Proxy configurator is not responding." appears only on the v1 conf path.
- [x] `session.launcher.ProxyConfiguratorNotResponding` is wired to that path.

## Out of scope

- Localizing the remaining hardcoded `AppLaunchError` messages in `useBackendAIAppLauncher.tsx` (local-proxy-not-ready, coordinator/worker connection failures, TCP-app shell copy). They predate this PR and are untouched by it; localizing them is a separate sweep to be filed as a follow-up FR.

## Tests

The failing environments (broken agent, mis-configured resource group, exhausted appproxy worker, …) are hard to reproduce live, so the backend's actual error responses — extracted from the backend.ai sources (main as of 2026-08-11) — are encoded as fixtures and driven through the **real** Backend.AI client (`_wrapWithPromise`) with only `fetch` stubbed:

- `react/src/hooks/__fixtures__/appLaunchBackendErrors.ts` — manager start-service (8), appproxy coordinator (6), worker (6), and network-level failure shapes.
- `react/src/hooks/useBackendAIAppLauncher.errors.test.ts` — 50 tests. The frontend side calls the same exported helpers the hook uses in production (`resolveStartServiceErrorMessage`, `buildAppLaunchFailureExtraDescription`) rather than re-composing their logic in the test, and message assertions are **equality**, not `toContain`, so the client's debug prefix cannot leak back in unnoticed. The `requestParameters` redaction is asserted against the real client rejection: the raw payload contains `login_session_token`, the notification details must not.

The catalog surfaced one behavioral refinement: start-service 404 can also mean **"no such app"** (`error_code: backendai_read_not-found`), which must not be presented as "session not accessible". `isSessionNotFoundError` excludes that one known app-not-found code (a denylist, so an unrecognized error_code fails safe to the FR-2586 behaviour instead of dropping the localized guidance), and keeps the FR-2586 404-only heuristic for older managers that send none.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, i18n structure, terminology)
`vitest run useBackendAIAppLauncher.errors.test.ts` → 50/50 passing

Related: BA-7258, BA-7261

[FR-3478]: https://lablup.atlassian.net/browse/FR-3478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/react/src/hooks/useBackendAIAppLauncher.errors.test.ts
+++ b/react/src/hooks/useBackendAIAppLauncher.errors.test.ts
@@ -1,0 +1,316 @@
+/**
+ * FR-3478 regression tests: every backend failure on the app-launch path must
+ * surface its real cause to the user, never the blanket
+ * "Proxy configurator is not responding." text (which describes only the v1
+ * `/conf` path).
+ *
+ * The error catalog in `__fixtures__/appLaunchBackendErrors.ts` was extracted
+ * from the backend.ai sources; these environments are hard to reproduce live
+ * (they need a broken agent, a mis-configured resource group, an exhausted
+ * appproxy worker, …), so the catalog + these tests stand in for them.
+ *
+ * The manager stage deliberately runs through the REAL Backend.AI client
+ * (`Client._wrapWithPromise`) with only `fetch` stubbed, because the client's
+ * error transformation — rejecting with a plain object instead of an `Error` —
+ * is the root cause of FR-3478. Re-implementing that transformation in the
+ * test would let the two drift apart. The real client is imported by relative
+ * path on purpose: the bare `backend.ai-client` specifier is aliased to a stub
+ * in vitest.config.ts.
+ */
+import { Client } from '../../../packages/backend.ai-client/src/client';
+import { ClientConfig } from '../../../packages/backend.ai-client/src/client-config';
+import {
+  coordinatorErrors,
+  managerErrors,
+  networkLevelFailures,
+  workerErrors,
+  type MockedErrorResponse,
+} from './__fixtures__/appLaunchBackendErrors';
+import {
+  AppLaunchError,
+  getAppProxyErrorMessage,
+  isSendRequestErrorResponse,
+  isSessionNotFoundError,
+  sendRequest,
+} from './useBackendAIAppLauncher';
+import { renderHook } from '@testing-library/react';
+import { useErrorMessageResolver } from 'backend.ai-ui';
+
+const LEGACY_PROXY_TEXT = 'Proxy configurator is not responding.';
+const FALLBACK = 'Failed to start the app service.';
+
+const STATUS_TEXT: Record<number, string> = {
+  400: 'Bad Request',
+  401: 'Unauthorized',
+  403: 'Forbidden',
+  404: 'Not Found',
+  409: 'Conflict',
+  500: 'Internal Server Error',
+  502: 'Bad Gateway',
+  503: 'Service Unavailable',
+};
+
+const stubFetchWith = (mock: MockedErrorResponse) => {
+  vi.stubGlobal(
+    'fetch',
+    vi.fn(
+      async () =>
+        new Response(JSON.stringify(mock.body), {
+          status: mock.status,
+          statusText: STATUS_TEXT[mock.status],
+          headers: mock.headers,
+        }),
+    ),
+  );
+};
+
+const getErrorMessage = () => {
+  const { result } = renderHook(() => useErrorMessageResolver());
+  return result.current.getErrorMessage;
+};
+
+/**
+ * Drive the real client's start-service call against the stubbed fetch and
+ * capture the rejection — the exact value `_resolveV2ProxyUri`'s catch sees.
+ */
+const startServiceRejection = async (
+  mock: MockedErrorResponse,
+): Promise<unknown> => {
+  stubFetchWith(mock);
+  const client = new Client(
+    new ClientConfig(
+      'AKIADUMMYACCESSKEY',
+      'dummy-secret',
+      'http://mock-manager',
+    ),
+    'test-agent',
+  );
+  try {
+    await client.computeSession.startService(
+      'login-session-token',
+      'e961a85c-29b3-429e-b140-b01dcd80b689',
+      'jupyter',
+    );
+  } catch (err) {
+    return err;
+  }
+  throw new Error('expected startService to reject');
+};
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+});
+
+describe('stage 1 — manager start-service rejections through the real client', () => {
+  it.each(Object.entries<MockedErrorResponse>(managerErrors))(
+    '%s: rejects with a plain object carrying statusCode/error_code',
+    async (_name, mock) => {
+      const err = await startServiceRejection(mock);
+
+      // Root cause of FR-3478: the client never rejects with an Error, so an
+      // `err instanceof Error` branch always falls through to its fallback.
+      expect(err).not.toBeInstanceOf(Error);
+      expect(err).toMatchObject({
+        statusCode: mock.status,
+        error_code: mock.body.error_code,
+      });
+    },
+  );
+
+  it.each(Object.entries<MockedErrorResponse>(managerErrors))(
+    '%s: resolves to the manager message + error_code, not the proxy text',
+    async (_name, mock) => {
+      const err = await startServiceRejection(mock);
+      const resolved = getErrorMessage()(err, FALLBACK);
+
+      expect(resolved).not.toBe(LEGACY_PROXY_TEXT);
+      expect(resolved).not.toBe(FALLBACK);
+      // The client folds `body.msg` (or `body.title` when msg is absent) into
+      // its `message`, and getErrorMessage appends the error_code.
+      expect(resolved).toContain(mock.body.msg ?? mock.body.title);
+      expect(resolved).toContain(`(${mock.body.error_code})`);
+    },
+  );
+
+  it.each(Object.entries<MockedErrorResponse>(managerErrors))(
+    '%s: AppLaunchError preserves the payload for the notification',
+    async (_name, mock) => {
+      const err = await startServiceRejection(mock);
+      const wrapped = new AppLaunchError(
+        getErrorMessage()(err, FALLBACK),
+        'configuring',
+        err,
+      );
+
+      expect(wrapped.statusCode).toBe(mock.status);
+      expect(wrapped.errorCode).toBe(mock.body.error_code);
+      // The original payload (incl. traceback/response) must survive for the
+      // notification's extraDescription.
+      expect(wrapped.originalError).toBe(err);
+    },
+  );
+
+  it('treats only session-domain 404s as "session not accessible"', async () => {
+    // 404 with error_code "session_read_not-found" → the session itself is
+    // gone (or created under another access key): show SessionNotAccessible.
+    expect(
+      isSessionNotFoundError(
+        await startServiceRejection(managerErrors.sessionNotFound),
+      ),
+    ).toBe(true);
+
+    // 404 with error_code "backendai_read_not-found" → the *app* was not
+    // found in service_ports; the session is fine. Masking this as "session
+    // not accessible" would misdirect the user.
+    expect(
+      isSessionNotFoundError(
+        await startServiceRejection(managerErrors.appNotFound),
+      ),
+    ).toBe(false);
+
+    // Pre-24.09 managers send no error_code — keep the FR-2586 heuristic.
+    expect(isSessionNotFoundError({ statusCode: 404 })).toBe(true);
+    expect(isSessionNotFoundError({ statusCode: 500 })).toBe(false);
+  });
+
+  it('distinguishes same-title 503 causes only via msg, never via title', async () => {
+    // noScalingGroup and noCoordinator share status, type, the (typo'd)
+    // title "Serivce unavailable." and even error_code — msg is the only
+    // discriminator, and it must reach the user.
+    const scalingGroupMsg = getErrorMessage()(
+      await startServiceRejection(managerErrors.noScalingGroup),
+      FALLBACK,
+    );
+    const coordinatorMsg = getErrorMessage()(
+      await startServiceRejection(managerErrors.noCoordinator),
+      FALLBACK,
+    );
+
+    expect(scalingGroupMsg).toContain('Session has no scaling group assigned');
+    expect(coordinatorMsg).toContain(
+      'No coordinator configured for this resource group',
+    );
+    expect(scalingGroupMsg).not.toBe(coordinatorMsg);
+  });
+});
+
+describe('stage 2/3 — appproxy coordinator & worker errors through sendRequest', () => {
+  const appproxyCatalog: Array<[string, MockedErrorResponse]> = [
+    ...Object.entries<MockedErrorResponse>(coordinatorErrors),
+    ...Object.entries<MockedErrorResponse>(workerErrors),
+  ];
+
+  it.each(appproxyCatalog)(
+    '%s: sendRequest returns a detectable error shape with the real message',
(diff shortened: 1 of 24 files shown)


react/src/hooks/useBackendAIAppLauncher.tsx
```diff
@@ -800,9 +806,19 @@ export const useBackendAIAppLauncher = (
...
           rejected: (error: any) => {
+            // Prefer the original rejection payload (manager problem+json with
+            // statusCode/error_code/traceback) over the AppLaunchError wrapper
+            // stack, which carries no diagnostic information (FR-3478).
+            const original = error?.originalError;
+            const extraDescription =
+              original instanceof Error
+                ? original.stack
+                : original
+                  ? JSON.stringify(original, null, 2)
```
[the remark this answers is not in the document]
ironAiken2: Fixed in 950fe40a1 — the notification's diagnostic payload now serializes the rejection with `requestParameters` omitted (it echoes the raw start-service body incl. `login_session_token`), while keeping statusCode/error_code/traceback/response intact.

react/src/hooks/useBackendAIAppLauncher.tsx
```diff
@@ -1084,35 +1110,63 @@ export const useBackendAIAppLauncher = (
...
+export function isSessionNotFoundError(err: unknown): boolean {
   if (!err || typeof err !== 'object') return false;
-  const e = err as { statusCode?: unknown };
-  return e.statusCode === 404;
+  const e = err as { statusCode?: unknown; error_code?: unknown };
+  if (e.statusCode !== 404) return false;
+  if (typeof e.error_code === 'string') {
+    return e.error_code.startsWith('session_');
+  }
+  return true;
```
agatha197, line 1121: Suggested flip to a denylist, so an unrecognized `error_code` keeps the FR-2586 behaviour instead of losing the localized guidance:

```suggestion
  // `backendai_read_not-found` means the *app* is absent from service_ports,
  // not the session. Deliberately a denylist rather than an allowlist on the
  // `session_` domain: this backend contract cannot be verified from the
  // frontend, so an unrecognized code must keep the FR-2586 behaviour instead
  // of silently dropping the localized recovery guidance.
  return e.error_code !== 'backendai_read_not-found';
```

The JSDoc just above would need a matching tweak — "when one is present we additionally require the `session` domain" describes the allowlist.

react/src/hooks/useBackendAIAppLauncher.errors.test.ts
```diff
@@ -0,0 +1,316 @@
...
+    // found in service_ports; the session is fine. Masking this as "session
+    // not accessible" would misdirect the user.
+    expect(
+      isSessionNotFoundError(
+        await startServiceRejection(managerErrors.appNotFound),
+      ),
+    ).toBe(false);
+
+    // Pre-24.09 managers send no error_code — keep the FR-2586 heuristic.
+    expect(isSessionNotFoundError({ statusCode: 404 })).toBe(true);
```
agatha197, line 212: Pins the fail-safe direction, which the fixture-driven cases structurally cannot:

```suggestion
    // Pre-24.09 managers send no error_code — keep the FR-2586 heuristic.
    expect(isSessionNotFoundError({ statusCode: 404 })).toBe(true);
    expect(isSessionNotFoundError({ statusCode: 500 })).toBe(false);

    // Fail-safe direction: an error_code we do not recognize (e.g. the manager
    // renames its domain prefix) must NOT strip the localized guidance. These
    // literals are intentionally not sourced from the fixture — reading them
    // from the same catalog the production code is matched against would make
    // the assertion self-fulfilling.
    expect(
      isSessionNotFoundError({
        statusCode: 404,
        error_code: 'compute-session_read_not-found',
      }),
    ).toBe(true);
```

agatha197 (changes requested):
The core fix is right — resolving the manager's payload instead of `instanceof Error` is exactly what BA-7258 needed. One change request, on the part that goes beyond it.

`isSessionNotFoundError` now **allowlists** the `session_` error_code domain. Since both 404s share `type: probs/object-not-found` and title-matching is (correctly) off the table, `error_code` genuinely is the only discriminator — the coupling is unavoidable. What *is* avoidable is its failure direction.

As an allowlist, a manager-side rename of the `session_` prefix silently drops `SessionNotAccessible` — and that string isn't a label, it's the recovery instruction ("switch to the access key that owns the session, or create a new session") in 22 languages. Users would get a bare "No such session." instead.

Flipping it to a **denylist** on the one known-misclassified code makes every drift path degrade to at worst the pre-FR-3478 behaviour, never a regression:

| scenario | allowlist (current) | denylist |
|---|---|---|
| `session_read_not-found` | guidance shown | same |
| `backendai_read_not-found` (app missing) | manager message | same |
| no error_code (older manager) | guidance shown | same |
| manager renames the `session_` prefix | **guidance lost** | guidance kept |
| manager renames the app-not-found code | fine | misclassified — i.e. today's behaviour |

Worth noting this is also the one thing the fixtures can't catch: the tests assert against the same fixture that defines the code, so a rename would keep all 43 green. The second suggestion pins the intent with literals instead — deliberately not sourced from the fixture.

react/src/hooks/useBackendAIAppLauncher.tsx
ironAiken2: Applied in 06fab54d8 — flipped to the denylist as suggested, and the JSDoc now reads "we exclude that one known app-not-found code" instead of the allowlist wording. One knock-on: the existing `coordinator 404 (token not found)` test asserted the allowlist direction (`agent_read_not-found` → false), so it now pins the fail-safe direction instead, with a comment noting that coordinator failures are routed through `sendRequest`/`getAppProxyErrorMessage` and never reach `isSessionNotFoundError` anyway.

react/src/hooks/useBackendAIAppLauncher.errors.test.ts
ironAiken2: Applied in 06fab54d8 verbatim — the `compute-session_read_not-found` literal is kept out of the fixture catalog as you specified. Full suite: 43/43 passing, `verify.sh` → ALL PASS.

react/src/hooks/useBackendAIAppLauncher.errors.test.ts
```diff
@@ -0,0 +1,331 @@
...
+      expect(resolved).toContain(mock.body.msg ?? mock.body.title);
+      expect(resolved).toContain(`(${mock.body.error_code})`);
+    },
+  );
+
+  it.each(Object.entries<MockedErrorResponse>(managerErrors))(
+    '%s: AppLaunchError preserves the payload for the notification',
+    async (_name, mock) => {
+      const err = await startServiceRejection(mock);
+      const wrapped = new AppLaunchError(
```
agatha197, line 150: These tests never call `_resolveV2ProxyUri` or the notification `rejected` handler — they re-implement the production composition right here, so reverting the fix at `useBackendAIAppLauncher.tsx:281-285` back to `err instanceof Error ? err.message : 'Proxy configurator is not responding.'` still leaves all 43 green.
The `requestParameters` redaction has no coverage at all; please extract the `extraDescription` builder and assert it never leaks `login_session_token`.

react/src/hooks/useBackendAIAppLauncher.tsx
```diff
@@ -267,15 +270,18 @@ export const useBackendAIAppLauncher = (
...
+      // The Backend.AI client rejects with a plain object
+      // (`{ statusCode, title, msg, error_code, ... }`), never an Error, so
+      // resolve the manager's own message instead of pattern-matching on
+      // `instanceof Error` — that check made every start-service failure
+      // surface as "Proxy configurator is not responding." (FR-3478).
       throw new AppLaunchError(
-        err instanceof Error
-          ? err.message
-          : 'Proxy configurator is not responding.',
+        getErrorMessage(err, t('session.appLauncher.FailedToStartAppService')),
```
agatha197: `_wrapWithPromise` fills `msg: respOverlay.msg` from the fetch `Response`, where `.msg` is never assigned (`client.ts:334`, `:494`), so `getErrorMessage` always falls through to `message` and the user gets `server responded failure: 503 Service Unavailable - Session has no scaling group assigned (backendai_generic_unavailable)`.
The clean text sits on the same rejection as `description` / `response.msg` — prefer those over `message`. The test's `toContain(msg)` (test:130) hides this; an equality assertion would not.

react/src/hooks/useBackendAIAppLauncher.tsx
```diff
@@ -1084,35 +1110,68 @@ export const useBackendAIAppLauncher = (
...
-class AppLaunchError extends Error {
+export class AppLaunchError extends Error {
   stage: 'detecting' | 'configuring' | 'requesting' | 'connecting';
-  originalError?: Error;
+  // The Backend.AI client rejects with a plain object, not an Error
+  // (see `_wrapWithPromise` in packages/backend.ai-client), so this must be
+  // `unknown` — typing it `Error` silently dropped the manager's payload
+  // (statusCode/error_code/traceback) at wrap time (FR-3478).
+  originalError?: unknown;
+  statusCode?: number;
```
agatha197: Nothing outside the new tests reads `statusCode`/`errorCode` — no production code branches on them, and `originalError` already carries the whole payload.
Please drop both (and the two guard tests that only cover them) until a real call site needs them.

react/src/hooks/useBackendAIAppLauncher.tsx
```diff
@@ -200,7 +200,7 @@ export const useBackendAIAppLauncher = (
     const response = await sendRequest(rqst);
     if (response === undefined) {
       throw new AppLaunchError(
-        'Proxy configurator is not responding.',
+        t('session.launcher.ProxyConfiguratorNotResponding'),
```
agatha197, line 203: Only 2 of the 7 `AppLaunchError` messages in this file end up localized — 187, 371, 377, 523 and 944 stay hardcoded English right next to these.
Also `_open_wsproxy`'s catch at 432-438 still uses the `err instanceof Error ? err.message : …` / `err instanceof Error ? err : undefined` shape this PR removes; either finish the pass or state it as out of scope in the description.

agatha197 (changes requested):
A second pass beyond the threads already resolved. Four points: the tests do not actually pin the FR-3478 regression, the resolved message still reaches the user as the client's debug string, two dead fields, and a half-finished pass over the file.

react/src/hooks/useBackendAIAppLauncher.errors.test.ts
ironAiken2: Fixed in d19cdf6ca — the message resolution and the extraDescription builder are now exported pure functions (`resolveStartServiceErrorMessage`, `buildAppLaunchFailureExtraDescription`) that the hook itself calls, and the tests call those same functions instead of re-composing the logic. Message assertions are equality now, so the debug prefix cannot slip back in. The redaction has direct coverage: a new per-fixture test asserts the raw client rejection DOES contain `login_session_token` (sanity) and the builder output does not, plus no `requestParameters` key at all.

react/src/hooks/useBackendAIAppLauncher.tsx
ironAiken2: Fixed in d19cdf6ca — confirmed: `respOverlay.msg` is never assigned anywhere in the catch (only `status`/`statusText` are), so the rejection's `msg` is always undefined and the resolver always fell through to the debug `message`. The new `resolveStartServiceErrorMessage` picks the clean text from `description` / `response.msg` / `response.title` (in that order, after `msg` itself so a future client that fills it wins) and slots it into `getErrorMessage`'s preferred field. The user now sees exactly `Session has no scaling group assigned (backendai_generic_unavailable)`. Kept at the call site rather than changing the shared BUI `getErrorMessage` priority, to avoid a blast radius over its other callers. The `toContain` assertions are now `toBe` equality against the full expected string.

react/src/hooks/useBackendAIAppLauncher.tsx
ironAiken2: Fixed in d19cdf6ca — dropped both fields, the constructor extraction block, and the two guard tests that only covered them. `originalError` remains the single carrier of the payload.

react/src/hooks/useBackendAIAppLauncher.tsx
ironAiken2: Split in d19cdf6ca: the `_open_wsproxy` catch is fixed in this PR — it is the same payload-dropping `instanceof Error` shape this PR exists to remove, so it now goes through `getErrorMessage` and always preserves `err` as `originalError`. Localizing the 5 remaining hardcoded messages (187, 371, 377, 523, 944) is declared out of scope in the PR description — they predate this PR, and a 22-language sweep for them belongs in a follow-up FR rather than widening this fix.

agatha197 (approved):
LGTM

PR #8675, opened 2026-08-11
fix(BAISelect): disable Astryx internal search when showSearch is an object (Fixes #8667)

### LINKED ISSUE #8667: Image environment selector: clicking an option does not select the image (Session Launcher, Deployment revision)
## Component

- **Name**: the image Environment / Version selector — `ImageEnvironmentSelectFormItems` (`react/src/components/ImageEnvironmentSelectFormItems.tsx`), which renders two `BAISelect` instances (`packages/backend.ai-ui/src/components/BAISelect.tsx`)
- **Astryx primitive**: `Select` (via `BAISelect`, children option API with `BAISelectOptionItem` / `BAISelectOptionGroup`)
- **Scope**: every usage — the reporter reproduced it on both the Session Launcher and the Deployment revision flow

## Where it is reachable

- Session Launcher — the "Environments / Version" step (menu: Sessions → Start)
- Deployment revision — `DeploymentAddRevisionModal.tsx`, the image selection field
- Also used in: 4 non-test files import `ImageEnvironmentSelectFormItems` or its `Image` type

(`SessionLauncherPage.tsx`, `DeploymentAddRevisionModal.tsx`,

`SessionFormItems/ResourceAllocationFormItems.tsx`, `hooks/useDefaultImagesWithFallback.ts`).

This is usage breadth from a grep, not a per-usage reproduction.

## Steps to reproduce

1. Open the Session Launcher (Sessions → Start) — or the Deployment revision modal.
1. Click the **Environments** selector; the option list opens.
1. Click any image option in the list.
1. Repeat on the **Version** selector.

## Observed

The image cannot be selected at all:

- clicking an option in the list has no effect — the option does not become

selected and the list stays as it was; and

- in the cases where a click does register, the chosen image is not reflected

back into the form field (the selector / form keeps its previous value).

Both the Environments and the Version selector are affected, and it happens on

the Session Launcher and the Deployment revision flow alike, so no session or

deployment can be created through the normal image-selection path.

## Expected

Clicking an option selects it: the dropdown closes, the trigger shows the chosen

environment/image, and the form value (`environments.environment`,

`environments.version`, `environments.image`) is updated so the Session Launcher

and the Deployment revision modal can proceed to the next step.

## Environment

- **Mode**: Unknown (not reported; not mode-specific as far as reported)
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown — reported on the current Astryx UI

## Impact

Blocks the task — image selection is a required step for both session creation

and deployment revision creation, and no workaround was reported. The reporter

flagged it as urgent.

## Evidence

- Screenshot / recording: none supplied.

## Notes

Reported in Korean; both symptom variants above are the reporter's own wording

("항목 클릭이 안 먹힘", "선택은 되지만 값이 반영 안 됨"). Recorded as one issue because

they were reported as one defect on the same control; split later if triage finds

two mechanisms.

JIRA Issue: FR-3499

waterWang:
## Summary

When `showSearch` is passed as an object (antd compatibility mode with `searchValue`/`onSearch`/`optionFilterProp`), the `hasSearch` prop was incorrectly set to `true` via the `showSearch !== false` condition. This activated the Astryx Selector's internal search mode, which:

- Intercepts option clicks, making them unselectable
- Leaves the popup open after clicking an option
- Prevents the image selection from being reflected in the form field

## Fix

Changed `hasSearch: showSearch !== false` to `hasSearch: showSearch === true`. This ensures the Astryx internal search is only enabled when `showSearch` is explicitly a boolean `true`. When it's an object, the calling component manages its own external search via `searchValue`/`onSearch`, and the Astryx internal search is not needed.

## Related Issue

Fixes #8667

### DIFF
--- a/packages/backend.ai-ui/src/components/BAISelect.tsx
+++ b/packages/backend.ai-ui/src/components/BAISelect.tsx
@@ -492,7 +492,7 @@ function BAISelect<ValueType = any, OptionType = BAISelectOption>({
     // `baseShowSearch ?? true`, so a `BAISelect` that says nothing about
     // search was type-to-filter. Treating `undefined` as "no search" turned
     // that off on every call site that omitted the prop.
-    hasSearch: showSearch !== false,
+    hasSearch: showSearch === true,
     status:
       status === 'error' || status === 'warning' ? { type: status } : undefined,
     size:


yomybaby:

PR #8680, opened 2026-08-11
fix(FR-3500): distinguish disabled action buttons on info notification card background

### LINKED ISSUE #8676: Session notification card: disabled action buttons indistinguishable from enabled ones on the info background
## Where

- **Page**: global — session notification popups/list (notification cards shown top-right; also rendered in the notification drawer)
- **Region**: the action icon-button row (app dialog / terminal / log / terminate) inside a session notification card
- **Component** (if identified): `BAIComputeSessionNodeNotificationItem` (`react/src/components/BAIComputeSessionNodeNotificationItem.tsx`) rendering `SessionActionButtons` (`react/src/components/ComputeSessionNodeItems/SessionActionButtons.tsx`)

## Observed

Session notification cards now render with an info (light blue) background. On that background, disabled action icon buttons look the same as enabled ones — in the reporter's screenshot of two PENDING session notifications, the 1st and 2nd buttons from the left are disabled but show no visible difference (same icon color/strength) from the enabled buttons next to them.

## Expected

Disabled action buttons are visually distinguishable from enabled ones (e.g. clearly dimmed icon and/or disabled surface), including when the card has the info background.

## Environment

- **Mode**: light
- **Viewport**: Unknown
- **Browser/OS**: Unknown

## Reproduction

1. Start a session so a session notification card appears (status PENDING).
1. Look at the action icon buttons on the notification card — some are disabled while the session is PENDING.
1. Compare disabled vs enabled buttons on the info-colored card background.

## Evidence

- Screenshot: reporter's capture `clip-20260811-170711.png` (to be attached in Jira)

## Notes

Same symptom family as FR-3494 (content contrast lost against an info banner background); unverified whether the surface is shared.

JIRA Issue: FR-3506

waterWang:
## Summary

Session notification cards render with an info (light blue) background. On that background, disabled action icon-buttons (app dialog / terminal / log / terminate) inside a session notification card look the same as enabled ones — the reporter's screenshot of two PENDING session notifications showed the 1st and 2nd buttons from the left disabled but visually identical to the enabled buttons next to them.

**Root cause.** Astryx `Button` applies `opacity: 0.5` when disabled, but on the info (light blue) background the icon still reads as enabled.

## Fix

Add a local CSS rule in `react/src/components/astryx-bui/astryxBui.css` that reduces the disabled opacity to `0.25` when the button sits inside an info-status notification stack item (`.bai-notification-stack-item[data-status="info"]`), so the disabled state is clearly distinguishable.

The selector targets both native `[disabled]` and `[aria-disabled="true"]` (used when a tooltip keeps the button focusable). It only applies inside the notification stack, so no other surface changes.

Closes #8676

### DIFF
--- a/react/src/components/astryx-bui/astryxBui.css
+++ b/react/src/components/astryx-bui/astryxBui.css
@@ -303,3 +303,13 @@
   min-width: 0;
   overflow-wrap: anywhere;
 }
+
+/* #8676 — Session notification card: disabled action buttons indistinguishable
+   from enabled ones on the info background.
+   Astryx Button already applies `opacity: 0.5` when disabled, but on the
+   info (light blue) background the icon still reads as enabled. Reduce opacity
+   further so the disabled state is visually distinct. */
+.bai-notification-stack-item[data-status="info"]
+  button:where([disabled], [aria-disabled="true"]) {
+  opacity: 0.25;
+}


yomybaby:

PR #8685, opened 2026-08-11
fix(FR-3181): reset auto-mounted folders when switching projects (Closes #7992)

### LINKED ISSUE #7992: Start Session: dot-folder automounts not reset when switching projects
## Bug Report

When switching projects (via the global project selector in the UI header) while the Start Session dialog is open, the previous project's dot-file VFolders (folders whose name starts with `.`) remain in the automount list instead of being replaced with the new project's dot files.

## Steps to Reproduce

1. Open the Start Session dialog (SessionLauncherPage) and navigate to the **Storage** step
1. Observe the `Automount Folders` section — note the dot-file folders listed for the current project
1. While the dialog is open, switch to a different project using the global project selector in the header
1. The automount section now shows the **previous project's dot files** in addition to (or instead of correctly showing) the new project's dot files

## Expected Behavior

When the project changes, the automount folder list should update to reflect **only** the new project's accessible dot-file VFolders.

## Root Cause

In `react/src/components/VFolderTable.tsx`, there are two `useEffect` hooks:

1. `useEffect([autoMountedFolderNames])` — calls `onChangeAutoMountedFolders(autoMountedFolderNames)` to sync the form field
1. `useEffect([currentProject.id])` — resets `selectedRowKeys` (manual mount selections) on project change

The project-change effect resets manual folder selections but does NOT explicitly reset the `autoMountedFolderNames` form field. If `autoMountedFolderNames` doesn't change reference (e.g., same user-owned dot folders accessible in both projects), the first effect won't fire on project change, leaving the form field with stale data.

## Affected Component

`react/src/components/VFolderTable.tsx` — effects around lines 303–317

JIRA Issue: FR-3181

waterWang:
### Problem

When switching projects via the global project selector in the UI header while the **Start Session** dialog is open, the previous project's dot-folder VFolders (folders whose name starts with `.`) remain in the automount list instead of being replaced with the new project's dot files.

### Root cause

`VFolderTableFormItem` keyed the `VFolderTable` component only by `ownerEmail`. Switching projects (same owner) did **not** remount the table, so:

1. The `allFolderList` query was not re-fetched for the new project.
2. The `autoMountedFolderNames` effect did not re-fire on project change, so the hidden `autoMountedFolderNames` form field kept the previous project's dot-folder names.

### Fix

- **`VFolderTableFormItem.tsx`**: Key `VFolderTable` by `ownerEmail` **and** `currentProject.id`. This forces the table to fully remount (and refetch) when the project changes.
- **`VFolderTable.tsx`**: Add `currentProject.id` to the `autoMountedFolderNames` effect dependencies so the parent form's hidden field is updated on project switch.

### Test plan

1. Open Start Session dialog -> Storage step.
2. Note the automount folders listed for the current project.
3. Switch to a different project via the header selector while the dialog is open.
4. The automount section now shows the new project's dot files (old project's dot files are gone).

Closes #7992

### DIFF
--- a/react/src/components/VFolderTable.tsx
+++ b/react/src/components/VFolderTable.tsx
@@ -303,7 +303,7 @@ const VFolderTable: React.FC<VFolderTableProps> = ({
       onChangeAutoMountedFolders(autoMountedFolderNames);
     // Omit `onChangeAutoMountedFolders` from deps so a parent re-render doesn't retrigger this effect
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [autoMountedFolderNames]);
+  }, [autoMountedFolderNames, currentProject.id]);
 
   useEffect(() => {
     // Only reset selectedRowKeys when currentProject changes if there are no controlled selectedRowKeys
--- a/react/src/components/VFolderTableFormItem.tsx
+++ b/react/src/components/VFolderTableFormItem.tsx
@@ -12,6 +12,7 @@ import VFolderTable, {
   vFolderAliasNameRegExp,
 } from './VFolderTable';
 import { useEventNotStable } from 'backend.ai-ui';
+import { useCurrentProjectValue } from '../hooks/useCurrentProject';
 import * as _ from 'lodash-es';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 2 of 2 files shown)


yomybaby:

Issue #8674, opened 2026-08-11
Header project selector: option hover paints a dark background in light mode, label unreadable


ironAiken2:
## Where

- **Page**: app header — global (observed on Sessions → Start a new session, `/session/start`)
- **Region**: the header project selector's dropdown — option hover state
- **Component** (if identified): `ProjectSelect` (`react/src/components/ProjectSelect.tsx`) → BUI `BAISelect` (`packages/backend.ai-ui/src/components/BAISelect.tsx`); rendered by `react/src/components/MainLayout/WebUIHeader.tsx`

## Observed

In light mode, hovering a non-selected option in the header project selector dropdown (e.g. `default`) paints a near-black opaque background while the option label stays dark, so the label is unreadable while hovered. Non-hovered options and the checked option render normally on the white dropdown surface.

## Expected

Hovering an option shows a light-mode hover highlight with the label still readable (subtle light hover token, as in other light-mode dropdown surfaces).

## Environment

- **Mode**: light
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: `feat/FR-3250-session-priority-admin-list` @ 2026-08-11 (includes `main` post-Astryx merge, FR-3482)

## Reproduction

1. Use light mode.
1. Click the project selector in the app header.
1. Hover any non-selected option in the dropdown.

## Evidence

- Screenshot: reporter's capture `clip-20260811-170607.png` (to be attached in Jira)

## Notes

Same symptom family as FR-3493 (header user dropdown hover inverts, hiding the label) and FR-3501 (header band button hover paints an opaque black block in light mode) — possibly the same header-band surface; unverified.

JIRA Issue: FR-3505

Issue #8670, opened 2026-08-11
Header band: content stays white in dark mode instead of inverting against the app mode


yomybaby:
## Where

- **Page**: every authenticated page — the band is the main layout header, so any route reproduces it (e.g. `/dashboard`, menu: any)
- **Region**: the whole orange header band — the project selector on the left (its border and its up/down chevron icon), the `Project` label, and the notification / theme toggle / help / user-name controls on the right
- **Component** (if identified): `WebUIHeader` (`react/src/components/MainLayout/WebUIHeader.tsx`) and the controls it mounts — `ProjectSelect`, `BAINotificationButton`, `WebUIThemeToggleButton`, `WEBUIHelpButton`, `UserDropdownMenu`

## Observed

The band's content keeps the **same light/white colouring in both app modes** —

the project selector's border and chevron glyph, and the text and icons of the

right-hand buttons, are white in light mode and stay white in dark mode.

## Expected

The band's content is **inverted against the app mode**, i.e. the opposite

polarity to the page:

- **light mode** — content is white (project selector border, chevron, button

text and icons)

- **dark mode** — content is black

Scope confirmed by the reporter: **all header band content**, including the

`Project` label on the left, not just the controls named above.

## Environment

- **Mode**: both (the two modes currently render the same, and dark is the one

that is wrong)

- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown (current Astryx UI)

## Reproduction

1. Log in.
1. Look at the header band in light mode — project selector border, its chevron, and the right-hand button labels/icons.
1. Switch to dark mode with the header theme button and look again.

## Evidence

- Screenshot: none

## Notes

The reporter describes the band as deliberately carrying a "reversed" theme, and

reads today's constant-white content as the regression. Worth checking at fix

time against whatever the band's polarity was specified to be — a previous round

may have settled on constant-white on purpose.

JIRA Issue: FR-3502

Issue #8669, opened 2026-08-11
Header band: button hover paints an opaque black block in light mode


yomybaby:
## Where

- **Page**: every authenticated page — the band is the main layout header, so any route reproduces it (e.g. `/dashboard`, menu: any)
- **Region**: the icon/label buttons on the right of the orange header band — notification, theme toggle, help, and the user-name button
- **Component** (if identified): `WebUIHeader` (`react/src/components/MainLayout/WebUIHeader.tsx`) and the controls it mounts — `BAINotificationButton`, `WebUIThemeToggleButton`, `WEBUIHelpButton`, `UserDropdownMenu`

## Observed

In **light mode**, hovering any of these buttons paints an opaque **black**

background on the brand-orange band.

In **dark mode** the hover reads correctly — a translucent black wash over the

band — and is not part of this report.

The reporter checked the pre-Astryx build and the light-mode hover did not look

like this there.

## Expected

The hover wash follows the band, with the polarity flipped per app mode:

- **light mode** — a translucent **white** overlay (lightens the orange)
- **dark mode** — a translucent **black** overlay (current behaviour, keep)

Translucent in both cases, so the brand orange stays visible through the hover

rather than being replaced by a flat block.

## Environment

- **Mode**: light (dark is correct)
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown (current Astryx UI)

## Reproduction

1. Log in.
1. Make sure the app is in light mode.
1. Hover the notification / theme / help / user-name button in the top-right of the header.

## Evidence

- Screenshot: none

## Notes

The header band's content colours are reported separately as a sibling issue —

same band, different symptom.

JIRA Issue: FR-3501

Issue #8662, opened 2026-08-11
Start page: announcement content (tables, code, quotes) blends into the info banner background


agatha197:
## Where

- **Page**: `/start` (menu: Start)
- **Region**: the announcement banner at the top of the page, above the start cards
- **Component** (if identified): `AnnouncementAlert` (`react/src/components/AnnouncementAlert.tsx`) — renders Astryx `Banner` with `status="info"`; the markdown body goes through `markdown-to-jsx` into the Banner's `description` slot. Only consumer is `StartPage.tsx`.

## Observed

The announcement content sits on the info-status Banner background, and everything inside the announcement renders in that same flat background color: table columns have no visible separation, code highlighting shows no distinct background, and blockquotes show no distinct marking. Because there is no color variation between these content elements and the banner surface, the structured markdown content is hard to distinguish.

## Expected

Only the banner's status icon carries the info color. The banner's background matches the surface color of the start cards below it, so the announcement's inner content (table column separation, code highlight background, blockquote marking) reads with normal contrast against a neutral card-like surface.

## Environment

- **Mode**: Unknown
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown

## Reproduction

1. Publish an announcement whose markdown body contains a table, a fenced code block, and a blockquote (superadmin: edit via the announcement editor).
1. Open the Start page (`/start`).
1. Look at the announcement banner: the table/code/quote regions blend into the info background.

## Evidence

- Screenshot: none provided

## Notes

The reporter suspects the uniform look comes from the announcement living inside an info alert surface. Related: FR-3402 covers a different symptom on the same announcement rendering surface (blockquotes render differently between editor preview and published banner).

JIRA Issue: FR-3494

PR #8697, opened 2026-08-11
fix(FR-3506): drop the fill from disabled session action buttons

### LINKED ISSUE #8676: Session notification card: disabled action buttons indistinguishable from enabled ones on the info background

yomybaby:
Resolves #8676 (FR-3506)

Supersedes **PR #8680**, which is **closed**. That PR lowered the disabled opacity `0.5 → 0.25` via scoped CSS on `.bai-notification-stack-item[data-status="info"]`. It aimed at the wrong lever (see *Why not lower the opacity* below), covered only one of the two surfaces these buttons appear on, and sat at the bottom rung of the `astryx-fix` fix ladder. Nothing from it is carried over.

## Mechanism

Not "opacity 0.5 is too weak" — a **token collapse inside the info banner**, plus the fact that Astryx's disabled state is opacity-only.

1. Astryx expresses disabled as **one flat `opacity: 0.5` on the whole button** and changes **no token** — `Button.tsx:88-96` `styles.disabled` is `{cursor, opacity: 0.5, backgroundImage: 'none', transform}`. A filled chip therefore survives disabling as a *solid, still-painted block*, just a step paler.
2. `neutralTheme`'s `banner → 'status:info'` rule (`@astryxdesign/theme-neutral/src/neutralTheme.ts:533-540`, shipped into `react/src/astryx-theme/built/backendai-default-built.css:477-483`) collapses three separate roles onto one value:

   ```css
   .astryx-banner.info {
     background-color: var(--color-background-blue);
     --color-accent-muted: transparent;
     --color-text-primary:   var(--color-text-blue);
     --color-text-secondary: var(--color-text-blue);
     --color-accent:         var(--color-text-blue);   /* a primary button's FILL */
   }
   ```

   So inside the notice a **primary button's fill** and a **secondary button's glyph** are literally the same colour, while a secondary button's own fill (`--color-neutral`, *not* remapped) is near-invisible against the tint.
3. Result, measured: the **disabled** primary chip out-contrasts every **enabled** secondary neighbour — 2.37:1 vs 1.21:1 in light, 3.21:1 vs 1.70:1 in dark. In dark it even out-contrasts the *enabled* terminate glyph (3.21 vs 2.18). The hierarchy is inverted, which is exactly "the 1st and 2nd buttons are disabled but look identical to the enabled ones".

Which buttons the reporter saw: for a **PENDING** session `isActive()` is true but `isAppSupported()` is false (empty `service_ports`), so `appLauncher` (primary) and `terminal` (secondary) are disabled while `logs` and `terminate` stay enabled.

### The fix

`react/src/components/ComputeSessionNodeItems/SessionActionButtons.tsx` only. Each button's disabled predicate is hoisted to a named const and now drives **both** `isDisabled` **and** `variant`:

```tsx
const variantWhenEnabled = (
  isDisabled: boolean,
  variant: 'primary' | 'secondary',
): 'primary' | 'secondary' | 'ghost' => (isDisabled ? 'ghost' : variant);
```

`ghost` is `backgroundColor: 'transparent'` (`Button.tsx:224`), so a dead control loses its chip entirely and can no longer out-shout a live one. Enabled buttons keep their existing variant, unchanged.

Fix level: **Astryx component prop** — rung 2 of the `astryx-fix` §1 ladder, reached without a theme edit and without any CSS. **No opacity value is touched or introduced**; `astryx docs`/`component IconButton` confirms `variant` is the only lever the component exposes for this (there is no disabled-emphasis prop, and Astryx's own IconButton guidance is "use ghost in toolbars and dense areas to reduce visual clutter").

### Why not lower the opacity (what #8680 did)

Measured on the same harness: `0.5 → 0.25` puts the *disabled glyph* at ~1.7:1 (light) / ~2.2:1 (dark) against the band. That trades "disabled looks enabled" for "disabled is unreadable" — a worse defect, not a fix. It is also StyleX-owned, so a class-only CSS override loses the specificity fight (StyleX pads selectors with `:not(#\#)` = (1,1,0)) and needs `!important`.

### Why not a theme edit

Two reasons. (a) **Scope**: WS-1 owns `backendAiTheme.ts` / `antdParity.ts` this round, so the theme was off-limits here by construction. (b) **It would be the wrong shape anyway**: the genuinely theme-level fix is to stop `.astryx-banner.info` collapsing `--color-accent` onto `--color-text-blue`, which is a deliberate upstream `neutralTheme` decision with app-wide blast radius across *every* info Banner, and is the shared root of sibling #8662. That belongs in its own PR with its own probe. Separately, a `components: { button: { ':disabled': …opacity } }` override is discarded in production builds — `useCSSLayers: false` puts a second, **unlayered** copy of `styles.disabled` in the entry stylesheet, which outranks every named layer (the same trap already written up in `react/src/components/BAISider.css:180-232`).

## Measured before/after — **live**, both modes

**Probe: Storybook, not the running app.** The notification stack is a logged-in app surface and this box has no Backend.AI endpoint or credentials (`e2e/envs/.env.playwright` absent, no `config.toml`; the Playwright-MCP seed test fails at the login form). So the surface was reproduced in BUI Storybook — Astryx `Banner status="info"` + `ButtonGroup` + `IconButton`s in the exact enabled/disabled/variant combinations, over a `var(--color-background-body)` panel, in both `Theme mode="light"` and `mode="dark"`. This is a **real browser measurement of the real mechanism** (Storybook's `astryxBrandTheme` `extends: neutralTheme`, so the identical `.astryx-banner.info` collapse applies), not a static calculation — but it is **not** a probe of the assembled page. The temporary story was deleted before committing; `git status` shows only the one changed file.

All numbers are WCAG 2.x contrast of the **composited** colour (element background/glyph → `opacity` → banner band) read from `getComputedStyle` in headless Chromium. **0 `pageerror` in both passes.**

### Light — band `#c4ddfb`

| Button | | chip vs band | glyph vs band |
|---|---|---|---|
| `appLauncher` **disabled** | before | **`#6291c4` 2.37:1** | `#e2eefd` 1.18:1 |
| | after | **`#c4ddfb` 1.00:1** (no chip) | `#6291c4` 2.37:1 |
| `terminal` **disabled** | before | `#bcd3f0` 1.10:1 | `#6291c4` 2.37:1 |
| | after | `#c4ddfb` 1.00:1 | `#6291c4` 2.37:1 |
| `logs` enabled | unchanged | `#b4cae4` 1.21:1 | `#00458c` **6.76:1** |
| `terminate` enabled | unchanged | `#b4cae4` 1.21:1 | `#ff4d4f` 2.35:1 |
| `appLauncher` enabled | unchanged | `#00458c` **6.76:1** | `#ffffff` 1.39:1 |

### Dark — band `#383743` (`--color-background-blue` `#9eb7ff3D` over `--color-background-body` `#191919`)

| Button | | chip vs band | glyph vs band |
|---|---|---|---|
| `appLauncher` **disabled** | before | **`#8085a1` 3.21:1** | `#9c9ba1` 4.23:1 |
| | after | **`#383743` 1.00:1** (no chip) | `#8085a1` 3.21:1 |
| `terminal` **disabled** | before | `#4a4852` 1.30:1 | `#8085a1` 3.21:1 |
| | after | `#383743` 1.00:1 | `#8085a1` 3.21:1 |
| `logs` enabled | unchanged | `#5c5961` 1.70:1 | `#c7d3ff` **7.87:1** |
| `terminate` enabled | unchanged | `#5c5961` 1.70:1 | `#be3d3f` 2.18:1 |
| `appLauncher` enabled | unchanged | `#c7d3ff` **7.87:1** | `#ffffff` 11.64:1 |

**Read this way:** before, the loudest chip in the row after the enabled primary was a **disabled** button (2.37 light / 3.21 dark, against enabled secondaries at 1.21 / 1.70). After, disabled chips sit at exactly 1.00:1 — the band itself — so chip presence is now a clean binary: enabled buttons have one, disabled buttons do not. The enabled primary's 6.76:1 / 7.87:1 is untouched.

### Screenshot (probe harness, all four panels)

FR-3506 probe: light/dark × before/after

Top to bottom: light-before, light-after, dark-before, dark-after. Button order in each row: `app` (disabled), `terminal` (disabled), `logs`, `terminate`, `app` (enabled). In both *-before* rows the first button is a solid filled chip; in both *-after* rows it is gone.

## Verification

| Command | Result |
|---|---|
| `bash scripts/verify.sh` | `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup, StyleX sentinel, `astryx theme build --check` ✓ up to date, Terminology 0 blocking / 0 warn) |
| `pnpm --filter backend.ai-ui build` | `✓ built in 1m 7s` (run because a fresh worktree has no `dist`; **no** `packages/backend.ai-ui/src` file is changed by this PR) |
| `pnpm --prefix ./react exec vitest run` | **66 files passed, 1166 tests passed** |
| `pnpm --filter backend.ai-ui exec vitest run` | **41 passed / 1 skipped files, 583 passed / 1 skipped tests** |
| `node scripts/migration-gates/astryx-token-gate.mjs --strict` | 259 declared / 760 checked / **undeclared: 2** — `BAIText.css:27` and `BAIText.tsx:230`. **Pre-existing**: verified by `git stash` → identical output on the clean base → `git stash pop`. Neither is in this diff. |

No theme edit **in this first commit**: it left `THEME_NAME_REV` at **11**. The follow-up commit below is a theme edit and bumps it to **12** — see that section for its own verification run. No `index.html` / `resources/webui.css` mirror update is needed in either commit — no background or surface colour moved.

## Residue — what this PR does **not** close

- **The `terminal` button's cue is unchanged on the info band.** Its variant is `secondary`, whose chip is already near-invisible there (1.10:1 light / 1.30:1 dark), so dropping it to `ghost` moves ~nothing on that one surface — the only cue it has, before and after, is the glyph dim (6.76 → 2.37 light, 7.87 → 3.21 dark; a 2.9× / 2.5× contrast drop). It *does* gain a visible cue on the white-card surface (`SessionDetailContent`), where the secondary chip is real. Making an **enabled** secondary chip readable on a tinted band requires the theme, which this PR deliberately does not touch.
- **`.astryx-banner.info`'s token collapse stays open**, and it is the shared root of **#8662** ("content contrast lost against an info banner background"). **This PR does not subsume #8662.** Two further consequences of the same collapse are also left open and are worth their own report: an **enabled** primary button inside an info banner paints `--color-on-accent` `#ffffff` on a `#c7d3ff` chip in dark = **1.39:1 light / 11.64:1 dark** — the light figure is illegible even when enabled; and the **enabled** `terminate` glyph sits at 2.35:1 light / 2.18:1 dark, barely above a disabled glyph.
- **No "why is it disabled?" tooltip.** Every button already passes `tooltip`, so stating the reason ("Available once the session is running") is a one-line ternary each — but it needs a new i18n key across 22 locale files and the report's stated Expected is visual. Left as a follow-up rather than bundled.
- **No live probe of the assembled page** — stated above and not glossed: the numbers are from a Storybook reproduction of the surface, in a real browser, not from the notification stack in the running app. A reviewer with a cluster should confirm on a PENDING session in both modes.
- **`aria-disabled`, not `disabled`.** Because every button passes `tooltip`, `Button.tsx:627-629` keeps them focusable and sets `aria-disabled="true"` with the native `disabled` attribute left undefined. Unchanged by this PR, but any e2e assertion must target `[aria-disabled="true"]` — a `:disabled` selector matches nothing here.
- **Sibling workstreams untouched**: no overlap with WS-1 (#8669 / #8670 / #8674, which own the theme this round), WS-2, WS-4a or WS-6.

Drive-by, per `.claude/rules/comment-density.md` "touch it, trim it": the 30-line `toAstryxSize` migration essay in this file is cut to the 4 lines that still constrain the code (the measured Q-19 glyph sizes and the recorded 8px box residue). No other file is touched.

---

## Follow-up commit — `style(FR-3506): outline the secondary button at the theme level`

The design decision this PR's first commit recorded as residue ("antd's non-primary buttons were `colorBgContainer` + a 1px `colorBorder` outline … worth a separate decision") has now been made: **the default (`secondary`) Button variant renders as an outline app-wide, while `primary` stays a borderless brand fill.**

### Mechanism

Astryx's `secondary` variant is a borderless solid fill (`--color-neutral`, `Button.tsx:205-214`). The parity target is antd's default button: `colorBgContainer` background + a 1px `colorBorder` rule. Pinned in `backendAiTheme.ts`'s `components.button` block as a `'variant:secondary'` override:

```ts
'variant:secondary': {
  '--color-neutral': 'var(--color-background-surface)',   // fill → antd colorBgContainer
  border: '1px solid var(--color-border-emphasized)',     // antd colorBorder
},
```

The fill is remapped via the **token the StyleX rule consumes**, not `backgroundColor`: production's unlayered StyleX copy outranks `@layer astryx-theme` for any property with a counterpart in the component's own base (the `BAISider.css:180-232` mechanism — a `backgroundColor` override would hold in dev and silently revert to the grey fill in the built app), while a custom property has no competitor in either build. The theme components block is shared by every role theme built from the recipe, so the **admin (blue) and secondary (teal) role themes get the same outline** — confirmed live on `/admin/users`.

`IconButton` is a thin wrapper over `Button` (same `astryx-button` class), so secondary icon buttons — including the session action buttons this PR is about — get the same outline. `reset.css` sets `box-sizing: border-box` universally, so the border changes no button dimension. `THEME_NAME_REV` 11 → **14** (12 landed on `main` in between and was merged in; each recipe change re-bumped) and `bai-r14-default-brand-hc43dtt.*` is the regenerated artifact.

**ButtonGroup: the outline moves to the group.** Per-member outlines cannot compose inside a joined group — seams double (measured: every session-action member carried both a 1px left and right border, a 2px seam), and a member rendered `ghost` by the disabled mapping leaves its bordered neighbour half-open. `react/src/index.css` therefore gives a group that contains a secondary member a single ring — a generated `::after` overlay, because members' opaque fills cover an inset outline on the wrapper — and members drop their own border, keeping a 1px leading divider. Two divider harmonisations ride along: every member of a ringed group shares the emphasized divider (a ghost member otherwise kept Astryx's fainter default next to the secondaries', so seam weights alternated), and a **solid primary group's divider** (`Create User` + more-menu) stays on-accent — white-family in both modes like the label, but at 35% alpha (`color-mix` over `--color-on-accent`), because the opaque white cut the fill too hard and a theme-flipping grey read as a foreign border. A disabled member's wholesale `opacity: 0.5` also dimmed its own divider, so seam brightness alternated against enabled neighbours; inside a ringed group the member stays at full opacity and the disabled state reads from the glyph via `--color-text-disabled` (icons follow `currentColor`) — antd's own disabled treatment. All rules use `!important` because production's unlayered StyleX copy outranks every layer (the `BAISider.css:180-232` mechanism); horizontal groups only — no vertical `ButtonGroup` exists in the app. Re-measured live: ring `1px #D9D9D9`/`#424242` with `radius 8px`, member dividers uniform `1px` emphasized in both modes, primary group divider `#D9D9D9`/`#424242`.

### Measured before/after — live app, both modes

After-values measured on `/session`, `/data`, `/environment` in the running dev app (`getComputedStyle` in headless Chromium, **0 `pageerror` in both passes**); before-values are the recipe's token values the same probe read prior to the change.

| Mode | Secondary before | Secondary after (measured) | Primary (unchanged, measured) |
|---|---|---|---|
| Light | `rgba(0,0,0,0.06)` fill, no border | `#FFFFFF` + `1px solid #D9D9D9` | `#FF7A00` fill, no border |
| Dark | `#262626` fill, no border | `#141414` + `1px solid #424242` | `#BE5E06` fill, no border |

Admin role theme (`/admin/users`, blue accent): the standalone secondary refresh button measures the same `#FFFFFF` + `1px #D9D9D9`, and the primary `Create User` + more-menu group keeps its borderless blue fill with the group's own `--color-on-accent` divider. The sider collapse chip (`.bai-sider-toggle`, `BAISider.css`) keeps its own scoped hairline (`--color-border`) by design — unaffected.

| Light | Dark |
|---|---|
| secondary outline, light | secondary outline, dark |

| Session action group (single seams) | Admin role theme (`/admin/users`) |
|---|---|
| session action group seams | admin users secondary outline |

| Ringed group, dark (mixed primary + secondary) | Primary group divider — on-accent 35%, light | Primary group divider — on-accent 35%, dark |
|---|---|---|
| group ring dark | create user divider light | create user divider dark |

### Verification (re-run after both follow-up commits)

| Command | Result |
|---|---|
| `bash scripts/verify.sh` | `=== ALL PASS ===` (incl. `astryx theme build --check` against the r13 artifact) |
| `pnpm --prefix ./react exec vitest run` | **66 files, 1166 tests passed** (re-run after the seam commit: exit 0) |
| `pnpm --filter backend.ai-ui exec vitest run` | **41 passed / 1 skipped files, 583 / 1 tests** (no BUI source is touched by the follow-ups) |
| `node scripts/migration-gates/astryx-token-gate.mjs --strict` | undeclared: **2** — the same pre-existing `BAIText.css:27` / `BAIText.tsx:230` pair documented above; every token the follow-ups add (`--color-neutral` remap, `--color-background-surface`, `--color-border-emphasized`) is declared |

### Effect on this PR's residue

The first residue bullet ("making an **enabled** secondary chip readable on a tinted band requires the theme") is what this commit addresses: `.astryx-banner.info` remaps neither `--color-background-surface` nor `--color-border-emphasized`, so an enabled secondary button inside the info banner now paints a real surface chip with a border instead of the 1.10:1 / 1.30:1 near-invisible neutral fill. **Caveat**: the banner surface itself was not re-measured live (no PENDING session was available during the probe) — the numbers above are from the table/toolbar surfaces; a reviewer with a cluster should still eyeball the notification stack in both modes. The `.astryx-banner.info` token collapse (#8662) itself remains open and is still not subsumed.


agatha197 (approved):
LGTM

PR #8698, opened 2026-08-11
fix(FR-3502): invert header band content with the app mode

### LINKED ISSUE #8669: Header band: button hover paints an opaque black block in light mode

### LINKED ISSUE #8670: Header band: content stays white in dark mode instead of inverting against the app mode

### LINKED ISSUE #8674: Header project selector: option hover paints a dark background in light mode, label unreadable

yomybaby:
Resolves #8670 (FR-3502)
Resolves #8669 (FR-3501)
Resolves #8674 (FR-3505)

Three reports on one surface, one theme edit and one `THEME_NAME_REV` bump — so they ship together rather than as three PRs racing for the same built artifact.

### FR-3502 — band content is pinned white instead of inverting

The brand band is a **reversed** surface, but every expression of that was a constant, not a `[light, dark]` pair:

| where | was |
|---|---|
| `WebUIHeader.tsx` ×2 | `<MediaTheme mode="dark">` — a literal |
| `WebUIHeader.tsx` band root | `color: 'var(--color-on-dark)'` |
| `UserDropdownMenu.tsx` trigger | `'data-astryx-media': 'dark'` |
| `BAINotificationButton.tsx` | `data-astryx-media="dark"` + `color: var(--color-on-dark)` ×2 |
| `BAISelect.css` `.bai-select-ghost` | `color` / `border-color: var(--color-on-dark)` |

`MediaTheme`'s whole mechanism is `<div data-astryx-media={mode} style="display:contents">`, and the theme keys `[data-astryx-media="dark"] { --color-text-primary: var(--color-on-dark); … }` off it. `--color-on-dark` is `#ffffff` in **both** schemes (`backendai-default-built.css:182`), so dark mode painted white on `#E88A28`. All of these now derive from `isDarkMode`.

The chevron was a fourth, different case: `Selector.tsx:148` puts `color: var(--color-icon-secondary)` on the indicator span and the `Icon` takes `color="inherit"`, so it already varied with the app mode — in the **wrong** direction (45% black on the orange in light). Astryx declares that in StyleX, which a class selector cannot outrank (`:not(#\#)` = specificity 1,1,0 — see `BAISider.css`), so `BAISelect.css` re-declares the **token** rather than re-stating `color`.

### FR-3501 — the light-mode hover wash was opaque

`ANTD_REVERSED_BAND_OVERLAYS.light['--color-overlay-hover']` was the literal `'#262626'` — fully opaque near-black, inherited from antd's `colorFillSecondary` dark seed. Astryx paints hover as `background-image: linear-gradient(var(--color-overlay-hover), var(--color-overlay-hover))` **over** the existing background, so an opaque value replaces the brand orange outright. The other three cells were already translucent and are untouched.

`rgba(255,255,255,0.16)` is not a new number: it is the exact wash `ANTD_HOVER_PARITY` already applies to `button.variant:primary` / `variant:destructive` (`backendAiTheme.ts:756,759`), one step under the existing pressed `0.18`. **This is the one taste call in the PR** — `rgb(255 255 255 / 10%)` (what `BAISelect.css` used for the ghost trigger) is the softer alternative if the designer prefers it.

### FR-3505 — the selector popup inherits the band, and #8678 does not reach it

Not closed by #8678, and its pattern is **not copyable one-for-one**:

- `BAISelect` renders `Selector` / `MultiSelector`, never `DropdownMenu`, so the `.astryx-dropdown-menu` pin cannot match. There is **zero** `astryx-selector` in the built CSS.
- `themeProps('dropdown-menu')` sits **on the panel** (`DropdownMenu.tsx:477`); `themeProps('selector')` sits on the **trigger** (`Selector.tsx:1162`) and the panel carries **no theme class at all**. `defineTheme` emits `.astryx-<name>` with pseudo-classes but never descendant combinators, so `components: { selector: … }` would reach the trigger only.
- The panel is a genuine DOM descendant: `usePopover` → `useLayer` renders a native `popover="auto"` element **inline** and promotes it with `showPopover()`. Top layer does not change DOM ancestry, and there is no `createPortal` anywhere in `Layer/` or `Popover/`. So it inherits `--color-overlay-hover` straight off the band.

`Field` — the div carrying `themeProps('field')` that wraps trigger **and** popup — is the only ancestor both share, and is where the neutral page pair is restored. It reuses `MENU_PANEL_PAGE_OVERLAYS` verbatim, which already reads both values out of `ANTD_NEUTRAL_SURFACES`, so **no literal is retyped** (the anti-drift rule #8678 established).

Everywhere except the band that declaration is a no-op — it restates the value the field already inherits from `:root`. Inside the band it would also neutralise the ghost trigger's own wash, which *is* band chrome, so `WebUIHeader` spreads `ANTD_REVERSED_BAND_OVERLAYS` onto `<ProjectSelect style>`; that lands on the trigger div (`Selector.tsx:1186`) as an **inline** declaration, which beats the class rule there and only there. `BAISelect.css`'s hand-rolled `background-color` hover is therefore deleted — the ghost trigger now hovers through the same token as the `IconButton`s beside it.

FR-3501 does **not** incidentally close this: with the band wash changed to translucent white, the symptom mutates from "unreadable row" to "no visible highlight".

## Measured before/after

Measured live in BUI Storybook (`select-baiselect--ghost-style`) with headless Chromium, reading `getComputedStyle`, `color-scheme` flipped on the `[data-astryx-theme]` root. **0 `pageerror` in every pass.**

BUI's Storybook mirrors the brand theme only partially — it carries no `components.field` (nor #8678's `dropdown-menu`) — so the generated `.astryx-field` rule was injected **verbatim from `backendai-default-built.css`** into `@layer astryx-theme`, and the band root was simulated by setting the JS-indexed pair on the Field's parent.

### Band content colour (measured)

| | light | dark |
|---|---|---|
| ghost trigger label — before | `rgb(255,255,255)` | `rgb(255,255,255)` |
| ghost trigger label — **after** | `rgb(255,255,255)` | **`rgb(23,23,23)`** |
| trigger border — before / after | `rgb(255,255,255)` / same | `rgb(255,255,255)` / **`rgb(23,23,23)`** |
| chevron — before | `rgba(0,0,0,0.45)` | `rgba(255,255,255,0.45)` |
| chevron — **after** | **`rgb(255,255,255)`** | **`rgb(23,23,23)`** |

Contrast against the band (`#FF9729` light / `#E88A28` dark, `resources/theme.json:26,61`), WCAG 2.x:

| | before | after |
|---|---|---|
| band text, dark | white 2.59:1 | **`#171717` 6.91:1** |
| band text, light | white 2.16:1 | white 2.16:1 *(unchanged — see Residue)* |
| chevron, light | `#8c5317` 2.89:1 | white 2.16:1 *(now matches the label instead of the page)* |
| chevron, dark | `#f2bf89` **1.55:1** | **`#171717` 6.91:1** |

### Selector option-row overlay (measured)

| app scheme | before | after |
|---|---|---|
| light | `rgba(255,255,255,0.16)` — the band wash | `light-dark(rgba(0,0,0,0.06), #262626)` → `rgba(0,0,0,0.06)` |
| dark | `rgba(0,0,0,0.06)` — the band wash | `light-dark(…)` → `#262626` |

Trigger keeps the band pair in both (`rgba(255,255,255,0.16)` / `rgba(0,0,0,0.06)`) — the inline override wins over `.astryx-field`, confirmed live.

Composited row, and the label on it:

| | before | after |
|---|---|---|
| light row on `#FFFFFF` | `#ffffff` — no visible highlight | `#f0f0f0`, label 16.17:1 |
| dark row on `#141414` | `#131313` — no visible highlight | `#262626`, label 15.13:1 |
| light row *as reported* (pre-FR-3501, `#262626`) | label **1.22:1** | — |

### Hover wash on the band, light mode (FR-3501)

`#262626` opaque over `#FF9729` → `#262626`, brand orange destroyed. After: 16% white → `#ffa84b`, orange preserved and lifted.

## Verification

```
bash scripts/verify.sh
  Relay PASS · Lint PASS · Format PASS · TypeScript PASS · Vite warmup PASS
  StyleX cssInjectionTarget PASS · Astryx theme build PASS · Terminology PASS
  === ALL PASS ===

pnpm --filter backend.ai-ui build      ✓ built in 33.55s   (run BEFORE verify.sh — its
                                        `astryx theme build --check` reads the built dist)

pnpm --prefix ./react exec vitest run   Test Files 66 passed (66) · Tests 1168 passed (1168)
pnpm --filter backend.ai-ui exec vitest run
                                        Test Files 41 passed | 1 skipped (42)
                                        Tests 583 passed | 1 skipped (584)

  declared 259 | checked 765 | undeclared: 2
    packages/backend.ai-ui/src/components/BAIText.css:27   var(--x)
    packages/backend.ai-ui/src/components/BAIText.tsx:230  var(--font-family-mono)
```

Both gate findings are **pre-existing on `main`** and neither file appears in this diff (`git status` lists no `BAIText.*`). Same two #8678 reported.

### Theme procedure (all five steps)

1. Recipe: `FIELD_PAGE_OVERLAYS` added to `backendAiTheme.ts`.
2. `THEME_NAME_REV` **11 → 12**, and `FIELD_PAGE_OVERLAYS` added to the `computeThemeName` seed-hash array.
3. Nothing non-colour was tupled — the new entry is two `light-dark()` strings built from `ANTD_NEUTRAL_SURFACES`.
4. `astryx theme build` regenerated; `built/index.ts` repointed to `bai-r12-default-brand-hc43dtt`; the three `bai-r11-default-brand-h165tsmd.*` artifacts deleted.
5. **KEEP-IN-SYNC mirrors not needed** — the built-CSS diff is exactly the theme name and the new `.astryx-field` rule; no background or surface colour moved, so `index.html`'s critical `<style>` and `resources/webui.css`'s `.splash` are untouched. Stated explicitly rather than left ambiguous.

A drift guard mirroring #8678's block was added to `backendAiTheme.test.ts`: without it, deleting the spread only changes the theme name and every other check goes green after a regenerate.

## Residue

- **No live app probe.** There is no Backend.AI endpoint or credentials on this box (`e2e/envs/.env.playwright` absent, no `config.toml`), and the header is a logged-in surface. Everything above was measured in BUI Storybook against the real `BAISelect`, plus statically from `backendai-default-built.css` and the shipped `@astryxdesign/core` sources. **The header band itself was never rendered.**
- **The `.astryx-field` rule could not be probed as the theme emits it** — BUI's Storybook theme mirror is partial. It was injected verbatim from the generated CSS. `WebUIHeader`, `UserDropdownMenu` and `BAINotificationButton` live in `react/src`, which Storybook does not cover at all; their changes are derivations of the same `MediaTheme` mechanism the Storybook probe confirms, not separately measured.
- **Light-mode band text stays at 2.16:1.** That is the product owner's stated intent (white on the orange in light mode), not a regression this PR introduces. It is below WCAG AA and worth its own decision.
- **`rgba(255,255,255,0.16)` is a taste call**, not a measurement. Flagged above.
- **`.astryx-field` is app-wide.** Every Astryx `Field` now re-declares the overlay pair. In all of them the declared value **equals** what they already inherit from `:root`, so the only behavioural delta is inside a band / `[data-astryx-media]` context. Not separately screenshotted across the app.
- **Does not subsume #8662** ("content contrast lost against an info banner background") — different mechanism (`.astryx-banner.info` collapsing `--color-accent` onto `--color-text-blue`), different surface.
- **Does not subsume the sibling `MultiSelector` case beyond the token.** `MultiSelector` uses the same `Field` and the same `--color-overlay-hover`, so it inherits the fix, but no multi-select sits on the band today and none was probed.
- `BAINotificationButton`'s tooltip content keeps `<MediaTheme mode="dark">` **constant** on purpose: `ANTD_HOVER_PARITY.tooltip` pins the bubble to `light-dark(rgba(0,0,0,0.85), #424242)` — dark in both schemes. Deriving it would produce white-on-white in dark, the exact defect that file's comment records. Only the trigger is band chrome.
- The badge-wrapper interception noted in `BAINotificationButton.tsx` was **not** re-verified live (no app probe); the glyph now reads `var(--color-icon-primary)`, which `MediaTheme` remaps on the button element itself, so the wrapper's own `color` no longer needs to be out-specified.

Per `comment-density.md`, the long migration-era blocks in every file this PR touches were trimmed to their load-bearing constraint (−243 lines, +139).


yomybaby:
## Live app probe — run after this PR was opened

The PR body says a live probe was impossible on the authoring box. A Backend.AI endpoint
became available afterwards, so this branch was probed **in the running app**, not in
Storybook: manager **26.7.0** at an internal webserver, logged in as a superadmin,
Chromium 1600×1000, dark entered through the header's own toggle with
`document.documentElement.dataset.theme` asserted `'dark'`.

**`data-astryx-theme` served = `bai-r12-default-brand-hc43dtt`**, so the regenerated
artifact is the one actually painting — not a stale `r11`. `bai-r11-*` is gone from the
tree and `built/index.ts` re-points to `r12`. **0 `pageerror` in every pass, both modes.**

Baseline ("before") was captured on `main` @ `e40fd8f47` with the identical script.

### FR-3502 — band content now inverts with the app mode

| element | before light | after light | before **dark** | after **dark** |
|---|---|---|---|---|
| `Project` label | `rgb(255,255,255)` | `rgb(255,255,255)` | `rgb(255,255,255)` ❌ | **`rgb(23,23,23)`** ✅ |
| project name | `rgb(255,255,255)` | `rgb(255,255,255)` | `rgb(255,255,255)` ❌ | **`rgb(23,23,23)`** ✅ |
| user name | `rgb(255,255,255)` | `rgb(255,255,255)` | `rgb(255,255,255)` ❌ | **`rgb(23,23,23)`** ✅ |
| bell / theme / help / user icons | `rgb(255,255,255)` | `rgb(255,255,255)` | `rgb(255,255,255)` ❌ | **`rgb(23,23,23)`** ✅ |
| band background | `rgb(255,151,41)` | unchanged | `rgb(232,138,40)` | unchanged |

`rgb(23,23,23)` is `--color-on-light` (`#171717`). The band background always tracked the
app mode; only the content did not. Dark-mode contrast of `#171717` on `#E88A28` ≈ **7.1:1**.

Bonus, unlisted in the issue: the selector chevron measured `rgba(0,0,0,0.45)` on the
orange band in light mode before, and is `rgb(255,255,255)` after.

### FR-3501 — hover overlay, measured from real screenshots

`getComputedStyle` on the button is **not** a valid measurement here — Astryx paints the
hover pill as an overlay, so the button's own `background-color` reads `rgba(0,0,0,0)`
hovered or not. These are pixels sampled from screenshots of each button's box.

| mode | band at rest | button hovered — before | after |
|---|---|---|---|
| light | `#ff9729` | **`#262626`** (opaque, orange fully occluded) ❌ | **`#ffa84b`** ✅ |
| dark | `#e88a28` | `#db8226` | `#db8226` (unchanged, was already correct) |

The resolved token behind it, read on the button itself:

| | before | after |
|---|---|---|
| app light → button `color-scheme` | `dark` | `dark` |
| app **dark** → button `color-scheme` | `dark` ❌ | **`light`** ✅ |
| app light → `--color-overlay-hover` | `#262626` | **`rgba(255,255,255,0.16)`** |
| app dark → `--color-overlay-hover` | `rgba(0,0,0,0.06)` | `rgba(0,0,0,0.06)` |

Arithmetic check on the chosen wash: `#ff9729` composited with white at 16% gives
R `255`, G `151 + 104×0.16 = 168` (`a8`), B `41 + 214×0.16 = 75` (`4b`) → `#ffa84b`.
**The sampled pixel equals the token arithmetic exactly**, which is what confirms the
overlay is translucent rather than merely a different opaque colour.

### FR-3505 — reproduced on `main`, closed here

This settles the open question from the handoff (whether #8678 had already closed it).
It had not. Project-selector dropdown, hovering a non-selected option:

| | option hover background | label colour | contrast |
|---|---|---|---|
| **before, light** | **`#262626`** | `rgb(20,20,20)` | **≈1.2:1 — unreadable** ❌ |
| after, light | `#f0f0f0` | `rgb(20,20,20)` | ≈15.5:1 ✅ |
| before, dark | `#1d1d1d` | `rgb(255,255,255)` | fine |
| after, dark | `#262626` | `rgb(255,255,255)` | fine |

The listbox panel's `color-scheme` now follows the page (`light`→`light`, `dark`→`dark`),
which is what `FIELD_PAGE_OVERLAYS` was for.

### Residue on the probe itself

- Light-mode band contrast is **not** changed by this PR and is worth a separate look:
  white `#FFFFFF` on `#FF9729` measures ≈**2.2:1**. That is pre-existing on `main` and
  is the *requested* design (white content on the orange band in light mode), so it is
  recorded here rather than changed.
- The screenshot-pixel method samples a point at 15% inset of each button's box to stay
  off the glyph; it is a fill measurement, not a per-pixel diff.

agatha197 (changes requested):
Please resolve conflicts

yomybaby:
@agatha197 Conflicts resolved and pushed (`6256cc903`) — ready for another look.

**What conflicted:** exactly one file, `react/src/components/BAINotificationButton.tsx`. #8701 (FR-3509) replaced the shortcut badge with Astryx `Kbd` in the same hunk this PR touches to invert the trigger's media context, so the JSX auto-merged cleanly and only the comment block plus the `bandMediaMode` declaration collided.

**How it was resolved:** kept `const bandMediaMode = isDarkMode ? 'light' : 'dark'` (this PR's functional change — the body already calls it via `data-astryx-media`), and merged the two comment versions into one that keeps every constraint both sides recorded:

- the media context must stay **off** `Tooltip` and the drawer — Astryx renders them as inline siblings, not portals, so a wrapper reaches their panels;
- it therefore sits on the trigger **button** via `data-astryx-media`;
- the tooltip *content*'s `mode="dark"` is **constant**, not the app's opposite, because `ANTD_HOVER_PARITY` pins the bubble to `colorBgSpotlight` (dark in both schemes).

Nothing else needed touching: `main` moved in 3 other files this branch also has (`BAIText*`, locales, `WebUISider`), but none overlapped. Theme artifacts are unchanged — still `bai-r12-default-brand-hc43dtt` only, `THEME_NAME_REV = 12`.

**Re-verified after the merge:**

- `pnpm --prefix ./react exec vitest run` → 66 files / **1168** tests pass
- `pnpm --filter backend.ai-ui exec vitest run` → 42 files / **590** tests pass
- `astryx-token-gate.mjs --strict` → 259 declared / 738 checked / **2 undeclared** (the pre-existing `BAIText.css:27` + `BAIText.tsx:230`)
- CI on the merge commit: green, 0 failures

Auto-merge (squash) is armed, so this lands as soon as the review clears.

agatha197 (approved):
LGTM

PR #8711, opened 2026-08-12
fix(FR-3520): surface review-step validation errors on the failing cards and stepper

### LINKED ISSUE #8710: Deployment preset Review step: validation failures show only a bare 1px border, and the stepper marks failing steps as completed
On the deployment preset wizard's Review step, a validation failure is barely indicated: the only cue is a bare 1px red outline on the failing cards. There is no icon, no message, and no list of which fields failed. The stepper on the right shows blue completion checkmarks for the very steps that hold the failing fields.

Reproduce

1. Admin > Deployments > Create Preset
1. Go straight to the review step with an empty form: /admin/deployments/deployment-presets/new?step=review

Observed

- Basic Info / Resources / Deployment cards get only border-color: rgb(255, 77, 79).
- The stepper shows "Basic Info" and "Model & Execution" as completed (blue check), not error.

Expected

- The failing cards carry the design system's error treatment.
- The stepper marks the steps that contain failing fields as errored.

Mechanism (measured)

1. AdminDeploymentPresetReviewSummary passes className="bai-card-error" plus an inline style=` borderColor: token.colorError ` instead of BAICard's status="error". So the cards never get the bai-card-~~error class that BAICard.css styles with var(~~-color-error), and never get the error icon BAICard puts on the action slot. The hardcoded rgb(255,77,79) is the antd parity fixture red, so it does not follow dark mode.
1. .bai-card-error carries zero CSS rules. It is only a tour anchor; BAICard emits it as an OUTPUT of status="error". The call site passes it as an input expecting it to paint.
1. stepErrors is computed with form.getFieldError(...), which returns empty for fields whose step is not currently mounted, so the per-step status="error" never fires.

Note: AdminDeploymentPresetValidationTour does spotlight the error card, but it is gated on the has_opened_tour_deployment_preset_validation user setting and fires once per user, ever.

JIRA Issue: FR-3520

agatha197:
Resolves #8710 (FR-3520)

## Summary

On the deployment preset **Review** step, a failed validation was almost invisible: the only cue was a bare 1px red border on the failing cards, and the stepper showed a **completion checkmark** next to the very step holding the failing fields.

## Mechanism

Two separate causes, both named below rather than guessed at.

**1. The stepper — `getFieldError` only sees mounted fields.**
`stepHasError` was `fields.some((f) => form.getFieldError(f).length > 0)`. `getFieldError` reports only fields that are currently **mounted**, so every step except the one on screen measured as clean and rendered `status="completed"`. It now also consults `errorFieldNames`, which is populated from the full `validateFields()` sweep and covers every step — the same source `AdminDeploymentPresetReviewSummary` already matched its per-card flags against, so no new plumbing.

**2. The cards hand-rolled the tint instead of using the design system.**
They passed `className="bai-card-error"` + inline `style={{ borderColor: token.colorError }}`. `.bai-card-error` has **zero CSS rules** — `BAICard` *emits* it as an OUTPUT of `status="error"` for the two validation tours to anchor on — so only the inline style painted anything. They now pass `status="error"` and pick up `bai-card--error`.

**3. The error glyph never rendered.** `BAICard` draws its error icon only in the `extraButtonTitle` branch; these cards supply a custom `extra` (the Edit link), so that branch is never taken. The action slot now renders a `CircleX` through `BAICard`'s existing `.bai-card-extra--error` hook (`color: var(--color-error)`), so the icon is themed, not hardcoded.

### Correction — this is not a colour fix

The issue as filed claimed the hardcoded `rgb(255,77,79)` did not follow dark mode. **That was wrong**, and I measured it on `main` before writing this: `theme.useToken()`'s `colorError` carries `[light, dark]` pairs, so the pre-fix inline style already resolved to `rgb(255,77,79)` light / `rgb(190,61,63)` dark. The border colour is **identical before and after**. Switching to `status="error"` is an API/consistency change (per `use-bai-card.md` rule 4), not a visual one. The user-visible wins are the stepper and the new glyph. FR-3520 carries a comment with the same correction.

## Measured, in the real app (light + dark)

Repro: `/admin/deployments/deployment-presets/new?step=review` with an empty form.

| | Before | After |
|---|---|---|
| Card classes | `bai-card--compact, bai-card-error` | `bai-card--compact, **bai-card--error**, bai-card-error` |
| Card inline style | `border-color: rgb(255,77,79)` | *(none)* |
| Computed border — light | `rgb(255, 77, 79)` | `rgb(255, 77, 79)` *(unchanged, by design)* |
| Computed border — dark | `rgb(190, 61, 63)` | `rgb(190, 61, 63)` *(unchanged, by design)* |
| Error glyph in action slot | absent | 3 × `rgb(255, 77, 79)` (`var(--color-error)`) |
| Stepper step 1 (holds the failing fields) | `Go to step 1: Basic Info, **completed**` | `Go to step 1: Basic Info, **error**` |
| Stepper step 2 — empty form (no failing fields) | `…, completed` | `…, completed` *(correct — genuinely clean)* |
| Stepper step 2 — **env var row with an empty value** | `…, **completed**` | `…, **error**` |
| Error cards — env var case | 3 | 4 (Model & Execution now included) |
| Stepper step 3 | `…, error` | `…, error` |
| `pageerror` count | — | 0 (both modes) |

Dark mode was entered through the header toggle with `document.documentElement.dataset.theme === 'dark'` asserted.

## Screenshots

After
 after light 
 after dark 

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup, StyleX, Astryx theme build, Terminology)
- `pnpm exec vitest run` in `react/` → **66 files, 1166 tests passed**
- `pnpm exec vitest run` in `packages/backend.ai-ui/` → **41 passed / 1 skipped, 583 tests passed / 1 skipped**
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → **fails with `undeclared: 2` — pre-existing.** Verified by `git stash`-ing this diff and re-running on the clean tree: identical exit code and identical `undeclared: 2`. Neither undeclared token is in a file this PR touches.
- Live probe in both modes, 0 `pageerror`.

### Field-list coverage

The step/card error lists are hand-maintained, and the first cut missed the two `Form.List`s whose **rows** carry the required rules — `environ` (Environment Variables) and `resourceOpts` (Resource Options). Both report row errors under the list's own name, so an empty env var value failed validation while the Model & Execution card and its stepper entry still rendered complete. Both are now listed.

Audited every `name=` on the form against the lists; the remaining uncovered fields (`description`, `openToPublic`, `revisionHistoryLimit`, `resourceSlots`) carry no `rules`, so they cannot produce an error.

**No manager-version branching is needed.** The two `@since` directives on this form (`presetValues`, `healthCheck.enable`) hydrate initial values from an existing preset — they do not gate whether a field renders or validates, and the top-level field names are static JSX. `errorFieldNames` only ever contains fields that actually mounted and validated, so a list entry for a field an older manager does not render is inert; the lists are safe as supersets. Nested paths collapse to their first segment, so `modelDefinition.*` (which contains the version-gated `enable`) is already covered by the `modelDefinition` entry.

## Residue — deliberately not closed here

- **No list of *which* fields failed.** The cards say *that* a section failed, not which input. That was raised as an option and deferred as a separate design decision.
- **`AdminDeploymentPresetValidationTour` is untouched.** It spotlights the error card, but is gated on the `has_opened_tour_deployment_preset_validation` user setting and fires **once per user, ever** — so for anyone past their first run it contributes nothing. Whether that gate is right is a separate question.
- **`SessionLauncherErrorTourProps` uses the same `.bai-card-error` anchor** and is **not** covered by this PR; I did not audit whether the session launcher has the same card/stepper symptom.
- `.bai-card-error` is left emitted by `BAICard` — both tours query it, so removing it would be a silent contract change.

**Checklist:**

- [ ] Documentation

### DIFF
--- a/react/src/components/AdminDeploymentPresetReviewSummary.tsx
+++ b/react/src/components/AdminDeploymentPresetReviewSummary.tsx
@@ -7,7 +7,6 @@ import { resolvesReadsVfolderConfigFiles } from '../helper/modelServiceCommand';
 import { useSuspendedBackendaiClient } from '../hooks';
 import { useAdminImageReference } from '../hooks/hooksUsingRelay';
 import { ResourceNumbersOfSession } from '../pages/SessionLauncherPage';
-import { theme } from '../theme-shim';
 import type {
   AdminDeploymentPresetFormValue,
   ModelServiceFormValue,
@@ -43,12 +42,21 @@ import { useTranslation } from 'react-i18next';
 // and is DROPPED everywhere below — the Astryx default density is the design.
 
 const BASIC_INFO_FIELDS = ['name', 'runtimeVariantId', 'imageId'] as const;
-const RESOURCES_FIELDS = ['cpu', 'mem', 'clusterMode', 'clusterSize'] as const;
+const RESOURCES_FIELDS = [
+  'cpu',
+  'mem',
+  'clusterMode',
+  'clusterSize',
+  'resourceOpts',
+] as const;
 const DEPLOYMENT_FIELDS = ['replicaCount'] as const;
+// `environ` and `resourceOpts` are Form.Lists whose rows carry their own
+// required rules, so they report errors under the list's own name.
 const STEP2_FIELDS = [
   'startupCommand',
   'bootstrapScript',
   'modelDefinition',
+  'environ',
 ] as const;
 
 interface PresetReviewSummaryProps {
@@ -80,7 +88,6 @@ const PresetReviewSummary: React.FC<PresetReviewSummaryProps> = ({
 }) => {
   'use memo';
   const { t } = useTranslation();
-  const { token } = theme.useToken();
   const baiClient = useSuspendedBackendaiClient();
   /**
    * FR-3481: mirrors the input form's placement of Service
--- a/react/src/components/AdminDeploymentPresetReviewSummary.tsx
+++ b/react/src/components/AdminDeploymentPresetReviewSummary.tsx
@@ -30,6 +30,7 @@ import {
   toLocalId,
 } from 'backend.ai-ui';
 import * as _ from 'lodash-es';
+import { CircleX } from 'lucide-react';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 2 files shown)


ironAiken2 (approved):
LGTM

PR #8714, opened 2026-08-12
fix(FR-3521): remove the portless atomic-write patch that breaks the daemon's routes.json watch

### LINKED ISSUE #8713: Remove the portless atomic-write patch: rename-based saves break the daemon's routes.json watch
The FR-3443 patch (react/patches/portless@0.15.5.patch, PR #8535) changed routes.json saves from an in-place writeFileSync to a tmp-file + renameSync atomic replace. The rename swaps the file's inode on every save, and the portless daemon watches routes.json with fs.watch, which stays attached to the old inode — so after the first save the daemon silently stops seeing route changes. New dev-server registrations then land in routes.json but never reach the running daemon, which serves its own 404 for them until it is restarted. Symptom: a dev server is up and registered, but its https://<name>.localhost:1355 URL returns a portless 404 whenever an older daemon (started by another checkout/session) owns port 1355.

Decision: remove the patch (revert to upstream portless@0.15.5 behaviour) — the original corruption concern from FR-3443 is accepted as the lesser problem compared to routes silently not propagating between sessions.

—

JIRA Issue: FR-3521

nowgnuesLee:
Resolves #8713 (FR-3521)

## What

Removes the FR-3443 portless patch (`react/patches/portless@0.15.5.patch`, introduced in #8535), reverting to upstream `portless@0.15.5` behaviour:

- deleted `react/patches/portless@0.15.5.patch`
- removed the `patchedDependencies` entry in `pnpm-workspace.yaml`
- regenerated `pnpm-lock.yaml` via `pnpm install` (patch hash dropped)

## Why

The patch changed routes.json saves from an in-place `writeFileSync` to a tmp-file + `renameSync` atomic replace. The rename swaps the file's inode on every save, but the portless daemon watches routes.json with `fs.watch`, which stays attached to the old inode — after the first save the daemon silently stops seeing route changes. New dev-server registrations land in routes.json but never reach the running daemon, which serves its own 404 for them until restarted.

Decision (FR-3521): the original corruption concern from FR-3443 is accepted as the lesser problem compared to routes silently not propagating between sessions.

## Verification

- Installed `node_modules/portless/dist/chunk-SLEZT6EJ.js` no longer contains the patched code (`PORTLESS_CORRUPT_ROUTES` marker: 0 occurrences) — upstream behaviour restored.
- No remaining references to the patch anywhere in the repo (only `pnpm-workspace.yaml` referenced it).
- `scripts/verify.sh`: Relay/Lint/Format/Terminology passed locally; the run was cut short before the final all-pass line at the operator's request (the diff contains no source code — patch removal + lockfile regen only), CI will run the full harness.

https://claude.ai/code/session_011RQ7wiYeAS19jvz4JM2Ufw


agatha197 (approved):
LGTM

PR #8724, opened 2026-08-12
style(FR-3529): give the Runtime Parameters category tabs room and a rail

### LINKED ISSUE #8723: Deployment Add Revision (Advanced Mode): Runtime Parameters category tabs sit flush against the alert with no bottom border
## Where

- **Page**: Deployments → (a deployment) → **Add Revision** → **Advanced Mode**

(`/deployments/:deploymentId`, modal). The runtime hint above the section

(`modelService.RuntimeVariantDefaultCommandAppliedNote`) is rendered only by

this modal, which is what pins the screenshot to it.

- **Region**: the category sub-tab strip inside the collapsible \*\*Runtime

Parameters\*\* section — the row reading `E2e Category` / `General`, sitting

between the `Unchanged parameters will use the runtime's default values.`

alert and the first parameter field.

- **Component** (if identified): `RuntimeParameterFormSection`

(`react/src/components/RuntimeParameterFormSection.tsx`), which renders an

Astryx `TabList size="sm"` directly after the alert.

- **Also reachable on**: Admin → Deployments → Deployment Presets → New/Edit

(`/admin/deployments/deployment-presets/new`), which mounts the same section.

Not separately reproduced — listed so the fix is not scoped to the modal only.

## Observed

The sub-tab strip is packed tightly into the space between the alert and the

first field: no vertical gap above it, no vertical gap below it, and no bottom

border under the tab row. The tabs therefore do not read as a tab bar — they

look like stray inline text wedged under the alert, and the whole section reads

as cramped.

## Expected

- Vertical margin above and below the tab strip, so it is visually separated

from the alert above and the first parameter field below.

- A bottom border under the tab row, so the strip reads as a tab bar and the

active tab's underline sits on a baseline.

## Environment

- **Mode**: light (dark not checked)
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: local dev server, WebUI 26.8.0-alpha.0 (build 8450)

## Reproduction

1. Open Deployments → any deployment → **Add Revision**.
1. Switch to **Advanced Mode**.
1. Pick a non-`custom` runtime (e.g. `cmd`) so the Runtime Parameters section

appears, and expand it.

1. Look at the category tab row directly under the alert.

Note: the tab strip only shows more than one tab when the runtime's presets

carry more than one `category`; the screenshot's `E2e Category` / `General`

tabs come from presets created with FR-3476's category field.

## Evidence

- Screenshot (reporter): `/tmp/clipimg/clip-20260812-144651.png` — to be

attached in Jira; the path is a temp clipboard dump and will not survive.

## Notes

Same missing-separation family as FR-3496 (MetadataList divider) and FR-3498

(in-card section separation), both of which came out of the antd → Astryx

conversion. Not verified to share a cause — noted only so triage can batch them

if it turns out they do.

JIRA Issue: FR-3529

agatha197:
Resolves #8723 (FR-3529)

> Stacked on #8717 (FR-3523). That PR is what makes the Runtime Variant select
> in this modal populate at all — without it you cannot pick a non-`custom`
> runtime, and the Runtime Parameters section this PR touches never renders.

## Symptom

In **Deployments → (a deployment) → Add Revision → Advanced Mode**, with a
non-`custom` runtime selected, the **Runtime Parameters** section's category tab
strip (`E2e Category` / `General`) is wedged between the warning banner above it
and the first parameter field below it: no vertical gap on either side, and no
rail under the tabs. It does not read as a tab bar.

## Mechanism

`RuntimeParameterFormSection` renders the Astryx `TabList` primitive directly.
Astryx's `hasDivider` defaults to `false`, so no rail is drawn — the BUI wrapper
`BAITabList` defaults it to `true`, but this call site does not go through the
wrapper. The strip is also a bare sibling of the `Banner` and the panel `div`s
inside `Collapsible`, which lay their children out with no gap container, so
there is nothing to supply vertical spacing either.

Both knobs are first-class Astryx props, so the fix stays at the component-prop
level — no CSS file, no wrapper swap:

- `hasDivider` draws the rail;
- `xstyle` carries a `marginBlock` pair built from `spacingVars['--spacing-4']`
  (tokens only; `stylex.create`, per the repo's existing `xstyle` idiom in
  `WebUISider.tsx`).

The nav already spans the full content width (672px measured), so `BAITabList`'s
block-level-nav rule was not needed here — the rail spans the bar as-is.

## Measured

Computed style on the `nav.astryx-tab-list`, live in the Add Revision modal with
runtime `cmd` selected:

| | before (light) | after (light) | before (dark) | after (dark) |
|---|---|---|---|---|
| `margin-top` | 0px | **16px** | 0px | **16px** |
| `margin-bottom` | 0px | **16px** | 0px | **16px** |
| `border-bottom` | `0px` | **`1px solid rgb(240,240,240)`** | `0px` | **`1px solid rgb(48,48,48)`** |
| gap to previous sibling | 0px | **16px** | 0px | **16px** |
| gap to next sibling | 0px | **16px** | 0px | **16px** |
| nav width | 672px | 672px | 672px | 672px |
| `pageerror` | 0 | 0 | 0 | 0 |

The rail colour differs between modes (`#F0F0F0` vs `#303030`), i.e. it resolves
through the theme rather than being hardcoded.

Dark mode was entered via the OS `prefers-color-scheme` the app's `system` theme
mode follows, with `document.documentElement.dataset.theme` asserted `'dark'`
before measuring. (Setting that attribute by hand does **not** re-theme the app —
an earlier pass that did so produced a light-looking "dark" screenshot and was
discarded.)

### Light

| before | after |
|---|---|
|  |  |

### Dark

| before | after |
|---|---|
|  |  |

## Verification

- `pnpm --prefix ./react exec vitest run` → **66 files, 1166 tests passed**
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → exits 1,
  **pre-existing**: the sole finding is
  `packages/backend.ai-ui/src/components/BAIText.tsx:230 var(--font-family-mono)`,
  untouched here. No finding names `RuntimeParameterFormSection`.
- No `packages/backend.ai-ui/src` change, so no BUI rebuild was required.

## Residue

- **Only this call site.** Four other `TabList` call sites render without
  `hasDivider` too — `SessionDetailContent`, `AgentDetailDrawerContent`,
  `StorageHostDetailDrawerContent`, `RoleDetailDrawerContent`. They were not
  touched and not reproduced; whether they *want* a rail is a per-surface
  decision, not something to sweep from one report. Worth a follow-up if the
  team wants the rail to be the default everywhere (i.e. route those call sites
  through `BAITabList`).
- The same section also renders on **Admin → Deployments → Deployment Presets →
  New/Edit**; it gets the fix for free but was not separately re-measured.
- The 16px value is a plain `--spacing-4` pick, not derived from a legacy
  reference — there is no pre-migration screenshot of this section to match
  against.

### DIFF
--- a/react/src/components/RuntimeParameterFormSection.tsx
+++ b/react/src/components/RuntimeParameterFormSection.tsx
@@ -23,10 +23,21 @@ import { Collapsible } from '@astryxdesign/core/Collapsible';
 import { IconButton } from '@astryxdesign/core/IconButton';
 import { Tab, TabList } from '@astryxdesign/core/TabList';
 import { Text } from '@astryxdesign/core/Text';
+import { spacingVars } from '@astryxdesign/core/theme/tokens.stylex';
+import * as stylex from '@stylexjs/stylex';
 import { BAIFlex, toLocalId } from 'backend.ai-ui';
 import { Undo2 } from 'lucide-react';
 import React, { useEffect, useEffectEvent, useState } from 'react';
 import { useTranslation } from 'react-i18next';
+
+const styles = stylex.create({
+  // The tab strip is a bare sibling of the Banner and the first field, so its
+  // breathing room has to come from the strip itself. FR-3529.
+  categoryTabs: {
+    marginBlockStart: spacingVars['--spacing-4'],
+    marginBlockEnd: spacingVars['--spacing-4'],
+  },
+});
 
 /** Convert category slug to a display-friendly label. */
 function formatCategoryLabel(category: string): string {
@@ -272,6 +283,8 @@ const RuntimeParameterFormSection: React.FC<
           already is. */}
       <TabList
         size="sm"
+        hasDivider
+        xstyle={styles.categoryTabs}
         value={effectiveActiveTab}
         onChange={(key) => setActiveTab(key)}
       >


ironAiken2 (approved):
LGTM

PR #8728, opened 2026-08-12
fix(FR-3517): return the whole-path copy to the "Path" label and size the path row down

### LINKED ISSUE #8706: Folder explorer metadata panel: Path row shows two indistinguishable copy icons, oversized trailing icon, and outsized path text
## Where

- **Page**: Data page → open a folder → the explorer's metadata panel ("메타데이터" tab) — the "경로" (Path) row.
- **Region**: the path breadcrumb value and the copy icons at its end.
- **Component** (if identified): `VFolderNodeDescriptionV2` (`react/src/components/VFolderNodeDescriptionV2.tsx:158-172`) rendering `VirtualFolderPathV2` (`react/src/components/VirtualFolderNodeItems/VirtualFolderPathV2.tsx`) plus a trailing full-path copy `IconButton`.

## Observed

Three things on the same row:

1. **Two adjacent copy icons at the end of the path.** The VFolder-ID segment's copy icon (orange, from `BAICopyableText`) sits directly next to a second, larger white copy icon (the full-path copy `IconButton`). Nothing distinguishes their targets, so they read as an accidental duplicate.
1. **The path text renders larger than the neighbouring metadata values** — the segments are drawn as monospace code boxes that visually outweigh the other rows in the same list.
1. **The trailing copy icon is larger than the other icons on the panel** (compare the segment copy icons and the panel's other inline icons).

## Expected

One visually coherent path row: copy affordances that are distinguishable (or a single one), path text at the same visual weight as sibling metadata values, and icons at the panel's standard icon size — matching the pre-migration rendering of this row.

## Environment

- **Mode**: dark (as reported); light not checked
- **Build**: `main` @ 2026-08-12 — `@astryxdesign/core` 0.3.0

## Reproduction

1. Open the Data page and open any folder.
1. In the explorer, look at the metadata panel's "경로" row.

## Evidence

- Attached: `path-row-copy-icons.png` — the row with the two adjacent copy icons and the oversized trailing one.

## Notes

- The two icons do have different targets today (segment copy vs. whole-path copy) — the defect as reported is that this is not perceivable.
- Possible size cue, hedged: the trailing `<CopyIcon />` is the only icon in `VFolderNodeDescriptionV2` without the file's usual `size="1em"`, so it renders at lucide's 24px default.

JIRA Issue: FR-3517

yomybaby:
Resolves #8706 (FR-3517)

> Now targets `main` directly. An earlier revision rendered this copy control
> as `variant:link` and was stacked on `FR-3524` (#8731), which owns that theme
> augmentation; the latest head uses `BAIText copyable` instead, so the
> dependency is gone and the branch was rebased onto main.

## Mechanism

Both real sub-defects come from the same commit — the Astryx migration
(7d7b42388) — and both are *dropped* legacy detail, not new Astryx behaviour.

**(1) Two indistinguishable copy icons.** Before the migration the whole-path
copy lived on the **"Path" LABEL** (`Typography.Text copyable` wrapping the
label), so it sat in the label column, spatially separated from the VFolder-ID
segment's own copy icon in the value column. The migration moved it into the
value cell, which parked two identical 16px copy glyphs 4px apart with
different targets.

It is now **back beside the `경로` label**, rendered with the standard
`BAIText copyable` idiom (`copyable={{ text: vfolderPath }}` on the label
text), so it is the exact same control as the two segment copies in the value
cell — same ghost/accent styling, same shared `general.button.Copy` label.

> **Revision note (latest head).** An earlier revision rendered this as an
> icon-only `Button variant="link"` with a dedicated `전체 경로 복사` tooltip;
> the measured tables and a11y snapshots below describe THAT revision. The
> latest head replaces it with `BAIText copyable` and deletes the
> `data.folders.CopyFullPath` key from all 22 locale bundles. The
> disambiguation between the three copy controls now comes from **position**
> (label cell vs value cell), not from a distinct tooltip.

### The "Astryx forces this" claim was wrong

An earlier revision of this PR justified the value-cell position by saying
`MetadataListItem.label` is typed `string`, so Astryx *forces* the control out
of the label. **That is a type-level constraint, not a structural one.**
`@astryxdesign/core/src/MetadataList/MetadataListItem.tsx:158` builds
`labelContent` and drops `{label}` into it verbatim, then renders it as a JSX
child in **both** layout branches — `<dt>{labelContent}</dt>` (inline) and
`<dt>{labelContent}</dt>` (stacked). There is **no attribute path at all**: the
prop never reaches an `aria-label` / `title` / `alt`, so nothing stringifies it
and a `ReactNode` renders correctly today.

That is the "child only → safe" row of
`.specs/FR-3482-astryx-migration/CONVERSION-IDIOMS.md` **§3**, ratified from
this exact mistake (it happened twice in one day: here and FR-3531).

### How the widening is done

Not a cast at the call site — one wrapper, per
`.claude/rules/component-props-extension.md`. **`BAIMetadataListItem`** is new,
and sits beside `BAIMetadataList` in the same BUI module:

```tsx
export interface BAIMetadataListItemProps
  extends Omit<MetadataListItemProps, 'label'> {
  label: ReactNode;
}
```

The cast happens once inside; the file header carries the one-line reason.

Two deliberate choices:

- **`label` stays REQUIRED.** Astryx requires it, and an `Omit` + optional
  redeclare would let a call site silently render an empty `<dt>`. Only the
  *type* is widened, not the contract.
- **It lives in BUI, not `react/src/components/astryx-bui/`**, so it sits with
  the list wrapper it belongs to. Safe because `react/`,
  `packages/backend.ai-ui/`, and the root all resolve
  `@astryxdesign/core@0.3.0` to one physical path in the pnpm store, so
  `MetadataListContext` is a single module instance and the item still reads
  its parent list's label config across the package boundary.

`VFolderNodeDescriptionV2` previously rendered its rows with
`label={item.label as string}`; the row type is widened to `ReactNode` and the
cast is gone.

**(2) Path text outweighs sibling metadata values.** Every segment in
`VirtualFolderPathV2` used to carry `style={{ fontSize: '0.9em' }}` — the
legacy render deliberately sized the path *down* so a monospace string would
not outweigh its neighbours. The conversion to `Text type="code"` dropped that
override, so the path rendered at the full body size (14px) *and* in
monospace. Restored with the Astryx `size="sm"` prop — no CSS, no theme edit.
After the rebase onto main with FR-3518 (#8730) merged, the copyable segments
render BUI `BAIText`, which had no `size`; this PR adds a one-line `size`
passthrough to its underlying Astryx `Text` (typed as `TextProps['size']`,
already forwarded at runtime via `...restProps`). Still measures 12px in both
modes.

**(3) "The trailing copy icon is larger than the other icons" — DOES NOT
REPRODUCE.** The issue hedged this on the source cue that the trailing
`<CopyIcon />` is the only icon in the file without `size="1em"`, so it might
render at lucide's 24px default. Measured, it does not: Astryx sizes its own
glyph, so **all three** copy icons render at exactly 16×16 — including the two
segment controls (measured on `BAICopyableText`, since converged by FR-3518
onto `BAIText copyable`), whose lucide icons also carry the bare `width="24"`
attribute. What differs is **contrast**, not size: in dark mode the whole-path
icon was `rgb(255,255,255)` while the segment icons are the dimmer accent
`rgb(190,94,6)`. No size change was made for this, and none was needed. (The
contrast difference is gone on the latest head by construction — all three
controls are the same `BAIText copyable` control, so they share one tint.)

## Measured before / after (live, real app, both modes)

Dev server on `127.0.0.1:4460` against `http://10.82.0.130:8090`, Korean UI,
Data page → folder → Metadata panel → `경로` row. "Before" is this PR's
previous head (`d0f35b37`) measured on the same running server by reverting
only the call site. Dark asserted via **both**
`document.body.classList.contains('dark-theme')` and
`document.documentElement.dataset.theme === 'dark'` after reload.

### The whole-path copy control

*(measured on the earlier `Button variant="link"` revision — see the revision
note above; the latest head renders the same 16×16 inline footprint via
`BAIText copyable`, with the segment controls' ghost/accent styling instead of
the link tint)*

| | before | after |
|---|---|---|
| position | value cell (`<dd>`), after the path segments | **label cell (`<dt>`), beside `경로`** |
| affordance | visible text button `경로 복사하기` | icon-only, tooltip **`전체 경로 복사`** |
| variant | `ghost` | **`link`** (`data-variant="link"`, FR-3524) |
| box | **129×24** | **16×16** |
| padding | `8px 12px 8px 12px` | **`0px`** |
| color — light | `rgb(20, 20, 20)` (ghost body colour) | **`rgb(255, 122, 0)`** (link tint) |
| color — dark | `rgb(255, 255, 255)` | **`rgb(190, 94, 6)`** (link tint) |
| hover bg — light | `linear-gradient(rgba(0,0,0,.06), rgba(0,0,0,.06))` | **`none`** |
| hover bg — dark | `linear-gradient(rgb(38,38,38), rgb(38,38,38))` | **`none`** |
| hover affordance | — (box only) | **`text-decoration: underline`** |
| hover box | 129×24 | **16×16** (no growth) |

`background-color` is `rgba(0, 0, 0, 0)` in every cell, before and after — the
ghost hover wash was painted as a `background-image` gradient, which is why the
"box" survived a transparent background colour. The link variant has no
built-in variant object to inherit it from, so it goes away by construction.

Of the four review points, the latest head keeps **(1)** back beside the label
and **(4)** inline footprint, and supersedes **(2)** link colour and **(3)**
the dedicated `전체 경로 복사` tooltip: the control now deliberately matches
the segment copies (shared accent tint, shared `Copy` label), with position
carrying the disambiguation.

### Path segment text — `Text type="code"` computed `font-size` (unchanged, kept)

| | before this PR | after | sibling metadata values | legacy (pre-migration) |
|---|---|---|---|---|
| light | 14px | **12px** (`data-size="sm"`) | 14px sans | 0.9em × 14px = 12.6px |
| dark | 14px | **12px** (`data-size="sm"`) | 14px sans | 0.9em × 14px = 12.6px |

`size="sm"` resolves to `--font-size-sm: 0.75rem` = 12px, **0.6px below** the
legacy 12.6px — `sm` is the nearest declared token and the gate forbids a raw
px, so the 0.6px is accepted rather than hardcoded.

## Accessibility snapshot (Playwright `ariaSnapshot`, Korean UI)

**Before** — the label cell was inert text and the control was a third button
in the value cell:

```yaml
# dt
- term: 경로
# dd
- definition:
  - text: (root) / f38dea2350fa...
  - button "Copy":
    - status
  - text: Quota Scope ID (User) / 7b / f8 / fc87...
  - button "Copy":
    - status
  - text: VFolder ID
  - button "경로 복사하기":
    - text: 경로 복사하기
    - status
```

**After** *(earlier revision; on the latest head the button's accessible name
is the shared `Copy` label rather than `전체 경로 복사`, structure unchanged)*
— the label still reads as `경로`, and the control is a sibling with its own
accessible name:

```yaml
# dt
- term:
  - text: 경로
  - button "전체 경로 복사":
    - status
  - tooltip "전체 경로 복사"
# dd
- definition:
  - text: (root) / f38dea2350fa...
  - button "Copy":
    - status
  - text: Quota Scope ID (User) / 7b / f8 / fc87...
  - button "Copy":
    - status
  - text: VFolder ID
```

Role-based queries on the live page:

- `getByRole('button', { name: '전체 경로 복사', exact: true })` → **1**
- `getByRole('term')` → **10**, and the `경로` term's text is `경로 전체 경로 복사`
  — i.e. the label text is still first and intact; the added control
  contributes its own name rather than replacing it.

`packages/backend.ai-ui/src/components/BAIMetadataList.test.tsx` (new) locks
both halves in: a string label still renders, and a `ReactNode` label keeps
`getByRole('term')` reading the label while
`getByRole('button', { name: 'Copy' })` resolves separately.

Idiom §3's other required check — that no upstream code sorts or keys off the
widened label — holds: `items` is only `.map`ped, and the React key is
`item.key`, never `item.label`. No `localeCompare` / `.sort()` touches it.

## Label-column alignment

`MetadataList` lays labels out in a CSS grid, so a control added to one label
could widen the track and shift every other row. It does not.

`grid-template-columns` is **`84px 620px`** — **identical** before and after,
in both light and dark. All 10 label cells, before and after:

| row | x | width | height |
|---|---|---|---|
| `경로` | 776 | 84 | 24 |
| `상태` | 776 | 84 | 24 |
| `호스트` | 776 | 84 | 24 |
| `소유` | 776 | 84 | 24 |
| `마운트 권한` | 776 | 84 | 24 |
| `소유자` | 776 | 84 | 24 |
| `사용자` | 776 | 84 | 24 |
| `복사 가능 여부` | 776 | 84 | 24 |
| `사용 방식` | 776 | 84 | 24 |
| `생성 시간` | 776 | 84 | 24 |

Every sibling label sits at the same `x` and the same `width`, and the `경로`
row's own height stays **24px** — the 16×16 inline control does not grow its
row. (This is exactly what the link variant's `height: auto` + zero padding
buys; the 24px-tall `ghost` control would have.)

## i18n

One new key, `data.folders.CopyFullPath`, added to **all 21 locales**
(`Copy full path` / `전체 경로 복사`). `button.CopySomething` was not reused:
interpolating it gives `경로 복사하기`, and the point of this label is to
distinguish this control from the two segment copies beside it.

## Reconciliation with #8730 (FR-3518)

FR-3518 makes both segment copy controls on this very row resolve their tooltip
from i18n, so both read **`복사`**. That is what this control's
**`전체 경로 복사`** finally distinguishes it from. Once both land, the `경로`
row reads:

```
경로 ⧉                    (root) / f38dea2350fa… ⧉ / 7b / f8 / fc87… ⧉
 ↑ 전체 경로 복사 (link)              ↑ 복사              ↑ 복사
 in the LABEL cell                    both in the VALUE cell
```

— one link-tinted control on the label naming the whole path, two identical
accent glyphs on the value naming their own segment. FR-3518 (#8730) has since
merged and converged every copy control — `BAICopyableText` is gone and the
segment copies are `BAIText code copyable`, with their control label translated
through BUI's `general.button.Copy`; this branch is rebased on top and only
adds the `size` passthrough to `BAIText`.

## (4) Follow-up on the same panel — the explorer resize handle collapsed onto the tab strip

Reported while reviewing this PR's dev server: a small grey bar overlapping
the metadata tab's top-left corner, and no vertical divider between the file
list and the info panel.

**Mechanism.** Astryx `ResizeHandle` sizes itself with `height: 100%`, but the
explorer's split row is sized by `min-height` only, so the percentage resolves
to `auto` and the handle collapses to its ~30px content box pinned at the top
of the row. Its two visuals land wrong at once: the pill grip (3×32px,
`--color-border`) parks over the "메타데이터" tab's top-left corner, and the
`hasDivider` 1px between-panel line exists only for those ~30px — measured off
the report screenshot by pixel scan (pill at x≈834–836 / y≈91–121 vs tab box
top y=106; no divider pixels anywhere below y≈150).

**Fix.** Wrap the handle in a stretched flex wrapper
(`display: flex; align-self: stretch`) — a stretched flex item has a definite
cross size, so the handle's `height: 100%` resolves and the divider/pill span
the full split height again. No new paint; the wrapper draws nothing.

## Verification

| check | result |
|---|---|
| `bash scripts/verify.sh` | **`=== ALL PASS ===`** (Relay, Lint, Format, TypeScript, Vite warmup, StyleX sentinel, Astryx theme build, Terminology) |
| `pnpm --filter backend.ai-ui build` | built — required before `verify.sh`, whose `astryx theme build --check` cannot resolve `backend.ai-ui` without a `dist` |
| `pnpm --prefix ./react exec vitest run` | **66 files / 1166 tests passed**, 0 failed |
| `pnpm --filter backend.ai-ui exec vitest run` | **43 passed + 1 skipped files / 605 passed + 1 skipped tests**, 0 failed |
| `node scripts/migration-gates/astryx-token-gate.mjs --strict` | exit 1 — **2 findings, both PRE-EXISTING on `main`** in `packages/backend.ai-ui/src/components/BAIText.{css,tsx}` (`var(--x)`, `var(--font-family-mono)`), untouched here |
| live probe | light + dark re-measured on the running dev server, **0 `pageerror` in both passes** |

The BUI counts moved from the previous revision's *42 files / 590 tests*
because the base changed: on `FR-3524` alone the suite is **42 files (+1
skipped) / 603 tests (+1 skipped)**, and this PR's new
`BAIMetadataList.test.tsx` adds **1 file / 2 tests**.

## Residue — what this does NOT close

- **Copy buttons are 16px while the panel's inline status glyphs are 14px.**
  Measured and real, but uniform across all three copy buttons and inherited
  from the Astryx icon recipe, not from this file. Left alone.
- **`react/src/components/VFolderNodeDescription.tsx`** (the V1 sibling) has
  the same shape, but its only consumer, `FolderExplorerModal.tsx`, is imported
  by nothing — dead code. Not touched.
- **`MetadataListItem.label` wants to be `ReactNode` upstream.** CONVERSION-IDIOMS §3
  measured 72 `string`-typed slot props rendered as children in
  `@astryxdesign/core@0.3.0`, 39 with no attribute path at all. Worth reporting
  to Astryx rather than carrying wrappers forever; `BAIMetadataListItem` is the
  local stopgap.
- The theme was **not** edited here, so `THEME_NAME_REV` is untouched — the REV
  13 bump belongs to #8731.

### DIFF
--- a/packages/backend.ai-ui/src/components/BAIMetadataList.test.tsx
+++ b/packages/backend.ai-ui/src/components/BAIMetadataList.test.tsx
@@ -1,0 +1,45 @@
+/*
+ FR-3517 — `BAIMetadataListItem` widens Astryx's `string`-typed `label` to a
+ `ReactNode` (CONVERSION-IDIOMS §3). These assert the two things idiom 3 says to
+ check after using it: the node actually renders into the label cell, and the
+ row's accessible name still reads as the label with the added control
+ separately reachable.
+*/
+import BAIMetadataList, { BAIMetadataListItem } from './BAIMetadataList';
+import { render, screen } from '@testing-library/react';
+
+describe('BAIMetadataListItem', () => {
+  it('renders a plain string label as before', () => {
+    render(
+      <BAIMetadataList>
+        <BAIMetadataListItem label="Path">/vroot/local</BAIMetadataListItem>
+      </BAIMetadataList>,
+    );
+    expect(screen.getByRole('term')).toHaveTextContent('Path');
+    expect(screen.getByRole('definition')).toHaveTextContent('/vroot/local');
+  });
+
+  it('renders a ReactNode label, keeping the label text and the control separate', () => {
+    render(
+      <BAIMetadataList>
+        <BAIMetadataListItem
+          label={
+            <>
+              Path
+              <button type="button" aria-label="Copy full path" />
+            </>
+          }
+        >
+          /vroot/local
+        </BAIMetadataListItem>
+      </BAIMetadataList>,
+    );
+
+    // The label cell still reads as the label…
+    expect(screen.getByRole('term')).toHaveTextContent('Path');
+    // …and the control is reachable under its own accessible name.
+    expect(
+      screen.getByRole('button', { name: 'Copy full path' }),
+    ).toBeInTheDocument();
+  });
+});
--- a/packages/backend.ai-ui/src/components/BAIMetadataList.test.tsx
+++ b/packages/backend.ai-ui/src/components/BAIMetadataList.test.tsx
@@ -26,7 +26,7 @@ describe('BAIMetadataListItem', () => {
           label={
             <>
               Path
-              <button type="button" aria-label="Copy full path" />
+              <button type="button" aria-label="Copy" />
             </>
           }
         >
@@ -38,8 +38,6 @@ describe('BAIMetadataListItem', () => {
     // The label cell still reads as the label…
     expect(screen.getByRole('term')).toHaveTextContent('Path');
     // …and the control is reachable under its own accessible name.
-    expect(
-      screen.getByRole('button', { name: 'Copy full path' }),
-    ).toBeInTheDocument();
+    expect(screen.getByRole('button', { name: 'Copy' })).toBeInTheDocument();
   });
 });
--- a/packages/backend.ai-ui/src/components/BAIMetadataList.tsx
+++ b/packages/backend.ai-ui/src/components/BAIMetadataList.tsx
@@ -25,9 +25,11 @@
 import './BAIMetadataList.css';
 import {
   MetadataList,
+  MetadataListItem,
+  type MetadataListItemProps,
   type MetadataListProps,
 } from '@astryxdesign/core/MetadataList';
-import React from 'react';
+import React, { type ReactNode } from 'react';
 
 export interface BAIMetadataListProps extends MetadataListProps {
   /**
@@ -83,5 +85,28 @@ const BAIMetadataList: React.FC<BAIMetadataListProps> = ({
 };
 
 BAIMetadataList.displayName = 'BAIMetadataList';
+
+/**
+ * Astryx types `label` as `string`, but `MetadataListItem` renders it as a JSX
+ * child with no attribute path, so a node works (CONVERSION-IDIOMS §3).
+ */
+export interface BAIMetadataListItemProps extends Omit<
+  MetadataListItemProps,
+  'label'
+> {
+  label: ReactNode;
+}
+
+export const BAIMetadataListItem: React.FC<BAIMetadataListItemProps> = ({
+  label,
+  ...metadataListItemProps
+}) => {
+  'use memo';
+  return (
+    <MetadataListItem {...metadataListItemProps} label={label as string} />
+  );
+};
+
+BAIMetadataListItem.displayName = 'BAIMetadataListItem';
 
 export default BAIMetadataList;
--- a/packages/backend.ai-ui/src/components/BAIText.tsx
+++ b/packages/backend.ai-ui/src/components/BAIText.tsx
@@ -93,6 +93,8 @@ export interface BAITextProps extends Omit<
   code?: boolean;
   disabled?: boolean;
   monospace?: boolean;
+  /** Astryx `Text` size step, forwarded as-is (antd had no counterpart). */
+  size?: TextProps['size'];
   /** CSS-based ellipsis (multi-line via `rows`), with an optional tooltip. */
   ellipsis?: boolean | BAITextEllipsisConfig;
   copyable?: boolean | BAITextCopyConfig;
--- a/packages/backend.ai-ui/src/components/index.ts
+++ b/packages/backend.ai-ui/src/components/index.ts
@@ -16,6 +16,8 @@ export { default as BAICompactGroup } from './BAICompactGroup';
 export type { BAICompactGroupProps } from './BAICompactGroup';
 export { default as BAIMetadataList } from './BAIMetadataList';
 export type { BAIMetadataListProps } from './BAIMetadataList';
+export { BAIMetadataListItem } from './BAIMetadataList';
+export type { BAIMetadataListItemProps } from './BAIMetadataList';
 export {
   default as BAIPropertyFilter,
   mergeFilterValues,
--- a/react/src/components/FolderExplorerModalV2.tsx
+++ b/react/src/components/FolderExplorerModalV2.tsx
@@ -530,13 +530,20 @@ const FolderExplorerModalV2: React.FC<FolderExplorerProps> = ({
                   <div style={{ flex: 1, minWidth: 0 }}>
                     {fileExplorerElement}
                   </div>
-                  <ResizeHandle
-                    direction="horizontal"
-                    isReversed
-                    hasDivider
-                    label={t('explorer.Metadata')}
-                    resizable={infoPanel.props}
-                  />
+                  {/* The handle's own `height: 100%` resolves to `auto` here —
+                      this row is sized by `min-height` only, which makes the
+                      percentage indefinite, collapsing the handle (divider +
+                      pill) to ~30px pinned at the top, over the tab strip. A
+                      stretched flex wrapper gives it a definite height. */}
+                  <div style={{ display: 'flex', alignSelf: 'stretch' }}>
+                    <ResizeHandle
+                      direction="horizontal"
+                      isReversed
+                      hasDivider
+                      label={t('explorer.Metadata')}
+                      resizable={infoPanel.props}
+                    />
+                  </div>
                   <div style={{ width: infoPanel.size, flexShrink: 0 }}>
                     {vFolderInfoPanelElement}
                   </div>
--- a/react/src/components/VFolderNodeDescriptionV2.tsx
+++ b/react/src/components/VFolderNodeDescriptionV2.tsx
@@ -10,7 +10,9 @@
 
  PILOT-DECISIONs:
  - The copy affordance on the "Path" LABEL moves next to the path VALUE:
-   `MetadataListItem.label` is a plain string (P2).
+   `MetadataListItem.label` is a plain string (P2). FR-3517: it carries a
+   visible label there, so its target is distinguishable from the segment
+   copies it now sits beside.
  - The disabled 'rw' option's inline Tooltip is dropped — Astryx `Selector`
    options take string labels; the client-side guard in `onChange` still
    blocks the restricted transition.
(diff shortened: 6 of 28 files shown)


agatha197 (dismissed):
Image

Image

i18n is only applied to `BAICopyableText`.

ironAiken2 (changes requested):
It appears that some sections are not available in the i18n translation.

ironAiken2 (approved):
LGTM

nowgnuesLee (approved):
LGTM

Issue #8702, opened 2026-08-11
BAIModal: proposal - render the loading body with BAISkeletonAstryx instead of the bare Astryx Skeleton


nowgnuesLee:
## Question

Proposal: should `BAIModal`'s `loading` state render `BAISkeletonAstryx` (the BUI skeleton wrapper) instead of the bare Astryx `Skeleton` it uses today?

## Where

- **Component**: `BAIModal` (`packages/backend.ai-ui/src/components/BAIModal.tsx:758`) — its `loading` prop renders `<Skeleton height={120} />` imported directly from `@astryxdesign/core/Skeleton`.
- **Proposed replacement**: `BAISkeletonAstryx` (`react/src/components/astryx-bui/BAISkeletonAstryx.tsx`) — the antd-parity skeleton wrapper (title + paragraph rows, avatar, input/button variants), used by 75 files.
- **Recurs in**: every modal that passes `loading` — the fallback is centralized in `BAIModal`, so one change covers all of them.

## Current behaviour

While a modal's content is loading, the body shows a single 120px-tall skeleton box. `BAISkeletonAstryx`, used across the rest of the app for loading fallbacks, renders the multi-part shimmer (title bar + paragraph rows) that antd's `Skeleton` produced before the migration.

## Why this is being raised

The modal's loading fallback is the one place still on the bare single-box primitive while the app-wide convention for content-shaped loading states is `BAISkeletonAstryx`, so modal bodies flash a visibly different (flat, single-block) placeholder than every other loading surface.

## Proposal

From the reporter, stated as one option:

> Use `BAISkeletonAstryx` in place of the bare Astryx `Skeleton` for `BAIModal`'s `loading` body.

## Options considered

1. Swap to `BAISkeletonAstryx` (the proposal).
1. Keep the bare single-box `Skeleton` as a deliberate, cheaper modal fallback.

## Impact if it stays as-is

Cosmetic inconsistency — modal loading states look flatter than the rest of the app's skeleton fallbacks.

## Environment

- **Build**: `main` @ 2026-08-12 — `@astryxdesign/core` 0.3.0

## Evidence

- `BAIModal.tsx:758`: `{loading ? <Skeleton height={120} /> : children`}
- `BAISkeletonAstryx` usage breadth: 75 files (e.g. `ResourceGroupSettingModal.tsx`, `VFolderTextFileEditorModal.tsx`, `DeploymentAutoScalingCard.tsx`).

## Notes

[...]

PR #8729, opened 2026-08-12
fix(FR-3514): render the modal loading body as antd's paragraph skeleton

### LINKED ISSUE #8703: BAIModal: loading body renders a flat single-box Skeleton instead of the app-wide multi-part skeleton fallback
## Where

- **Page**: any modal opened with the `loading` prop — e.g. setting modals while their query resolves.
- **Region**: the modal body during loading.
- **Component** (if identified): `BAIModal` (`packages/backend.ai-ui/src/components/BAIModal.tsx:758`) — its `loading` branch renders `<Skeleton height={120} />` imported directly from `@astryxdesign/core/Skeleton`.

## Observed

While a modal's content loads, the body shows a single flat 120px skeleton box.

## Expected

The multi-part content-shaped shimmer (title bar + paragraph rows) used by loading fallbacks everywhere else in the app — the established convention is `BAISkeletonAstryx` (`react/src/components/astryx-bui/BAISkeletonAstryx.tsx`, 75 consuming files), which reproduces what antd's `Skeleton` rendered in modals before the migration.

## Environment

- **Mode**: both (not mode-specific)
- **Build**: `main` @ 2026-08-12 — `@astryxdesign/core` 0.3.0

## Reproduction

1. Open any modal whose content loads asynchronously (a settings modal backed by a query).
1. Watch the body before the content arrives.

## Evidence

- `BAIModal.tsx:758`: `{loading ? <Skeleton height={120} /> : children`}
- Convention comparison: `BAISkeletonAstryx` is the loading fallback in 75 files (e.g. `ResourceGroupSettingModal.tsx`, `VFolderTextFileEditorModal.tsx`, `DeploymentAutoScalingCard.tsx`).

## Notes

- Fix-time consideration: `BAIModal` lives in `packages/backend.ai-ui` while `BAISkeletonAstryx` lives in `react/src/components/astryx-bui/`, so the swap crosses a workspace boundary — where the shared skeleton lives is for the fixer to settle.

JIRA Issue: FR-3514

yomybaby:
Resolves #8703 (FR-3514)

## Mechanism

**Astryx's `Skeleton` is one box; antd's was a composition — and the composition lived on the wrong side of the workspace boundary.**

antd's `Modal` rendered `<Skeleton active title={false} paragraph={{ rows: 4 }} />` for `loading` ([upstream source](https://github.com/ant-design/ant-design/blob/master/components/modal/Modal.tsx)). Astryx's `Skeleton` has no `paragraph`/`title` concept at all — it is a single `width`/`height`/`radius` box — so the migration substituted `<Skeleton height={120} />`, a flat grey slab.

The multi-part shimmer *does* exist in this repo: `BAISkeletonAstryx` composes title bar + N paragraph rows (last one short) with an incrementing `index` so the wave reads as one motion, and it is already the loading fallback in **74 files**. It just lived in `react/src/components/astryx-bui/`, and `packages/backend.ai-ui` cannot import from `react/src` — so `BAIModal` had no way to reach it. That import direction, not the visual choice, is why the modal ended up with the primitive.

**How the mechanism was established (and what measuring changed):** the report points at `BAIModal.tsx:758`. Fixing only that changed **nothing** on screen — the first live re-probe came back byte-identical to the before state. There are **two** modal implementations with the same line: `packages/backend.ai-ui/src/components/BAIModal.tsx` (BUI) and `react/src/components/astryx-bui/BAIModalAstryx.tsx:209` (the pilot copy still backing the 12 VFolder-area modals, including the `/data` → *Create Folder* modal used to reproduce the bug). Both are fixed here; fixing one alone would have shipped a diff that reads correct and is invisible in half the app.

## Where the shared skeleton lives (the call the issue left open)

`BAISkeletonAstryx` **moves into `packages/backend.ai-ui/src/components/`** and is exported from the BUI barrel. `react/src/components/astryx-bui/BAISkeletonAstryx.tsx` becomes a 6-line re-export of it, so **all 74 existing import sites keep working untouched** and there is exactly **one** implementation of the shimmer. The alternative — composing the same shape inline in `BAIModal` — was rejected: it would have created a second, divergent copy of the very thing this issue is about.

`hasTitle={false}` is deliberate and matches antd: the dialog header already carries the title, so a title-bar shimmer in the body would double it up.

## Measured before / after (live, real app, both modes)

Reproduced by holding the modal's own query open with a Playwright route delay. Values are computed styles read off the live DOM inside `dialog[open]`.

**`BAIModal` (BUI) — "Bulk Edit Resource Groups", `/agent?tab=resourceGroup`**

| | before (light) | after (light) | before (dark) | after (dark) |
|---|---|---|---|---|
| skeleton boxes in body | **1** | **4** | **1** | **4** |
| box heights | 120px | 16 / 16 / 16 / 16px | 120px | 16 / 16 / 16 / 16px |
| box widths | 472px | 472 / 472 / 472 / **287.9**px | 472px | 472 / 472 / 472 / **287.9**px |
| stack `gap` | `normal` (no stack) | `16px` | `normal` (no stack) | `16px` |
| `border-radius` | 8px | 6px | 8px | 6px |
| `animation-delay` | `1s` | `1s / 1.1s / 1.2s / 1.3s` | `1s` | `1s / 1.1s / 1.2s / 1.3s` |
| `background-color` | `rgba(0,0,0,0.15)` | `rgba(0,0,0,0.15)` | `rgba(255,255,255,0.18)` | `rgba(255,255,255,0.18)` |
| dialog height | 266px | 258px | 266px | 258px |

**`BAIModalAstryx` (react) — "Create a new storage folder", `/data`**

| | before (light) | after (light) | before (dark) | after (dark) |
|---|---|---|---|---|
| skeleton boxes in body | **1** | **4** | **1** | **4** |
| box heights | 120px | 16 / 16 / 16 / 16px | 120px | 16 / 16 / 16 / 16px |
| box widths | 602px | 602 / 602 / 602 / **367.2**px | 602px | 602 / 602 / 602 / **367.2**px |
| stack `gap` | `normal` (no stack) | `16px` | `normal` (no stack) | `16px` |
| `border-radius` | 8px | 6px | 8px | 6px |
| `animation-delay` | `1s` | `1s / 1.1s / 1.2s / 1.3s` | `1s` | `1s / 1.1s / 1.2s / 1.3s` |
| `background-color` | `rgba(0,0,0,0.15)` | `rgba(0,0,0,0.15)` | `rgba(255,255,255,0.18)` | `rgba(255,255,255,0.18)` |
| dialog height | 266px | 258px | 266px | 258px |

The staggered `animation-delay` (1.0 → 1.3s) is the part a static diff cannot show: the four rows shimmer as one wave rather than four synchronised pulses. Colours are unchanged in both modes — this is a shape fix, not a colour fix. Dark measurements were taken with `document.body.classList.contains('dark-theme') === true` asserted (the app keys dark off `body.dark-theme`, not `documentElement.dataset.theme`). **Zero `pageerror` in all four passes.**

### Screenshots (`BAIModal`, Bulk Edit Resource Groups)

| | before | after |
|---|---|---|
| light | before light | after light |
| dark | before dark | after dark |

## Verification

| gate | result |
|---|---|
| `bash scripts/verify.sh` | **`=== ALL PASS ===`** (Relay, Lint, Format, TypeScript, Vite warmup, StyleX sentinel, Astryx theme build `--check`, Terminology) |
| `pnpm --filter backend.ai-ui build` | ✅ built (required — `packages/backend.ai-ui/src` changed) |
| `pnpm --prefix ./react exec vitest run` | **66 files / 1166 tests passed**, 0 failed |
| `pnpm --filter backend.ai-ui exec vitest run` | **42 files passed + 1 skipped / 590 tests passed + 1 skipped**, 0 failed |
| `node scripts/migration-gates/astryx-token-gate.mjs --strict` | exit 1 with **2 undeclared**, both in `BAIText.css:28` / `BAIText.tsx:225` — **pre-existing**; verified byte-identical output on `main` with this diff removed. **0 new.** |
| live probe | 4/4 passes (2 modals × light/dark), 0 `pageerror` |

No theme edit, so no `THEME_NAME_REV` bump and no `bai-r*-*` artifact regeneration. Nothing under root `scripts/` was touched.

## Residue — what this does *not* close

- **The `Astryx` suffix stays.** `BAISkeletonAstryx` keeps its migration-era name in its new BUI home so the 74 call sites and their default-import identifier are untouched. Renaming it to `BAISkeleton` is a separate, mechanical follow-up; BUI already exports other `*Astryx`-suffixed components (`BAIProjectSelectAstryx`, `BAIBucketSelectAstryx`, …) so this is not a new inconsistency.
- **The app-side re-export shim stays.** There is now one import *path* per workspace but only one *implementation*. Migrating the 74 sites to `import { BAISkeletonAstryx } from 'backend.ai-ui'` and deleting the shim is out of scope here — it would bury a 4-line behavioural fix under a 74-file diff.
- **`react/src/components/FolderCreateModal.tsx` (V1) is dead code** — it passes `loading` to BUI's `BAIModal`, but nothing imports it (`FolderCreateModalV2` replaced it everywhere). It benefits from the fix; its deletion is not proposed here.
- **The two modal implementations are not merged.** `BAIModalAstryx` (react, `isLoading`/`onAction`/`isOpen`) and `BAIModal` (BUI, antd-shaped `loading`/`onOk`/`open`) remain separate components with separate prop vocabularies. This PR makes their loading bodies agree; consolidating them is [FR-3515](https://lablup.atlassian.net/browse/FR-3515)'s territory, not this one.
- **Not subsumed:** [FR-3495](https://lablup.atlassian.net/browse/FR-3495) (modal header title size, already Done) and [FR-3515](https://lablup.atlassian.net/browse/FR-3515) (`onOk`/`onCancel` + footer render-prop contract, still To Do) — this PR touches neither the header nor the footer contract.
- **Other flat-`Skeleton` call sites are untouched.** `height={120}` existed in exactly these two places; the ~20 BUI select fragments that render the Astryx `Skeleton` primitive directly are input-shaped placeholders where a single box is correct.

## Reviewer: look hardest at

1. **The re-export shim** (`react/src/components/astryx-bui/BAISkeletonAstryx.tsx`) — it re-exports through the BUI *barrel* rather than a deep path. That is the only public entry (`backend.ai-ui`), and `routes.tsx` already imports the barrel, so no new module-graph weight, but it is worth a second opinion.
2. **`hasTitle={false}`** — the issue's "Expected" says "title bar + paragraph rows". I dropped the title bar because that is what antd's `Modal` actually did and the dialog header already renders the title. If the team prefers the title bar, it is a one-word change.
3. **`BAIModalAstryx` was not in the issue's scope** but is the implementation behind the modal you can actually reach in the UI. Confirm you want both changed in one PR rather than split.

[FR-3515]: https://lablup.atlassian.net/browse/FR-3515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


react/src/components/astryx-bui/BAISkeletonAstryx.tsx
agatha197, line 1: It just exports BUI components. Is it needed?

agatha197 (changes requested):

ironAiken2:
Closing as a duplicate: this and FR-3513 (#8702) report the same defect (`BAIModal.tsx:758` bare `<Skeleton height={120} />`). Superseded by #8740, which additionally settles the package-boundary question both issues flagged — `BAISkeletonAstryx` moved into `backend.ai-ui` as `BAISkeleton` (alias dropped, all 74 call sites updated) and `BAIModal`'s loading body renders it. Closed per @sungchulhong's direction; apologies for the crossed wires, and please leave feedback on #8740 if anything from this PR should carry over.

PR #8732, opened 2026-08-12
fix(FR-3525): keep the lab Tour callout above its spotlight backdrop

### LINKED ISSUE #8719: Astryx lab Tour: step description card renders behind the backdrop mask (deployment preset validation tour)
## Where

- **Page**: `/admin/deployments/deployment-presets/new?step=review` (menu: Admin → Deployments → Deployment Presets → create, Review step) — the validation tour
- **Region**: the tour's step description card (title "입력 필드 검증 오류", step counter "1 of 3", Next button)
- **Component** (if identified): Astryx lab `Tour` / `TourStep` (`@astryxdesign/lab`), via `react/src/components/AdminDeploymentPresetValidationTour.tsx`

## Observed

The element spotlight renders correctly (the target card is cut out of the

backdrop at full contrast), but the tour's description card is dimmed — it

appears **behind** the backdrop mask. It remains interactive (Next is clickable),

but visually it reads as part of the dimmed background rather than the

foreground of the tour.

## Expected

The description card renders above the backdrop mask at full contrast, like the

spotlighted element — the two foreground layers of a tour.

## Environment

- **Mode**: light (screenshot); dark Unknown
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: dev, 2026-08-12 (Korean UI; observed on the FR-3520 branch dev server)

## Reproduction

1. Open the deployment preset create page, Review step, with an empty form.
1. Trigger validation so the error tour starts.
1. Compare the spotlighted card's contrast with the description card's.

## Evidence

- Screenshot: `/tmp/clipimg/clip-20260812-141550.png` (reporter's capture, to be attached in Jira)

## Notes

The same lab `Tour` is used by `SessionLauncherErrorTourProps.tsx` (usage

breadth only — not verified there). Scope as reported: this tour.

JIRA Issue: FR-3525

yomybaby:
Resolves #8719 (FR-3525)

## Mechanism — top-layer promotion order, inverted by StrictMode's double-invoke

The tour draws two things: a **spotlight overlay** (`TourHighlight` in
`@astryxdesign/lab/src/Tour/TourStep.tsx`, a `popover="manual"` div whose inner
`[data-testid="tour-highlight"]` box carries
`box-shadow: … , 0 0 0 9999px var(--color-overlay)` — that 9999px spread *is*
the dim) and the **callout** (a core `Popover`). Both are promoted into the
browser top layer with `showPopover()`, and **top-layer paint order is
promotion order**, not z-index — neither element has a z-index at all
(`z-index: auto` measured on both). The file's own comment states the intended
invariant: the highlight is "promoted once (never re-promoted) so that ordering
holds", with the callout promoted after it.

It was not promoted once. The promotion lived in a mount layout effect whose
cleanup called `hidePopover()`. React StrictMode (this app mounts under
`<React.StrictMode>`, `react/src/index.tsx`) double-invokes effects in dev:
cleanup-all, then re-run-all. The callout's `useLayer.show()` is guarded by its
own `isOpenRef` and does **not** re-promote, so the highlight came back **last**
and its dim painted over the callout.

Instrumenting `HTMLElement.prototype.showPopover` on the live page, before the
fix:

```
SHOWPOPOVER #1 tour-backdrop
SHOWPOPOVER #2 <callout "Input field validation errors…">
SHOWPOPOVER #3 tour-backdrop      ← StrictMode re-run; overlay now on top
```

Two controls confirm the mechanism rather than a look-alike:

- **Remove StrictMode locally** (temporary, reverted): only `#1` + `#2` fire and
  the callout renders at full contrast.
- **Re-promote the callout by hand** (`hidePopover(); showPopover()` on the
  callout, backdrop untouched): the callout surface goes from `rgb(140,140,140)`
  back to `rgb(255,255,255)` while the page stays dimmed at `rgb(140,140,140)`.

It is *not* an opacity or isolation problem: `opacity: 1` and
`isolation: auto` on the callout throughout. And the card stays clickable
because the overlay root sizes to **0×0** — `inset: 0` cannot beat the UA
popover rule's `width/height: fit-content`, and its only child is
`position: absolute` — so it captures no pointer events even though it paints
across the viewport.

## The fix

A pnpm patch on `@astryxdesign/lab@0.3.0-canary.12db2a1` (the repo already
carries three such patches under `react/patches/`): promote the overlay only
when it is not already open, and drop the `hidePopover()` cleanup — detaching
the node removes it from the top layer on its own. That makes the code do what
its comment already claimed.

```diff
-    el.showPopover();
-    return () => {
-      if (el.isConnected && typeof el.hidePopover === 'function') {
-        el.hidePopover();
-      }
-    };
+    if (!el.matches(':popover-open')) {
+      el.showPopover();
+    }
```

There is no fix available at the theme / prop / composition level: `hasBackdrop`
is the only knob, and turning it off drops the spotlight entirely. **This should
go upstream to facebook/astryx** (lab `Tour`); the patch is the diff to send.

## Measured before / after

Live on the dev server, `/admin/deployments/deployment-presets/new?step=review`
with an empty form, 1600×1000. Values are screenshot pixel samples (sRGB) at the
callout's bottom-left interior, with two references in the same frame.

| Sample | Light before | Light after | Dark before | Dark after |
|---|---|---|---|---|
| **Callout surface** | `rgb(140,140,140)` ❌ | `rgb(255,255,255)` ✅ | `rgb(17,17,17)` ❌ | `rgb(31,31,31)` ✅ |
| Spotlighted target (undimmed reference) | `rgb(255,255,255)` | `rgb(255,255,255)` | `rgb(20,20,20)` | `rgb(20,20,20)` |
| Plain page (dimmed reference) | `rgb(140,140,140)` | `rgb(140,140,140)` | `rgb(11,11,11)` | `rgb(11,11,11)` |

Before, the callout surface is pixel-identical to the dimmed page. After, it
matches the untinted surface token, and the backdrop still dims everything else
— `0.55 × 255 = 140` and `0.55 × 31 = 17` are exactly the
`--color-overlay: rgba(0,0,0,0.45)` composites, so the "before" numbers are the
dim landing on the card and nothing else.

Interaction re-exercised after the fix (light): all three steps, Next → Next,
callout surface `rgb(255,255,255)` on every step against a `rgb(140,140,140)`
page, `body.dark-theme` asserted for the dark pass, **0 pageerrors** in every
run.

Screenshots were not attached: the shared `assets/images` branch is checked out
by a concurrent agent worktree and `img.sh` could not commit. The pixel samples
above are the same evidence in numeric form.

## Verification

| Check | Result |
|---|---|
| `bash scripts/verify.sh` | `=== ALL PASS ===` |
| `pnpm --filter backend.ai-ui build` | built (required first — the fresh worktree had no `dist`, which fails `astryx theme build --check`) |
| `pnpm --prefix ./react exec vitest run` | 66 files, **1166 passed** |
| `pnpm --filter backend.ai-ui exec vitest run` | 42 files + 1 skipped, **590 passed / 1 skipped** |
| `node scripts/migration-gates/astryx-token-gate.mjs --strict` | exit 1 — **2 pre-existing findings**, both in `packages/backend.ai-ui/src/components/BAIText.{css,tsx}` (`var(--x)`, `var(--font-family-mono)`), untouched by this PR |
| `pnpm install --frozen-lockfile` | clean (patch path lives in `react/patches/` next to the existing three; the lockfile records the content hash, not the path) |

## Residue

- **This is a development-build defect.** StrictMode's effect double-invoke is
  dev-only, so a production bundle promoted the two layers in the right order
  already. The report came from a dev server, and the patch also removes the
  latent fragility for any future re-promotion (remount, fast refresh) —
  but do not read it as a production regression fix.
- **`SessionLauncherErrorTourProps.tsx` is fixed by the same change.** It uses
  the identical `<Tour hasBackdrop>` composition, and both the defect and the
  fix live in lab's shared `TourHighlight`, not at the call site. It could not
  be re-measured live: the session launcher's default form validates clean, so
  its validation tour never opens on an untouched cluster.
- **Not closed: FR-3526** (off-viewport spotlight targets not scrolled into view
  on step change). Different mechanism, untouched here.
- **Not closed: a tour-open race in our own call sites.** Both tour components
  resolve their targets in a single `requestAnimationFrame` after `isActive`
  flips and give up permanently if `.bai-card-error` has not painted yet, so the
  tour silently does not appear on a slow load. It cost several probe retries to
  reproduce. Pre-existing, out of scope, worth its own issue.
- The patch pins `0.3.0-canary.12db2a1`; the next `@astryxdesign/lab` bump will
  need it re-applied or dropped if upstream lands the fix.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -192,6 +192,7 @@ overrides:
   lodash: ^4.18.0
 
 patchedDependencies:
+  '@astryxdesign/lab@0.3.0-canary.12db2a1': 0e5e0cbe89a52670bf3b7b7324794cf671a257ce501858d58f346a993af51e08
   '@cloudscape-design/board-components@3.0.176': 06f6366f1016801dd543a4b968e088fbcbf34ee1a4473bc09a07e35206597a98
   ansi_up@6.0.6: 99856fc7fa4571c6ef4716cfd2dbf1151a025690c737775e248e8b7fe08a4459
   relay-runtime@20.1.1: cc791a2ea160bc7c87571401d2f1430d50527847c44039d45579a5bfa60effa6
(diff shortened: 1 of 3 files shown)


agatha197 (approved):
LGTM

Issue #8739, opened 2026-08-13
astryx-bui layer: proposal — consolidate host-side BAI gap components into the backend.ai-ui package


nowgnuesLee:
## Question

Proposal: consolidate the host app's `react/src/components/astryx-bui/` gap components into the `backend.ai-ui` (BUI) package — or explicitly decide which components are allowed to live host-side, and by what rule.

## Where

- **Directory**: `react/src/components/astryx-bui/` — 18 components + a shared stylesheet (`astryxBui.css`)
- **Component** (examples): `BAICardAstryx`, `BAIModalAstryx`, `BAIDeleteConfirmModalAstryx`, `BAISkeletonAstryx`, `BAIBadgeCountAstryx`, `BAIListAlertAstryx`, `BAIPopconfirmAstryx`, `BAINameActionCellAstryx`, …
- **Recurs in**: 130 files under `react/src` import from `astryx-bui/` (usage breadth from `grep -rl`, not verified per-site)

## Current behaviour

The app has two homes for BAI-branded components:

1. the `backend.ai-ui` workspace package (BUI) — the component library proper, with its own i18n instance, stories, and tests;
1. `react/src/components/astryx-bui/` — host-app-side components created during the Astryx migration, several of which are same-named twins of BUI components (`BAICardAstryx` vs BUI `BAICard`, `BAIModalAstryx` vs BUI `BAIModal`, `BAIDeleteConfirmModalAstryx` vs BUI `BAIDeleteConfirmModal`, `BAIBadgeCountAstryx` vs BUI `BAIBadge`, `BAIListAlertAstryx` vs BUI `BAIListAlert`).

A call site picks one of the two by import path, and nothing enforces which one is current.

## Why this is being raised

The two-home layout has already produced real divergence: the copy control existed once in BUI's `BAIText` (localized) and once host-side in `astryx-bui/BAICopyableText` (hardcoded English label), and the two rendered side by side on the folder explorer's Path row until FR-3518 retired the host-side copy. Same-named twins invite the same drift for modals, cards, skeletons, and badges, and each new feature has to guess which home is authoritative.

## Proposal

Stated as one option, not a decision:

> Treat BUI as the single home for reusable BAI components. For each `astryx-bui` component, either promote it into `backend.ai-ui` (when it is generic) or fold it into its BUI twin (when one exists), and leave host-side only what genuinely depends on host-app context (router, host i18n, app-shim) — with that rule written down.

## Options considered

[...]

PR #8741, opened 2026-08-13
feat(FR-3533): close BUI twin gaps ahead of the astryx-bui fold

nowgnuesLee:
References #8739 (FR-3533)

First PR of the FR-3533 stack (consolidate `react/src/components/astryx-bui/` into `backend.ai-ui`). Closes the two BUI-twin capability gaps the fold in the next PR depends on:

**`BAIDeleteConfirmModal`**
- New `cannotBeUndoneText` prop (host twin had it; all 3 host call sites pass `t('dialog.warning.CannotBeUndone')`).
- QA-FINDINGS Q-17 warning placement ported from the host twin: with the typed-confirm input present, the warning renders as `Text color="danger"` directly under the input (legacy position); the `Banner status="error"` is kept only for the no-input variant and now renders after `extraContent` instead of before it (it used to sit above the option checkboxes — after the thing it warns about, on `PurgeUsersModal`).
- `needsInput` now also requires a non-empty resolved `confirmText`, so an explicitly empty `confirmText` (target row not yet resolved) cannot arm the gate pre-satisfied.

**`BAICard`**
- `padding` / `width` passthrough to the Astryx `Card`. `use-bai-card.md` already documents `padding` as the supported per-card inset knob; the prop now exists. Needed by `AllocationHistoryStatistics`'s `GraphCard` (`padding={4}`, `width="100%"`).

**Behavior note**: the Q-17 placement applies to every existing `BAIDeleteConfirmModal` call site — input-gated delete modals now show the danger text under the input instead of a trailing banner. This is the layout the QA report asked for and what the host-side variant already shipped on the VFolder pages.

**Verification**
- BUI vitest: 611 passed, 1 skipped (incl. `destructiveConfirmFlow.test.tsx`, the typed-confirm gate contract)

### DIFF
--- a/packages/backend.ai-ui/src/components/BAICard.tsx
+++ b/packages/backend.ai-ui/src/components/BAICard.tsx
@@ -160,6 +160,10 @@ export interface BAICardProps extends Omit<
   /** Callback function triggered when the extra button is clicked */
   onClickExtraButton?: () => void;
   size?: 'default' | 'small';
+  /** Astryx `Card` padding step; overrides the `size`-derived default (6, or 3 for `size="small"`). */
+  padding?: React.ComponentProps<typeof Card>['padding'];
+  /** Astryx `Card` width passthrough. */
+  width?: React.ComponentProps<typeof Card>['width'];
   /** antd's nested/inner card treatment. */
   type?: 'inner';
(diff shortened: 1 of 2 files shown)


nowgnuesLee:
Folded into #8742 — the BUI gap-closing commit is now that PR's first commit (stack restructured per review feedback).

PR #8742, opened 2026-08-13
refactor(FR-3533): fold the astryx-bui twin components onto their BUI originals

nowgnuesLee:
References #8739 (FR-3533)

Folds the seven host-side `astryx-bui` components that had same-named BUI twins onto their `backend.ai-ui` originals, and deletes the host files: `BAIModalAstryx`, `BAIDeleteConfirmModalAstryx`, `BAICardAstryx`, `BAIListAlertAstryx`, `BAINameActionCellAstryx`, `BAISelectionLabel`, `BAIQuestionIconWithTooltipAstryx`. The BUI twins were created as the intended originals during FR-3482 ("This is the BUI original of the component the pilot rebuilt as…"); the host copies were pilot leftovers that had already produced one real divergence bug (FR-3518).

**First commit — BUI twin gap-closing** (was #8741, folded in here): before deleting the host twins, the two capabilities only the host copies had are added to the BUI originals so the fold is lossless.
- `BAIDeleteConfirmModal` gains `cannotBeUndoneText` and the QA-FINDINGS Q-17 warning placement (danger `Text` directly under the typed-confirm input; `Banner` only for the no-input variant, after `extraContent`). This applies to every existing `BAIDeleteConfirmModal` call site — the QA-requested layout the host variant already shipped on the VFolder pages. `needsInput` now also requires a non-empty resolved `confirmText`.
- `BAICard` gains `padding` / `width` passthrough to the Astryx `Card` (the knobs `use-bai-card.md` documents), needed by `AllocationHistoryStatistics`'s `GraphCard`.

**Prop translation applied at call sites** (pilot dialect → frozen antd-v6-shaped BUI surface):

| pilot | BUI |
|---|---|
| `actionLabel` / `cancelLabel` | `okText` / `cancelText` |
| `onAction` | `onOk` |
| `isActionLoading` / `isActionDisabled` | `confirmLoading` / `okButtonProps.disabled` |
| `actionVariant="destructive"` | `okButtonProps={{ danger: true }}` |
| `isLoading` | `loading` |
| `onAfterOpen` | `afterOpenChange` (open-edge guarded) |
| `inputPlaceholder` | `inputProps.placeholder` |
| `style` (modal Layout) | `styles.container` (FolderExplorer min-height) |
| NameActionCell `confirm` (`confirmLabel`/`cancelLabel`) | `popConfirm` (`okText`/`cancelText`) |

Every folded modal call site adds `maskClosable={false}` (BUI defaults to backdrop-dismiss `purpose='info'`; the pilot hardcoded `'form'`) and `footer={null}` where the pilot rendered no footer (BUI generates an OK/Cancel footer by default). `DeleteVFolderModal`'s list alert adds `banner` to keep the section-container rendering.

**BUI change — `Table/BAINameActionCell`**: an action's tooltip now rides the button itself (`title` → Astryx `tooltip`), which keeps a disabled action focusable via `aria-disabled` instead of native `disabled`. This is the FR-3423 contract the host cell carried (disabled-with-reason must stay hover/keyboard reachable); the external `Tooltip` wrapper could not provide it because Astryx `Button` only takes the `aria-disabled` path when it owns the tooltip. The FR-3423 contract test moved to the BUI rendering (menu label folds the reason as "title — reason"; the More button is BUI's localized label).

**Deliberate deltas (residue)**:
- The pilot modal's OK button was promise-driven (`clickAction` spinner). BUI's `onOk` is a plain handler; loading comes from `confirmLoading`, which every call site that had a pending state already passes. No call site relied on reject-keeps-open.
- More-actions glyph: horizontal ellipsis → BUI's vertical ellipsis; overflowed disabled reason renders in the label instead of a `description` slot.
- `BAIListAlert` scrollbar thumb: host `--color-text-secondary`/full radius → BUI `--color-text-disabled`/3px (BUI sheet was already globally loaded; which rule won was import-order dependent before).
- `BAIQuestionIconWithTooltip` (BUI) renders a non-focusable span where the host version was a real `<button>` — a small keyboard-a11y regression at 6 sites; candidate follow-up on the BUI twin.
- e2e safety: `.bai-name-action-cell-actions` / `-title-icon` selectors are still emitted by BUI; the 5 specs querying them are unaffected.

**Verification**
- react vitest: 88 files / 1405 passed; BUI vitest: 611 passed, 1 skipped

**Live probe** (dev server `fr-3533`, logged in against a real cluster; dark then light, `data-theme` asserted):
| Surface | Result |
|---|---|
| Data page (`BAINameActionCell` + `BAICard` + table) | renders; row title is a real link (`to`-branch), hover actions present |
| Share modal (folded `BAIModal`) | title/body render, **no footer** (`footer={null}`), backdrop click does NOT dismiss (`maskClosable={false}`), Escape does |
| Folder-create modal | custom footer intact (Reset/Cancel/Create), `loading` resolved |
| Delete-forever modal (`BAIDeleteConfirmModal`) | typed-confirm gate: OK disabled until exact name typed; placeholder from `inputProps.placeholder`; Q-17 warning is a danger `Text` **directly under the input** (no Banner); danger color = `colorError` `#BE3D3F` (dark) |
| Restore action `popConfirm` | popover shows title/description/취소/확인 per the translation table; confirming restored the folder (full trash round-trip, state restored) |
| Notification drawer (`BAIDrawerAstryx`) + bell badge (`BAIBadgeCountAstryx`) | drawer opens with header/body; badge wrapper present (no dot — no unread) |
| `dangerAction.css` via BUI barrel | `.bai-name-action-cell-danger` computes `#89001A` in light mode (declared `--color-text-red`) |
| Console | zero uncaught page errors in both modes (only pre-existing CSP font notice and the backend's known `Group`/`UserGroup` Relay id warning) |


react/src/components/VFolderNodesV2.tsx
```diff
@@ -45,6 +41,10 @@ import { Text } from '@astryxdesign/core/Text';
 import { Tooltip } from '@astryxdesign/core/Tooltip';
 import {
   BAIAlertIconWithTooltip,
+  BAIModal,
+  BAINameActionCell,
+  type BAINameActionCellAction,
+  BAIQuestionIconWithTooltip,
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed: BUI `BAIQuestionIconWithTooltip` now renders its glyph inside a reset `<button type="button">` (aria-label flattened from `title`), restoring the deleted host twin's keyboard-focusable trigger for all its call sites.

packages/backend.ai-ui/src/components/BAIDeleteConfirmModal.tsx
```diff
@@ -190,6 +191,16 @@ const BAIDeleteConfirmModal: React.FC<BAIDeleteConfirmModalProps> = ({
       ? (extractTextFromNode(items[0]?.label) ?? t('general.button.Delete'))
       : t('general.button.Delete'));
 
+  // An explicitly empty `confirmText` (e.g. the target row is not resolved yet)
+  // must not arm the gate with an already-satisfied empty comparison.
+  const needsInput =
+    !reversible &&
+    (items.length > 1 || requireConfirmInput) &&
+    !!resolvedConfirmText;
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed as suggested: split into `wantsInput` (intent) and `needsInput` (input rendered). When confirmation is wanted but the resolved confirm text is empty, the OK button now stays disabled (`!resolvedConfirmText || typedText !== resolvedConfirmText`) instead of falling through to the item-count check.

packages/backend.ai-ui/src/components/BAICard.tsx
```diff
@@ -160,6 +160,10 @@ export interface BAICardProps extends Omit<
   /** Callback function triggered when the extra button is clicked */
   onClickExtraButton?: () => void;
   size?: 'default' | 'small';
+  /** Astryx `Card` padding step; overrides the `size`-derived default (6, or 3 for `size="small"`). */
+  padding?: React.ComponentProps<typeof Card>['padding'];
```
agatha197, line 167: The new `padding` override doesn't participate in the `--bai-card-inset` contract. The full-bleed tab strip reads its inset from `BAICard.css`, where `--bai-card-inset` is derived from `--spacing-6` (or `--spacing-3` via `.bai-card--compact`) — i.e. keyed on `size`, not on the actual padding step. So a call site that combines a custom `padding` with `tabList` gets a silently misaligned tab rail: the bleed/inset identity (`border + padding == padding token`) the CSS relies on no longer holds.

Today's only consumer (`AllocationHistoryStatistics`'s `GraphCard`, `padding={4}`) has no tabs, so nothing is visibly broken — but the prop surface now allows the combination.

Suggest either:
1. republish the inset from the *resolved* padding (e.g. also emit a `bai-card--pad-{n}` class or set `--bai-card-inset` inline from the resolved step), or
2. at minimum, document the constraint in the prop's JSDoc ("do not combine with `tabList`; the tab strip inset assumes the `size`-derived padding") so the trap is at least written down at the API boundary.

agatha197 (changes requested):
One latent trap found in the BUI gap-closing commit — everything else in the fold checked out (verified the `onOk` self-closing at all folded call sites, the `title` → Astryx `tooltip` path in `BAIButton`, and that the kept `.bai-name-action-cell-danger` rule is still loaded and consumed). Details inline.

packages/backend.ai-ui/src/components/BAICard.tsx
nowgnuesLee: Good catch — confirmed the trap is real: with an explicit `padding={4}` and `tabList`, the strip's bleed (`margin-inline: -23px` from the size-derived inset) would hang 8px outside the card's borders on each side, since Astryx `Card` doesn't clip overflow. Fixed in 627e457c5 with your option 1: when `padding` is explicit, `--bai-card-inset` is republished inline from the resolved step (`max(0px, calc(var(--spacing-{step}) - var(--border-width)))`, clamped for `padding={0}`; fractional steps map to `--spacing-0-5` / `--spacing-1-5`, which `astryx.css` ships). Inline wins over both class declarations, so the `size`-derived defaults and all existing call sites are untouched, and a consumer-passed `style` still takes precedence. The prop's JSDoc and the BAICard.css contract comment now document the republish. `verify.sh` ALL PASS, BAICard vitest 59/59.

nowgnuesLee: Follow-up in 7cb67eb47: simplified the fix one level deeper. Astryx `Card` already reflects its resolved padding step as `--container-padding-inline-start` on the card element (both explicit-step and theme-default paths), so `--bai-card-inset` now derives from that var in a single CSS rule — the inline republish from 627e457c5 and the step-enumerating `.bai-card` / `.bai-card--compact` declarations are gone, and the identity holds for every padding step with one owner. `verify.sh` ALL PASS, BAICard vitest 59/59.

agatha197 (approved):
LGTM

PR #8747, opened 2026-08-13
fix(FR-3534): namespace app-compiled StyleX atomics away from Astryx's precompiled classes

### LINKED ISSUE #8746: Production build only: compiled StyleX is emitted unlayered, promoting Astryx atomics above project CSS (session-list card tabs lose their square bottom corners)
## Component

- **Name**: StyleX compiler wiring (`react/vite.config.ts` -> `@stylexjs/unplugin`), cascade-layer bootstrap (`react/src/index.css` `@layer` order statement)
- **Astryx primitive**: compiled `xstyle` / `stylex.create` output vs Astryx's precompiled atomics in `@layer astryx-base`
- **Scope**: app-wide; the visible instance is `BAITabList type="card"` (`packages/backend.ai-ui/src/components/BAITabList.css`)

## Where it is reachable

- Sessions list (`/session`) tab strip, production build only. The dev server renders it correctly.

## Steps to reproduce

1. `pnpm run build`, serve `build/web`, log in, open `/session`.
1. Compare the tab strip with `pnpm dev` at the same commit.

## Observed

Inactive card tabs keep rounded bottom corners in the build, so the accent rail shows through as notches under each tab.

Measured on the non-selected tab — same element, same class list in both:

| |border-bottom-left/right-radius|
|---|---|
|dev|`0px` (correct)|
|build|`8px`|

## Mechanism (measured)

StyleX derives an atomic class name by hashing the declaration, so app code authoring `borderRadius: 'var(--radius-element)'` compiles to `.xh6dtrn` — the same class Astryx's precompiled `astryx.css` already ships inside `@layer astryx-base`. With `useCSSLayers: false` the unplugin appends that atomic to the entry stylesheet UNLAYERED and carrying StyleX's `:not(#\#)` priority ladder, i.e. at ID-level specificity outside every layer.

That outranks `.bai-tab-list--card .astryx-tab` (unlayered, specificity 0-2-0), which only ever won because Astryx's copy was confined to `astryx-base`. Net effect: an Astryx default is silently promoted above the project CSS written to override it.

In the built entry stylesheet the class appears three times — twice inside `@layer astryx-base` (Astryx's own) and once unlayered at offset 272318 (the appended StyleX block). Deleting only that third rule and reloading made the tab strip byte-identical to dev (md5 `614ea7ba…`), which is what identified it.

It reproduces in the build only because the dev server serves just the rules from modules it has transformed so far (14 atomics measured) whereas the build emits all of them (37). `.xh6dtrn` comes from a lazy route — the StyleX probe page (`react/src/pages/AstryxStylexProbePage.tsx:70`) — that a normal dev session never loads.

**17 of the 37 atomics in the appended block collide with class names that already exist in Astryx's layered CSS**, so the tab strip is the visible instance, not the only exposure.

## Expected

Compiled StyleX participates in the cascade layers rather than escaping them, so an Astryx default can never outrank project CSS merely because the app happened to author the same declaration.

## Environment

- **Mode**: both (mechanism is mode-independent)
- **Build**: branch off main @ 9c00663c8 (2026-08-13)
- Vite 6.4.3, `@stylexjs/unplugin` 0.19.0, `@astryxdesign/core` 0.3.0

## Notes

Fix direction, both taken from upstream's `astryxStylex()` preset (`astryx/packages/build/src/vite.ts`):

1. `useCSSLayers: true` — the re-emitted atomic lands in `@layer priority1..N` instead of outside every layer. `xstyle` still beats `astryx-base`, because nothing names the priority layers in the order statement and an unnamed layer registers last.
1. Move the `@layer` order statement to an inline `<style>` at the top of the project-root `index.html`. It currently lives only in `react/src/index.css` and `backend.ai-ui.css`, which reach the parser through the JS module graph, so their parse position differs between dev and build; layer precedence is fixed by first appearance and no later statement can reorder it.

Upstream never hits this because its preset compiles Astryx from source with `stylexPrefix: 'astryx'` (`libraryPrefix`), so library and product atomics can never share a class name. Adopting that would mean switching this repo from Astryx's precompiled `dist` to source consumption — a larger, separate migration worth its own ticket.

Related but distinct: FR-3532 (dev-side `@layer` order / brand theme). Verified by A/B on the fix branch that the two do NOT overlap — FR-3532's tab-count-badge symptom (dev `rgb(0,116,226)` vs build `rgb(255,122,0)`) is unchanged by the fix above, and removing the new `index.html` statement does not reproduce it either.

JIRA Issue: FR-3534

yomybaby:
Resolves #8746 (FR-3534)

## Problem

The same commit renders differently under `pnpm dev` and a production build. Two visible instances:

- **Card tab strips** (session list, Folders): inactive tabs keep rounded *bottom* corners in the build, so the accent rail shows through as notches. Dev is correct (`0px`).
- **Folders "Active" tab count badge**: dev renders the theme's info blue `rgb(109, 156, 254)`; the build repaints it accent orange `rgb(190, 94, 6)`.

## Mechanism

StyleX derives an atomic class name from the declaration's content hash, and both compilers feeding this app — Astryx's precompiled `astryx.css` and our `@stylexjs/unplugin` pass — use the default `x` prefix. Any app-authored declaration identical to a library one therefore compiles to the very class Astryx components already wear: `.xh6dtrn` (`border-radius: var(--radius-element)`) and `.x1ewilqj` (`background-color: var(--color-accent)`), both re-emitted by the lazy StyleX probe page.

The app's copies are appended to the entry stylesheet **unlayered**, and unlayered CSS outranks every named layer. Those copies therefore override both kinds of intended winner at once — the unlayered project CSS (`.bai-tab-list--card .astryx-tab`, lower specificity than the `:not(#\#)` ladder) *and* the theme's component overrides (`.astryx-badge.info` in `@layer astryx-theme`) — production-wide. Dev diverges only because the dev server injects each module's rules when the module loads, and the colliding module is a lazy route a normal session never visits.

17 of the 37 atomics in the appended block collide with classes already present in Astryx's layered CSS (measured on the pre-fix build), so the two instances above are the visible sample, not the extent.

## Fix

One option — compile the app's StyleX under its own class-name prefix:

```ts
classNamePrefix: 'webui',
```

App atoms become `.webui…` while Astryx components keep wearing `.x…`, so the collision — and every hijack it enabled, layered or unlayered victim alike — is impossible by construction. This is the same principle upstream `@astryxdesign/build` states for its own split compilation: distinct library/product prefixes "so library and product atoms never collide".

The cascade-layer rework from the first two commits (`useCSSLayers: true`, HTML-owned layer order statement, new verify guards) is **reverted**: moving the colliding copies into `@layer priority1..N` rescued the unlayered-CSS victims but still left the copies above `@layer astryx-theme`, so the badge-class symptoms survived that approach. The prefix closes both classes with a one-option diff. The layer work's orthogonal benefits (generated CSS no longer outranking hand-written unlayered CSS, specificity hygiene, the verify.sh guard hardening) are real but separable — deferred to their own issue.

## Measured (production build, dark mode, /data)

| | pre-fix build | this PR's build | dev (reference) |
|---|---|---|---|
| Badge `info` background | `rgb(190, 94, 6)` — theme lost | `rgb(109, 156, 254)` | `rgb(109, 156, 254)` |
| Card tab `border-radius` | bottom corners 8px (notch) | `8px 8px 0 0` | `8px 8px 0 0` |

Static check on the entry stylesheet: the appended StyleX block now carries `webui`-prefixed atoms only (44), and no unlayered rule on a collision class remains anywhere in the bundle.

## Residue

- Upstream `@astryxdesign/build`'s Vite integration documents its library/product layer split as a **dev-endpoint** mechanism (`/virtual:stylex.css`); production Vite builds are outside its current envelope. Adopting it (with the Vite 8 upgrade it requires) was evaluated live and abandoned during this fix — the production bundle placed all 1,500 library atoms above `@layer astryx-theme`.
- Rollup can emit more than one `assets/index-*.css` (a lazy chunk entered through an `index.ts` shares the name), and `cssInjectionTarget` takes the first match. The reverted guard commit had hardened this; re-introducing that guard is part of the deferred layer-work issue.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (after the pivot, pre-build)
- `pnpm --prefix ./react run build` → OK; the live measurements above were taken on the served bundle
- An independent verification pass is being run by the reviewer

**Checklist:**

- [ ] Documentation
- [x] Minimum requirements to check during review: on a production build, the Folders "Active" badge must be the theme blue and card tabs must have square bottom corners — identical to dev
- [x] Test case(s) to demonstrate the difference of before/after: measured table above (automated regression guard deferred with the layer-work issue)

### DIFF
--- a/index.html
+++ b/index.html
@@ -13,15 +13,6 @@
        policy. The web server substitutes {{nonce}} per request (empty in dev). -->
   <meta property="csp-nonce" nonce="{{nonce}}">
   <base href="/">
-  <!-- Cascade-layer order. This is the AUTHORITATIVE copy and it must stay the
-       first stylesheet the parser sees: a layer named later than a statement
-       can never be reordered, and every other copy lives in a JS-imported CSS
-       module whose parse position differs between dev and build. Compiled
-       StyleX (`xstyle`) lands in `@layer priority1..N`, which is named here by
-       nothing and therefore registers last — above `astryx-base`. -->
-  <style nonce="{{nonce}}">
-    @layer reset, theme, base, astryx-base, astryx-theme, components, utilities;
-  </style>
   <link rel="icon" href="manifest/favicon.ico">
   <link rel="apple-touch-icon" sizes="256x256" href="manifest/backend-ai.iconset/icon_256x256@2x.png">
   <link rel="icon" type="image/png" sizes="32x32" href="manifest/backend-ai.iconset/icon_32x32@1x.png">
--- a/index.html
+++ b/index.html
@@ -13,6 +13,15 @@
        policy. The web server substitutes {{nonce}} per request (empty in dev). -->
   <meta property="csp-nonce" nonce="{{nonce}}">
   <base href="/">
+  <!-- Cascade-layer order. This is the AUTHORITATIVE copy and it must stay the
+       first stylesheet the parser sees: a layer named later than a statement
+       can never be reordered, and every other copy lives in a JS-imported CSS
+       module whose parse position differs between dev and build. Compiled
+       StyleX (`xstyle`) lands in `@layer priority1..N`, which is named here by
+       nothing and therefore registers last — above `astryx-base`. -->
+  <style nonce="{{nonce}}">
+    @layer reset, theme, base, astryx-base, astryx-theme, components, utilities;
+  </style>
   <link rel="icon" href="manifest/favicon.ico">
   <link rel="apple-touch-icon" sizes="256x256" href="manifest/backend-ai.iconset/icon_256x256@2x.png">
   <link rel="icon" type="image/png" sizes="32x32" href="manifest/backend-ai.iconset/icon_32x32@1x.png">
--- a/packages/backend.ai-ui/src/styles/backend.ai-ui.css
+++ b/packages/backend.ai-ui/src/styles/backend.ai-ui.css
@@ -30,13 +30,14 @@
  * That relative order only holds if the FULL order is declared before any
  * layered rule appears, because layer precedence is fixed by first appearance
  * and later `@layer` statements can only append names, never reorder ones
- * already seen. Since `src/index.ts` imports this file first, BUI's CSS is
- * frequently the first layered sheet in the bundle — so the statement lives
- * here as well as in the app.
+ * already seen. The app declares it authoritatively in the project-root
+ * `index.html`; this copy is what covers a consumer loading BUI's `dist`
+ * without that HTML.
  *
- * !! KEEP IN SYNC with `react/src/index.css`, which declares the identical
- * !! list. Two identical statements are idempotent (the second is a no-op);
- * !! two DIFFERENT ones silently give you whichever loads first.
+ * !! KEEP IN SYNC with the project-root `index.html` and `react/src/index.css`,
+ * !! which declare the identical list. Two identical statements are idempotent
+ * !! (the second is a no-op); two DIFFERENT ones silently give you whichever
+ * !! loads first.
  */
 @layer reset, theme, base, astryx-base, astryx-theme, components, utilities;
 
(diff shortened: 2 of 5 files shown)


nowgnuesLee (approved):
LGTM

PR #8752, opened 2026-08-13
fix(FR-3510): stop row selection from flashing the fetch-key button's loading state

### LINKED ISSUE #8684: Data page: selecting a table row puts the fetch-key refresh button into its loading state
## Component

- **Name**: `AutoUpdateFetchKeyButton` / `BAIFetchKeyButton` as wired on `VFolderNodeListPage` (`react/src/pages/VFolderNodeListPage.tsx:503`)
- **Astryx primitive**: Button (via BUI wrappers)
- **Scope**: this page only / Unknown — other list pages with both row selection and a fetch-key button were not checked

## Where it is reachable

- `/data` (menu: Data) — the "Folders" card's toolbar, the refresh (fetch-key) button on the right end, next to the trash and expand controls.

## Steps to reproduce

1. Open the Data page (Folders card, "활성" tab).
1. Select a row with its checkbox — do nothing else (no refresh click, no filter change).
1. Watch the refresh button in the toolbar.

## Observed

Selecting a row puts the refresh button into its loading state, even though no refetch was requested by the user. The selection itself works (the "1개 선택됨" toolbar appears).

## Expected

Row selection is a purely client-side state change and should not affect the refresh button. The button's loading state should reflect an in-flight refetch only — as it did before the migration.

## Environment

- **Mode**: dark (as reported); light not checked
- **Browser/OS**: Chrome / macOS
- **Build**: `main` @ 2026-08-11

## Impact

Workaround exists (the state is cosmetic and clears) / misleading — the UI signals a fetch that the user did not initiate, and during it the button is not clickable.

## Evidence

- Attached: `folder-list-row-selected.png` — one row selected, toolbar showing "1개 선택됨".

## Notes

- The button's `loading` prop on this page is driven by deferred-value comparisons (`deferredQueryVariables !== queryVariables || deferredFetchKey !== fetchKey`), so any render where those diverge shows as loading. Recorded as an observation of the wiring, not a root-cause claim.

JIRA Issue: FR-3510

### LINKED ISSUE #8893: Admin/project-admin pages: React Compiler bailouts flash the fetch-key button's loading state on row selection
Sibling of FR-3510 (same symptom, different bailout causes). A compiler sweep over all 48 pages plus the 93 risky components (files using useDeferredValue / FetchKeyButton / react-hooks suppressions) found admin and project-admin surfaces where the React Compiler bails out (or is never enabled), so `queryVariables` loses memoization and any unrelated re-render (e.g. row selection) flashes the `deferred !== current` loading state through BAIFetchKeyButton's 700ms anti-flicker hold.

Confirmed instances and causes:

- **AdminVFolderNodeListPage** (`/admin/data`) — render-phase ref write (`queryMapRef.current[...] = ...`), a `react-hooks/refs` violation -> `CompileError: Cannot access refs during render`. Fix: move the write into a `useEffect`.
- **ProjectAdminDataPage** — identical render-phase ref write. Same fix.
- **AdminComputeSessionListPage** (`/admin/session`) — inline count-filter strings with `\"` escapes inside the `graphql` tag; a tagged template whose cooked text differs from raw makes the compiler bail (`Todo: Handle tagged template where cooked value is different from raw value`). Fix: pass the count filters as query variables (mirrors `filterForActiveCount` precedent in AdminVFolderNodeListPage).
- **AdminUserCredentialList**, **ContainerRegistryList** — no `'use memo'` directive at all, so never compiled; both compile cleanly once annotated. Fix: add the directive.

All five fixed with compiler `CompileSuccess` verified per file; `@ts-ignore` badge-count suppressions in the touched files replaced with `keyof` narrowing.

Out of scope here, recorded for follow-up triage (bailouts found by the same sweep but with no deferred-comparison flash surface, or outside admin scope): AdminDeploymentPresetSettingPage (try/finally), AssignRoleModal, ProjectAdminSettingModal, UserSettingModal, FolderExplorerModal/V2, LoginSession (various try/optional-chaining bailouts, deferred surface present — worth checking), DashboardPage, ReservoirArtifactDetailPage, MyResourceWithinResourceGroup, TotalResourceWithinResourceGroup, PendingSessionNodeList, AdminUserCredentialList-style missing directives outside admin scope.

JIRA Issue: FR-3594

ironAiken2:
Resolves #8684 (FR-3510)
Resolves #8893 (FR-3594)

## Mechanism

The bug is a chain of three facts, none sufficient alone:

1. **A `react-hooks/exhaustive-deps` suppression made the React Compiler skip the whole component.** `babel-plugin-react-compiler` (annotation mode) refuses to compile a function containing a React ESLint suppression — verified by compiling `VFolderNodeListPage.tsx` with the compiler and inspecting output: **no memo cache was emitted** (`CompileError: Suppression` pointing at the suppression line; after the fix: `CompileSuccess`, memo cache present).
2. Without compiler memoization, `queryVariables` (a plain object literal) got a **new identity on every render**. Any unrelated state update — row selection via `setSelectedFolderList` — made `useDeferredValue(queryVariables)` return the previous render's object for one urgent render pass, so `deferredQueryVariables !== queryVariables` flashed `true`.
3. `BAIFetchKeyButton` holds its loading icon for a **minimum of 700ms** (anti-flicker), amplifying that single render pass into a visible, unclickable loading state.

The fix replaces the suppression with the project's `useEffectEvent` idiom (`use-effect-event.md`, same shape as `DeploymentAuditLogTab.tsx`), which restores compiler memoization. No query, variable, or refetch semantics change.

## Measured before/after (live dev server, headless Chromium, 50–60ms polling for 1.5s after clicking a row checkbox)

| Probe | Before | After |
|---|---|---|
| Light — refresh button loading after row select | **loading seen**, ~483–920ms window | **none** (0/11 samples) |
| Dark (`data-theme` asserted `'dark'`) — same | **loading seen**, ~332–807ms window | **none** (0/19 samples) |
| Selection toolbar ("1 selected") | works | works |
| Manual refresh click still shows loading | — | **yes**, 171–1154ms busy window (legit loading intact) |
| `pageerror` count (both modes, both passes) | 0 | 0 |

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup, StyleX sentinel, Astryx theme build, Terminology)
- `pnpm --prefix ./react exec vitest run` → 88 files, **1405 passed**
- `pnpm --filter backend.ai-ui exec vitest run` → 46 passed | 1 skipped, **611 passed | 1 skipped** (BUI untouched; run for completeness)
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 2 findings in `BAIText` (`var(--x)`, `var(--font-family-mono)`) — **pre-existing**: identical findings on current `main` (e82347f3e), unrelated to this diff

## Sibling fixes folded in (FR-3594)

Originally stacked as #8895; folded here per review preference. Same symptom, three different React Compiler bailout causes.

The chain above also applies when the React Compiler skips a component → `queryVariables` (a plain object literal) gets a new identity every render → any unrelated state update (row selection) makes `useDeferredValue` return the previous object for one urgent render pass → `deferredQueryVariables !== queryVariables` flashes `true` → `BAIFetchKeyButton`'s 700ms anti-flicker hold amplifies it into a visible, unclickable loading state.

A compiler sweep over all 48 pages plus the 93 risky components (files using `useDeferredValue` / `FetchKeyButton` / react-hooks suppressions) found the admin/project-admin surfaces where that chain is live, with three distinct bailout causes:

| File | Bailout cause | Fix |
|---|---|---|
| `AdminVFolderNodeListPage` (/admin/data) | render-phase `queryMapRef.current[…] = …` write → `CompileError: Cannot access refs during render` | move the write into a `useEffect` (the tab-change handler only reads the map on user interaction) |
| `ProjectAdminDataPage` | **two defects**: (1) identical render-phase ref write; (2) even with compiler memoization restored, `combinedFilter` was rebuilt through `filterOutEmpty([...])` / a switch returning fresh literals — the compiler keeps calls with fresh **mutable** arguments un-memoized, so the object filter (and `queryVariables`) changed identity every render. String-filter pages are immune because the memo guard compares strings by value. | (1) same; (2) hoist usage-mode filters to module constants and build `combinedFilter` from literals only, which the compiler memoizes |
| `AdminComputeSessionListPage` (/admin/session) | inline count-filter strings need `\"` escapes in the `graphql` tag; cooked ≠ raw tagged template is a compiler Todo bailout | pass count filters as query variables (the `filterForActiveCount` precedent) |
| `AdminUserCredentialList` | no `'use memo'` directive — compiler never ran | add the directive (compiles cleanly) |
| `ContainerRegistryList` | no `'use memo'` directive — compiler never ran | add the directive (compiles cleanly) |

Also replaced the `@ts-ignore` badge-count suppressions in the touched files with `keyof` narrowing, per the FR-3510 review direction of removing suppressions rather than adding them.

### Verification (sibling fixes)

- Per-file annotation-mode compile check (`babel-plugin-react-compiler`): all five files now report `CompileSuccess` (before: `CompileError`/never compiled)
- **Live probe** (headless Chromium against the dev server, 50ms sampling for 1.5s after clicking a row checkbox):

  | Probe | Before | After |
  |---|---|---|
  | `/project/default/admin/data` — refresh button loading after row select | **8/20 samples, 0–599ms window** | **0/22 samples** |
  | `/admin-data` — same | (fixed by commit 1) | **0/20 samples** |
  | Selection toolbar appears | works | works |
  | Manual refresh click still shows loading | — | yes, 0–592ms window (legit loading intact) |
  | `pageerror` count | 0 | 0 |
- `bash scripts/verify.sh` → `=== ALL PASS ===` (the earlier Relay drift was the regenerated `AdminComputeSessionListPageQuery` artifact, committed here)
- `pnpm --prefix ./react exec vitest run` → 88 files, **1405 passed** (0 failed)

### Residue (recorded in FR-3594 for follow-up triage)

The same sweep found bailouts **without** a deferred-comparison flash surface (no user-visible bug known, memoization only): `AdminDeploymentPresetSettingPage` (try/finally), `DeploymentAddRevisionModal`, `ChatCard`, and several others. And bailouts **with** a deferred surface but outside the admin scope of this issue: `AssignRoleModal`, `ProjectAdminSettingModal`, `UserSettingModal`, `FolderExplorerModal`/`V2`, `LoginSession`, plus missing-directive files `DashboardPage`, `ReservoirArtifactDetailPage`, `MyResourceWithinResourceGroup`, `TotalResourceWithinResourceGroup`, `PendingSessionNodeList`. Those need per-file restructuring (try/finally, optional-chaining-in-try, TSAsExpression keys) and are listed in the Jira issue.


agatha197 (changes requested):
Please check admin pages too. (ex. data, session page)

yomybaby (approved):
LGTM

PR #8753, opened 2026-08-13
fix(FR-3530): fire Form.Provider onFormChange for unnamed forms

### LINKED ISSUE #8733: Session Launcher: form item changes are not synced to the URL (formValues query param)
## Component

- **Name**: Session Launcher form → URL sync (`react/src/pages/SessionLauncherPage.tsx` — the debounced `formValues` query-param sync, `syncFormToURLWithDebounce`)
- **Astryx primitive**: Unknown (page-level form behavior; the form controls are Astryx-based)
- **Scope**: this page only

## Where it is reachable

- `/session/start` (menu: Sessions → Start)

## Steps to reproduce

1. Open `/session/start`.
1. Change any form item (session name, environment, resources, …).
1. Watch the address bar / the `formValues` query parameter.

## Observed

Changing form items does not update the URL — the `formValues` query parameter does not reflect the change (reported as "form 아이템을 바꾸는데 URL이랑 싱크가 안 된다").

## Expected

Shortly after each form change (the sync is debounced), the URL's `formValues` query parameter reflects the current form state, so reloading or sharing the URL restores the same form — as this page behaved before the migration.

## Environment

- **Mode**: Unknown
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown (reported 2026-08-13, post-FR-3482 migration)

## Impact

Workaround exists — launching the session itself is unaffected, but reload/share loses the entered form state.

## Evidence

- none

## Notes

The sync path exists in code (`syncFormToURLWithDebounce` at `SessionLauncherPage.tsx:328`, invoked from the form change handler at `:620`); whether it stopped firing or stopped writing is not diagnosed here.

JIRA Issue: FR-3530

ironAiken2:
Resolves #8733 (FR-3530)

## Mechanism

The self-hosted form engine's `FormProvider` (`packages/backend.ai-ui/src/form-engine/Form.tsx`) guarded `triggerFormChange` / `triggerFormFinish` with `onFormChange && formName`, so a `<Form>` without a `name` prop never reached `Form.Provider onFormChange`. rc-field-form — the engine's oracle — fires the callback unconditionally, passing `name` as `undefined` for unnamed forms.

SessionLauncherPage's debounced `formValues` URL sync (`syncFormToURLWithDebounce`) rides `Form.Provider onFormChange` on a **nameless** `<Form form={form}>`, so the guard silently killed the sync after the antd → engine switch. Acceptance test 25 was written for this exact page but its fixture used `name="launcher"`, which is why the suite stayed green.

Fix: drop the name guard (both channels, for parity), widen the callback types to `name: string | undefined`, and pin the unnamed case with acceptance test **25b**.

## Before / after

| Probe | Before (main) | After (this branch) |
|---|---|---|
| Acceptance test 25b (unnamed form → `onFormChange`) | **fails** (`expected "vi.fn()" to be called at least once`) | passes (30/30) |
| `/session/start`: type session name | URL unchanged (reported symptom) | `"sessionName":"fr3530-probe"` appears in `formValues` after the 500ms debounce |
| Switch session type to Batch | URL unchanged | `"sessionType":"batch"` + batch subtree appear immediately |
| Reload the synced URL | entered state lost | session name, Batch type, and batch config all restored |

Live probe on the shared test backend (`10.82.0.130:8090`), non-destructively (no session launched). The fix is form-event plumbing with no theme/mode interaction; light/dark parity is unaffected by construction.

reload restores the synced form — session name, Batch type, batch config

## Verification

- `pnpm --prefix ./react exec vitest run` → 88 files, **1406 passed** (includes the new 25b)
- `pnpm --filter backend.ai-ui build` → OK; `pnpm --filter backend.ai-ui exec vitest run` → **611 passed**, 1 skipped
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 2 findings in `BAIText` (**pre-existing on main**, verified by stashing this diff and re-running; untouched by this PR)

## Residue

- The related report FR-3499 (image environment selector) is already Done and not touched here.
- While probing, the e2e login helper (`test-util.ts login()`) could not find the `Endpoint` field after clicking `Advanced` on the login screen, and does not tolerate an already-authenticated storage state — pre-existing harness friction, not addressed in this PR.

### DIFF
--- a/packages/backend.ai-ui/src/form-engine/Form.tsx
+++ b/packages/backend.ai-ui/src/form-engine/Form.tsx
@@ -292,11 +292,11 @@ const Form = React.forwardRef(InternalForm) as <Values = any>(
 export interface FormProviderProps {
   validateMessages?: ValidateMessages;
   onFormChange?: (
-    name: string,
+    name: string | undefined,
     info: { changedFields: FieldData[]; forms: Record<string, FormInstance> },
   ) => void;
   onFormFinish?: (
-    name: string,
+    name: string | undefined,
     info: { values: Store; forms: Record<string, FormInstance> },
   ) => void;
   children?: React.ReactNode;
@@ -323,16 +323,14 @@ export const FormProvider: React.FC<FormProviderProps> = ({
       ...parent.validateMessages,
       ...validateMessages,
     },
+    // Unnamed forms fire too, with `name` undefined — rc-field-form parity.
+    // Guarding on the name silenced SessionLauncherPage's URL sync (FR-3530).
     triggerFormChange: (formName, changedFields) => {
-      if (onFormChange && formName) {
-        onFormChange(formName, { changedFields, forms: formsRef.current });
-      }
+      onFormChange?.(formName, { changedFields, forms: formsRef.current });
       parent.triggerFormChange(formName, changedFields);
     },
     triggerFormFinish: (formName, values) => {
-      if (onFormFinish && formName) {
-        onFormFinish(formName, { values, forms: formsRef.current });
-      }
+      onFormFinish?.(formName, { values, forms: formsRef.current });
       parent.triggerFormFinish(formName, values);
     },
     registerForm: (formName, formInstance) => {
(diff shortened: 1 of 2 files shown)


agatha197 (approved):
LGTM

PR #8754, opened 2026-08-13
fix(FR-3526): scroll off-viewport tour spotlight targets into view on step change

### LINKED ISSUE #8720: Astryx lab Tour: off-viewport spotlight targets are not scrolled into view on step change
## Component

- **Name**: Astryx lab `Tour` / `TourStep` (`@astryxdesign/lab`), via `react/src/components/AdminDeploymentPresetValidationTour.tsx`
- **Astryx primitive**: lab `Tour`
- **Scope**: this tour (as reported) / other lab Tour usages Unknown

## Where it is reachable

- `/admin/deployments/deployment-presets/new?step=review` (Admin → Deployments → Deployment Presets → create, Review step) — the validation tour
- Also used in: `SessionLauncherErrorTourProps.tsx` (usage breadth only)

## Steps to reproduce

1. Open the deployment preset create page, Review step, with an empty form on a

short viewport so at least one failing card is below the fold.

1. Trigger validation so the error tour starts.
1. Advance the tour to a step whose spotlight target is outside the viewport.

## Observed

The spotlight anchors to the off-viewport element, but the page does not

scroll; the user sees a spotlight pointing at nothing until they scroll

manually.

## Expected

Each step's spotlight target is scrolled into view automatically when the step

activates — as the pre-migration antd Tour did (`scrollIntoViewOptions`

defaults).

## Environment

- **Mode**: light (screenshot); dark Unknown
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: dev, 2026-08-12 (Korean UI; observed on the FR-3520 branch dev server)

## Impact

Workaround exists (scroll manually), but the tour's guidance is lost for

off-screen steps.

## Evidence

- Screen recording: none; screenshot `/tmp/clipimg/clip-20260812-141550.png` shows the tour context

## Notes

The migrated tour appears to scroll only the first target on mount ("antd

scrollIntoViewOptions parity" comment in the file); per-step behaviour is

unverified — hedged observation, not a diagnosis.

JIRA Issue: FR-3526

ironAiken2:
Resolves #8720 (FR-3526)

## Mechanism

The lab `TourStep` (`@astryxdesign/lab`) draws its spotlight at the target's *current* box and re-measures on scroll/resize, but never scrolls the target on screen when the step activates. Both migrated validation tours only compensated with a one-shot `scrollIntoView` on the **first** target at mount, so advancing to a step whose target sat below the fold left the user staring at a dimmed page with the spotlight (and callout) rendered off-viewport.

The lab package publicly exports `useTour()` with `activeStepIndex`, so the fix is an app-side wrapper — no lab patch needed:

- **`react/src/components/astryx-bui/BAITourAstryx.tsx`** (new): wraps lab `Tour`; a render-null child observes `useTour().activeStepIndex` and, when a step activates with its target outside the viewport, calls `target.scrollIntoView({ block: 'center' })`. Skips the scroll when the target is already fully visible — the same if-needed semantics antd's rc-tour applied with its `scrollIntoViewOptions` default (`isInViewPort` check + `scrollIntoView`).
- **`AdminDeploymentPresetValidationTour.tsx`** / **`SessionLauncherErrorTourProps.tsx`**: render `BAITourAstryx` with `scrollTargets` (one entry per rendered step, nulls filtered in step order). The wrapper also covers step 0, so each file's one-shot mount scroll is removed.

## Measured before/after

Live probe on the dev server (`/admin/deployments/deployment-presets/new?step=review`, empty form, viewport 1280×600 so the step-navigation bar starts below the fold at `top: 940`):

| Step (target) | Before — nav rect | After — nav rect |
|---|---|---|
| 1 (error card) | card `top:174` in viewport; nav `top:940` off-viewport | same (no scroll needed) |
| 2 (card `extra`) | nav `top:940` off-viewport | same (target already visible → no scroll, if-needed) |
| **3 (step navigation)** | **nav stays `top:940` — spotlight points off-screen** | **page scrolls, nav at `top:540` (centered, in viewport)** |

Dark mode (entered via the header toggle, `document.documentElement.dataset.theme === 'dark'` asserted): identical measurements (nav 940 → 540). Zero `pageerror` in all passes (light before/after, dark after).

### Screenshots

| Before — step 3 (light) | After — step 3 (light) |
|---|---|
| before step3 | after step3 |

| After — step 1 (light, unchanged) | After — step 3 (dark) |
|---|---|
| after step1 | after dark step3 |

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Vite warmup / StyleX sentinel / Astryx theme build / Terminology)
- `pnpm --prefix ./react exec vitest run` → 88 files, 1405 passed
- `pnpm --filter backend.ai-ui exec vitest run` → 611 passed, 1 skipped
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → exits 1 with the same 2 pre-existing `BAIText.css`/`BAIText.tsx` findings as `main` (verified by running the gate on `main`); this change adds no CSS/tokens

## Residue

- The session-launcher validation tour (`SessionLauncherErrorTourProps.tsx`) gets the identical wrapper + `scrollTargets` conversion but was verified statically only (verify.sh + vitest); the live probe exercised the deployment-preset tour. Both tours share the same three-step structure and the same scroller code path.
- The lab `Tour` itself still doesn't scroll; if more lab Tour usages appear they must use `BAITourAstryx` (or the lab component should grow the behavior upstream).

### DIFF
--- a/react/src/components/AdminDeploymentPresetValidationTour.tsx
+++ b/react/src/components/AdminDeploymentPresetValidationTour.tsx
@@ -3,7 +3,8 @@
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
 import { useBAISettingUserState } from '../hooks/useBAISetting';
-import { Tour, TourStep } from '@astryxdesign/lab';
+import BAITourAstryx from './astryx-bui/BAITourAstryx';
+import { TourStep } from '@astryxdesign/lab';
 import React, { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
@@ -67,8 +68,6 @@ const PresetValidationTour: React.FC<PresetValidationTourProps> = ({
         setTargets(null);
         return;
       }
-      // antd `scrollIntoViewOptions` parity: bring the first target on screen.
-      card.scrollIntoView({ block: 'center' });
       setTargets({
         card,
         extra: card.querySelector<HTMLElement>('.bai-card__extra'),
@@ -88,7 +87,16 @@ const PresetValidationTour: React.FC<PresetValidationTourProps> = ({
   };
 
   return (
-    <Tour isActive hasBackdrop isStepCountShown onDismiss={handleDismiss}>
+    <BAITourAstryx
+      isActive
+      hasBackdrop
+      isStepCountShown
+      onDismiss={handleDismiss}
+      // One entry per rendered step, in step order (nulls are skipped below).
+      scrollTargets={[targets.card, targets.extra, targets.nav].filter(
+        (el): el is HTMLElement => el != null,
+      )}
+    >
       <TourStep
         targetRef={{ current: targets.card }}
         heading={t('tourGuide.deploymentPreset.ValidationErrorTitle')}
@@ -111,7 +119,7 @@ const PresetValidationTour: React.FC<PresetValidationTourProps> = ({
           {t('tourGuide.deploymentPreset.FixErrorAndTryAgainText')}
         </TourStep>
       ) : null}
-    </Tour>
+    </BAITourAstryx>
   );
 };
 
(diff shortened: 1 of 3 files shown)


agatha197 (approved):
LGTM

PR #8760, opened 2026-08-13
chore(FR-3538): probe FSEvents health at dev startup and ask before polling fallback

### LINKED ISSUE #8759: Detect broken macOS FSEvents at dev startup and offer polling fallback
pnpm dev silently loses all file watching (Vite HMR, tsc watch, relay watch) when macOS fseventsd refuses new FSEventStream registrations — e.g. after leaked dev processes exhaust its client table. The watcher failure is invisible: the dev server starts normally, serves stale transforms even on full refresh, and only a restart picks up edits.

## Why This Work is Needed

Diagnosing this today took a full layer-by-layer investigation (chokidar -> libuv -> raw FSEventStreamStart). Developers hitting it just see 'HMR is broken' with zero signal. The failure mode recurs on watcher-heavy machines, so dev startup should detect it and say so.

## Expected Outcomes

- pnpm dev probes FSEvents health at startup (macOS only) via a reusable scripts/fsevents-health.mjs (also runnable standalone as a diagnostic)
- When broken, it prints the diagnosis + heal steps and ASKS whether to fall back to stat-polling for that run (never silently enables polling); the opt-in covers Vite, nodemon/relay, and tsc watch
- Non-interactive sessions get the warning and continue unchanged

JIRA Issue: FR-3538

yomybaby:
Resolves #8759 (FR-3538)

## Summary

- `scripts/fsevents-health.mjs` (new): macOS-only FSEvents health probe — watches a fresh temp dir and requires a real event within 1.5s, catching both failure modes fseventsd exhibits when its client table is exhausted (async `EMFILE` on directory watches, and silently-dead streams). Hard no-op reporting healthy on every other platform (Linux/CI safe). Also runs standalone: `node scripts/fsevents-health.mjs` (exit 0 healthy / 1 broken). `BAI_FSEVENTS_HEALTH=broken|ok` overrides the verdict for testing the prompt flow (macOS only).
- `scripts/dev.mjs`: runs the probe before spawning watchers (darwin only — it runs **even when `VITE_WATCH_USE_POLLING` is set**, since that flag covers only Vite and the tsc/relay watchers would otherwise stay dead undiagnosed). When FSEvents is broken it prints the diagnosis + heal steps (kill leaked watchers → `watchman shutdown-server` → `sudo pkill fseventsd` → reboot) and **asks** before enabling stat-polling for that run — never silently, since polling costs steady CPU and hides the machine-level problem.
- `scripts/fsevents-health.test.ts` (new): pins the probe's platform contract — off darwin it reports healthy and `BAI_FSEVENTS_HEALTH` cannot make a CI run look broken; on darwin the override decides.

### Every way out of the prompt

| Input | Result |
|---|---|
| `y` | polling enabled for this run, message states exactly what it covers |
| `n` | continue unchanged |
| **Ctrl+C** | **startup cancelled** — exit 130, no dev tree spawned |
| **Ctrl+D** / closed stdin | continue unchanged ("No") |
| **90s with no answer** | continue unchanged ("No"), and says so on stdout |
| non-TTY, or `CI` set | warning only, no prompt |

readline holds the TTY in raw mode, so both Ctrl+C and Ctrl+D reach the interface rather than the process and each needs wiring: `rl.on('SIGINT')` exits, and the question's signal is `AbortSignal.any([close, AbortSignal.timeout(90_000)])`. The deadline is a liveness backstop for an unattended pty, not a decision deadline — it takes the safe default (No, behaviour unchanged) and prints that it did, rather than deciding silently. 90s leaves room to read the four heal steps printed above the prompt.

### What a `y` actually covers

Polling is not a universal substitute for FSEvents, so the prompt and the messages only claim what they deliver:

| Subsystem | Covered by `y`? | How |
|---|---|---|
| Vite HMR | ✅ | `VITE_WATCH_USE_POLLING=1` (`react/vite.config.ts`) |
| tsc watch — **files** | ✅ | `TSC_WATCHFILE=DynamicPriorityPolling` — **unless** `TSC_WATCHFILE` is explicitly pinned, in which case the pin wins and the confirmation message says so instead of claiming tsc is covered |
| tsc watch — **directories** | ❌ | macOS hardcodes `fsSupportsRecursiveFsWatch`, so `watchDirectory()` short-circuits to recursive `fs.watch`; `TSC_WATCHDIRECTORY` is only read on the unreachable non-recursive path and would be dead config here |
| `relay:watch` schema files (nodemon) | ✅ | `CHOKIDAR_USEPOLLING=1` |
| `relay-compiler --watch` sources | ❌ | requires Watchman (`WatchmanFileSource`; watch mode refuses to start without it) and has no polling file source — heal instructions only |

That is why healing the machine, not polling, is presented as the real fix.

## Why

Diagnosed live (FR-3538): a machine where leaked dev processes (3× Playwright test-server + a relay LSP, days old) exhausted fseventsd's client table — `FSEventStreamStart()` refused every new stream, so each freshly-started `pnpm dev` had a permanently silent watcher: HMR dead, full refresh serving stale transforms, tsc/relay watch blind, until a dev-server restart. Killing the four leaked processes instantly restored FSEvents (differential confirmation); the same machine re-exhausted within minutes, so startup must detect the condition every run and tell the developer instead of failing silently.

## Tests

- `pnpm exec vitest run scripts/fsevents-health.test.ts` → 2 passed, 1 skipped (the darwin-only healthy-probe case runs on macOS; the non-darwin case is the one that runs on CI)
- Every row of the "way out of the prompt" table above driven under a **real pty** (`pty.openpty()` + a truncated copy of `dev.mjs` that stops at the spawn point) with the probe forced BROKEN:
  - Ctrl+C → `cancelled at the FSEvents prompt — dev server not started`, **exit 130**, spawn point never reached
  - Ctrl+D → `continuing WITHOUT polling`, spawn point reached
  - unanswered (harness deadline shortened to 2s) → `no answer in 2s — taking the default (No).`, then `continuing WITHOUT polling`, spawn point reached
  - `y` → child env carries `VITE_WATCH_USE_POLLING=1` / `CHOKIDAR_USEPOLLING=1` / `TSC_WATCHFILE=DynamicPriorityPolling`
  - `y` with `TSC_WATCHFILE=UseFsEvents` → pin preserved in the child env, and the message now reads `tsc keeps its pinned TSC_WATCHFILE=UseFsEvents` instead of claiming tsc file watching is covered
  - `n` → `continuing WITHOUT polling`
  - `VITE_WATCH_USE_POLLING=1` + `y` → prompt narrows to the tsc watcher, polling extended
- `BAI_FSEVENTS_HEALTH=broken` + non-TTY → warning + "continuing without polling", no prompt
- `BAI_FSEVENTS_HEALTH=ok` → zero `[dev]` output, startup identical to before
- `node scripts/fsevents-health.mjs` on this (healthy) machine → `FSEvents OK`, exit 0

## Verification

## Review notes

- The whole feature is inert unless `process.platform === 'darwin'` **and** the probe fails; `BAI_FSEVENTS_HEALTH=broken node scripts/dev.mjs` is how to see it without breaking your machine.
- To check the Ctrl+C fix by hand: `BAI_FSEVENTS_HEALTH=broken pnpm run dev`, then Ctrl+C at the prompt — the shell should come straight back (exit 130) with no Portless/tsc/Vite children started.
- The messages are the review surface: each branch of the `y` / `n` / non-TTY text should claim only the subsystems the coverage table above says are covered.

- [x] Documentation — `DEV_ENVIRONMENT.md` troubleshooting entry
- [ ] Minium required manager version — n/a, dev tooling only
- [x] Specific setting for review — `BAI_FSEVENTS_HEALTH=broken pnpm run dev` forces the broken verdict without breaking your machine
- [x] Minimum requirements to check during review — macOS; on Linux the whole block is skipped
- [x] Test case(s) to demonstrate the difference of before/after — see **Tests** above (Ctrl+C now exits 130 instead of starting the dev tree; a pinned `TSC_WATCHFILE` is no longer reported as covered)


scripts/dev.mjs
```diff
@@ -7,9 +7,69 @@
...
 const env = { ...process.env };
 
+// fseventsd can refuse every new FSEventStream registration (client-table
+// exhaustion by leaked dev processes — see scripts/fsevents-health.mjs). Every
+// watcher this script spawns then starts permanently silent: HMR dead, stale
+// transforms on refresh, tsc/relay watch blind. Probe up front and ASK before
+// falling back to stat-polling — polling costs steady CPU and hides the real
+// problem, so it is never enabled silently. Skipped when the user already
+// opted into polling via VITE_WATCH_USE_POLLING.
+if (process.platform === 'darwin' && !env.VITE_WATCH_USE_POLLING?.trim()) {
```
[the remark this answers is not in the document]
yomybaby: Fixed in 1cfa8d6 — the probe now runs on darwin regardless of `VITE_WATCH_USE_POLLING` (that flag indeed covers only Vite's chokidar). When Vite already polls, the warning/prompt target just the uncovered subsystems: the TTY prompt offers to extend polling to the tsc/relay watchers, and the non-TTY warning names the exact env vars (`CHOKIDAR_USEPOLLING=1`, `TSC_WATCHFILE=DynamicPriorityPolling`) to cover them explicitly. Verified both new paths (non-TTY warning, TTY `y` → "polling fallback enabled") alongside the original five scenarios.

scripts/dev.mjs
```diff
@@ -7,9 +7,82 @@
...
+      try {
+        answer = await rl.question(
+          viteCovered
+            ? '[dev] Vite already stat-polls (VITE_WATCH_USE_POLLING), but the tsc/relay\n' +
+                '[dev] watchers still rely on FSEvents. Extend polling to them for THIS run? [y/N] '
+            : '[dev] Fall back to stat-polling for THIS run (works, but steady CPU cost\n' +
+                '[dev] and the machine stays broken for every other tool)? [y/N] ',
+        );
+      } catch {
+        // Ctrl+D / stdin closed mid-question — treat as "No".
```
agatha197: **`rl.question()` never settles on Ctrl+D — this hangs `pnpm dev` forever.**

The `catch` here says Ctrl+D is treated as "No", and the test plan lists `TTY, answer n / Ctrl+D -> "continuing WITHOUT polling", no crash (AbortError handled)`. That is not what actually happens.

`readline/promises`' `rl.question()` called **without a `signal`** never settles — neither on stdin EOF nor on `rl.close()`. No `AbortError` is thrown, so this `catch` never runs and the `await` on line 46 blocks forever. A TTY Ctrl+D on an empty line calls `rl.close()`, so it takes exactly this path: the dev server never starts.

Reproduced on Node v24.15.0 — both `input.end()` (EOF) and an explicit `rl.close()` with a pending question left the promise pending until a 2s watchdog fired; neither `resolve` nor `reject` was ever called.

Fix: wire an `AbortSignal` to the interface's `close` event so the existing `catch` actually runs.

```suggestion
      const ac = new AbortController();
      rl.once('close', () => ac.abort());
      let answer = '';
      try {
        answer = await rl.question(
          viteCovered
            ? '[dev] Vite already stat-polls (VITE_WATCH_USE_POLLING), but the tsc/relay\n' +
                '[dev] watchers still rely on FSEvents. Extend polling to them for THIS run? [y/N] '
            : '[dev] Fall back to stat-polling for THIS run (works, but steady CPU cost\n' +
                '[dev] and the machine stays broken for every other tool)? [y/N] ',
          { signal: ac.signal },
        );
      } catch {
        // Ctrl+D / stdin closed mid-question — treat as "No".
```

scripts/dev.mjs
```diff
@@ -7,9 +7,82 @@
...
+        '[dev] event stream, so file watching (Vite HMR, tsc watch, relay watch)\n' +
+        '[dev] will be silently dead. Usual cause: leaked dev processes exhausting\n' +
+        "[dev] fseventsd's client table. To heal the machine:\n" +
+        "[dev]   1. kill stale watchers:  ps aux | grep -E 'test-server|vite|relay'\n" +
+        '[dev]      (also close days-old editor windows), then re-run pnpm dev\n' +
+        '[dev]   2. still broken:        sudo pkill fseventsd   (auto-restarts)\n' +
+        '[dev]   3. last resort:         reboot\n' +
+        '[dev] Health check any time:    node scripts/fsevents-health.mjs\n',
+    );
+    if (process.stdin.isTTY) {
```
agatha197: **The prompt has no timeout and no CI guard — an unattended TTY blocks startup indefinitely.**

`process.stdin.isTTY` is not the same as "a human is watching". Automation that runs `pnpm dev` under a pty (tmux, and this repo's own `dev-server` skill) has a TTY but nobody to answer, so on a broken machine startup waits forever. Combined with the Ctrl+D issue above, there is currently no path out of this prompt other than typing into it.

Please add a bounded wait and an explicit non-interactive escape:

- honor `process.env.CI` (take the non-TTY branch when set), and
- give the prompt a timeout (e.g. 15s) that defaults to "No", using the same `AbortController` suggested above (`setTimeout(() => ac.abort(), 15_000)`).

That keeps the "never enable polling silently" guarantee while ensuring the dev server always eventually starts.

scripts/dev.mjs
```diff
@@ -7,9 +7,82 @@
...
+                '[dev] and the machine stays broken for every other tool)? [y/N] ',
+        );
+      } catch {
+        // Ctrl+D / stdin closed mid-question — treat as "No".
+      } finally {
+        rl.close();
+      }
+      if (/^y(es)?$/i.test(answer.trim())) {
+        env.VITE_WATCH_USE_POLLING = '1'; // Vite chokidar (react/vite.config.ts)
+        env.CHOKIDAR_USEPOLLING = '1'; // nodemon's chokidar (relay watch)
```
agatha197: **`CHOKIDAR_USEPOLLING` does not cover the relay watcher — the PR's "a yes covers the whole dev tree" claim only half holds.**

`relay:watch` is `nodemon --watch data/schema.graphql --watch data/client-directives.graphql --exec 'pnpm run relay --watch'`. Two different watchers are involved:

1. **nodemon** (chokidar `^3.5.2`, so it does read `CHOKIDAR_USEPOLLING`) — but it only watches those **two schema files**.
2. **`relay-compiler --watch`**, which is what actually watches the `.tsx` sources for `graphql` tags — and it uses **Watchman**, not chokidar (`node_modules/relay-compiler/relay-compiler-config-schema.json` carries the Watchman saved-state / SCM-aware clock settings).

Watchman on macOS is itself FSEvents-based, so with a broken `fseventsd` the relay compiler's change detection stays dead even after the user answers `y`. The comment on this line ("nodemon's chokidar (relay watch)") reads as if relay is covered; it isn't.

Suggested: correct the wording here and in the PR body, and add `watchman shutdown-server` to the heal steps in the warning above (and in `DEV_ENVIRONMENT.md`), since restarting Watchman is the relay-side counterpart to `sudo pkill fseventsd`.

scripts/dev.mjs
```diff
@@ -7,9 +7,82 @@
...
+        if (!env.TSC_WATCHFILE?.trim()) {
+          env.TSC_WATCHFILE = 'DynamicPriorityPolling'; // tsc watch children
+        }
```
agatha197: **Only `TSC_WATCHFILE` is set — tsc's *directory* watchers stay on FSEvents.**

TypeScript reads `TSC_WATCHFILE` and `TSC_WATCHDIRECTORY` as two independent env vars (`createSystemWatchFunctions` in `typescript.js` takes `tscWatchFile: process.env.TSC_WATCHFILE` and `tscWatchDirectory: process.env.TSC_WATCHDIRECTORY`). On macOS `fsSupportsRecursiveFsWatch` is `true`, so directory watching goes through recursive `fs.watch`, i.e. FSEvents.

So after the fallback is enabled, tsc still misses **newly created and deleted files** (the directory-watch path), which is a common edit during development. `DynamicPriorityPolling` is a valid directory-watch strategy too, so it can be set the same guarded way:

```suggestion
        if (!env.TSC_WATCHFILE?.trim()) {
          env.TSC_WATCHFILE = 'DynamicPriorityPolling'; // tsc watch children
        }
        if (!env.TSC_WATCHDIRECTORY?.trim()) {
          env.TSC_WATCHDIRECTORY = 'DynamicPriorityPolling'; // tsc directory watchers
        }
```

The non-TTY guidance on lines 76-78 should name `TSC_WATCHDIRECTORY` as well, so a developer copying those env vars gets the same coverage.

(The ordering relative to the existing `UseFsEventsOnParentDirectory` block further down is correct — that block is guarded by `!env.TSC_WATCHFILE?.trim()` and runs after this one, so the polling value is not overwritten.)

scripts/dev.mjs
```diff
@@ -7,9 +7,82 @@
...
 
 const env = { ...process.env };
 
+// fseventsd can refuse every new FSEventStream registration (client-table
+// exhaustion by leaked dev processes — see scripts/fsevents-health.mjs). Every
+// watcher this script spawns then starts permanently silent: HMR dead, stale
+// transforms on refresh, tsc/relay watch blind. Probe up front and ASK before
+// falling back to stat-polling — polling costs steady CPU and hides the real
+// problem, so it is never enabled silently. VITE_WATCH_USE_POLLING covers
+// ONLY Vite (react/vite.config.ts), so the probe still runs with it set —
```
agatha197, line 31: **The PR description still documents the pre-`1cfa8d6` behavior and now contradicts this comment.**

This comment correctly states that the probe **still runs** when `VITE_WATCH_USE_POLLING` is set. The PR body still says the opposite:

> `scripts/dev.mjs`: runs the probe before spawning watchers (darwin only, **skipped when `VITE_WATCH_USE_POLLING` is already set**)

That was true before the fix for Copilot's review comment, but not any more. A reviewer reading only the description would be approving behavior that no longer exists. Please update the body (the summary bullet and the corresponding test-plan entry) to match the shipped code — per CLAUDE.md, a changed approach should be reflected in the PR description.

scripts/dev.mjs
```diff
@@ -7,9 +7,82 @@
...
+// the tsc/relay watchers would otherwise stay dead undiagnosed; the prompt
+// then targets just those uncovered subsystems.
+if (process.platform === 'darwin') {
+  const healthy = await probeFsevents();
+  if (!healthy) {
+    const viteCovered = !!env.VITE_WATCH_USE_POLLING?.trim();
```
agatha197: **`viteCovered` uses a looser truth test than the config it is claiming to describe.**

`react/vite.config.ts:889` is `usePolling: process.env.VITE_WATCH_USE_POLLING === '1'` — a strict equality against `'1'`. This line treats any non-empty value as "covered", so `VITE_WATCH_USE_POLLING=0` or `=true` makes the script report "Vite already stat-polls" and offer to extend polling only to tsc/relay, while Vite is in fact **not** polling. The user then declines and gets the message on line 68-69 telling them Vite HMR polls, which is false.

Match the consumer's check:

```suggestion
    const viteCovered = env.VITE_WATCH_USE_POLLING === '1';
```

agatha197 (changes requested):
Solid direction — the failure mode is real, the diagnosis is well-researched, and the Linux/CI no-op keeps it safe for everyone else. Six things before merge; findings 1 and 4 were verified locally against Node v24.15.0 and the installed `typescript` / `nodemon` / `relay-compiler`.

1. **`rl.question()` never settles on Ctrl+D** — `pnpm dev` hangs forever. Reproduced; the test plan currently marks this path as passing.
2. **No prompt timeout and no `CI` guard** — an unattended TTY (tmux, the `dev-server` skill) blocks startup indefinitely.
3. **`CHOKIDAR_USEPOLLING` does not cover relay** — `relay-compiler --watch` uses Watchman (FSEvents-backed), so relay change detection stays dead after a `y`.
4. **`TSC_WATCHDIRECTORY` is missing** — tsc's directory watchers remain on FSEvents, so file creation/deletion is still missed.
5. **The PR description contradicts the code** after `1cfa8d6` (still says the probe is skipped when `VITE_WATCH_USE_POLLING` is set).
6. **`viteCovered` is looser than `vite.config.ts`'s `=== '1'`** check, so `=0` / `=true` produce a wrong diagnosis.

Details inline.

scripts/dev.mjs
yomybaby: Confirmed and fixed in 42f091d7. Thanks — this one was real and I had it marked as passing.

Reproduced independently on Node v24.18.1: `rl.question()` without a `signal` stays pending on both `input.end()` (EOF) and an explicit `rl.close()`. Run in the actual script shape, though, the symptom is not a hang but an exit — Node reports `Detected unsettled top-level await` and terminates with code 13. Same practical outcome you describe: the dev tree never spawns.

Applied your suggestion as written (`AbortController` + `rl.once('close', …)`), then re-verified the three interactive paths end to end under a real pty (`script -qec`) with the probe forced BROKEN:

- **Ctrl+D** (0x04 into the pty) → `continuing WITHOUT polling`, reaches the spawn point, exit 0
- **`y`** → `polling fallback enabled…`, and the child env carries `VITE_WATCH_USE_POLLING=1` / `CHOKIDAR_USEPOLLING=1` / `TSC_WATCHFILE=DynamicPriorityPolling`
- **`n`** → `continuing WITHOUT polling`, reaches the spawn point

The test-plan entry was wrong. The PR body now records that this path did **not** work before 42f091d7, rather than claiming it passed.

scripts/dev.mjs
yomybaby: Taking the `CI` half, declining the timeout.

The `CI` guard is in 42f091d7 — `if (process.stdin.isTTY && !env.CI?.trim())`, verified with `CI=1` under a pty (takes the non-interactive branch even though `isTTY` is true). One condition, removes a whole class of objection; worth having.

The timeout I'd rather not add:

1. **The premise about this repo's tooling doesn't hold.** `.claude/skills/dev-server/SKILL.md` starts the server with the Bash tool's `run_in_background: true`, which does not allocate a pty — that path already lands in the non-TTY branch. And nothing under `.github/workflows/` or `e2e/` runs `pnpm dev` at all, so CI never reaches this prompt (the guard above is belt-and-braces, not a fix for an observed case).
2. **"No path out of this prompt other than typing into it"** was true only because of the Ctrl+D bug you found in the sibling thread. With that fixed, Ctrl+D is a working escape — and it's the one a developer already reaches for reflexively.
3. **A 15s deadline on an interactive question is its own failure mode.** The warning above the prompt is deliberately long (four heal steps); a developer who actually reads it before answering, or steps away for a moment, gets a decision made for them with no indication it happened. tmux is the one remaining TTY case, but a tmux pane is a human's terminal, not an unattended one.

If an unattended-pty caller does turn up, I'd rather add an explicit opt-out env var than a wall-clock guess. Resolving on that basis — happy to reopen if you have a concrete caller I've missed.

scripts/dev.mjs
yomybaby: Correct, and fixed in 42f091d7 — the claim was wrong and the comment was misleading.

Confirmed from the shipped binary rather than the docs: `node_modules/relay-compiler/linux-x64/relay` contains `WatchmanFileSource`, `not implemented: watch-mode (subscribe) is not available for non-watchman file sources.`, and `Cannot run relay in watch mode if 'watchman' is not available (or explicitly disabled).`. So relay watch isn't merely Watchman-*preferring* — there is no polling file source to fall back to at all.

That also settles Copilot's version of the same finding (run one-shot compilation behind a polling watcher, or use a supported polling file source for Relay): the first is a rebuild of `relay:watch` well outside this PR's scope, and the second doesn't exist.

What changed instead:

- the comment is now `// nodemon's chokidar (relay:watch schema files)`
- the `y` message states what it actually covers (Vite HMR + tsc **file** watching) and says plainly that `relay-compiler --watch` cannot be covered
- `watchman shutdown-server` added as step 2 of the heal steps, and to `DEV_ENVIRONMENT.md`
- the PR body no longer says "a yes covers the whole dev tree" — there's a per-subsystem coverage table instead

One nuance I wrote into the messaging: with a Watchman server already watching the repo root, a new relay subscription reuses the existing watch and can survive a broken `fseventsd`. So relay is *at risk*, not certainly dead — which is why it gets heal instructions rather than a warning that overstates the case.

scripts/dev.mjs
yomybaby: The gap you're pointing at is real, but the proposed fix doesn't work on macOS — it would be dead configuration inside a darwin-only block, so I'm not applying it.

`TSC_WATCHDIRECTORY` is read (`typescript.js:8449`), but the only consumer of `tscWatchDirectory` is `updateOptionsForWatchDirectory`, which is called only from `nonRecursiveWatchDirectory` (`:8221`), which is reached only via `hostRecursiveDirectoryWatcher` — constructed only in the `!fsSupportsRecursiveFsWatch` branch. And `watchDirectory()` never gets that far on macOS:

```js
const fsSupportsRecursiveFsWatch = process.platform === "win32" || isMacOs;   // :8431

function watchDirectory(directoryName, callback, recursive, options) {
  if (fsSupportsRecursiveFsWatch) {                                            // :8197
    return fsWatch(directoryName, /*Directory*/ 1, …, recursive, 500, getFallbackOptions(options));
  }
  // ← nonRecursiveWatchDirectory path, unreachable on darwin
}
```

`fsSupportsRecursiveFsWatch` is a plain platform test; `TSC_WATCHDIRECTORY` does not influence it. That matches TS's own documentation of the option — "the strategy for how entire directory trees are watched **under systems that lack recursive file-watching functionality**." Your premise ("on macOS `fsSupportsRecursiveFsWatch` is `true`, so directory watching goes through recursive `fs.watch`") is exactly what makes the remedy a no-op here.

I'd guess the local verification ran on Linux, where `fsSupportsRecursiveFsWatch` is `false` and the variable genuinely does take effect — a darwin-only path checked on a non-darwin box.

Two things that also soften the underlying gap:

- `fsWatch` is called with `getFallbackOptions(options)` (`:123271`), so when `fs.watch` errors — the async `EMFILE` mode this PR describes — TS already falls back to polling that directory at `PriorityPollingInterval`. Only the silently-dead-stream mode is left uncovered, and no `TSC_WATCH*` value fixes that one either.
- Since it can't be fixed by configuration, I documented it: `DEV_ENVIRONMENT.md` and the PR body's coverage table both now state that tsc **directory** watching stays on FSEvents on macOS.

Resolving on that basis — glad to revisit if you can show `TSC_WATCHDIRECTORY` changing behaviour on darwin.

scripts/dev.mjs
yomybaby: Right — fixed. The description now describes the post-1cfa8d6 behaviour (the probe runs on darwin regardless of `VITE_WATCH_USE_POLLING`), and the test plan was rewritten alongside it:

- the Ctrl+D entry no longer claims a pass it never had — it records that the path was broken until 42f091d7
- `CI=1`, `VITE_WATCH_USE_POLLING=true|0|1`, and the `n` path are now listed as separate entries
- a per-subsystem coverage table was added, so "what does answering `y` actually cover" is answerable from the description alone rather than by reading the script

scripts/dev.mjs
yomybaby: Fixed in 42f091d7, exactly as suggested: `const viteCovered = env.VITE_WATCH_USE_POLLING === '1';`, mirroring `react/vite.config.ts:889`.

Verified both ways with the probe forced BROKEN: `VITE_WATCH_USE_POLLING=true` now takes the not-covered messaging, `=1` takes the covered one.

The blast radius was limited to messaging — the `y` branch overwrites the variable with `'1'` regardless — but the `n` / non-TTY text was telling the developer that Vite polls when it doesn't, which is the worst place in this script to be wrong.

yomybaby:
Thanks for the review — it caught a genuine startup-breaking bug. Dispositions below; details are in each thread, and the code changes are in 42f091d7.

| # | Finding | Disposition |
|---|---|---|
| 1 | `rl.question()` never settles on Ctrl+D | **Fixed.** Reproduced on Node v24.18.1; applied the `AbortController` suggestion and re-verified Ctrl+D / `y` / `n` end to end under a real pty |
| 2 | No prompt timeout, no `CI` guard | **Partly taken.** `CI` guard added. Timeout declined — the `dev-server` skill runs `pnpm dev` via background Bash (no pty), no workflow runs `pnpm dev` at all, and with #1 fixed Ctrl+D is a working escape; a 15s deadline would decide for a developer who is reading the heal steps |
| 3 | `CHOKIDAR_USEPOLLING` doesn't cover relay | **Fixed.** Confirmed from the relay binary that watch mode requires Watchman and has no polling file source. Comment, messages, heal steps, `DEV_ENVIRONMENT.md` and the PR body all corrected |
| 4 | `TSC_WATCHDIRECTORY` missing | **Declined, with evidence.** On darwin `fsSupportsRecursiveFsWatch` is a hardcoded platform test and `watchDirectory()` short-circuits to recursive `fs.watch`, so `TSC_WATCHDIRECTORY` is only consulted on a path that is unreachable here — it would be dead config. Documented the limitation instead |
| 5 | PR description contradicts the code | **Fixed.** Description and test plan rewritten; a per-subsystem coverage table added |
| 6 | `viteCovered` looser than `=== '1'` | **Fixed.** Matches `react/vite.config.ts:889` now; verified with `=true` and `=1` |

The one thing worth flagging back: the PR previously promised more coverage than stat-polling can deliver. Two of the five watcher paths (tsc *directory* watching, `relay-compiler --watch`) cannot be redirected to polling on macOS at all, so the script now says so explicitly and leans on the heal steps rather than on the fallback.

scripts/dev.mjs
```diff
@@ -7,9 +7,97 @@
...
+        input: process.stdin,
+        output: process.stdout,
+      });
+      // rl.question() without a signal NEVER settles — not on stdin EOF, not on
+      // rl.close() — so Ctrl+D would strand this await and dev would never start.
+      const ac = new AbortController();
+      rl.once('close', () => ac.abort());
+      let answer = '';
+      try {
+        answer = await rl.question(
```
agatha197, line 80: Ctrl+C at this prompt does not cancel startup: readline is in raw mode with no `SIGINT` listener, so Node closes the interface, the abort path runs, and the dev tree spawns anyway (reproduced — `concurrently` was reached with `TSC_WATCHFILE=UseFsEventsOnParentDirectory`). A silent stdin also blocks startup indefinitely, since nothing but `rl.close()` ever fires `ac`.

An `rl.on('SIGINT', ...)` that exits, plus an `AbortSignal.timeout()` folded into the signal passed here, closes both without changing the Ctrl+D behaviour.

scripts/dev.mjs
```diff
@@ -7,9 +7,97 @@
...
+        rl.close();
+      }
+      if (/^y(es)?$/i.test(answer.trim())) {
+        env.VITE_WATCH_USE_POLLING = '1'; // Vite chokidar (react/vite.config.ts)
+        env.CHOKIDAR_USEPOLLING = '1'; // nodemon's chokidar (relay:watch schema files)
+        if (!env.TSC_WATCHFILE?.trim()) {
+          env.TSC_WATCHFILE = 'DynamicPriorityPolling'; // tsc watch children
+        }
+        console.warn(
+          '[dev] polling fallback enabled for this run (Vite HMR + tsc file watching).\n' +
```
agatha197: When `TSC_WATCHFILE` is explicitly pinned — the escape hatch documented at L125 — the guard at L72 correctly preserves the pin, but this message still claims tsc file watching is covered. Reproduced: with `TSC_WATCHFILE=UseFsEvents` and answering `y`, the spawned child receives `UseFsEvents`, not `DynamicPriorityPolling`.

Worth making the line conditional on whether the pin was applied, so it keeps to the PR's own rule of claiming only what it delivers.

agatha197 (changes requested):
Checked the probe design and the coverage table against the sources: `react/vite.config.ts:889` really is `=== '1'`, nodemon resolves to chokidar 3.6.0 which does honor `CHOKIDAR_USEPOLLING`, the non-darwin path is a hard no-op, and the later `TSC_WATCHFILE` default at L127 does not clobber the one set at L73. Two issues in the prompt/env-plumbing layer, both reproduced by running this file under a pty with the platform forced to darwin.

scripts/dev.mjs
yomybaby: Both halves fixed in 3b2c601b — you were right on the Ctrl+C one, and right enough on the deadline that I stopped arguing.

**Ctrl+C.** Reproduced exactly as described: readline holds the TTY in raw mode, SIGINT went to the interface, nothing listened, `close` fired, the abort path ran, and `concurrently` spawned anyway. `rl.on('SIGINT')` now closes the interface, prints `cancelled at the FSEvents prompt — dev server not started`, and exits 130. Verified under a real pty (`pty.openpty()` driving a copy of `dev.mjs` truncated at the spawn point, probe forced BROKEN): exit 130, spawn point never reached. Ctrl+D is unchanged — still "No", still reaches the spawn point, exit 0.

**Deadline.** Taken, and folded into the same signal as you suggested: `AbortSignal.any([ac.signal, AbortSignal.timeout(PROMPT_TIMEOUT_MS)])`. Two deltas from your sketch, both aimed at the objection I raised last round (a decision made for the developer with no indication it happened):

- 90s, not 15s — the warning above the prompt is four heal steps long and is meant to be read before answering;
- it announces itself: `no answer in 90s — taking the default (No).`

The default it takes is the no-op one, so the cost of hitting the deadline is a `pnpm dev` restart, not a wrong polling config. Verified with the harness deadline shortened to 2s: the line prints, then `continuing WITHOUT polling`, then the spawn point.

The full input → outcome matrix (`y` / `n` / Ctrl+C / Ctrl+D / deadline / non-TTY / `CI`) is now a table in the PR description.

scripts/dev.mjs
yomybaby: Confirmed and fixed in 3b2c601b. Reproduced your case first: `TSC_WATCHFILE=UseFsEvents` + `y` → the child env carries `UseFsEvents`, while the message claimed tsc file watching was covered.

The message is now conditional on whether the pin was applied:

```
[dev] polling fallback enabled for this run (Vite HMR). tsc keeps its pinned
[dev] TSC_WATCHFILE=UseFsEvents — unset it to let the fallback cover
[dev] tsc file watching too.
```

and the unpinned case keeps the original `(Vite HMR + tsc file watching)` wording. The pin itself still wins — that is the documented escape hatch at L125 and the guard is unchanged; only the claim moved. Named the pin's actual value in the line so the developer can see which variable to unset without going looking for it.

Verified under a pty both ways with the probe forced BROKEN: pinned → the message above and `TSC_WATCHFILE=UseFsEvents` in the child env; unpinned → the old message and `TSC_WATCHFILE=DynamicPriorityPolling`. The PR body's coverage table now carries the same caveat on the tsc-files row.

PR #8782, opened 2026-08-14
fix(FR-3486): keep notices above modal backdrops and drop the backdrop blur

### LINKED ISSUE #8636: Notifications and toasts are blurred and inert behind an open modal (Astryx native dialog top-layer ordering)
Notifications, toasts and inline error messages are unreadable and unusable while any modal is open. Reported twice in the [Frontend Teams thread](https://teams.microsoft.com/l/message/19:14c484402d874dafb15806d093b95a82@thread.skype/1786339263879?tenantId=13c6a44d-9b52-4b9e-aa34-0513ee7131f2&groupId=74ae2c4d-ec4d-4fdf-b2c2-f5041d1e8631&parentMessageId=1786326333187&teamName=devops&channelName=Frontend):

- "Modal 이 떠 있는 경우 notification message 가 블러 처리되어서 보이지 않음"
- "Bulk Create Users from CSV 에서 잘못된 csv 파일을 가져와서 에러 메시지가 떠야 하는데, 블러처리 되어 메시지 내용을 확인할 수 없음"

Split out of the `to-astryx` QA pass because it cannot be closed by a QA-sized fix.

## Measured mechanism

Astryx `Dialog` is a **non-portalled native `<dialog>`** promoted with `showModal()`, and its `::backdrop` carries `backdrop-filter: blur(2px)` plus `background-color: rgba(0,0,0,0.45)`. The CSS top layer is ordered **by entry**, so a modal dialog's backdrop paints above the whole normal-flow document **and** above every top-layer element that entered earlier.

Two separate consequences, measured in both modes:

1. `.bai-notification-stack` is a plain `<div>` in normal flow (`popover` attribute null). Its `z-index: 1000` is meaningless against the top layer, so it is blurred and dimmed.
1. Astryx's `ToastViewport` — which `message.error()` reaches through the app-shim — **is** promoted (`popover="manual"`, `:popover-open` true), but it calls `showPopover()` **on mount**, i.e. always before any dialog. Entering the top layer first puts it under the backdrop just the same.

Isolated proof: a synthetic `popover="manual"` shown **before** the dialog is blurred; an identical one shown **after** is crisp. Both are `:popover-open`.

1. Separately, `showModal()` makes everything outside the dialog **inert**. Even a correctly-ordered, perfectly crisp notice has unclickable Close / Retry / Cancel / "View folder" controls.

## Why this is not a QA-row fix

- Removing the blur (`dialog::backdrop { backdrop-filter: none `}) restores legibility and is exact parity with legacy — the antd `BAIModal` passed `mask=` blur: false ``. This is already catalogued as `REGRESSION-CATALOG.md` **O-5**. It is worth doing, but it only addresses symptom 1's readability, not the ordering or the inertness.
- Re-entering the top layer on every dialog open (`hidePopover(); showPopover()` driven by a `dialog[open]` count) is measured to work, but the toast viewport is Astryx-owned, so it needs either a ref out of `BAIAppProvider`, an upstream `isTopLayer` refresh hook, or a `MutationObserver` on `[popover][role="region"]`.
- **Full parity is unreachable by either.** Restoring antd's behaviour (notification z-1010 above the z-1000 mask, fully interactive) requires mounting the notice host **inside** the top-most open dialog, or not using `showModal()` at all. Both are structural changes to `BAIModal` / `NotificationHost`.

## Suggested scope

Decide whether interactive notifications over modals are a requirement. If yes, this is a design + architecture task on the dialog/notification hosting model. If no, stage 1 (drop the backdrop blur, per O-5) plus stage 2 (top-layer re-entry) gets legible, non-interactive notices.

## Evidence

Probes and measured JSON under `.scratch/astryx-migration/qa8/` on the `to-astryx` branch: `probe-modal-toplayer.mjs`, `probe-modal-toplayer-order.mjs`, `before-modal-toplayer.json`, `before-modal-toplayer-order.json`. Pixel comparison: `before-toplayer-notice-crop-nodialog-light.png` (crisp) vs `before-toplayer-notice-crop-light.png` (blurred).

Related: `REGRESSION-CATALOG.md` O-5.

JIRA Issue: FR-3486

nowgnuesLee:
Resolves #8636 (FR-3486)

## Mechanism

The CSS top layer stacks **by entry order**, so a native `<dialog>` promoted with `showModal()` paints its `::backdrop` above the whole document and above every top-layer element that entered earlier:

- `.bai-notification-stack` was a normal-flow `div` — its `z-index: 1000` cannot compete with the top layer at all.
- Astryx's `ToastViewport` (`popover="manual"`) enters the top layer **on mount**, i.e. always before any dialog, so it lands under the backdrop just the same.
- Astryx hardcodes `backdrop-filter: blur(2px)` on the backdrop (compiled StyleX inside `@layer astryx-base`, not tokenized), which is what made the buried notices unreadable.

## Fix

1. **`BAINotificationStackAstryx`** becomes a `popover="manual"` that shows itself on attach, entering the top layer. A `data-bai-top-layer` attribute opts it into re-promotion. CSS fallback (`display: flex` + `z-index: 1000`) keeps it rendered where the Popover API is unavailable.
2. **`BAIAppProvider`'s bridge mount** (BUI app-shim — the owner of the toast viewport, so Storybook surfaces get the fix too) listens for the capturing `toggle` event; when a dialog opens modally (`:modal`), it re-enters both notice surfaces (`hidePopover()` → `showPopover()`) so they stack above the newest backdrop. The toast viewport selector `[role="region"][popover="manual"]` is Astryx 0.4.0 `ToastViewport` internal DOM — no refresh hook is exposed upstream; the coupling is pinned in a comment.
3. **`backend.ai-ui.css`** drops the backdrop blur inside `@layer components`, which outranks `astryx-base` by layer order. The dim mask (`rgba(0,0,0,0.45)`) is kept — exact parity with the antd-era `BAIModal` mask (catalogued as O-5).

## Measured before/after (live dev server, login screen, `modal.confirm` + notification + `message.error`)

| Property | Before (issue-measured) | After (light) | After (dark) |
|---|---|---|---|
| `::backdrop` `backdrop-filter` | `blur(2px)` | `none` | `none` |
| `::backdrop` `background-color` | `rgba(0,0,0,0.45)` | `rgba(0,0,0,0.45)` (unchanged) | same |
| Notification stack | normal-flow, under backdrop | `:popover-open`, **above** backdrop | same |
| Toast viewport (shown before modal) | under backdrop | re-promoted on modal open, **above** backdrop | same |
| Notice buttons after modal closes | — | hit-testable again (`elementFromPoint` ✓) | same |
| `pageerror` | — | 0 (only a pre-existing CSP font warning) | 0 |

## Verification

- `pnpm --filter backend.ai-ui build` → built (dist regenerated)
- react vitest: **88 files / 1405 tests passed**
- BUI vitest: **46 files passed, 1 skipped / 611 passed, 1 skipped**
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 1 finding, **pre-existing on main** (`BAIText.css:28 var(--x)`, FR-3540 leftover — verified identical on a clean `origin/main` checkout, unrelated to this diff)

## Residue

- **Notices above an open modal are still inert** (visible and legible, but Close/Retry/Cancel are unclickable until the modal closes). `showModal()` makes everything outside the dialog inert by spec; lifting that requires changing the dialog/notification hosting model — explicitly left as the team decision FR-3486 names ("interactive notifications over modals"). This PR delivers the issue's stage 1 (blur removal, O-5) + stage 2 (top-layer re-entry).
- The toast viewport and the notification stack both anchor bottom-right and can overlap; pre-existing layout, unchanged here.
- Browsers without the dialog `toggle` event degrade to the status quo (notices below the backdrop), never an error; the app's effective floor (Astryx requires the Popover API + CSS anchor positioning) already implies `toggle` support.

### Light — notification + modal open (notice crisp above the dimmed page)
notices above backdrop, light

### Dark — notification shown before the modal, re-promoted above the backdrop
notice above backdrop, dark

### Dark — `message.error` toast shown before the modal, re-promoted above the backdrop
toast above backdrop, dark

https://claude.ai/code/session_01FysW3SbMGEsgTm4rG9vnhU

### DIFF
--- a/packages/backend.ai-ui/src/app-shim/appShim.test.tsx
+++ b/packages/backend.ai-ui/src/app-shim/appShim.test.tsx
@@ -17,8 +17,8 @@ import { BAIAppProvider } from './index';
 import { message } from './message';
 import { modal } from './modal';
 import type { ShowToastFn, ToastOptions } from '@astryxdesign/core/Toast';
-import { cleanup, render } from '@testing-library/react';
-import { afterEach, describe, expect, it, vi } from 'vitest';
+import { render } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 interface ShownToast {
   options: ToastOptions;
@@ -121,71 +121,80 @@ describe('app-shim message', () => {
 });
 
 describe('app-shim top-layer re-entry (FR-3486)', () => {
-  // jsdom has neither the Popover API nor the `:modal` / `:popover-open`
-  // selectors, so the fixtures carry instance-level stand-ins. The Astryx
-  // toast viewport BAIAppProvider mounts has no stubs, which also exercises
-  // the listener's `typeof showPopover` guard.
-  function mountNotice({ isPopoverOpen }: { isPopoverOpen: boolean }) {
+  // jsdom lacks the Popover API and the `:modal` / `:popover-open` selectors,
+  // so fixtures carry instance-level stand-ins. Never hoist them into a
+  // setupTests prototype polyfill: the listener branches on the method's
+  // ABSENCE, and a global polyfill would make that branch untestable.
+  const fixtures: HTMLElement[] = [];
+
+  function mountNotice({ isPopoverOpen = false, hasPopoverApi = true } = {}) {
     const el = document.createElement('div');
     el.setAttribute('popover', 'manual');
     el.setAttribute('data-bai-top-layer', '');
-    const showPopover = vi.fn();
-    const hidePopover = vi.fn();
-    Object.assign(el, { showPopover, hidePopover });
-    const nativeMatches = el.matches.bind(el);
     el.matches = (selector: string) =>
-      selector === ':popover-open' ? isPopoverOpen : nativeMatches(selector);
+      selector === ':popover-open' && isPopoverOpen;
+    const calls: string[] = [];
+    if (hasPopoverApi) {
+      Object.assign(el, {
+        showPopover: () => calls.push('show'),
+        hidePopover: () => calls.push('hide'),
+      });
+    }
     document.body.appendChild(el);
-    return { showPopover, hidePopover };
+    fixtures.push(el);
+    return calls;
   }
 
-  function dispatchToggle(
-    target: HTMLElement,
-    { isModal = true, newState = 'open' } = {},
-  ) {
-    if (target instanceof HTMLDialogElement) {
-      const nativeMatches = target.matches.bind(target);
-      target.matches = (selector: string) =>
-        selector === ':modal' ? isModal : nativeMatches(selector);
-    }
+  function dispatchToggle({
+    tag = 'dialog',
+    isModal = true,
+    newState = 'open',
+  } = {}) {
+    const target = document.createElement(tag);
+    target.matches = (selector: string) => selector === ':modal' && isModal;
     document.body.appendChild(target);
+    fixtures.push(target);
     const event = new Event('toggle');
     Object.assign(event, { newState });
     target.dispatchEvent(event);
   }
 
+  beforeEach(() => {
+    render(<BAIAppProvider />);
+  });
+
   afterEach(() => {
-    cleanup();
-    document.body.innerHTML = '';
+    fixtures.splice(0).forEach((el) => el.remove());
   });
 
   it('re-enters an open notice surface when a modal dialog opens', () => {
-    render(<BAIAppProvider />);
-    const notice = mountNotice({ isPopoverOpen: true });
-    dispatchToggle(document.createElement('dialog'));
-    expect(notice.hidePopover).toHaveBeenCalledTimes(1);
-    expect(notice.showPopover).toHaveBeenCalledTimes(1);
-    expect(notice.hidePopover.mock.invocationCallOrder[0]).toBeLessThan(
-      notice.showPopover.mock.invocationCallOrder[0],
-    );
+    const calls = mountNotice({ isPopoverOpen: true });
+    dispatchToggle();
+    expect(calls).toEqual(['hide', 'show']);
   });
 
   it('shows a not-yet-promoted notice without hiding it first', () => {
-    render(<BAIAppProvider />);
-    const notice = mountNotice({ isPopoverOpen: false });
-    dispatchToggle(document.createElement('dialog'));
-    expect(notice.hidePopover).not.toHaveBeenCalled();
-    expect(notice.showPopover).toHaveBeenCalledTimes(1);
+    const calls = mountNotice();
+    dispatchToggle();
+    expect(calls).toEqual(['show']);
   });
 
-  it('ignores dialog close, non-modal dialogs, and non-dialog toggles', () => {
-    render(<BAIAppProvider />);
-    const notice = mountNotice({ isPopoverOpen: true });
-    dispatchToggle(document.createElement('dialog'), { newState: 'closed' });
-    dispatchToggle(document.createElement('dialog'), { isModal: false });
-    dispatchToggle(document.createElement('div'));
-    expect(notice.hidePopover).not.toHaveBeenCalled();
-    expect(notice.showPopover).not.toHaveBeenCalled();
+  it('skips a notice without the Popover API instead of throwing', () => {
+    const bare = mountNotice({ isPopoverOpen: true, hasPopoverApi: false });
+    const stubbed = mountNotice({ isPopoverOpen: true });
+    dispatchToggle();
+    expect(bare).toEqual([]);
+    expect(stubbed).toEqual(['hide', 'show']);
+  });
+
+  it.each([
+    { name: 'dialog close', options: { newState: 'closed' } },
+    { name: 'a non-modal dialog', options: { isModal: false } },
+    { name: 'a non-dialog toggle', options: { tag: 'div' } },
+  ])('ignores $name', ({ options }) => {
+    const calls = mountNotice({ isPopoverOpen: true });
+    dispatchToggle(options);
+    expect(calls).toEqual([]);
   });
 });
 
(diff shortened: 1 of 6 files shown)


packages/backend.ai-ui/src/app-shim/index.tsx
```diff
@@ -87,6 +88,38 @@ const BAIAppBridgeMount: React.FC<{
     return () => registerBridge(null);
   }, [showToast]);
 
+  // The CSS top layer stacks by entry order, so a `showModal()` backdrop
+  // paints over notice popovers that entered earlier; re-enter them on each
+  // modal open. Notices stay inert while a modal is open (FR-3486).
+  // `[role="region"][popover="manual"]` is Astryx 0.4.0 ToastViewport's
+  // internal DOM (src/Toast/ToastViewport.tsx) — no refresh hook is exposed.
+  useEventListener(
```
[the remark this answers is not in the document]
nowgnuesLee: Added in 07fc24b98: three jsdom tests in `appShim.test.tsx` now pin the re-entry contract — (1) an already-promoted notice is hidden then shown again when a modal dialog fires `toggle` with `newState: 'open'`, (2) a not-yet-promoted notice gets `showPopover()` without a prior hide, and (3) dialog close events, non-modal dialogs, and non-dialog `toggle` sources are all ignored. jsdom lacks the Popover API and the `:modal`/`:popover-open` selectors, so the fixtures carry instance-level stand-ins; the unstubbed Astryx toast viewport that `BAIAppProvider` mounts doubles as coverage for the `typeof showPopover` guard, which the same commit also made explicit instead of relying on the catch to swallow the missing-API selector error.

packages/backend.ai-ui/src/app-shim/appShim.test.tsx
```diff
@@ -118,6 +120,84 @@ describe('app-shim message', () => {
...
+  // setupTests prototype polyfill: the listener branches on the method's
+  // ABSENCE, and a global polyfill would make that branch untestable.
+  const fixtures: HTMLElement[] = [];
+
+  function mountNotice({ isPopoverOpen = false, hasPopoverApi = true } = {}) {
+    const el = document.createElement('div');
+    el.setAttribute('popover', 'manual');
+    el.setAttribute('data-bai-top-layer', '');
+    el.matches = (selector: string) =>
+      selector === ':popover-open' && isPopoverOpen;
```
agatha197: TS2322 here breaks the `typecheck (backend.ai-ui)` CI job: the DOM lib types `Element.matches` with type-predicate overloads, so a plain `(selector: string) => boolean` is not assignable. Same `Object.assign` pattern as the `showPopover`/`hidePopover` stubs below avoids it:

```suggestion
    Object.assign(el, {
      matches: (selector: string) =>
        selector === ':popover-open' && isPopoverOpen,
    });
```

packages/backend.ai-ui/src/app-shim/appShim.test.tsx
```diff
@@ -118,6 +120,84 @@ describe('app-shim message', () => {
...
+    return calls;
+  }
+
+  function dispatchToggle({
+    tag = 'dialog',
+    isModal = true,
+    newState = 'open',
+  } = {}) {
+    const target = document.createElement(tag);
+    target.matches = (selector: string) => selector === ':modal' && isModal;
```
agatha197: Second TS2322 of the same shape:

```suggestion
    Object.assign(target, {
      matches: (selector: string) => selector === ':modal' && isModal,
    });
```

agatha197 (changes requested):
Verified this implementation end to end on a live dev server (login screen + logged-in main app against the test backend):

- `::backdrop` measures `backdrop-filter: none` with the `rgba(0,0,0,0.45)` dim kept — O-5 parity confirmed.
- The notification stack mounts as `popover="manual"` + `data-bai-top-layer`, `:popover-open` true.
- On `modal.confirm` open, both the notification stack and the Astryx toast viewport re-enter the top layer with exactly one `hidePopover → showPopover` pair (instrumented live), and a toast shown *before* the modal ends up crisp above the backdrop — the exact regression scenario from #8636.
- Local runs: `scripts/verify.sh` ALL PASS, BUI vitest 617 passed, react vitest 1405 passed.

One blocker before this can merge: **the `typecheck (backend.ai-ui)` CI job is red on this PR.** `pnpm --filter backend.ai-ui exec tsc --noEmit` fails with TS2322 at `appShim.test.tsx:134` and `:154` — directly assigning `el.matches = (selector: string) => boolean` is not assignable to the DOM lib's `Element.matches`, whose tag-name overloads are type predicates. `scripts/verify.sh` only typechecks the `react` workspace, which is why it reported ALL PASS locally while CI (per-package tsc) catches it.

The fixtures already use the `Object.assign(el, { showPopover, hidePopover })` pattern two lines below, which sidesteps the assignability check; the inline suggestions apply the same pattern to the two `matches` stubs. Everything else looks correct and within scope.

packages/backend.ai-ui/src/app-shim/appShim.test.tsx
nowgnuesLee: Thank you! I fixed it.

packages/backend.ai-ui/src/app-shim/appShim.test.tsx
nowgnuesLee: Thank you! I fixed it.

agatha197 (changes requested):
Notices now land above the `<dialog>` itself, not just its backdrop, and they stay inert — a bottom-right notice can cover a 90%-width modal's footer controls and cannot be dismissed until the modal closes. Please call that out in the Residue section.

nowgnuesLee:
I'm closing this PR since I'm going to implement the Astryx dialog using a portal instead.

PR #8785, opened 2026-08-14
fix(FR-3546): retry tour target resolution until the anchors exist

### LINKED ISSUE #8784: Deployment preset Review step: the validation tour never appears on first entry (Skip to Review)
## Component

- **Name**: `PresetValidationTour` (`react/src/components/AdminDeploymentPresetValidationTour.tsx`)
- **Astryx primitive**: `Tour` / `TourStep` (`@astryxdesign/lab`), via `BAITourAstryx`
- **Scope**: this page only — the session-launcher tour (`SessionLauncherErrorTourProps.tsx`) is a near-verbatim copy of the same shape, but it was not verified in this report

## Where it is reachable

- `/admin/deployments/deployment-presets/new?step=review` (menu: Admin → Deployments → Deployment Presets → Create)
- The same tour shape also exists in: `react/src/components/SessionLauncherErrorTourProps.tsx` (Sessions → Start)

## Steps to reproduce

1. Log in as an admin.
1. Open Admin → Deployments → Deployment Presets → Create.
1. Leave the form empty. On the first step, click **Skip to Review**.
1. Observe the Review step.
1. Now click **Previous**, then **Next**, to return to the Review step.

## Observed

At step 4 the Review step renders its three validation-error cards correctly, but

the validation tour never appears — no spotlight, no callout, no backdrop. The

page just sits there with the error cards.

At step 5, on the second entry into the Review step, the same tour appears

immediately and works normally.

Measured with a 100ms poll over 4s after each entry (error-card count / tour

callout present):

| Entry | Error cards in DOM | Tour visible |
|---|---|---|

|—|—|—|
|---|---|---|

| 1st, via Skip to Review | 3 | no — for the full 4s |
|---|---|---|

| 2nd, via Previous → Next | 3 | yes — immediately |
|---|---|---|

## Expected

The validation tour appears on the **first** entry into the Review step,

regardless of how the user got there (Skip to Review, Next, or a direct

`?step=review` URL). It already does this on the second entry, and guiding the

user to their validation errors on first arrival is the tour's entire purpose.

## Environment

- **Mode**: both (not mode-specific)
- **Viewport**: 1280×720
- **Build**: `FR-3526` @ `aeb1eccbc`, since merged to `main`

## Impact

Workaround exists — navigate away from the Review step and back — but the user

has no way to know that, so in practice first-time users simply never see the

tour.

## Evidence

- Repro run logs (100ms poll, both paths) captured locally; screenshots

`repro-A-skip.png` / `repro-B-prevnext.png` to be attached in Jira.

## Notes

Suspicion only, from instrumenting the component during the repro: the tour

resolves its DOM targets in a single `requestAnimationFrame` whose effect depends

only on `isActive`, and on first entry that frame appears to run before the review

summary has painted its `.bai-card-error` cards. Mechanism to be confirmed at fix

time.

JIRA Issue: FR-3546

agatha197:
Resolves #8784 (FR-3546)

## Mechanism

Both validation tours resolved their DOM anchors inside a **single**
`requestAnimationFrame`, in an effect that depends only on `isActive`:

```tsx
useEffect(() => {
  const frame = requestAnimationFrame(() => {
    const card = document.getElementsByClassName('bai-card-error')?.[0];
    if (!card) { setTargets(null); return; }   // ← gives up forever
    setTargets({ card, … });
  });
  return () => cancelAnimationFrame(frame);
}, [isActive]);
```

On the **first** entry into the Review step that frame runs before the review
summary has painted — `AdminDeploymentPresetReviewSummary` suspends on a Relay
query (`useAdminImageReference`) — so `.bai-card-error` is not in the DOM yet.
`targets` is set to `null`, and because `isActive` stays `true` the effect never
re-runs. The tour is silently dropped.

Entering the step a **second** time worked only because the Relay cache was warm
by then, so the summary painted in the same commit that flipped the tour on.
That is why "Previous → Next" appeared to fix it.

Instrumented log from the repro, first entry (before this change):

```
[TOURDBG] render open=true isActive=true targets=null
[TOURDBG] rAF isActive=true cardFound=false cardCount=0    ← one shot, missed
```

versus the second entry:

```
[TOURDBG] rAF isActive=true cardFound=true cardCount=3
[TOURDBG] render open=true isActive=true targets=set
```

## Change

- **`react/src/components/astryx-bui/useTourTargets.ts`** (new): resolves a
  tour's anchors once `isActive` turns on and, if they are not there yet,
  re-runs the resolver on **every DOM mutation** until it succeeds, then
  disconnects. A `null` return now means "not ready yet" instead of "give up".
- **`AdminDeploymentPresetValidationTour.tsx`** / **`SessionLauncherErrorTourProps.tsx`**:
  use the hook instead of their own single-shot `rAF`. Both files carried the
  identical defect, so the resolution lives in one place rather than a third copy.

**Why a MutationObserver and not a timeout.** The first version of this fix
retried each frame for 3s. That is a wall-clock guess about how long a Relay
query takes: a slow backend blows through it and the tour is lost again — the
same bug, just with a different threshold. Waiting on the anchor itself has no
such tunable, and it matches the actual intent ("show the tour once the error
cards exist"). The observer only lives while a tour is pending and disconnects
the moment it resolves.

Per `comment-density.md` ("touch it, trim it"), the two migration-travelogue
comment blocks in the files being edited were shortened to the constraint they
still carry.

## Measured before/after

Playwright repro against a dev server, polling `.bai-card-error` count (`c`) and
whether the tour callout is present (`tour`):

| Scenario | Before | After |
|---|---|---|
| 1st entry — **Skip to Review** | `0ms c=3 tour=0` — **stays 0 for the full 4s** | `0ms c=0 tour=0` → `100ms c=3 tour=1` |
| 2nd entry — Previous → Next | `0ms c=3 tour=1` | `0ms c=3 tour=1` (unchanged) |
| 1st entry, **GraphQL delayed 6s** | n/a — would miss any deadline under 6s | `0ms c=0 tour=0` → `6000ms c=3 tour=1` |

The third row is the one that justifies the observer: with all `**/admin/gql`
responses stalled by 6s, the anchors land far past any 3s deadline and the tour
still appears **in the same poll sample the cards do**.

Before the fix the error cards were present in *both* of the first two rows —
the tour was the only thing missing, which rules out "the cards were not
rendered".

## Verification

- `bash scripts/verify.sh` → Relay / Lint / Format / TypeScript / Vite warmup /
  StyleX sentinel / Terminology all **PASS**.
  `Astryx theme build` **FAILs**, but it fails identically on a clean `main`
  checkout with this branch's changes stashed — pre-existing, not from this change.
- `pnpm --prefix ./react exec vitest run` → 88 files, **1405 passed**
- Live repro on the dev server (table above), both the normal and the
  6s-stalled-backend path. The session-launcher tour shares the same code path
  but was verified statically only.

## Residue

- The two tour components remain near-verbatim copies of each other (same
  three-step structure, differing only in identifiers, i18n keys, and the
  second-step anchor). This change removes one copy of the resolution logic;
  merging the components themselves is a separate refactor.

### DIFF
--- a/react/src/components/AdminDeploymentPresetValidationTour.tsx
+++ b/react/src/components/AdminDeploymentPresetValidationTour.tsx
@@ -4,8 +4,9 @@
  */
 import { useBAISettingUserState } from '../hooks/useBAISetting';
 import BAITourAstryx from './astryx-bui/BAITourAstryx';
+import { useTourTargets } from './astryx-bui/useTourTargets';
 import { TourStep } from '@astryxdesign/lab';
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 
 /**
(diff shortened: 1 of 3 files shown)


ironAiken2 (approved):
LGTM

PR #8806, opened 2026-08-15
fix(FR-3554): restore top-center message placement and an opaque dark banner surface

### LINKED ISSUE #8798: Data: consecutive notifications render on top of each other instead of stacking (create folder then immediately delete)
## Where

- **Page**: `/data` (menu: Data) — folder list
- **Region**: the notification / toast overlay stack
- **Component** (if identified): Unknown — the two cards are the folder-created notification and the "Moved the ... folder to the trash bin." message (`data.folders.MovedToTrashBin`, raised from `react/src/components/VFolderNodesV2.tsx`)

## Observed

When a second notification appears while the first one is still on screen, the two cards render on top of each other instead of stacking. In the screenshot the white "Moved the aaa folder to the trash bin." card sits directly over the dark "Folder: aaa" card, covering most of it — only the first card's top edge and part of its title remain readable.

The two cards also do not share a surface style: the one underneath is dark with an orange folder name, the one on top is white.

## Expected

Consecutive notifications stack vertically with spacing between them, each card fully visible (or the earlier one dismissed) — no card drawn over another.

## Environment

- **Mode**: dark (light mode not checked)
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown

## Reproduction

1. Open Data (`/data`).
1. Create a folder (e.g. `aaa`).
1. Immediately, while the "folder created" notification is still visible, delete that folder (move to trash bin).
1. The trash-bin notification appears overlapping the previous one.

## Evidence

- Screenshot: `/home/ubuntu/.claude/jobs/81280346/pasted-1.png` (to be attached in Jira)

## Notes

Same notification surface as FR-3465 (closed notifications reappearing) and FR-3486 (notifications blurred behind an open modal), but a different symptom.

JIRA Issue: FR-3554

agatha197:
Resolves #8798 (FR-3554)

Creating a folder and immediately deleting it put a notification card and a
message toast on screen at once, and the toast painted on top of the card.

The two notices come from **different APIs**, and antd placed them in different
places:

| | API | antd placement | Renders into |
|---|---|---|---|
| "Folder created" | `upsertNotification` | bottom-right | `.bai-notification-stack` |
| "Moved … to the trash bin." | app-shim `message.success` | **top-center** | Astryx `ToastViewport` |

Two fixes, both on those notice surfaces: **(1)** `message` goes back to
top-center, **(2)** the dark info-banner surface — what the notification cards
are drawn on — becomes opaque.

---

# 1. `message` was rendering bottom-right

## Mechanism

The migration mapped `message` onto `ToastViewport` and left it on its default
`bottomEnd` anchor — the corner the notification stack has always occupied — so
two independently-laid-out overlays landed on top of each other. The viewport is
promoted into the browser top layer via `popover="manual"`, so it painted over
the cards whatever z-index the stack carried (raising the stack's z-index would
not have helped).

## The change

`ToastPosition` is four corners (`topEnd | topStart | bottomEnd | bottomStart`)
and `inset` takes px offsets, so a **centered anchor is not expressible through
the component API**. So:

- `DefaultProviders` passes `toast={{ position: 'topStart' }}` as the base;
- `astryxBui.css` centers the viewport — the last-resort scoped-CSS tier,
  because no prop reaches it.

```css
[popover][role='region'] {
  inset-inline-start: 50%;
  transform: translateX(-50%);
  align-items: center;
}
```

**Why that selector is safe:** it matches the toast viewport and nothing else —
the only other `role="region"` in Astryx is `Carousel`, which is not a popover.
Both attributes are part of the component's rendered contract rather than a
hashed StyleX class, and the specificity (0,2,0) beats StyleX's single-class
atomics (0,1,0), so it does not depend on source order or on whether StyleX is
emitted layered (cf. FR-3534).

## Measured

| | Toast centre X | Window centre X | Toast top | Overlaps the stack? |
|---|---|---|---|---|
| **Before** | — (anchored bottom-right, over the stack) | 640 | — | **yes** |
| **After, light** | **640** | 640 | 16px | **no** |
| **After, dark** | **640** | 640 | 16px | **no** |

Notification stack unchanged at bottom-right (`872–1256 × 612–696`).

---

# 2. Dark info-banner surface was translucent

Reported while reviewing the first fix: in dark mode, page content reads
through the notification cards.

## Mechanism

Astryx repaints `.astryx-banner.info` with `--color-background-blue`, and only
that token's **dark** value carries an alpha suffix:

| | `--color-background-blue` | measured banner fill |
|---|---|---|
| Light | `#c4ddfb` | `rgb(196, 221, 251)` — **opaque** |
| Dark | `#9eb7ff`**`3D`** (24%) | `rgba(158, 183, 255, 0.24)` — **76% see-through** |

Light was never affected, which is what marks this as a token asymmetry rather
than an intended translucent surface.

## The change

Override the banner's info background in the theme's `components:` block —
tier 1 of the fix ladder, and the mechanism `@layer astryx-theme` already uses
to outrank `astryx-base`:

```ts
banner: { 'status:info': { backgroundColor: 'light-dark(#c4ddfb, #393f50)' } }
```

`#393f50` **is** the existing fill composited over the dark page (`#9eb7ff` at
24% over `#191919`), so the colour is unchanged to the eye; only the
see-through goes away. Light keeps `#c4ddfb` verbatim.

Scoped to the banner rather than to `--color-background-blue` itself, which
other call sites (e.g. `SessionTemplateModal`'s selected row) may want
translucent.

`THEME_NAME_REV` 15 → 16 with the artifacts regenerated and the stale
`bai-r15-*` files deleted — without the bump a stale registration or stale
built CSS silently masks a recipe change. The live probe confirms the new
registration is the active one:
`document.documentElement` carries `data-astryx-theme="bai-r16-default-brand-hc43dtt"`.

## Measured

| | Notification card fill | Alpha |
|---|---|---|
| Light, before & after | `rgb(196, 221, 251)` | 1 (unchanged) |
| Dark, **before** | `rgba(158, 183, 255, 0.24)` | **0.24** |
| Dark, **after** | `rgb(57, 63, 80)` | **1** |

---

## Screenshots — the reported repro (create folder → immediately move to trash)

**Dark, before the opacity fix.** `message` is already top-center, but the
notification card is see-through: the pagination row and the "Folder created"
line behind it are legible straight through the card.

dark before

**Dark, after.** Same repro, same placement, card now opaque.

dark after

**Light.** Unchanged by the opacity fix; shown for the placement.

light

The toast auto-hides after 4s, shorter than the screenshot round-trip, so each
capture held it open by stretching the page's `setTimeout`. That is a **capture
aid only** — every number above comes from uninstrumented runs.

## Verification

- `pnpm --prefix ./react exec vitest run` → **88 files / 1405 tests passed**
- `pnpm --filter backend.ai-ui exec vitest run` → **47 passed / 1 skipped (613 passed, 1 skipped)**
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 1 undeclared `var()` at `packages/backend.ai-ui/src/components/BAIText.css:28`. **Pre-existing on main**; not in this diff.
- No `backend.ai-ui` source change, so no BUI rebuild is required to consume this.
- Console during the probes: only pre-existing noise (Mixed Content from `client.ts`, `RelayResponseNormalizer` warnings, a Google Fonts block from `react-grab`, `/func/totp` 404). No new `pageerror`.

## Residue

- `success` / `warning` / `error` banners still fall through to Astryx's
  `--color-*-muted`, which is alpha in **both** modes (0x33 light / 0x3F dark).
  Only `info` — the status the notification stack uses — is fixed here.
- The alpha asymmetry is upstream in Astryx's stock token; this overrides it at
  the app's theme layer rather than patching the dependency.
- Does not subsume FR-3486 (notifications blurred/inert behind an open modal) or
  FR-3465 (closed notifications reappearing).
- The centering rule is global by construction: any future Astryx toast viewport
  in this app is centered too. That is the intent — the app has one `message` host.


yomybaby (approved):
LGTM

PR #8809, opened 2026-08-16
fix(FR-3557): make the dark neutral hover overlay translucent so it survives every surface

### LINKED ISSUE #8805: Dark mode: ghost/text button hover is invisible on #262626 surfaces (opaque --color-overlay-hover equals the table row background)
## Component

- **Token**: `--color-overlay-hover` (`ANTD_NEUTRAL_SURFACES`, `react/src/astryx-theme/backendAiTheme.ts:181`)
- **Astryx primitive**: any `Button` / `IconButton` `variant="ghost"` (Astryx paints hover as `background-image: linear-gradient(var(--color-overlay-hover), var(--color-overlay-hover))`)
- **Scope**: app-wide in **dark mode**, wherever the underlying surface is `#262626`

## Where it is reachable

- Sessions -> a session -> **Session Scheduling History** -> hover the expand chevron (the reported surface)
- Any table row ghost action button in dark mode

## Observed

In dark mode the hover highlight does not appear **at all**. Light mode is fine.

Measured live (dark, Session Scheduling History modal):

- hover paints `linear-gradient(rgb(38,38,38), rgb(38,38,38))` = **#262626**
- the table row it paints on: `background-color: rgb(38,38,38)` = **#262626**
- **the overlay colour is byte-identical to the surface** -> contrast ratio **1.00:1**, zero visible change

Token value: `--color-overlay-hover: light-dark(rgba(0,0,0,0.06), #262626)`.

- light half `rgba(0,0,0,0.06)` is **translucent** -> composites over any surface -> works everywhere
- dark half `#262626` is **opaque** -> only reads as a wash on the surface it was measured against

## Root cause

The dark half is an **opaque, pre-resolved** colour. `backendAiTheme.ts` documents this as deliberate: **"audit 1 (catalog G-4) pinned the opaque dark value precisely because a translucent wash is invisible on `#141414`"**. That reasoning holds for the **page** surface (`#141414`) and only for it. Cards, modals and table rows sit at **`#262626`**, where the same opaque value becomes a no-op.

An opaque overlay can only be correct on exactly one surface; Astryx uses this token as a **universal** overlay across all of them.

## Expected

The ghost hover reads as a visible step on every dark surface, as it does in light.

## Suggested fix (needs a decision - contradicts a recorded deliberate choice)

Make the dark half **translucent white** so it composites instead of replacing:

- `rgba(255,255,255,0.08)` over `#141414` -> `#272727`, i.e. it **reproduces the pinned `#262626` within 1/255** on the page surface, preserving audit 1s measured intent
- the same value over `#262626` -> `~#373737`, a clearly visible step

So the translucent form is a strict improvement: identical where the opaque value was correct, and correct everywhere it was not.

Because `backendAiTheme.ts` explicitly says the global token **"must stay as it is"**, this should be confirmed by the team before landing, and it needs the full theme procedure (THEME_NAME_REV bump + `astryx theme build` artifact regeneration + KEEP-IN-SYNC mirrors).

## Environment

- **Mode**: dark only (light unaffected)
- **Build**: `FR-3556` @ 26.9.0-alpha.0.8646, Astryx 0.4.0

## Notes

Found while verifying FR-3556 (expand-chevron hover pill sheared by the cell clip). FR-3556 fixes the **geometry** in both modes; this issue is why the pill is still not **visible** in dark. The two are independent.

`ANTD_REVERSED_BAND_OVERLAYS` (`packages/backend.ai-ui/src/theme-shim/antdParity.ts:180`) already documents this exact trap for the header band and works around it locally with translucent values - the same reasoning generalises.

JIRA Issue: FR-3557

agatha197:
Resolves #8805 (FR-3557)

## Problem

In **dark mode** the ghost/text button hover state does not appear at all. Reported on the Scheduling History modal's collapsible chevron, but it is app-wide.

| Before (dark) | After (dark) | After (light — unchanged) |
|---|---|---|
|  |  |  |

## Mechanism

Astryx paints every hover state as one composited layer — `background-image: linear-gradient(var(--color-overlay-hover), var(--color-overlay-hover))`. So `--color-overlay-hover` is a **universal overlay**, applied over whatever surface the element happens to sit on.

Its dark half was the **pre-resolved opaque `#262626`**. An opaque overlay is only ever correct on the one surface it was resolved against:

- on the `#141414` **page** surface it read as antd's wash — which is why audit 1 (catalog G-4) pinned it;
- on a `#262626` **card / modal / table row** it is *byte-identical to the surface*.

Measured live in the Session Scheduling History modal (dark):

| | value |
|---|---|
| chevron hover paint | `linear-gradient(rgb(38,38,38), rgb(38,38,38))` = `#262626` |
| row background underneath | `rgb(38,38,38)` = `#262626` |
| contrast | **1.00:1 — no visible change** |

## Fix

Dark half becomes **`rgba(255,255,255,0.08)`** — translucent, so it composites instead of replacing.

| surface | before (opaque) | after (0.08 white) |
|---|---|---|
| `#141414` page | `#262626` | **`#272727`** — within **1/255** of the pinned value |
| `#262626` card / modal / row | `#262626` (invisible) | **`~#373737`** — visible step |
| light mode | `rgba(0,0,0,0.06)` | **unchanged** |

So it is strictly better: identical where the opaque value was correct, correct where it was not. Verified live — `resolvedOverPage` computed `#272727` against the previously pinned `#262626`, and the light overlay still reports `rgba(0,0,0,0.06)`.

> This reverses a decision recorded in `backendAiTheme.ts` ("`--color-overlay-hover` must stay as it is globally"). That comment is updated in place and marked **SUPERSEDED (FR-3557)** rather than deleted, with the reason, so the pin is not simply restored later. Audit 1's *measured intent* is preserved — only its opaque encoding changes.

## Blast radius

Every dark-mode ghost/text/menu hover. The FR-3493 (`dropdown-menu`) and FR-3505 (`field`) overrides **spread** this token rather than copying it, so they follow automatically — `backendAiTheme.test.ts` asserts exactly that relationship and it passed unchanged throughout. Only the test's *hardcoded literal* needed updating. The `button` / `tab` component-scope overrides set their own values and are untouched.

## Theme procedure (all steps)

- recipe edited in `react/src/astryx-theme/backendAiTheme.ts`
- **`THEME_NAME_REV` 15 → 16**; artifacts regenerated via `astryx theme build` → `bai-r16-default-brand-h4glqj5` (the hash moved, so this does not collide with any other in-flight r16)
- `built/index.ts` re-export updated; the stale `bai-r15-*` artifacts deleted
- value is a colour, so the `[light, dark]` tuple is legal
- **no KEEP-IN-SYNC mirror update needed** — `index.html`'s critical style and `resources/webui.css`'s `.splash` carry page/splash surfaces, not interaction overlays (grepped: neither references this token)

## Verification

- `bash scripts/verify.sh` — **`=== ALL PASS ===`** (incl. `astryx theme build --check`)
- `pnpm --prefix ./react exec vitest run` — **1405 passed** (88 files)
- `pnpm --filter backend.ai-ui exec vitest run` — **613 passed**, 1 skipped
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` — 1 undeclared `var()` in `packages/backend.ai-ui/src/components/BAIText.css:28`, **pre-existing and not in this diff**
- Live probe, dark + light, on the real app; 0 `pageerror`

## Notes

- Found while verifying #8803 (FR-3556), which fixes the same chevron's **geometry** (hover pill sheared by the cell clip). The two are independent and neither subsumes the other: FR-3556 makes the pill the right *shape*, FR-3557 makes it *visible* in dark. FR-3556 is not required to review or land this.
- The "before (dark)" shot was taken on the FR-3556 branch, so its expand column is the fixed 56px; the variable under test here is the overlay's visibility, not the column width.


agatha197:
Combined into #8803 at the reviewer's request — both fixes touch the same chevron's hover state, so they review and land together. The FR-3557 commit is cherry-picked onto that branch unchanged (`Resolves #8805 (FR-3557)` is on #8803), and the combined branch is re-verified: verify.sh ALL PASS, react 1405 passed, backend.ai-ui 613 passed.

Closing this PR only; FR-3557 stays open in Jira until #8803 lands. Branch `FR-3557` is left in place for reference.

PR #8811, opened 2026-08-16
feat(FR-3558): generate a build-time search index of pages, tabs, and setting items

### LINKED ISSUE #8807: Global search palette (Cmd+K) with a build-time page/tab/setting index
## Summary

## Why

## Design (decided)

## Delivery

JIRA Issue: FR-3558

nowgnuesLee:
Resolves #8807 (FR-3558)

## What

- `react/scripts/build-search-index.mjs` (`pnpm run search-index`) walks `routes.tsx`, type-strips each module with `vite.transformWithEsbuild`, and follows value imports to collect the `t('…')` keys reachable from every route entry.
- Emits `react/src/generated/searchIndex.json` — 34 entries, `?tab=` deep links for 8 pages / 24 tabs, 4 setting pages / 54 items, 1,701 body keys after dropping keys shared by ≥10 entries (140 kB raw, 18.5 kB gzip).
- `scripts/verify.sh` gains a "Search index" drift gate next to the Relay one; `scripts/dev.mjs` runs the build once on start.
- The index is committed and marked generated (`.gitattributes`, `.prettierignore`, `.claude/rules/review-ignored-paths.md`).
- **ADR 0003** — the committed index is the single source: `build:only` does not regenerate it (guarded by `searchIndex.test.ts`), `typecheck.yml` rebuilds it after relay-compiler and fails on drift next to the `__generated__` check, and both that step and `verify.sh` share `scripts/check-generated-drift.sh`. A key the index references but `en.json` lacks fails the build (`KNOWN_MISSING_KEYS` is the explicit allowlist), and esbuild / import-scan failures are counted and warned about. See `docs/adr/0003-committed-search-index-artifact.md`.

## Why

The app is an SPA whose other routes are `React.lazy` chunks — they are not mounted when the palette opens, so their text does not exist in the DOM to search. The index is therefore derived from source at build time, and it stores i18n **keys**, not strings, so all 20 locales work and switching language needs no re-index.

## Verification

- vitest: 4 tests (`searchIndex.test.ts`)
- Two consecutive builds are byte-identical; the drift gate fails when the committed index is stale.

---

## Review fixes

Applied in `fix(FR-3558): keep page-owned keys, index optional-param routes, and harden the extractor` (8f78dfa).

**Blockers**

1. **The ≥10-entry noise filter deleted page-owned vocabulary.** The walk is transitive with no depth limit, so `session.launcher.SharedMemory` reaches 19 entries exactly as `button.Cancel` does — a count-only filter dropped both. Measured before the fix: `session.launcher.{SharedMemory,ClusterMode,Memory}` absent from the session launcher, `data.{Foldername,UsageMode,Host}` absent from Data, `session.SessionName` absent everywhere. Placement now uses the depth already recorded in `keyMap`: an entry **owns** a key when it declares it within `OWNER_DEPTH` (4) import hops of the route component. A key owned by ≥ `NOISE_THRESHOLD` entries is chrome and dropped everywhere; any other shared key survives only on its owners; a key owned by nobody survives where it is shallowest. All seven measured keys are back on their pages, and `button.*`, `time.*`, `errorBoundary.*`, `general.ExtendLoginSession` are gone from all entries.
2. **`chat/:id?` was skipped as parametrised**, so Chat — a first-class sidebar item — had no entry. Optional (`?`-suffixed) params no longer disqualify a route; it is indexed at the path with the optional segment omitted (`/project/:projectName/chat`).

**Should-fix**

3. Tab recognizer accepts `value` as a `key` alias, so radio-strip tabs (`{ label, value }`) get their labels: **26/53 → 41/53** tabs labelled. `TAB_OVERRIDES` was re-checked and **kept** — the Data page's `statusCategory` is `parseAsString` with labels built by `_.map`, so nothing is statically detectable (verified: removing the override leaves the page with 0 tabs). It is now guarded by an assertion instead.
4. Startup assertions for the hardcoded config: every configured path (`CUT_PATH_*`, `EXTERNAL_PREFIXES`, `SHIM_COMPONENT_FILES`, `ROUTE_CHROME_COMPONENTS`, `routes.tsx`, `en.json`) must exist, and every `TAB_OVERRIDES` key must match an indexed route — otherwise the build throws with the stale entry named. All of these previously failed silently on rename.
5. Both can't-fail tests replaced. The route-coverage test compared the extractor's output to the extractor's own route list; it now checks `VALID_MENU_KEYS` against the index, with a documented exclusion list (`summary`/`job` aliases, plugin-rendered `pipeline`) that is itself asserted not to be stale. "Drops keys on every page" was structurally always true; it is now two tests that assert the ownership rule directly — named page keys present on their pages, named chrome keys absent from all.
7. `--check` was dead (never called, and it ignored `--out`). **Deleted**, along with `diffSummary` and the unused `search-index:check` script — `verify.sh`'s rebuild-then-`git status` gate is the real check, mirroring Relay.

**Nits**

- (9) The `component` file-path array is no longer emitted; it is `--verbose` report output only. (10) Dead `fileCount` removed. (11) `revisionTab` removed from `TAB_LIKE_PARAMS` — only reachable from a skipped parametrised route.
- (12) `missingFromEnJson` no longer warns on `t('key', 'Fallback')`, which renders the fallback: **4 warnings → 1**, and that one (`storageHost.permission.SelectionLimitExceeded`, called as `t(key, { options })`) is a genuine miss.
- (15) `objectElements` takes the right operand only for `&&` `BinaryExpression`s, not for `??`/`||`/`+`. (16) `dedupeSettings` keys on key+group, so the same title in two groups no longer collapses.

**Measured effect**

| | before | after |
|---|---|---|
| entries | 34 | **35** (`/project/:projectName/chat` added) |
| distinct body keys | 1,701 | **1,940** |
| key placements | 2,840 | **3,913** |
| keys dropped | 229 (by count) | **39** (chrome, by ownership) |
| tabs labelled | 26/53 | **41/53** |
| raw / gzip | 141 kB / 18.3 kB | **178 kB / 22.8 kB** |

The artifact grows because page-owned vocabulary that was being deleted is now retained; dropping the `component` array offsets part of it.

**Deliberately not applied**

- **(13) `dev.mjs` rewriting the committed index on every dev-server start.** Accepted as-is: the build is ~0.7 s, never blocks startup, and a developer who changes a route wants the palette to match without a manual step. The risk it raises is a merge conflict on the regenerated artifact, which is addressed instead by a new `.claude/rules/search-index-conflicts.md` — on conflict, take either side and re-run `pnpm run search-index`; never hand-merge.
- **(14) Cross-machine determinism.** The index is deterministic for a given source tree (asserted byte-for-byte by `searchIndex.test.ts`), but it is derived by walking the filesystem, so it is only reproducible across machines to the extent that the checkout and the resolved `typescript`/`vite` versions match — both are pinned by the lockfile, and `verify.sh` re-derives it in CI, so a machine-specific build would fail the gate rather than ship.
- **(8) Renaming the script** — skipped as churn.


react/scripts/build-search-index.mjs
```diff
@@ -0,0 +1,1131 @@
...
+assertConfiguredPaths();
+
+// ---------------------------------------------------------------------------
+// workspace deps (zero new dependencies)
+// ---------------------------------------------------------------------------
+const require_ = createRequire(path.join(REACT_DIR, 'package.json'));
+const ts = require_('typescript');
+const { transformWithEsbuild } = await import(require_.resolve('vite'));
+
+const rel = (f) => path.relative(ROOT, f);
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in 33ddf8abc — `rel()` now normalizes to POSIX separators (`path.relative(ROOT, f).split(path.sep).join("/")`), so the CUT/SHIM/chrome/external comparisons and the emitted IDs hold on Windows. The regenerated index is byte-identical on POSIX.

react/package.json
```diff
@@ -99,6 +99,7 @@
     "build:only": "pnpm run relay && NODE_OPTIONS='--max-old-space-size=4096' vite build",
     "test": "vitest run",
     "relay": "relay-compiler",
+    "search-index": "node scripts/build-search-index.mjs",
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in 33ddf8abc — `build:only` now runs `pnpm run search-index` between `relay` and `vite build`, so a production build regenerates the index from the source being built, matching the Relay treatment.

react/src/components/GlobalSearchPalette/searchIndex.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    const settingPages = index.entries.filter((e) => e.settings.length > 0);
+    const settingItems = index.entries.reduce(
+      (a, e) => a + e.settings.length,
+      0,
+    );
+    console.log(
+      `?tab=: ${tabPages.length} pages / ${tabKeys} keys; ` +
+        `tabs labelled: ${allTabs.filter((t) => t.labelKey).length}/${allTabs.length}; ` +
+        `settings: ${settingPages.length} pages / ${settingItems} items`,
+    );
```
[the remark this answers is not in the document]
nowgnuesLee: Removed in 33ddf8abc. One correction for the record: `no-console` is not enforced by this repo's ESLint config (the lint gate passed with the log in place) — but the log was a debug artifact and the values are asserted right below, so it is gone.

agatha197 (changes requested):
please resolve conflicts

react/scripts/build-search-index.mjs
```diff
@@ -0,0 +1,1193 @@
...
+};
+
+/** Unwrap `filterOutEmpty([...])`, `_.compact([...])`, `[...].filter(...)`. */
+function asArrayLiteral(node) {
+  let n = node;
+  for (let i = 0; i < 4 && n; i++) {
+    if (ts.isArrayLiteralExpression(n)) return n;
+    if (ts.isCallExpression(n)) {
+      const arg = n.arguments[0];
+      if (arg) {
```
agatha197, line 333: The doc comment says `[...].filter(...)` is unwrapped, but a call with any argument always descends into `arguments[0]` — for `.filter(fn)` that is the callback, so the `n.expression.expression` branch is never reached and the array comes back `null` (the whole tab/setting list is silently dropped). Check `PropertyAccessExpression` before the argument, or drop that case from the comment.

agatha197 (changes requested):
Ran the extractor, the vitest file, and a byte-identical rebuild locally: all clean. One latent mismatch in the extractor.

yomybaby (dismissed):
### Merge conflict with `main`

This PR currently conflicts with the latest `main` and cannot be merged as is. It is also the base of the FR-3558 stack, so the rest of the stack stays blocked until this one is resolved.

**Conflicting files**
- `package.json` (content — the scripts block)

Please bring in the latest `main` (merge or rebase), resolve the conflicts, and re-run `bash scripts/verify.sh` before pushing. If `pnpm-lock.yaml` also conflicts after the `package.json` resolution, take `main`'s lockfile and re-run `pnpm install` rather than hand-merging it (`.claude/rules/pnpm-lockfile-conflicts.md`).

---
_Generated by [Claude Code](https://claude.ai/code)_

react/scripts/build-search-index.mjs
nowgnuesLee: Fixed in 2f1ace522 — `asArrayLiteral` now tries the call's receiver first (so `[...].filter(fn)` unwraps to the array instead of descending into `fn`) and falls back to `arguments[0]`, which still covers `filterOutEmpty([...])` / `_.compact([...])`. The committed index is byte-identical: no route currently shapes its list that way, so this was latent.

agatha197 (approved):
LGTM

react/scripts/build-search-index.mjs
```diff
@@ -0,0 +1,1186 @@
...
+    return 0;
+  }
+
+  const built = await buildEntries();
+  const index = toIndex(built);
+  const text = serialize(index);
+
+  const missing = missingFromEnJson(index);
+  if (missing.length)
+    console.warn(
```
agatha197, line 1177: This only warns, so a typoed key still builds and shows up as a raw key in the palette. It is 0 today — cheapest moment to make it throw (with an explicit allowlist if false positives appear).

react/scripts/build-search-index.mjs
```diff
@@ -0,0 +1,1186 @@
...
+  let code = src;
+  try {
+    ({ code } = await transformWithEsbuild(src, file, {
+      loader: file.endsWith('x') ? 'tsx' : 'ts',
+      jsx: 'automatic',
+      sourcemap: false,
+      target: 'esnext',
+    }));
+  } catch {
+    /* keep raw source; this file's import list is approximate */
```
agatha197: When esbuild (or `preProcessFile` below) fails on a file, its entire import edge silently disappears and index coverage shrinks unnoticed. Please count these failures and warn when non-zero.

react/scripts/build-search-index.mjs
```diff
@@ -0,0 +1,1186 @@
...
+const NOISE_THRESHOLD = 10;
+
+/**
+ * An entry OWNS a key when it declares it within this many import hops of the
+ * route component. Ownership — not the raw entry count — decides where a
+ * shared key survives: chrome (`time.*`, `button.Cancel`) is owned by many
+ * pages at once, while page vocabulary (`session.launcher.SharedMemory`) is
+ * owned by one or two and merely leaks deep into the rest.
+ */
+const OWNER_DEPTH = 4;
```
agatha197, line 61: Coverage is very uneven under this heuristic: `/admin/session` gets 162 keys but `/project/:projectName/admin/session` only 12 — thin wrapper pages search poorly. Consider treating components directly rendered by the route component as owners regardless of depth.

react/package.json
```diff
@@ -96,9 +96,10 @@
     "vitest": "vitest run",
     "vitest:watch": "vitest",
     "build": "pnpm run build:only && cp -r ./build/* ../build/web/",
-    "build:only": "pnpm run relay && NODE_OPTIONS='--max-old-space-size=4096' vite build",
+    "build:only": "pnpm run relay && pnpm run search-index && NODE_OPTIONS='--max-old-space-size=4096' vite build",
```
agatha197: Regenerating the index inside `build:only` can silently ship an index that differs from the committed one, bypassing the verify.sh drift gate. Either drop it here and rely on the gate, or stop committing the artifact — one or the other.

agatha197 (changes requested):
Reviewed the generator (not the 5,185-line generated JSON, per review-ignored-paths). Determinism and the committed artifact both reproduce locally — nice work on `assertConfiguredPaths` and the stale-override checks. A few things to settle before merge, plus two decisions worth an explicit answer in the PR: (1) keeping the artifact committed means nearly every PR touching user-facing text will also touch this one big JSON (constant conflicts in stacks) — gitignore + generate-before-verify would remove that conflict class; (2) this new build contract + verify gate looks like ADR territory per `.claude/rules/adr.md`.

react/scripts/build-search-index.mjs
nowgnuesLee: Fixed in `542e4cd5f` — an indexed key absent from `en.json` is now an error (exit 1, before the file is written). `KNOWN_MISSING_KEYS` is the explicit allowlist for a deliberate absence; it is empty today.

react/scripts/build-search-index.mjs
nowgnuesLee: Fixed in `542e4cd5f` — esbuild and import-scan failures are counted per file and reported as a warning when non-zero (zero on the current tree), naming the consequence: pages rendering through those files index thinner than they are.

react/scripts/build-search-index.mjs
nowgnuesLee: Looked into this before touching the heuristic, and the depth rule is not what makes `/project/:projectName/admin/session` thin. Its route component (`ProjectAdminSessionPage.tsx`, 388 lines) is not a wrapper: it declares 7 `t()` keys of its own and renders its table and filters from `backend.ai-ui`, which the extractor never enters by design (`EXTERNAL_PREFIXES` — BUI has its own i18next instance and locale files). `/admin/session` gets 162 keys because its UI is still host-side. Route entries that import another route's entry (the cut that would hurt a thin wrapper) exist in four places — `/admin/data` and `/project/…/admin/data` → `VFolderNodeListPage`, `/admin/session` → the legacy log opener, `/project/…/start` → `SessionLauncherPage` — and none of them lose coverage from it; the importing pages carry 160–190 keys of their own. So the fix for the thin pages would be indexing BUI's locale keys for the BUI components a page renders, which is a scope decision rather than a heuristic tweak. Leaving this open until the author decides.

react/package.json
nowgnuesLee: Fixed in `4f6b8e6f1` — `build:only` no longer regenerates the index; the committed file is what ships, and `searchIndex.test.ts` asserts the script stays that way. The drift gate that makes this safe now also runs in CI: `typecheck.yml` builds the index after relay-compiler and fails on drift next to the `__generated__` check (verify.sh is the local harness, no workflow runs it), with `resources/i18n/en.json` in the path filter because the build fails on a key it cannot find. The rebuild-and-diff idiom is one `scripts/check-generated-drift.sh` shared by verify.sh and the workflow. Decision and rejected alternatives are ADR 0003 (`docs/adr/0003-committed-search-index-artifact.md`).

nowgnuesLee:
Answering the two questions in @agatha197's review body:

1. **Committed artifact vs gitignore + generate-before-verify** — keeping it committed, and making the committed file the single source: `build:only` no longer regenerates it (`4f6b8e6f1`), `typecheck.yml` rebuilds and fails on drift the way it does for `__generated__`, and `searchIndex.test.ts` guards the build script. gitignore + generate was rejected because `searchIndex.types.ts` statically imports the JSON (a fresh checkout would fail `tsc`/vitest before generation) and because the repo's Relay artifacts already follow the commit + drift-gate convention. The conflict cost is real and stays; it is paid with the regenerate-never-hand-merge rule.
2. **ADR** — agreed; ADR 0003 (`docs/adr/0003-committed-search-index-artifact.md`, indexed in `docs/ARCHITECTURE.md`) records the decision, the three rejected alternatives (gitignore + generate, regenerate in build, `--check` in build) and the limits (advisory CI check, BUI keys not indexed).

react/scripts/build-search-index.mjs
nowgnuesLee: Decision from the author: BUI locale keys stay out of the index for now. The thin pages come from UI that `backend.ai-ui` renders (the extractor never enters it — `EXTERNAL_PREFIXES`, BUI has its own i18next instance), not from the ownership depth, so widening coverage means indexing BUI's locale files per rendered component — a scope of its own. We will take that up if a concrete "search cannot find X" report shows the gap matters in practice; ADR 0003 records the limit under Consequences.

agatha197 (approved):
LGTM

PR #8812, opened 2026-08-16
feat(FR-3558): add the global search palette — index-backed search core, UI, and actions

### LINKED ISSUE #8807: Global search palette (Cmd+K) with a build-time page/tab/setting index

nowgnuesLee:
Resolves #8807 (FR-3558)

## What

The complete global search palette, folded from three stacked layers (#8813, #8814 — closed, see their "Folded into" comments) because they edited each other's files and only review coherently as one diff:

- **Search core** — `buildHits` joins the generated index (#8811) with `useWebUIMenuItems()` into page / tab / setting-item hits; `visibility` reuses the sidebar's gating (`inactiveList` hidden, 4 runtime-gated tabs via an override map); `rank` resolves i18n keys at query time (current locale + English) with fuse.js — ASCII fuzzy, CJK substring; recents persisted via `useBAISettingUserState`.
- **Palette UI** — header trigger + `mod+k` (Astryx `useHotkeys`), Astryx `CommandPalette` rows with icon / title / scope breadcrumb / "Found in …" lines, empty state = recents + grouped pages; opening runs in a `startTransition` (trigger shows `isLoading` while pending) instead of an empty Suspense fallback; `useSettingArrival` scrolls + transiently highlights `?setting=<key>` targets; palette hover matches Astryx's neutral overlay (band-inherited overlay scoped out via the theme's `command-palette-item` block, THEME_NAME_REV 17, rebuilt on Astryx 0.4.3).
- **Actions** — static registry (Create: session / folder / deployment / FastTrack; Appearance: theme light-dark-system, UI language; Panels & help: notifications, sidebar, manual, user settings); sider + notification drawer state lifted to Jotai (`useShellPanels`, `[`/`]` unchanged); `useHelpURL` extracted from `WEBUIHelpButton`; `?action=add` opener on Data / Deployments.
- All palette copy translated in the 20 locales (terminology matched per file).

## Why

Routes are lazy chunks, so a palette cannot search rendered pages; it searches the build-time index and resolves keys through `t()` at query time — every locale works without re-indexing.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (includes the search-index drift gate)
- vitest: palette suite green (85 tests at fold time); every action live-verified on a dev cluster in en + ko
- Screenshots in the fold-source PRs #8813 / #8814 and `~/.claude/jobs/8ba9997e/tmp/palette-*/`

## Review fixes

Applied from a blind review of this layer (commit `254a1d0`):

- **Locale parity** — `webui.menu.groupName.General` existed only in en+ko while `useGlobalSearchSource` uses it for the ungrouped heading *and* as the `/usersettings` fallback group; added to the other 19 locales with each file's established term.
- **Gate consistency** — actions now declare the `menuKey` they act on and inherit that page's visibility in `isHitVisible`. Fixes two holes: the FastTrack action ignored `blockList` (its sidebar entry is `pipeline`), and no action consulted `inactiveList` at all, so "Create Folder" survived an inactivated Data page.
- **Lazy palette chunk** — `GlobalSearchPaletteButton` now `React.lazy`s the palette, so the 141 KB generated index + fuse.js leave the entry bundle (verified: a dedicated 188 KB `GlobalSearchPalette-*.js` chunk, zero index bytes in `index-*.js`). The existing `startTransition` open path absorbs the chunk suspend exactly like the client suspend — still no local Suspense.
- **Twin rows** — the three "Data" and two "Sessions" page hits carried no breadcrumb and two shared a sidebar group, so they rendered as byte-identical rows. The admin twins (`/admin/…` → `webui.search.scope.Administration`, `/project/x/admin/…` → `webui.search.scope.ProjectAdministration`) now carry their scope as a trailing marker on the right of the row — its own keys rather than the sidebar's `webui.menu.Administration`, because that label's Korean "관리" reads as "management" standing alone (the marker says "관리자" / "프로젝트 관리자"); the project-scoped page carries nothing — the palette only ever runs in the active project and cannot switch it, so a "Project: <name>" line (an earlier revision) told the user nothing actionable and was dropped on review.
- **Backdrop** — Astryx `Dialog` blurs its `::backdrop`; the modal and drawer masks dropped that blur and keep the tint only (FR-3585). The palette overrides the same StyleX `::backdrop` key with `backdrop-filter: none`, so it dims the page like every other overlay instead of blurring it.
- **Matching tightened** — fuzzy threshold 0.35 → 0.2 (one edit per five characters: "sesions" still finds Sessions, "kernel" no longer reaches "in·ternal" / "ex·ternal" body sentences); multi-word queries go through Fuse extended search so each word is its own pattern joined by AND ("create session" no longer matches "Terminate Session"), with operator characters stripped from the query; and a page surfaces at most one "Found in …" row, its best-scoring sentence, instead of up to three clones.
- **Keyword noise** — dropped `menuKey` from tab / setting-item keywords (weight 0.6), which was floating 15 unrelated `/admin/settings` rows above real body matches for `settings`. The page hit still matches its own key at weight 1.0.
- **Dead code** — removed `SearchConfigFlags.hideAgents` / `.enableReservoir`, `SearchContext.isAdmin` / `.projectName`, and `RecentSearchHit.labelKey` (re-resolved live now; old persisted entries carrying it are ignored). The duplicated config assembly is now one `toSearchConfigFlags()`.
- **Mobile deep link** — `?setting=` no longer expires against an item that was never painted: `SettingList` passes only the on-screen titles, so below `md` (nav view) or on a non-matching group tab the param survives until the item actually renders.
- **Tests** — the twin-page fixture now mirrors production's shared admin group instead of fabricating a unique label per key, and asserts the trailing scope marker is what separates the rows; added a positive `mod+k` open test and inactiveList/blockList action-gate tests. Recents duplicating in the empty state is kept deliberately (VS Code behaviour) and noted in-line.

`bash scripts/verify.sh` → `=== ALL PASS ===` on this branch and on both upstack branches; `vitest run src/components/GlobalSearchPalette src/hooks` green (458 tests).

- **Update — urgent open restored** (supersedes the `startTransition` note in the lazy-palette bullet above): per requester preference after live testing, both open paths set state urgently again and the palette is wrapped in a local `<Suspense fallback={null}>`, which absorbs both the lazy chunk and the client suspend; the trigger loading state went away with the transition.

- **Update — first-open flash fixed** (`7d2703f`): Astryx renders `CommandPaletteEmpty` with `emptyBootstrapText` whenever `search === '' && optimisticResults.length === 0`, which is always true on the first paint after mount (`bootstrap()` commits in a transition, after the dialog has painted). We passed only `emptySearchText`, so the opening dialog painted Astryx's default *"Type to search"* — reading as an empty result set — for the whole open animation (CDP screencast: frames 003–009 of the first open; 134 ms in the rAF sampler). `emptyBootstrapText` now carries three token-only skeleton rows shaped like the real ones, and the trigger warms `import('./GlobalSearchPalette')` from an idle callback on mount. Measured over 3 trials: 8 palette module requests on first open before, **0** after; no frame shows the default copy.

- **Update — open-path waste removed and bootstrap single-commit** (`afd62d3`): profiled the first open with a Playwright + CDP sampler (dev server, 6× CPU throttle, medians of 3). The data layer was not the stall — `buildHits` is 2–7 ms per call under throttle and the Fuse index is built lazily on the first keystroke, not at mount — but the palette rebuilt the whole hit list on all **8** of its renders per open (each rebuild also discarding the ranker's Fuse index, which is keyed on the hit array's identity), and `bootstrap()` returned a fresh array on every call Astryx makes, so the 40-row list committed twice. Hits, their id lookup and the bootstrap rows now come from a module-level cache (`searchArtifacts.ts`) that outlives unmount; hit builds per open **8 → 1**, and `bootstrap()` returns the same array so later calls are `setState` bail-outs. Translators are memoised per (i18next instance, language) — a locale switch still invalidates everything — and `warmGlobalSearch`, re-exported from the lazy chunk so the index stays out of the entry bundle, resolves the whole index against both locales in the trigger's existing idle callback (9.5 ms unthrottled); the palette additionally warms the ranker on an idle callback after it opens. Measured (6× throttle, median of 3): click→painted 1558 → 1442 ms, blocked 1256 → 1166 ms, **first keystroke 501 → 390 ms**, reopen→painted 1012 → 996 ms. The reported 28-then-40 growth did not reproduce here (40 in one array on every run, before and after); the residual open cost is the React render of the palette and its 40 rows, which the CPU profile shows as `jsx` / `createElement` / `astryxVars`.

- **Update — reverted: Astryx `CommandPalette` is used as shipped.** An earlier revision swizzled `CommandPalette` (+ context, Input, Item, List) into `GlobalSearchPalette/astryx/` to seed the mount commit from a synchronous `bootstrapSync()`, so the first painted frame carried the whole list. Review found the fork drifting from the installed 0.5.2 with nothing able to report it, and reversing the repo's recorded swizzle rejections without an ADR — so the fork, `bootstrapSync`, and the prettier/eslint ignores are gone. The first frame of every open is the skeleton placeholder again (the update above), replaced by the rows on the bootstrap transition's commit.


react/src/components/GlobalSearchPalette/GlobalSearchPaletteButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+  // `import()` caches its module promise, and `warmGlobalSearch` resolves the
+  // whole index against the current locale and English, so neither the first
+  // open nor the first keystroke pays for them.
+  useEffect(() => {
+    const warm = () => {
+      void importPalette().then((palette) => palette.warmGlobalSearch(i18n));
+    };
+    if (typeof requestIdleCallback === 'function') requestIdleCallback(warm);
+    else setTimeout(warm, 200);
+  }, [i18n]);
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in 051a0a2ae — the warm-up effect now also depends on `i18n.resolvedLanguage`, so a language switch re-resolves the index for the newly selected locale instead of leaving it cold until the next mount.

react/src/hooks/useShellPanels.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Open/closed state of the two app-shell panels. Both used to live inside the
+ * component that owns the `[` / `]` shortcut, which left the global search
+ * palette unable to drive them from outside the layout.
+ */
+export const notificationDrawerOpenState = atom(false);
+
+export const useNotificationDrawerState = () =>
+  useAtom(notificationDrawerOpenState);
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in 051a0a2ae — `useNotificationDrawerState` now uses a block body with the `'use memo'` directive, consistent with the other hooks in the file.

react/src/components/GlobalSearchPalette/GlobalSearchPaletteButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const { t, i18n } = useTranslation();
+  const { isDarkMode } = useThemeMode();
+  const [isOpen, setIsOpen] = useState(false);
+
+  // `allowInInputs` so the palette is reachable while a form field has focus,
+  // which is where a user most often reaches for it.
+  useHotkeys([
+    { keys: 'mod+k', allowInInputs: true, onPress: () => setIsOpen(true) },
+  ]);
```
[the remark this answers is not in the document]
nowgnuesLee: The E2E layer exists — it is the top PR of this stack, #8864 (`e2e/global-search/global-search-palette.spec.ts`, 7 scenarios including `mod+k` open, navigation/tab/setting arrivals, and the theme action, plus the coverage-report update). It is split out per the stacked-PR workflow rather than folded into this PR.

react/src/components/SettingItem.tsx
```diff
@@ -83,10 +92,16 @@ const SettingItem: React.FC<SettingItemProps> = ({
...
+  isArrivalTarget = false,
 }) => {
   'use memo';
 
+  const rootRef = useRef<HTMLDivElement>(null);
   const [isRowHovered, setIsRowHovered] = useState(false);
+
+  useEffect(() => {
+    if (isArrivalTarget) rootRef.current?.scrollIntoView({ block: 'center' });
+  }, [isArrivalTarget]);
```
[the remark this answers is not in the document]
nowgnuesLee: Considered and not applying in this PR. The arrival cue is deliberately visual-only for now: moving focus after a palette-driven navigation is a user-visible focus-behavior change that interacts with the dialog's own focus restoration, and we prefer to keep this PR's arrival contract minimal. If a11y feedback confirms the need, it can land as its own change.

react/src/components/GlobalSearchPalette/visibility.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * The single visibility predicate: pages are visible iff their menu key
+ * survived `useWebUIMenuItems()`, `inactiveList` entries are hidden, tabs and
+ * setting items inherit the page gate, and actions inherit the gate of the page
+ * they act on before their own runs.
+ */
+export const isHitVisible = (hit: SearchHit, ctx: SearchContext): boolean => {
+  const menuKey = hit.menuKey;
+
+  if (hit.kind === 'action') {
```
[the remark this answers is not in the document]
nowgnuesLee: Considered and not applying. Recording per-key tab ownership means changing the index schema, the extractor walk, and the visibility filter — a substantial design change out of proportion to the exposure: the matched strings are i18n labels already shipped to every client in the bundle, and arriving on the page still cannot open the gated tab (the tab hit itself is filtered). So this is UX polish (a hit that lands without its matched content), not an information leak, and the current key placement is accepted as-is.

agatha197 (changes requested):
Image The project name appears without any explanation and it is too duplicated info. Can you improve it?

react/src/components/GlobalSearchPalette/astryx/CommandPalette/CommandPalette.tsx
```diff
@@ -0,0 +1,655 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+
+'use client';
+/**
+ * SWIZZLED from @astryxdesign/core CommandPalette @ 0.4.3, for one behaviour
```
agatha197: The header says 0.4.3 but the workspace installs 0.5.2, and this fork already drifts from the installed source beyond the SYNC-SEED blocks (interaction-overlay hover, cursor behavior, `useDialogContext`) — with prettier/eslint both ignoring this directory, nothing will ever catch that. Please re-swizzle from 0.5.2, or better: express the ~15-line delta as a `pnpm patch` (like the existing `@astryxdesign__core@0.5.2.patch`), which hard-fails on a core bump instead of drifting silently.

react/src/components/GlobalSearchPalette/GlobalSearchPaletteButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+    // be its own dependency for a language switch to re-warm the index.
+  }, [i18n, i18n.resolvedLanguage]);
+
+  const bandMediaMode = isDarkMode ? 'light' : 'dark';
+
+  return (
+    <>
+      <Tooltip
+        content={
+          <MediaTheme mode="dark">
```
agatha197: `BAINotificationButton` deliberately avoids a `MediaTheme` wrapper here — its comment cites FR-3726 (the dark palette's `--color-neutral` equals the bubble). Drop the wrapper so the two header buttons stay consistent?

react/src/components/GlobalSearchPalette/GlobalSearchPaletteButton.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  // Warm the chunk AND the artifacts it builds while the header is idle:
+  // `import()` caches its module promise, and `warmGlobalSearch` resolves the
+  // whole index against the current locale and English, so neither the first
+  // open nor the first keystroke pays for them.
+  useEffect(() => {
+    const warm = () => {
+      void importPalette().then((palette) => palette.warmGlobalSearch(i18n));
+    };
+    if (typeof requestIdleCallback === 'function') requestIdleCallback(warm);
```
agatha197: The idle/timeout handle is not captured, so the warm-up cannot be cancelled on unmount or locale change — the sibling effect in `useGlobalSearchSource` does clean up. Please return a cleanup with `cancelIdleCallback` / `clearTimeout`.

react/src/hooks/useShellPanels.ts
```diff
@@ -0,0 +1,42 @@
...
+  boolean,
+  (value: SetStateAction<boolean>) => void,
+] => {
+  'use memo';
+
+  const [compactSidebarActive] = useBAISettingUserState('compact_sidebar');
+  const [override, setOverride] = useAtom(siderCollapsedOverrideState);
+
+  return [
+    override ?? !!compactSidebarActive,
```
agatha197, line 34: Behavior change worth confirming: `compact_sidebar` used to be read once at mount; now it applies live *until* the first manual toggle, after which setting changes are ignored for the rest of the session — toggling it in UserSettings will appear to do nothing. Intended?

react/src/components/GlobalSearchPalette/searchArtifacts.ts
```diff
@@ -0,0 +1,322 @@
...
+ */
+const syncIcons = (
+  hits: ReadonlyArray<SearchHit>,
+  menuSources: ReadonlyArray<MenuHitSource>,
+): void => {
+  const sourceByKey = _.keyBy(menuSources, 'key');
+  _.forEach(hits, (hit) => {
+    if (hit.kind === 'action' || !hit.menuKey) return;
+    const icon = sourceByKey[hit.menuKey]?.icon;
+    if (icon !== undefined && icon !== hit.icon) hit.icon = icon;
```
agatha197: This mutates cached hits during render, and the icon write alone cannot trigger a repaint — it works because `useWebUIMenuItems` happens to re-render alongside. Safer to drop `icon` from hits and look it up by `menuKey` in `renderItem`.

react/src/components/GlobalSearchPalette/visibility.ts
```diff
@@ -0,0 +1,63 @@
...
+  [tabGateKey('admin-deployments', 'tab', 'prometheus-preset')]: (ctx) =>
+    ctx.supports('prometheus-query-preset'),
+  [tabGateKey('admin-deployments', 'tab', 'deployment-presets')]: (ctx) =>
+    ctx.supports('deployment-preset'),
+};
+
+/** Whether the menu still offers the page, i.e. neither blocked nor inactive. */
+const isMenuKeyOffered = (menuKey: string, ctx: SearchContext): boolean =>
+  !ctx.disabledMenuKeys.has(menuKey) &&
+  (ctx.visibleMenuKeys.has(menuKey) || ALWAYS_VISIBLE_MENU_KEYS.has(menuKey));
```
agatha197, line 39: `blockList` entries are `_.remove`'d from the menu rather than disabled, so a blocked `usersettings` still passes this check and its setting hits stay searchable. Should the always-visible set respect the blockList too?

agatha197 (changes requested):
Solid feature work — all 21 locales filled, no antd imports, no GraphQL changes, thick tests, and the 73-file scope is all justified by the palette. The blocker is the vendored CommandPalette fork (version drift + it reverses the repo's three recorded swizzle rejections without an ADR — `.claude/rules/adr.md` asks for that to be flagged). The rest are smaller alignment fixes.

react/src/components/GlobalSearchPalette/astryx/CommandPalette/CommandPalette.tsx
nowgnuesLee: Fixed in `36bd371b9` — the vendored `CommandPalette` (+ context, Input, Item, List) is gone, along with `bootstrapSync` and the prettier/eslint ignore entries; the palette uses Astryx's `CommandPalette` as shipped. The first painted frame is the skeleton placeholder again, replaced by the rows on the bootstrap transition's commit — we took that trade rather than carrying a patch for it.

react/src/components/GlobalSearchPalette/GlobalSearchPaletteButton.tsx
nowgnuesLee: Fixed in `691a781e4` — the `MediaTheme` wrapper around the tooltip content is gone; the `Kbd` is coloured by the theme's tooltip block, the same way `BAINotificationButton`'s is.

react/src/components/GlobalSearchPalette/GlobalSearchPaletteButton.tsx
nowgnuesLee: Fixed in `691a781e4` — the effect captures the `requestIdleCallback` / `setTimeout` handle and returns a cleanup that cancels it, matching the sibling effect in `useGlobalSearchSource`.

react/src/components/GlobalSearchPalette/searchArtifacts.ts
nowgnuesLee: Fixed in `691a781e4` — `syncIcons` is gone and hits no longer carry menu icons. `useGlobalSearchSource` exposes `getIcon(hit)`, which reads the live menu by `menuKey` (`toMenuIcons`) on every render; action hits keep their own static glyph.

react/src/components/GlobalSearchPalette/visibility.ts
nowgnuesLee: Fixed in `691a781e4` — `SearchContext` gains `blockedMenuKeys` (from `useWebUIMenuItems().blockList`), `isMenuKeyOffered` checks it before the always-visible whitelist, and it is part of the artifact cache signature. Test added: a blocked `usersettings` hides both the page hit and its setting hits.

nowgnuesLee:
Addressed the review from @agatha197 on the duplicated project name: the "Project: <name>" second line is gone (`470450843`) — the palette only runs in the active project and cannot switch it, so it told the user nothing actionable. The admin twins of a project page now carry their scope ("Administration" / "Project administration") as a trailing marker on the right of the row instead, with their own i18n keys (`cadfa5665`).

react/src/hooks/useShellPanels.ts
nowgnuesLee: Behaviour is unchanged from `main`, only the state moved. The hook alone reads as you describe — `override ?? setting`, and a boolean override stops following — but the re-sync that makes the setting win is still in `MainLayout.tsx` (the `prevCompactSidebarActive` block, same as on `main`): whenever `compact_sidebar` changes it calls `setSideCollapsed(!!compactSidebarActive)`, which now writes the override, so the sidebar follows a settings change after a manual toggle exactly as before. `main` never read the setting only once either — the same block re-applied it on every change. What the PR did is lift `sideCollapsed` into a Jotai atom so the palette's "Toggle sidebar" action can drive it from outside the layout; the follow-the-setting rule is split between the hook and the layout, which is what made it read as a regression. Leaving the code as is for this PR.

agatha197 (approved):
LGTM

PR #8813, opened 2026-08-16
feat(FR-3558): add the global search palette UI with header trigger, ⌘K, and setting arrival

### LINKED ISSUE #8807: Global search palette (Cmd+K) with a build-time page/tab/setting index

nowgnuesLee:
Resolves #8807 (FR-3558)

## What

- `GlobalSearchPalette` on Astryx `CommandPalette` + `GlobalSearchPaletteButton` in the header (`mod+k` via Astryx `useHotkeys`, `⌘K` hint in the input). Rows are icon + title + scope breadcrumb, or "Found in …" for a body match; the empty state is recents plus the grouped page list.
- ~120 ms input debounce (`useDebouncedSearchSource`) — `CommandPalette` calls `search()` on every keystroke and has no debounce of its own, so a fast typist re-ranked the whole index per character. Only the trigger is delayed; the ranker stays synchronous and the input echoes keystrokes immediately.
- The row icon sits in a fixed, non-shrinking square slot so a long secondary line truncates the text column instead of squashing the glyph.
- `useSettingArrival()` implements the `?setting=<i18nKey>` arrival: scroll to the item, mark it for 1.5 s, then strip the param.
- i18n: `webui.search.*` in en + ko.

## Why

Two of the three fixes come from the requester's live test (2026-08-16): the palette re-ranked on every keystroke, and row icons shrank next to long "found in" lines.

## Verification

- vitest: 59 tests (palette rows, debounce, setting arrival, plus the core suite)
- Live-checked on `https://fr-3558.localhost:1355` (dark mode, Korean UI) — the two findings above are what this PR fixes; screenshots of the pre-fix state are in the FR-3558 working notes.

---

### Mechanism

`WebUIHeader` declares the band's **reversed** interaction overlays (`ANTD_REVERSED_BAND_OVERLAYS`) as inline literals on `.bai-webui-header`. Astryx's `CommandPalette` renders a **non-portalled native `<dialog>`**, and the top layer does not change DOM ancestry — so the palette, mounted by `GlobalSearchPaletteButton` inside that band, inherited the band's literal on every row.

`light-dark()` never participated: the inherited value was already resolved for the band's *inverted* surface, which is why the dark-mode row painted 6% **black**.

### Fix

Re-declare the pair on `command-palette-item` in the theme — the same treatment `MENU_PANEL_PAGE_OVERLAYS` already applies to a `DropdownMenu` panel ("a floating panel is a PAGE surface") — and lift the active row to `--color-accent-muted`, the token Astryx's own `CommandPaletteItem` uses for a palette's current item and the only one clearly visible on both popover surfaces. `THEME_NAME_REV` 15 → 16, built artifacts regenerated, stale `bai-r15-*` deleted.

Hover and keyboard highlight are the same state in Astryx (`handleMouseEnter` sets `highlightedIndex`), so one token fixes both.

### Measured, live, hovering a palette row

| Mode | Popover surface | Hovered row — before | Hovered row — after |
|---|---|---|---|
| Light | `rgb(255,255,255)` | `rgba(255,255,255,0.16)` → white on white, **invisible** | `color(srgb 1 0.478431 0 / 0.2)` = brand orange 20% → ≈ `rgb(255,228,204)` |
| Dark | `rgb(31,31,31)` | `rgba(0,0,0,0.06)` → ≈ `rgb(29,29,29)`, **Δ2/255** | `color(srgb 0.745098 0.368627 0.0235294 / 0.25)` → ≈ `rgb(71,47,25)` |

Dark mode entered via the header theme toggle with `document.documentElement.dataset.theme === 'dark'` asserted; `data-astryx-theme` confirmed `bai-r16-default-brand-hu4alvq` (i.e. the new artifact is the one being served, not a stale r15).

### Also — debounce raised 120 ms → 500 ms

120 ms is shorter than an ordinary typing cadence, so it coalesced nothing and results still re-ranked per character. Live check: 7 keystrokes in 29 ms echoed all 7 characters into the input immediately and produced **one** results-list recompute after the burst (14 → 23 rows), not one per key. The exact 500 ms could not be timed in the driven tab — it reports `visibilityState: hidden`, where Chrome clamps `setTimeout` to 1000 ms; the value itself is pinned by `useDebouncedSearchSource.test.ts` under fake timers.

### Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` on `FR-3558-palette-ui`, `FR-3558-actions`, `FR-3558-e2e` after `gh stack rebase --upstack`
- `pnpm --prefix ./react exec vitest run` → **96 files / 1464 tests passed**
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → exit 1 on **one pre-existing** finding, `packages/backend.ai-ui/src/components/BAIText.css:28 var(--x)` (introduced by FR-3518, untouched here). Verified identical output with this commit stashed, so this change adds no new undeclared token.
- Zero `pageerror` / console exceptions across both mode passes.

### Residue

- The band-overlay leak itself is untouched — any *other* non-portalled floating surface mounted inside `WebUIHeader` inherits the same literals. This fix follows the established per-component pattern (`dropdown-menu`, `field`, now `command-palette-item`) rather than changing how the band declares them.
- `--color-overlay-pressed` is restored to the page value rather than an accent, so `:active` stays distinguishable from the highlight.

---
**Correction (2026-08-17):** the earlier debounce note is superseded — the input debounce (120→500 ms) was **removed** in `c6e519d18`. Astryx `CommandPalette` internally narrows the previous results on every keystroke (`runSearch` optimistic narrowing, no opt-out prop), so a debounced source produced a two-phase repaint per pause (instant substring preview → delayed ranked commit), which read as flicker. With the synchronous local index the palette commits final ranked results in one paint per keystroke, matching the Astryx docs palette UX. Requested by the reporter after live testing.

> **Correction (2026-08-17):** the accent-muted hover from the fix above was reverted to Astryx's stock neutral overlay per review — stock `CommandPaletteItem` uses `--color-overlay-hover` for both `:hover` and `itemHighlighted` and reserves `--color-accent-muted` for picker-mode selection. The scope fix (re-declaring the overlays against the app scheme instead of inheriting the header band's reversed literals) is unchanged; measured hovered row is now `rgba(0,0,0,0.06)` on white in light and `rgb(38,38,38)` on `rgb(31,31,31)` in dark. `THEME_NAME_REV` 16 → 17. The debounce note above is also superseded — the debounce was removed entirely (Astryx's internal optimistic narrowing + a delayed commit is what produced the two-phase repaint users saw as flicker; a synchronous source commits in one paint).

---
**Update (2026-08-18) — opening runs in a transition, not a Suspense fallback (`7e30a7ee`).** The palette tree suspends via `useSuspendedBackendaiClient`, and a Suspense boundary mounted for the first time *inside* a transition still shows its fallback — so the empty-dialog flash could not be fixed by keeping the local boundary. The local `<Suspense>` around `<GlobalSearchPalette>` is removed and both open paths (`mod+k` `onPress` and the trigger `onClick`) go through `startTransition`, so React holds the commit until the subtree is ready: the header stays painted and the palette mounts whole. Closing stays urgent. A ≤2-line comment at the open paths records that every open path must stay inside `startTransition` — an urgent open propagates the suspension to MainLayout's ancestor boundary (`<Suspense fallback={null}>` around sider + header) and blanks the header.

Pending feedback on the trigger uses Astryx `IconButton`'s own props — `isLoading={isPending}` plus `isInterruptible` — no hand-rolled spinner and no extra `aria-busy` prop: `isInterruptible` keeps the button enabled and un-deduped while the transition holds (it stays clickable-looking), Astryx renders its `Spinner` on the built-in delay so a fast open shows nothing, and `Button` derives `aria-busy` from `isLoading` itself (it sets the attribute *after* the props spread, so a consumer-passed `aria-busy` would be inert). In practice the pending window is only visible while the backend client is still connecting; the state clears when the palette commits.

`GlobalSearchPalette.test.tsx` gains a test that models the real ancestor boundary — the button rendered inside `<Suspense fallback={<span>header-blanked</span>}>` with a gated `useGlobalSearchSource` mock — and asserts that while the subtree is held there is no dialog, no rows, and no fallback, then the rows appear on release. Mutation-checked: reverting `open` to an urgent `setIsOpen(true)` fails that test. Verified with the bare-render version first, which passed either way — without an ancestor boundary the two are indistinguishable.


nowgnuesLee:
Folded into #8812 — the core/UI/actions layers edited each other's files, so they are now one reviewable PR. No content was lost; the folded branch is byte-identical to this PR's final head.

PR #8815, opened 2026-08-16
e2e(FR-3558): cover the global search palette

### LINKED ISSUE #8807: Global search palette (Cmd+K) with a build-time page/tab/setting index

nowgnuesLee:
Resolves #8807 (FR-3558)

## What

- `e2e/global-search-palette.spec.ts` — 7 cases: open from the header button, open with `mod+k`, Escape closes, a page hit navigates (arrow-selected), a tab hit lands on `?tab=`, a setting hit arrives with `data-arrival` set and then strips `?setting=`, the theme action flips the header toggle, and a selected hit comes back under "Recent".
- `e2e/utils/test-util.ts`: `login()`'s `getByLabel('Endpoint')` also matched the endpoint-history menu, its trigger, and the "About Endpoint" button — five elements, so every login threw a strict-mode violation. Narrowed to the textbox role.
- `e2e/E2E_COVERAGE_REPORT.md` updated (new section, summary row, totals, date).

## Why

The palette's value is the arrival — the URL it produces and the state it leaves behind — which only an end-to-end run can assert.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (on this branch and on all four below it)
- vitest: 77 tests across the stack
- **Playwright run blocked by the environment, not by the spec.** Against `https://fr-3558.localhost:1355` + webserver `http://10.82.0.6:8090`, the UI login never completes: `POST /server/login` returns `400 {"title": "You must provide the username field."}` — the webserver cannot read the client's AES-encoded login payload, while a direct API login with the same credentials returns `200 authenticated: true`. An existing untouched spec (`e2e/information/information.spec.ts`) fails identically on the same host pair, so the spec still needs a run against a cluster where UI login succeeds.


nowgnuesLee:
Superseded by #8864: deleting the folded `FR-3558-actions` base branch auto-closed this PR during the stack restructure. Same branch, same content.

PR #8842, opened 2026-08-18
style(FR-3572): give the deployment Scheduling History controls the accent icon-button treatment

### LINKED ISSUE #8841: Deployment detail: Scheduling History controls are default-coloured link buttons, not the accent icon button the session drawer uses
## Component

- **Names**: `DeploymentBasicInfoCard` (Lifecycle row) and `DeploymentReplicasCard` (replica row Lifecycle cell), `react/src/components/`
- **Astryx primitive**: currently `BAIButton type="link"`; should be `IconButton variant="ghost" size="sm"` + `.bai-action-accent`
- **Reference**: `SessionDetailContent` Status row, already converted

## Where it is reachable

- Deployments -> a deployment -> **Basic Information -> Lifecycle** row
- Same page, **Replicas** table -> a replica row -> Lifecycle cell

## Observed

Measured live on the deployment detail page (dark, 1600x1100):

- the Lifecycle row control renders **145 x 24** with a visible `Scheduling History` **text label**, preceded by a vertical `Divider`
- computed `color` is **`rgb(255,255,255)`** — the default text colour, not the accent
- class list is just `astryx-button`; no `.bai-action-accent`
- the replica-row control is icon-only but is also a `BAIButton type="link"`, wrapped in a separate `Tooltip`, and its accessible name resolves to `"Action"` rather than the scheduling-history label

The session drawer equivalent is a **24 x 24 icon-only** `IconButton` carrying `.bai-action-accent` and a `tooltip`.

## Expected

Both deployment controls match the session drawer: icon-only accent `IconButton` with a tooltip.

## Root cause

The same defect the session drawer already fixed as **FR-3482 QA finding Q-37**: antd `type="link"` buttons painted `colorLink` `#FF7A00`; the Astryx conversion dropped the tint to `--color-text-primary`, which reads as non-interactive. `.bai-action-accent` restores it through `--color-text-accent` (see `packages/backend.ai-ui/src/styles/actionAccent.css`). The deployment page was never given the same treatment.

## Environment

- **Build**: `FR-3556` @ 26.9.0-alpha.0.8646, Astryx 0.4.0
- **Mode**: both (the tint is missing in each)

## Notes

Reported alongside FR-3556 / FR-3557, which fix the chevron inside the modal these buttons open. This issue is about the openers, not the modal.

JIRA Issue: FR-3572

agatha197:
Resolves #8841 (FR-3572)

## Problem

The two **Scheduling History** openers on the deployment detail page were still antd-era `BAIButton type="link"`. The Astryx conversion drops that tint to `--color-text-primary`, so they read as plain non-interactive text — the same defect the session drawer's Status row already fixed as **FR-3482 QA finding Q-37**, which restores it with `.bai-action-accent` (`--color-text-accent`).

After — the Lifecycle row now matches the session drawer's status row (badge + accent history icon):

## Measured, in the real app (deployment detail, dark)

| | before | after | session drawer (reference) |
|---|---|---|---|
| Lifecycle row control | **145 × 24**, visible `Scheduling History` text | **24 × 24**, icon-only | 24 × 24, icon-only |
| computed `color` | `rgb(255,255,255)` (default) | **`rgb(190,94,6)`** | `rgb(190,94,6)` |
| `.bai-action-accent` | absent | **present** | present |
| tooltip | none | **yes** | yes |
| replica-row accessible name | **`"Action"`** | **`"Replica Scheduling History"`** | — |

`--color-text-accent` is `light-dark(#FF7A00, #be5e06)`, so `rgb(190,94,6)` is exactly its dark half.

## Changes

- **Basic Information → Lifecycle**: drop the text label and the leading vertical `Divider`; render the accent `IconButton` with a `tooltip`.
- **Replicas table → Lifecycle cell**: same treatment. The button was already icon-only but wrapped in a bare `Tooltip`, which left it with **no accessible name** — its computed name resolved to `"Action"`. `IconButton`'s own `label` / `tooltip` fix both the tint and the a11y name.
- `action` → Astryx's `clickAction`, which preserves the async pending state both call sites relied on (the replica one kicks off a render-as-you-fetch query).
- `Divider` and `BAIButton` imports dropped where they became unused.

## Verification

- `bash scripts/verify.sh` — **`=== ALL PASS ===`**
- `pnpm --filter backend.ai-ui build` — built
- `pnpm --prefix ./react exec vitest run` — **1405 passed** (88 files)
- Live probe on the deployment detail page; both controls measured as in the table above

## Scope

Only the two deployment-page openers. The modal they open is untouched here — its chevron is fixed separately in #8803 (FR-3556 / FR-3557).

### DIFF
--- a/react/src/components/DeploymentBasicInfoCard.tsx
+++ b/react/src/components/DeploymentBasicInfoCard.tsx
@@ -16,8 +16,8 @@ import DeploymentSchedulingHistoryModal, {
 } from './DeploymentSchedulingHistoryModal';
 import DeploymentSettingModal from './DeploymentSettingModal';
 import { ButtonGroup } from '@astryxdesign/core/ButtonGroup';
-import { Divider } from '@astryxdesign/core/Divider';
 import { DropdownMenu } from '@astryxdesign/core/DropdownMenu';
+import { IconButton } from '@astryxdesign/core/IconButton';
 import {
   MetadataList,
   MetadataListItem,
@@ -95,20 +95,19 @@ const DeploymentOverviewContent: React.FC<{
               status={deployment.metadata.status as BAIDeploymentStatus}
             />
             {onClickSchedulingHistory && (
-              <>
-                <Divider orientation="vertical" />
-                <BAIButton
-                  type="link"
-                  size="small"
-                  icon={<History size="1em" />}
-                  style={{ padding: 0 }}
-                  action={async () => {
-                    await onClickSchedulingHistory();
-                  }}
-                >
-                  {t('deployment.SchedulingHistory')}
-                </BAIButton>
-              </>
+              // Same control as the session drawer's Status row: an accent
+              // icon button, not a default-tinted link (FR-3482 Q-37 / FR-3572).
+              <IconButton
+                className="bai-action-accent"
+                variant="ghost"
+                size="sm"
+                icon={<History size="1em" />}
+                label={t('deployment.SchedulingHistory')}
+                tooltip={t('deployment.SchedulingHistory')}
+                clickAction={async () => {
+                  await onClickSchedulingHistory();
+                }}
+              />
             )}
           </BAIFlex>
         ) : (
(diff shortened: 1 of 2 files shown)


ironAiken2 (approved):
LGTM

PR #8845, opened 2026-08-18
fix(FR-3573): keep tooltip text inside the tooltip bubble

### LINKED ISSUE #8843: Sessions: session-name truncation tooltip text overflows the tooltip bubble
## Where

- **Page**: `/session` (menu: Sessions)
- **Region**: the session list's Name column — the truncation tooltip that appears when hovering a session name that is too long for the cell
- **Component** (if identified): Astryx `Text` with `maxLines={1} hasTruncateTooltip` in `SessionInfoCell` (`react/src/components/SessionListColums/SessionInfoCell.tsx`)

## Observed

Hovering a long session name opens the truncation tooltip, but the tooltip's

bubble background is narrower than the text it carries. The text runs past the

right edge of the rounded bubble and the trailing characters render directly on

the page background, outside the bubble. In the reported case the name

`mock-chat-b3217a3e-06c1-404f-abd8-d2daebd9af4b` had roughly its last nine

characters sitting outside the bubble.

The text is not wrapped and not ellipsised — it simply overflows the surface

that is supposed to contain it.

## Expected

The tooltip surface contains the whole string: the bubble either grows to fit

the text or the text wraps onto additional lines inside the bubble, with the

background covering every character. No glyph renders outside the bubble.

## Environment

- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown (reported 2026-08-18)

## Reproduction

1. Open Sessions (`/session`).
1. Find a session whose name is long enough to be truncated in the Name column

(e.g. a name ending in a UUID).

1. Hover the truncated name to open its tooltip.

## Evidence

- Screenshot: `/home/ubuntu/.claude/jobs/76abdad9/pasted-1.png` (to be attached in Jira)

## Notes

Suspected, not verified: the tooltip content may not be honouring the width cap

it is given, so the text lays out wider than the bubble. Mechanism to be

measured at fix time.

JIRA Issue: FR-3573

agatha197:
Resolves #8843 (FR-3573)

## Summary

Hovering a truncated session name opened a tooltip whose bubble was narrower than its own text — the trailing characters painted outside the surface, directly on the page background.

Fixed at the theme level: `.astryx-tooltip` now pins `white-space: normal`, so the bubble can never inherit `nowrap` from the context its trigger happens to live in.

## Mechanism (measured, live app)

Astryx tooltips are **not portalled** — `useLayer` renders the popover in place (`position: fixed` + anchor positioning), so the bubble is a DOM descendant of whatever the trigger sits inside. Three facts compose:

1. Astryx caps tooltip content at `max-width: 300px` (`useTooltip` `styles.content`, and `truncationTooltipStyles.content` for the truncate variant) and **relies on wrapping** to stay inside that box.
2. A single-line truncating `Text` (`maxLines={1}`) carries `white-space: nowrap`, and `BAITableAstryx` additionally wraps every truncate-mode cell's content in a `white-space: nowrap` span.
3. `white-space` inherits. Measured on the tooltip's ancestor chain: `nowrap` on the `astryx-text` span, `.bai-nac-title-area`, `.bai-nac-wrapper`, the table-cell wrapper span and the `<td>` — and it computed to `nowrap` on the bubble, its content div and its content span.

So the text could not wrap, laid out as one long line, and overflowed the 300px box. `overflow` is `visible` on the bubble, so it painted rather than clipped.

This is the same class of defect as **FR-3537** (`text-align: center` leaking from a `<button>` trigger into the bubble), one property over — hence the fix sits on the same theme `components` block, immediately below that pin.

## Measured before/after

Session list, name `mock-chat-b3217a3e-06c1-404f-abd8-d2daebd9af4b`, viewport 1100px.

| | bubble width | text ink width | painted outside bubble | `white-space` | lines |
|---|---|---|---|---|---|
| before — light | 300px | 330px | **+38px** | `nowrap` | 1 |
| before — dark | 300px | 330px | **+38px** | `nowrap` | 1 |
| after — light | 300px | 239px | −53px (inside) | `normal` | 2 |
| after — dark | 300px | 239px | −53px (inside) | `normal` | 2 |

### Before / after

| | Before | After |
|---|---|---|
| Dark | before dark | after dark |
| Light | before light | after light |

## Theme procedure

- `THEME_NAME_REV` 15 → 16 (static recipe changed).
- Artifacts regenerated with `astryx theme build`; `built/index.ts` re-export updated to `bai-r16-default-brand-hc43dtt`; the stale `bai-r15-*.{js,d.ts,variants.d.ts}` deleted.
- No seed-hash entry needed — `ANTD_HOVER_PARITY` is static recipe, not a seed (hash `hc43dtt` unchanged).
- No KEEP-IN-SYNC mirror update needed — no background/surface colour moved, so `index.html`'s critical `<style>` and `resources/webui.css`'s `.splash` are untouched.
- `white-space` is a plain string, not a `[light, dark]` tuple (tuples are colours only).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, tsc, Vite warmup, StyleX sentinel, `astryx theme build --check`, Terminology).
- `pnpm --prefix ./react exec vitest run` → **88 files / 1405 tests passed**, including `backendAiTheme.test.ts` (recipe ↔ built-artifact name parity).
- `pnpm --filter backend.ai-ui exec vitest run` → **47 passed / 1 skipped (48 files), 613 passed / 1 skipped**.
- `pnpm --filter backend.ai-ui build` → clean.
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 1 undeclared-var finding, `packages/backend.ai-ui/src/components/BAIText.css:28`. **Pre-existing and unrelated** — that file is not in this diff (`git diff --name-only HEAD` has zero `BAIText` hits) and the same finding is reported on `main`.
- Live probe on the running app, **both modes**, dark entered via the header toggle with `document.documentElement.dataset.theme === 'dark'` asserted. Zero `pageerror` in both passes. Fresh reload shows no `[astryx-theme] prebuilt brand theme is stale` warning and `data-astryx-theme="bai-r16-default-brand-hc43dtt"`.

## Scope — this is not a session-list fix

The defect is reachable from any tooltip nested in a `nowrap` context, so a theme-level pin closes the whole family rather than one call site. Two read-only sweeps of the codebase found:

- **31 call sites** on the truncation-tooltip path (8 direct `hasTruncateTooltip`, 21 via `BAIText ellipsis={{tooltip:true}}`, 3 via `BAILink ellipsis`), **19 of them high-risk** because their content is long and unbroken: full image references (`ImageList.tsx:330`), session UUIDs (`SessionDetailContent.tsx:386`), SSH public / access / secret keys (`MyKeypairManagementModal.tsx:703-719`), PromQL templates (`PrometheusQueryPresetTable.tsx:161`), startup commands (`AdminDeploymentPresetTable.tsx:184`). Two are the shared table cells `BAINameActionCell` / `BAINameActionCellAstryx`, which multiply across every name column in the app.
- **~175 tooltip render sites** overall; the long-unbroken-content ones (`SessionTemplateModal.tsx:247` image refs, `SessionStatusTag.tsx:129` raw backend error text, `VFolderNameTitle.tsx:31` folder names, `BAITagList.tsx:94` tag lists) share the same failure shape.

Only the session-list case was reproduced and measured; the rest are listed as the surface this pin covers, not as separately verified repros.

## Residue

- **Astryx `Tooltip` still exposes no width knob** (unlike `Popover`, which takes `width`). This PR makes the 300px cap work as intended; it does not add a way for a call site to ask for a wider bubble. A tooltip carrying a genuinely long string now grows in height instead.
- **The upstream asymmetry is untouched**: the bubble background lives on the outer container while the width cap lives on an inner div. Pinning `white-space` removes the only way we found to trigger the mismatch, but the shape remains in `@astryxdesign/core` 0.4.0.
- The migration comment at `SessionInfoCell.tsx:154-156` ("Astryx's truncation tooltip sizes itself") was the assumption that broke. Left in place — it is accurate again now.
- No sibling report is subsumed: FR-3537 and FR-3527 are colour/alignment fixes on the same surface and are already merged; this is linked to both as `relates`.

## Test plan

- [ ] Sessions → hover a session name long enough to truncate: the whole name is inside the bubble, wrapping onto a second line, light **and** dark.
- [ ] A short tooltip (any icon-button label) still renders on one line, unchanged.
- [ ] The cluster-mode help tooltip (FR-3537's surface) is still left-aligned and readable.


yomybaby:
Resolved the merge conflict with `main` and pushed `cd0127192` to this branch.

**What conflicted.** `main` took `THEME_NAME_REV` **16** from #8756 (FR-3536), which emits `bai-r16-default-brand-h161l5oh.*`. This branch also claimed 16, but from a different recipe — so both sides registered the same `data-astryx-theme` identity from different inputs, which is exactly the collision the REV counter exists to prevent (first registration silently wins). Git surfaced it as a rename/rename conflict on all three artifacts plus content conflicts in `backendai-default-built.css` and `built/index.ts`.

**How it was resolved.** Not by hand-picking either side's generated files — a stale `bai-r*` left behind is invisible at review time and wins at runtime. Instead:

- kept both recipes (this branch's tooltip `whiteSpace: 'normal'` pin + everything `main` added),
- rebumped `THEME_NAME_REV` **16 → 17**,
- deleted every superseded `bai-r15-*`/`bai-r16-*` artifact,
- re-ran `pnpm --filter backend.ai-ui build` then `astryx theme build`, emitting `bai-r17-default-brand-h161l5oh.*`,
- repointed `built/index.ts` at the new module.

The hash is unchanged (`h161l5oh`) because `whiteSpace` is not part of the seed-hash array — which is precisely why the REV bump was load-bearing here.

**Verification on the merged tree:** `bash scripts/verify.sh` → `=== ALL PASS ===`, including `astryx theme build --check` ("Theme outputs are up to date"). `white-space: normal` is present once in the regenerated `backendai-default-built.css`. No dependency delta against `main` (`pnpm-workspace.yaml` / `react/package.json` identical — `main` is already on Astryx 0.4.3).

**Heads up on the neighbour:** #8804 (FR-3555) had the same collision and is now resolved at **REV 18**, so the two open theme PRs no longer collide with each other either. Whichever of the two merges second will still need a fresh `astryx theme build` on top of the new `main` — the artifact filenames are derived, so they cannot be merged textually.

A dev server for this branch (post-merge) is running at https://pr-8845.localhost:1357 for visual review.

nowgnuesLee (approved):
LGTM

PR #8854, opened 2026-08-18
fix(FR-3576): size width-less table columns from their content

### LINKED ISSUE #8853: Tables: rich cells (id + tag + actions) truncate because every width-less column gets an equal 1/N share
## Where

Any `BAITableAstryx`. Reported on **Deployment detail → Revision History** and the **Replicas** card, where the `Revision (ID)` column shows `#1 (…` and clips the `Current` tag to `Cu`.

## Observed

A column whose `render` returns rich content — an id + a copy button + a `BAITag` + action buttons — is squeezed to the point that the tag and the actions are unreadable, while neighbouring one-word columns sit half empty.

## Expected

Column width follows what the cell actually renders, so the content is visible by default. Columns stay drag-resizable for anyone who wants them narrower.

## Mechanism

Nothing measures content. A column that declares neither `width` nor `minWidth` gets Astryx `proportional(1)` — a flat 1/N split of the table with only a 120px floor. Measured on the Replicas table at a 1632px table width: all 7 columns exactly 233px each, while their natural content widths are 347 / 110 / 130 / 129 / 361 / 378 / 175 — summing to 1630. The space is already there; it is just handed out equally instead of by need.

The sweep found ~55 production tables in this state, so this is a component-level defect, not a per-page one.

## Scope

Not specific to React-node cells as such — plain-text columns are split equally too. Rich cells simply need the most width, so they are where the flat split shows.

JIRA Issue: FR-3576

agatha197:
Resolves #8853 (FR-3576)

## The mechanism

A `BAITableAstryx` column that declares neither `width` nor `minWidth` was handed Astryx `proportional(1)` — **a flat 1/N split of the table with only a 120px floor**. Nothing measured what a cell renders, so a column packing `#1 (uuid) [copy] [Current] [apply] [⋮]` got exactly the same share as a one-word status column.

The original report guessed that React-node cells are *measured badly*. They are not measured **at all** — and neither is plain text. Rich cells are simply the ones that need the most room, so they are where the flat split shows.

Measured on the deployment-detail **Replicas** table at a 1632px table width, before the fix:

| Column | allotted | natural content width |
|---|---|---|
| Replica ID | 233 | 347 |
| Lifecycle | 233 | 110 |
| Health Status | 233 | 130 |
| Traffic Status | 233 | 129 |
| Session | 233 | 361 |
| Revision (ID) | 233 | 378 |
| Created At | 233 | 175 |
| **total** | **1631** | **1630** |

The space was already there. It was just handed out equally instead of by need.

A sweep of every `BAITableAstryx` call site found **~55 production tables** in this state, so this is a component-level defect, not a per-page one — per-column pixel widths across 55 files would be neither tractable nor maintainable.

## The fix

`columnFit` now defaults to `'content'`. It reuses the **`scroll.x` machinery that antd reached through `scroll={{ x: 'max-content' }}`** rather than opening a second layout path: the table goes `table-layout: auto` with `min-width: 100%`, and the browser sizes each width-less column from what it actually renders — React nodes included. Columns with a numeric or drag-resized `width` stay pixel-pinned exactly as before.

Two bounds ride on the cell's **content wrapper**, because a `<td>`'s own `max-width` is ignored by the automatic table layout while a block child's is not:

- a **400px cap**, so one free-text column (an error message, a start command) cannot take the viewport and starve the rest;
- `column.minWidth`, which `proportional` can no longer carry in this mode (FR-3500).

`columnFit="equal"` restores the old flat split for any table that wants uniform columns.

## Measured after

Same table, same 1632px width — space moved from the three status columns to the three rich ones, with **no horizontal scrollbar introduced**:

| Column | before | after |
|---|---|---|
| Replica ID | 233 | 252 |
| Lifecycle | 233 | 173 |
| Health Status | 233 | 195 |
| Traffic Status | 233 | 197 |
| Session | 233 | 277 |
| Revision (ID) | 233 | 274 |
| Created At | 233 | 264 |

Revision History tab (the reported table), 4 visible columns: `Revision (ID)` **408 → 741**, so the `Current` tag — previously clipped to `Cu` — is fully readable.

### Regression probes on wide tables

| Page | cols | table width | h-scroll | widest column |
|---|---|---|---|---|
| Sessions | 10 | 1632 = viewport | no | 500 |
| Data (folders) | 7 | 1632 = viewport | no | 610 |
| Admin → Users | 21 | 3198 vs 1632 | **yes** | 352 |

Admin → Users already scrolled horizontally before this change (21 columns pinned to the 120px floor = 2452px), so the scrollbar is not new — the columns are simply readable now instead of all being 120px. The 400px cap held everywhere: no column's intrinsic contribution exceeded it.

**Drag-to-resize still works** under the automatic layout: dragging `Revision (ID)`'s handle 200px left moved the columns `[741, 332, …] → [541, 532, …]`.

## Verification

- `pnpm --filter backend.ai-ui build` → built, 0 errors
- `pnpm --filter backend.ai-ui exec vitest run` → **617 passed**, 1 skipped (48 files)
- `pnpm --prefix ./react exec vitest run` → **1405 passed** (88 files)
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 1 undeclared `var(--x)` in `BAIText.css:28`. **Pre-existing** — verified identical with these changes stashed on `HEAD`.
- Live probe on the dev server, **light and dark** (`document.documentElement.dataset.theme === 'dark'` asserted), zero new `pageerror`.

`BAITableAstryx.scrollX.test.tsx` gained coverage for the new default, for `columnFit="equal"`, for explicit `scroll.x` still winning, and for the cap landing on the wrapper of content-sized columns only.

## Screenshots

**Before** — as reported. `Revision (ID)` clips the uuid *and* the `Current` tag (to `Cu`), and the action buttons are squeezed; `Runtime` and `Cluster Mode` sit half empty:

before

**After** — Revision History, light. The `Current` tag and both action buttons are fully readable:

after light

**After** — same table, dark:

after dark

**After** — Replicas card. `#1 (0cc64c83-b… [copy] )` now reads complete instead of `#1 (…`:

after replicas

## Residue

- **`BAIId`'s own `maxWidth: 100`** still truncates the uuid inside the cell (`0cc64c83-b…`). That is deliberate and unchanged — this PR makes the surrounding tag/actions visible, not the full uuid.
- **No call sites were touched.** The three existing `minWidth` declarations keep working (they now ride on the wrapper). The `minWidth: 320` override in `AdminUserManagement.tsx` is now redundant rather than wrong — left in place to keep this diff to the mechanism.
- The 400px cap is a single measured constant, not a per-column knob. A column that genuinely needs more should declare a numeric `width`.

### DIFF
--- a/packages/backend.ai-ui/src/components/Table/BAITable.scrollX.test.tsx
+++ b/packages/backend.ai-ui/src/components/Table/BAITable.scrollX.test.tsx
@@ -17,16 +17,43 @@ describe('BAITable scroll.x', () => {
     expect(layer.style.getPropertyValue('--bai-table-scroll-x')).toBe(expected);
   });
 
-  it('stays off when scroll is absent or carries no x', () => {
-    const { container: withoutScroll } = renderScrollTable();
+  it.each([[undefined], [{ y: 500 }]])(
+    'content-fits by default when scroll carries no x (%s)',
+    (scroll) => {
+      const { container } = renderScrollTable(scroll ? { scroll } : {});
+      const layer = dimLayerOf(container);
+      expect(layer).toHaveClass('bai-table-astryx-scroll-x');
+      expect(layer.style.getPropertyValue('--bai-table-scroll-x')).toBe(
+        'max-content',
+      );
+    },
+  );
+
+  it('stays off when columnFit is equal', () => {
+    const { container: withoutScroll } = renderScrollTable({
+      columnFit: 'equal',
+    });
     expect(dimLayerOf(withoutScroll)).not.toHaveClass(
       'bai-table-astryx-scroll-x',
     );
 
-    const { container: yOnly } = renderScrollTable({ scroll: { y: 500 } });
+    const { container: yOnly } = renderScrollTable({
+      columnFit: 'equal',
+      scroll: { y: 500 },
+    });
     const layer = dimLayerOf(yOnly);
     expect(layer).not.toHaveClass('bai-table-astryx-scroll-x');
     expect(layer.style.getPropertyValue('--bai-table-scroll-x')).toBe('');
+  });
+
+  it('explicit scroll.x still wins over the columnFit default', () => {
+    const { container } = renderScrollTable({
+      columnFit: 'equal',
+      scroll: { x: 800 },
+    });
+    const layer = dimLayerOf(container);
+    expect(layer).toHaveClass('bai-table-astryx-scroll-x');
+    expect(layer.style.getPropertyValue('--bai-table-scroll-x')).toBe('800px');
   });
 
   it('releases max-width on auto columns only', () => {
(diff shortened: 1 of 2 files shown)


nowgnuesLee (changes requested):
I think using the `scroll` option would be the right approach here. Is there a specific reason this PR is needed?

agatha197:
> I think using the `scroll` option would be the right approach here. Is there a specific reason this PR is needed?

User cannot see the whole contents as a default. For example, user cannot know the current revision.
image

agatha197:
Not planned.

PR #8855, opened 2026-08-18
fix(FR-3575): make the folder-explorer upload dropzone fill the overlay

### LINKED ISSUE #8852: Folder explorer: the drag-and-drop upload overlay no longer fills the explorer, and its caption covers the title
## Where

- **Page**: Data → folder explorer (the Explorer modal/panel opened from a folder row)
- **Region**: the drag-and-drop upload overlay that appears while dragging files over the explorer
- **Component** (if identified): `DragAndDrop` (`packages/backend.ai-ui/src/components/baiClient/FileExplorer/DragAndDrop.tsx`), rendered by `BAIFileExplorer` while `isDragMode` is on

## Observed

Dragging files over the folder explorer no longer covers the whole explorer area with one dropzone.

- The blurred backdrop does span the full area, but the dashed dropzone box inside it only occupies a band near the top — the lower part of the explorer stays blurred but is not part of the visible dropzone.
- The dropzone's caption text is rendered **above** the box, on top of the explorer/modal title, in two stacked copies of the same sentence ("Drag and drop files to this area to upload." as a bold heading and again as smaller helper text). The heading sits over the title row and makes it unreadable.

## Expected

The same shape as before the Astryx migration: a single dropzone that fills the entire explorer area edge to edge, with the upload icon and the caption centered **inside** the box, and nothing painted over the explorer title.

## Environment

- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: branch `FR-3554` (post `to-astryx`, on top of `main`)

## Reproduction

1. Open Data and enter a folder's file explorer.
1. Drag one or more files from the OS file manager over the explorer.
1. Observe the upload overlay while still dragging.

## Evidence

- Screenshot: `/home/ubuntu/.claude/jobs/aae93a67/pasted-1.png` (to be attached in Jira)

## Notes

Pre-migration the overlay was a single antd `Upload.Dragger` that itself carried `position: absolute; inset: 0`, so the dropzone chrome was the full-area element; the caption lived inside it.

JIRA Issue: FR-3575

agatha197:
Resolves #8852 (FR-3575)

Dragging files over the folder explorer showed a dropzone that covered only a shallow band near the top of the explorer, with two copies of its caption painted over the explorer title.

## Mechanism

The antd `Upload.Dragger` this replaced (`to-astryx` W2-D) was **itself** the `position: absolute; inset: 0` element, so the dropzone chrome was the full-area box and the icon + caption lived inside it.

The migration moved the positioning onto a new wrapper `<div>` and left `FileInput` with only `width="100%"`:

- Astryx's `FileInput` has no height knob, so the drop area collapsed to its intrinsic **100px** and sat **50px below** the overlay top — the remaining 552px stayed blurred but was not part of the dropzone.
- `label` and `description` render in the surrounding `Field`, **above** the drop area, not inside it. Both were set to the same string, which is the two stacked captions over the title.

The fix routes the caption through `placeholder` (which `renderDropzoneContent()` puts inside the box, next to the upload icon), keeps `label` for accessibility behind `isLabelHidden`, and moves the positioning onto `FileInput`'s `style` — Astryx's `mergeProps` applies `style` to the dropzone element itself, so the drop area is the full-area box again.

## Measured before/after

Explorer overlay box: `1152x652 @64,34` in both passes.

| | Dropzone box | Covers overlay | Caption location |
|---|---|---|---|
| Before | `1152x100 @64,84` | 15% (100 of 652px), offset 50px down | `Field`, above the box, ×2 — over the explorer title; box read "Choose files" |
| After | `1152x652 @64,34` | 100% — exact match | inside the box, ×1, centered under the icon |

Identical in light and dark (`document.documentElement.dataset.theme === 'dark'` asserted after entering dark through the header button).

## Screenshots

Captured on the current head of this branch, on `/project/default/data?folder=65af57bb-…`. A stray Astryx tooltip left over in the app header is suppressed in these shots so it does not read as part of the overlay.

**Before** — the dropzone is a shallow band at the top; the caption renders twice in the `Field` above it, over the explorer title, and the box itself only says "Choose files".

| Light | Dark |
|---|---|
| before light | before dark |

**After** — one dropzone filling the explorer, caption and icon centered inside it, nothing over the title.

| Light | Dark |
|---|---|
| after light | after dark |

## Second defect — the overlay, the drop, and the upload

Reported after the first fix landed: dropping a file made a *second* upload prompt appear, and then that uploads stopped working entirely. Both come from the same place — **who closes the overlay, and when**.

Astryx's `FileInput` calls `e.stopPropagation()` on its own `drop` and `dragleave`. `BAIFileExplorer` listened for both on `document`, which produced three distinct failures depending on the phase:

| Wiring | Symptom |
|---|---|
| `drop` on **bubble** (original) | `FileInput` stops it, `document` never sees it, `isDragMode` never clears — the overlay stays up and reads as a second upload prompt. Latent before this PR: the dropzone was a 100px band, so most drops missed it, bubbled, and closed the overlay — while also never uploading. |
| `drop` on **capture** | `drop` is discrete priority, so React flushes the state update *mid-dispatch*. The dropzone unmounts before React reaches `FileInput`'s handler and the file is silently discarded — **no upload request at all**. |
| `dragleave` on **capture** | The browser fires relatedTarget-less `dragleave`s as the cursor moves inside the dropzone. `FileInput` stopping them is what shielded the handler; on capture they close the overlay mid-drag, leaving nothing to drop onto. |

Final shape: the document listener leaves `drop` alone and keeps `dragleave` on bubble; `useDragOverlay` exposes `close`; and `DragAndDrop`'s `onDropCapture` calls it from a `setTimeout`, i.e. after the event finishes. `FileInput` takes the file first, the tus upload runs, the overlay closes right after.

### Measured with a real browser drag

Not synthetic DOM events — CDP `Input.dispatchDragEvent` with a real file, on `/project/default/data?folder=65af57bb-…`:

| | overlay on dragEnter | survives dragOver | overlay after drop | file in folder |
|---|---|---|---|---|
| `drop` bubble | shown | yes | **still shown** | no |
| `drop` capture | shown | yes | gone | **no** (`files=1` reached the dropzone; it was already unmounted) |
| `dragleave` capture | shown | **no — dies mid-drag** | n/a | no |
| Final | shown | yes | gone | **yes** |

The trace that pinned it: `["capture files=1"]` with no following `onChange` when closing synchronously, versus `["capture files=1", "onChange n=1"]` and `uploaded: true` when deferred.

Test uploads were deleted afterwards; both folders are back to their original contents.

### Tests

- `useDragOverlay.test.tsx` (new, 8 cases) pins each listener phase against a stand-in that stops propagation the way `FileInput` does. Verified it fails on the broken wiring, not just passes on the fixed one.
- `e2e/vfolder/file-upload-dnd.spec.ts` was **stale** — written against `.ant-modal-body` / `.ant-upload-drag`, which no longer exist, and it bypassed the drop entirely via `setInputFiles` on the hidden input, so it could not have caught any of this. Rewritten onto the Astryx dropzone, with cursor movement inside the zone and a real `drop`. Not executed here: the e2e suite's global teardown sweeps leftover vfolders on the shared cluster, which is not safe to run ad hoc from this branch.

## Verification

- `pnpm --filter backend.ai-ui build` → built in 23.06s
- `pnpm --filter backend.ai-ui exec vitest run` → 613 passed, 1 skipped (48 files)
- `pnpm --prefix ./react exec vitest run` → 1405 passed (88 files)
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 1 undeclared `var()`, `BAIText.css:28` — **pre-existing on `main`**, untouched by this PR (the only file changed here is `DragAndDrop.tsx`)
- Live probe on `https://fr-3575.localhost:1336`, light and dark, zero `pageerror` in both passes
- `pnpm --filter backend.ai-ui exec vitest run` re-run after the drag work → 621 passed, 1 skipped (49 files)
- Real-drag probe (CDP) after the final commit: overlay opens once, survives the drag, closes on drop, file lands in the folder, zero `pageerror` (the drop path is theme-independent)

## Residue

- The de-dupe guard `fileList !== lastFileListRef.current` is carried over unchanged. `onChange` hands back a fresh array each call, so the comparison is always true — inert, and out of scope here.
- The dropped-`directory` PILOT-DECISION stands: picking a whole folder from the file dialog is still unavailable. Dropping a folder works (that path is `DataTransfer`).
- This does not touch FR-3516 (the explorer's resizable file-list/metadata split) or FR-3517 (the metadata Path row); both are separate folder-explorer reports.

## Third defect — a cancelled drag left the overlay stuck

Reported in review (@yomybaby): dragging a file in and then cancelling — Escape, or dragging back out of the window — left the overlay up for good, with the explorer unusable behind it.

Same root as the rest: the dropzone now fills the overlay, and Astryx `FileInput` calls `stopPropagation()` on its own `dragleave`, so every leave that ends the drag fires on the dropzone and never reaches the `document` listener. While the dropzone was a 100px band most leaves missed it and bubbled, which is why this surfaced only after the resize.

Measured on Chromium with CDP-driven drags against the running app:

| | what the page receives |
|---|---|
| drag leaves the window | `dragleave` on the dropzone, `relatedTarget` null — swallowed on bubble |
| pointer moves inside the page | `dragleave` always names the entered element in `relatedTarget` |
| drag cancelled | **nothing** — no `dragleave`, no `dragend`, and `dragover` does not auto-repeat while the pointer is still |

That third row is why a heartbeat/timeout cannot work, and it corrects an assumption in the "Second defect" section above: the relatedTarget-less mid-drag `dragleave` described there is not what Chromium sends.

Two changes:

- `dragleave` moves to the **capture** phase, where the dropzone cannot hide it. The existing `relatedTarget` guard was already correct — it just never ran.
- For the cancel case, mouse events stay suppressed for the whole drag and resume the moment it ends, so the first `mousemove` while the overlay is up is proof the drag is over. That listener is scoped to while the overlay is open.

`drop` deliberately stays on bubble, for the reason given above.

### Verified

7/7 live browser checks pass; the two cancel checks fail without the change (overlay stays up) and a real drop still uploads. The two new `useDragOverlay` tests likewise fail on the old wiring. The e2e spec's synthetic mid-drag leaves now carry a `relatedTarget`, matching what the browser sends, and the window-exit and cancel paths are covered as their own steps.

- `pnpm --filter backend.ai-ui exec vitest run` → 622 passed, 1 skipped (49 files)
- `pnpm --prefix ./react exec vitest run` → 1405 passed (88 files)

A separate defect found while testing this — dismissing the upload **file chooser** closes the whole modal, via `<input type="file">`'s bubbling `cancel` — is fixed in #8868 (FR-3579), not here.

## Third defect — the cancel dismissal broke the upload

Reported on the review thread: after the fix above, dropping a file no longer
uploaded it — the browser opened the file in a new tab instead — and the drop
hint read "Drop files here" in a Korean session.

### The upload

The `mousemove` dismissal added in the previous section rests on "mouse events
stay suppressed for the whole drag". That was measured with **CDP-driven
drags**, which emit no mouse events at all, so the measurement could not
observe the case it was asserting. With a real pointer the assumption does not
hold, and one `mousemove` mid-drag unmounts the dropzone — the drop then has
nothing to land on, no upload runs, and the browser falls back to its default
action.

Measured against the running app, CDP drag plus an explicit
`Input.dispatchMouseEvent`, on `/project/default/data?folder=20c7799d-…`:

| | overlay before | after mousemove | uploads on drop | overlay after drop |
|---|---|---|---|---|
| Before | 1 | **0** | **0** | — |
| After | 1 | 1 | 3 (`request-upload` + tus `PUT` + relist) | 0 |

A cancelled drag still fires no drag event, so it still needs a non-drag
signal — but one the browser **cannot** deliver mid-drag. `mousedown` /
`keydown` / `wheel` all require the button to have been released first, so the
overlay now comes down on the user's next interaction rather than on the next
pointer twitch.

### The English drop hint

`FileInput` swaps its caption to its own `dropHint` string the moment the
pointer is over the drop area, so the `placeholder` this PR routes the caption
through is only visible for the instant before that — what the user actually
reads is Astryx's string, not ours. And the eight `@astryx.fileInput.*` strings
FileInput paints inside the drop area were never in our override catalogs (en
284 keys vs 250 per locale); only `clearLabel`, `required` and
`triggerWithFiles` were, so the rest fell through to the shipped `en`.

`@astryxdesign/core@0.4.0` **hardcoded** all eight
(`isDragOver ? 'Drop files here' : displayPlaceholder`), so no catalog could
have reached them. They route through `t()` from **0.4.1** upstream, and #8857
(FR-3577) moved us 0.4.0 → **0.4.3** on main. **This branch is rebased onto that**, so filling the
group in all 20 catalogs is what closes the report.

Measured in a Korean session on the running app:

| | caption |
|---|---|
| after `dragEnter` | `파일을 이곳에 드래그하여 업로드하세요.` (this PR's `placeholder`) |
| while the pointer is over the drop area | `여기에 파일을 놓으세요` (was `Drop files here`) |

The remaining 26 missing keys (listInput, dateInput, numberInput,
commandPalette, step, sideNavItem) are untouched — a separate sweep.

### Why the e2e spec did not catch either

`file-upload-dnd.spec.ts` dispatches **synthetic** `DragEvent`s at `document`.
Those never reach `FileInput`'s own handlers, so `isDragOver` stays false —
the spec asserts the `placeholder` caption that a real drag never shows, and it
cannot observe a `mousemove` the browser only emits during a real drag. Both
defects are invisible to it by construction.

### Verified

- `pnpm --filter backend.ai-ui exec vitest run` → 628 passed, 1 skipped (49 files); `useDragOverlay.test.tsx` 12 passed, and its mousemove-tolerance test fails on the old wiring
- Live drag on `https://fr-3575.localhost:1336`, Korean session: caption reads `여기에 파일을 놓으세요`, the overlay survives 5 `mousemove`s, the drop uploads (3 requests), and the overlay closes after it

## The accent wash at overlay scale

Reported after the drop path was working: with the dropzone now filling the
explorer, dragging a file tinted the whole file list orange.

### Mechanism

Astryx paints a dropzone's drag-over state from the brand accent
(`FileInput.tsx`, `dropzoneActive`):

```
borderColor:     var(--color-accent)        // #FF7A00 light / #be5e06 dark
backgroundColor: var(--color-accent-muted)  // color-mix(accent 20%, transparent)
```

Nothing here is misconfigured — the fill is simply **area-dependent**. Astryx's
own two templates (`FileInputBasic`, `FileInputShowcase`) render at 350px wide,
and neither passes `mode="dropzone"` at all, so a 20% accent fill has never been
exercised at the size this PR gave it. Stretched over a 1152x652 overlay the same
value stops reading as a hint on a control and starts reading as a filter over
the content, which the `backdropFilter: blur(4px)` behind it amplifies.

### The fix

The border is a line — its cost does not scale with the area — so it keeps the
accent and stays the drop cue. Only the fill is neutralized, by re-defining the
custom property on the dropzone's own class:

```css
.bai-file-explorer-dropzone {
  --color-accent-muted: var(--color-background-muted);
}
```

Two things that choice buys:

- **It does not fight StyleX.** Astryx's compiled CSS lives in
  `@layer astryx-base` and literally emits `background-color: var(--color-accent-muted)`,
  so redefining what the property resolves to on that element wins without a
  specificity or `!important` contest. This is the same technique the theme
  already uses for `.astryx-banner.info`.
- **It stays scoped to this call site.** The repo's other `mode="dropzone"` —
  `BulkCreateUserFromCSVModal` — is an inline-scale field inside a modal, where
  the accent feedback is correct. A theme-level `components: { 'file-input': … }`
  override would have neutralized that one too.

`--color-background-muted` is translucent in **both** modes
(`rgba(0,0,0,0.04)` / `rgba(255,255,255,0.08)`), unlike `--color-overlay-hover`,
whose dark value is the opaque `#262626` and would have hidden the file list.

| | Drag-over fill | Border |
|---|---|---|
| Before | `--color-accent-muted` — 20% #FF7A00 over the whole explorer | accent, dashed |
| After | `--color-background-muted` — neutral, translucent both modes | accent, dashed (unchanged) |

### Verified

- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 277 declared
  properties, 898 `var()` usages, 1 undeclared: `BAIText.css:28` — the same
  pre-existing entry reported above, untouched here
- pre-commit `lint-staged` (lint + format) passed on both changed files
- Confirmed on the running app by the reporter: the tint is gone

**Not re-run for this commit** — `scripts/verify.sh`, the vitest suites, and
`pnpm --filter backend.ai-ui build` were skipped at the author's request, so the
counts in the sections above predate these two files. The change is a 3-line CSS
rule plus its `import`, with no TypeScript surface; CI covers the BUI build that
carries the rule into the production bundle.


e2e/vfolder/file-upload-dnd.spec.ts
```diff
@@ -98,64 +98,79 @@ test.describe(
...
+        );
       });
 
-      // 5. Verify the drag-and-drop overlay appears with text "Drag and drop"
-      const dragOverlay = page.locator('.ant-upload-drag');
+      // 5. The Astryx dropzone covers the explorer and carries the caption
+      //    inside it (FR-3575).
+      const dragOverlay = page.locator(
+        '.astryx-file-input[data-mode=dropzone]',
+      );
```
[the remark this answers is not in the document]
agatha197: Fixed in 59d69be3e.

You are right that `.astryx-file-input` violates the selector policy in `.github/instructions/e2e.instructions.md` ("Never against a framework-internal class").

The overlay wrapper — which this repo owns — now carries `data-testid="folder-explorer-drop-overlay"`, and the spec locates that.

One constraint worth recording: the drop area itself cannot take a `data-testid`. Astryx `FileInput` spreads unknown props onto its hidden `<input>`, and only `className` / `style` are merged onto the dropzone element. So the dropzone gets the repo-owned `bai-file-explorer-dropzone` class instead, which is the policy's sanctioned fallback ("only against a class the repo itself owns"). It is scoped inside the testid rather than queried globally.

e2e/vfolder/file-upload-dnd.spec.ts
```diff
@@ -98,64 +98,79 @@ test.describe(
...
       await expect(dragOverlay).toBeVisible({ timeout: 5000 });
-
-      // Verify the overlay text
       await expect(
-        page.getByText('Drag and drop files to this area to upload.'),
+        dragOverlay.getByText('Drag and drop files to this area to upload.'),
       ).toBeVisible();
```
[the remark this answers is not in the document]
agatha197: Fixed in 59d69be3e. Good catch — the test was asserting the symptom's neighbourhood rather than the regression itself.

The spec now compares bounding boxes before the drop, on all four sides:

```ts
const overlayBox = await dropOverlay.boundingBox();
const dropzoneBox = await dragOverlay.boundingBox();
for (const side of ['x', 'y', 'width', 'height'] as const) {
  expect(
    Math.abs(dropzoneBox![side] - overlayBox![side]),
    `dropzone ${side} should match the overlay`,
  ).toBeLessThanOrEqual(1);
}
```

Measured live on the branch, this is the exact delta the ticket is about:

| | dropzone | overlay |
|---|---|---|
| before the fix | `1152x100 @64,84` | `1152x652 @64,34` |
| after | `1152x652 @64,34` | `1152x652 @64,34` |

So a collapse back to the 100px band now fails on `height` and `y`, not just silently passes.

packages/backend.ai-ui/src/components/baiClient/FileExplorer/DragAndDrop.tsx
yomybaby, line 1: If canceled while dragging a file into the area using Drag & Drop, the view gets stuck as follows.

Image

yomybaby (changes requested):

packages/backend.ai-ui/src/components/baiClient/FileExplorer/DragAndDrop.tsx
agatha197: Fixed in 26fb2738d.

The cancel path had no handler at all. Astryx `FileInput` calls `stopPropagation()` on its own `dragleave`, and this PR made the dropzone fill the overlay — so every leave that ends the drag now fires on the dropzone and never reaches the `document` listener. Before the resize the dropzone was a 100px band, so most leaves missed it and bubbled, which is why this surfaced only now.

Measured on Chromium with CDP-driven drags against the running app, rather than assumed:

| | what the page receives |
|---|---|
| leave the window | `dragleave` on the dropzone, `relatedTarget` null — swallowed on bubble |
| move inside the page | `dragleave` always names the entered element in `relatedTarget` |
| cancel (Escape / drag back out) | **nothing** — no `dragleave`, no `dragend`, and `dragover` does not auto-repeat while the pointer is still |

So two changes, not one:

- `dragleave` moves to the **capture** phase, where the dropzone cannot hide it. The existing `relatedTarget` guard was already correct — it just never ran.
- For the cancel case there is no drag event left to react to, and a timeout cannot see it either. Mouse events stay suppressed for the whole drag and resume the moment it ends, so the first `mousemove` while the overlay is up is proof the drag is over. That listener is scoped to while the overlay is open.

`drop` deliberately stays on bubble — a capture listener there flushes the close synchronously and unmounts the dropzone before React reaches `FileInput`'s handler, discarding the file.

Verified live on the branch: 7/7 checks pass, and the two cancel checks fail without the change (overlay stays up). A real drop still uploads. The two new hook tests likewise fail on the old wiring. `scripts/verify.sh` → ALL PASS; `backend.ai-ui` vitest → 622 passed, 1 skipped.

yomybaby: Please check once again. This time, when I drop (a file), instead of the file being uploaded, the file opens in a new tab. And even though it’s being displayed in Korean, it says “drop files here.”

yomybaby (changes requested):

packages/backend.ai-ui/src/components/baiClient/FileExplorer/DragAndDrop.tsx
agatha197: Both are fixed — `2bf7f740e` and `dfec8bff1`, pushed after your report.

**The file opening in a new tab.** The previous fix dismissed a cancelled drag on the first `mousemove` while the overlay was up, resting on a measurement taken with CDP-driven drags. Those emit no mouse events at all, so the measurement could not observe the case it was asserting. With a real pointer the assumption does not hold: one stray `mousemove` mid-drag unmounted the dropzone, the drop then had nothing to land on, no upload ran, and the browser fell back to its default action and opened the file.

The dismissal signal is now `mousedown` / `keydown` / `wheel` — none of which the browser can deliver before the button is released, so none can fire mid-drag. The overlay comes down on your next interaction instead of on the next pointer twitch. A cancelled drag still fires no drag event, so it still needs a non-drag signal; it just has to be one that cannot arrive while the drag is in flight.

Measured on the running app with a CDP drag plus an explicit `Input.dispatchMouseEvent`:

| | overlay before | after mousemove | uploads on drop |
|---|---|---|---|
| Before | 1 | **0** | **0** |
| After | 1 | 1 | 3 (`request-upload` + tus `PUT` + relist) |

The overlay still closes on the drop itself (asserted 0 after).

**"Drop files here" in a Korean session.** `FileInput` swaps its caption to its own `dropHint` the moment the pointer is over the drop area, so what you read there is Astryx's string, not the `placeholder` this PR routes our caption through — ours is only visible for the instant before that. The eight `@astryx.fileInput.*` strings it paints inside the drop area were missing from our override catalogs (en 284 keys vs 250 per locale); only `clearLabel`, `required` and `triggerWithFiles` were present.

Worth noting that filling them earlier would not have helped: `@astryxdesign/core@0.4.0` **hardcoded** all eight (`isDragOver ? 'Drop files here' : displayPlaceholder`), so no catalog could have reached them. They route through `t()` from **0.4.1** upstream, and #8857 (FR-3577) moved us 0.4.0 → 0.4.3 on main, which this branch is rebased onto. `dfec8bff1` fills the group in all 20 locales — a Korean session now reads `여기에 파일을 놓으세요`.

The remaining 26 missing keys (listInput, dateInput, numberInput, commandPalette, step, sideNavItem) are untouched here — a separate sweep.

yomybaby (approved):
LGTM

nowgnuesLee (approved):
LGTM

PR #8857, opened 2026-08-18
chore(FR-3577): update Astryx to 0.4.3 and move the pins into the pnpm catalog

### LINKED ISSUE #8856: Update Astryx to 0.4.3
Bump `@astryxdesign/core`, `@astryxdesign/theme-neutral` and `@astryxdesign/cli` from 0.4.0 to 0.4.3 (latest stable, published 2026-08-16).

Follow-up to FR-3541 (0.4.0 adoption).

Scope:

- Bump the three pinned versions in `react/package.json`, `packages/backend.ai-ui/package.json` and `pnpm-workspace.yaml`.
- Refresh the `minimumReleaseAgeExclude` entries (0.4.3 is inside the 7-day quarantine window until 2026-08-23).
- Run `astryx upgrade --apply` and rebuild the theme.
- Re-sync the generated ASTRYX block in `AGENTS.md` / `react/AGENTS.md` / `CLAUDE.md`.
- `@astryxdesign/lab` stays pinned at 0.3.0-canary.12db2a1 — there is still no stable lab release.

JIRA Issue: FR-3577

yomybaby:
Resolves #8856 (FR-3577)

Astryx 0.4.3 was published on 2026-08-16. This bumps `@astryxdesign/core`, `@astryxdesign/theme-neutral` and `@astryxdesign/cli` from 0.4.0, and — as a second, separable change — moves every `@astryxdesign` pin into the pnpm catalog.

Follow-up to #8772 (FR-3541), which adopted 0.4.0.

## Changes

- **Version bump**: 0.4.0 → 0.4.3 across `react/package.json`, `packages/backend.ai-ui/package.json` and `pnpm-workspace.yaml`.
- **Catalog**: the four `@astryxdesign` specs (`core`, `theme-neutral`, `cli`, and the `lab` canary) now live in the workspace catalog and the manifests say `catalog:`. The version was previously duplicated across five declarations in two manifests, which is how a partial bump becomes possible — `theme-neutral` peer-requires the *exact* `core` version, so a half-updated set fails at install rather than at review.
- **`minimumReleaseAgeExclude`**: swapped the 0.4.0 entries for 0.4.3 ones. 0.4.3 is inside the 7-day quarantine window; remove after **2026-08-23**. Also refreshed the stale `theme-neutral@0.4.0` reference in the `lucide-react` catalog comment (its `lucide-react ^1.18.0` requirement is unchanged at 0.4.3).
- **Codemods**: none. `astryx upgrade --from 0.4.0` reports `No codemods available for this version range` — 0.4.1–0.4.3 are patch releases with no API renames. No source file changed.
- **Agent docs**: `react/AGENTS.md` refreshed by `astryx upgrade --apply` to the v0.4.3 block; root `AGENTS.md` (= `CLAUDE.md`, a symlink) re-synced by hand to match. The upgrade regenerates the block from scratch and drops the project-specific **MIGRATION RELAXATION** line, so it was restored in both copies — worth knowing for the next bump.
- **Theme artifacts**: rebuilt `react/src/astryx-theme/built/*` with the 0.4.3 CLI. The theme name is unchanged (`bai-r15-default-brand-hc43dtt` — recipe and seeds untouched), so `built/index.ts` needed no edit and there are no orphaned artifacts to delete.

## The one visual change: Banner corners

The theme rebuild produces a real, intentional rendering change, and it is the only one I could measure.

**Banner (and therefore `BAIAlert`) goes from square corners to `border-radius: 8px`.**

0.4.3 fixes a theme reaching the banner's radius: `--_banner-radius` was declared in the docs and in `derivedVarRegistry.ts`, but no rule read it, so a theme's `borderRadius` on the `banner` target expanded into a variable nothing consumed. The four card-silhouette radii read it now, falling back to `--radius-container` — and this project's built theme sets `--radius-container: 8px` (`backendai-default-built.css:246`). So banners now pick up the same container radius every other surface already had.

The generated CSS also rewrites the info banner's fill from a direct `background-color` plus `--color-accent-muted: transparent` to `--color-accent-muted: var(--color-background-blue)`. Upstream did this because under stock `theme-neutral` the info banner painted nothing (a plain CSS property written by a theme lands in `@layer astryx-theme`, which StyleX's `@layer priority4` outranks). **For this project it is a no-op** — measured, not inferred.

### How that was measured

Rather than reason about the cascade, I rendered the real `BAIAlert` markup (all four statuses, via `@testing-library/react`, so the StyleX class list is the app's own) into a static page and loaded it in headless Chromium twice: once against `@astryxdesign/core@0.4.0` + the previous built theme CSS, once against 0.4.3 + the rebuilt one. Light and dark, 760×620 @2x.

Computed `background-color` on `.astryx-banner.info`:

| | 0.4.0 | 0.4.3 |
|---|---|---|
| light | `rgb(196, 221, 251)` | `rgb(196, 221, 251)` |
| dark | `rgba(158, 183, 255, 0.24)` | `rgba(158, 183, 255, 0.24)` |

Identical, as are the icon colors, and as are all four statuses' `font-size`, `line-height`, `font-weight`, `color`, padding and every element bounding box.

A full-page pixel diff of the two screenshots is **1335 / 1,884,800 px (0.07%)** in light and 1091 px in dark. Those pixels group into exactly 8 bands, which line up with the top and bottom edges of the 4 banners — i.e. the corner arcs, and nothing else. (A ninth light-mode band is the `0.4.0`/`0.4.3` caption I put on the probe page itself.)

## Not changed

`@astryxdesign/lab` stays on `0.3.0-canary.12db2a1`. There is still no stable `lab` release (`latest` on npm is `0.0.0-bootstrap.0`), and the 0.4.x canary line peer-requires a canary `core` build, so it cannot match a stable core either way — the same accepted mismatch as at 0.3.0 and 0.4.0 adoption (MIGRATION-SPEC §1-④).

## Catalog change is contract-neutral

`backend.ai-ui` is published to npm and declares `@astryxdesign/*` as peer dependencies, so `catalog:` must not leak into the published manifest. `pnpm pack` on this branch writes:

```json
"peerDependencies": {
  "@astryxdesign/core": "0.4.3",
  "@astryxdesign/theme-neutral": "0.4.3",
  ...
}
```

i.e. the same shape as before, resolved. `pnpm install` after the catalog switch reported `Already up to date` — no resolved version moved.

## Worth a reviewer's eye

Neither of these is changed by this PR; both are places where 0.4.1–0.4.3 moved something underneath existing hand-written CSS, and `tsc` cannot see either.

- `packages/backend.ai-ui/src/components/BAIAlert.css:57` — `.bai-alert .astryx-banner { align-items: flex-start }`, a deliberate specificity fight against StyleX. It still matches (the class is still on the same element in the probe markup), but it is the rule most exposed to a future Banner restructure.
- `packages/backend.ai-ui/src/components/BAIColorPicker.css:44` — a hand-rolled `:focus-visible` 2px inset ring. 0.4.1 made the focus ring themeable (`--focus-outline-*`) and standardized it at 3px offset; this project sets none of those tokens, so this one control is now the only ring in the app that does not match the system one. Left alone here to keep the PR a dependency bump.

Other surfaces I checked and found clean: no `BannerStatusMap` augmentation, no `--focus-outline-*` override, no CSS targeting the popup/context-menu theme targets that 0.4.1–0.4.2 relocated, no `useContainerReveal` or `TreeList` usage, and no `TextArea` call site that passes `maxLength` (so 0.4.3's grapheme-aware character counter has no live consumer yet).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / Vite warmup / StyleX / Astryx theme `--check` / Terminology)
- `vitest`: `backend.ai-ui` 613 passed | 1 skipped · `react` 1405 passed · root 98 passed | 6 skipped
- `pnpm pack` on `backend.ai-ui` — published `peerDependencies` unchanged (above)
- Headless-Chromium before/after probe of all four Banner statuses, light and dark (above)


nowgnuesLee (approved):
LGTM

PR #8875, opened 2026-08-19
feat(FR-3558): gate the global search palette behind an experimental setting

### LINKED ISSUE #8807: Global search palette (Cmd+K) with a build-time page/tab/setting index

nowgnuesLee:
Resolves #8807 (FR-3558)

## What

Puts the global search palette behind an experimental user setting,
`experimental_global_search`, which ships **off**.

- `GlobalSearchPaletteButton` reads the flag and, while it is off, renders no
  trigger and feeds `useHotkeys` an empty list — so `mod+k` stays with the
  browser instead of being bound to a disabled handler.
- User Settings → General → *Experimental features* gains a **Global search**
  checkbox alongside AI Agents and Import from Hugging Face.
- `userSettings.GlobalSearch` is translated in all 21 locales, reusing each
  file's existing word for "search". No `__NOT_TRANSLATED__` markers.
- The new setting item is picked up by the build-time search index, so the
  toggle is itself findable from the palette once enabled.

## Why

The palette is a new, prominent header affordance that changes a global
keyboard shortcut. Gating it lets the feature land in the tree and be dogfooded
without changing the header for every user on day one, and gives us a per-user
off switch if `mod+k` collides with something in the field.

The gate lives in the button rather than at its `WebUIHeader` mount point so it
travels with the component: any future mount is gated by construction, and the
hotkey registration — which lives inside the button — disappears with it.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, search index, lint,
  format, tsc, StyleX, Astryx theme build, terminology).
- `pnpm vitest run src/components/GlobalSearchPalette src/components/MainLayout/WebUIHeader.test.tsx`
  → 8 files, 87 tests passing, including two new contracts: the trigger renders
  while the setting is on, and with it off there is no trigger and neither
  `Control+k` nor `Meta+k` opens a dialog.
- `react/src/generated/searchIndex.json` regenerated and committed
  (`pnpm run search-index`); verify.sh's drift gate is clean.

## Stack

Sits between `FR-3558-search-core` (#8812) and `FR-3558-e2e` (#8864); the e2e
spec enables the flag via `page.addInitScript` in its `beforeEach`.

## Review fixes

- **The "binds no mod+k while the setting is off" test could not fail.** It
  asserted the absence of a dialog, which the render gate already guarantees —
  the assertion held whether or not the hotkey was registered. Astryx
  `useHotkeys` calls `preventDefault()` *before* its handler, so
  `defaultPrevented` on a dispatched cancelable ⌘K is the registration's only
  observable trace. Both directions now assert on it: flag off → `false` (the
  browser keeps its own ⌘K), flag on → `true`. `navigator.platform` is pinned to
  Apple for the suite because jsdom reports none, which would resolve `mod` to
  ctrl.
- **The comment above `useHotkeys` overstated what the empty array does.** The
  hook attaches its window listener unconditionally; `[]` avoids *matching*, not
  *registering*. Reworded.
- **`isOpen` survived a toggle-off.** The gate unmounted the palette but left
  the flag set, so re-enabling the setting would have re-opened the palette
  unprompted. Cleared while rendering the disabled branch (an effect trips
  `react-hooks/set-state-in-effect`). The gate stays in the button, not hoisted
  to the header. Covered by a new test that fails without the fix.

Verified after the fixes and the trunk sync: `bash scripts/verify.sh` →
`=== ALL PASS ===`; `pnpm --prefix react exec vitest run
src/components/GlobalSearchPalette` → 7 files, 78 tests passing.

Renamed the toggle from "Global search" to "Command palette" (`userSettings.GlobalSearch` → `userSettings.CommandPalette` across all 21 locales, testid `items-experimental-command-palette`, search index regenerated); the `experimental_global_search` settings key, the component names, and the header tooltip (`webui.menu.Search`) are unchanged.


agatha197 (approved):
LGTM

ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3558-pr8875-gate-global-search.sungchul.10-82-0-159.sslip.io)

PR #8889, opened 2026-08-19
fix(FR-3591): restore the full drag height of the folder explorer's resize handle

### LINKED ISSUE #8888: ResizeHandle: only the top ~40% of the divider is draggable, but the resize cursor covers its whole height
## Component

- **Name**: Astryx `ResizeHandle` (`@astryxdesign/core/Resizable`, v0.4.3)
- **Astryx primitive**: `ResizeHandle` (the `astryx-resize-handle` separator and its internal grab zone)
- **Scope**: every usage — the defect is in the component's own styles, not in a call site

## Where it is reachable

- The folder explorer modal, opened from Data (menu: Data → a folder row → folder name) — the divider between the file list and the metadata panel, at `xl` and wider
- Also rendered by: `react/src/components/FolderExplorerModalV2.tsx`, `react/src/components/FolderExplorerModal.tsx` (V1, currently unreachable)

## Steps to reproduce

1. Open Data and click a folder name to open the folder explorer modal, wide enough for the two-pane layout.
1. Put the pointer on the divider near its **bottom** — e.g. level with the "Created At" row of the metadata panel. The cursor turns into the resize cursor.
1. Press and drag sideways.
1. Repeat near the **top** of the same divider.

## Observed

The drag only engages in roughly the top 40% of the divider. Lower down nothing

happens on press-and-drag, even though the resize cursor is shown for the

divider's entire height, so the divider looks draggable everywhere.

Measured in the live app (viewport 1600, folder explorer split, divider spanning

y=124–544, 420px tall). Dragging 60px sideways from five heights:

| position on the divider | y | info panel width change |
|---|---|---|

|—|—|—|
|---|---|---|

| 10% | 166 | 60px |
|---|---|---|

| 30% | 250 | 60px |
|---|---|---|

| 50% | 334 | 0 |
|---|---|---|

| 70% | 418 | 0 |
|---|---|---|

| 90% (≈ the "Created At" row) | 502 | 0 |
|---|---|---|

## Expected

Press-and-drag engages anywhere the resize cursor is shown — i.e. over the whole

height of the divider, as the cursor advertises.

## Environment

- **Mode**: both (geometry, not colour)
- **Viewport**: 1600×1000
- **Build**: `FR-3590` @ `origin/main` 60d7799b0, `@astryxdesign/core` 0.4.3

## Impact

Workaround exists — drag from the upper part of the divider. But nothing tells

the user where that boundary is, so it reads as an intermittently broken drag.

## Evidence

- The separator element is 420px tall. Its internal grab zone

(`position: absolute`, `touch-action: none`, `z-index: 1` — the element that

actually receives the drag) measures y=-86 to y=334: the right height, shifted

up by 210px, exactly half of it. Its computed transform is

`matrix(1, 0, 0, 1, 0.5, -210)`.

- Setting that element's `transform` to `none` in devtools restores its box to

y=124–544, matching the separator.

## Notes

Suspicion, not a diagnosis: in `ResizeHandle.tsx` the static `hitAreaHorizontal`

style stretches the grab zone with `top: 0; bottom: 0`, while the dynamic

`hitAreaOffsetX(dir)` adds `translate(…, -50%)`. A block-axis `-50%` pairs with a

`top: 50%` anchor, which this element no longer has. `hitAreaOffsetY` has the

mirror-image shape on the inline axis, so a vertical handle is likely affected

the same way.

**Not fixed upstream**: `@astryxdesign/core` 0.4.4 (latest at the time of

writing) ships these two style blocks unchanged, so a version bump does not

close it.

JIRA Issue: FR-3591

agatha197:
Resolves #8888 (FR-3591)

> Stacked on #8887 (FR-3590). Review that one first; this branch is based on it.

On the folder explorer's split divider, press-and-drag only engaged in roughly
the top 40% of the handle. The resize cursor covers the divider's whole height,
so the dead lower part looks draggable and the drag reads as intermittently
broken. Reported while reviewing #8887; **pre-existing** — measured identical
with and without that PR's change.

## Mechanism

In `@astryxdesign/core` `Resizable/ResizeHandle.tsx`, the grab zone — the
absolutely-positioned child with `touch-action: none` that actually receives the
drag — gets its height from a static style:

```ts
hitAreaHorizontal: { width: …, top: 0, bottom: 0, cursor: 'col-resize' }
```

and its offset from a dynamic one:

```ts
hitAreaOffsetX: (dir) => ({
  insetInlineStart: 0,
  transform: { default: `translate(calc(${dir} * (…) - 6.5px), -50%)`, … },
}),
```

The block-axis `-50%` pairs with a `top: 50%` anchor centring a *fixed-height*
zone. This element is stretched by `top/bottom: 0` instead, so the `-50%` shifts
a full-height box up by half its own height. `cursor: col-resize` lives on the
separator root, which is not shifted — hence cursor and hit area disagree.

Measured live (viewport 1600, divider spanning y=124–544, 420px):

| | y range | computed transform |
|---|---|---|
| separator | 124 – 544 | `none` |
| grab zone | **-86 – 334** | `matrix(1, 0, 0, 1, 0.5, -210)` |

Setting the grab zone's `transform` to `none` in devtools restored it to
124–544, confirming the shift is the whole story.

`hitAreaOffsetY` has the mirror-image shape on the inline axis, so a vertical
handle is likely affected the same way — untested, we have no vertical handle.

## Fix

`ResizeHandle` already exposes the way out. `hitAreaBiasDir()` returns `null`
when the effective pill side is `center`, and that branch styles the grab zone
with `rtlStyles.centerInline('0px')` — `transform: translate(-50%, 0px)` —
instead of `hitAreaOffsetX(dir)`:

```js
hitBiasDir == null
  ? isHorizontal ? rtlStyles.centerInline('0px') : styles.hitAreaCenteredY
  : isHorizontal ? dynamicStyles.hitAreaOffsetX(hitBiasDir) : ...
```

So the stray block-axis `-50%` is never applied at all. The fix is one prop on
the existing call site:

```tsx
<ResizeHandle direction="horizontal" isReversed hasDivider pillPlacement="center" … />
```

No CSS file, no `!important`, no reach into Astryx's internal DOM — and
`centerInline` uses the direction-symmetric `left: 50%` + `translateX(-50%)`
pair, so it is also better-behaved under RTL than the biased path it replaces.

**The only visual delta** is the pill: it centers on the divider (x-center
767.5, exactly the separator's) instead of sitting 8px toward the metadata
panel. Grab zone and pill stay concentric.

> An earlier revision of this PR shipped a co-located
> `FolderExplorerModalV2.css` that cancelled the shift with `translate: 0 50%`.
> @ironAiken2 pushed back on editing CSS directly, and he was right — that file
> and its marker `className` are deleted.

## Measured before / after

Same divider, drag 40–60px sideways from five heights:

| position | y | before | after |
|---|---|---|---|
| 10% | 166 | ✅ moves | ✅ moves |
| 30% | 250 | ✅ moves | ✅ moves |
| 50% | 334 | ❌ **0px** | ✅ moves |
| 70% | 418 | ❌ **0px** | ✅ moves |
| 90% (≈ the "Created At" row) | 502 | ❌ **0px** | ✅ moves |

Grab zone box: `-86 – 334` → **`124 – 544`**, exactly matching the separator.

Not regressed by this change, re-measured after it: the pill stays at y=318–350
with `pointer-events: none`; #8887's containment still holds (0px row overflow,
0 explorer elements hit inside the info-panel band).

## Verification

- `pnpm --prefix ./react exec vitest run src/components/FolderExplorerModalV2.test.tsx` → 3 passed
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → unchanged: the
  same single pre-existing `var(--x)` at `BAIText.css:28` (the gate matching that
  file's own doc comment, present on `origin/main`). The new CSS declares no
  `var()`.

## Review follow-up (@ironAiken2)

> Didn't we agree that we'd use Stylex rather than editing the CSS files
> directly when changing the style?

Correct, and the CSS is gone — see **Fix** above. My first answer claimed
StyleX could not express this; that was wrong. I had checked that `xstyle`
reaches only the separator root and that StyleX has no descendant selectors,
and concluded "impossible" without checking `ResizeHandle`'s own
`pillPlacement` branch, which sidesteps the defect entirely.

### Measured

| | grab zone | transform | css translate |
|---|---|---|---|
| before | -86 – 334 | `matrix(1,0,0,1,0.5,-210)` | — |
| CSS override (earlier revision) | 124 – 544 | `matrix(1,0,0,1,0.5,-210)` | `0px 50%` |
| **`pillPlacement="center"`** | **124 – 544** | **`matrix(1,0,0,1,-8,0)`** | **none** |

Drag from five heights, all live (50/70/90% were dead before):

| position | 10% | 30% | 50% | 70% | 90% |
|---|---|---|---|---|---|
| info panel delta | +50px | +50px | +50px | +50px | +50px |

- `pnpm --prefix ./react exec vitest run` → **93 files, 1472 tests passed**
- Zero `pageerror`.

## Residue

- The underlying Astryx defect is still there — `hitAreaOffsetX` /
  `hitAreaOffsetY` keep the stray block-axis `-50%` against a stretched box.
  This PR routes around it via `pillPlacement`; FR-3591 records the mechanism
  for an upstream fix.
- `pillPlacement="center"` is a documented prop, so nothing here depends on
  Astryx's internal DOM or class names any more.
- Vertical handles are untested; `hitAreaOffsetY` has the same shape, and
  `pillPlacement="center"` would route around it there too (the `center` branch
  uses `hitAreaCenteredY`, which is anchored for a stretched box).


ironAiken2 (changes requested):
Didn’t we agree that we’d use Stylex rather than editing the CSS files directly when changing the style?

agatha197:
@ironAiken2 Agreed on the rule — StyleX first. In this one case it cannot express the fix, so I kept the `.css` and want to check that reasoning with you.

Re-read the installed `@astryxdesign/core` 0.4.4 source (`src/Resizable/ResizeHandle.tsx`):

- `xstyle` is merged onto the **separator root** only; it never reaches the children.
- The element that needs fixing is the internal grab-zone `<div>`. Unlike the pill it carries **no `themeProps(...)`**, so it has no `astryx-resize-handle-*` class of its own to target.
- `children` replaces only the pill — the grab zone is always rendered.
- StyleX has no descendant selectors; it styles the element you apply it to, and this element belongs to Astryx internals.

So there is no StyleX expression for "the grab zone inside someone else's component". CLAUDE.md names this case directly:

> A style that props/xstyle cannot express goes in a co-located `.css` file the component imports (P17), with `var(--…)` Astryx tokens; never a runtime style engine.

That is what the 3-line `FolderExplorerModalV2.css` is — scoped to this one handle, and written to be deleted when upstream drops the stray `-50%`.

The one genuinely StyleX-native alternative is `astryx swizzle ResizeHandle`: eject the component into our repo and fix it at source. That trades a deletable 3-line rule for a forked component we own and keep in sync, which felt like the worse deal for an upstream one-liner — but it is your call, and I'll switch if you prefer the fork.

Also re-verified after rebasing onto the updated #8887: grab zone now spans 124–544, exactly matching the separator, and dragging works at all five heights (50/70/90% were dead before). Zero `pageerror`.

agatha197:
Correction — @ironAiken2 was right, and my previous comment was wrong.

There **is** a props-only fix. I missed a branch in `ResizeHandle`:

```js
hitBiasDir == null
  ? isHorizontal ? rtlStyles.centerInline('0px') : styles.hitAreaCenteredY
  : isHorizontal ? dynamicStyles.hitAreaOffsetX(hitBiasDir) : ...
```

`hitAreaBiasDir()` returns `null` when the effective pill side is `center`, and that branch styles the grab zone with `centerInline('0px')` — `transform: translate(-50%, 0px)`. The stray block-axis `-50%` is simply never applied. So `pillPlacement="center"` avoids the defect at source; no CSS needed.

Pushed as `pillPlacement="center"` and **deleted `FolderExplorerModalV2.css`** along with its marker `className`.

Measured (viewport 1600, divider 124–544):

| | grab zone | transform | css translate |
|---|---|---|---|
| before | -86 – 334 | `matrix(1,0,0,1,0.5,-210)` | — |
| CSS override (previous commit) | 124 – 544 | `matrix(1,0,0,1,0.5,-210)` | `0px 50%` |
| **`pillPlacement="center"`** | **124 – 544** | **`matrix(1,0,0,1,-8,0)`** | **none** |

Drag engages at 10/30/50/70/90% of the handle height (50/70/90% were dead before).

The only visual delta: the pill centers on the divider (x-center 767.5, exactly the separator's) instead of sitting 8px toward the metadata panel. Grab zone and pill stay concentric, and `centerInline` uses the direction-symmetric `left: 50%` + `translateX(-50%)` pair, so this is also better under RTL than the biased path it replaces.

`verify.sh` ALL PASS; react 93 files / 1472 tests pass.

Sorry for the noise on the first pass — I checked `xstyle` and the descendant-selector limitation but not the `pillPlacement` branch, and asserted "impossible" on incomplete evidence.

ironAiken2 (approved):
LGTM

PR #8895, opened 2026-08-19
fix(FR-3594): restore compiler memoization on admin/project-admin list surfaces

### LINKED ISSUE #8893: Admin/project-admin pages: React Compiler bailouts flash the fetch-key button's loading state on row selection

ironAiken2:
Resolves #8893 (FR-3594)

Sibling of #8752 (FR-3510) — same symptom, three different React Compiler bailout causes. Stacked on that PR.

## Mechanism

FR-3510 established the chain: the React Compiler skips a component → `queryVariables` (a plain object literal) gets a new identity every render → any unrelated state update (row selection) makes `useDeferredValue` return the previous object for one urgent render pass → `deferredQueryVariables !== queryVariables` flashes `true` → `BAIFetchKeyButton`'s 700ms anti-flicker hold amplifies it into a visible, unclickable loading state.

A compiler sweep over all 48 pages plus the 93 risky components (files using `useDeferredValue` / `FetchKeyButton` / react-hooks suppressions) found the admin/project-admin surfaces where that chain is live, with three distinct bailout causes:

| File | Bailout cause | Fix |
|---|---|---|
| `AdminVFolderNodeListPage` (/admin/data) | render-phase `queryMapRef.current[…] = …` write → `CompileError: Cannot access refs during render` | move the write into a `useEffect` (the tab-change handler only reads the map on user interaction) |
| `ProjectAdminDataPage` | **two defects**: (1) identical render-phase ref write; (2) even with compiler memoization restored, `combinedFilter` was rebuilt through `filterOutEmpty([...])` / a switch returning fresh literals — the compiler keeps calls with fresh **mutable** arguments un-memoized, so the object filter (and `queryVariables`) changed identity every render. String-filter pages are immune because the memo guard compares strings by value. | (1) same; (2) hoist usage-mode filters to module constants and build `combinedFilter` from literals only, which the compiler memoizes |
| `AdminComputeSessionListPage` (/admin/session) | inline count-filter strings need `\"` escapes in the `graphql` tag; cooked ≠ raw tagged template is a compiler Todo bailout | pass count filters as query variables (the `filterForActiveCount` precedent) |
| `AdminUserCredentialList` | no `'use memo'` directive — compiler never ran | add the directive (compiles cleanly) |
| `ContainerRegistryList` | no `'use memo'` directive — compiler never ran | add the directive (compiles cleanly) |

Also replaced the `@ts-ignore` badge-count suppressions in the touched files with `keyof` narrowing, per the FR-3510 review direction of removing suppressions rather than adding them.

## Verification

- Per-file annotation-mode compile check (`babel-plugin-react-compiler`): all five files now report `CompileSuccess` (before: `CompileError`/never compiled)
- **Live probe** (headless Chromium against the dev server, 50ms sampling for 1.5s after clicking a row checkbox):

  | Probe | Before | After |
  |---|---|---|
  | `/project/default/admin/data` — refresh button loading after row select | **8/20 samples, 0–599ms window** | **0/22 samples** |
  | `/admin-data` — same | (fixed by commit 1) | **0/20 samples** |
  | Selection toolbar appears | works | works |
  | Manual refresh click still shows loading | — | yes, 0–592ms window (legit loading intact) |
  | `pageerror` count | 0 | 0 |
- `bash scripts/verify.sh` → `=== ALL PASS ===` (the earlier Relay drift was the regenerated `AdminComputeSessionListPageQuery` artifact, committed here)
- `pnpm --prefix ./react exec vitest run` → 88 files, **1405 passed** (0 failed)

## Residue (recorded in FR-3594 for follow-up triage)

The same sweep found bailouts **without** a deferred-comparison flash surface (no user-visible bug known, memoization only): `AdminDeploymentPresetSettingPage` (try/finally), `DeploymentAddRevisionModal`, `ChatCard`, and several others. And bailouts **with** a deferred surface but outside the admin scope of this issue: `AssignRoleModal`, `ProjectAdminSettingModal`, `UserSettingModal`, `FolderExplorerModal`/`V2`, `LoginSession`, plus missing-directive files `DashboardPage`, `ReservoirArtifactDetailPage`, `MyResourceWithinResourceGroup`, `TotalResourceWithinResourceGroup`, `PendingSessionNodeList`. Those need per-file restructuring (try/finally, optional-chaining-in-try, TSAsExpression keys) and are listed in the Jira issue.

### DIFF
--- a/react/src/__generated__/AdminComputeSessionListPageQuery.graphql.ts
+++ b/react/src/__generated__/AdminComputeSessionListPageQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<5cd4a94f9a2089a82f2810de0e978003>>
+ * @generated SignedSource<<1dac9fcf4fe49833725628c52c60a88f>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -12,6 +12,11 @@ import { ConcreteRequest } from 'relay-runtime';
 import { FragmentRefs, Result } from "relay-runtime";
 export type AdminComputeSessionListPageQuery$variables = {
   filter?: string | null | undefined;
+  filterForAllCount?: string | null | undefined;
+  filterForBatchCount?: string | null | undefined;
+  filterForInferenceCount?: string | null | undefined;
+  filterForInteractiveCount?: string | null | undefined;
+  filterForSystemCount?: string | null | undefined;
   first?: number | null | undefined;
   offset?: number | null | undefined;
   order?: string | null | undefined;
@@ -55,21 +60,46 @@ var v0 = {
   "name": "filter"
 },
 v1 = {
-  "defaultValue": 20,
+  "defaultValue": null,
   "kind": "LocalArgument",
-  "name": "first"
+  "name": "filterForAllCount"
 },
 v2 = {
-  "defaultValue": 0,
+  "defaultValue": null,
   "kind": "LocalArgument",
-  "name": "offset"
+  "name": "filterForBatchCount"
 },
 v3 = {
   "defaultValue": null,
   "kind": "LocalArgument",
+  "name": "filterForInferenceCount"
+},
+v4 = {
+  "defaultValue": null,
+  "kind": "LocalArgument",
+  "name": "filterForInteractiveCount"
+},
+v5 = {
+  "defaultValue": null,
+  "kind": "LocalArgument",
+  "name": "filterForSystemCount"
+},
+v6 = {
+  "defaultValue": 20,
+  "kind": "LocalArgument",
+  "name": "first"
+},
+v7 = {
+  "defaultValue": 0,
+  "kind": "LocalArgument",
+  "name": "offset"
+},
+v8 = {
+  "defaultValue": null,
+  "kind": "LocalArgument",
   "name": "order"
 },
-v4 = [
+v9 = [
   {
     "kind": "Variable",
     "name": "filter",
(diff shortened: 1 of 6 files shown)


ironAiken2:
Folded into #8752 — the sibling fixes now ship on the FR-3510 branch directly (single PR instead of a stack).

PR #8907, opened 2026-08-20
fix(FR-3599): apply the deployment card's secondary accent to its icon and title

### LINKED ISSUE #8904: Start page: Start Deployment card icon/title render in the primary (orange) accent instead of the secondary (green) button accent
## Where

- **Page**: `/start` (Start page)
- **Region**: the "Start Deployment" board card — icon badge and title text
- **Component**: `ActionItemContent` (`react/src/components/ActionItemContent.tsx`), used from `react/src/pages/StartPage.tsx`

## Observed

The Start Deployment card's action button correctly renders in the SECONDARY

(green) accent, but the card's icon and title render in the PRIMARY (orange)

accent — the same color as every other Start page card (Create Folder, Start

Session, Start Batch Session, Start from URL).

## Expected

The icon and title should render in the same SECONDARY (green) accent as the

button, so the whole card reads as one visually consistent unit — matching the

pre-migration (antd `ThemeSecondaryProvider`) behavior, where a single accent

override applied to the button, icon, and title together.

## Environment

- **Mode**: both (root cause is theme-token plumbing, not a mode-specific style)
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: `main` @ 2026-08-20

## Reproduction

1. Open the Start page (`/start`).
1. Locate the "Start Deployment" card.
1. Compare its icon/title color against its own button color.

## Evidence

- None attached.

## Notes

Root cause identified while triaging: `StartPage.tsx` wraps the deployment

card in `AstryxSecondaryTheme` (a nested Astryx `<Theme>`), which the Astryx

`Button` correctly picks up via the CSS cascade. `ActionItemContent`'s icon and

title, however, read their color from `theme.useToken()` — the antd-shim token

object mounted once near the app root — which is not per-subtree-aware, so

they keep the app-wide primary accent regardless of the nested theme. The

comment already in `StartPage.tsx` at the deployment card flagged this as a

known partial-fidelity gap left over from the Ant Design → Astryx migration

(FR-3482). Fixed in a follow-up PR: `ActionItemContent` now applies `themeColor`

consistently to icon + title (previously only the title honored it), and

`StartPage` passes `themeColor={token.colorSuccess`} explicitly for the

deployment card.

JIRA Issue: FR-3599

agatha197:
Resolves #8904 (FR-3599)

## Summary

- The Start page's "Start Deployment" card wraps itself in `AstryxSecondaryTheme`, which correctly gives its button the SECONDARY (green) accent. Its icon and title, however, kept the PRIMARY (orange) accent — the same defect on every other card — because `ActionItemContent` reads icon/title colour from `theme.useToken()`, the app-root `theme-shim`, which is not aware of nested Astryx `<Theme>` subtrees.
- `ActionItemContent`: consolidated the icon and title color logic into a single `accentColor`/`accentColorWithAlpha` derivation, so the icon now honours `themeColor` (it previously ignored it and always used `token.colorPrimary`).
- `StartPage`: passes `themeColor={token.colorSuccess}` to the deployment card's `ActionItemContent`, matching the pre-migration antd `ThemeSecondaryProvider` behavior where one accent override covered button, icon, and title together.
- Dropped the `itemRole` (`'user' | 'admin'`) prop from `ActionItemContent`. It switched icon/title colour between `colorPrimary`/`colorInfo`, but no call site has passed `itemRole="admin"` since FR-587 (Feb 2025) moved the deployment card onto an explicit secondary-theme wrapper instead of the role-based blue/orange split — it was dead code, confirmed by a full-repo grep. `themeColor` is the only live colour override now.

This was already flagged as a known partial-fidelity gap in a code comment on `StartPage.tsx` left over from the Ant Design → Astryx migration (FR-3482); this PR closes it.

## Test plan

- [x] `bash scripts/verify.sh` → `=== ALL PASS ===`
- [ ] Visual check: Start page → Start Deployment card icon/title render in green, matching the button

### DIFF
--- a/react/src/components/ActionItemContent.tsx
+++ b/react/src/components/ActionItemContent.tsx
@@ -19,7 +19,6 @@ interface StartItemContentProps {
   onClick?: () => void;
   to?: To;
   themeColor?: string;
-  itemRole?: 'user' | 'admin';
   type?: 'simple' | 'default';
   style?: React.CSSProperties;
 }
@@ -33,13 +32,11 @@ const ActionItemContent: React.FC<StartItemContentProps> = ({
   to,
   themeColor,
   type = 'default',
-  itemRole = 'user',
   style,
 }) => {
   const { token } = theme.useToken();
   const containerRef = useRef<HTMLDivElement>(null);
-  const accentColor =
-    themeColor ?? (itemRole === 'user' ? token.colorPrimary : token.colorInfo);
+  const accentColor = themeColor ?? token.colorPrimary;
   const accentColorWithAlpha = `rgba(${parseInt(accentColor.slice(1, 3), 16)}, ${parseInt(accentColor.slice(3, 5), 16)}, ${parseInt(accentColor.slice(5, 7), 16)}, 0.15)`;
 
   // PILOT-DECISION: antd `Button type="primary"` hand-painted with
--- a/react/src/components/ActionItemContent.tsx
+++ b/react/src/components/ActionItemContent.tsx
@@ -38,7 +38,9 @@ const ActionItemContent: React.FC<StartItemContentProps> = ({
 }) => {
   const { token } = theme.useToken();
   const containerRef = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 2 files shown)


nowgnuesLee (approved):
LGTM

Issue #8905, opened 2026-08-20
Container Logs modal: text and button colors differ when opened from a notification action button vs the session detail page


agatha197:
## Where

- **Page**: any page — opened via the notification panel's action button row (Header → Notifications, or a toast), compared against the session detail page's action button row
- **Region**: the Container Logs modal header (title text) and its action buttons (kernel select, download, auto-update)
- **Component**: `ContainerLogModal` (`react/src/components/ComputeSessionNodeItems/ContainerLogModal.tsx`), opened via `SessionActionButtons` (`react/src/components/ComputeSessionNodeItems/SessionActionButtons.tsx`). The notification-side trigger is `BAIComputeSessionNodeNotificationItem` (`react/src/components/BAIComputeSessionNodeNotificationItem.tsx`), rendered inside an Astryx `Banner` (`react/src/components/astryx-bui/BAINotificationStackAstryx.tsx`)

## Observed

When the Container Logs modal is opened from a notification's action button, its title text and buttons render in a different color than when the same modal is opened from the session detail page's action button — even though both entry points render the exact same `ContainerLogModal` component.

## Expected

The Container Logs modal should have identical text and button colors regardless of where it was opened from — matching the style it has when opened from the session detail page's action button.

## Reproduction

1. Open a session's Container Logs from the session detail page's action button row. Note the title/button colors.
1. Open the Container Logs modal for the same (or another) session via a notification's action button instead (Header → Notifications, or a toast).
1. Compare — the title text and button colors differ between the two.

## Evidence

- Container Logs modal (as seen when opened via the notification path): `/home/ubuntu/.claude/jobs/df106900/pasted-1.png`
- Notification card with the action buttons used to open it: `/home/ubuntu/.claude/jobs/df106900/pasted-2.png`

To be attached in Jira by the reporter.

## Notes

`SessionActionButtons.tsx` carries an existing comment: "inside an info Banner the theme collapses `--color-accent` onto the text colour" (FR-3506), and notification items render inside an Astryx `Banner`. This suggests the Banner's locally-scoped color-token overrides may be leaking into the modal content rather than being scoped to the Banner's own DOM subtree — same suspected mechanism as FR-3580, but on the modal's colors rather than an inline action button's fill. Unconfirmed; for triage to verify.

JIRA Issue: FR-3600

PR #8909, opened 2026-08-20
fix(FR-3601): give the Container Logs modal a height budget that fits

### LINKED ISSUE #8906: Container Logs modal: the log viewer shows a vertical scrollbar by default (body height too short)
## Where

- **Page**: any page — the Container Logs modal, regardless of where it is opened from
- **Region**: the log viewer body area inside the modal
- **Component**: `ContainerLogModal` (`react/src/components/ComputeSessionNodeItems/ContainerLogModal.tsx`) — the log container `<div>` wrapping `ScrollFollow`/`LazyLog`, sized via `height: 'calc(100% - 50px)'` inside a modal body fixed to `height: 'calc(100vh - 100px)'`

## Observed

The Container Logs modal's log viewer area shows a vertical scrollbar by default.

## Expected

The modal/log area height should be increased so that, by default, there is no vertical scrollbar in the log viewer.

## Reproduction

1. Open Container Logs for any session (from either the session detail page or a notification action button).
1. Observe the log viewer area — a y-scrollbar is present by default.

## Evidence

- Container Logs modal showing the default y-scroll: `/home/ubuntu/.claude/jobs/df106900/pasted-1.png`

To be attached in Jira by the reporter.

JIRA Issue: FR-3601

agatha197:
Resolves #8906 (FR-3601)

## Mechanism

Astryx `Dialog` caps its own height at its `maxHeight` prop, which **defaults to `75vh`** (`@astryxdesign/core/src/Dialog/Dialog.tsx`, `maxHeight = '75vh'`).

`ContainerLogModal` never passed a `maxHeight`, but asked for a body of `calc(100vh - 100px)`. At a 950px viewport that is an 850px body inside a dialog clamped to 712.5px — so the body overflowed the dialog's content area, and Astryx `LayoutContent` (`overflow-y: auto`) turned the excess into a permanent vertical scrollbar.

This is migration residue: antd's `Modal` had no such cap, so `calc(100vh - 100px)` used to size the body directly.

## Approach — follow the existing full-height-viewer idiom

The two sibling viewers already solved this. Both cap at **`95vh`**:

| Component | cap | content box |
|---|---|---|
| `VFolderTextFileEditorModal` (Monaco file viewer) | `maxHeight={'95vh'}` | `height: calc(100vh - 220px)` |
| `FolderExplorerModalV2` | `maxHeight={EXPLORER_MAX_HEIGHT}` = `'95vh'` | `minHeight: calc(95vh - var(--astryx-dialog-padding-block-start) - var(--astryx-dialog-padding-block-end))` |

This PR takes `95vh` from both, and the **token-based chrome subtraction** from `FolderExplorerModalV2` — the dialog's own block padding is published by the theme as `--astryx-dialog-padding-block-*`, so it is not re-measured as a literal here.

Two things it deliberately does *not* copy:

- **`FolderExplorerModalV2`'s `minHeight`-only fill.** Tried and measured: with no definite height on the body, `height: 100%` does not resolve through `LayoutContent`, the log pane collapsed to 120px and `react-lazylog` rendered blank.
- **`VFolderTextFileEditorModal`'s `calc(100vh - 220px)`.** Its 220px reserves space for a Save/Cancel footer that this modal does not have (`footer={null}`), so copying it wastes ~88px. It is also viewport-relative while the cap is proportional, and the two cross over above ~1120px of viewport height.

So the body is measured against **the cap** rather than the viewport. Only the header row stays a literal (52px, measured).

## Measured before / after

Live app, `hf-model-to-Qwen2-5-0-5B-Instruct-vPK` → Container Logs. `LayoutContent` is the box that was scrolling; `lazyLog` is the visible log pane:

| Viewport | | dialog height | LayoutContent scrollH / clientH | scrollbar | visible log |
|---|---|---|---|---|---|
| 1600×950 | before | 713px (75vh cap) | 882 / 661 → **221px overflow** | yes | ~629px |
| 1600×950 | after | 902px (95%) | 762 / 762 | **no** | **724px** |
| 1600×700 | after | 665px (95%) | — | **no** | 486px |
| 1600×1080 | after | 1026px (95%) | — | **no** | 847px |
| 1600×1440 | after | 1368px (95%) | — | **no** | 1189px |

The dialog lands at exactly 95% of the viewport at all four heights, so the fix scales instead of drifting.

Dark mode re-measured at 1600×950 (`document.documentElement.dataset.theme === 'dark'`): identical numbers, `scrolls: false`, visible log 724px. Zero `pageerror` in both passes.

### Before

The log pane runs past the modal's bottom edge and the modal body scrolls:

Container Logs modal before the fix

### After

The modal fits the viewport and the log pane ends inside it:

Container Logs modal after the fix

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup, StyleX sentinel, Astryx theme build, Terminology)
  - `Astryx theme build` failed on the first run in a fresh worktree with `No "exports" main defined in .../backend.ai-ui/package.json` — the known stale-BUI-dist trap, not this change. `pnpm --filter backend.ai-ui build`, then `ALL PASS`.
- `pnpm --prefix ./react exec vitest run` → 93 files, **1472 passed**
- `pnpm --filter backend.ai-ui exec vitest run` → 52 files, **677 passed, 1 skipped**
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 1 undeclared `var()`, in `packages/backend.ai-ui/src/components/BAIText.css:28`. **Pre-existing** — verified by running the gate on `main`, where it fails identically. The two `--astryx-dialog-padding-block-*` tokens this PR adds are declared and pass.

## Residue

- `LOG_MODAL_HEADER_HEIGHT` (52px) is still a literal — there is no published token for the `DialogHeader` row. If the title ever wraps to a second line (very narrow viewports; the modal is `width: 100%`, so this needs a phone-width screen) the body overshoots by one row and `LayoutContent` scrolls, which is the same graceful degradation as today.
- The `react-lazylog` pane keeps its own inner scrollbar. That is the log *content* being longer than the pane — inherent to a log viewer, and not what the report was about.
- **Does not** address FR-3600 (#8905), the sibling report that this modal's text/button colors differ when opened from a notification action button. Not reproduced: reaching the notification path needs a session-creation mutation on the shared cluster, and two cheap proxies (tinting the mount-point ancestor; opening the modal under the admin/blue `AstryxAdminTheme` route scope) both left the colors unchanged — so it is neither plain `color` inheritance nor the route theme scope. It looks specific to the Astryx `Banner` surface the notification renders on, which FR-3580 (#8870, In Review) already covers.

### DIFF
--- a/react/src/components/ComputeSessionNodeItems/ContainerLogModal.tsx
+++ b/react/src/components/ComputeSessionNodeItems/ContainerLogModal.tsx
@@ -26,11 +26,19 @@ interface ContainerLogModalProps extends BAIModalProps {
   defaultKernelId?: string;
 }
 
-// Astryx `Dialog` caps its height at 75vh unless given `maxHeight`, so the tall
-// body this viewer asks for overflowed into a scrollbar. The two budgets must
-// stay in step: the body plus the header chrome has to fit the cap (FR-3601).
-const MODAL_MAX_HEIGHT = 'calc(100vh - 64px)';
-const LOG_BODY_HEIGHT = 'calc(100vh - 160px)';
+/**
+ * Height cap for the log dialog — the `95vh` the sibling full-height viewers
+ * use (`VFolderTextFileEditorModal`, `FolderExplorerModalV2`). Astryx `Dialog`
+ * otherwise caps itself at 75vh, well under the body this viewer asks for.
+ *
+ * The body is measured against that same cap rather than the viewport, so the
+ * two cannot cross over on a tall screen. It subtracts only the chrome above
+ * it: the dialog's own block padding (a published theme token) and the header
+ * row (measured 52px; this modal renders `footer={null}`).
+ */
+const LOG_MODAL_MAX_HEIGHT = '95vh';
+const LOG_MODAL_HEADER_HEIGHT = '52px';
+const LOG_BODY_HEIGHT = `calc(${LOG_MODAL_MAX_HEIGHT} - ${LOG_MODAL_HEADER_HEIGHT} - var(--astryx-dialog-padding-block-start) - var(--astryx-dialog-padding-block-end))`;
 
 const ContainerLogModal: React.FC<ContainerLogModalProps> = ({
   sessionFrgmt,
@@ -149,7 +157,7 @@ const ContainerLogModal: React.FC<ContainerLogModalProps> = ({
         </BAIFlex>
       }
       width={'100%'}
-      maxHeight={MODAL_MAX_HEIGHT}
+      maxHeight={LOG_MODAL_MAX_HEIGHT}
       styles={{
         header: {
           width: '100%',
--- a/react/src/components/ComputeSessionNodeItems/ContainerLogModal.tsx
+++ b/react/src/components/ComputeSessionNodeItems/ContainerLogModal.tsx
@@ -26,6 +26,12 @@ interface ContainerLogModalProps extends BAIModalProps {
   defaultKernelId?: string;
 }
 
+// Astryx `Dialog` caps its height at 75vh unless given `maxHeight`, so the tall
+// body this viewer asks for overflowed into a scrollbar. The two budgets must
+// stay in step: the body plus the header chrome has to fit the cap (FR-3601).
+const MODAL_MAX_HEIGHT = 'calc(100vh - 64px)';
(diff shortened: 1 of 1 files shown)


ironAiken2 (approved):
LGTM

PR #8921, opened 2026-08-20
fix(FR-3606): adapt the folder explorer to live viewport resizes

### LINKED ISSUE #8920: Folder explorer: layout does not adapt to live viewport resizes (stuck or broken until reload)
## Where

The folder explorer modal (Data → folder name → explorer), `react/src/components/FolderExplorerModalV2.tsx`.

## Steps to reproduce

1. Open the folder explorer on a wide viewport (≥1200px) — the two-pane split layout renders.
1. Shrink the browser window below the `xl` breakpoint (1200px), or start narrow and grow past it.

## Observed

- The layout does not follow the viewport: it either keeps the previous layout verbatim (e.g. grows back to 1600px but stays in the single-column stacked layout, no split, no divider) or lands in a broken in-between state where panes/toolbar buttons overlap and button labels wrap mid-word. It stays that way until a full reload.
- Measured on the FR-3591 dev build: 1600 → 1000 → 1600 leaves the modal stacked at 1600px (screenshots attached to the PR).

## Root cause

`useBAIBreakpoint`'s store (`packages/backend.ai-ui/src/theme-shim/breakpoints.ts`) registers one `MediaQueryList` handler **per subscriber**, but all handlers share the module-level `cachedSnapshot`. On a breakpoint flip the first handler to run updates the cache and notifies its own component; every later handler then computes "no change" against the already-updated cache and never calls its `onStoreChange`. Net effect: exactly one component in the app reacts to a media-query flip; everything else keeps the stale snapshot until something else happens to re-render it. (The shrink direction sometimes works in the explorer only because the split-row `ResizeObserver` forces a re-render; that observer is unmounted in the stacked layout, so the grow direction never recovers.)

## Expected

1. Every `useBAIBreakpoint` subscriber re-renders when a breakpoint flips — the explorer switches split ↔ stacked live.
1. The explorer's action buttons (Create Folder / Create File / Upload, header FileBrowser / SFTP) adapt to the size as designed (labels ≥ `lg`, icon-only below) without a reload.
1. The file-list/metadata divider changes direction with the layout: vertical handle in the side-by-side layout, horizontal handle above the metadata panel in the stacked layout (FR-3516's proposal).
1. The explorer toolbar (breadcrumbs + action buttons) never overlaps or clips its labels at narrow widths.

JIRA Issue: FR-3606

agatha197:
Resolves #8920 (FR-3606)

The folder explorer froze in whatever breakpoint state it first rendered in: growing the window back past `xl` left it stacked (no split, no divider) until a full reload, and crossing breakpoints sometimes left a broken mixed state (overlapping toolbar buttons, mid-word label wraps).

## Mechanism

`useBAIBreakpoint`'s store (`packages/backend.ai-ui/src/theme-shim/breakpoints.ts`) registered one `MediaQueryList` handler **per subscriber**, all sharing the module-level `cachedSnapshot`. On a flip, the first handler to run updated the cache and notified its own component; every later handler then computed "no change" against the already-updated cache and never called its `onStoreChange`. Net effect: exactly one subscriber in the whole app reacted to any breakpoint flip. (Shrinking sometimes worked in the explorer only by accident — the split row's `ResizeObserver` forced a re-render; that observer is unmounted in the stacked layout, so growing back never recovered.)

The store now attaches **one shared handler** and fans out to a listener set, so every subscriber re-renders on a flip. The cache is refreshed on first subscribe in case it went stale while detached. The regression test renders three subscribers and asserts all of them flip — it fails against the old implementation (verified by stashing the fix: `1 failed | 1 passed`).

## Stacked-layout divider (FR-3516)

Below `xl` the two panes stack, and the split now flips to the block axis: a horizontal `ResizeHandle direction="vertical"` sits above the metadata panel and drags its height (default 440px, min 160px). `pillPlacement="center"` routes the grab zone away from `hitAreaOffsetY` — the inline-axis mirror of the FR-3591 offset bug — so the drag engages across the handle's full width (verified at x=100 and x=400 of a 672px handle).

The toolbar-overlap half of the report is already handled one PR down (FR-3590 added `wrap="wrap"` to the breadcrumb/actions row); with the store fixed, the `lg` icon-only button switch now also happens live.

## Evidence (measured on this branch's dev build)

1600 → 1000 → 1600, modal open the whole time:

| before: stuck stacked after growing back to 1600 | after: split restored at 1600 |
|---|---|
| before | after |

| after: live shrink to 1000 — stacked + horizontal divider | after: live shrink to 800 — icon-only buttons, no overlap |
|---|---|
| stacked | narrow |

- xl divider still drags after breakpoint round-trips, including low on the handle (FR-3591 intact): `aria-valuenow` 720 → 800.
- Vertical handle drag: 440 → 540 (up), 540 → 490 (down).

## Follow-up: header wrap at very narrow widths

Reported during review: at ~430px the FileBrowser/SFTP groups overflowed the header row and clipped into the modal's close button. The header row now wraps — the groups drop under the title, right-aligned (`margin-left: auto`), with a 180px readable title slice as the wrap trigger. Wide layout unchanged (title and actions measured on one row at 1600px).

| after: header wrapped at 430px |
|---|
|  |

## Verification

- BUI vitest: 53 files / 679 tests passed (incl. the new `breakpoints.test.tsx`)
- `react` vitest `FolderExplorerModalV2.test.tsx`: 3 passed

**Checklist:**

### DIFF
--- a/packages/backend.ai-ui/src/theme-shim/breakpoints.test.tsx
+++ b/packages/backend.ai-ui/src/theme-shim/breakpoints.test.tsx
@@ -46,13 +46,16 @@ beforeAll(() => {
       matches: evaluateQuery(query, viewportWidth),
       evaluate: () => evaluateQuery(query, viewportWidth),
       listeners,
-      addEventListener: (_type: string, cb: () => void) => listeners.add(cb),
-      removeEventListener: (_type: string, cb: () => void) =>
-        listeners.delete(cb),
+      addEventListener: (_type: string, cb: () => void) => {
+        listeners.add(cb);
+      },
+      removeEventListener: (_type: string, cb: () => void) => {
+        listeners.delete(cb);
+      },
     };
     mockLists.push(list);
     return list;
-  }) as typeof window.matchMedia;
+  }) as unknown as typeof window.matchMedia;
 });
 
 const Probe: React.FC<{ id: string }> = ({ id }) => {
--- a/packages/backend.ai-ui/src/theme-shim/breakpoints.test.tsx
+++ b/packages/backend.ai-ui/src/theme-shim/breakpoints.test.tsx
@@ -1,0 +1,92 @@
+import { useBAIBreakpoint } from './breakpoints';
+import { act, render, screen } from '@testing-library/react';
+import { beforeAll, describe, expect, it } from 'vitest';
+
+/**
+ * matchMedia mock with a mutable viewport width: `matches` follows the query,
+ * and crossing a boundary fires that list's 'change' listeners — the part of
+ * the real MediaQueryList the store subscribes to.
+ */
+type MockMQL = {
+  matches: boolean;
+  media: string;
+  addEventListener: (type: string, cb: () => void) => void;
+  removeEventListener: (type: string, cb: () => void) => void;
+};
+
+let viewportWidth = 1600;
+const mockLists: Array<
+  MockMQL & { listeners: Set<() => void>; evaluate: () => boolean }
+> = [];
+
+function evaluateQuery(query: string, width: number): boolean {
+  const min = query.match(/min-width:\s*([\d.]+)px/);
+  if (min) return width >= parseFloat(min[1]);
+  const max = query.match(/max-width:\s*([\d.]+)px/);
+  if (max) return width <= parseFloat(max[1]);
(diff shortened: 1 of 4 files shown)


ironAiken2 (approved):
LGTM

PR #8925, opened 2026-08-20
fix(FR-3607): pin the session-drawer launch notice visible and interactive above the drawer

### LINKED ISSUE #8922: Session detail drawer: launching an app (e.g. Terminal) shows no progress notification
## Component

- **Name**: app launch actions in `SessionActionButtons` (`react/src/components/ComputeSessionNodeItems/SessionActionButtons.tsx`), which launch via `useBackendAIAppLauncher` (`launchAppWithNotification` / `runTerminal`); the progress surface is the BAI notification stack (`react/src/hooks/useBAINotification.tsx`)
- **Astryx primitive**: Unknown (notification surface: `BAINotificationStack` / ToastViewport)
- **Scope**: session detail drawer only, as reported / Unknown elsewhere

## Where it is reachable

- Sessions page (`/session`) → click a session name → session detail drawer (`SessionDetailDrawer.tsx` → `SessionDetailContent.tsx` → `SessionActionButtons`)
- Also used in: `SessionDetailContent.tsx` (drawer and detail page); the same buttons appear on the session detail page route

## Steps to reproduce

1. Open the Sessions page and click a running interactive session's name to open the session detail drawer.
1. In the drawer, launch an app — e.g. Terminal (or App launcher → any app).
1. Watch the notification area for a launch progress notification.

## Observed

The app launch proceeds, but no progress notification is displayed while the app is being prepared.

## Expected

A progress notification appears in the notification stack tracking the app launch (the launch flow is literally `launchAppWithNotification`; app launches show launch progress feedback elsewhere in the app).

## Environment

- **Mode**: Unknown
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown

## Impact

Workaround exists — the launch itself appears to proceed; only the progress feedback is missing, so the user cannot tell whether anything is happening until the app opens.

## Evidence

- None provided.

## Notes

The drawer is open while launching; this may share a surface with FR-3578 / FR-3486 (notification surfaces hidden or inert behind an open dialog top-layer). Not verified — recorded as a suspicion only.

JIRA Issue: FR-3607

ironAiken2:
Resolves #8922 (FR-3607)

## Mechanism

Launching an app (e.g. Terminal) from the session detail drawer produced no visible
progress notification. The launch flow itself was never broken — `launchAppWithNotification`
upserts the notice into the stack unconditionally. The notice was *painted over*: the
scrimmed drawer opened with `showModal()`, which promotes the `<dialog>` into the CSS
top layer, and the top layer paints above every z-index in the document. The fixed
`.bai-notification-stack` anchors to the bottom-right corner — exactly the 800px-wide
drawer panel's footprint — so the notice was fully covered by the opaque panel (and
inert besides, per the [modal-dialog inertness spec](https://html.spec.whatwg.org/multipage/interaction.html#modal-dialogs-and-inert-subtrees) FR-3578 documented).

#8881 (FR-3585) removes the cause: scrimmed drawers render through `BAIDrawerPortal`
(`show()`, no top layer), and the z ladder pins `notification` (11000) above the whole
modal band (`modalBase` 1100). This PR stacks on it and pins the *reported scenario* —
launch notice over an open drawer — with a regression suite, so the drawer can never
silently return to the top layer without CI naming this exact symptom.

## What the suite pins (`notificationOverDrawer.test.tsx`)

1. An open scrimmed `BAIDrawerAstryx` calls `show()` and never `showModal()` — nothing
   is promoted above the notification stack, nothing outside the drawer is inerted.
2. A launch-progress notice (`backgroundTask` pending, determinate percent) upserted
   through the real `useBAINotification` wiring while the drawer is open renders in the
   stack, description and progress bar included.
3. The notice stays interactive: its dismiss button closes it, and the drawer stays open.

**Before-state proof**: with `BAIDrawerAstryx` checked out from `main` (the `showModal()`
version), assertion 1 fails — `showModal` is called — while the notice still "renders" in
jsdom, which is precisely why a DOM-presence test alone could not have caught this bug.

## Live probe (real cluster, this branch)

Sessions → running session → detail drawer → Terminal launch button:

| | Light | Dark |
|---|---|---|
| Notice over the open drawer | light | dark |

- Notice box `x:1192 y:768 w:384` vs drawer panel `x:800 y:0 w:800 h:900` — the notice
  sits fully inside the drawer's footprint and paints on top of it.
- `document.elementFromPoint()` at the notice's center resolves inside the notice in both
  modes — visible *and* hit-testable, not just painted.
- The Terminal popup opened in both passes; zero `pageerror` in both.

## Verification

- `react` vitest: 97 files / 1542 tests passed (includes the 3 new tests)
- `backend.ai-ui` vitest: 55 files / 711 passed, 1 skipped
- `astryx-token-gate --strict`: 1 pre-existing finding (`BAIText.css:28`), identical on
  the untouched base branch — not introduced here (verified by stashing the diff and re-running)

## Residue

- The behavioral fix lands in #8863/#8881; this PR adds no production code. If the base
  PRs change approach, the suite here is the contract their replacement must still satisfy.
- Keyboard reachability of notices while a modal/drawer holds the focus trap is FR-3578's
  documented design (pointer + dismiss reachable, Tab stays trapped) — unchanged here.


ironAiken2:
Closing: the behavioral fix lands in #8881 (FR-3585), which now carries its own regression coverage for this scenario. The suite added here is superseded by that PR's stack.

Issue #8884, opened 2026-08-19
Data > New Folder modal: the auto-mount help tooltip renders at the viewport top-left instead of anchored to its "?" icon


yomybaby:
## Component

- **Name**: `BAIQuestionIconWithTooltipAstryx` (`react/src/components/astryx-bui/BAIQuestionIconWithTooltipAstryx.tsx`)
- **Astryx primitive**: `Tooltip` (wrapping a `<button>` + `Icon` / `CircleHelpIcon`)
- **Scope**: observed inside a modal dialog (New Folder). The reporter's wording ("in modals like folder creation") suggests it is not limited to this one modal. Behaviour outside modals: Unknown — not probed at report time.

## Where it is reachable

- `/data` (menu: Storage → Data) — the "새 폴더 추가 / Add new folder" modal, the "?" icon next to the **자동 마운트 / Auto-mount** option of the **사용 방식 / Usage mode** radio group (`FolderCreateModalV2.tsx:555`, tooltip text = `data.AutomountFolderCreationDesc`)
- Also used in: 10 files (e.g. `SchedulerPage.tsx`, `VFolderNodesV2.tsx`, `MyKeypairManagementModal.tsx`, `InviteFolderSettingModal.tsx`, `ChatParametersSliders.tsx`) — usage breadth only, not reproduced at each site

## Steps to reproduce

1. Open Storage → Data.
1. Click the button that opens the "새 폴더 추가 / Add new folder" modal.
1. Hover (or focus) the "?" icon next to the "자동 마운트 / Auto-mount" option in the "사용 방식 / Usage mode" row.

## Observed

The tooltip bubble appears pinned to the **top-left corner of the viewport** (roughly 0,0, over the app header), completely detached from the "?" icon that triggered it. The tooltip text itself is correct ("자동 마운트 폴더는 새로운 세션이 시작될 때 자동으로 마운트됩니다."); only its position is wrong. There is no connecting arrow or any other visual relationship to the trigger.

## Expected

The tooltip is anchored to the "?" icon — positioned adjacent to it (above/below/beside per Astryx's placement logic), inside or over the modal, so the reader can tell which control the hint belongs to.

## Environment

- **Mode**: light (dark: Unknown)
- **Viewport**: Unknown (screenshot is a crop; the modal is right-of-centre and the tooltip sits at the far top-left, so the window is wide)
- **Browser/OS**: Unknown
- **Build**: Unknown

## Impact

[...]

PR #8929, opened 2026-08-20
fix(FR-3609): make the deployment card-title help tooltips hoverable

### LINKED ISSUE #8924: Deployment detail: the Replicas / AutoScaling / Access Tokens card-title help tooltips never appear on hover
## Component

- **Name**: Astryx `Tooltip` as used directly in three deployment detail cards —

`DeploymentReplicasCard.tsx`, `DeploymentAutoScalingCard.tsx`,

`DeploymentAccessTokensCard.tsx` (`react/src/components/`)

- **Astryx primitive**: `Tooltip` (`@astryxdesign/core/Tooltip`), with a bare

`lucide-react` `CircleHelp` `<svg>` as the trigger child

- **Scope**: this page only (as reported)

## Where it is reachable

- `/deployments/:deploymentId` (menu: Deployments → pick a deployment) — the

help "?" icon next to each of the three card titles:

- **Replicas** (복제본) — `DeploymentReplicasCard.tsx:129`
- **AutoScaling** (오토스케일링) — `DeploymentAutoScalingCard.tsx:98`
- **Access Tokens** (액세스 토큰) — `DeploymentAccessTokensCard.tsx:146`

- The page is registered in project, project-admin and admin scopes

(`react/src/routes.tsx:383`, `:561`, `:705`), so all three see it.

## Steps to reproduce

1. Open Deployments and click any deployment to reach its detail page.
1. Hover the "?" icon beside the **Replicas** card title.
1. Repeat for the **AutoScaling** and **Access Tokens** card titles.

## Observed

No tooltip appears. The icon is drawn, but hovering it produces nothing —

neither the bubble nor a `cursor: help` affordance. Keyboard focus does not

surface the text either, so the description is unreachable by any input.

## Expected

Hovering the "?" icon shows the description bubble, the same way the "?" icons

in the **Replica Lifecycle / Health / Traffic / Revision** column headers on

this very page do (`DeploymentReplicasCard.tsx:370`, `:416`, `:441`, `:493`) —

and the same way these three card titles behaved before the Astryx migration.

The strings exist and are translated (`deployment.tab.description.Replicas`,

`.AutoScaling`, `.AccessTokens` in `resources/i18n/en.json`).

## Environment

- **Mode**: both
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: `main` (post-FR-3482 Astryx migration)

## Impact

Workaround exists — the information is only explanatory, but there is no way to

read it from the UI at all.

## Evidence

- Screenshot: none supplied.
- Same-page comparison: the four column-header "?" icons in the Replicas table

work, so the working and non-working cases sit side by side on one screen.

## Notes

Suspicion only, to be confirmed at fix time: the working icons go through

`BAIQuestionIconWithTooltip`, which wraps the glyph in a `<span>`; the three

failing ones pass the bare `<svg>` to `Tooltip` directly. Related shapes exist

in other files, so this may be worth batching.

JIRA Issue: FR-3609

agatha197:
Resolves #8924 (FR-3609)

## Problem

On the deployment detail page the help "?" icons beside the **Replicas**,
**AutoScaling** and **Access Tokens** card titles never show their tooltip —
by hover or by keyboard.

All three passed a bare lucide `<svg>` straight to Astryx `Tooltip`:

```tsx
<Tooltip content={t('deployment.tab.description.Replicas')}>
  <CircleHelp style={{ color: token.colorTextDescription }} size="1em" />
</Tooltip>
```

Astryx wraps a non-text child in a `display: contents` div and attaches its
`mouseenter` / `mouseleave` listeners to `wrapper.firstElementChild` — here, the
`<svg>` itself. lucide renders `fill="none"`, and SVG hit-testing defaults to
`pointer-events: visiblePainted`, so only the glyph's ~1.5px stroke is
hoverable; the pointer never "enters" the element. `useTooltip`'s `isFocusable()`
accepts only `A/BUTTON/INPUT/SELECT/TEXTAREA`, `[tabindex]` or contenteditable,
so with the default `focusTrigger="auto"` there is no focus fallback either.

The wiring is otherwise correct — the elements do get `anchor-name` and
`aria-describedby` — which is why this reads as "the tooltip is missing" rather
than "the tooltip is broken".

Regression origin: 7d7b42388 (`feat(FR-3482): migrate the WebUI from Ant Design
to Astryx`) swapped antd's `QuestionCircleOutlined` (a real `<span role="img">`,
with handlers attached via `cloneElement`) for the lucide `<svg>`.

## Evidence

Measured against a running dev server, hovering each trigger's centre and
reading back the open `[popover]` elements:

| Trigger shape | Where | computed `cursor` | Tooltip on hover |
|---|---|---|---|
| bare `<svg>` in a `display:contents` wrapper | Dashboard board items (`BAIBoardItemTitle`) | `auto` | 0 / 6 |
| glyph wrapped in a real element (`BAIQuestionIconWithTooltip`) | Admin → Resource Policy column header | `help` | 1 / 1 |

The deployment detail page itself could not be exercised end to end — the shared
test backend currently has no deployments — so the reproduction above uses the
identical trigger shape on a reachable page. Note the same page already renders
four *working* "?" icons (Replica Lifecycle / Health / Traffic / Revision column
headers) through `BAIQuestionIconWithTooltip`, so the working and broken shapes
sat side by side on one screen.

## Change

Converge the three card titles onto `BAIQuestionIconWithTooltip`, the shared BUI
component (63 call sites) that wraps the glyph in a `Text` element with
`cursor: help` and `display: inline-flex`:

```tsx
<BAIQuestionIconWithTooltip title={t('deployment.tab.description.Replicas')} />
```

This also drops each file's `theme.useToken()` read, the `Tooltip` /
`CircleHelp` imports and the hand-rolled tint — the component takes its colour
from `Text color="placeholder"`, which is that shim's intended end state.

Net: 14 insertions, 30 deletions across 3 files.

## Scope

Deliberately limited to the three cards named in the issue. The same bare-`<svg>`
trigger shape is confirmed broken elsewhere (`BAIBoardItemTitle`,
`BAIAlertIconWithTooltip`, `ClusterModeFormItems`, `ResourceAllocationFormItems`,
`SessionDetailContent`, `ResourcePresetSelect`, `UserInfoModal`); that sweep is
filed separately rather than folded in here.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript,
Vite warmup paths, StyleX cssInjectionTarget, Astryx theme build, Terminology).

### DIFF
--- a/react/src/components/DeploymentAccessTokensCard.tsx
+++ b/react/src/components/DeploymentAccessTokensCard.tsx
@@ -8,13 +8,11 @@ import { DeploymentAccessTokensCardListQuery } from '../__generated__/Deployment
 import { DeploymentAccessTokensCard_deployment$key } from '../__generated__/DeploymentAccessTokensCard_deployment.graphql';
 import { App } from '../app-shim';
 import { Form } from '../form-engine';
-import { theme } from '../theme-shim';
 import BAIFormItem from './BAIFormItem';
 import { AstryxFormSelector } from './astryxFormControls';
 import { DateTimeInput } from '@astryxdesign/core/DateTimeInput';
 import type { ISODateTimeString } from '@astryxdesign/core/DateTimeInput';
 import { Text } from '@astryxdesign/core/Text';
-import { Tooltip } from '@astryxdesign/core/Tooltip';
 import {
   BAISkeleton,
   BAIButton,
@@ -24,6 +22,7 @@ import {
   BAIFlex,
   BAIModal,
   BAINameActionCell,
+  BAIQuestionIconWithTooltip,
   BAITable,
   BAIText,
   BAIUnmountAfterClose,
@@ -36,7 +35,7 @@ import {
   useMutationWithPromise,
 } from 'backend.ai-ui';
 import dayjs from 'dayjs';
-import { Trash2, CircleHelp, PlusIcon } from 'lucide-react';
+import { Trash2, PlusIcon } from 'lucide-react';
 import React, {
   Suspense,
   useDeferredValue,
@@ -77,7 +76,6 @@ const DeploymentAccessTokensCard: React.FC<DeploymentAccessTokensCardProps> = ({
 }) => {
   'use memo';
   const { t } = useTranslation();
-  const { token } = theme.useToken();
   const { message } = App.useApp();
   const { logger } = useBAILogger();
   const [isPendingRefetch, startRefetchTransition] = useTransition();
@@ -147,12 +145,9 @@ const DeploymentAccessTokensCard: React.FC<DeploymentAccessTokensCardProps> = ({
         title={
           <BAIFlex gap="xs" align="center">
             {t('deployment.tab.AccessTokens')}
-            <Tooltip content={t('deployment.tab.description.AccessTokens')}>
-              <CircleHelp
-                style={{ color: token.colorTextDescription }}
-                size="1em"
-              />
-            </Tooltip>
+            <BAIQuestionIconWithTooltip
+              title={t('deployment.tab.description.AccessTokens')}
+            />
           </BAIFlex>
         }
         extra={
--- a/react/src/components/DeploymentAutoScalingCard.tsx
+++ b/react/src/components/DeploymentAutoScalingCard.tsx
@@ -13,10 +13,8 @@ import { App } from '../app-shim';
 import { useCurrentUserInfo } from '../hooks/backendai';
 import { useBAIPaginationOptionStateOnSearchParam } from '../hooks/reactPaginationQueryOptions';
 import { useBAISettingUserState } from '../hooks/useBAISetting';
-import { theme } from '../theme-shim';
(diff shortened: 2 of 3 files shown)


agatha197:
Reviewer note — unrelated defect visible while testing this PR.

On **Chromium >= 151** (e.g. Arc), every Astryx tooltip on this page renders at the viewport top-left instead of beside its trigger. That is **not** caused by this PR: measured on the deployment detail page with this branch applied, 13/13 tooltips land at `(0, 4)` on Chromium 151 and 13/13 anchor correctly on Chromium 145 — including the four Replicas column-header tooltips this PR does not touch.

Tracked in FR-3589 (#8884), which I have re-measured and commented on. On Chromium 145 the three tooltips this PR fixes anchor correctly (trigger y=843 -> bubble y=787), so the change here is verifiable on a browser without that regression.

nowgnuesLee (approved):
LGTM

PR #8949, opened 2026-08-21
test: e2e auto-heal — login-flow locators after the Astryx migration

lablup-octodog:
Produced by the daily digest heal pipeline (manual invocation, 2026-08-21) — unblocks the e2e login flow that gates every spec and the digest preflight, broken since the Astryx migration.

## Root causes (both from #8626 / FR-3482)

1. **`getByLabel('Endpoint')` became ambiguous.** With `connectionMode: 'SESSION'` and no saved endpoint, the endpoint section is expanded on load. The old locator substring-matches 5 elements once open (incl. the "Endpoint History" / "About Endpoint" buttons); the strict-mode violation was swallowed by `.catch(() => false)`, so the helper clicked **Advanced** believing the section was collapsed — collapsing the open section and hanging on `fill()`. A human never hits this: they see the field and type. → `getByRole('textbox', { name: 'Endpoint' })` in the shared helper and every local `fillEndpoint` copy.
2. **Astryx `Button` exposes no `aria-label`** (accessible name = visible text), so `getByLabel('Login', { exact: true })` matched **zero** elements. → `getByRole('button', { name: 'Login', exact: true })`, 16 call sites.

Yesterday's "trusted-click app bug" hypothesis is retracted — instrumented runs (listener walk, MutationObserver, presence polling from t=0) showed the clicks worked fine; they were collapsing an already-open section.

## Merged with `main` (2026-08-24)

`main` had since landed a partial fix of its own in `e2e/utils/test-util.ts` — it switched the submit click to `getByRole` and added `exact: true` to the `getByLabel('Endpoint')` lookup — which conflicted with this branch. Resolved in favour of this branch, which is the superset: `main`'s retry path still probes `getByLabel('Login', { exact: true })` for `stillOnLoginForm`, and that matches **zero** elements on an Astryx `Button`, so the flag is always `false` and a rejected login waits 60 s for a dropdown that will never appear instead of re-submitting.

## Verification — re-measured after the merge (2026-08-24)

Backend `10.82.0.130:8090`; WebUI served from this branch (post-merge tree) via `pnpm run dev`. Every number below is a real run, `--workers=1`, on the exact tree that is pushed here. Baseline = the same specs at `main`'s revision, same server, same session.

Direct locator measurement on the live login page:

| Locator | Matches |
|---|---|
| `getByLabel('Login', { exact: true })` (what `main` still uses) | **0** |
| `getByRole('button', { name: 'Login', exact: true })` (this PR) | **1** |

Spec runs:

| Scope | `main` baseline | This PR |
|---|---|---|
| `e2e/auth/login.spec.ts` | 6 failed / 3 passed | ✅ **9 passed / 0 failed** |
| `password-expiry` + `login-error-messages` + `concurrent-login-guard` | 15 failed / 2 passed / 5 not-run | **2 failed / 15 passed / 5 not-run** |
| `e2e/user/user-crud.spec.ts` | — | blocked at `:100`, see below |

`bash scripts/verify.sh` → `=== ALL PASS ===`.

**Known remaining failures — separate drift family, untouched by this PR:**

- `concurrent-login-guard:151` and `login-error-messages:331` both fail on `getByRole('dialog', { name: 'Logged in elsewhere' })` (Astryx modal role/name drift). That locator is a context line in this diff, not a changed one. The `concurrent-login-guard` suite is `describe.serial`, so its 4 remaining tests are skipped as a consequence.
- `user-crud:100` fails on `getByRole('tab', { name: 'Users' })` (Astryx tab drift). The login itself now succeeds and the run reaches the credential page — the failure is downstream. Because that suite is serial too, the one line this PR changes in `user-crud.spec.ts` (`:385`) is never reached in a full-suite run.
- The global cleanup teardown still reaches for `locator('.ant-modal-confirm')` in `test-util.ts:666` — a dead antd locator from the same migration backlog. It degrades gracefully (skip-and-continue) but leaves active vfolders behind.

These go to the nightly heal pipeline; none are caused by this PR (before it, they failed earlier, at the login click itself).


yomybaby (commented):
LGTM

yomybaby (approved):

PR #8952, opened 2026-08-21
fix(FR-3589): stop all:unset from poisoning tooltip anchor-name on Chromium 151

### LINKED ISSUE #8884: Data > New Folder modal: the auto-mount help tooltip renders at the viewport top-left instead of anchored to its "?" icon

agatha197:
Resolves #8884 (FR-3589)

Stacked on #8929 (`fix/FR-3609-deployment-card-tooltips`) — that PR makes the deployment card-title tooltips open at all; this one makes every tooltip using the unstyled-button trigger pattern anchor correctly on Chromium ≥ 151.

## Problem

On Chromium ≥ 151 (Arc, current Chrome), Astryx tooltips render at the viewport top-left, detached from their "?" trigger. Measured on the deployment detail page: 0/13 tooltips anchored on Chromium 151.0.7922.34, 13/13 on 145.0.7632.6 — same build, same probe.

## Mechanism (isolated and confirmed)

The tooltip trigger is an unstyled `<button style={{ all: 'unset', … }}>`. Astryx's `addAnchorName` composes `anchor-name` as a list: read `el.style.anchorName` → append `--astryx-layer-*` → write back.

Chromium changed the CSSOM getter for longhands covered by an inline `all: unset` between 145 and 151:

| | `style.anchorName` pre-read | composed write | computed |
|---|---|---|---|
| 145.0.7632.6 | `""` | `--astryx-layer-x` | `--astryx-layer-x` ✅ |
| 151.0.7922.34 | `"unset"` | `unset, --astryx-layer-x` — CSS-wide keyword in a list → **whole declaration invalid, write rejected** | `none` → top-left ❌ |

With `anchor-name: none` the popover's `position-area` never resolves and the layer stays at its static `top: 0 / left: 0`.

## Change

Extract the trigger mechanics into a generic BUI **`BAIIconWithTooltip`** — any glyph (`icon`, required) in a focusable unstyled button, carrying the explicit UA-style reset (`appearance / background / border / margin / padding / font / color`) instead of `all: 'unset'`, plus the FR-3589 warning comment. It has **no legacy call sites**, so its props extend Astryx's own `TooltipProps` directly (`content`, `isOpen`, `delay` in ms, native `placement`/`alignment`) rather than carrying an antd-shaped surface.

**`BAIQuestionIconWithTooltip`** becomes its CircleHelp specialization — it keeps its frozen antd-shaped surface (`title`, `open`, `mouseEnterDelay` in seconds, `placement: AntdPlacement`) for the 63 existing call sites unchanged, and translates it to `BAIIconWithTooltip`'s native props at the boundary (the same `splitAntdPlacement` it always used). No `icon` prop leaks onto it, so a "question icon" component can no longer render a key glyph or a badge.

The hand-rolled `Tooltip` + `<button>` wrappers converge onto them:

- `react/src/pages/SchedulerPage.tsx` — `BAIQuestionIconWithTooltip` (a plain question icon; also drops its `Tooltip` / `CircleHelp` / `theme.useToken` imports)
- `react/src/components/VFolderNodesV2.tsx` — `BAIIconWithTooltip icon={<StorageUsageBadge …/>}`
- `react/src/components/MyKeypairManagementModal.tsx` — `BAIIconWithTooltip icon={<KeyRoundIcon …/>}`, `style={{ cursor: 'default' }}`

No copy of the reset remains outside `BAIIconWithTooltip`.

## Does Astryx 0.4.7 (now on main, FR-3651) fix this upstream?

No — verified live after rebasing onto main's 0.4.7 bump, with an A/B on the running app (Chromium 151, deployment detail page):

| Astryx 0.4.7 + trigger style | anchored | top-left |
|---|---|---|
| `all: 'unset'` (pre-fix state, temporarily restored) | 0/13 | **13/13** |
| explicit reset (this PR) | **13/13** | 0/13 |

The installed 0.4.7's `Layer/anchorName.js` is byte-identical to 0.4.4 for `readAnchorNames`/`addAnchorName` — still `el.style.anchorName ?? ''` with no CSS-wide-keyword filtering. The 0.4.5–0.4.7 changelogs (touch-trigger support, ARIA rename, overlay padding) never touch anchor-name composition. The upstream defect is now tracked separately as **FR-3656** (defensive CSS-wide-keyword filtering in `readAnchorNames`; the filter was prototyped and verified — 13/13 anchored on both engines with it applied). This PR also adds a **lint guard**: a `no-restricted-syntax` rule (react + BUI configs) rejecting any inline `all:` property in a JSX `style` attribute, with this mechanism in the message, so the pattern cannot silently return while the upstream fix is pending. stylex `all: unset` (class-based) stays allowed — classes never reach the inline-style CSSOM getter.

Side effect worth noting: `all: unset` also suppressed the focus outline; the explicit reset leaves the default focus indicator, restoring the visible keyboard-focus state the button wrapper exists to provide.

## Not done here

Astryx's `readAnchorNames` should defensively filter CSS-wide keywords (`unset`, `initial`, `inherit`, `revert`, …) before composing the list — any future `all: unset` on a Tooltip trigger reintroduces this. That is an upstream `@astryxdesign/core` change (0.4.5 diff shows `Layer/` and `Tooltip/` untouched, so no released fix exists); worth filing upstream, not patching here.

## Verification

- Chromium 151: 13/13 deployment-detail tooltips anchored (was 0/13).
- Chromium 145: 13/13 still anchored (no regression).
- `bash scripts/verify.sh` → `=== ALL PASS ===`.

### DIFF
--- a/packages/backend.ai-ui/eslint.config.js
+++ b/packages/backend.ai-ui/eslint.config.js
@@ -28,6 +28,20 @@ export default [
     },
   },
 
+  {
+    files: ["src/**/*.tsx"],
+    rules: {
+      "no-restricted-syntax": [
+        "error",
+        {
+          selector: "JSXAttribute[name.name='style'] Property[key.name='all']",
+          message:
+            "Inline `all:` resets break Astryx anchor positioning on Chromium >= 151 — the anchor-name CSSOM getter returns the CSS-wide keyword, which poisons addAnchorName's list and detaches the layer to the viewport top-left (FR-3589). Use an explicit reset, or BAIIconWithTooltip for tooltip triggers.",
+        },
+      ],
+    },
+  },
+
   // Enforce that BUI components access translations through the internal
   // `useBAIi18n` hook only. Direct imports of i18n primitives from
   // `react-i18next` would re-introduce React-Context-based lookup, which
--- a/packages/backend.ai-ui/eslint.config.js
+++ b/packages/backend.ai-ui/eslint.config.js
@@ -28,20 +28,6 @@ export default [
     },
   },
 
-  {
-    files: ["src/**/*.tsx"],
-    rules: {
-      "no-restricted-syntax": [
-        "error",
-        {
-          selector: "JSXAttribute[name.name='style'] Property[key.name='all']",
-          message:
-            "Inline `all:` resets break Astryx anchor positioning on Chromium >= 151 — the anchor-name CSSOM getter returns the CSS-wide keyword, which poisons addAnchorName's list and detaches the layer to the viewport top-left (FR-3589). Use an explicit reset, or BAIIconWithTooltip for tooltip triggers.",
-        },
-      ],
-    },
-  },
-
   // Enforce that BUI components access translations through the internal
   // `useBAIi18n` hook only. Direct imports of i18n primitives from
   // `react-i18next` would re-introduce React-Context-based lookup, which
--- a/react/eslint.config.js
+++ b/react/eslint.config.js
@@ -89,6 +89,11 @@ export default [
           message:
             'Direct <style> elements are forbidden (CSP nonce safety). Import a co-located .css file instead, and drive runtime-variable values through CSS custom properties.',
         },
+        {
+          selector: "JSXAttribute[name.name='style'] Property[key.name='all']",
+          message:
+            "Inline `all:` resets break Astryx anchor positioning on Chromium >= 151 — the anchor-name CSSOM getter returns the CSS-wide keyword, which poisons addAnchorName's list and detaches the layer to the viewport top-left (FR-3589). Use an explicit reset, or BAIIconWithTooltip for tooltip triggers.",
+        },
       ],
     },
   },
--- a/react/eslint.config.js
+++ b/react/eslint.config.js
@@ -89,11 +89,6 @@ export default [
           message:
             'Direct <style> elements are forbidden (CSP nonce safety). Import a co-located .css file instead, and drive runtime-variable values through CSS custom properties.',
         },
-        {
-          selector: "JSXAttribute[name.name='style'] Property[key.name='all']",
-          message:
-            "Inline `all:` resets break Astryx anchor positioning on Chromium >= 151 — the anchor-name CSSOM getter returns the CSS-wide keyword, which poisons addAnchorName's list and detaches the layer to the viewport top-left (FR-3589). Use an explicit reset, or BAIIconWithTooltip for tooltip triggers.",
-        },
       ],
     },
   },
--- a/packages/backend.ai-ui/src/components/BAIIconWithTooltip.tsx
+++ b/packages/backend.ai-ui/src/components/BAIIconWithTooltip.tsx
@@ -1,0 +1,64 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+
+ Generic icon-with-tooltip trigger: any glyph (a lucide icon, a status badge)
+ wrapped in a focusable unstyled button so the hint is keyboard-reachable.
+ `BAIQuestionIconWithTooltip` specializes this with the CircleHelp glyph.
+
+ No legacy call sites to keep compiling, so the props extend Astryx's own
+ `TooltipProps` directly rather than an antd-shaped surface (unlike
+ `BAIQuestionIconWithTooltip`, a FRONTIER COMPONENT).
+*/
+import { nodeToAccessibleLabel } from '../helper/astryxLabel';
+import { Text } from '@astryxdesign/core/Text';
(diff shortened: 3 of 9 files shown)


agatha197:
Considered fixing this at the source instead — a pnpm patch on `@astryxdesign/core`'s `Layer/anchorName.ts` (`readAnchorNames`) so it filters CSS-wide keywords, protecting every `useLayer` consumer (Popover, DropdownMenu, HoverCard, Selector, ...) rather than just Tooltip. Prototyped and confirmed it works (isolated repro: with the filter, an `all: unset` element's anchor-name write succeeds on Chromium 151).

Dropped it after checking exposure: `all: 'unset'` (or any other CSS-wide keyword) appears in exactly one place project-wide — the tooltip-trigger pattern this PR touches. Zero of the 49 Popover/DropdownMenu/HoverCard/Selector call sites use it. So the wider protection a dependency patch buys is theoretical here, not observed, and doesn't justify the ongoing cost of carrying a patch across every future `@astryxdesign/core` bump. Staying with the app-side fix, scoped to where the bug actually shows up.

nowgnuesLee (approved):
LGTM

packages/backend.ai-ui/eslint.config.js
nowgnuesLee, line 1: Please add it to `packages/eslint-config-bai`

nowgnuesLee (changes requested):

packages/backend.ai-ui/eslint.config.js
agatha197: Simplified per offline feedback: the rule now sits inline (warn) in `eslint-config-bai`'s shared react preset — no exported constant — so `backend.ai-ui` needs no local block. The react workspace keeps its inline copy in the CSP `no-restricted-syntax` block, which redeclares the rule (flat config replaces rule configs rather than merging them). Head: b9bf83f8d.

react/eslint.config.js
nowgnuesLee, line 1: Please delete this changes.

nowgnuesLee (changes requested):

react/eslint.config.js
agatha197: Done!

nowgnuesLee (approved):
LGTM

PR #8968, opened 2026-08-21
fix(FR-3629): restore per-file isolation for the vitest suites

### LINKED ISSUE #8967: Re-enable per-file isolation for the react vitest suite (revert isolate: false)
The react vitest suite runs with `isolate: false` (`react/vitest.config.ts`), set by FR-3614. Since then `react-vitest` fails intermittently on CI across many unrelated branches, with a different test file failing on each run.

## Evidence — isolation is the deciding variable

Controlled experiment, same sequential file order (`--no-file-parallelism`), only isolation toggled:

| Run | Result | Test phase |
|---|---|---|

|—|—|—|
|---|---|---|

| `--isolate` | 96 files / 1559 tests all pass | 13.9s |
|---|---|---|

| current config (`isolate: false`) | 1 failed (`VFolderNodes.test.tsx`) | 16.4s |
|---|---|---|

The failing file passes on its own, so it is an inter-file effect, not a defect in the test.

## Why it looks flaky

Without isolation a worker reuses one jsdom and module environment across every file it runs, so files scheduled later in a worker run slower. Every observed failure is a `waitFor` / `userEvent` budget miss — "mutation was never dispatched", "called 0 times", "focus is null" — not a wrong assertion.

This explains the observed pattern: CI runners have fewer cores, so more files per worker and deeper accumulation; a dev box with more cores passes. Forcing a single worker locally reproduces it. Which file fails changes run to run because worker placement changes.

## What was ruled out

Listener accumulation on the shared `document` is NOT the mechanism. Removing the `node_modules` exclusion in `react/src/setupTests.ts` so that all listeners are cleaned made things worse — 1 failure became 8, and the test phase went 16.4s to 46.5s — because react-dom caches "already registered per document" and will not re-register what is removed. That exclusion is load-bearing.

FR-3617 raised `asyncUtilTimeout` to 5s and `testTimeout` to 15s, which delays the symptom but does not remove it: `DeploymentSettingModal` — named in that fix — still failed afterwards with the same `expected [] to have a length of 1`.

## Proposal

Restore per-file isolation for the react suite. The cost is wall-clock time: on one dev box a sequential run went 33s to 102s, which is the speed FR-3614 bought. If that cost is unacceptable, an alternative is to cap files per worker so accumulation stays shallow, but isolation is the only option verified to make the suite green.

Related: FR-3614 (introduced `isolate: false`), FR-3617 (timeout mitigation).

—

Captured while working on branch: FR-3585

JIRA Issue: FR-3629

nowgnuesLee:
Resolves #8967 (FR-3629)

## What & why

Reverts the `isolate: false` setup introduced by FR-3614 (#8938). Since that change, `react-vitest` failed intermittently on CI across unrelated branches — a different test file each run, always a `waitFor` / `userEvent` budget miss. FR-3629 documents the controlled experiment: with the same sequential file order, `--isolate` is green (96 files / 1559 tests) while the no-isolate config fails, so isolation is the deciding variable. Without isolation a worker reuses one jsdom + module environment across all its files, and later files accumulate enough slowdown to miss their async budgets on low-core CI runners.

## Changes

- `react/vitest.config.ts`, `packages/backend.ai-ui/vitest.config.ts` — drop `isolate: false` (back to vitest's default per-file isolation).
- `react/src/setupTests.ts`, `packages/backend.ai-ui/setupTests.ts` — remove the no-isolate compensation machinery (`vi.resetModules()`, DOM/storage scrubbing, the project-code listener tracker, explicit RTL `cleanup()` registration). All of it existed only to simulate what real isolation now provides.

## Deliberately kept from #8938 / FR-3617

- FR-3617's `asyncUtilTimeout: 5000` + `testTimeout: 15_000` — separate issue; still valid headroom for slow CI runners.
- The `global-stores.test.ts` listener-handle fixes and the `NetworkStatusBanner.test.tsx` probe assertion — those fixed genuinely broken test code (`removeEventListener` called with a fresh arrow function never unregisters) and are correct under isolation too.

## Cost

Isolation is the slower mode — that speed is what FR-3614 bought. On this run the react suite took 29s wall (16.98s test phase) parallelized, vs. the issue's measured 33s→102s sequential comparison. If the CI wall-clock cost turns out to matter, capping files per worker is the fallback discussed in FR-3629 — but isolation is the only option verified to keep the suite green.

## Verification

- `react` vitest: **94 files / 1548 tests passed**
- `backend.ai-ui` vitest: **692 passed | 1 skipped**
- `bash scripts/verify.sh`: `=== ALL PASS ===`

https://claude.ai/code/session_01M1ZB6YUQs7pU6ZDxrt4PQo


yomybaby (approved):
LGTM

Issue #8832, opened 2026-08-17
Spec: session list resource-grid view — production implementation (from map #8786)


yomybaby:
## Parent

Production follow-up of wayfinder map #8786. The agreed direction is the resolution of decision ticket #8790; the working prototype lives on branch [`prototype/session-resource-grid`](https://github.com/lablup/backend.ai-webui/tree/prototype/session-resource-grid) (`ddc3872e6`), iterated over 18 live reaction rounds (#8789).

## Problem Statement

A session list rendered as a table answers "which sessions exist" but not "what do they hold and how hard are they working". An operator who wants to survey resource pressure — how much CPU/memory/accelerator each session occupies, which sessions are hot and which are idle, where PENDING work is queued — must open sessions one by one or mentally aggregate table columns. There is no at-a-glance occupancy/utilization overview, and pagination hides most of the picture.

## Solution

Add a **grid view** to the session list pages (user page and `/admin/session`), toggleable against the existing table view. Every session in the current filter renders as a **group of small squares on one shared grid**: 1 CPU core = 1 square, memory quantized per selectable unit (1/2/4/8 GiB), 1 accelerator device = 1 square with fractional shares as partial fills. Square color encodes live utilization with the app's stepped semantic convention; group identity is a muted palette with per-session override. Hovering a group shows an anchored mini session detail popover; clicking opens the session detail drawer. The whole filtered set loads at once (cap 100 with a truncation banner) — an overview, not a page.

The grid renderer ships as a **standalone reusable `backend.ai-ui` component** (`BAIResourceUnitGrid`) that knows nothing about sessions: it takes groups of quantized units with resolved colors plus popover/legend slots. Session semantics (GraphQL query, slot parsing, kernel-level live_stat aggregation, mode/metric controls, drawer opening) stay in the app. Other surfaces (agents, deployments) can compose the same component later.

This is a **port + refactor of the prototype**, not a rebuild — the 18-round tuning (serpentine packing, merged group plates, popover anchoring) is preserved.

## User Stories

1. As a session-list user, I want a view toggle (Table | Grid) next to the refresh button, so that I can switch between detail-per-row and overview-per-filter without leaving the page.
2. As a session-list user, I want each of my sessions drawn as a connected group of unit squares sized by its allocation, so that I can compare session sizes at a glance.
3. As a session-list user, I want 1 CPU core to equal exactly 1 square in CPU mode, so that the grid reads as countable units rather than abstract areas.
[...]

PR #9006, opened 2026-08-24
feat(FR-3652): show the resource grid view in the agent detail panel

### LINKED ISSUE #9005: Show the session resource grid view in the agent detail panel
## Problem

The session resource grid view (FR-3569..FR-3571, spec https://github.com/lablup/backend.ai-webui/issues/8832) ships on the user session list and `/admin/session`, behind the `experimental_session_resource_grid` opt-in. Rolling it out to the other session lists was explicitly left out of scope by that spec.

The agent detail panel gained its own session list in FR-3252 ("sessions assigned to the agent"). That list is the one place where "what is this agent actually holding, and how hard is it working" is the whole question — exactly what the grid answers — but it has no grid view.

## Solution

Add the same Table | Grid toggle to the agent detail drawer's Sessions tab, gated behind the same experimental setting and using the same icons, tooltips and i18n keys as the two shipped pages.

The drawer's table reads `agentsV2 -> sessions` (Strawberry `SessionV2`), while the grid reads the legacy `compute_session_list`. The grid keeps its own query and is scoped to the agent through the legacy queryfilter's `agent_ids` field, which the manager maps to an array column (`ComputeSession._queryfilter_fieldspec` -> `ArrayFieldItem("sessions_agent_ids")`, compiled to `EXISTS(SELECT item FROM unnest(col) WHERE item = :val)`).

A `SessionV2`-native grid is not possible today: `live_stat` exists on no Strawberry type, so every cell would render in the "no data" grey that the grid uses to mean "no telemetry" — a false statement drawn as a feature.

## Notes

- The drawer instance runs the grid in controlled mode (`viewParams` / `onChangeViewParams`), so it writes none of the five app-global grid URL query keys that an uncontrolled grid owns. Without this, opening the drawer over a session list page already in grid view would silently repaint the page behind it.
- Prerequisite fix in `BAIResourceUnitGrid`: the hover popover is `position: fixed` and positioned from viewport coordinates, but the drawer's `<dialog>` carries `transform: translateX(0)` while open, which makes it the containing block for fixed descendants. Inside the drawer the popover therefore landed off-screen and was clipped away by the dialog's `overflow: hidden` — the grid's entire detail affordance (and its palette picker) silently did nothing on hover.

## Accepted limitations

- A multi-node session with kernels on other agents matches the filter and is drawn whole; the session-level `occupied_slots` has no per-kernel breakdown, so narrowing only kernel mode would make the two modes disagree.
- The grid caps at 100 sessions (`offset: 0`, no pagination) while the table paginates at 10, and the two views' status vocabularies differ because they read different graphs.
- The toggle is component-local state, so it resets when the drawer closes.

JIRA Issue: FR-3652

yomybaby:
Resolves #9005 (FR-3652)

## What

The session resource grid (FR-3569..FR-3571) ships on the user session list and `/admin/session`; its spec (#8832) explicitly left rolling it out to the other session lists out of scope. This adds it to the session list the agent detail panel gained in FR-3252 (#8613), behind the **same** `experimental_session_resource_grid` opt-in, with the same icons, tooltips and i18n keys.

No new i18n keys, no `graphql` tag touched, no `__generated__` churn.

| | |
|---|---|
| Flag off | tab renders exactly as it does today, no toggle |
| Flag on | Table \| Grid segmented control between the running/finished radio group and the refresh button |

## Since the `CHANGES_REQUESTED` review (@agatha197 — "I still can't see the tooltip")

The branch was stale and no longer built against `main`, which is the most likely reason the popover behaved as it did before the fix on the reviewer's box.

- **Rebased onto `main`** (was 2 weeks behind; `AgentSessions.tsx` conflicted with the `session-preemption-statuses` status buckets that landed in the meantime).
- **Fixed a build break:** FR-3655 (#9015) landed the day after this branch was last pushed and renamed `BAIDrawerAstryx` → `BAIDrawer`. The drawer story still imported the old name, so `packages/backend.ai-ui` no longer type-checked against `main` — and `backend.ai-ui` is consumed from `dist`, so a checkout in that state serves the **pre-fix** grid to anything that is not the react dev server (which aliases BUI to source).
- **Re-measured the popover on the real components in the real drawer** after the rebase (headless Chromium, 1920px viewport, panel at viewport left 1120), hovering the same plate — the numbers below reproduce exactly, and the flipped-above placement was measured too (grid pushed 300px / 600px down the panel): popover stays inside the panel, hit-testable, in all three.

If it still does not show after a fresh `pnpm install && pnpm --filter backend.ai-ui build`, please say which view you were on and roughly where on the screen — that would mean a second failure mode I have not found.

### 1. `fix(FR-3652)` — `BAIResourceUnitGrid` popover anchoring

A prerequisite, and a real defect in the shared component rather than in the new call site.

The grid's hover popover is `position: fixed` and writes coordinates measured with `getBoundingClientRect()` — viewport space. That is only correct while no ancestor establishes a containing block for fixed descendants, which every host so far satisfied. `@astryxdesign/lab`'s `Drawer` applies `transform: translateX(0)` to the `<dialog>` element itself while open, so the dialog takes over as the containing block — and its `overflow: hidden` clips whatever survives.

Measured on the **real components in a real drawer** (the new Storybook story, panel at viewport left 1120, 1920px viewport), hovering the same plate:

| | popover `left`..`right` | inside the panel | in viewport | hit-testable |
|---|---|---|---|---|
| before | **2268..2468** | ✗ | ✗ | ✗ |
| after | **1147..1347** | ✓ | ✓ | ✓ |

<table>
<tr><th>before — hover produces nothing</th><th>after</th></tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

So inside the drawer the grid's entire detail affordance — session detail *and* the palette picker — silently did nothing on hover.

The fix measures the containing block's origin with a zero-size fixed probe (the idiom this file already uses to resolve colors inside its own cascade) and subtracts it. The probe reports `(0, 0)` whenever the viewport really *is* the origin, so the two shipped pages and the dashboard panel are unchanged — confirmed by the same measurement on an untransformed host, where before and after are byte-identical.

Positioning in place rather than portalling to `document.body` keeps the popover inside the component's own cascade (theme tokens, `light-dark()`) and leaves the z-index ladder untouched.

### 3. `test(FR-3652)` — the Storybook story the measurement above runs against

Note for whoever opens that story: Storybook does not load `@astryxdesign/lab`'s stylesheet, so the class the dialog already carries for `side="end"` has no rule there and the panel sits at the *start* edge — where the origin is 0 and this bug cannot show. The measurement restores that one shipped rule (`.xhi6v0a { inset-inline-start: auto }`, present in `lab.css`) in the page first.

### 4. `fix(FR-3652)` — complete the minilang escaping

CodeQL's `js/incomplete-sanitization`. The escape handled `"` but not `\`, and lark's `ESCAPED_STRING` refuses to close a token after a lone trailing backslash — so such an id would not mis-scope the query, it would make the whole expression unparseable and degrade the grid to its error banner. Escaping the backslash first turns that into a value that parses and matches nothing. Practically unreachable either way (agent ids are operator-configured host-style labels), but the failure mode is now the harmless one.

### 5. `fix(FR-3652)` — point the drawer story at the renamed `BAIDrawer`

The rebase repair described in the review-response section. The same commit corrects the story's own comment, which implied the bug is visible in Storybook when it is not.

## Why the two views in one tab read different graphs

The drawer's table reads `agentsV2 -> sessions` (Strawberry `SessionV2`). The grid keeps its own legacy `compute_session_list` query, scoped through that queryfilter's `agent_ids` field — the manager maps it to an array column (`ComputeSession._queryfilter_fieldspec` → `ArrayFieldItem("sessions_agent_ids")`, compiled to `EXISTS(SELECT item FROM unnest(col) WHERE item = :val)`), present since 23.03.9, so no manager version gate is needed.

Rebuilding the grid on `SessionV2` was considered and is **not possible today**: `live_stat` exists on no Strawberry type. Every cell would render in the "no data" grey that the grid uses to mean *no telemetry* — a false statement drawn as a feature. This is also why the two views' status vocabularies and page sizes differ.

## URL query-state

An uncontrolled grid keeps its five settings (`gridMode`, `gridResource`, `gridMetric`, `gridMemUnit`, `gridLayout`) in **app-global** nuqs keys. A drawer opened over a session list page already in grid view would share them and silently repaint the page behind it.

The drawer instance therefore runs the grid **controlled** (`viewParams` / `onChangeViewParams` from local state) — the seam `SessionResourceGridPanel` already uses for the same reason — so it writes zero URL keys. The toggle is local state too: `AgentList` is mounted on three pages whose URLs would outlive the drawer.

## Polling

Exactly one of the two queries polls at a time. The refresh button reloads the preloaded table query only in table view, and switching back to table forces a `network-only` reload rather than showing rows that stopped refreshing.

## Accepted limitations

- **Multi-node bleed.** A session with kernels on other agents matches the filter and is drawn whole. Session-level `occupied_slots` has no per-kernel breakdown, so narrowing only kernel mode would make the two modes disagree with each other. Say the word if this is unacceptable on an agent panel — it is ~5 lines client-side, using the `containers.agent` field the query already selects.
- **Cap vs. pagination.** Grid = first 100, newest first, with the existing truncation banner; table = page N of 10. Both shipped pages already have this mismatch.
- **Status vocabulary.** The grid sends the complement form (`status != "TERMINATED" & status != "CANCELLED"`), matching both shipped pages, which also admits legacy-only statuses the table's 10-enum bucket does not. Counts can legitimately differ for the same radio setting.
- **Ordering** is fixed to `-created_at`; the table's sorter speaks camelCase `SessionV2OrderBy` and the legacy list speaks snake_case. The grid has no sortable headers, so this only decides *which* 100 come back.
- The toggle and grid settings reset when the drawer closes.

## Verification

Re-run after the rebase onto `main`:

- `backend.ai-ui` 1967 passed / 1 skipped (76 files); `react` 2270 passed (130 files)
- Unit test drives the popover fix through mocked rects and **fails on the pre-fix code** (`1144` instead of `24`)
- Real-component measurement in headless Chromium against the `InsideDrawer` story, panel forced to the end edge: **pre-fix `2268..2468`** (off-screen, `elementFromPoint` misses it) vs. **post-fix `1147..1347`** (inside the panel, hit-testable) — plus the flipped-above placement at 300px / 600px down the panel, all inside the panel
- `git status` clean outside the five changed files — no `resources/i18n`, no `__generated__`

## Still needs a live look (superadmin + real cluster)

Not reachable from this box:

- Grid toolbar wrapping inside the fixed 800px drawer on a cluster with ≥2 accelerator families
- Grid group count vs. the table's `count` for the same radio setting (expect a legitimate difference per the status note above)
- How the multi-node bleed actually reads on a real multi-node session
- `agent_ids` has no index (`69c059996cbd_add_agent_ids_col_to_session.py` adds the column without `create_index`), so both the count and the slice are sequential scans — fine for a drawer, worth measuring on a large deployment

## Review notes

- Turn on **Settings → Experiments → session resource grid**, open an agent's detail drawer, and switch to the **Sessions** tab — the Table | Grid control appears between the running/finished radio group and the refresh button. With the flag off, the tab is byte-identical to today.
- Hover a plate in grid view: the popover must open **inside the 800px panel**, and stay hoverable long enough to reach it. That is the FR-3652 popover fix; it is the one thing worth re-checking by hand.
- Open the drawer from a session list page that is already in grid view — the page behind it must keep its own `gridMode` / `gridResource` / … URL keys untouched (the drawer runs the grid controlled).

- [ ] Documentation
- [x] Specific setting for review (eg., KB link, endpoint or how to setup) — `experimental_session_resource_grid` opt-in, see Review notes
- [x] Minimum requirements to check during review — see Review notes
- [x] Test case(s) to demonstrate the difference of before/after — `BAIResourceUnitGrid.test.tsx` popover-rebasing test + the `InsideDrawer` story


agatha197 (changes requested):
Image I still can't see the tooltip.

yomybaby:
@agatha197 — thanks, and sorry for the delay. I could not reproduce "no tooltip" from the code, so here is what I found and what changed.

**Most likely cause: the branch no longer built against `main`.** FR-3655 (#9015) landed on 2026-08-24, the day after this branch was last pushed, and renamed `BAIDrawerAstryx` → `BAIDrawer`. The drawer story in this PR still imported the old name, so `packages/backend.ai-ui` stopped type-checking once `main` was in the picture. That matters here because `backend.ai-ui` is consumed from `dist` — only the react dev server aliases it to source — so a checkout in that state serves the **pre-fix** `BAIResourceUnitGrid`, which is exactly the "hover does nothing" behaviour.

**What I did**

- Rebased onto `main` (resolved the `AgentSessions.tsx` conflict with the `session-preemption-statuses` status buckets).
- Fixed the stale import, and corrected the story's comment — Storybook does not load lab's stylesheet, so `side="end"` has no rule there and the panel sits at the *start* edge, where the fixed-positioning origin is 0 and this bug cannot show at all.
- Re-measured on the real components in the real drawer, headless Chromium, 1920px viewport, panel forced to the end edge (`dialog { inset-inline-start: auto }`), hovering the same plate:

| | popover `left`..`right` | inside the panel | hit-testable (`elementFromPoint`) |
|---|---|---|---|
| pre-fix | 2268..2468 | no | no |
| post-fix | 1147..1347 | yes | yes |

  The flipped-above placement was measured too (grid pushed 300px and 600px down the panel): the popover stays inside the panel and hit-testable in all three cases.

- `bash scripts/verify.sh` → `=== ALL PASS ===`; `backend.ai-ui` 1967 passed / 1 skipped, `react` 2270 passed.

**Please re-check on the rebased branch** after `pnpm install && pnpm --filter backend.ai-ui build`. If the tooltip still does not show, could you say which view you were on (grid in the agent drawer, or one of the session list pages) and roughly where on the screen the plate was? That would mean a second failure mode I have not found, and the two numbers above would let me chase it.

agatha197:
@copilot fix conflicts

agatha197:
Dev server on box `sujin`: stopped (was http://fr-3652-pr9006-show-resource-grid.sujin.10-82-0-159.sslip.io)

agatha197 (approved):
LGTM

PR #9016, opened 2026-08-24
chore(FR-3657): use cached eslint and incremental tsc in verify.sh

### LINKED ISSUE #9014: Speed up verify.sh with ESLint cache and incremental tsc
## Problem

`scripts/verify.sh` (the local/agent verification harness) takes ~146s per run. Agents run it before every submit/push, so this cost is paid repeatedly.

Measured breakdown (8-core dev box):

| Step | Time | Share |
|---|---|---|

|—|—|—|
|---|---|---|

| Lint (`pnpm -r lint`, uncached) | 100.3s | 69% |
|---|---|---|

| TypeScript (`tsc --noEmit`, full check) | 36.6s | 25% |
|---|---|---|

| Relay + Format + theme build + misc gates | ~9s | 6% |
|---|---|---|

## Changes

1. **Lint: use the existing cached `lint:ci` scripts** (`--cache --cache-strategy content`) instead of uncached `lint`. Warm-cache measurements: react 71.9s → 3.6s, backend.ai-ui 30.3s → 3.9s. CI workflows (vitest-react.yml, vitest-backend-ai-ui.yml) and the pre-commit lint-staged hook already rely on the same cache strategy. Add a `lint:ci` script to backend.ai-client (currently missing).
1. **TypeScript: add `--incremental`** to the `tsc --noEmit` step (TS 5.5 supports it with noEmit). Measured: 36.8s → 8.8s warm. Add `*.tsbuildinfo` to .gitignore.

Cache safety: all active ESLint rules are single-file rules (no type-aware linting, no cross-file rules), and `--cache-strategy content` re-lints every modified file by content hash. tsc incremental tracks the type dependency graph itself.

## Expected result

verify.sh warm runs drop from ~146s to ~30s (5x). Parallelizing independent checks (a further ~15s) is out of scope here.

JIRA Issue: FR-3657

ironAiken2:
Resolves #9014 (FR-3657)

## What

`scripts/verify.sh` (the local/agent verification harness) took **~146s** per run; agents run it before every submit/push. Two changes cut warm runs to **~25–31s** (5x), plus three review-driven guards:

1. **Lint** — run the existing cached `lint:ci` scripts (`--cache --cache-strategy content`) instead of uncached `lint`. `backend.ai-client` and `backend.ai-agent-cli` gain `lint:ci` scripts so `pnpm -r lint:ci` keeps the same package coverage as `pnpm -r lint`.
2. **TypeScript** — add `--incremental` to the `tsc --noEmit` step (`react/tsconfig.tsbuildinfo`, gitignored via a new `*.tsbuildinfo` entry).
3. **Coverage gate** (review follow-up) — `scripts/lint-ci-coverage-gate.mjs`, run by verify.sh before Lint, fails when any workspace package defines `lint` without `lint:ci`. `pnpm -r` silently skips (exit 0) packages lacking the script, and `backend.ai-agent-cli` had already landed on main in exactly that state — without the gate, every future package would reopen the same hole.
4. **Schema-inline regression test** (review follow-up) — `packages/backend.ai-ui/src/locale/__tests__/i18nSchemaCacheTripwire.test.ts` pins the cache-tripwire below.
5. **Tripwire trigger paths** (review follow-up) — `.github/workflows/vitest-backend-ai-ui.yml` gains `packages/backend.ai-ui/eslint.config.js` and `i18n.schema.json` in its `paths:`. The suite's filter listed neither and verify.sh runs no vitest step, so a PR touching only the ESLint config — exactly the cleanup-to-a-path-reference item 4 guards against — would never have run that test. The schema entry also restores locale revalidation in CI for a schema-only PR.

## Why the caches are sound

ESLint's content-hash cache re-lints every modified file and invalidates on config/version changes; the one unsound case is a rule whose verdict on file A depends on the content of file B. Review found the places this repo hits that case; all are closed in this PR:

- **`backend.ai-client` and `backend.ai-agent-cli` are deliberately uncached** (their `lint:ci` repeats their plain `lint` command): both packages enable type-aware `@typescript-eslint/no-floating-promises` (`projectService`), so changing an imported function's return type to a Promise can make an unchanged caller invalid — which a content-hash cache would skip. CI already runs `backend.ai-client`'s lint uncached (`typecheck.yml`); the two cost ~5s each locally.
- **`backend.ai-ui`'s locale schema content is embedded in the ESLint config**: `src/locale/*.json` are validated against `i18n.schema.json` (`eslint-plugin-json-schema-validator`), and a schema-only edit does not touch the locale files' content. The schema's *content* is inlined into the `json-schema-validator/no-invalid` rule options, so editing the schema changes the locale files' config hash and invalidates exactly their cache entries. Verified empirically: with a warm cache, breaking the schema makes the next `lint:ci` fail with 11,156 violations; `.ts` cache entries stay valid because their config slice is unchanged.
  - Review caveat (agatha197): the rule resolves `$schema || options || catalog`, and every locale file carries `"$schema"` — so the inline copy is never used for *validation*; it works purely as a cache-hash tripwire. The new regression test asserts the rule options carry the schema **content** (an object deep-equal to the on-disk file, not a path string), so a later "cleanup" back to a path reference fails the test instead of silently restoring the stale-cache bug.

The remaining cached lint surfaces (react, backend.ai-ui) use single-file rules only, and this matches what the project already trusts: CI (`vitest-react.yml`, `vitest-backend-ai-ui.yml`) runs `lint:ci` with an actions cache keyed on the lockfile + eslint configs, and the pre-commit lint-staged hook runs `lint --cache`.

tsc `--incremental` tracks the type dependency graph itself, so a change in file A still surfaces type errors in dependent file B.

## Measurements (8-core dev box)

| Step | Before | After (warm) |
|---|---|---|
| react eslint | 71.9s | 3.6s |
| backend.ai-ui eslint | 30.3s | 3.9s |
| backend.ai-client eslint | ~5s | ~5s (deliberately uncached) |
| backend.ai-agent-cli eslint | ~5s (was covered by `pnpm -r lint`) | ~5s (deliberately uncached) |
| lint:ci coverage gate | — | <0.1s |
| tsc --noEmit | 36.8s | 8.8s |
| **verify.sh total** | **~146s** | **~30–41s** |

Cold runs (fresh caches) are unchanged (~142s). The warm total grew from the original ~25–31s only by the agent-cli lint/tsc lanes — coverage main already pays for, not new cost (the 41s upper bound was measured while a dev server shared the box). Parallelizing the independent checks is deliberately out of scope.

## Verification

- `bash scripts/verify.sh` — `=== ALL PASS ===` in **41s** warm (dev server sharing the box) after the rebase onto main and the review fixes.
- Coverage gate: removing `backend.ai-agent-cli`'s `lint:ci` makes the gate fail naming that package; restoring it returns green.
- Tripwire test: swapping the inline schema for a path string fails with `expected './i18n.schema.json' to be type of 'object'`; restoring the inline passes.
- Trigger paths: `packages/backend.ai-ui/eslint.config.js` and `i18n.schema.json` now appear in the `Vitest (backend.ai-ui)` `paths:` list, so a config-only or schema-only PR runs the suite.

https://claude.ai/code/session_014rDBVANzHZz7yupv4KL7X8
https://claude.ai/code/session_01SdS1y7NFyftDPZBWc2Cd59

### DIFF
--- a/packages/backend.ai-client/package.json
+++ b/packages/backend.ai-client/package.json
@@ -47,7 +47,7 @@
     "vitest": "vitest run",
     "vitest:watch": "vitest",
     "lint": "eslint ./src",
-    "lint:ci": "eslint ./src --cache --cache-strategy content --cache-location .eslintcache",
+    "lint:ci": "eslint ./src",
     "prepublishOnly": "pnpm lint && pnpm test && pnpm build"
   },
   "dependencies": {
--- a/packages/backend.ai-client/package.json
+++ b/packages/backend.ai-client/package.json
@@ -47,7 +47,7 @@
     "vitest": "vitest run",
     "vitest:watch": "vitest",
     "lint": "eslint ./src",
-    "lint:ci": "eslint ./src",
+    "lint:ci": "npm run lint --silent",
     "prepublishOnly": "pnpm lint && pnpm test && pnpm build"
   },
   "dependencies": {
--- a/packages/backend.ai-client/package.json
+++ b/packages/backend.ai-client/package.json
@@ -47,6 +47,7 @@
     "vitest": "vitest run",
     "vitest:watch": "vitest",
     "lint": "eslint ./src",
+    "lint:ci": "eslint ./src --cache --cache-strategy content --cache-location .eslintcache",
     "prepublishOnly": "pnpm lint && pnpm test && pnpm build"
   },
   "dependencies": {
--- a/packages/backend.ai-client/package.json
+++ b/packages/backend.ai-client/package.json
@@ -47,7 +47,7 @@
     "vitest": "vitest run",
     "vitest:watch": "vitest",
     "lint": "eslint ./src",
-    "lint:ci": "npm run lint --silent",
+    "lint:ci": "eslint ./src",
     "prepublishOnly": "pnpm lint && pnpm test && pnpm build"
   },
   "dependencies": {
--- a/packages/backend.ai-ui/eslint.config.js
+++ b/packages/backend.ai-ui/eslint.config.js
@@ -3,6 +3,14 @@ import jsonSchemaValidator from "eslint-plugin-json-schema-validator";
 import storybookPlugin from "eslint-plugin-storybook";
 import jsoncParser from "jsonc-eslint-parser";
 import globals from "globals";
+import fs from "node:fs";
+
+// Embedded inline (not referenced by path) so the schema CONTENT is part of
+// the resolved config: editing i18n.schema.json then invalidates `--cache`d
+// lint results for the locale files, which a `$schema` path alone would not.
+const i18nSchema = JSON.parse(
+  fs.readFileSync(new URL("./i18n.schema.json", import.meta.url), "utf8"),
+);
 
 export default [
   ...base,
@@ -109,6 +117,14 @@ export default [
     },
     rules: {
       ...jsonSchemaValidator.configs.recommended.rules,
+      "json-schema-validator/no-invalid": [
+        "error",
+        {
+          schemas: [
+            { fileMatch: ["**/src/locale/*.json"], schema: i18nSchema },
+          ],
+        },
+      ],
     },
   },
 
--- a/packages/backend.ai-ui/src/locale/__tests__/i18nSchemaCacheTripwire.test.ts
+++ b/packages/backend.ai-ui/src/locale/__tests__/i18nSchemaCacheTripwire.test.ts
@@ -1,0 +1,60 @@
+import { execFileSync } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+// Guards the lint-cache tripwire: eslint.config.js must inline i18n.schema.json's
+// CONTENT (not a path) so schema edits invalidate the locale files' eslint cache.
+describe('i18n schema lint-cache tripwire', () => {
+  it(
+    'inlines the schema content into the no-invalid rule options',
+    { timeout: 30_000 },
+    () => {
+      // jsdom gives this file a non-file import.meta.url; testPath is the real one.
+      const pkgRoot = path.resolve(
+        path.dirname(expect.getState().testPath ?? ''),
+        '../../..',
+      );
+      // Loaded in a subprocess: vitest's transform breaks the config's own
+      // `new URL(..., import.meta.url)`; native node import matches ESLint.
+      const configUrl = pathToFileURL(
+        path.join(pkgRoot, 'eslint.config.js'),
+      ).href;
+      const stdout = execFileSync(
+        process.execPath,
+        [
+          '--input-type=module',
+          '-e',
+          `const config = (await import(${JSON.stringify(configUrl)})).default;
+           const entry = config.find(
+             (e) =>
+               Array.isArray(e.files) &&
+               e.files.includes("src/locale/*.json") &&
+               e.rules?.["json-schema-validator/no-invalid"],
+           );
+           console.log(
+             JSON.stringify(
+               entry?.rules["json-schema-validator/no-invalid"]?.[1] ?? null,
+             ),
+           );`,
+        ],
+        { encoding: 'utf8' },
+      );
+      const options = JSON.parse(stdout);
+      expect(options).not.toBeNull();
+
+      const inline = options.schemas[0].schema;
+
+      // A "cleanup" back to a path string would silently restore the stale-cache
+      // bug: the rule validates via the files' own $schema either way, so only
+      // this inline copy makes the schema content part of the cache hash.
+      expect(inline).toBeTypeOf('object');
+      expect(inline).toEqual(
+        JSON.parse(
+          fs.readFileSync(path.join(pkgRoot, 'i18n.schema.json'), 'utf8'),
+        ),
+      );
+    },
+  );
+});
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -208,6 +208,8 @@ check_z_index_ladder() {
 run_check "Relay" check_relay_drift
 # lint:ci = the cached eslint variant CI runs (content-hash cache; modified
 # files are always re-linted). Uncached equivalent: `pnpm -r lint`.
+# backend.ai-client's lint:ci is deliberately uncached: its type-aware
+# no-floating-promises rule can flag a caller whose own content is unchanged.
 run_check "Lint" pnpm -r --stream lint:ci
 run_check "Format" pnpm run format
 run_check "TypeScript" pnpm --prefix ./react exec tsc --noEmit --incremental
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -206,9 +206,11 @@ check_z_index_ladder() {
 }
 
 run_check "Relay" check_relay_drift
-run_check "Lint" pnpm -r --stream lint
+# lint:ci = the cached eslint variant CI runs (content-hash cache; modified
+# files are always re-linted). Uncached equivalent: `pnpm -r lint`.
+run_check "Lint" pnpm -r --stream lint:ci
 run_check "Format" pnpm run format
-run_check "TypeScript" pnpm --prefix ./react exec tsc --noEmit
+run_check "TypeScript" pnpm --prefix ./react exec tsc --noEmit --incremental
 # The react lane reaches backend.ai-{ui,client} through tsconfig `paths`,
 # but nothing pulls in the agent CLI, so it gets its own lane.
 run_check "TypeScript (agent-cli)" pnpm --filter backend.ai-agent-cli exec tsc --noEmit
--- a/scripts/verify.sh
+++ b/scripts/verify.sh
@@ -208,8 +208,11 @@ check_z_index_ladder() {
 run_check "Relay" check_relay_drift
 # lint:ci = the cached eslint variant CI runs (content-hash cache; modified
 # files are always re-linted). Uncached equivalent: `pnpm -r lint`.
-# backend.ai-client's lint:ci is deliberately uncached: its type-aware
-# no-floating-promises rule can flag a caller whose own content is unchanged.
+# backend.ai-client's and backend.ai-agent-cli's lint:ci are deliberately
+# uncached: their type-aware no-floating-promises rule can flag a caller whose
+# own content is unchanged. The coverage gate fails when a package defines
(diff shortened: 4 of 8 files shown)


scripts/verify.sh
```diff
@@ -184,9 +184,11 @@ check_z_index_ladder() {
 }
 
 run_check "Relay" check_relay_drift
-run_check "Lint" pnpm -r --stream lint
+# lint:ci = the cached eslint variant CI runs (content-hash cache; modified
+# files are always re-linted). Uncached equivalent: `pnpm -r lint`.
+run_check "Lint" pnpm -r --stream lint:ci
```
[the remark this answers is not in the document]
ironAiken2: Fixed in 624374c6b by making the schema part of the config hash instead of re-linting without cache: `packages/backend.ai-ui/eslint.config.js` now reads `i18n.schema.json` and passes its **content** as an inline schema in the `json-schema-validator/no-invalid` rule options. ESLint invalidates cached results when a file's resolved config changes, so a schema-only edit now invalidates exactly the locale files' cache entries (the `.ts` files' config slice is unchanged, so their cache stays warm). Verified empirically: with a warm cache, breaking the schema made the next `lint:ci` fail with 11,156 violations; restoring it returned green. This keeps the warm run at ~4.6s vs paying an uncached locale re-lint on every verify.

packages/backend.ai-client/package.json
```diff
@@ -47,6 +47,7 @@
     "vitest": "vitest run",
     "vitest:watch": "vitest",
     "lint": "eslint ./src",
+    "lint:ci": "eslint ./src --cache --cache-strategy content --cache-location .eslintcache",
```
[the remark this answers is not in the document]
ironAiken2: Fixed in 624374c6b: `backend.ai-client`'s `lint:ci` is now plain `eslint ./src` (no cache flags), exactly as you suggest — the type-aware `no-floating-promises` rule makes a content-hash cache unsound for this package. This also matches CI, which already runs this package's lint uncached (`typecheck.yml` → `pnpm --filter backend.ai-client run lint`). Cost is ~5s per warm verify.sh run (total stays ~25–31s). The script is kept (rather than removed) so `pnpm -r lint:ci` retains the same package coverage as `pnpm -r lint`; a comment in `scripts/verify.sh` records why this one is uncached, and the PR body's safety rationale is updated.

packages/backend.ai-client/package.json
```diff
@@ -47,6 +47,7 @@
     "vitest": "vitest run",
     "vitest:watch": "vitest",
     "lint": "eslint ./src",
+    "lint:ci": "eslint ./src",
```
agatha197, line 50: This line is duplicated with line 49.

agatha197 (changes requested):

packages/backend.ai-client/package.json
ironAiken2: @copilot please check this

ironAiken2: Please revert ur changes. @copilot

ironAiken2: Intentional, though it does look odd in isolation. Two constraints meet here:

1. `lint:ci` has to exist in this package: verify.sh runs `pnpm -r --stream lint:ci`, and pnpm `-r` silently skips any package that doesn't define the script — removing it would drop this package from the Lint step without any visible failure.
2. Unlike react / backend.ai-ui (where `lint:ci` adds `--cache`), this package's `lint:ci` is deliberately **uncached**: its type-aware `no-floating-promises` rule can flag a caller whose own content is unchanged, so a content-hash cache is unsound (see the Copilot thread above). That makes it textually identical to `lint`.

An alias form (`lint:ci` → `run lint`) was tried in f35da98 and reverted in e543253 — keeping the explicit command. The rationale is recorded in the comment above the Lint step in `scripts/verify.sh` (lines 187–190).

scripts/verify.sh
```diff
@@ -184,9 +184,13 @@ check_z_index_ladder() {
 }
 
 run_check "Relay" check_relay_drift
-run_check "Lint" pnpm -r --stream lint
+# lint:ci = the cached eslint variant CI runs (content-hash cache; modified
+# files are always re-linted). Uncached equivalent: `pnpm -r lint`.
+# backend.ai-client's lint:ci is deliberately uncached: its type-aware
+# no-floating-promises rule can flag a caller whose own content is unchanged.
+run_check "Lint" pnpm -r --stream lint:ci
```
agatha197, line 216: `pnpm -r run` errors only when *no* selected package has the script; packages that lack it are skipped with no warning and exit 0 (verified: `pnpm --filter backend-ai-webui-react --filter backend.ai-agent-cli --stream run lint:ci` runs react only). `backend.ai-agent-cli` has landed on main with a `lint` but no `lint:ci`, so after the rebase this line silently drops that package from the Lint gate — please add its `lint:ci` and a guard here that fails when a package has `lint` without `lint:ci`, otherwise every future package reopens the same hole.

packages/backend.ai-ui/eslint.config.js
```diff
@@ -109,6 +117,14 @@ export default [
     },
     rules: {
       ...jsonSchemaValidator.configs.recommended.rules,
+      "json-schema-validator/no-invalid": [
+        "error",
+        {
+          schemas: [
+            { fileMatch: ["**/src/locale/*.json"], schema: i18nSchema },
```
agatha197, line 124: This schema is never used for validation: `no-invalid` picks `$schema || options || catalog` (`lib/rules/no-invalid.js:373`) and every locale file carries `"$schema": "../../i18n.schema.json"`, so `$schema` always wins. The inline copy works purely as a cache-hash tripwire and nothing pins that behaviour — a later "cleanup" back to a path reference would silently restore the stale-cache bug this PR is fixing, so it needs a regression guard (or `mergeSchemas` so the option is actually live).

agatha197 (changes requested):
Checked the premises and they hold: the eslint cache hash does cover the resolved config (eslint 9.39.5 `lib/cli-engine/lint-result-cache.js:56`), the two cached packages carry no type-aware or cross-file rules, and `tsc --incremental --noEmit` on TS 6.0.3 still reports errors on a repeat run and re-derives them across file boundaries. Two things below block merge — and since the branch is 126 commits behind main and conflicts in `verify.sh`, the rebase is exactly where the first one bites.

scripts/verify.sh
ironAiken2: Fixed in 202e05f8a (on top of a rebase onto main so the package exists on this branch):

- `backend.ai-agent-cli` gains `lint:ci` — **uncached**, same soundness reason as `backend.ai-client`: it enables type-aware `no-floating-promises` with `projectService`, so a content-hash cache could skip a caller invalidated by a dependency's return-type change.
- New guard `scripts/lint-ci-coverage-gate.mjs`, run by verify.sh right before the Lint step: it walks the `pnpm-workspace.yaml` package globs and fails when any package defines `lint` without `lint:ci`, naming the offenders. Verified both ways: removing agent-cli's `lint:ci` makes the gate fail with `Packages with a \`lint\` script but no \`lint:ci\`: packages/backend.ai-agent-cli`; restoring it returns green. Costs <0.1s per run.

Thanks for catching this — the silent-skip behavior I described in the other thread had already bitten this branch.

packages/backend.ai-ui/eslint.config.js
ironAiken2: Confirmed against the plugin source (`no-invalid.js`: `validatorsCtx.$schema || validatorsCtx.options || validatorsCtx.catalog`) — the inline copy is validation-dead and works purely as a cache-hash tripwire. Pinned in 202e05f8a with a regression test rather than `mergeSchemas`:

- `packages/backend.ai-ui/src/locale/__tests__/i18nSchemaCacheTripwire.test.ts` loads the flat config in a node subprocess (vitest's transform breaks the config's own `import.meta.url`, and native import matches how ESLint loads it) and asserts the `no-invalid` rule options carry the schema **content** — an object deep-equal to the on-disk `i18n.schema.json` — not a path string. A cleanup back to a path reference fails with `expected './i18n.schema.json' to be type of 'object'` (verified by making that exact edit).
- Chose the guard over `mergeSchemas` because merging would run both validators on every locale file — double validation cost on cache misses and duplicate error reports for each violation — while buying nothing the $schema validation doesn't already provide.

The PR body's soundness section now records this caveat too.

packages/backend.ai-ui/src/locale/__tests__/i18nSchemaCacheTripwire.test.ts
```diff
@@ -0,0 +1,60 @@
+import { execFileSync } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+// Guards the lint-cache tripwire: eslint.config.js must inline i18n.schema.json's
+// CONTENT (not a path) so schema edits invalidate the locale files' eslint cache.
+describe('i18n schema lint-cache tripwire', () => {
```
agatha197, line 9: `Vitest (backend.ai-ui)`'s `paths:` filter is `src/**`, `package.json`, `vitest.config.ts` and the workflow file, and verify.sh runs no vitest step — so a PR touching only `packages/backend.ai-ui/eslint.config.js`, which is exactly the cleanup-to-a-path-reference this test guards against, never runs it.

Please add `packages/backend.ai-ui/eslint.config.js` and `packages/backend.ai-ui/i18n.schema.json` to that workflow's `paths:`; the latter also restores locale revalidation in CI for a schema-only PR.

agatha197 (changes requested):
Re-verified both earlier findings end to end: the coverage gate flags `backend.ai-agent-cli` and `backend.ai-client` on main's tree and is green here (8 packages scanned), the tripwire test fails with `expected './i18n.schema.json' to be type of 'object'` when the inline schema is reverted to a path, and on a warm cache a schema-only edit re-lints exactly the 21 locale files (2.9s) while the `.ts` entries stay warm. One gap left in what actually runs the new guard.

packages/backend.ai-ui/src/locale/__tests__/i18nSchemaCacheTripwire.test.ts
ironAiken2: Fixed in 435ce2bf: added `packages/backend.ai-ui/eslint.config.js` and `packages/backend.ai-ui/i18n.schema.json` to the workflow's `paths:`, so a config-only cleanup PR now runs the tripwire test and a schema-only PR revalidates the locale files in CI. Thanks — the guard was blind to exactly the PR shape it exists for.

agatha197 (approved):
LGTM

PR #9062, opened 2026-08-25
fix(FR-3670): move folder-explorer row actions onto BAINameActionCell

### LINKED ISSUE #9051: Folder explorer: the third row action (more "⋮" button) is clipped by the Controls column's default width
## Where

- **Page**: the folder explorer (file browser) — opened from the Data page by

entering a folder; rendered in `FolderExplorerModal` / `FolderExplorerModalV2`

- **Region**: the file table's `Controls` column (per-row action buttons)
- **Component** (if identified): `BAIFileExplorer` /`FileItemControls`

(`packages/backend.ai-ui/src/components/baiClient/FileExplorer/`)

## Observed

Each row's `Controls` cell holds three actions — download, delete, and a "more"

(⋮) button. At the column's default width only the first two fit; the ⋮ button

is cut off at the column edge and cannot be clicked. It only becomes visible

after the column/table is widened (first screenshot). The table shows a

horizontal scrollbar in this state.

## Expected

All three row actions are visible and clickable at the default column width,

without widening the table or scrolling horizontally.

## Environment

- **Mode**: dark (from the screenshots; light not checked)
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: Unknown

## Reproduction

1. Open the Data page and enter a folder to open the folder explorer.
1. Look at any file row's `Controls` column at the default width.

## Evidence

- Screenshot — three buttons, after widening: `/home/ubuntu/.claude/jobs/cadd255b/pasted-1.png` (to be attached in Jira)
- Screenshot — default width, ⋮ clipped: `/home/ubuntu/.claude/jobs/cadd255b/pasted-2.png` (to be attached in Jira)

## Notes

Possibly related to the column's fixed width vs. the number of rendered actions;

not verified.

JIRA Issue: FR-3669

### LINKED ISSUE #9052: Folder explorer: proposal — replace the Controls column with BAINameActionCell, as most other tables use
## Question

Should the folder explorer's file table drop its separate `Controls` column and

put the row actions on the name cell via `BAINameActionCell`, the way most other

tables in the app do?

## Where

- **Page**: the folder explorer (file browser) — opened from the Data page by

entering a folder

- **Component** (if identified): `BAIFileExplorer` / `FileItemControls`

(`packages/backend.ai-ui/src/components/baiClient/FileExplorer/`)

- **Recurs in**: `BAINameActionCell` is used by ~69 files across `react/src` and

`packages/backend.ai-ui/src` (e.g. `VFolderNodes.tsx`, `ResourcePresetList.tsx`,

`AdminUserManagement.tsx`, `KeypairResourcePolicyList.tsx`,

`AutoScalingRuleListNodes.tsx`). The folder explorer is one of the tables that

still carries a dedicated `Controls` column instead.

## Current behaviour

The file table renders a dedicated `Controls` column whose cell holds the

download / delete / more (⋮) buttons for that row. Other list tables in the app

render their row actions through `BAINameActionCell` on the name column, which

handles the visible-action set and the overflow more-menu itself.

## Why this is being raised

The folder explorer is inconsistent with the rest of the app's tables, and the

`Controls` column is where the clipped ⋮ button was observed — the shared cell

already owns action layout and overflow, so the divergence is what makes this

table need its own answer to it.

## Proposal

> Replace the folder explorer's `Controls` column with `BAINameActionCell` on

> the name column, matching the other tables.

## Impact if it stays as-is

Cosmetic inconsistency, with a maintenance cost: row-action behaviour has to be

solved twice.

## Environment

- **Mode**: Unknown (not mode-specific)
- **Build**: Unknown

## Evidence

- Screenshot — the `Controls` column: `/home/ubuntu/.claude/jobs/cadd255b/pasted-2.png` (to be attached in Jira)

## Notes

Prior art: FR-2408 added `BAINameActionCell`, FR-2409 / FR-2411 rolled it onto

`VFolderNodes` and other tables. FR-3528 is an open question about the same

cell's disabled-action styling.

JIRA Issue: FR-3670

agatha197:
Resolves #9052 (FR-3670)
Resolves #9051 (FR-3669)

## What

The folder explorer's file table had its own `Controls` column pinned at
`width: 80`, holding up to three row actions (download, delete, "more"). Almost
every other table in the app puts its row actions on the name cell through
`BAINameActionCell`. This moves the explorer onto that component and drops the
`Controls` column.

## After

| Light | Dark |
|---|---|
| after light | after dark |

All three actions visible per row, no `Controls` column. The "before" is the
pair of screenshots on #9051.

## Mechanism

The clipping was **not** a styling problem — it was a fixed width that could not
know how many actions the cell held. Measured live, the three action buttons
occupy **76px**; the cell adds 8px of padding each side, so the row needs ~92px
inside an 80px column. The third button (`⋮`) fell ~12px past the column edge
and became unclickable — which is why widening the column would only have moved
the failure to the next narrower layout.

`BAINameActionCell` measures its own container with a `ResizeObserver` and
collapses whatever does not fit into an overflow menu, so the fixed width that
caused the clipping is gone rather than enlarged.

## Measured, live (admin, 26.9.0-alpha.0 backend)

Per row, in the folder explorer modal. `overflowRight` is the action group's
right edge minus the cell's — negative means fully inside the cell.

| Viewport | title area | actions | buttons shown | overflowRight |
|---|---|---|---|---|
| 1920 (light) | 87px | 76px | **3** (download, delete, ⋮) | −8px |
| 1920 (dark) | 87px | 76px | **3** | −8px |
| 1280 (light) | 46px | 50px | **2** + overflow menu | −8px |

Before, the same three buttons had a hard 80px box and the third was cut off
(see the screenshots on #9051). Nothing is clipped at any of the three sizes
now; when space runs out the actions collapse into `⋮` instead.

Dark mode entered through the header toggle with
`document.documentElement.dataset.theme === 'dark'` asserted. Action colours
resolve to `rgb(3,135,191)` (`colorInfo`) and `rgb(190,61,63)` (`colorError`) —
they come from `BAINameActionCell`'s own palette now, so the per-icon inline
colours the old cell carried are gone.

## Three things that fell out of sharing one cell

All caught by the live probe, not by the type checker:

1. **The name has to yield width.** With the raw title node the file name
   rendered at its intrinsic 190–197px inside an 87px title area and painted
   *over* the buttons. The name now shrinks and ellipsizes (with the truncate
   tooltip it already had).
2. **Row actions must keep `stopPropagation`.** The old buttons called it
   themselves; `BAINameActionCellAction.onClick` receives no event, and the
   explorer table has its own row `onClick`. Without the guard, clicking `⋮`
   also ticked the row's checkbox. Verified: clicking `⋮` leaves 0 rows
   selected, clicking the Size cell still selects 1.

3. **The folder icon and the directory name broke onto two lines.** The
   directory branch had them as inline siblings inside one `BAILink`,
   separated by a space, so the narrowed cell wrapped between them and left
   the icon alone above the name. The file branch never had this — it already
   lays the two out as flex children. The directory branch now matches it
   (icon beside the link, `flex-shrink: 0`), and the link carries its own
   truncation rather than `BAILink`'s `ellipsis` wrapper, whose `BAIText` gets
   no `min-width: 0` and so will not shrink as a flex item. Verified: icon and
   name share a line (tops within 4px) and the name block is 24px — one line —
   at both 1920 and 1280, for a short Hangul name and a 39-char one.

## Exercised

Non-destructively, except one scratch directory created and then deleted
through the new delete action itself:

- directory navigation (the FR-3602 handler, moved verbatim) — breadcrumb
  advances, row count changes;
- the `⋮` overflow menu opens and contains `Edit File`;
- the delete action opens `BAIDeleteConfirmModal` with the typed-confirm gate
  ("Type `<name>` to confirm", "This action cannot be undone");
- 0 `pageerror` across every pass, light and dark.

## Component changes

`BAINameActionCell` gains one optional action field, and `renderTitle` gets
narrower about what it wraps:

- **`loading` on an action** — for progress the cell does not own (the
  explorer's background delete tracked via `deletingFilePaths`). Additive and
  defaulted off. Without it the in-flight delete spinner would have been
  silently dropped.
- **Only string titles get the default `BAIText` treatment.** The explorer's
  title is an inline-rename field, and `BAIText`'s truncation + truncate-tooltip
  would swallow it. Rather than add a flag for that, the component now trusts
  what the prop type already says: a `ReactNode` title brings its own
  presentation. Of the six call sites passing a node, **four already hand it a
  `BAIText` they configured themselves** (`ProjectStoragePermissionTable`,
  `KeypairResourcePolicyStoragePermissionTable[V2]`,
  `DeploymentAccessTokensCard`) and were getting a second `BAIText` wrapped
  around it — verified in the DOM: those title cells now contain exactly one
  `.astryx-text` node instead of two, and render unchanged. The one call site
  that genuinely relied on the outer wrapper, `QuotaScopeTable`'s
  `<code>{value}</code>`, now states its `ellipsis`/tooltip locally. All 34
  string-title call sites are untouched.

`FileItemControls` is replaced by `FileNameCell` (it renders the name cell now,
not a control strip, and still owns the download mutation and its Suspense
boundary).

## i18n

Adds `comp:FileExplorer.Download` across all 21 locales, taking the value each
already carried for `DownloadSelected` (the bare verb in every language — no new
translation invented). Removes `comp:FileExplorer.Controls` and
`comp:FileExplorer.MoreOptions`, both orphaned by this change.

## Behaviour changes worth a reviewer's eye

- The `⋮` button is no longer disabled on directories. It used to be, which
  would now hide overflowed actions on directory rows — the same class of bug
  as #9051. The `Edit File` row inside it stays disabled and states its reason.
- The `Edit File` row is only offered where editing is permitted at all;
  previously a permanently-disabled row was the entire menu when it was not.

## Residue

- The Name column is narrow in the modal layout (87px of title area at 1920,
  because the metadata panel takes half the dialog), so longer names ellipsize
  earlier than they used to *appear* to — previously they overflowed the column
  and collided with the controls instead. The truncate tooltip shows the full
  name. Rebalancing the explorer's column widths is a separate layout question;
  I tried `width: '40%'` on the Name column and measured **no effect** (identical
  87px), so I reverted it rather than leave an inert prop in the tree.
- The picker mode's non-directory rows are unchanged — they render a disabled
  name with no actions, as before.

## Verification

- `pnpm --filter backend.ai-ui exec vitest run` → 759 passed, 1 skipped (63 files)
- `pnpm --prefix ./react exec vitest run` → 1627 passed (102 files)
- `pnpm --filter backend.ai-ui build` → clean
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → exits 1, but on
  the same 3 pre-existing files as `main` (`BAIDialog.css`, `BAIDrawerPortal.css`,
  `BAIText.css`); this PR adds no CSS and no file of mine is flagged.
- Terminology warn on `comp:BAIResourceUnitGrid` is pre-existing and unrelated.


nowgnuesLee (approved):
LGTM

PR #9066, opened 2026-08-25
fix(FR-3677): default deployment presets to single-node and warn on multi-node

### LINKED ISSUE #9065: Convert multi-node x1 to single-node x1 when saving a deployment preset
The session launcher already normalizes a multi-node x1 selection to `single-node` before sending the create request (FR-2381), because multi-node requires an overlay network that all-in-one deployments may not have configured. `DeploymentAddRevisionModal` does the same for deployment revisions.

The deployment preset setting page does not. `handleSubmit` in `react/src/pages/AdminDeploymentPresetSettingPage.tsx` passes `values.clusterMode` straight into both the create and the update mutation input, so a preset saved with **Multi-node** and **cluster size 1** is persisted as `MULTI_NODE` / `1`. The form default is exactly that combination (`clusterMode: MULTI_NODE, clusterSize: 1`), so this is the state an admin gets without touching the cluster fields at all.

## Expected

Saving a preset with cluster mode `MULTI_NODE` and cluster size `1` should persist `SINGLE_NODE`, matching the session launcher and the deployment revision paths.

## Scope

- `react/src/pages/AdminDeploymentPresetSettingPage.tsx` — normalize `clusterMode` in `handleSubmit` for both `commitCreate` and `commitUpdate`.

JIRA Issue: FR-3677

agatha197:
Resolves #9065 (FR-3677)

## Problem

The deployment preset form defaults to cluster mode **Multi-node** with **cluster size 1** (`AdminDeploymentPresetSettingPageContent.tsx`). Multi-node requires an overlay network that all-in-one deployments may not have configured, so a preset saved with that untouched default produces sessions that fail to start — and the admin gets no signal at save time.

## Change

Previous revision of this PR silently rewrote `MULTI_NODE` x1 to `SINGLE_NODE` in `handleSubmit`, mirroring the session launcher (FR-2381). **That normalization is reverted** — `AdminDeploymentPresetSettingPage.tsx` is back to `main`. This is an admin surface: the persisted value must be exactly what the admin selected, not a value the UI quietly corrected.

Two changes instead:

1. **Default is now `SINGLE_NODE`.** The safe mode is what an admin gets without touching the cluster fields.
2. **Selecting `MULTI_NODE` raises a `warningOnly` `Form.Item` rule** on the cluster-mode field:

   > If multi-node is not configured on the cluster, sessions created from this preset will fail to start.

   Same mechanism the session launcher already uses for its cluster-size-1 hint (`ClusterModeFormItems.tsx` → `session.launcher.ClusterSizeOneMultiNodeConvertInfo`). It informs without blocking submit — a multi-node preset on a properly configured cluster stays perfectly valid.

New i18n key `adminDeploymentPreset.MultiNodeNotConfiguredWarning`, added to all 21 locales.

## Scope

`AdminDeploymentPresetSettingPageContent.tsx` (default + one rule) and the 21 locale files. No mutation-input behavior change.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup, StyleX, Astryx theme build, z-index mirrors, Terminology).

Warning message with 'Multi Node'.
image
Default value is 'Single Node'.
image

### DIFF
--- a/react/src/components/AdminDeploymentPresetSettingPageContent.tsx
+++ b/react/src/components/AdminDeploymentPresetSettingPageContent.tsx
@@ -663,7 +663,9 @@ const AdminDeploymentPresetSettingPageContent: React.FC<
       };
     }
     return {
-      clusterMode: 'MULTI_NODE' as const,
+      // Single-node by default: multi-node needs an overlay network that
+      // all-in-one deployments may not have (FR-3677).
+      clusterMode: 'SINGLE_NODE' as const,
       clusterSize: 1,
       // Model definition is off by default (optional). Seed one empty model so
       // it renders once the switch is turned on.
@@ -1095,10 +1097,12 @@ const AdminDeploymentPresetSettingPageContent: React.FC<
                 )}
               </Form.List>
             </BAIFormItem>
+            {/* Top-aligned: bottom alignment would push Cluster Size down by
+                the height of Cluster Mode's validation message. */}
             <BAIFlex
               gap="md"
               wrap="wrap"
-              style={{ alignItems: 'flex-end', marginTop: token.marginMD }}
+              style={{ alignItems: 'flex-start', marginTop: token.marginMD }}
             >
               <BAIFormItem
                 name="clusterMode"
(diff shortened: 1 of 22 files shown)


yomybaby (approved):
LGTM

PR #9070, opened 2026-08-25
feat(FR-3675): pick the model mount subpath with a directory picker

### LINKED ISSUE #9063: Provide path selection option when specifying a path inside a model type folder
## Background

When a model's files live in a subfolder of a model-type vfolder, the user must either specify a mount **subpath** or move the files out to the folder root. There are flows where a path inside a model-type folder must be specified — e.g., creating a revision in a deployment (`DeploymentAddRevisionModal`, `modelMountConfig.subpath`) — but the subpath is currently entered as a plain text input, which is error-prone.

## Task

Provide a path **selection** option (browse & pick, instead of free typing) in the places where a path inside a model-type folder must be specified.

- Reuse the existing component `BAIVFolderPathPicker` (`packages/backend.ai-ui/src/components/baiClient/FileExplorer/BAIVFolderPathPicker.tsx`) — a select-like trigger that opens `BAIDirectoryPickerModal` to pick a directory inside the chosen vfolder (`''` = root, `"inner/path"` below it).
- Apply it to the deployment revision creation flow first (`modelMountConfig.subpath`), and to other model-type-folder path inputs where applicable.

JIRA Issue: FR-3675

ironAiken2:
Resolves #9063 (FR-3675)

## What & why

When a model's files live in a subfolder of the model-type vfolder, the deployment revision form requires a mount **subpath** — but it was a free-typed text input, so a typo or a nonexistent path only surfaced at serve time. This replaces the input with **`BAIVFolderPathPicker`** (its first application), which opens a directory-only picker modal on the selected model folder, so only a real directory can be picked (`''` = folder root).

## Changes (`DeploymentAddRevisionModal.tsx`)

- The Custom form's `modelSubpath` field now renders `BAIVFolderPathPicker`, keyed to the watched `modelFolderId` (`toLocalId`) and disabled until a model folder is chosen.
- A subpath only makes sense inside the folder it was picked from, so:
  - changing the model folder via the select resets the subpath field;
  - the create-new-folder callback (programmatic `setFieldValue`, which skips `onChange`) resets it too.
- Prefills from an existing revision (edit/duplicate) keep their stored subpath, as before; submit semantics are unchanged (`'' | undefined` → `null`).

The field still renders only when the manager supports `model-mount-subpath`, and the Preset form is untouched (it has no subpath field).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, i18n terminology)

https://claude.ai/code/session_013BhoynodeDDhwqQzUDV6rY

### DIFF
--- a/packages/backend.ai-ui/src/components/baiClient/FileExplorer/BAIVFolderPathPicker.stories.tsx
+++ b/packages/backend.ai-ui/src/components/baiClient/FileExplorer/BAIVFolderPathPicker.stories.tsx
@@ -296,7 +296,7 @@ The value is the **sub path inside the vfolder** — \`''\` for the vfolder root
 | \`value\` | \`string\` | - | Selected sub path (\`''\` = vfolder root) |
 | \`defaultValue\` | \`string\` | - | Initial value for uncontrolled usage |
 | \`onChange\` | \`(selectedSubPath?: string) => void\` | - | Fired when a location is confirmed in the modal |
-| \`selectProps\` | \`BAISelectProps\` subset | - | Forwarded to the sub path trigger select |
+| \`label\` | \`string\` | "Select a path" | Accessible name of the trigger (visually hidden; the surrounding Form.Item renders the visible label) |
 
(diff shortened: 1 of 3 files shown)


agatha197 (approved):
LGTM

PR #9141, opened 2026-08-26
fix(FR-3673): gate PREEMPTED / RESCHEDULING session statuses on manager 26.8.0

### LINKED ISSUE #9060: Session status filter sends PREEMPTED / RESCHEDULING to managers without those enum values, breaking the Sessions tab
## Summary

The session status filter hardcodes `PREEMPTED` and `RESCHEDULING` in its "running" status bucket. Managers that do not define those values in the `SessionV2Status` enum reject the whole query, so the screen always falls into the error boundary. Reproduced against **manager 26.4.9**.

## Reproduction

1. Connect the WebUI to a manager on **26.4.9**.
1. Open the Sessions tab (agent/node detail drawer -> Sessions, and the project-admin Sessions page take the same path).
1. The list never renders; the error boundary is shown.

Response:

```json
{
  "errors": [
    {
      "message": "Variable \"$sessionFilter\" got invalid value \"PREEMPTED\" at \"sessionFilter.status.in[7]\"; Value \"PREEMPTED\" does not exist in \"SessionV2Status\" enum. Did you mean the enum value \"PREPARED\"?",
      "locations": [{ "line": 1, "column": 53 }]
    },
    {
      "message": "Variable \"$sessionFilter\" got invalid value \"RESCHEDULING\" at \"sessionFilter.status.in[8]\"; Value \"RESCHEDULING\" does not exist in \"SessionV2Status\" enum. Did you mean the enum value \"SCHEDULED\"?",
      "locations": [{ "line": 1, "column": 53 }]
    }
  ]
}
```

Indices 7 and 8 match `PREEMPTED` / `RESCHEDULING` in `RUNNING_STATUSES` exactly.

## Root cause

`RUNNING_STATUSES` is a compile-time constant typed against the schema the WebUI is built with (`data/schema.graphql`, which does contain both values). It is sent verbatim regardless of the connected manager's version, and a GraphQL enum coercion failure is a whole-request error, not a partial one — so a single unknown value blanks the page instead of degrading.

Affected call sites (the same list is duplicated three times):

- `react/src/components/AgentNodeItems/AgentSessions.tsx:36` - `RUNNING_STATUSES`
- `react/src/pages/ProjectAdminSessionPage.tsx:43` - `RUNNING_STATUSES`
- `react/src/components/DashboardPanels/sessionFilterProperties.ts:33` - `SESSION_STATUSES` (queryfilter string, lower risk, but the same drift)

## Proposed fix

Gate the two values on the manager version, following the existing idiom (`react/src/components/PendingSessionNodeList.tsx:53`):

```
const runningStatuses = baiClient.isManagerVersionCompatibleWith('<min-version>')
  ? RUNNING_STATUSES
  : RUNNING_STATUSES.filter((s) => s !== 'PREEMPTED' && s !== 'RESCHEDULING');
```

Preferably factor the bucket into one shared helper so the three copies cannot drift again.

## Open question

The exact manager version that introduced `PREEMPTED` / `RESCHEDULING` into `SessionV2Status` needs confirmation from the manager side. 26.4.9 does not have them; the WebUI's bundled schema does. That version is the threshold for the gate.

## Acceptance criteria

- On a manager without `PREEMPTED` / `RESCHEDULING`, the Sessions tab and the project-admin Sessions page load and filter without hitting the error boundary.
- On a manager that has them, both statuses are still included in the running bucket.
- The running/finished status buckets live in one place rather than being duplicated per call site.

JIRA Issue: FR-3673

agatha197:
Resolves #9060 (FR-3673)

## Problem

On manager **26.4.9**, the Sessions tab (agent detail drawer) and the project-admin Sessions page always land in the error boundary:

```json
{
  "errors": [
    { "message": "Variable \"$sessionFilter\" got invalid value \"PREEMPTED\" at \"sessionFilter.status.in[7]\"; Value \"PREEMPTED\" does not exist in \"SessionV2Status\" enum. Did you mean the enum value \"PREPARED\"?" },
    { "message": "Variable \"$sessionFilter\" got invalid value \"RESCHEDULING\" at \"sessionFilter.status.in[8]\"; Value \"RESCHEDULING\" does not exist in \"SessionV2Status\" enum. Did you mean the enum value \"SCHEDULED\"?" }
  ]
}
```

Indices 7 and 8 are exactly where `PREEMPTED` / `RESCHEDULING` sat in the hardcoded `RUNNING_STATUSES` array.

## Cause

`RUNNING_STATUSES` is a compile-time constant typed against the schema the WebUI is *built* with, and it was sent verbatim to whatever manager the user is *connected* to. A GraphQL enum coercion failure rejects the whole request, not the offending value — so one unknown status blanks the page instead of degrading.

The threshold is manager **26.8.0**, verified against the manager tags:

| manager tag | `SessionV2StatusGQL` contains PREEMPTED / RESCHEDULING |
|---|---|
| 26.4.10 | no |
| 26.7.0 | no |
| 26.8.0 | yes |

Introduced by lablup/backend.ai#13126 (`feat(BA-6749): route preemption victims through PREEMPTED and RESCHEDULING`), merged before the 26.8.0 version freeze.

## Change

- `client.ts` declares a `session-preemption-statuses` feature flag in the existing `26.8.0` block of `_updateSupportList()`. Version comparison lives with every other manager capability, and `supports()` memoizes the list on first call rather than re-comparing per render.
- `getSessionV2StatusBuckets(supportsPreemptionStatuses)` in `react/src/helper/sessionStatusBuckets.ts` owns the running / finished buckets and drops the two statuses when the flag is off. It is a plain function, not a hook: each consumer reads `supports('session-preemption-statuses')` itself and passes the boolean in — the shape `buildModelDefinitionInput` already uses in `AdminDeploymentPresetSettingPage`.
- The bucket list had been copy-pasted into `AgentSessions.tsx` and `ProjectAdminSessionPage.tsx`. One home means the gate cannot be applied to one copy and missed on the other — the duplication is what the issue's third acceptance criterion asks to remove.
- `SessionV2Status` is now exported from `backend.ai-ui`, so the shared helper does not pin the union to an unrelated page's Relay artifact.

## Out of scope

`react/src/components/DashboardPanels/sessionFilterProperties.ts` lists the same two statuses, but it targets the **legacy queryfilter string** API (a different enum), it only fires when a user explicitly picks that value in a custom dashboard panel, and gating it would mean threading a version flag through the whole panel-resource registry's `getStringFilterProperties` contract. Left for a follow-up; all three acceptance criteria on the issue are about the V2 enum path fixed here.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, StyleX, Astryx theme build, z-index mirrors, Terminology)
- `react/src/helper/sessionStatusBuckets.test.ts` → 3 passed: includes both statuses when the flag is on, drops them when it is off, finished bucket unchanged either way. The helper is pure, so the test calls it directly and needs no client mock.


react/src/helper/sessionStatusBuckets.ts
nowgnuesLee, line 1: I don't think this needs to be extracted into a custom hook.

nowgnuesLee (changes requested):

react/src/helper/sessionStatusBuckets.ts
agatha197: But it is used in 3 components...

nowgnuesLee: Being used in multiple places doesn't automatically mean this should be a custom hook — the logic is simple enough as is. If you'd still like to extract it, I'd prefer a plain helper function that takes `isSupportPreemptionStatus` as an argument.

nowgnuesLee (changes requested):

react/src/helper/sessionStatusBuckets.ts
agatha197: Agreed — done in cd979ef.

It is now `getSessionV2StatusBuckets(supportsPreemptionStatuses)`, a plain function in `react/src/helper/sessionStatusBuckets.ts`. Each consumer reads `supports("session-preemption-statuses")` itself and passes the boolean in, which matches how `buildModelDefinitionInput` takes its flags in `AdminDeploymentPresetSettingPage`. You were right that there was no precedent for the hook: every existing hook that calls `supports()` (`useCSVExport`, `useStartSession`, `useWebUIMenuItems`) has its own substantive job and reads the flag along the way — none exists solely to wrap a capability gate.

The buckets still live in one module, so the duplication the third acceptance criterion asks to remove stays removed. The test dropped its hook mock and calls the function directly. `scripts/verify.sh` passes.

nowgnuesLee (approved):
LGTM

PR #9157, opened 2026-08-26
docs: update admin_menu docs (2026-08-26)

lablup-octodog:
## What changed and why

**fix(FR-3718): target resource presets by id in delete and modify mutations (#9143)** — commit `71b7a41f` — changed the resource-preset admin UX in two user-visible ways:

1. The `resource-presets-per-resource-group` capability gate around the **Resource Group** field in `ResourcePresetSettingModal.tsx` was removed — the field now always renders, stays clearable, and an empty value makes the preset the manager's *global* preset (per the code comment retained on the field).
2. The client-side duplicate-name validator (and its `resourcePreset.PresetNameAlreadyExists` key, deleted from all locales) is gone; mutations now target the preset **id**. Names must still be unique, but a duplicate is now rejected server-side and surfaced as an error message on **Create**, not as an inline field error.

The "Manage resource preset" section still described the old behavior in all four languages. This PR updates it:

- **Resource Group** bullet: dropped "(Conditional)"; documents the field as optional/clearable with the global-preset semantics.
- Replaced "You cannot create the same resource preset name that already exists, since it is the key value…" with: names must be unique, and a duplicate is rejected by the server with an error message on **Create**.
- **th only**: corrected the Create Preset button name from **สร้างพรีเซต** (matches no live label in this namespace) to **สร้างค่าที่กำหนดไว้ล่วงหน้า** (`resourcePreset.CreatePreset` in `resources/i18n/th.json`), per the terminology-precedence rule.

Scope: `src/{en,ko,ja,th}/admin_menu/admin_menu.md`.

## Deliberately skipped

- Six recapture-only images planned for this chapter (`detailed_agent_node_usage_information.png`, `information_page.png`, `user_detail_dialog.png`, `keypair_detail_dialog.png`, `storage_host_detail_drawer.png`, `fair_share_usage_bucket_modal.png`, all FR-3694 `BAIMetadataList` restyles, `c16477c9e`) — the screenshot environment (`http://10.82.1.75:8090`) was unreachable this run, so the batch produced no files. Existing images left in place; re-queue on the next nightly run with a live environment.
- No other prose in this chapter is affected by the range; other FR-3718 hunks (e.g. `QuotaScopeTable.tsx` id ellipsis) were assessed and skipped per the plan.

## Review findings (independent, by Reporter)

- **Prose claims verified against source.** I confirmed in commit `71b7a41f` and the current `ResourcePresetSettingModal.tsx`: the capability gate is removed and the field renders unconditionally with `allowClear`; the duplicate-name validator and its i18n key are deleted; the only create-failure path is `message.error(...)` driven by server errors. Both new claims are supported; no unsupported behavior is asserted.
- **Translation parity: OK.** All four languages carry the same two edits and both `TODO(screenshot)` markers with matching structure.
- **UI labels verified** against live i18n in all four languages (**Create** = 作成/생성/สร้าง; **Create Preset** = プリセットの作成/프리셋 생성/สร้างค่าที่กำหนดไว้ล่วงหน้า). The Thai label fix is correct.
- **No scope creep** — 4 files, this chapter only, no image churn.
- ⚠️ **Screenshots owed**: `create_resource_preset_dialog.png` and `modify_resource_preset_dialog.png` were **not** recaptured (environment unreachable). The existing captures predate FR-3718, so on servers that gated the field they may not show the Resource Group field the updated prose now describes. Both locations carry explicit `<!-- TODO(screenshot): … -->` markers naming the route and required content; a human capture is needed. Same for the six FR-3694 manifest images above. Not a merge blocker, but the markers should not be forgotten.
- Minor nit (non-blocking): "names must **still** be unique" reads slightly changelog-relative for a fresh manual reader; consistent across all four languages, safe to keep or smooth in review.

## Verified by the wrapper

- Diff touches only `packages/backend.ai-webui-docs/` within this chapter's scope.
- `bash scripts/verify.sh` passes (`=== ALL PASS ===`).

## Convention note

This nightly docs-sync PR carries no `FR-` number by design; it deviates from the repo's `prefix(FR-XXXX)` title convention and will not be linked by `project-status-sync`.

---

### DIFF
--- a/packages/backend.ai-webui-docs/src/en/admin_menu/admin_menu.md
+++ b/packages/backend.ai-webui-docs/src/en/admin_menu/admin_menu.md
@@ -1399,19 +1399,20 @@ would not be shown.
 The resource preset dialog includes:
 
 - **Preset Name**: A unique name for the preset (only alphanumeric characters, periods, hyphens, and underscores allowed).
-- **Resource Group**: (Conditional) Associate the preset with a specific resource group.
+- **Resource Group**: Associate the preset with a specific resource group. This field is optional — leave it empty (or clear it) to keep the preset global, so it applies regardless of the resource group.
 - **Resource Preset**: Dynamic fields for each available resource type (CPU, Memory, GPU, etc.). Memory fields support dynamic unit input (`MiB`, `GiB`, `TiB`, `PiB`).
 - **Shared Memory**: The amount of shared memory allocated for the preset. This value must be less than the **Memory** value.
 
 ![](../images/modify_resource_preset_dialog.png)
+<!-- TODO(screenshot): /environment -> Resource Presets tab -> Edit (pencil) action on a preset, showing the Edit Resource Preset modal with the Resource Group field. Not captured: the screenshot environment was unreachable on this run. -->
 
 You can also create a resource preset by clicking the **Create Preset** button in the
-right top of the Resource Presets tab. You cannot create the same resource
-preset name that already exists, since it is the key value for distinguishing
-each resource preset.
+right top of the Resource Presets tab. Resource preset names must still be
+unique. If you enter a name that already exists, the server rejects the request
+when you click **Create** and an error message is displayed.
 
 ![](../images/create_resource_preset_dialog.png)
-<!-- TODO: Re-capture create_resource_preset_dialog.png — needs update. -->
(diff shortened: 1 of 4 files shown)


nowgnuesLee (approved):
LGTM

PR #9162, opened 2026-08-27
fix(FR-3712): stop accelerator slot fields failing as "not a valid undefined"

### LINKED ISSUE #9133: Resource preset create/edit: every accelerator slot fails validation with "<label> is not a valid undefined", blocking save
## Where

Admin → Resources → **Resource Presets** → create / edit modal (`ResourcePresetSettingModal`). Every accelerator slot field — **CUDA-capable GPU** (`cuda.device`), `cuda.shares`, `rocm.device`, `atom.device`, … i.e. every `resource_slots` key that is **not** `cpu` or `mem`.

## Observed

Typing any value into an accelerator field immediately shows a red validation error and the preset cannot be saved:

```
CUDA-capable GPU 유효하지 않은 undefined
```

(en: `CUDA-capable GPU is not a valid undefined`)

CPU and RAM validate fine — only the accelerator fields fail.

## Expected

The value is accepted and the preset saves. No validation error, and certainly no `undefined` in user-facing copy.

## Root cause

Two defects stack. The first blocks the save; the second is why the message reads `undefined`.

### 1. An empty placeholder rule is no longer a no-op (blocking)

`ResourcePresetSettingModal.tsx:360` builds the rules array with an empty object for non-cpu/mem slots:

```
rules={[
  _.includes(["cpu", "mem"], resourceSlotKey)
    ? { required: true, message: t("data.explorer.ValueRequired") }
    : {},                                   // <- accelerator slots
  { validator(__, value) { /* mem-only 300 PiB check */ } },
]}
```

BUI form-engine `pickMethod` (`packages/backend.ai-ui/src/form-engine/validate.ts:389`) resolves which validator handles a rule:

```
const keys = Object.keys(rule).filter((key) => key !== "message");
if (keys.length === 1 && keys[0] === "required") return "required";
if (rule.type === undefined && rule.pattern instanceof RegExp) return "pattern";
return rule.type ?? "string";           // <- `{}` lands here
```

- cpu/mem get `{ required: true, message `} → keys are exactly `["required"]` → method `"required"` → only the required check runs. **Safe.**
- accelerators get `{`} → no keys → falls through to `rule.type ?? "string"` → method `"string"` → the default branch runs `checkType`, which rejects anything that is not a `string`.

`AstryxFormNumberInput` emits a **number** (`astryxFormControls.tsx:473` — `onChange?: (value: number | null) => void`), so the implicit string check always fails.

**Why it worked before:** pre-migration this field was antd `InputNumber` with **`stringMode`** (`git show 7d7b42388^:react/src/components/ResourcePresetSettingModal.tsx:373`), so the stored value was a **string** and the same `{`} rule passed its implicit string check. The Astryx migration (`7d7b42388`) dropped `stringMode` as a documented PILOT-DECISION — correctly, since slot counts fit in float precision — but the `{`} placeholder rule was left behind and silently flipped from no-op to a failing string assertion.

### 2. `${type`} interpolates to the literal string "undefined"

The template is `form.validateMessages.types.String` (`packages/backend.ai-ui/src/locale/ko.json:660`):

```
"String": "${label} 유효하지 않은 ${type}"
```

`validateRule` (`validate.ts:417`) normalises the rule before running the validator…

```
const effectiveRule: NormalizedRule = { ...rule, type: method === "required" ? rule.type : method };
```

…but builds the interpolation map from the **raw** rule (`validate.ts:453`):

```
const kv: Record<string, any> = { ...rule, name, ...messageVariables };
```

For `{`} the raw rule has no `type`, so `replaceMessage` resolves `${type`} to `undefined` and stringifies it. `${label`} resolves fine because it comes from `messageVariables`.

This is independent of defect 1: **any** rule that relies on an inferred type rather than a declared one will print `undefined` when it fails.

## Fix

1. **Call site** — drop the `{`} placeholder. Build the rules array conditionally (e.g. `filterOutEmpty` / spread) so accelerator slots carry no rule at all, or give them a real `{ type: "number" `} rule matching what the control emits.
1. **Engine** — build `kv` from `effectiveRule`, not `rule`, so `${type`} resolves to the normalised type. Optionally have `replaceMessage` leave the placeholder intact instead of emitting `undefined` when a key is missing.

## Impact

Resource presets cannot be created or edited on any cluster that exposes an accelerator slot — the modal never passes validation. `ResourcePresetSettingModal.tsx:366` is the only empty-rule form site in `react/src` (the `{`} at `ProjectResourcePolicyList.tsx:169` is a table column placeholder, unrelated).

## Affected files

- `react/src/components/ResourcePresetSettingModal.tsx`
- `packages/backend.ai-ui/src/form-engine/validate.ts`
- `packages/backend.ai-ui/src/locale/*.json` (`form.validateMessages.types.*`, reference only)

JIRA Issue: FR-3712

agatha197:
Resolves #9133 (FR-3712)

## Problem

In the resource preset create/edit modal, typing any value into an accelerator slot field (`cuda.device`, `cuda.shares`, `rocm.device`, …) immediately failed validation with `CUDA-capable GPU is not a valid undefined`, so presets could not be created or edited on any cluster exposing an accelerator slot. CPU/RAM validated fine.

## Fix

Two stacked defects, fixed at both levels:

1. **Call site** (`ResourcePresetSettingModal.tsx`) — non-cpu/mem slots got an empty `{}` placeholder rule. The form engine's `pickMethod` types an empty rule as `string`, and since the Astryx migration dropped antd `InputNumber` `stringMode`, `AstryxFormNumberInput` emits a **number** — so the implicit string check always failed. The placeholder is replaced with a conditional spread, so accelerator slots now carry no type rule at all.

2. **Engine** (`packages/backend.ai-ui/src/form-engine/validate.ts`) — the `${type}` message interpolation map was built from the **raw** rule, which has no `type` when the type is inferred, so the template printed the literal string `undefined`. The map is now built from the normalised rule (`effectiveRule`), so an inferred type resolves to its actual name for **any** rule relying on type inference, not just this call site.

## Tests

- New `packages/backend.ai-ui/src/form-engine/validate.test.ts` — 3 regression tests for `${type}` interpolation with inferred/declared types (all pass).
- All existing form-engine tests pass (15/15).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, terminology)
- `pnpm vitest run src/form-engine` in `packages/backend.ai-ui` → 3 files, 15 tests passed

### DIFF
--- a/packages/backend.ai-ui/src/form-engine/validate.test.ts
+++ b/packages/backend.ai-ui/src/form-engine/validate.test.ts
@@ -1,0 +1,44 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+
+ FR-3712 — `${type}` interpolation for rules whose type is inferred rather
+ than declared. `validateRule` normalises `{}` to a `string`-typed rule before
+ validating, so the failure message must interpolate the normalised type, not
+ the raw rule's `undefined`.
+ */
+import { defaultValidateMessages } from './messages';
+import { validateRules } from './validate';
+import { describe, expect, it } from 'vitest';
+
+// Default (parallel) mode always rejects with the per-rule summaries.
+const collectErrors = (value: unknown, rules: any[]) =>
+  validateRules(['resource_slots', 'cuda.device'], value, rules, {
+    ...defaultValidateMessages,
+  }).then(
+    () => {
+      throw new Error('validateRules resolved in parallel mode');
+    },
+    (summaries: { errors: any[] }[]) => summaries.flatMap((s) => s.errors),
+  );
+
+describe('validateRules — inferred-type message interpolation (FR-3712)', () => {
+  it('interpolates ${type} as the normalised type for an empty rule', async () => {
+    const errors = await collectErrors(1, [{}]);
+    expect(errors).toEqual([
+      "'resource_slots.cuda.device' is not a valid string",
+    ]);
+  });
+
+  it('still resolves ${type} for an explicitly typed rule', async () => {
+    const errors = await collectErrors('one', [{ type: 'number' }]);
+    expect(errors).toEqual([
+      "'resource_slots.cuda.device' is not a valid number",
+    ]);
+  });
+
+  it('accepts a number under a declared number rule', async () => {
+    const errors = await collectErrors(1, [{ type: 'number' }]);
+    expect(errors).toEqual([]);
+  });
+});
--- a/packages/backend.ai-ui/src/form-engine/validate.ts
+++ b/packages/backend.ai-ui/src/form-engine/validate.ts
@@ -405,6 +405,9 @@ async function validateRule(
   messageVariables?: Record<string, string>,
 ): Promise<any[]> {
   let rawErrors: any[] = [];
+  // Kept in sync with the declarative branch so `${type}` interpolates the
+  // normalised type, not `undefined`, when the rule's type was inferred.
+  let effectiveRule: NormalizedRule = rule;
 
   if (rule.validator) {
     const cbArg = await runCustomValidator(rule, value);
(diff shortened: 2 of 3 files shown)


nowgnuesLee (approved):
LGTM

PR #9163, opened 2026-08-27
fix(FR-3710): feed the sort order back into the Deployment Presets table

### LINKED ISSUE #9130: Admin → Deployments → Deployment Presets: sorting is stuck ascending and the header shows no sort indicator
## Where

Admin → Deployments → **Deployment Presets** tab (`AdminDeploymentPreset`). Sortable columns: **Name**, **Created At**.

## Observed

- Clicking a sortable header never renders an active sort indicator — the header always looks unsorted.
- Every click re-sorts **ascending**. Descending is unreachable: clicking Name repeatedly keeps returning the same ascending order instead of cycling asc → desc → none.

## Expected

Same behaviour as the sibling tabs (Deployments, Prometheus Presets, Runtime Variant Presets): the clicked column shows its asc/desc arrow, and clicking cycles ascending → descending → unsorted.

## Root cause

`BAITable` treats a table as **server-sorted (controlled)** as soon as `onChangeOrder` is wired:

```
// packages/backend.ai-ui/src/components/Table/BAITable.tsx
const isOrderControlled = !!onChangeOrder || order != null;
const activeOrder = isOrderControlled ? order : uncontrolledOrder;
```

`AdminDeploymentPreset.tsx` passes `onChangeOrder` but **never passes `order` back down**, so `activeOrder` is permanently `undefined`. Consequences:

1. `sortState` is computed from `activeOrder` and is therefore always `[]` → no header arrow.
1. `useTableSortable` starts every click from the "unsorted" state, so `onSortChange` always emits `ascending` → the descending step of the cycle can never be reached.

The round trip itself is fine — `convertToOrderBy` produces a valid `DeploymentRevisionPresetOrderBy` (`NAME` / `CREATED_AT`) and the query re-fetches; only the controlled `order` feedback edge is missing.

## Fix

Mirror `AdminRuntimeVariantPreset.tsx:121`, which derives the value from the query ref and passes it to the table:

```
const order = convertFirstOrderByToString(queryRef.variables.orderBy);
// ...
<AdminDeploymentPresetTable order={order} ... />
```

`AdminDeploymentPresetTable` already spreads `...tableProps` into `BAITable` and does not `Omit` `order`, so no change is needed there.

## Notes

- `AdminDeployment.tsx:297`, `AdminPrometheusPreset.tsx:187` and `AdminRuntimeVariantPreset.tsx:201` all pass `order` — this tab is the odd one out.
- Unrelated but adjacent: `availablePresetSorterKeys` in `AdminDeploymentPresetTable.tsx` lists `rank`, but the table renders no Rank column.

## Affected files

- `react/src/components/AdminDeploymentPreset.tsx`
- `react/src/components/AdminDeploymentPresetTable.tsx` (reference only)

JIRA Issue: FR-3710

agatha197:
Resolves #9130 (FR-3710)

## Problem

On **Admin → Deployments → Deployment Presets**, the sortable **Name** and **Created At** headers never showed an active sort indicator, and every click re-sorted **ascending** — descending was unreachable.

## Root cause

`BAITable` treats a table as server-sorted (controlled) as soon as `onChangeOrder` is wired:

```ts
// packages/backend.ai-ui/src/components/Table/BAITable.tsx
const isOrderControlled = !!onChangeOrder || order != null;
const activeOrder = isOrderControlled ? order : uncontrolledOrder;
```

`AdminDeploymentPreset.tsx` passed `onChangeOrder` but never passed `order` back down, so `activeOrder` was permanently `undefined`:

1. `sortState` is derived from `activeOrder`, so it was always `[]` → no header arrow.
2. `useTableSortable` therefore restarted every click from the unsorted state, so `onSortChange` always emitted `ascending` → the descending step of the cycle was unreachable.

The round trip itself was already fine — `convertToOrderBy` produced a valid `DeploymentRevisionPresetOrderBy` and the query refetched. Only the controlled `order` feedback edge was missing.

## Fix

Derive `order` from the query ref and pass it to the table, mirroring `AdminRuntimeVariantPreset.tsx:121` — the sibling tabs (`AdminDeployment.tsx:297`, `AdminPrometheusPreset.tsx:187`, `AdminRuntimeVariantPreset.tsx:201`) all already do this; this tab was the odd one out.

```diff
+ const order = convertFirstOrderByToString(queryRef.variables.orderBy);
...
  <AdminDeploymentPresetTable
    loading={isRefetching}
+   order={order}
```

`AdminDeploymentPresetTable` already spreads `...tableProps` into `BAITable` and does not `Omit` `order`, so it needed no change.

The unsorted step of the cycle round-trips correctly: `onChangeOrder(undefined)` → `orderBy: undefined` → `convertFirstOrderByToString` reads back `null` → `sortState` is `[]`. Unlike the sibling defect in FR-3711, the state lives in the Relay query variables rather than a nuqs parser carrying `.withDefault(...)`, so there is no default to swallow the unsorted state.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Astryx theme/integration, z-index, terminology). The one terminology warning is pre-existing and unrelated (`BAIResourceUnitGrid^ChangeGroupColor`).
- Control-flow traced end to end through `BAITable`'s `isOrderControlled` / `activeOrder` / `sortState` / `useTableSortable` path.
- Not exercised against a live cluster — the change is a three-line prop wiring that mirrors an already-working sibling component.

## Out of scope

`availablePresetSorterKeys` in `AdminDeploymentPresetTable.tsx` lists `rank`, but the table renders no Rank column. Noted in the issue as unrelated; left alone.

### DIFF
--- a/react/src/components/AdminDeploymentPreset.tsx
+++ b/react/src/components/AdminDeploymentPreset.tsx
@@ -12,7 +12,7 @@ import { App } from '../app-shim';
 import AdminDeploymentPresetTable, {
   type DeploymentPresetNodeInList,
 } from '../components/AdminDeploymentPresetTable';
-import { convertToOrderBy } from '../helper';
+import { convertFirstOrderByToString, convertToOrderBy } from '../helper';
 import { buildPath } from '../helper/pathBuilder';
 import { useSuspendedBackendaiClient, useWebUINavigate } from '../hooks';
 import { theme } from '../theme-shim';
@@ -91,6 +91,7 @@ const AdminDeploymentPreset = ({
     useState<DeploymentPresetNodeInList | null>(null);
 
   const filter = queryRef.variables.filter ?? undefined;
+  const order = convertFirstOrderByToString(queryRef.variables.orderBy);
   const pageSize = queryRef.variables.limit ?? 10;
   const offset = queryRef.variables.offset ?? 0;
   const current = pageSize ? Math.floor(offset / pageSize) + 1 : 1;
@@ -186,6 +187,7 @@ const AdminDeploymentPreset = ({
             _.map(deploymentRevisionPresets?.edges, 'node'),
           )}
           loading={isRefetching}
+          order={order}
           onEdit={handleEditPreset}
           onDelete={handleDeletePreset}
           tableSettings={tableSettings}


nowgnuesLee (approved):
LGTM

PR #9164, opened 2026-08-27
fix(FR-3711): let the auto-scaling Created At sort reach ascending and unsorted

### LINKED ISSUE #9131: Deployment detail → Auto-Scaling: the Created At sort is frozen at descending, every header click is a no-op
## Where

Deployment detail page → **Auto-Scaling** card (`DeploymentAutoScalingCard` / `AutoScalingRuleListNodes`). **Created At** is the only sortable column in that table.

## Observed

Clicking the **Created At** header does nothing. The arrow stays on descending and the row order never changes — ascending is unreachable.

## Expected

The header cycles ascending → descending → unsorted like every other sortable table in the app, and the rows re-order accordingly.

## Root cause

A three-state sort cycle is being written into a two-state URL parameter that cannot represent "unsorted".

Astryx cycles the header like this (`@astryxdesign/core`, `Table/plugins/sortable/useTableSortable.js`, `getNextDirection`), and `BAITable` opts into the unsorted step with `allowUnsortedState: true`:

```
null → ascending → descending → null
```

`DeploymentAutoScalingCard.tsx:157` declares the sort param with a non-nullable default:

```
order: parseAsStringLiteral(["createdAt", "-createdAt"] as const)
  .withDefault("-createdAt"),
```

and `DeploymentAutoScalingCard.tsx:344` maps the unsorted emission to `null`:

```
onChangeOrder={(order) => {
  setQueryParams({ order: order ? (order as "createdAt" | "-createdAt") : null });
}}
```

Writing `null` to a nuqs parser that carries `.withDefault(...)` clears the search param, so the very next read returns the default again. The sequence is therefore:

1. Initial state is `-createdAt` → **descending**.
1. Click → `getNextDirection("descending", true)` returns `null` → `BAITable` emits `onChangeOrder(undefined)`.
1. The handler writes `order: null` → param cleared → reads back as `-createdAt` → **descending**.

The state machine never leaves descending, so the ascending step is unreachable and the click has no visible effect.

## Fix

Make the parameter nullable so the unsorted state round-trips, mirroring `AdminDeploymentPage.tsx`, where the tab order param is nullable and the initial descending sort is applied only as a load-time seed (`params.order ?? DEFAULT_ORDER_BY_TAB[tab]`):

- drop `.withDefault("-createdAt")` from the `order` parser,
- seed the first query with `-createdAt` when the param is absent,
- build `orderBy` as `undefined` when `order` is `null` instead of always emitting a hardcoded `CREATED_AT` entry (`DeploymentAutoScalingCard.tsx:181`).

## Notes

- Distinct from FR-3710 (Admin → Deployments → Deployment Presets), which fails because `order` is never passed down at all. This card does pass `order` — the defect is the non-nullable default swallowing the unsorted state.
- Any other table that pairs `onChangeOrder` with a `withDefault` order param has the same trap.

## Affected files

- `react/src/components/DeploymentAutoScalingCard.tsx`
- `react/src/components/AutoScalingRuleListNodes.tsx` (sortable column definition, reference only)

JIRA Issue: FR-3711

agatha197:
Resolves #9131 (FR-3711)

## Problem

On the deployment detail page → **Auto-Scaling** card, the **Created At** header was frozen on descending. Every click was a visual no-op and ascending was unreachable.

A three-state sort cycle was being written into a two-state URL parameter. `BAITable` hardcodes `allowUnsortedState: true`, so Astryx cycles the header `null → ascending → descending → null` and emits `onChangeOrder(undefined)` on the unsorted step. But the param was declared with a non-nullable default:

```ts
order: parseAsStringLiteral(['createdAt', '-createdAt'] as const)
  .withDefault('-createdAt'),
```

The handler mapped the unsorted emission to `order: null`, which clears the search param — and a `withDefault` parser reads a cleared key back as its default. So the state machine could never leave descending:

1. Initial state `-createdAt` → descending.
2. Click → `getNextDirection('descending', true)` → `null` → `onChangeOrder(undefined)`.
3. Handler writes `null` → param cleared → reads back `-createdAt` → descending again.

## Fix

Move the card's sort, filter, and pagination **off the URL and into component state**.

The Auto-Scaling card is one section among several on the deployment detail page, not a page in its own right, so its table state is not page-level navigation and does not belong in the search params. That was the review call on this PR, and it dissolves the bug instead of working around it: plain `useState` with a nullable order represents the unsorted step directly, so the `withDefault` trap has nothing to trap.

- `useQueryStates({ order, filter })` → two `useState` hooks, `order` typed `'createdAt' | '-createdAt' | null` and seeded with `'-createdAt'`, so the header arrow matches the rows on first load;
- `useBAIPaginationOptionStateOnSearchParam` → `useBAIPaginationOptionState`. Both return the same `BAIPaginationOptionState`, so no call site changed;
- `order={orderString}` passes the real `null` down, so the header arrow clears on the unsorted step (`BAITable`'s `order` prop is already `string | null`);
- `orderBy` is built with the shared `convertToOrderBy<AutoScalingRuleOrderBy>(orderString)` helper (the one `DeploymentReplicasCard` already uses) instead of a hand-rolled `{ field: 'CREATED_AT', direction }` literal. It maps `null` to `undefined`, so the query carries **no `orderBy` while the header is unsorted** — "unsorted" means unsorted, not a silently re-applied default;
- the `nuqs` import is gone from this module.

### Resulting header cycle

| Click | `order` state | Header | Rows |
|---|---|---|---|
| (load) | `-createdAt` | descending | newest-first |
| 1 | `null` | no arrow | server default order (no `orderBy`) |
| 2 | `createdAt` | ascending | oldest-first |
| 3 | `-createdAt` | descending | newest-first |

Every click changes the header state, and the header always reflects what the rows are actually sorted by. Ascending takes two clicks (through the unsorted step) — the cost of starting the cycle on the descending default rather than on an arrow-less newest-first seed.

## Scope

`react/src/components/DeploymentAutoScalingCard.tsx` only (17 insertions, 30 deletions).

- `AutoScalingRuleListNodes.tsx` is purely presentational — it spreads `order` / `onChangeOrder` onto `BAITable` and needed no change.
- `DeploymentDetailPage.tsx` holds no URL state, and nothing else on the page reads the `order` / `filter` / `current` / `pageSize` keys this card was writing, so removing them has no other consumer.
- The sibling `DeploymentReplicasCard` keeps its URL state, namespaced under an `r` prefix (`rOrder`, `rCurrent`, `rPageSize`, `rFilter`, `rStatusCategory`). Applying the same card-local treatment there is left to a follow-up — it carries five pieces of state, and the tab-like `rStatusCategory` may be worth keeping in the URL, so it needs its own decision.

Distinct from FR-3710 (Admin → Deployments → Deployment Presets), which fails because `order` is never passed down at all.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup paths, StyleX, Astryx theme build, Astryx integration, z-index mirrors, Terminology).

Not verified in a live browser — reproducing needs a deployment carrying auto-scaling rules. A reviewer with suitable data should confirm the header cycle above.

### DIFF
--- a/react/src/components/DeploymentAutoScalingCard.tsx
+++ b/react/src/components/DeploymentAutoScalingCard.tsx
@@ -5,11 +5,13 @@
 import { DeploymentAutoScalingCardDeleteMutation } from '../__generated__/DeploymentAutoScalingCardDeleteMutation.graphql';
 import {
   AutoScalingRuleFilter,
+  AutoScalingRuleOrderBy,
   DeploymentAutoScalingCardListQuery,
 } from '../__generated__/DeploymentAutoScalingCardListQuery.graphql';
 import { DeploymentAutoScalingCardPresetsQuery } from '../__generated__/DeploymentAutoScalingCardPresetsQuery.graphql';
 import { DeploymentAutoScalingCard_deployment$key } from '../__generated__/DeploymentAutoScalingCard_deployment.graphql';
 import { App } from '../app-shim';
+import { convertToOrderBy } from '../helper';
 import { useCurrentUserInfo } from '../hooks/backendai';
 import { useBAIPaginationOptionState } from '../hooks/reactPaginationQueryOptions';
 import { useBAISettingUserState } from '../hooks/useBAISetting';
@@ -147,11 +149,9 @@ const DeploymentAutoScalingCardContent: React.FC<
 
   // Card-local state, not URL state: this card is one section among several on
   // the detail page, so its sort/filter/page are not page-level navigation.
-  // BAITable order string: "createdAt" (ASC) | "-createdAt" (DESC) | null
-  // (unsorted).
   const [orderString, setOrderString] = useState<
     'createdAt' | '-createdAt' | null
-  >(null);
+  >('-createdAt');
   const [graphQLFilter, setGraphQLFilter] = useState<
     AutoScalingRuleFilter | undefined
   >(undefined);
@@ -162,21 +162,11 @@ const DeploymentAutoScalingCardContent: React.FC<
     setTablePaginationOption,
   } = useBAIPaginationOptionState({ current: 1, pageSize: 10 });
 
-  // Newest-first seeds the list while the sort is unset — on first load, and
-  // again once the header cycles back to unsorted.
-  const effectiveOrder = orderString || '-createdAt';
-
   const queryVariables = {
     deploymentId,
     offset: baiPaginationOption.offset,
     limit: baiPaginationOption.limit,
-    orderBy: [
-      {
-        field: 'CREATED_AT' as const,
-        direction: (effectiveOrder.startsWith('-') ? 'DESC' : 'ASC') as
-          'ASC' | 'DESC',
-      },
-    ],
+    orderBy: convertToOrderBy<AutoScalingRuleOrderBy>(orderString),
     filter: graphQLFilter ?? null,
   };
   const deferredQueryVariables = useDeferredValue(queryVariables);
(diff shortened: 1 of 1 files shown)


nowgnuesLee (approved):
LGTM

PR #9170, opened 2026-08-27
fix(FR-3722): derive row-action `disabled` from `disabledReason`

### LINKED ISSUE #9169: Row actions can render disabled with no reason: make BAINameActionCell derive `disabled` from `disabledReason`
`BAINameActionCellAction` takes `disabled?: boolean` and `disabledReason?: string` as two independent optional fields, so every call site has to keep them in sync by hand. Three call sites currently do not, and render a disabled row action with **no explanation** — the same defect FR-3679 fixed on the Access Tokens card.

## Drifted call sites

| Location | Disabled with no reason when |
|---|---|

|—|—|
|---|---|

| `react/src/components/DeploymentRevisionHistoryTab.tsx:392` | deployment is stopped, or a rollback is in flight — `isDeployDisabled` adds `isDeploymentInStoppedCategory(...)` and `rollingBackRevisionId === record.id`, which `deployDisabledReason` does not cover |
|---|---|

| `react/src/components/VFolderNodes.tsx:238` (restore) | `vfolder?.status !== 'delete-pending'` |
|---|---|

| `react/src/components/VFolderNodesV2.tsx:236` (restore) | `vfolder?.vfolderStatus !== 'DELETE_PENDING'` |
|---|---|

The safe shape already exists in the same files — `start-service` in both `VFolderNodes` and `VFolderNodesV2` writes `disabled: !!noDeployTooltip, disabledReason: noDeployTooltip`, so the reason is the single source of truth and cannot drift.

## Proposed fix

Default `disabled` to `!!disabledReason` inside `BAINameActionCell` so the reason alone is enough, and drop the redundant boolean at call sites. This is **non-breaking**: both render paths already guard on `action.disabled` (`title={action.disabled ? action.disabledReason : action.title`} at :520, and the more-menu label at :352), so existing call sites that pass both keep working unchanged.

Then fix the three drifted sites so every disabled path names its reason.

## Scope

- `packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx` (+ doc/stories)
- the three call sites above

Follow-up to FR-3679 (PR #9166), which fixed the same defect class on the deployment Access Tokens card.

JIRA Issue: FR-3722

agatha197:
Resolves #9169 (FR-3722)

Follow-up to #9166 (FR-3679), which fixed "disabled button with no reason" on the
deployment Access Tokens card. This fixes the same defect class at its source.

## The problem

`BAINameActionCellAction` took `disabled?: boolean` and `disabledReason?: string`
as two independent optional fields, so every call site had to keep them in sync by
hand. Five did not, and shipped a disabled row action with **no explanation at
all**:

| Call site | Disabled with no reason when |
|---|---|
| `DeploymentRevisionHistoryTab.tsx` — Apply | deployment is stopped, or a rollback is in flight. `isDeployDisabled` added `isDeploymentInStoppedCategory(...)` and `rollingBackRevisionId === record.id`; `deployDisabledReason` covered only `isCurrent \|\| isDeploying`. |
| `VFolderNodes.tsx` — Restore | `vfolder?.status !== 'delete-pending'` |
| `VFolderNodesV2.tsx` — Restore | `vfolder?.vfolderStatus !== 'DELETE_PENDING'` |
| `VFolderNodes.tsx` — Delete forever | `status !== 'delete-pending'` (i.e. `delete-ongoing` / `-complete` / `-error`); the reason covered only `isProjectFolderManagedElsewhere` |
| `VFolderNodesV2.tsx` — Delete forever | **every** disabled case — the action carried no `disabledReason` whatsoever |

The safe shape was **already in the tree** — `start-service` in both `VFolderNodes`
and `VFolderNodesV2` writes `disabled: !!noDeployTooltip, disabledReason: noDeployTooltip`,
so the reason is the single source of truth and cannot drift. It was the exception;
this makes it the default.

## The change

```ts
disabled?: boolean | { reason: string };
```

The reason lives **inside** the flag that disables, so "a reason but enabled" is
unrepresentable and the two cannot disagree. A bare `true` still disables without
a reason — now a deliberate choice rather than two fields that drifted apart.
`FileNameCell`'s download and delete, gated on plain capability booleans, are the
legitimate case for it.

`disabledReason` is **removed**, not deprecated. `tsc` names every call site, and
keeping both fields alive would recreate exactly the drift this PR is about.

A `disabledReason(disabled)` accessor is exported alongside it, for callers that
thread a `disabled` value through their own props and have to read the reason
back out somewhere — #9166 above this PR does exactly that on a `BAIButton`,
whose `disabled` is a plain boolean beside a `title` tooltip. Without the export
each such caller re-implements the narrowing inline, which is how the union
starts leaking back into two values.

> An earlier revision of this PR instead defaulted `disabled` to
> `!!disabledReason`. That left the invariant a *convention*: `disabled: false`
> beside a reason stayed representable, and a call site passing only a reason read
> as if it had not asked for anything to be disabled. Review moved it into the
> type.

One behavior improvement rode along: the visible button's tooltip was
`action.disabled ? action.disabledReason : action.title`, so an action disabled
*without* a reason lost its tooltip entirely. It now falls back to the action
title.

## Call-site fixes

Each of the five now states `disabled: { reason }` on every branch that blocks it,
with `false` as the fall-through — so no branch can go silent by omission.

Both `delete` (move-to-trash) actions had an **unconditional** `disabledReason`
beside a narrower boolean — the two disagreed, and the reason shown was whichever
string the ternary happened to reach. Folding them into one field forced the
branches to line up. `VFolderNodesV2`'s no-permission branch was dead
code (its `disabled` was only ever `isPipelineFolder`, so the string never rendered)
and is removed rather than preserved; the `TODO(needs-backend)` note above it is
trimmed to the constraint it records.

BUI's own `FileExplorer/FileNameCell` Edit action had the identical shape —
`disabled: isEditDisabled` beside a `disabledReason` ternary that always yields a
string — and gets the same treatment. It was the last live call site sitting one
deletion away from a permanently disabled button.

## i18n

Three keys added and one removed, across all 21 locales:

- `deployment.ApplyDisabledDeploymentStopped`
- `deployment.ApplyDisabledWhileRollingBack`
- `data.folders.DeletionAlreadyStarted`
- ~~`data.folders.CannotRestoreNotInTrash`~~

The removed key ("Only folders in the trash bin can be restored") would have been
**unreachable as written**. Restore renders only when `isDeletedCategory(status)`
holds, and that covers `delete-ongoing`, `delete-complete` and `delete-error` as
well as `delete-pending` — so in every case the message could appear, the folder
*is* in the trash; its deletion has merely passed the restorable point.
`DeletionAlreadyStarted` ("Deletion has already started for this folder.") is true
for all three, and is shared with the two delete-forever actions.

## Scope note

The gating *rules* themselves (`isDeploymentInStoppedCategory`, `status !== 'delete-pending'`,
`isProjectFolderManagedElsewhere`) stay frontend-side and are unchanged here — this
PR is about how a blocked action explains itself, not about who decides. Moving the
decision server-side would need an entity-level action-permission field on `VFolder`
and the deployment types, of the kind `Agent` / `Session` / `ImageV2` already expose
as `permissions: [XPermission!]!`. `VFolder.accessControl.permission` is mount-level
(RO/RW/RW_DELETE) and cannot gate these buttons. Worth a backend FR; out of scope
for this one.

## Verification

- `pnpm --filter backend.ai-ui build` → built
- `pnpm --prefix ./react exec vitest run` → **1732/1732 pass** (110 files)
- New `BAINameActionCell.test.tsx` → **6/6**, and load-bearing: stubbing the
  reason accessor to `undefined` fails the tooltip and overflow-label cases.
- `VFolderNodes.test.tsx` → **5/5** (two new cases cover the trash-bin pair
  across `delete-pending` and `delete-ongoing`)
- `FileExplorer` suite → **12/12**

The new test covers both render paths — the visible button (with `clientWidth`
stubbed, since jsdom reports 0 and would collapse every action into the overflow
menu) and the overflow-menu row.

## Reviewing order

This is the bottom of the stack — the API change every layer above depends on.
#9166 (FR-3679) sits directly on it and is the first consumer: its Access Tokens
row action writes `disabled: { reason }` against the type introduced here. Read
this PR first; it is the smaller and more load-bearing of the two.


packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx
```diff
@@ -36,12 +36,15 @@ export interface BAINameActionCellAction {
...
+   * Whether the action is disabled. Defaults to `!!disabledReason`, so naming
+   * a reason is enough — pass this only to disable without one.
+   */
   disabled?: boolean;
   /** Loading spinner for progress this cell does not own (e.g. a background
    * delete tracked by the parent). Use `action` when the click itself awaits. */
   loading?: boolean;
-  /** Tooltip text when disabled */
+  /** Why the action is disabled. Becomes the button tooltip. */
   disabledReason?: string;
```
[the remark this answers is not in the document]
agatha197: Fixed in 3dd6bbd41. Both spots in `BAINameActionCell.doc.ts` now describe the reason as what disables: the guidance entry reads "Name \`disabledReason\` and the action disables itself … Pass \`disabled\` only to disable without a reason, or \`disabled: false\` to keep an action enabled despite one", and the `actions` prop description says "\`disabledReason\` (which alone disables) with an optional \`disabled\` override".

react/src/components/VFolderNodes.tsx
```diff
@@ -233,16 +233,14 @@ const VFolderNameCell: React.FC<VFolderNameCellProps> = ({
...
-            isProjectFolderManagedElsewhere,
           disabledReason: isPipelineFolder
             ? t('data.folders.CannotRestorePipelineFolder')
             : isProjectFolderManagedElsewhere
               ? (projectFolderAdminHint ??
                 t('data.folders.NoRestorePermission'))
-              : undefined,
+              : vfolder?.status !== 'delete-pending'
+                ? t('data.folders.DeletionAlreadyStarted')
+                : undefined,
```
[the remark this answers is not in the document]
agatha197: Fixed in 3dd6bbd41. `renderTable` now takes vfolder overrides, and a new describe block covers the trash-bin pair: `delete-pending` leaves Restore and Delete enabled, `delete-ongoing` disables both with `data.folders.DeletionAlreadyStarted` on each. 5/5 pass.

react/src/components/DeploymentRevisionHistoryTab.tsx
```diff
@@ -389,13 +389,16 @@ const DeploymentRevisionHistoryTab: React.FC<
...
-          isCurrent ||
-          isDeploying ||
-          isDeploymentInStoppedCategory(deploymentStatus) ||
-          rollingBackRevisionId === record.id;
+          isCurrent || isDeploying
+            ? t('deployment.ApplyDisabled')
+            : isDeploymentInStoppedCategory(deploymentStatus)
+              ? t('deployment.ApplyDisabledDeploymentStopped')
+              : rollingBackRevisionId === record.id
+                ? t('deployment.ApplyDisabledWhileRollingBack')
```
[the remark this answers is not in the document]
agatha197: Confirmed and fixed in 68954df5f. `Apply` is the cell's only auto action, so a narrow row overflows it into the menu whose trigger `moreMenuDisabled` locks when stopped — the new reason really was unreachable there. Took the `minVisibleActions={1}` option: Apply stays a visible button, so its tooltip is always readable. The menu's own item keeps its blanket disable and the TODO tracking the move to a per-item reason; that case predates this PR and is not one of the five sites it fixes.

packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx
```diff
@@ -66,6 +69,11 @@ export interface BAINameActionCellAction {
   popConfirm?: BAIPopconfirmConfig;
 }
 
+const isActionDisabled = (action: BAINameActionCellAction) => !!action.disabled;
```
nowgnuesLee: This function adds more complexity than it removes — I'd just inline it.

packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx
```diff
@@ -66,6 +69,11 @@ export interface BAINameActionCellAction {
...
+
+const actionDisabledReason = (action: BAINameActionCellAction) =>
+  typeof action.disabled === 'object' ? action.disabled.reason : undefined;
```
nowgnuesLee:
```suggestion
const disabledReason = (disabled: boolean | { reason: string }) =>
  typeof disabled === 'object' ? disabled.reason : undefined;
```

nowgnuesLee (changes requested):

packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx
agatha197: Agreed — once `disabled` became the union it was just a wrapper around `!!action.disabled`. Inlined at all three call sites in ca4863318.

packages/backend.ai-ui/src/components/Table/BAINameActionCell.tsx
agatha197: Applied in ca4863318 — it takes the `disabled` value now and is renamed to `disabledReason` to match. One deviation from the suggestion: the parameter is typed `BAINameActionCellAction['disabled']` rather than `boolean | { reason: string }`, because `disabled` is optional and every call site passes it directly, so the narrower type does not compile.

nowgnuesLee (approved):
LGTM

ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3722-pr9170-derive-row-action.sungchul.10-82-0-159.sslip.io)

PR #9183, opened 2026-08-27
style(FR-3728): render the preset Required field as a checkbox

### LINKED ISSUE #9180: Runtime variant preset modal: render the Required field as a checkbox, not a switch
The runtime variant preset create/edit modal (`BAIRuntimeVariantPresetSettingModal`) renders its **Required** field with `AstryxFormSwitch`.

A switch is for a setting that takes effect immediately; `Required` is a boolean attribute of the preset being authored, submitted with the rest of the form. Astryx's own guidance for `CheckboxInput` is the same: "Don't use a checkbox for actions that take effect immediately; use a toggle switch or button instead" — and the inverse holds here.

## Change

- Render the `required` form field with a checkbox instead of a switch.
- BUI has no checkbox form-control adapter yet (the host app has `AstryxFormCheckbox` in `react/src/components/astryxFormControls.tsx`); add the mirrored `AstryxFormCheckbox` to `packages/backend.ai-ui/src/components/astryxFormControls.tsx`.

No behavioral change — the submitted value and the `runtime-variant-preset-required` capability gate stay as they are.

## Where

- `packages/backend.ai-ui/src/components/fragments/BAIRuntimeVariantPresetSettingModal.tsx`
- `packages/backend.ai-ui/src/components/astryxFormControls.tsx`

Related: FR-3087 (introduced the Required field).

JIRA Issue: FR-3728

agatha197:
Resolves #9180 (FR-3728)

The runtime variant preset create/edit modal drove its `Required` field with `AstryxFormSwitch`. A switch reads as a setting that takes effect on flip; `required` is a boolean attribute of the preset being authored and is only persisted on submit — so a checkbox is the right control. Astryx's own `CheckboxInput` guidance draws the same line ("Don't use a checkbox for actions that take effect immediately; use a toggle switch or button instead").

## What changed

- `BAIRuntimeVariantPresetSettingModal` renders the `required` field with a checkbox.
- The field label and the control now say different things: the `Form.Item` label is a new `Requirement` key ("필수 여부" / "Requirement"), and the checkbox carries the visible `Required` label ("필수" / "Required") instead of hiding it. `Requirement` is added to all 21 BUI locales.
- The item is marked `required` **for display only** — a checkbox always carries a value, so `requiredMark="optional"` would otherwise render the label as "필수 여부 (선택)". No validation rule is added, and the rendered checkbox carries no `aria-required`.
- BUI had no checkbox form-control adapter, so `AstryxFormCheckbox` is added to `packages/backend.ai-ui/src/components/astryxFormControls.tsx`, mirroring the host adapter of the same name minus its `onValueChange` escape hatch.

The field keeps `valuePropName="checked"` and its tooltip, so the submitted value and the `runtime-variant-preset-required` capability gate are unchanged.

## Verification

- Live check on the dev server (Admin → Deployments → Runtime Variant Presets → Create Preset), en-US and ko-KR: the modal reports 1 checkbox / 0 switches, no `aria-required` on the control, no "(optional)" / "(선택)" suffix on the label, and clicking the box flips the bound form value `false → true`.

## Screenshots

| Before (switch) | After — English | After — 한국어 |
|---|---|---|
| before | after-en | after-ko |


yomybaby (approved):
LGTM

PR #9190, opened 2026-08-27
docs: update deployment docs (2026-08-27)

lablup-octodog:
## What changed

FR-3689 (`6a06823`, #9083) reworked the **Runtime Variant Presets** admin surface (`/admin/deployments`):

- The separate *Runtime Variant* / *Runtime Variant ID* table columns merged into one **Runtime Variant (ID)** column (falling back to the bare-ID title on servers without the runtime-variant field).
- A **UI Type** column was added directly after **Value Type**; **Default Value** is now shown by default; the capability gate was dropped from the read-only **Category** / **Display Name** columns.
- The substantive change: UI Type and Value Type are now constrained to compatible pairings (`UI_TYPE_TO_ALLOWED_VALUE_TYPES`), with new validation/warning messages, and the setting modal was reordered so **UI Type** sits directly under **Value Type**. A stored-mismatch preset degrades to a plain text input in the deployment form.

The manual described the old two-column table and the old, unconstrained modal, so `deployment/deployment_presets.md` was rewritten in all four languages (en/ko/ja/th): column list, hidden-columns sentence, capability `:::note`, create/edit field list, and a new `:::note` on the pairing constraint.

## Driving commits

- `6a06823` — fix(FR-3689): constrain the runtime variant preset uiType/valueType pairing (#9083) — the only commit driving this chapter's prose.
- `c16477c9e` (FR-3694, #9104) and `29fc723f3` (FR-3670, #9062) — drove the chapter's eight screenshot-only refreshes in `deployment/deployment.md`, which did not happen this run (see below).

## Review findings (independent review)

**No defects found.** Everything was spot-checked against the source at the range head:

- All pairing claims match `packages/backend.ai-ui/src/helper/runtimeVariantPresetUI.ts` exactly (Slider/Number Input: Integer/Float; Checkbox: Boolean/Flag; Select: String; Text Input: String/Integer/Float).
- Table claims match `BAIRuntimeVariantPresetTable.tsx`: merged column with copyable ID in parentheses and the `RuntimeVariantId` fallback title, UI Type shown-by-default with verbatim display of unrecognized controls and `-` when none, Default Value no longer `defaultHidden`, Category/Display Name gates removed on the read side.
- Modal claims match `BAIRuntimeVariantPresetSettingModal.tsx`: field order (Value Type → UI Type → Key → Default Value), incompatible options disabled, `onValuesChange` unconditionally clearing `uiType` when Value Type changes, the stored-mismatch warning banner shown regardless of server version, and the older-server reverse constraint (`isValueTypeConstrained` restricting Value Type by the stored control).
- The deployment-form fallback sentence matches `RuntimeParameterFormSection.tsx`'s `effectiveUIType` degrade.
- The writer's **out-of-plan modal-field reorder is correct**: the plan's field order predated checking the modal source; the doc's claim to list fields "in the order they appear" made the reorder necessary, not scope creep.
- **Translation parity:** all four languages carry the same six edits with matching structure. Every quoted UI string was checked verbatim against the BUI locale files (`RuntimeVariantWithID`, `UIType*`, `UITypeIncompatibleWithValueType`, `StoredValueTypeMismatch`) — ko correctly uses the live label "런타임 (ID)" rather than a literal translation.
- No terminology drift (no `terminology.json` entry covers these terms; the docs quote the live i18n labels, which take precedence).
- Nit, not a blocker: the EN mismatch note slightly compresses when the field-level error appears (the warning banner shows on open; the field marker appears on validation, at latest on save). Not misleading.

## Screenshots — list captured ✅, modal still owed ⚠️

`runtime_variant_preset_list.png` was **recaptured in all four languages** (2026-08-28) against `http://10.82.130.172:8090`; the original capture host (`http://10.82.1.75:8090`) is still unreachable.

- The new image shows the **UI Type** and **Default Value** columns the prose describes — the image on `main` predates FR-3689 and has neither. UI Type renders real values (Checkbox / Select / Slider).
- Framing matches the previous image's scope (the card only, tab rail through the pagination row) at the chapter's 2x density, per `SCREENSHOT-GUIDELINES.md`; height is unchanged at 1114px, width grew to fit the two added columns without the table scrolling.
- **One caveat, recorded in a narrowed TODO:** the capture server runs manager `26.8.0rc1`, which does not serve the runtime-variant field, so the column appears in its documented bare-ID fallback form (**Runtime Variant ID**) rather than the qualified **Runtime Variant (ID)** form. Recapture on a server that serves the field to show the qualified form.
- **The rows show realistic sample data, not the live list.** The shared test server's preset list is full of leftover `e2e-*` fixtures and a row literally named `test`, which reads as a test-server capture in a manual. The row text (Name, Preset Target, Value Type, UI Type, Key, Default Value) is substituted at capture time with real vLLM engine parameters, per `SCREENSHOT-GUIDELINES.md`'s "realistic but non-sensitive sample data". Nothing was written to the shared server, so no other session's fixtures were touched. The localized cells use the verbatim `comp:BAIRuntimeVariantPresetTable` strings from each locale file, so all four languages are correct; every Value Type and UI Type the chapter lists appears at least once and each pairing is legal per `UI_TYPE_TO_ALLOWED_VALUE_TYPES`.
- The **Create Preset modal** shot remains blocked. Its UI Type fields require manager **26.9.0**; every reachable server was probed and runs `26.8.0rc1` or `26.8.1` (`dogbowl.cloud.backend.ai`). Its `TODO(screenshot)` marker was refreshed with that finding rather than removed.
- The eight `deployment/deployment.md` refresh-only images (FR-3694/FR-3670 restyles) are still deferred; existing images verified present and untouched. No prose depends on them.

- `deployment/deployment.md` prose — its plan entry was screenshot-refresh-only; correctly left unedited.
- FR-3689's `astryxFormControls.tsx` / parity-test hunks — internal to the form engine, no user-visible effect beyond what is documented.

## Verified

- `bash scripts/verify.sh` passed (`=== ALL PASS ===`).
- The diff touches only `packages/backend.ai-webui-docs/src/{en,ko,ja,th}/deployment/deployment_presets.md` and the four `images/runtime_variant_preset_list.png` files.
- `pnpm -w run lint:terminology`: **0 blocking findings** (remaining warn/report items are pre-existing i18n noise, untouched by this PR).
- Prettier reports style issues on these four `.md` files both before and after the change — pre-existing, not a regression; the docs package is outside every `lint-staged` glob.

## Follow-up in this revision

Per review feedback, the `:::note` stating that **Category** / **Display Name** are listed regardless of server version (and that writing them, plus **UI Type**, needs a newer server) was **removed in all four languages**. The column bullet still documents the bare-ID fallback, which is what the new screenshot shows.

> **Note:** This PR carries no `FR-` number by design (nightly docs sync), so it deviates from the repo's `prefix(FR-XXXX)` title convention and will not be linked by `project-status-sync`.

---


agatha197 (approved):
LGTM

PR #9209, opened 2026-08-28
fix(FR-3106): neutralize CSV formula injection in exported cells

### LINKED ISSUE #7842: Neutralize CSV formula injection in CSV export helpers
## Context

Found during review of PR #7837 (FR-3098), pre-existing gap (not introduced by that PR). None of the CSV builders neutralize cells starting with `=`, `+`, `-`, `@`, tab, or CR:

- `escapeCsvValue` in `react/src/helper/csv-util.ts`
- DiagnosticsPage's local `escCsv` (`react/src/pages/DiagnosticsPage.tsx`)
- keypair row builder in `react/src/components/MyKeypairManagementModal.tsx`

User-controlled data (emails, names, backend export rows) flows into files explicitly intended for Excel, so a crafted value like `=HYPERLINK(...)` executes as a formula on open (CSV/formula injection).

## Fix

Prefix dangerous leading characters with `'` (single quote) or a space in `escapeCsvValue` and route all CSV cell escaping through it (the central `downloadCSV` path added by PR #7837 makes this a one-place change). Add Vitest cases.

JIRA Issue: FR-3106

eastagiletracker:
This PR proposes neutralizing CSV formula injection in the cells written by `escapeCsvValue`, so an exported value that begins with `=`, `+`, `-`, `@`, tab or CR is no longer evaluated as a formula when the download is opened in a spreadsheet (fixes #7842). We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/470. You can sign in with your GitHub ID to claim ownership of the project.

## What is wrong

`escapeCsvValue` in `react/src/helper/csv-util.ts` wraps a cell in quotes and doubles the quotes inside it, but the leading character reaches the file untouched. Excel, LibreOffice and Sheets treat a leading `=` as the start of a formula, and Excel keeps Lotus compatibility so `+`, `-` and `@` open one too; a leading tab or CR hides one of those behind whitespace the importer strips. Wrapping the field in quotes does not prevent any of it. `react/src/pages/DiagnosticsPage.tsx` and `react/src/components/MyKeypairManagementModal.tsx` each carry their own local escaper with exactly the same gap, which is the set of three sites #7842 lists.

Reproduced on `main` at `f376cc1`, with no changes applied:

```
cd react
node --experimental-strip-types --input-type=module \
  -e "import { JSONToCSVBody } from './src/helper/csv-util.ts'; console.log(JSONToCSVBody([{ name: '=1+2' }]));"
```

```
"name"
"=1+2"
```

The value is written through verbatim, so a spreadsheet evaluates it on open. Emails, user and project names, and resource-policy rows all reach these exports, and the four `exportCSVWithFormattingRules` callers (`GeneratedKeypairListModal`, `KeypairResourcePolicyList`, `UserResourcePolicyList`, `ProjectResourcePolicyList`) plus the two local builders are the download buttons users press.

## The change

`neutralizeCsvFormula` is added to `csv-util.ts` and applied inside `escapeCsvValue`, prefixing an affected cell with a single quote. `escapeCsvValue` is now exported, and the local escaper in `DiagnosticsPage` and the row builder in `MyKeypairManagementModal` are routed through it, so the rule lives in one place as #7842 suggests.

Two behavioural details worth your attention, since both are choices rather than mechanics:

- A blanket prefix would also rewrite the `-` placeholder your `format_rules` emit for empty values in the three resource-policy exports, and every negative number. Neither can carry an expression, so a value that is exactly `-` or a plain signed number is left alone. Both cases are covered by tests, so the behaviour is pinned rather than incidental.
- `DiagnosticsPage` used to quote a cell only when it contained a comma, quote or newline. Going through the shared helper means every cell is quoted. That is still RFC 4180 and parses identically, but it does change the shape of that one file, so it should be a deliberate call rather than a surprise.

## Verification

Everything below was run in `react/`, before and after the change, on this branch.

- `pnpm exec vitest run`: 102 files / 1631 tests passing on the untouched tree, 102 files / 1650 passing after. The 19 extra are the new cases; no previously passing test changed state.
- `pnpm exec eslint ./src --max-warnings=0 --no-warn-ignored` and `pnpm exec tsc --noEmit`: clean both before and after. `prettier --check` passes on all four files.
- The new tests genuinely fail without the fix. Reverting only the `neutralizeCsvFormula(...)` call inside `escapeCsvValue` and re-running `pnpm exec vitest run src/helper/csv-util.test.ts` gives `8 failed | 32 passed` — the 8 are exactly the neutralization assertions, while the `-` placeholder and negative-number controls stay green.
- Neither `DiagnosticsPage` nor `MyKeypairManagementModal` has a test file, so the suite does not actually cover those two edits. We ran their exact row-building expressions against the real helper instead:

```
DiagnosticsPage:
"ID","Severity","Category","Title","Description","Remediation"
"csp-1","warning","CSP","Header missing","'=1+2",""
"ep-2","info","Endpoint","Reachable, fast","Latency 20ms","-"

MyKeypairManagementModal:
"AKIAEXAMPLE","'=1+2","ssh-rsa AAAAB3"
```

One merge note: #7941 is open against `react/src/helper/csv-util.test.ts` for `parseCSV` / `exportCSVWithFormattingRules` coverage. It adds cases in a different part of that file and does not touch the helper itself, so the two changes are compatible, but whichever lands second will want a rebase.

## How this was managed

We imported this repository's issues and pull requests into a board — 6,889 stories, with your milestones as epics — and ran this change on it as the story [Neutralize CSV formula injection in CSV export helpers](https://eastagiletracker.com/projects/470/stories/427396), which is the imported copy of #7842. The board is live at https://eastagiletracker.com/projects/470

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

### DIFF
--- a/react/src/components/MyKeypairManagementModal.tsx
+++ b/react/src/components/MyKeypairManagementModal.tsx
@@ -13,7 +13,7 @@ import { MyKeypairManagementModalRevokeMyKeypairMutation } from '../__generated_
 import { MyKeypairManagementModalSwitchMainKeyMutation } from '../__generated__/MyKeypairManagementModalSwitchMainKeyMutation.graphql';
 import { App } from '../app-shim';
 import { convertToOrderBy } from '../helper';
-import { downloadCSV } from '../helper/csv-util';
+import { downloadCSV, escapeCsvValue } from '../helper/csv-util';
 import { useBAIPaginationOptionState } from '../hooks/reactPaginationQueryOptions';
 import { useBAISettingUserState } from '../hooks/useBAISetting';
 import { theme } from '../theme-shim';
@@ -90,7 +90,7 @@ const downloadCredentialCSV = (credential: KeypairCredential) => {
     credential.secretKey,
     credential.sshPublicKey,
   ]
-    .map((v) => `"${v.replace(/"/g, '""')}"`)
+    .map((value) => escapeCsvValue(value))
     .join(',');
 
   const csvContent = `${header}\n${row}\n`;
--- a/react/src/helper/csv-util.test.ts
+++ b/react/src/helper/csv-util.test.ts
@@ -3,7 +3,13 @@
  Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
  */
 // csv-util.test.ts
-import { downloadCSV, JSONToCSVBody, parseCSV, UTF8_BOM } from './csv-util';
+import {
+  downloadCSV,
+  escapeCsvValue,
+  JSONToCSVBody,
+  parseCSV,
+  UTF8_BOM,
+} from './csv-util';
 
 describe('JSONToCSVBody', () => {
   it('should convert JSON data to CSV format without formatting rules', () => {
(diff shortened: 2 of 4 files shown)


nowgnuesLee:
Thanks for your interest in this repository. Unfortunately, we have to close this PR — it appears to be part of a pattern of automated, high-volume submissions across unrelated repositories, which we don't accept.

PR #9289, opened 2026-08-30
chore(FR-3796): remove dead client-era leftovers under src/lib

nowgnuesLee:
Resolves #9288 (FR-3796)

## Problem

The Backend.AI client moved to `packages/backend.ai-client/` in FR-2529, but `src/lib/` still carried client-era files nothing imports any more.

## What changed

Deleted (zero references across `react/`, `packages/`, `electron-app/`, `Makefile`, `scripts/`):

- `src/lib/json_to_csv.js`, `src/lib/json_to_csv.ts`
- `src/lib/TabCounter.js` — the live port is `react/src/lib/TabCounter.ts`

Dropped the `.gitignore` entries for `src/lib/` artifacts that no longer exist (`backend.ai-client-ingen-es6.js`, `bundles/`, `web-animations-js/`, `webcomponents-loader.js`). The `src/lib/backend.ai-client-node.js` entry stays — it is a real build artifact.

## What deliberately stays

`src/lib/backend.ai-client-node.ts` is the Electron local-proxy dependency: `src/wsproxy/manager.js` instantiates `ai.backend.Client` / `ClientConfig` from it, `src/wsproxy/lib/backend.ai-ws-appproxy.js` subclasses `Client`, and `make compile_client_node_ts` → `compile_wsproxy` compiles it into `src/wsproxy/dist/wsproxy.js`. The `backend.ai-client` package is ESM-only and browser-bound (FR-3754), so moving the wsproxy onto it is separate work, not cleanup. `src/.prettierrc.json` also stays — it governs `src/wsproxy` formatting.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===`
- `make compile_wsproxy` (Electron local-proxy build) → exit 0; only the pre-existing `express/lib/view.js` critical-dependency webpack warning
- `pnpm exec vitest run src/wsproxy/manager.test.ts` → 18 passed
- `git status` clean after the build — both artifacts (`src/lib/backend.ai-client-node.js`, `src/wsproxy/dist/`) remain ignored

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9290, opened 2026-08-30
refactor(FR-3796): move the Electron local proxy onto the backend.ai-client package

### LINKED ISSUE #9288: Remove leftover client sources under src/ superseded by the backend.ai-client package
The Backend.AI client library now lives in the workspace package `packages/backend.ai-client/`, but remnant client source files still remain under the legacy `src/` tree.

## Task

- Audit the client-related files left under `src/` and identify which ones are still consumed by the Electron app build (some of the legacy client code is believed to be used there).
- Keep only what the Electron build genuinely depends on; delete the rest so `packages/backend.ai-client/` is the single home for the client.
- Verify the web build and the Electron build (`make dep` / electron packaging) still succeed after the removal.

## Notes

- Do not remove anything the Electron build path still imports — confirm via the build scripts before deleting.

—

Captured while working on branch: FR-3703

JIRA Issue: FR-3796

nowgnuesLee:
Resolves #9288 (FR-3796)

Second half of FR-3796, stacked on #9289. That PR removed the dead `src/lib` leftovers; this one removes the last one — the 5,147-line `src/lib/backend.ai-client-node.ts` — by moving its only consumer, the Electron local proxy, onto the `backend.ai-client` workspace package. `src/lib/` no longer exists.

## What changed (three commits, reviewable bottom-up)

1. **`backend.ai-client` becomes constructible and requireable from Node.** The `Client` constructor read `localStorage` unconditionally; every `localStorage` access in `client.ts` (14 sites) now goes through a 13-line `safeStorage` helper that no-ops without the global, keeping each read site's existing fallback and still letting a real `QuotaExceededError` reach the quota-recovery catch. tsup now emits ESM + CJS with `crypto-es` inlined (it is ESM-only, so an external `require('crypto-es')` would fail on the node18 `pkg` target); `exports` gains a `require` condition.
2. **wsproxy requires the package.** Root `devDependencies` gains `backend.ai-client: workspace:*`; `manager.js` (still lazily, so the unit test imports without a build) and `backend.ai-ws-appproxy.js` require `backend.ai-client`; `console-appproxy.js` drops a require it never used; `compile_wsproxy` now depends on `compile_client_pkg` (`pnpm --filter backend.ai-client build`). Also sets `this._env = env` in the ws-appproxy constructor — it was never assigned (console-appproxy does it), so `this._env.ext_proxy_url` on the TCP path dereferenced `undefined`. Pre-existing bug, found while checking parity.
3. **Delete the legacy file** plus its Makefile targets, `.gitignore` / `.prettierignore` / `eslint` / `vitest` / `labeler` entries, and re-point the live docs (`CSP.md`, the theme-family spec, the client package's type-mismatch notes) at the package.

## Behaviour deltas, all deliberate

- HMAC signing is **byte-identical** between the two clients (`sign` / `getSignKey` / `getAuthenticationString` diffed and probed with the same key, date and path). The one signed input that changes is the declared API version **constant**: `X-BackendAI-Version` never tracked the `api_version` the web UI sends to `/conf` (the `APIMajorVersion` setter mirrors only `_apiVersionMajor`, in the old client too — it drives the `kernel/` vs `session/` path prefix) — the header was always the client's `_apiVersion` constant, `v4.20190615` before, `v8.20240915` now. The proxy signs and sends the same header, so validation holds either way; a pre-v8 manager receiving v8-signed requests is the one new combination, and that sits below the supported LTS floor (26.4).
- The standalone local-proxy `pkg` target moves node18 → node22, aligning the shipped binary with the package's `engines >= 22`; the node22-macos-arm64 binary was built and answered the API-mode `PUT /conf` probe locally.
- Release-pipeline wiring (human review catch): desktop release jobs restore `src/wsproxy/dist/` from the web-build artifact, which used to skip the client package build entirely — `pkg` then packed a binary that only *warned* about the unresolvable `backend.ai-client` and exited 0. `dep_web` now guards on `packages/backend.ai-client/dist/index.cjs` and builds the package itself (once per job, ahead of the six parallel `compile_localproxy` runs, so tsup's `clean: true` cannot race). The failure mode and the fix were both reproduced locally.
- The package strips a trailing `/` from the endpoint before it reaches the sign key; the web UI's client does the same, so the proxy now matches it.
- `clientVersion` in `User-Agent` goes `20.11.0` → `22.09.0` (unsigned, unread).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` on every commit.
- `make compile_wsproxy` (Electron bundle) → green; `src/wsproxy/dist/wsproxy.js` contains the package's code and no `backend.ai-client-node` string.
- `pnpm exec vitest run src/wsproxy/manager.test.ts` → 18/18.
- Raw-source `node src/wsproxy/local_proxy.js` + `PUT /conf` (API mode) → 200 with a token, which is only produced after `new ClientConfig` + `new Client` succeed.
- **Standalone binary:** `pnpm exec pkg src/wsproxy/local_proxy.js --targets node18-macos-arm64` → zero resolve warnings; the binary answers `PUT /conf` → 200 when run from an empty directory with no `node_modules` on the lookup path, so the package is inside the snapshot.
- Node probes on both `dist/index.js` and `dist/index.cjs` with no shims produce identical `signKey` / signature.

## CodeQL follow-up (commit 5)

CodeQL flagged the new `safeStorage.setItem` as clear-text storage of passwords. Most of that is noise — request bodies are masked by `redactRequestParameters` before they reach storage, which CodeQL's taint tracking does not credit — but auditing its sources exposed one real, pre-existing gap: `update_password()` sends `new_password2`, and the exact-match `SENSITIVE_LOG_KEYS` list never contained that spelling, so the repeated new password was persisted unmasked. The redaction now masks any key containing a sensitive **stem** (`password`, `secret`, `token`, `private_key`, …, case-insensitive), so snake_case and camelCase spellings (`new_password2`, `refreshToken`, `clientSecret`) are all covered; regression tests pin the exact `update_password` body plus camelCase/nested variants.

Known limitation, deferred to [FR-3836](https://lablup.atlassian.net/browse/FR-3836): API mode does not propagate `EXT_HTTP_PROXY` to the TCP gateway (the gateway receives a `ClientConfig`, which has no `ext_proxy_url`). Before this PR that path crashed outright; the `TODO(FR-3836)` marks the assignment site.

**Not verified here:** a live API-mode app-proxy connection against a real manager. Every available test box runs the webserver in SESSION mode (where the proxy uses the cookie-based console gateway and never instantiates the client), and none exposes the manager's API port to my machine — a signed `GET /auth/test` through the webserver returns the webserver's own 401 regardless of the signature. Someone with an API-mode deployment or shell access to a manager host should run the proxy once with a keypair before this leaves draft.

[FR-3836]: https://lablup.atlassian.net/browse/FR-3836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/backend.ai-client/src/safe-storage.ts
```diff
@@ -0,0 +1,13 @@
+const hasStorage = () => typeof localStorage !== 'undefined';
+
+/** `localStorage` that no-ops where the global is absent (Node, the Electron local proxy). */
+export const safeStorage = {
+  getItem: (key: string): string | null =>
+    hasStorage() ? localStorage.getItem(key) : null,
+  setItem: (key: string, value: string): void => {
+    if (hasStorage()) localStorage.setItem(key, value);
```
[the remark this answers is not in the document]
nowgnuesLee: Triaged flow by flow. 7 of the 8 sources are cut before this sink: request bodies pass through `redactRequestParameters` (client.ts, applied right before the persist at the `_wrapWithPromise` log block), which masks any key containing "password" plus the secret/token key list — CodeQL does not model that custom sanitizer, so the flows appear live. The `login()` path additionally pre-masks at its call site via `opts.log`. The 8th source was a real pre-existing gap this alert helped surface: `update_password()` sends `new_password2`, which the old exact-match key list missed — fixed in 91cb114 by switching to substring matching, with a regression test (`redact.test.ts`) pinning the exact `update_password` body. Verified live: the persisted log shows masked values for both the login and the GraphQL password-change entries. Leaving the code-scanning alert itself for a human to dismiss.

packages/backend.ai-client/src/safe-storage.ts
```diff
@@ -0,0 +1,13 @@
...
+  },
+  removeItem: (key: string): void => {
+    if (hasStorage()) localStorage.removeItem(key);
```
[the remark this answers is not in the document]
nowgnuesLee: Added in 10007bd: `node-compat.test.ts` runs in vitest's node environment (asserts `typeof localStorage === 'undefined'` first), constructs `ClientConfig` + `Client`, checks the session-id fallback, pins a byte-stable golden sign key (getCurrentDate is UTC-based so the golden value is TZ-safe), and exercises safeStorage's no-op setItem/removeItem path. A future direct `localStorage` access on the construct-and-sign path now fails CI.

packages/backend.ai-client/package.json
```diff
@@ -19,14 +19,20 @@
...
-      "types": "./dist/index.d.ts",
-      "import": "./dist/index.js"
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in 10007bd: the Runtime support section now documents the ESM + CommonJS entry points (with crypto-es bundled into both), Node >= 22 matching `engines`, and the DOM-free construct/sign guarantee. The Node >= 18 claim predated this PR but contradicted `engines` — aligned to >= 22.

nowgnuesLee:
Re: Copilot's pass-2 suppressed finding on `src/wsproxy/lib/backend.ai-ws-appproxy.js:29` (API mode does not propagate `ext_proxy_url` — the gateway receives a `ClientConfig`, which has no such field, so the external-proxy branch never fires there).

Acknowledged and deliberately **not fixed in this PR**. Context: before this PR that code path crashed outright (`this._env` was never assigned), so nothing regresses — the behavior moves from "crash" to "direct connection, corporate proxy setting ignored", which is also the pre-existing state SESSION mode never had (it receives `ext_proxy_url` via `cf`). Plumbing the proxy URL through `ClientConfig` (or passing the full config to the gateway) is a functional change that needs an API-mode + corporate-proxy environment to verify, which this PR's scope (legacy client removal, behavior-equivalent) does not include. A `TODO` now marks the spot at the assignment site (commit 378408f).

src/wsproxy/manager.js
```diff
@@ -240,7 +240,7 @@ class Manager extends EventEmitter {
         cf['mode'] = 'API';
         cf['access_key'] = req.body.access_key;
         cf['secret_key'] = req.body.secret_key;
-        const ai = require('../lib/backend.ai-client-node');
+        const ai = require('backend.ai-client');
```
agatha197, line 243: Narrower than the PR body implies, but worth stating: the signed `X-BackendAI-Version` has never tracked the `api_version` sent here — the `APIMajorVersion` setter mirrors only `_apiVersionMajor`, in the old client too. What changes is the constant, so `sessionPrefix` keeps following the web UI while the header goes from always-v4 to always-v8. That fixes the modern-manager case; the new combination is a v4 manager getting `/stream/kernel/…` signed as `v8.20240915`. The "same as the web UI's own client" argument doesn't cover it, since the web UI converges via `get_manager_version()` and this proxy never calls it.

src/wsproxy/lib/backend.ai-ws-appproxy.js
```diff
@@ -26,6 +26,9 @@ i18next
 module.exports = proxy = class Proxy extends ai.backend.Client {
   constructor(env) {
     super(env);
+    // TODO: in API mode `env` is a ClientConfig, which carries no ext_proxy_url,
+    // so the external-proxy branch in _start never fires (SESSION mode gets it via cf).
+    this._env = env;
```
agatha197, line 31: This makes `EXT_HTTP_PROXY` a silent no-op in API mode: `env` is `aiclient._config` (manager.js:320), a `ClientConfig` with no `ext_proxy_url`, so the branch at :154 never fires. Replacing a TypeError with silence is an improvement, but `this.extHttpProxyURL` is in scope at the call site — either pass it through, or give the TODO an issue reference per the PR checklist.

packages/backend.ai-client/README.md
```diff
@@ -49,8 +49,11 @@ and password in place of access key / secret key.
 
 ## Runtime support
 
-- ESM-only package (`"type": "module"`)
-- Node.js >= 18 (uses the global `fetch` API)
+- Ships ESM and CommonJS entry points (`"type": "module"` with `import` /
+  `require` export conditions; `crypto-es` is bundled into both)
+- Node.js >= 22 (uses the global `fetch` API). Constructing `Client` and
```
agatha197, line 54: This raises the stated minimum to Node 22 in the same PR that makes a `pkg --targets node18-*` binary (Makefile:163) the package's first Node consumer. It works today, but nothing gates it — a later node20+ API in the client would pass CI and break only the shipped local-proxy binary.

packages/backend.ai-client/src/client.ts
```diff
@@ -85,9 +83,11 @@ function redactSensitiveValues(value: unknown): unknown {
       if (key === '__proto__' || key === 'constructor' || key === 'prototype') {
         continue;
       }
-      redacted[key] = SENSITIVE_LOG_KEYS.has(key.toLowerCase())
-        ? REDACTED_PLACEHOLDER
-        : redactSensitiveValues(val);
+      const lowerKey = key.toLowerCase();
+      redacted[key] =
+        SENSITIVE_LOG_KEYS.has(lowerKey) || lowerKey.includes('password')
```
agatha197: Consistency point rather than a live leak — I could not find a request body in this app that carries a camelCase secret key today. Still, only `password` was generalized to substring matching; `token` / `secret` / `private_key` stay exact-match, so `refreshToken`, `clientSecret` or `secretKey` would pass through unmasked if one ever appears. The new test asserting `confirmPassword` / `newPassword2` shows camelCase bodies were already in scope, so extending the same rule to the rest of the list seems cheap.

agatha197 (changes requested):
Verified locally in a worktree: `scripts/verify.sh` → `=== ALL PASS ===`, client package 11/11, `manager.test.ts` 18/18, and `dist/index.d.cts` is really emitted so `exports.require.types` resolves. Four things below — the `ws-appproxy` one is the only one I would call blocking.

src/wsproxy/manager.js
nowgnuesLee: You're right, and thanks for the precision — the setter has never fed the signed header; what changed is the `_apiVersion` constant (always-v4 → always-v8), while `sessionPrefix` keeps following the web UI's `api_version`. I've rewritten the PR body's behaviour-delta bullet to say exactly that, including the new combination (a pre-v8 manager receiving `/stream/kernel/…` signed as v8) and why we accept it: that manager generation sits below the supported LTS floor (26.4). No code change.

src/wsproxy/lib/backend.ai-ws-appproxy.js
nowgnuesLee: Filed [FR-3836](https://lablup.atlassian.net/browse/FR-3836) for the propagation (plumb `extHttpProxyURL` into what the API-mode gateway receives + verify behind a real corporate proxy) and updated the marker to `TODO(FR-3836)` in 97545e5, per the checklist. We kept the fix out of this PR deliberately: it's a functional change on a path we cannot verify from here (needs API mode + `EXT_HTTP_PROXY`), and this PR's contract is behaviour-equivalence plus the crash fix.

packages/backend.ai-client/README.md
nowgnuesLee: Fixed at the root in 97545e5: the local-proxy `pkg` target moves node18 → node22 (Makefile), so the shipped binary and the package's `engines >= 22` now agree and there's no un-gated node18 constraint left. Verified locally: the node22-macos-arm64 binary builds with zero resolve warnings and answers the API-mode `PUT /conf` probe with a token (package resolved inside the snapshot). @yao-pkg/pkg 6.14 ships node22 base binaries for all six release targets.

packages/backend.ai-client/src/client.ts
nowgnuesLee: Agreed — done in 97545e5: the exact-match list is gone; `redactSensitiveValues` now substring-matches every sensitive stem (`password`, `secret`, `token`, `otp`, `authorization`, `passphrase`, `private_key`/`privatekey`) case-insensitively, so `refreshToken`, `clientSecret`, `secretKey`, `sshPrivateKey` are all masked. A regression test pins the camelCase spellings of each stem.

Makefile
```diff
@@ -164,7 +160,7 @@ compile_localproxy:
 		printf "$(YELLOW)local-proxy $(os)-$(arch) already built, skipping$(NC)\n"; \
 	else \
 		rm -rf ./app/backend.ai-local-proxy-$(BUILD_VERSION)-$(os)-$(arch)$(local_proxy_postfix); \
-		pnpm exec pkg ./src/wsproxy/local_proxy.js --targets node18-$(os)-$(arch) --output ./app/backend.ai-local-proxy-$(BUILD_VERSION)-$(os)-$(arch)$(local_proxy_postfix) --compress Brotli; \
+		pnpm exec pkg ./src/wsproxy/local_proxy.js --targets node22-$(os)-$(arch) --output ./app/backend.ai-local-proxy-$(BUILD_VERSION)-$(os)-$(arch)$(local_proxy_postfix) --compress Brotli; \
```
agatha197, line 168: `compile_client_pkg` is a prerequisite of `compile_wsproxy` only, but `dep_web` skips `compile_wsproxy` whenever `src/wsproxy/dist/wsproxy.js` already exists — always the case in `build_mac` / `build_desktop`, since the `web-build` artifact ships `src/wsproxy/dist/` (package.yml:84). In run 32944408565 the `make dep_electron` step produced no webpack output, confirming the skip, so `packages/backend.ai-client/dist/` is never built in the jobs that run this line; pkg 6.14.1 then exits 0 with only `Warning Cannot find module 'backend.ai-client'`, shipping a binary whose API-mode `PUT /conf` throws MODULE_NOT_FOUND.

Suggest building the package from `dep_web` (guarded on `dist/index.cjs`) rather than as a prerequisite of `compile_localproxy` — `compile_all_localproxy` runs that target six times in parallel, and tsup's `clean: true` would race on one `dist/`.

Not a regression: the same gap exists today through the gitignored `src/lib/backend.ai-client-node.js`, which is what the `Entry 'main' not found ... manager.js` warnings in that run are (reproduced locally). But the PR body's "zero resolve warnings / the package is inside the snapshot" check only holds on a box where `dist/` was already built, so it does not carry over to the release pipeline.

agatha197 (changes requested):
One blocking item on the release build wiring. The rest checked out: `safeStorage` covers every `localStorage` site on the construct-and-sign path, tsup does emit `index.d.cts` for the new `require` condition (verified by building to a scratch dir), `crypto-es` is fully gone from the root, and the two new suites do run in CI via typecheck.yml's `client-test` lane.

Makefile
nowgnuesLee: Fixed in 29ed495 exactly as suggested: `dep_web` now guards on `packages/backend.ai-client/dist/index.cjs` and builds the package itself — once per job, before the six parallel `compile_localproxy` invocations, so tsup's `clean: true` never touches a `dist/` another pack is reading. Reproduced your failure mode locally first (removed `dist/`, kept `src/wsproxy/dist/wsproxy.js` to mimic the web-build artifact): `pkg` emitted the two `Cannot find module 'backend.ai-client'` warnings and exited 0. After `make dep_web` with the guard: zero warnings, and the packed node22 binary answers the API-mode `PUT /conf` with a token. Thanks for chasing this into the actual release run — the local-only scope of my earlier "zero warnings" claim was exactly the blind spot.

agatha197 (approved):
LGTM

PR #9295, opened 2026-08-31
fix(FR-3705): keep form values when the registry modal's project select suspends

### LINKED ISSUE #9123: Add registry modal: toggling 'Set as global registry' re-enables the unchecked 'SSL verification'
In the **Add container registry** modal, uncheck **SSL verification**, then click the **Set as global registry** checkbox to turn it off. The SSL verification checkbox unexpectedly flips back to checked (enabled) — toggling one checkbox resets the other field's value.

**Expected:** each checkbox keeps its own state; toggling "Set as global registry" must not change "SSL verification".

**Observed:** SSL verification reverts to checked as soon as the global-registry checkbox is clicked (see attached screenshot of the two fields).

—

JIRA Issue: FR-3705

nowgnuesLee:
Resolves #9123 (FR-3705)

applied server: https://fr-3735.seungwon.10-82-0-159.sslip.io/

## Bug

In the **Add/Modify container registry** modal, uncheck **SSL verification**, then uncheck **Set as global registry** — SSL verification flips back to checked.

## Root cause

Unchecking `is_global` mounts `ProjectSelectForAdminPage`, whose `network-only` `useLazyLoadQuery` always suspends on mount. The nearest `Suspense` boundary sat **above the form**, so React hid the whole form subtree while the query loaded. A Suspense hide/show cycle runs every effect cleanup, which unregisters every form field — and with the form's `preserve={false}`, each unregistration resets the field to its initial value. On reveal, the fields re-register with initial values: `ssl_verify` (and `is_global` itself) revert to checked.

## Fix

Move the `Suspense` boundary **inside** the form, wrapping only the conditional allowed-projects item, with a loading `BAISelect` fallback that keeps the real control's chrome. The suspension no longer reaches the form, so sibling fields keep their values.

## Regression test

`ContainerRegistryEditorModal.test.tsx` renders the real modal under an outer Suspense boundary (as the app does) with a suspending stand-in for the project select:

- unchecks SSL verification, then unchecks Set-as-global,
- resolves the suspension,
- asserts both checkboxes stay unchecked.

Verified to **fail against the previous code** and pass with the fix.

## Verification

- `pnpm vitest run src/components/ContainerRegistryEditorModal.test.tsx src/form-engine/formEngineAcceptance.test.tsx` → 31 passed

### DIFF
--- a/react/src/components/ContainerRegistryEditorModal.test.tsx
+++ b/react/src/components/ContainerRegistryEditorModal.test.tsx
@@ -1,0 +1,118 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+import '../../__test__/matchMedia.mock.js';
+import '../../__test__/resizeObserver.mock.js';
+import ContainerRegistryEditorModal from './ContainerRegistryEditorModal';
+import '@testing-library/jest-dom';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+import { RelayEnvironmentProvider } from 'react-relay';
+import { createMockEnvironment } from 'relay-test-utils';
+
+vi.mock('react-i18next', async () => {
+  const React = await import('react');
+  return {
+    useTranslation: () => ({
+      t: (key: string) => key,
+      i18n: { language: 'en', changeLanguage: () => new Promise(() => {}) },
+      ready: true,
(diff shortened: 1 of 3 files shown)


agatha197 (changes requested):
`SSL Verification` is clicked with label, but `Set as Global Registry` isn't. And `Set as Global Registry` is not clicked once.

agatha197 (approved):
LGTM

PR #9304, opened 2026-08-31
fix(FR-3800): gate Folder Explorer write and delete on the folder permission

### LINKED ISSUE #9299: Read-only shared folder allows file write and delete in Folder Explorer
## Summary

A virtual folder shared with **read-only** permission still allows the recipient to modify and delete files from the Folder Explorer. The WebUI does not gate write/delete on the folder's share permission at all.

## Affected versions

## Reproduction

1. As user A, create a folder and share it with user B with **read-only** permission.
1. As user B, open the folder in the Folder Explorer.
1. Rename / create / delete files — the actions are enabled and succeed.

## Root cause

`react/src/components/FolderExplorerModalV2.tsx:320-321` hardcodes both capabilities:

```
const hasDeleteContentPermission = true;
const hasWriteContentPermission = true;
```

These feed `enableDelete` / `enableWrite` on `BAIFileExplorer` (`:383-386`). Only upload and the in-app text editor are gated, and they are gated on the **storage host** capability `upload-file` from `allowed_vfolder_hosts` (`:307-310`) — a different permission axis that has nothing to do with how the folder was shared.

There are three distinct permission axes, and the explorer currently consults only the first:

| Axis | Source | Consulted? |
|---|---|---|

|—|—|—|
|---|---|---|

| Storage host capability (`upload-file`, `download-file`) | `allowed_vfolder_hosts` in the resource policy | Yes — upload/edit only |
|---|---|---|

| Folder share permission (`ro` / `rw` / `wd`) | `vfolderV2.accessControl.permission` (`READ_ONLY` / `READ_WRITE` / `RW_DELETE`) | **No — not even queried** |
|---|---|---|

| Effective content permission (`write_content`, `delete_content`) | legacy `vfolder_node.permissions` | **No — absent from the V2 schema** |
|---|---|---|

This is a regression introduced by the V2 migration (FR-2619). The pre-migration component `react/src/components/FolderExplorerModal.tsx:152-159` — now dead code, no longer imported — derived both flags correctly from `vfolder_node.permissions`:

```
const hasDeleteContentPermission = _.includes(vfolder_node?.permissions, 'delete_content');
const hasWriteContentPermission = _.includes(vfolder_node?.permissions, 'write_content');
```

`FolderExplorerOpener.tsx:9` lazily loads `FolderExplorerModalV2`, so the V2 path is the only live one.

The `TODO(needs-backend)` at `FolderExplorerModalV2.tsx:311-319` documents the deliberate decision: because the V2 schema exposes only the **mount** permission and not an effective permission set, the code chose to allow everything and defer to server-side authorization.

## Proposed fix

Short term — add `accessControl { permission `} to the `FolderExplorerModalV2Query` and derive:

- `hasWriteContentPermission` = permission is not `READ_ONLY`
- `hasDeleteContentPermission` = permission is `RW_DELETE`
- gate upload/edit on the storage-host capability **and** the share permission

Long term — have the backend expose a proper effective permission set on `vfolderV2` (the gap FR-2619 recorded), and derive from that instead of the mount permission.

## Server-side follow-up (out of scope for this issue)

The customer observed the file actually being modified, which means the manager accepted the write too. A WebUI-only fix hides the affordance but leaves the API path open. Manager-side enforcement of the read-only share permission needs to be verified separately.

## Verification

- A folder shared as `ro` shows write/delete/upload/edit actions disabled in the Folder Explorer.
- A folder shared as `rw` allows write but not delete; `wd` allows both.

JIRA Issue: FR-3800

agatha197:
Resolves #9299 (FR-3800)

## Problem

A folder shared **read-only** still let the recipient rename, create, delete, upload and edit files from the Folder Explorer.

`FolderExplorerModalV2` hardcoded both capabilities:

```ts
const hasDeleteContentPermission = true;
const hasWriteContentPermission = true;
```

Upload and the in-app editor *were* gated — but on the storage **host** capability `upload-file` from `allowed_vfolder_hosts`, which is a different permission axis and says nothing about how the folder was shared. Three axes exist; the explorer consulted only the first:

| Axis | Source | Before |
|---|---|---|
| Storage host capability | `allowed_vfolder_hosts` in the resource policy | upload/edit only |
| Folder share permission (`ro`/`rw`/`wd`) | `vfolderV2.accessControl.permission` | **not queried** |
| Effective content permission | legacy `vfolder_node.permissions` | **absent from V2 schema** |

This is a regression from the V2 migration (FR-2619). The pre-migration `FolderExplorerModal.tsx` — now dead code, no longer imported — derived both flags from `vfolder_node.permissions` (`write_content` / `delete_content`). V2 dropped that field with nothing equivalent, so the code allowed everything and deferred to server-side authorization. The customer report shows the write actually landing, so that deferral was not holding.

Reported by a customer (KNUT) on **WebUI 26.4.9**; still reproducible on **v26.4.10** and `main`.

## Fix

Query `accessControl { permission }` and derive the gates from it:

```ts
const folderPermission = vfolderNode?.accessControl?.permission;
const hasWriteContentPermission =
  folderPermission === 'READ_WRITE' || folderPermission === 'RW_DELETE';
const hasDeleteContentPermission = folderPermission === 'RW_DELETE';
const hasUploadContentPermission =
  hasUploadHostPermission && hasWriteContentPermission;
```

Resulting matrix:

| Share permission | write (rename/mkdir/create) | delete | upload / editor save |
|---|---|---|---|
| `READ_ONLY` | ✗ | ✗ | ✗ |
| `READ_WRITE` | ✓ | ✗ | ✓ (with host `upload-file`) |
| `RW_DELETE` | ✓ | ✓ | ✓ (with host `upload-file`) |

Two deliberate choices:

- **Allow-list, not `!== 'READ_ONLY'`** — an unknown future enum value (`%future added value`) fails closed rather than silently granting write.
- **Upload/editor require host capability AND folder write** — both write through the upload API, so either axis alone is insufficient.

`accessControl.permission` is the folder's *mount* permission, not a true effective permission set, so the `TODO(needs-backend)` stays — narrowed to that one point and still pointing at the FR-2619 follow-up. It is nonetheless exactly the `ro`/`rw`/`wd` the sharer picked, so gating on it restores the pre-migration behaviour.

## Scope note — server-side enforcement is not in this PR

The customer observed the file actually being modified, so the manager accepted the write too. This PR removes the affordance; it does not close the API path. Manager-side enforcement of the read-only share permission needs to be verified separately and is called out in FR-3800.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / theme / terminology)
- `vitest run src/components/FolderExplorerModalV2.test.tsx` → 3 passed
- Relay artifact regenerated and committed (`FolderExplorerModalV2Query.graphql.ts`)

### DIFF
--- a/react/src/__generated__/FolderExplorerModalV2Query.graphql.ts
+++ b/react/src/__generated__/FolderExplorerModalV2Query.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<4cdc1fafd49e59bd654336f967435926>>
+ * @generated SignedSource<<1f4374b44a312b5064c4884e8a815025>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -10,11 +10,15 @@
 
 import { ConcreteRequest } from 'relay-runtime';
 import { FragmentRefs } from "relay-runtime";
+export type VFolderMountPermission = "READ_ONLY" | "READ_WRITE" | "RW_DELETE" | "%future added value";
 export type FolderExplorerModalV2Query$variables = {
   vfolderId: string;
 };
 export type FolderExplorerModalV2Query$data = {
   readonly vfolderNode: {
+    readonly accessControl: {
+      readonly permission: VFolderMountPermission;
+    };
     readonly host: string;
     readonly id: string;
     readonly metadata: {
@@ -77,27 +81,34 @@ v5 = {
   "alias": null,
   "args": null,
   "kind": "ScalarField",
+  "name": "permission",
+  "storageKey": null
+},
+v6 = {
+  "alias": null,
+  "args": null,
+  "kind": "ScalarField",
   "name": "name",
   "storageKey": null
 },
-v6 = [
-  (v5/*: any*/)
+v7 = [
+  (v6/*: any*/)
 ],
-v7 = {
+v8 = {
   "alias": null,
   "args": null,
   "kind": "ScalarField",
   "name": "projectId",
   "storageKey": null
 },
-v8 = {
+v9 = {
   "alias": null,
   "args": null,
   "concreteType": "ProjectBasicInfo",
   "kind": "LinkedField",
   "name": "basicInfo",
   "plural": false,
-  "selections": (v6/*: any*/),
+  "selections": (v7/*: any*/),
   "storageKey": null
 };
 return {
@@ -121,11 +132,23 @@ return {
           {
             "alias": null,
             "args": null,
+            "concreteType": "VFolderAccessControlInfo",
+            "kind": "LinkedField",
+            "name": "accessControl",
+            "plural": false,
+            "selections": [
+              (v5/*: any*/)
+            ],
+            "storageKey": null
+          },
+          {
+            "alias": null,
+            "args": null,
             "concreteType": "VFolderMetadataInfo",
             "kind": "LinkedField",
             "name": "metadata",
             "plural": false,
-            "selections": (v6/*: any*/),
+            "selections": (v7/*: any*/),
             "storageKey": null
           },
           {
@@ -136,7 +159,7 @@ return {
             "name": "ownership",
             "plural": false,
             "selections": [
-              (v7/*: any*/),
+              (v8/*: any*/),
               {
                 "alias": null,
                 "args": null,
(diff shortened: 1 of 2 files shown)


agatha197:
it is handled with #9315

PR #2357, opened 2024-04-18
feat: add new resource policy page (keypair, user, project)

agatha197:
# New resource policy page
resolves https://github.com/lablup/backend.ai-webui/issues/2351

### Description
Replace the original keypair resource policy page with the react component.
### Checklist
- [ ] list page: added idle timeout, max session lifetime
- [ ] modify/create modal
  - [ ] If the value is unlimited, checkbox is checked and input is disabled.
  - [ ] If the initial value exists -> check `Unlimited`: form value is undefined -> uncheck `Unlimited`: form value is reset.
- [ ] modify modal: host name is disabled.
- [ ] create modal: host name is enabled.
- [ ] allowed host
  - [ ] It displays only host name.
  - [ ] allowed permission list follows original value.
  - [ ] If the allowed permission is undefined, the default value is `{}` as follows server value.
- [ ] you can delete keypair resource policy by clicking the delete icon in the controls column.
   
### Screenshots
- keypair resource policy list

- Update keypair resource policy modal 

- Create keypair resource policy modal

- `AllowedHostNamesSelect`

## Project resource policy page

**Checklist:** (if applicable)

- [ ] Mention to the original issue
- [ ] Documentation
- [ ] Minium required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [x] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

PR #9306, opened 2026-08-31
fix(FR-3801): lift the keypair max session lifetime cap off 100 (26.8)

### LINKED ISSUE #9300: Keypair resource policy: Max session lifetime input is hardcoded to max 100
## Summary

In the keypair resource policy setting modal, the **Max session lifetime (sec)** input is capped at `100`. Any larger value configured elsewhere (e.g. Control Panel) is clamped to 100 as soon as an admin saves the policy from the WebUI.

## Affected versions

## Reproduction

1. In Control Panel, set a keypair resource policy's max session lifetime to a value above 100 (e.g. `3600`).
1. In the WebUI, go to Resource Policy → Keypair, open that policy's setting modal.
1. Save the policy — the value is clamped to `100`.

## Root cause

`react/src/components/KeypairResourcePolicySettingModal.tsx:488-492`:

```
<AstryxFormNumberInput
  label={t('resourcePolicy.MaxSessionLifetime')}
  min={0}
  max={100}
/>
```

`100` is a hardcoded literal. Every sibling numeric field in the same modal uses the shared constant `SIGNED_32BIT_MAX_INT` (imported at `:17`):

- `max_containers_per_session` (Cluster size) — `:477`
- `max_pending_session_count` — `:506`
- `max_concurrent_sessions` (Concurrency) — `:521`
- `max_concurrent_sftp_sessions` — `:549`

The literal has been there since the resource policy page was first introduced (PR #2357) and survived the Ant Design → Astryx migration (PR #8626) unchanged.

There is no product reason for a 100-second ceiling on a session lifetime; the field is expressed in seconds, so 100 is under two minutes.

## Proposed fix

Replace `max={100`} with `max={SIGNED_32BIT_MAX_INT`} to match the sibling fields.

## Verification

- A policy with max session lifetime above 100 (e.g. 3600) can be saved from the WebUI and round-trips without clamping.
- An existing policy created in Control Panel with a value above 100 is not silently clamped when its modal is opened and saved.

JIRA Issue: FR-3801

agatha197:
Resolves #9300 (FR-3801)

Backport of #9301 to the **26.8** release line.

## Why a separate PR instead of a cherry-pick

#9301 targets `main`, where this input is the post-Astryx `AstryxFormNumberInput`. The 26.8 branch predates the Ant Design → Astryx migration and still renders antd's `InputNumber`, so the commit does not apply. Same one-value change, re-applied against this branch's component.

## Problem

The **Max session lifetime (sec)** input in the keypair resource policy setting modal was bounded by a hardcoded `max={100}`. The field is expressed in seconds, so the ceiling is under two minutes — and it clamps silently: an operator who sets `3600` in Control Panel loses it the moment anyone opens that policy's modal in the WebUI and saves.

Reported by a customer (KNUT) during a training session on **WebUI 26.4.9**; reproducible on **26.8.1** and on `main` (fixed there by #9301).

## Fix

```diff
- <InputNumber min={0} max={100} style={{ width: '100%' }} />
+ <InputNumber
+   min={0}
+   max={SIGNED_32BIT_MAX_INT}
+   style={{ width: '100%' }}
+ />
```

`SIGNED_32BIT_MAX_INT` is already imported in this file and is what every sibling numeric field in the same modal uses — cluster size (`max_containers_per_session`), `max_pending_session_count`, concurrency (`max_concurrent_sessions`), and `max_concurrent_sftp_sessions`. Max session lifetime was the lone outlier.

The literal came in with the original resource policy page (#2357). No product rule is attached to it.

`unlimitedValue={0}` on the wrapping `FormItemWithUnlimited` is untouched, so the "unlimited" affordance still writes `0`.

## Verification

- `bash scripts/verify.sh` on this branch → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / terminology)
- No Relay or schema change, so no regenerated artifacts.
- Manual: a policy saved with `3600` round-trips without clamping; an existing policy created above 100 in Control Panel is no longer clamped when its modal is opened and saved.

### DIFF
--- a/react/src/components/KeypairResourcePolicySettingModal.tsx
+++ b/react/src/components/KeypairResourcePolicySettingModal.tsx
@@ -490,7 +490,11 @@ const KeypairResourcePolicySettingModal: React.FC<
                   label={t('resourcePolicy.MaxSessionLifetime')}
                   style={{ margin: 0, width: '100%' }}
                 >
-                  <InputNumber min={0} max={100} style={{ width: '100%' }} />
+                  <InputNumber
+                    min={0}
+                    max={SIGNED_32BIT_MAX_INT}
+                    style={{ width: '100%' }}
+                  />
                 </FormItemWithUnlimited>
               </Col>
               {baiClient.supports('max-pending-session-count') ? (


yomybaby (approved):
LGTM

PR #9312, opened 2026-08-31
feat: notification drawer empty state + header docs recapture (2026-08-31)

lablup-octodog:
### What changed and why

This PR started as the **header** chapter's share of the operator-directed outage recovery — the screenshot captures lost to the 2026-08-27 / 2026-08-30–31 screenshot-server outage, re-run against `main`. The recapture surfaced a real product regression, which this PR now also fixes, because the screenshot cannot be correct without it.

**Scope changed from the original screenshot-only plan**: it now carries one source change alongside the eight images.

#### 1. Product fix — the notification drawer had no empty state

`notification_collapse.png` came back with a **blank white body**. That was not a failed capture: under antd the "no data" placeholder was supplied by `List`'s built-in `locale.emptyText`, and the to-astryx migration replaced `List` with a plain `VStack` map (`WEBUINotificationDrawer.tsx`), which renders nothing for an empty array. The placeholder went with it.

The fix renders Astryx `EmptyState` when the active filter has no rows, matching how the rest of the app handles this (`SettingList.tsx`, `RoleScopePermissionEditModal.tsx`, `QuotaPerStorageVolumePanelCard.tsx` — all `EmptyState`, `isCompact` in constrained slots). Both the **All** and **In progress** tabs now explain themselves instead of showing a void.

The string is the **existing** `notification.NoNotification` key. It was orphaned when `List` went away but is still present in all 21 locale files, so no new translation is required.

#### 2. Screenshot recapture

| Image | Locales | Dimensions | Scope |
|---|---|---|---|
| `header.png` | en, ko, ja, th | 2440×120 (unchanged) | Top header bar strip only |
| `notification_collapse.png` | en, ko, ja, th | 280×900 (was 279×900) | Notification drawer only |

`notification_collapse.png` was recaptured against a build carrying the fix above, so the shipped image shows the empty state rather than a blank panel. Framing scope is unchanged (drawer element only); the 1px width delta is the element bounding box, not a broadened scope. Each locale is a genuine per-locale capture (`Notifications` / `알림` / `通知` / `การแจ้งเตือน`).

### Driving commits

None for the docs half. The commit range (`6f6f51d..b3f3913`) contains only previously merged documentation PRs. The source change is driven by what the recapture found, not by a commit in the range.

### Deliberately skipped

- No prose edits — the chapter contains no `TODO(screenshot)` / `TODO: Re-capture` markers, and the prose makes no empty-state claim that needed changing.

### Known content delta

`header.png` no longer shows the **login session timer** — the capture server's backend does not enable login session extension. The chapter's existing `:::note` already states the timer is conditionally visible and the timer section embeds no dedicated image, so the prose remains accurate.

### Verified

- Screenshots recaptured live in all four locales via `window.switchLanguage()` from a single drawer state, per SCREENSHOT-GUIDELINES.md.

### Conventions note

This PR carries **no `FR-` number**, so it deviates from the repo's `prefix(FR-XXXX)` title convention and will not be linked by `project-status-sync`. That was by design for the nightly docs-sync job; the source change added here would normally warrant its own FR — flagging it for the reviewer to decide whether to split it out.


agatha197 (approved):
LGTM

PR #9315, opened 2026-08-31
fix(FR-3800): gate Folder Explorer write/delete on the legacy per-user permission set

### LINKED ISSUE #9299: Read-only shared folder allows file write and delete in Folder Explorer

ironAiken2:
Resolves #9299 (FR-3800)

## Problem

A folder shared **read-only** still let the recipient rename, create, delete, upload and edit files from the Folder Explorer. `FolderExplorerModalV2` hardcoded:

```ts
const hasDeleteContentPermission = true;
const hasWriteContentPermission = true;
```

## Fix

Keep the V2 modal, but fetch the **legacy `vfolder_node.permissions`** (the per-user RBAC evaluation the pre-FR-2619 modal used) in the same lazy-load query, and derive the gates from it:

```graphql
# TODO(needs-backend): stopgap for FR-3800 — vfolderV2 exposes only
# the mount permission (accessControl.permission), not the caller's
# effective permission set, so write/delete gating reads the legacy
# per-user RBAC list here. Replace with the V2 field once the
# backend adds it (FR-2619 follow-up).
legacyVFolderNode: vfolder_node(id: $vfolderGlobalId) {
  id
  permissions
}
```

```ts
const hasDeleteContentPermission = _.includes(legacyVFolderNode?.permissions, 'delete_content');
const hasWriteContentPermission = _.includes(legacyVFolderNode?.permissions, 'write_content');
const hasUploadContentPermission = hasUploadHostPermission && hasWriteContentPermission;
```

Upload / editor save now require the storage-host `upload-file` capability **and** folder-level write — both write through the upload API. All V2-only features (audit-log tab, ADR-0001 project-context handling, layout fixes) are untouched.

## How to review

A read-only share is already prepared on the team test backend (`http://10.82.0.130:8090`): the folder **`backend.ai-webui`**, owned by `admin@lablup.com`, is shared **read-only (R)** with the standard e2e **user** account (`user@lablup.com` — see `e2e/utils/test-util.ts` seed accounts).

1. Run this branch (`pnpm run dev`) and log in as the **user** account against `http://10.82.0.130:8090`.
2. Data page → the `backend.ai-webui` folder shows **Read only (R)** in the mount-permission column. Open it in the Folder Explorer.
3. **Expected with this PR:** upload (file/folder), create folder/file, per-file rename and delete, bulk delete, drag-and-drop upload, and editor save are all **disabled**.
   **On `main`:** the same actions are enabled and actually succeed — the customer-reported bug.
4. Positive cross-check: open one of the user's **own** folders — every action stays enabled (owner permission).

Note on `rw` shares: the manager's RBAC map (`models/vfolder/row.py`, `LEGACY_PERMISSION_TO_RBAC_PERMISSION_MAP`) grants `delete_content` at `rw` as well — `rw` vs `wd` differ in *mount* capability, not content actions. So an `rw`-shared folder correctly shows both write **and** delete enabled; only `ro` locks the explorer. The UI mirrors exactly what the manager enforces.

## Why the legacy field, not `vfolderV2.accessControl.permission`

Verified against the manager source (`lablup/backend.ai` @ `46de869e57`):

- **`vfolderV2.accessControl.permission` is not a per-viewer value.** The resolver path (`api/gql/vfolder_v2/resolver/query.py:78` → `repositories/vfolder/repository.py:221-231`, `:1175-1202`) copies the **`vfolders.permission` column verbatim** — the owner's default mount permission, marked `# legacy` in the model and defaulting to `'rw'`. The recipient's `vfolder_permissions` share row is used **only as a visibility filter** (`models/vfolder/scopes.py:69-87`), never as a value. Two different users querying the same folder get the identical enum.
- Consequence: for a folder shared **read-only**, the recipient sees `READ_WRITE` (or `RW_DELETE`) — so gating on `accessControl.permission` (the approach in #9304) **cannot fix the customer scenario**. It also *over*-blocks: an owner whose folder column is `'rw'` (the default) would lose the delete affordance on their own folder.
- **Legacy `vfolder_node.permissions` is a genuine per-user RBAC evaluation** (`api/gql_legacy/vfolder.py:281-315` → `get_permission_ctx` → `VFolderPermissionContextBuilder`, `models/vfolder/row.py:1363-1388`). For a non-admin recipient the share row **overrides** the scope-derived set (`row.py:1210-1249`), so a read-only share yields exactly `["read_attribute", "read_content", "mount_ro"]` — no `write_content` / `delete_content`. (Admin recipients get a superset by design.)

The V2 schema exposes no per-viewer permission set at all today, so the extra legacy fetch is the only correct source; the `TODO(needs-backend)` on the query records the switch-over plan.

## Relation to #9304

#9304 gates the V2 modal on `accessControl.permission`. Per the manager-source findings above, that field cannot distinguish a read-only share from the owner's own view, so the read-only-share scenario in the customer report would remain broken (and owners of `'rw'` folders would lose delete). Happy to consolidate however the team prefers.

## Scope note — server-side enforcement is not in this PR

The customer observed the file actually being modified, so the manager accepted the write too. This PR restores the correct UI affordances; manager-side enforcement of the read-only share permission needs separate verification (called out in FR-3800).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay / Lint / Format / TypeScript / theme / terminology)
- `vitest run src/components/FolderExplorerModalV2.test.tsx` → 3 passed
- Relay artifact regenerated and committed (`FolderExplorerModalV2Query.graphql.ts`)
- Manual matrix: `ro` share → write/delete/upload/edit disabled; `rw` / `wd` share → write and delete enabled (matches the manager's RBAC map — see "How to review"); owner → all enabled.

https://claude.ai/code/session_01QkB2dJUVmnSiaHb2vQw95F

### DIFF
--- a/react/src/components/FolderExplorerModalV2.tsx
+++ b/react/src/components/FolderExplorerModalV2.tsx
@@ -49,6 +49,7 @@ import {
   BAIModal,
   type BAIModalProps,
   BAIUnmountAfterClose,
+  toGlobalId,
   useFetchKey,
   useInterval,
   VFolderFile,
@@ -186,36 +187,52 @@ const FolderExplorerModalV2: React.FC<FolderExplorerProps> = ({
   // dashed form via the shared `formatToUUID` helper.
   const vfolderUuid =
     vfolderID.length === 32 ? formatToUUID(vfolderID) : vfolderID;
-  const { vfolderNode } = useLazyLoadQuery<FolderExplorerModalV2Query>(
-    graphql`
-      query FolderExplorerModalV2Query($vfolderId: UUID!) {
-        vfolderNode: vfolderV2(vfolderId: $vfolderId) {
-          unmanagedPath
-          host
-          id
-          metadata {
-            name
+  const { vfolderNode, legacyVFolderNode } =
+    useLazyLoadQuery<FolderExplorerModalV2Query>(
+      graphql`
+        query FolderExplorerModalV2Query(
+          $vfolderId: UUID!
+          $vfolderGlobalId: String!
+        ) {
+          # TODO(needs-backend): stopgap for FR-3800 — vfolderV2 exposes only
+          # the mount permission (accessControl.permission), not the caller's
+          # effective permission set, so write/delete gating reads the legacy
+          # per-user RBAC list here. Replace with the V2 field once the
+          # backend adds it (FR-2619 follow-up).
+          legacyVFolderNode: vfolder_node(id: $vfolderGlobalId) {
+            id
+            permissions
           }
-          ownership {
-            projectId
-            project {
-              basicInfo {
-                name
+          vfolderNode: vfolderV2(vfolderId: $vfolderId) {
+            unmanagedPath
+            host
+            id
+            metadata {
+              name
+            }
+            ownership {
+              projectId
+              project {
+                basicInfo {
+                  name
+                }
               }
             }
+            ...FolderExplorerHeaderV2Fragment
+            ...VFolderNodeDescriptionV2Fragment
           }
-          ...FolderExplorerHeaderV2Fragment
-          ...VFolderNodeDescriptionV2Fragment
         }
-      }
-    `,
-    { vfolderId: vfolderUuid },
-    {
-      // Only fetch when both deferredOpen and modalProps.open are true to prevent unnecessary requests during React transitions
-      fetchPolicy:
-        deferredOpen && modalProps.open ? 'store-and-network' : 'store-only',
-    },
-  );
+      `,
+      {
+        vfolderId: vfolderUuid,
+        vfolderGlobalId: toGlobalId('VirtualFolderNode', vfolderID),
+      },
+      {
+        // Only fetch when both deferredOpen and modalProps.open are true to prevent unnecessary requests during React transitions
+        fetchPolicy:
+          deferredOpen && modalProps.open ? 'store-and-network' : 'store-only',
+      },
+    );
 
   // Permission calculation follows the folder's own ownership project when
   // the folder is project-owned (what the user can do must not depend on the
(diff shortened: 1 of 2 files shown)


react/src/components/FolderExplorerModalV2.tsx
```diff
@@ -300,25 +317,25 @@ const FolderExplorerModalV2: React.FC<FolderExplorerProps> = ({
...
+  // Share-permission gating (FR-3800) reads the legacy per-user RBAC list —
+  // see the TODO(needs-backend) on the query above.
+  const hasDeleteContentPermission = _.includes(
+    legacyVFolderNode?.permissions,
+    'delete_content',
+  );
+  const hasWriteContentPermission = _.includes(
+    legacyVFolderNode?.permissions,
+    'write_content',
+  );
```
agatha197, line 334: `FolderExplorerModalV2.test.tsx` is untouched, so this gating has no test. Please add a case that resolves the operation with `permissions: ['read_content']` and asserts write / delete / upload are disabled, so the bypass cannot regress.

But you can do that with another PR.

agatha197 (changes requested):

agatha197 (approved):

PR #9341, opened 2026-09-01
fix(FR-3820): answer per id in bulk vfolder and user mutations

### LINKED ISSUE #9340: Adopt per-id results from bulk vfolder/user mutations and surface partial failures
Backend PR [lablup/backend.ai#14040](https://github.com/lablup/backend.ai/pull/14040) (BA-7511) made three bulk mutations answer for **every requested id**, instead of returning a bare count.

## Schema changes (all additive, `Added in 26.9.0`)

- `BulkDeleteVFoldersV2Payload` — new `items: [VFolder!]!` and new `failed: [BulkDeleteVFolderV2Error!]!`; `deletedCount` is now `@deprecated`.
- `BulkPurgeVFoldersV2Payload` — new `successes: [UUID!]!`; `purgedCount` is now `@deprecated`.
- `BulkPurgeUsersV2Payload` — new `successes: [UUID!]!`; `purgedCount` is now `@deprecated`.

Note the GraphQL field names are `items` / `successes` — `deletedIds` / `purgedIds` / `purgedUserIds` are the internal Python DTO names, not the exposed API.

## Why the WebUI must act

The same PR gave vfolder bulk-delete a per-item guard. Previously the first failure aborted the whole call (surfacing as a 403 error toast) and `deleted_count` was computed from `len(input.ids)`. Now a partial failure returns **success** with the failures only in `failed`.

`DeleteVFolderModalV2` selects only `deletedCount`, so a partial failure now shows a green **"Deleted 3 folders out of 5"** with no indication of why the other 2 failed — a silent regression.

## Scope

1. `DeleteVFolderModalV2` — select `failed` and report which folders failed, with the reason.
1. `DeleteForeverVFolderModalV2` — same. Its payload has had `failed` since 26.4.4 but it was never selected (pre-existing gap).
1. Migrate all three call sites off the deprecated count fields to `items` / `successes`. Also requires syncing `data/schema.graphql`, which predates #14040.

## Out of scope

Keeping failed rows selected for retry, and updating the Relay store directly from `items` instead of a full refetch. Tracked separately as follow-up.

## Dependency

Requires manager >= 26.9.0. The `items` / `successes` selections will fail against older managers, so this must roll out with the backend.

JIRA Issue: FR-3820

agatha197:
Resolves #9340 (FR-3820)

Backend PR [lablup/backend.ai#14040](https://github.com/lablup/backend.ai/pull/14040) (BA-7511) made three bulk mutations answer for every requested id instead of returning a bare count.

## The regression this fixes

The same PR gave vfolder bulk-delete a per-item guard. Previously the first failure aborted the whole call and surfaced as a 403 error toast, and `deleted_count` was computed from `len(input.ids)`. A partial failure now returns **success** with the failures reported only in `failed`.

`DeleteVFolderModalV2` selected only `deletedCount`, so a partial failure showed a green *"Deleted 3 folders out of 5"* with no indication of why the other two failed.

## Changes

- **`DeleteVFolderModalV2`** — selects `failed` and names the folders that failed together with the backend reason. `failed.vfolderId` is a UUID while the fragment carries a global Relay ID, so names are matched through `toLocalId`.
- **`DeleteForeverVFolderModalV2`** — same treatment. Its payload has carried `failed` since 26.4.4 but never selected it, so per-folder purge errors were silently swallowed. Pre-existing gap, fixed here since it is the same code path.
- **All three call sites** move off the now-`@deprecated` count fields to `items` / `successes`.
- **Per-item failure modal** — the two vfolder modals and `PurgeUsersModal` hand `failed` to `BAIBulkErrorModal` (FR-3334) instead of joining the reasons into one `message.error` toast, which was unreadable past two items and, for users, never said which user each reason belonged to. One row per folder / user: name or email (matched from the failed id through `toLocalId`) and the backend message. It renders as a sibling of the confirm modal so it survives the parent's refetch after a partial success; when every item fails the confirm modal stays open behind it. The count-only fallback below keeps the existing toast for vfolders, since it carries no per-folder reason; the purge-users payload has carried `failed` since 26.3.0, so it needs no gate.
- **Manager-version gate** — `items` / `successes` and the delete payload's `failed` are documented as *"Added in 26.9.0"* (the purge payload's `failed` since 26.4.4), so selecting them unconditionally would make every bulk delete / purge fail GraphQL validation against an older manager. `@include` / `@skip` cannot help here: validation runs before execution, so the field must not be in the document at all. The selections carry the client-side `@since(version:)` directive and the deprecated `deletedCount` / `purgedCount` carry `@deprecatedSince(version:)`; `RelayEnvironment`'s transformer removes each field from the document before the request when the manager is below that version. A `bulk-mutation-per-id-results` capability (26.9.0, `packages/backend.ai-client/src/client.ts`) then picks which count to read. Both branches were executed against a live manager before merge.
- **`data/schema.graphql`** predated #14040, so the four affected types are synced from the merged supergraph. Only those types were touched — the file was not wholesale replaced — and the result is byte-identical to upstream for them.

Three i18n keys are added for the failure modal — `data.folders.DeleteFailureDescription`, `data.folders.ErrorMessage` and `credential.PurgeUsersPartialFailureDescription` — in all 21 languages. The count-only fallback keeps the existing `data.folders.FailedToDeleteFolders` toast.

Note the GraphQL field names are `items` / `successes`. `deletedIds` / `purgedIds` / `purgedUserIds` are the internal Python DTO names, not the exposed API.

## Screenshots

Captured on the project admin **Data** page (`/project-data`, Trash tab) against a manager that serves the 26.9.0 per-id fields. Two trashed folders were selected for **Delete Forever**; while the modal was open, one of them was restored to `ready` out of band, so the backend genuinely refuses to purge it. This is a real per-item failure from the manager, not a stubbed response.

| Purge confirmation | Partial failure reported per folder |
|---|---|
| purge modal | partial failure |

Before this PR the right-hand case showed only the green *"Permanently deleted 1 folders out of 2."* toast. The success toast is unchanged; the failures now open a `BAIBulkErrorModal` with one row per folder (name / error message) instead of being appended to a second toast. The move-to-trash modal shares the same code path; a partial failure there could not be reproduced against this manager because `bulkDeleteVfoldersV2` answers success even for an already purged id.

## Not in this PR

Keeping failed rows selected for retry, and updating the Relay store directly from `items` instead of the current full refetch. Both are separable behaviour changes and are left as follow-up.

## Dependency

The per-id fields need **manager >= 26.9.0**. Older managers are handled by the capability gate above rather than breaking, so this no longer has to roll out in lockstep with the backend.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, agent-cli TS, Vite warmup, StyleX, Astryx theme/integration, z-index, Agent mappings, Terminology).

https://claude.ai/code/session_01CccNgVxWEsXEXWKD5A35WX

🔗 **Dev server**: http://fr-3820-pr9341-answer-per-id.sujin.10-82-0-159.sslip.io


react/src/components/DeleteVFolderModalV2.tsx
```diff
@@ -53,9 +60,17 @@ const DeleteVFolderModalV2: React.FC<DeleteVFolderModalV2Props> = ({
...
         bulkDeleteVfoldersV2(input: $input) {
-          deletedCount
+          items @include(if: $supportsPerIdResults) {
+            id
+          }
+          failed @include(if: $supportsPerIdResults) {
+            vfolderId
+            message
+          }
+          deletedCount @skip(if: $supportsPerIdResults)
```
yomybaby: **Blocker.** `@include`/`@skip` are execution-time directives — validation still requires the field to exist in the server schema, so a manager < 26.9.0 rejects the whole document (`Cannot query field "items"` / `"failed"` on `BulkDeleteVFoldersV2Payload`) and the delete fails outright.

Use the project's client directives instead — the transformer removes them from the AST before sending (`graphql-transformer.ts:141-158`, applied at `RelayEnvironment.ts:74`), whereas it leaves `include`/`skip` untouched:

```graphql
items @since(version: "26.9.0") { id }
failed @since(version: "26.9.0") { vfolderId message }
deletedCount @deprecatedSince(version: "26.9.0")
```

react/src/components/DeleteForeverVFolderModalV2.tsx
```diff
@@ -51,9 +58,15 @@ const DeleteForeverVFolderModalV2: React.FC<
...
+        $supportsPerIdResults: Boolean!
       ) {
         bulkPurgeVfoldersV2(input: $input) {
-          purgedCount
+          successes @include(if: $supportsPerIdResults)
+          failed @include(if: $supportsPerIdResults) {
+            vfolderId
+            message
+          }
+          purgedCount @skip(if: $supportsPerIdResults)
```
yomybaby: Same blocker — `successes` is 26.9.0+, so an older manager rejects the document despite `@include(if: false)`. `@since(version: "26.9.0")` on `successes`/`failed`, `@deprecatedSince(version: "26.9.0")` on `purgedCount`.

(Gating `failed` here is right — the purge payload's `failed` only goes back to 26.4.4.)

react/src/components/PurgeUsersModal.tsx
```diff
@@ -65,11 +66,21 @@ const PurgeUsersModal: React.FC<PurgeUsersModalProps> = ({
...
     useMutation<PurgeUsersModalBulkMutation>(graphql`
-      mutation PurgeUsersModalBulkMutation($input: BulkPurgeUsersV2Input!) {
+      mutation PurgeUsersModalBulkMutation(
+        $input: BulkPurgeUsersV2Input!
+        $supportsPerIdResults: Boolean!
+      ) {
         adminBulkPurgeUsersV2(input: $input) {
-          purgedCount
+          successes @include(if: $supportsPerIdResults)
+          purgedCount @skip(if: $supportsPerIdResults)
```
yomybaby: Same blocker — `successes` is 26.9.0+. `@since` / `@deprecatedSince` here too. Leaving `failed` ungated below is correct.

react/src/components/DeleteVFolderModalV2.tsx
```diff
@@ -86,22 +101,43 @@ const DeleteVFolderModalV2: React.FC<DeleteVFolderModalV2Props> = ({
...
+                  folderNames: _.map(failed, (f) =>
+                    nameByLocalId[f.vfolderId]
+                      ? `${nameByLocalId[f.vfolderId]} (${f.message})`
+                      : f.message,
+                  ).join(', '),
                 }),
               );
+            }
+            if (deletedCount === 0) {
+              if (failed.length === 0) {
```
yomybaby: Nit: the nested `if` duplicates the toast call. Computing the names once, then `if (deletedCount === 0) return;` after, reads flatter. Same shape in `DeleteForeverVFolderModalV2`.

yomybaby (commented):
`@include`/`@skip` are execution-time directives, so `FieldsOnCorrectType` validation still requires the field to exist in the server schema. Validating the text this PR actually sends against a pre-26.9.0 schema (graphql-js) gives `Cannot query field "items"` / `"failed"`, and the query transformer only strips `since` / `deprecatedSince` / `skipOnClient` — not `include`/`skip`. So on a manager < 26.9.0 all three mutations now fail outright, a regression against the `deletedCount`-only selection. Fix: `@since(version: "26.9.0")` / `@deprecatedSince(version: "26.9.0")`.

The rest checks out — `toLocalId` matches `vfolderId: UUID!`, the three modals are the only call sites, the schema sync touches only the four affected types.

react/src/components/DeleteVFolderModalV2.tsx
agatha197: Fixed in 824edebc6 — replaced `@include`/`@skip` with the client directives: `items` / `failed` are now gated with `@since(version: "26.9.0")` and `deletedCount` carries `@deprecatedSince(version: "26.9.0")`, so the transformer strips them before the document reaches an older manager. The `$supportsPerIdResults` variable is dropped from the document; the runtime flag stays only to pick which count to read from the response.

react/src/components/DeleteForeverVFolderModalV2.tsx
agatha197: Fixed in 824edebc6 — `successes` / `failed` are gated with `@since(version: "26.9.0")` and `purgedCount` carries `@deprecatedSince(version: "26.9.0")`, replacing the execution-time directives.

react/src/components/PurgeUsersModal.tsx
agatha197: Fixed in 824edebc6 — `successes` gated with `@since(version: "26.9.0")` and `purgedCount` marked `@deprecatedSince(version: "26.9.0")`; `failed` stays ungated as you noted.

react/src/components/DeleteVFolderModalV2.tsx
agatha197: Flattened in 824edebc6 — the folder names are computed once, a single error toast covers both the partial-failure and all-failed cases, followed by an early `if (deletedCount === 0) return;`. Applied the same shape to `DeleteForeverVFolderModalV2`.

ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3820-pr9341-answer-per-id.sungchul.10-82-0-159.sslip.io)

agatha197:
Dev server on box `sujin`: stopped (was http://fr-3820-pr9341-answer-per-id.sujin.10-82-0-159.sslip.io)

react/src/components/DeleteForeverVFolderModalV2.tsx
```diff
@@ -53,7 +74,12 @@ const DeleteForeverVFolderModalV2: React.FC<
         $input: BulkPurgeVFoldersV2Input!
       ) {
         bulkPurgeVfoldersV2(input: $input) {
-          purgedCount
+          successes @since(version: "26.9.0")
+          failed @since(version: "26.9.0") {
```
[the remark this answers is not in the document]
agatha197: Confirmed against `data/schema.graphql`: the purge payload's `failed` is 26.4.4+, only `successes` is 26.9.0+. Gated `failed` at `@since(version: "26.4.4")` in 20fa4c024 so 26.4.4–26.8 managers get the per-folder purge failure modal too; `successes` stays at 26.9.0 and the count falls back to `purgedCount` there. Comment above the flag updated to match. (The delete payload's `failed` is genuinely 26.9.0+, so `DeleteVFolderModalV2` is unchanged.)

react/src/components/DeleteVFolderModalV2.tsx
```diff
@@ -55,7 +76,14 @@ const DeleteVFolderModalV2: React.FC<DeleteVFolderModalV2Props> = ({
         $input: BulkDeleteVFoldersV2Input!
       ) {
         bulkDeleteVfoldersV2(input: $input) {
-          deletedCount
+          items @since(version: "26.9.0") {
```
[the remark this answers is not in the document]
agatha197: Updated the *Manager-version gate* bullet in the PR description: it now explains that `@include`/`@skip` cannot help because validation precedes execution, that the selections carry client-side `@since` / `@deprecatedSince` directives which `RelayEnvironment`'s transformer strips from the document below the given manager version, and that the `bulk-mutation-per-id-results` capability only picks which count to read. It also notes the 26.4.4 gate on the purge `failed` field.

ironAiken2 (approved):
LGTM

PR #9343, opened 2026-09-01
test(FR-3824): add regression test for Folder Explorer share-permission gating

### LINKED ISSUE #9342: Add regression test for Folder Explorer share-permission gating
Follow-up to FR-3800 (PR #9315). Review comment after merge:

> FolderExplorerModalV2.test.tsx is untouched, so this gating has no test. Please add a case that resolves the operation with permissions: ['read_content'] and asserts write / delete / upload are disabled, so the bypass cannot regress.

Add a test case to `react/src/components/FolderExplorerModalV2.test.tsx` that resolves the `FolderExplorerModalV2Query` with a legacy `vfolder_node.permissions` of `['read_content']` and asserts the file explorer receives `enableWrite` / `enableDelete` / `enableUpload` / `enableEdit` all disabled, plus a positive control with `write_content` / `delete_content` present.

JIRA Issue: FR-3824

ironAiken2:
Resolves #9342 (FR-3824)

## Context

Follow-up to #9315 (FR-3800). Post-merge review comment:

> FolderExplorerModalV2.test.tsx is untouched, so this gating has no test. Please add a case that resolves the operation with permissions: ['read_content'] and asserts write / delete / upload are disabled, so the bypass cannot regress.

## What the test does

The stubbed `BAIFileExplorer` in `FolderExplorerModalV2.test.tsx` now records the props it receives, and `renderModal` gained a `legacyPermissions` knob that feeds the `VirtualFolderNode` mock (the legacy per-user RBAC list the FR-3800 gating reads). Two new cases:

- **Read-only share** — `permissions: ['read_content']` → asserts `enableWrite` / `enableDelete` / `enableUpload` / `enableEdit` are all `false`. The fixture's storage host **does** carry the `upload-file` capability, so the `false` on upload/edit proves the folder-level write gate participates in the AND, not just the host gate.
- **Positive control** — `permissions` containing `write_content` / `delete_content` → asserts all four are `true`, guarding against the gating collapsing to always-`false` (and proving the mock actually flows through).

## Mutation check (test-of-the-test)

Reintroducing the pre-FR-3800 bypass (`const hasDeleteContentPermission = true || …`) in `FolderExplorerModalV2.tsx` makes the read-only case fail — the exact regression this test exists to catch.

## Verification

- `vitest run src/components/FolderExplorerModalV2.test.tsx` → 5 passed (3 existing + 2 new)

https://claude.ai/code/session_01QkB2dJUVmnSiaHb2vQw95F

### DIFF
--- a/react/src/components/FolderExplorerModalV2.test.tsx
+++ b/react/src/components/FolderExplorerModalV2.test.tsx
@@ -7,7 +7,7 @@ import '../../__test__/resizeObserver.mock.js';
 import FolderExplorerModalV2 from './FolderExplorerModalV2';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import '@testing-library/jest-dom';
-import { fireEvent, render, screen } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { Suspense } from 'react';
 import { RelayEnvironmentProvider } from 'react-relay';
 import { MemoryRouter } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


agatha197 (approved):
LGTM

PR #9348, opened 2026-09-01
test(FR-3830): cover each share permission type in the Folder Explorer gating test

### LINKED ISSUE #9347: Cover each share permission type in the Folder Explorer gating test
Follow-up to FR-3824 (PR #9343), which added the first regression tests for the FR-3800 share-permission gating in `react/src/components/FolderExplorerModalV2.test.tsx`.

Those two cases do pin the bypass regression, but they do not distinguish the individual permission types. Mutation testing against the merged test file (patch `FolderExplorerModalV2.tsx`, re-run `vitest run src/components/FolderExplorerModalV2.test.tsx`, check whether any test fails):

Caught by the current tests:

- `hasDeleteContentPermission = true || ...` (reintroduce the FR-3800 bypass)
- `hasWriteContentPermission = true || ...`
- `hasUploadContentPermission = hasUploadHostPermission` (drop the folder-level write gate from the AND)

Survives the current tests:

- delete gate reads `write_content` instead of `delete_content`
- write gate reads `delete_content` instead of `write_content`
- `hasUploadContentPermission = hasWriteContentPermission` (drop the host `upload-file` gate from the AND)
- `hasDownloadContentPermission` forced to false

Three gaps cause this:

1. The positive control turns `write_content` and `delete_content` on together, so cross-wiring the two gates is invisible.
1. `renderModal` has a `legacyPermissions` knob for the folder RBAC list, but the storage-host capability axis is hardcoded in the `Domain` resolver as `{'local:volume1': ['download-file', 'upload-file']`}. The host side of the `enableUpload` / `enableEdit` AND cannot be exercised.
1. `enableDownload` is the only one of `BAIFileExplorer`'s five permission props that no case asserts.

Scope:

- Add a `hostPermissions` knob to `renderModal` that feeds the `Domain` resolver's `allowed_vfolder_hosts`.
- Add a case with `write_content` but no `delete_content`, asserting `enableDelete` is false while `enableWrite` stays true.
- Add a case with `delete_content` but no `write_content`, asserting `enableWrite` / `enableUpload` / `enableEdit` are false while `enableDelete` stays true.
- Add a case with `write_content` present but a host lacking `upload-file`, asserting `enableUpload` / `enableEdit` are false while `enableWrite` stays true.
- Assert `enableDownload` in both the download-file-present and download-file-absent directions.

Each new case must kill the corresponding surviving mutation listed above.

JIRA Issue: FR-3830

agatha197:
Resolves #9347 (FR-3830)

## Context

Follow-up to #9343 (FR-3824), which added the first regression tests for the FR-3800 share-permission gating in `FolderExplorerModalV2`.

Those two cases do pin the bypass regression, but they turn `write_content` and `delete_content` on together and leave the storage-host capability axis hardcoded in the `Domain` resolver. Mutation testing against the merged file — patch `FolderExplorerModalV2.tsx`, re-run the spec, and see whether anything fails — showed four rewiring bugs slipping through.

## Mutation results

| Mutation | Before (#9343) | After |
|---|---|---|
| `hasDeleteContentPermission = true \|\| …` (the FR-3800 bypass) | caught | caught |
| `hasWriteContentPermission = true \|\| …` | caught | caught |
| `hasUploadContentPermission = hasUploadHostPermission` (drop folder write gate) | caught | caught |
| delete gate reads `write_content` | **survived** | caught |
| write gate reads `delete_content` | **survived** | caught |
| `hasUploadContentPermission = hasWriteContentPermission` (drop host `upload-file` gate) | **survived** | caught |
| `hasDownloadContentPermission` forced false | **survived** | caught |
| download gate reads `upload-file` | **survived** | caught |

## What changed

`renderModal` gains a `hostPermissions` knob feeding the `Domain` resolver's `allowed_vfolder_hosts`, so the storage-host axis can be varied independently of the folder RBAC list. Four cases follow from it:

- **`write_content` without `delete_content`** — write/upload/edit enabled, delete disabled.
- **`delete_content` without `write_content`** — delete enabled, write/upload/edit disabled. Together with the case above this separates the two gates, so cross-wiring them is no longer invisible.
- **host without `upload-file`, folder grants `write_content`** — upload/edit disabled while write stays enabled. This is the other side of the `enableUpload` AND; #9343 only covered the folder side.
- **host without `download-file`** — `enableDownload` false while the folder-level grants are unaffected. `enableDownload` was the one permission prop of `BAIFileExplorer`'s five that no case asserted.

## Verification

- `vitest run src/components/FolderExplorerModalV2.test.tsx` → 9 passed (5 existing + 4 new)
- `prettier --check` on the touched file → clean
- All eight mutations above re-run against the extended file → every one fails at least one test

https://claude.ai/code/session_017v17auom1vvawqTKvAj4qy

### DIFF
--- a/react/src/components/FolderExplorerModalV2.test.tsx
+++ b/react/src/components/FolderExplorerModalV2.test.tsx
@@ -218,9 +218,11 @@ const VFOLDER_UUID = '11111111-2222-3333-4444-555555555555';
 const renderModal = ({
   ownershipProjectId,
   legacyPermissions,
+  hostPermissions,
 }: {
   ownershipProjectId: string | null;
   legacyPermissions?: string[];
+  hostPermissions?: string[];
 }) => {
   const environment: RelayMockEnvironment = createMockEnvironment();
   const resolver = (operation: any) =>
@@ -248,9 +250,12 @@ const renderModal = ({
         },
       }),
       KeyPair: () => ({ resource_policy: 'default' }),
+      // The storage-host capability axis. `enableUpload` / `enableEdit` are the
+      // AND of this and the folder-level `write_content`, so both sides need a
+      // knob to be gated independently.
       Domain: () => ({
         allowed_vfolder_hosts: JSON.stringify({
-          'local:volume1': ['download-file', 'upload-file'],
+          'local:volume1': hostPermissions ?? ['download-file', 'upload-file'],
         }),
       }),
       Group: () => ({ allowed_vfolder_hosts: '{}' }),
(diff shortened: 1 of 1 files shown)


ironAiken2 (approved):
LGTM

PR #9352, opened 2026-09-01
fix(FR-3833): send schema-valid order fields from the fair share steps

### LINKED ISSUE #9351: Fair share ordering: user step sends invalid order enum values; sorter keys unchecked against schema
Follow-up to FR-3430 (#8516). Reports that ordering does not work on the fair share steps. Investigation found four webui-side defects (backend-side defects are tracked separately):

1. **User step sends invalid enum values.** `UserFairShareTable` sorter keys are `email` / `username`, and `convertToOrderBy` converts them to `EMAIL` / `USERNAME`, which do not exist in `UserFairShareOrderField` (`USER_EMAIL` / `USER_USERNAME`). GraphQL variable validation rejects the whole query, so clicking those sorters breaks the table. Broken since FR-1968 (#5139).
1. **`convertToOrderBy` bypasses type checking.** It casts its result `as TOrderBy`, so the Relay-generated order-field string unions never constrain the produced field name — which is why defect 1 compiled.
1. **JSDoc/implementation mismatch.** The `convertToOrderBy` JSDoc documents a second field-mapping argument for server compatibility, but the implementation does not have that parameter.
1. **Sorter key lists are hand-maintained and can drift from the schema.** Each fair share table hardcodes `available*SorterKeys` with nothing tying them to the generated `*OrderField` enum types.

Fix: implement the documented `fieldMap` parameter on `convertToOrderBy`, declare a key→order-field map per fair share table typed as `Record<SorterKey, *OrderField>` (compile-time exhaustiveness + schema binding), and pass the maps at the step-level `convertToOrderBy` calls (resource group / domain / project / user).

Note: even after this fix, project-name and user-name ordering will stay inert until the backend restores the dropped order-field mappings (silently ignored since backend 26.4.0, BA-5127 adapter migration) — this fix removes the query-breaking error and makes the webui side correct and drift-proof.

JIRA Issue: FR-3833

ironAiken2:
Resolves #9351 (FR-3833)

## Problem

Follow-up to FR-3430 (#8516): ordering was reported broken on the fair share steps, including a hard crash. Live reproduction against a cluster pinned down the full chain; the webui carried these defects (backend-side defects tracked separately — see Scope):

1. **Clicking the Fair Share Factor sorter crashed the step** (all of domain / project / user). BAITable builds its order string from the dataIndex path joined with `.` — the antd-era table joined with `,` — so the column with `dataIndex: ['calculationSnapshot', 'fairShareFactor']` emitted `calculationSnapshot.fairShareFactor`. `convertToOrderBy`'s comma-only split never fired, snake-casing produced the nonexistent enum value `CALCULATION_SNAPSHOT_FAIR_SHARE_FACTOR`, the server answered 400, and Relay rethrew during render into the error boundary. (The sorter-values allowlist didn't catch it: nuqs re-runs `parseAsStringLiteral` only when reading the URL, not on same-page sets, and the table's `onChangeOrder` cast the emitted string in blind.)
2. **`convertToOrderBy` bypasses type checking.** Its `as TOrderBy` cast means the Relay-generated order-field unions never constrain the produced field name — which is why none of this failed `tsc`.
3. **JSDoc documented a `fieldNameMap` parameter that did not exist.** The mapping escape hatch was designed but never implemented.
4. **Sorter key lists were hand-maintained with nothing tying them to the schema.** Each fair share table hardcoded `available*SorterKeys` free to drift from the `*OrderField` enums, and the keys the table actually *emits* (dataIndex-derived) drifted from those lists too — e.g. the user table emitted `userEmail` / `userUsername` and only reached the correct `USER_EMAIL` / `USER_USERNAME` values by snake-case accident.

## Changes

- **`BAIColumnType` gains a `sortKey` override** (BUI): the order-string field name for a column whose dataIndex is a nested display path. `sortKeyOf` prefers it; default behavior unchanged. Pinned by the new `BAITable.sortKey.test.tsx`.
- **Fair share columns emit canonical keys**: the three fair-share-factor columns set `sortKey: 'fairShareFactor'`, the user table's email/username columns set `sortKey: 'email'` / `'username'`. Emitted keys now match the sorter-values allowlist, so the URL round-trips (sort survives reload) and the header sort indicator lights up.
- **`convertToOrderBy` now implements the documented `fieldNameMap` parameter** (`Record<string, NonNullable<TOrderBy['field']>>`): a mapped key wins; unmapped keys keep the snake-case fallback, so all ~35 existing call sites are unaffected.
- **`convertToOrderBy` splits joined dataIndex paths on `.` as well as `,`**, restoring the antd-era "last segment names the field" contract for any table still emitting a joined path.
- **Each fair share table exports a key→order-field map** typed `Record<SorterKey, *OrderField>` against the Relay-generated enum union (`resourceGroupOrderFieldMap`, `domainFairShareOrderFieldMap`, `projectFairShareOrderFieldMap`, `userFairShareOrderFieldMap`), passed at the step-level `convertToOrderBy` call sites. Adding a sorter key without a mapping, or a schema enum rename, now fails `tsc` instead of failing at runtime.
- **Unit tests** for `convertToOrderBy` (previously untested) and the BAITable `sortKey` override.

## Live verification (cluster at 10.82.0.130)

Before: clicking Fair Share Factor on the domain step → URL `order=calculationSnapshot.fairShareFactor`, request variables `{"field":"CALCULATION_SNAPSHOT_FAIR_SHARE_FACTOR"}`, HTTP 400, error boundary.
After: URL `order=fairShareFactor` / `-fairShareFactor`, request variables `{"field":"FAIR_SHARE_FACTOR","direction":"ASC|DESC"}`, rows re-sort (1.00 → 0.88 on DESC), sort indicator lit, no errors.

## Scope note (backend)

`projectName` / `email` / `username` ordering stays visually inert until the backend restores the order-field mappings dropped in the BA-5127 adapter migration (silently ignored since backend 26.4.0) — tracked in BA-7544. This PR removes the crash, makes every fair share sort send schema-valid values, and wires the whole path drift-proof, so those sorts start working the moment the backend fix lands.

## Verification

- `pnpm vitest run src/helper/index.test.tsx` (react) → 140 passed
- `pnpm vitest run src/components/Table/` (backend.ai-ui) → 61 passed
- Live before/after probe as above

https://claude.ai/code/session_01PpTzZ4dhqVaoQyX5Mp1JPt

### DIFF
--- a/packages/backend.ai-ui/src/components/Table/BAITable.doc.ts
+++ b/packages/backend.ai-ui/src/components/Table/BAITable.doc.ts
@@ -51,7 +51,7 @@ export const docs = {
       name: 'columns',
       type: 'BAIColumnsType<RecordType>',
       description:
-        'antd-shaped column model — title, dataIndex, key, render, sorter, width, align, fixed. Nested children groups are flattened, with the group title rendered as a muted caption above each child header.',
+        'antd-shaped column model — title, dataIndex, key, render, sorter, width, align, fixed. `sortKey` overrides the order-string field name when the dataIndex path is not the field the server sorts by. Nested children groups are flattened, with the group title rendered as a muted caption above each child header.',
     },
     {
       name: 'dataSource',
--- a/packages/backend.ai-ui/src/components/Table/BAITable.sortKey.test.tsx
+++ b/packages/backend.ai-ui/src/components/Table/BAITable.sortKey.test.tsx
@@ -1,0 +1,80 @@
+/*
+ FR-3833 — the order-string field name a sortable column emits.
+
+ The order string is built from the dataIndex path joined with '.', so a
+ column whose dataIndex is a nested DISPLAY path (['calculationSnapshot',
+ 'fairShareFactor']) used to emit 'calculationSnapshot.fairShareFactor' — a
+ string no server order-field enum accepts. `sortKey` overrides the emitted
+ field name; these tests pin the override and the default.
+
+ Structural, not visual — jsdom's lack of layout does not matter.
+*/
+import BAITable from './BAITable';
+import type { BAIColumnsType } from './tableTypes';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+interface Row {
+  id: string;
+  snapshot: { factor: number };
+}
+
+const ROWS: Array<Row> = [
+  { id: '1', snapshot: { factor: 0.5 } },
+  { id: '2', snapshot: { factor: 1.0 } },
+];
+
+const renderSorted = (
+  columns: BAIColumnsType<Row>,
+  onChangeOrder: (order?: string) => void,
+) =>
+  render(
+    <BAITable<Row>
+      rowKey="id"
+      dataSource={ROWS}
+      columns={columns}
+      order={null}
+      onChangeOrder={onChangeOrder}
+    />,
+  );
+
+describe('BAITable sortable column order-string field name', () => {
+  it('emits the sortKey override instead of the dataIndex path', async () => {
+    const onChangeOrder = vi.fn();
+    renderSorted(
+      [
+        {
+          key: 'fairShareFactor',
+          title: 'Factor',
+          dataIndex: ['snapshot', 'factor'],
+          sortKey: 'fairShareFactor',
+          sorter: true,
+        },
+      ],
+      onChangeOrder,
+    );
+
+    await userEvent.click(screen.getByText('Factor'));
+    expect(onChangeOrder).toHaveBeenCalledTimes(1);
+    expect(onChangeOrder.mock.calls[0][0]).toMatch(/^-?fairShareFactor$/);
+  });
+
+  it('defaults to the dot-joined dataIndex path without sortKey', async () => {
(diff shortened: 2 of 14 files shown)


ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3833-pr9352-send-schema-valid.sungchul.10-82-0-159.sslip.io)

agatha197 (changes requested):
Even though I changed name sorter, it doesn't work. And please resolve CI failure.

ironAiken2:
> Even though I changed name sorter, it doesn't work. And please resolve CI failure.

https://lablup.atlassian.net/browse/BA-7544
There is currently an issue whereby the core ignores the ordering field provided in enums without generating an error. I'll check the ci test though.

ironAiken2:
The failing `backend-ai-ui-vitest` check is not from this PR: main's `BAIVFolderPathPicker.doc.ts` documents a prop removed in FR-3675 (#9070), so the doc/source contract test fails on every merge commit. Fix: #9358 (FR-3835). Re-run this check after it merges.

agatha197 (approved):
LGTM

PR #9360, opened 2026-09-01
feat(FR-3813): review overlay read side — pins from GitHub with resolved state, panel and deep links

### LINKED ISSUE #9354: [pr-devserver-review] Review overlay read side — open a `#bai=v3` link on the dev server (deep link to the pinned element)
## Parent

FR-3313 — Per-PR dev server review (Revision 3). Decisions: FR-3785 (block format, self-contained anchor), FR-3791 (deep link, retry ladder). **Rescoped 2026-09-01 (driver decision):** the read side is the deep link only — see below.

## What to build

A `#bai=v3` block pasted on a PR comment, in Teams or in a Claude prompt carries a link whose anchor is self-contained (path, query, selector, testid, text, rect, component). Opening that link on the dev server is the whole read side:

- **Deep link**: on load and on `hashchange`, read the full `#bai=v3.<id>.<anchor>` hash → apply path + query first → find the element (selector → testid → text/rect fallbacks) with a retry ladder while the SPA renders → pin it, highlight it (the highlight fires when the element is found and survives a React re-render), and show a small card with what the link carries: the block's quoted label (route › landmark › tag "on-screen text") and the component. The reviewer's note is not in the anchor — it stays in the comment the link was clicked from. Element not found → one toast. Old `#bai-review=` v1 links → a short toast.
- **Untrusted input**: the hash comes from anywhere — bound the base64 length and the inflated size, validate `v === 3` and field types through the codec before using anything, keep `p` path-safe.
- **Tests**: unit tests for hash parsing / validation / resolution order; one Playwright smoke against a running dev server that opens a self-contained link and asserts the pin.
- `/__review/state` from FR-3811 stays as it is. **No** `/__review/pins`, no GitHub read from the dev server, no polling, no panel list, no pin states, no short id-only form, no served-set logic on the read side. FR-3816 (Teams as a pin source) is Not Planned.

## Why it shrank

The original scope (pins from GitHub with resolved state, panel, polling, short form) existed to draw a list of all pins on the dev server. That list duplicates GitHub's unresolved threads and the board, and it was ~90 % of the code, all of it server-side reads with token / cache / private-repo concerns. The link alone meets the requirement: paste it anywhere, click it, land on the element. PR #9360 (the original scope) is closed as superseded by #9368.

## Acceptance criteria

- [ ] Opening `<url>#bai=v3.<id>.<anchor>` from a fresh tab lands on the element with its pin highlighted and the block's label visible; a link whose element is not on the page shows one toast; a v1 link shows the "old link" toast.
- [ ] An oversized or malformed anchor is ignored without an error in the console beyond one warning.
- [ ] `GET /__review/pins` does not exist (404); `GET /__review/state` is unchanged.
- [ ] Unit tests and the Playwright smoke pass; `bash scripts/verify.sh` passes.

## Blocked by

- FR-3811 — the write side (v3 codec, id, anchor) this ticket reads.

yomybaby:
Resolves #9354 (FR-3813)

## What

The write side (#9337) gave people a `#bai=v3` block to paste into a PR comment. This reads those blocks back: every one pasted on the PRs this dev server serves shows up as a numbered pin on the element it was picked from, with its GitHub state, and a link from a comment or a Claude prompt opens the exact element.

- **`GET /__review/pins`** — every block found on the served PRs, merged by id. **`GET /__review/state`** grows the served set and `isPrivate` alongside the fields it already had. Both take zero request parameters, answer GET only (anything else → 405), and are refused outright for a private repository.
- **GitHub read** — one GraphQL query per served PR per poll: issue comments, review bodies and review threads with `isResolved` / `isOutdated`, `isMinimized` + `minimizedReason` (Hide → Resolved, case-insensitive), and reactions as a hint only. Links are matched after percent-decoding and HTML-unescaping; a pin outside a quote block is accepted and flagged.
- **Merge** — one id is one pin: the same block again (word content equal) is a second source with its own PR badge, a later comment carrying the id is a reply, and resolved / outdated / hint are ORed. Word content is taken per quote block, so a comment carrying two blocks gives each pin its own words, and a GitHub "Quote reply" is both — the block again plus, as the reply, whatever was written under the quote. A PR is badged once however many copies it holds. Numbers are the `createdAt` rank, fixed per payload.
- **Client** — poll at 25 s while visible, 120 s once quiet, nothing while hidden, an immediate read on focus; pins drawn at the anchor with a badge when it cannot be found; one flat panel list with source-PR badges, a per-channel status line and a click-to-highlight that survives re-renders; deep links in both forms, `#bai-review=` v1 ignored with a toast.
- **Served set** — the boot record's `served[]` (the whole stack below the running branch) when the record names this branch, else `gh pr list --head`; MERGED / CLOSED layers drop out on the next poll.

## How

Server, all under `react/vite-plugins/review-overlay/`:

| File | Owns |
| --- | --- |
| `pins/extract.ts` | the link pattern, display text, the normalisation that decides "same block again", `isAnchorV3` |
| `pins/github.ts` | `PR_QUERY` and one PR's response → occurrences, with `gh` injected |
| `pins/merge.ts` | the merge rules, the `createdAt` numbering, anchor decode + validation |
| `served.ts` | the served set and the merged/closed drop |
| `pins-service.ts` | the 15 s cache with a single shared upstream fetch, the private gate |
| `index.ts` | the two routes; nothing else moved |

Client, under `client/`: `resolve.ts` (the anchor ladder), `pins-state.ts` (the six states), `deeplink.ts` (fragment, path+query, retry ladder), `poll.ts`, `panel.ts` (pin layer + panel), plus wiring in `main.ts` and a pins count / panel toggle in `ui.ts`.

Four decisions worth flagging for review:

1. **The fragment is read once, synchronously, at module scope.** React Router keeps it on load, but the app's own `<Navigate replace>` redirects (`/` → default menu, `/start`, the scope roots) rewrite the URL without it as soon as a logged-in session resolves — a later read finds nothing. **The hash is left in the URL rather than cleared**: it is what makes the link re-openable and shareable, and the session pages that forward `hash: location.hash` on their own `navigate()` calls keep it alive across in-app navigation. Because it survives that navigation, the jump it describes is **followed once per tab** (the applied id is remembered in `sessionStorage`): a later reload of a page the reviewer navigated to themselves shows the pin in the panel instead of pulling them back to the pinned view.
2. **Path and query are applied before anchoring, with `location.assign`** — a full reload. `pushState` alone does not re-render React Router, and re-running the overlay boot is cheap; `navigatedForHash` guards the loop.
3. **A private repository disables `/__review/state` too**, not just `/__review/pins` (R3.4 says both). The repository check **fails closed**: an unreadable repo is treated exactly like a private one, so a broken `gh` shuts the endpoints instead of opening them.
4. **`orphan` sits below `replied` in the state ladder**, so "somebody answered you and is waiting" is never masked by "we cannot draw it here". The "not found on this page" badge still renders independently.

Trust boundary: pin text, author logins, resolver names and reply bodies come from anyone who can comment on a public PR, so every one of them reaches the DOM through `textContent` — `panel.test.ts` asserts an `` in a body stays text. `decodeAnchor` only guarantees `v`, `s` and a safe `p`, so `isAnchorV3` type-checks every remaining field on the server before the payload leaves it. The response is a whitelist built field by field; an upstream failure becomes the marker `upstream`, never the `gh` message, which can carry a token or a URL.

## Verification

```
bash scripts/verify.sh
  Relay PASS · Lint PASS · Format PASS · TypeScript PASS · TypeScript (agent-cli) PASS
  Vite warmup paths PASS · StyleX PASS · Astryx theme build PASS · Astryx integration PASS
  z-index ladder mirrors PASS · Agent mappings PASS · Terminology PASS
  === ALL PASS ===

pnpm --filter ./react exec vitest run vite-plugins
  Test Files  15 passed (15)
       Tests  221 passed (221)

E2E_REVIEW_OVERLAY_SMOKE=1 E2E_WEBUI_ENDPOINT=http://127.0.0.1:9081 \
  pnpm exec playwright test e2e/review-overlay-deeplink.spec.ts --project chromium
  ✓ @smoke a self-contained link pins the element and opens its panel item (16.0s)
  ✓ the endpoints take no parameters, answer GET only and share one read (276ms)
  2 passed
```

The smoke ran against the dev server this PR advertises (the worktree's own
`pnpm dev` with `VITE_DEV_REVIEW_OVERLAY=1`), and passed 3 runs out of 3. The
two `[cleanup]` teardown failures in that run are the suite's pre-existing
global vfolder sweep, which logs in against a backend this server does not
have; they are unrelated to this file.

It failed 2/2 before this branch's last two commits, for reasons that were
about the test and one real product gap, not the deep link: a Playwright
`button:visible` wait pierces the overlay's shadow root and matched its own
dock button, so the anchor was captured on the transient `/` rather than the
`/project/<name>/start` this logged-in dev server settles on ~5 s in; the
candidate filter picked react-grab's full-viewport `[data-testid="react-grab-overlay"]`
div; and it asserted on the first `.pin` in DOM order, which on this very PR is
a real pin anchored to `/`, hidden as an orphan. The settle is now polled
in-page, the filter skips both overlays and prefers a visible interactive
element with text, and both the pin marker and the panel item are addressed by
`data-pin-id`.

The product gap the smoke exposed: the deep link's highlight fired when the
pins payload arrived (~0.5 s) and `HIGHLIGHT_MS` is 2600, so it had expired by
the time the anchor retry ladder placed the pin (~5 s) — the link landed on the
element with nothing highlighted but the element's own flash. `locatePin` now
takes `highlight`, the retry ladder passes it, and the expiry stops the marker
pulsing as well as the item highlighting. Two new unit tests in
`client/panel.test.ts` pin the ordering (highlight at placement, not at
payload) and its survival across a poll (219 → 221).

New unit coverage (120 of those tests): extraction and the anchor field check, the GraphQL mapping (Hide → Resolved case-insensitively, a 👍 as a hint only, thread replies), the merge rules and numbering, the served set and the merged drop, the cache (one fetch for concurrent callers, failures cached, no token in the payload, a whitelisted key set), the anchor ladder, the six states, the fragment forms and the retry ladder, the poll cadence, and the panel's rendering, escaping and highlight stability. The review round added the rest: the per-block attribution and the quote-reply's own answer, a source counted once per PR, a thread reply carrying the pin link counted once, an unsubmitted review never read, a merged layer's pins frozen and an empty served set re-asked, the anchor size bound (a 200 KB payload that fits in one comment is refused), a selector hit with no text refused, the quiet cadence with a fresh `fetchedAt` on every answer, and the pin layer keeping what the full ladder found across a reposition and two flashes.

Live checks against a dev server booted from this worktree with `VITE_DEV_REVIEW_OVERLAY=1` are in a comment below.

## Not in this PR

- **Teams as a pin source** — FR-3816. The merge is written around a `channel` union and a `sources[]` list so it folds in without reshaping a pin, but nothing here reads Teams.
- **Pagination** — the query is one page per PR (50 / 50 / 50 × 10, the spec's 2-point shape); more than that is flagged as `truncated` in the panel's status line, never paged.
- **`dev.mjs` and the dev-server skill are untouched.** `VITE_DEV_REVIEW_OVERLAY` stays a manual opt-in, and "dev.mjs is unchanged and learns nothing new" stays true (R3.2).

https://claude.ai/code/session_0165N2Y5PnF5AnzvuTx96k64

### DIFF
--- a/DEV_ENVIRONMENT.md
+++ b/DEV_ENVIRONMENT.md
@@ -120,7 +120,7 @@ Two endpoints back it, both **GET only, with no request parameters**, both dev-o
 | `/__review/state` | the running PR, the served set, the repository and `isPrivate` |
 | `/__review/pins`  | every `#bai=v3` block found on the served PRs, merged by id    |
 
-The served set comes from the dev-server skill's boot record (`BAI_REVIEW_BOOT_RECORD`) when there is one — that is the whole stack below the running branch — and from `gh pr list --head <branch>` otherwise. Reads go through the box's own `gh` credentials, are cached for 15 s with one shared upstream fetch however many people are looking, and are **refused entirely for a private repository**. The token never reaches the browser.
+The served set comes from the dev-server skill's boot record (`BAI_REVIEW_BOOT_RECORD`) when there is one — that is the whole stack below the running branch — and from `gh pr list --head <branch>` otherwise. Reads go through the box's own `gh` credentials, are cached for 15 s with one shared upstream fetch however many people are looking, and are **refused entirely for a private repository**. Only what the PR page already shows is served: a review you have started but not submitted is skipped. The token never reaches the browser.
 
 ## CLI login (`/cli-login`)
 
--- a/DEV_ENVIRONMENT.md
+++ b/DEV_ENVIRONMENT.md
@@ -103,6 +103,25 @@ VITE_THEME_HEADER_COLOR=#7C3AED
 
 Vite auto-loads `VITE_*` vars from this file and exposes them on `import.meta.env` for the React app, tinting the header so you can tell multiple instances apart at a glance. You can also export `VITE_THEME_HEADER_COLOR` in the shell — same effect, no file edit needed.
 
+## Review overlay (`VITE_DEV_REVIEW_OVERLAY`)
+
+Off by default. Set it in `.env.development.local` or the shell before `pnpm run dev`:
+
+```bash
+VITE_DEV_REVIEW_OVERLAY=1
+```
+
+With it on, the dev server injects the review overlay (`react/vite-plugins/review-overlay/`). ⌘⌃C — or the dock's **📍 Review** button — picks an element and copies a `#bai=v3` markdown block to paste into the PR comment, the PR's Teams thread, or a Claude prompt. Every such block already posted on the PRs this server serves comes back as a numbered pin on the element it was picked from, with its GitHub state (resolved / replied / reacted / outdated), and opening a block's link lands on that element.
+
+Two endpoints back it, both **GET only, with no request parameters**, both dev-only:
+
+| Endpoint          | Answers                                                        |
+| ----------------- | -------------------------------------------------------------- |
+| `/__review/state` | the running PR, the served set, the repository and `isPrivate` |
+| `/__review/pins`  | every `#bai=v3` block found on the served PRs, merged by id    |
+
+The served set comes from the dev-server skill's boot record (`BAI_REVIEW_BOOT_RECORD`) when there is one — that is the whole stack below the running branch — and from `gh pr list --head <branch>` otherwise. Reads go through the box's own `gh` credentials, are cached for 15 s with one shared upstream fetch however many people are looking, and are **refused entirely for a private repository**. The token never reaches the browser.
+
 ## CLI login (`/cli-login`)
 
 `bai-agent login` (`packages/backend.ai-agent-cli`, the Backend.AI WebUI Agent CLI) hands the CLI the session this dev server is already logged in with, via the `/cli-login` page. The route is always mounted; it is not linked from any menu.
--- a/e2e/review-overlay-deeplink.spec.ts
+++ b/e2e/review-overlay-deeplink.spec.ts
@@ -30,13 +30,6 @@ test.describe('@review-overlay review overlay deep link', () => {
         (window as unknown as { __baiReviewOverlay?: boolean })
           .__baiReviewOverlay,
     );
-    // Let the SPA settle on its final route before the anchor is captured —
-    // the deep link reproduces that route, so it must be the resolved one.
-    await page
-      .locator('button:visible, a[href]:visible')
-      .first()
-      .waitFor({ timeout: 30_000 });
-    await page.waitForTimeout(2000);
 
     const link = await page.evaluate(
       async ([anchorUrl, codecUrl, idUrl]) => {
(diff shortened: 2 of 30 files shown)


yomybaby:
> 📍 **Start › sidebar › a "Backend.AI Logo"** · `c_k6ojb42`
> FR-3813 end-to-end check: this block was posted from the dev box and should come back as pin on the running server.
> [Open on dev server](http://fr-3813-pr9360-review-overlay-read.jongeun.10-82-0-159.sslip.io/#bai=v3.c_k6ojb42.q1YqU7Iy1lEqVrJSSozOKEpNs41R0i8uSSwqiVGKVdJRKlCyUtJX0lEqSUwHKQGxKkqUrJScEpOzU_NS9Bw9FXzy0_OVagE)

yomybaby:
Fixed in `0000000` — end-to-end reply check for FR-3813; this carries the same pin id so it should fold in as a reply: #bai=v3.c_k6ojb42

yomybaby:
Dev server on box `jongeun`: stopped (was http://fr-3813-pr9360-review-overlay-read.jongeun.10-82-0-159.sslip.io)

yomybaby:
## Live verification on a dev server booted from this branch

Server: `VITE_DEV_REVIEW_OVERLAY=1 PORT=9081 pnpm dev` from this worktree, with `BAI_REVIEW_BOOT_RECORD` from the dev-server skill's `advertise.sh boot-env`; `portless-doctor` clean afterwards.

**The whole loop, end to end.** The block two comments up was built on this box with the modules the dev server itself serves (`/__review/codec.js`, `/__review/id.js`, `/__review/block.js`) and posted with `gh`. It came back as a pin within one poll:

```
$ curl -s http://127.0.0.1:9081/__review/pins | jq '.pins[0] | {id, number, sourcePr, quoted, author}'
{ "id": "c_k6ojb42", "number": 1, "sourcePr": 9360, "quoted": true, "author": "yomybaby" }
```

Then, in order:

| Step | Result |
| --- | --- |
| `Fixed in …` comment carrying the same id | `replyCount: 1`, `latestReply` = that comment, `sources` = 2 — a reply, not a second pin |
| Hide → Resolved on the block comment (`minimizeComment(classifier: RESOLVED)`) | `resolved: true` on the next poll (un-hidden again afterwards so the block stays readable) |
| 8 concurrent `GET /__review/pins` | one identical `fetchedAt` for all of them — one upstream read, not one per viewer |
| `GET /__review/pins?anything=1` | byte-identical to the parameterless response |
| `POST /__review/pins` | `405` |
| `GET /__review/state` | `{"pr":9360,"repo":"lablup/backend.ai-webui","branch":"feat/FR-3813-overlay-read-side","source":"gh","served":[{"pr":9360,…}],"isPrivate":false}` |

**Playwright smoke** (`E2E_REVIEW_OVERLAY_SMOKE=1 pnpm exec playwright test e2e/review-overlay-deeplink.spec.ts --project chromium`) — **2 passed**. It builds the anchor in the page with the real codec, opens `<path>#bai=v3.<id>.<anchor>` in a fresh tab, and asserts the pin and the panel item. The 2 additional failures in that run are the repo's `global-cleanup.teardown.ts` project, which logs into a backend this box has no credentials for — unrelated to this PR.

Two honest notes:

- `source` stayed `gh` rather than `boot-record` above, because `/__review/state` resolves once at boot and `advertise.sh` writes the record only *after* the server is up. Correct by design (the record describes a name Portless has actually claimed), and the served set is the same either way for a one-layer stack. The stack case is covered by `served.test.ts`.
- The stacked-PR path (a block posted on a lower layer showing on the top layer's server with that PR's badge) is unit-tested end to end in `pins-service.test.ts`, not exercised live — this branch is unstacked.

https://claude.ai/code/session_0165N2Y5PnF5AnzvuTx96k64

yomybaby:
Superseded by #9368 — the read side was rescoped to the deep link only (driver decision 2026-09-01, spec frontend-board#182, FR-3813 updated). Branch kept for reference; the dev server this PR advertised is stopped.

PR #9361, opened 2026-09-01
feat(FR-3605): convert theme.json to the v2 appearance document and retire the theme-shim

### LINKED ISSUE #8919: Convert theme.json to the v2 {theme, branding} appearance document (Astryx vocabulary)
## Scope (restructured 2026-09-01 — bottom PR of the FR-1964 app-config stack)

Convert the theme document to the **v2 `{theme, branding}` appearance document** (Astryx vocabulary), so no antd-shaped document is ever stored and no data migration is needed when it is served from app config (FR-1964). PR: #9361.

- Convert `resources/theme.json` + `theme.schema.json` to v2: `schemaVersion: 2`, `theme.families.<family>.seeds` (string = both schemes, `[light, dark]` tuple splits them), `headerBg`, `fontFamily`, `siderMode`; `branding.logo` / `companyName` / `brandName` / `familyLabels`.
- Loader (`customThemeConfig`) reads **v2 only**; a v1 document logs a loud console error and falls back to the built-in defaults. **Breaking** for operators with a custom `theme.json` — migration guide + release note required (v1→v2 field map in the PR description).
- Roles stay derived in code (brand←accent, admin←info, secondary←success); the document stores brand seeds only. `computeThemeName` hash inputs unchanged, so the prebuilt brand theme artifact still matches by name.
- Branding editor **format adaptation** rides along (color pickers write seed tuples, `ThemeJsonConfigModal` validates the v2 schema, stale pre-v2 localStorage draft reseeded) so the editor never produces a document the loader rejects. The dead Text Color item is dropped.
- **Remove `custom_primary_color` / `ThemeAccentColorPicker`** (2026-08-19 decision) — the accent override mechanism is antd's palette algorithm over the v1 document and dies with the internal-currency switch. `usePrimaryColors` stays (reads shim tokens, not the setting). **(Moved here from FR-3834: engineering coupling.)**
- Delete `antdThemeConfig.schema.json` and its generator script.

## Out of scope (follow-ups)

- Serving the document from app config + bootstrap unification → FR-1964 (PR #8860).
- Theme family via userConfig + Branding Apply-to-app-config → FR-3834 (PR #9362).
- Migrating remaining `theme.useToken()` call sites and deleting the antd-shaped theme **shim** internals (items 2–3 of this ticket's original goal) — split out, to be re-ticketed when scheduled.

JIRA Issue: FR-3605

ironAiken2:
Resolves #8919 (FR-3605)

## Summary

Bottom PR of the FR-1964 app-config stack (restructured out of #8860 per the 2026-09-01 decision). Two things land together, because the second is what makes the first the app's *only* theme currency:

1. **The v2 `{theme, branding}` appearance document** in Astryx vocabulary replaces the antd-shaped `theme.json`, so when FR-1964 serves it from app config no antd-shaped document is ever stored and no data migration is needed later.
2. **The theme-shim is retired.** Every `theme.useToken()` consumer (189 files across `react/` and `backend.ai-ui`) reads Astryx tokens through `useTheme().token('<css-var>')`; the antd-vocabulary `ThemeShimProvider` and its runtime palette machinery are deleted, and `<AstryxBrandTheme>` is the app's single token provider. What the shim used to compute for names Astryx has no token for now lives in the theme itself as custom custom tokens.

```jsonc
{
  "schemaVersion": 2,
  "theme": {                       // feeds the Astryx pipeline
    "fontFamily": "'Ubuntu', Roboto, sans-serif",
    "siderMode": "dark",           // optional (was sider.theme)
    "families": {
      "default": {                 // explicit entry; top-level light/dark synthesis removed
        "seeds": {                 // string = both schemes, [light, dark] tuple splits them
          "accent": ["#FF7A00", "#DC6B03"],
          "link": ["#FF7A00", "#DC6B03"],
          "info": ["#028DF2", "#009BDD"],
          "error": ["#FF4D4F", "#DC4446"],
          "success": ["#00BD9B", "#03A487"],
          "warning": ["#FAAD14", "#FAAD14"]   // declared explicitly now: nothing in code fills in a seed the document omits
        },
        "headerBg": ["#FF9729", "#E88A28"]   // was components.Layout.headerBg
      }
    }
  },
  "branding": {                    // structural metadata no token can express
    "logo": { /* unchanged LogoConfig */ },
    "companyName": "Lablup Inc.",
    "brandName": "Backend.AI",
    "familyLabels": { "stained": "Stained" }   // was families[*].label
  }
}
```

### How the document reaches the screen

`theme.json` → `loadCustomThemeConfig` → `useCustomThemeConfig` → `AstryxBrandTheme` → `resolveRoleTheme` → `buildBackendAITheme` → Astryx `defineTheme` → `<Theme>`. The seeds run through Astryx's generator (derived accent ramp) and the recipe pins the parity tokens on top. **The document is the only source of brand values**: the recipe carries no defaults, a seed the document omits is simply not pinned (Astryx's token applies; the custom vocabulary falls back to Astryx `var()` references), and the prebuilt artifact is built from `resources/theme.json` itself. Nothing gates the first paint on the document: the first visible screen is already held by `useSuspendedBackendaiClient` (config + client init), which `theme.json` beats, so the default theme is never on screen before the document applies (checklist 3). The shipped document still hits the precompiled CSS (`bai-r27-…`, `THEME_NAME_REV` 22 → 27: 23 for the custom token layer, 24 for the verbatim dark values and the muted pin below, 25 for dropping the `--bai-` prefix from the custom tokens, 26 for the `color-mix()` tints, 27 for renaming `--color-warning-hover` to `--color-warning-border-hover` so both palette-step-4 tokens share the `-border-hover` spelling).

### The custom token layer (`packages/backend.ai-ui/src/theme/baiCustomTokens.ts`)

Custom tokens carry **no prefix** (`--color-info`, `--header-bg`, `--primary-5`, …): a `useTheme().token()` read looks the same whether the token is Astryx's or ours, and a name Astryx later adds is overridden by our value by design. Component-local CSS custom properties (`--bai-gap`, `--bai-form-item-*`, `--bai-z-*`) are not theme tokens and keep their prefix.

Shared by the app recipe and the Storybook brand theme:

| Token | Source |
|---|---|
| `--color-info`, `--color-link` | brand seeds Astryx has no token family for (+ `options.link`, hash input) |
| `--header-bg` | family `headerBg`, verbatim per scheme, with a default-family fallback the shim path never had |
| `--color-{error-bg, info-bg, warning-hover, success-border-hover, primary-bg, error-border}` | `color-mix()` tints of the pinned base token over `--color-background-surface` (bg ≈ antd step 1, border ≈ step 3, hover ≈ step 4), so they follow a rebrand with no palette code |
| `--primary-5` | the accent at 85% (light) / 65% (dark) over the surface — the ramp step progress fills consume |
| `--color-text-*`, `--color-fill-*`, `--color-bg-container-disabled`, `--preset-*` | antd's neutral alpha ramp and the preset steps, verbatim |

**Dark values render as declared.** The antd `darkAlgorithm` re-mapping of the dark seed (`ANTD_DARK_ALGORITHM_OUTPUT` / `resolveDarkSeed`, `#DC6B03` → `#be5e06`) is gone with the shim: the `[light, dark]` tuple is pinned verbatim, so what an operator writes is what renders. The vendored antd palette generator (`antdColors.ts`, its fixture and test) is deleted with it: the derived bg/border/hover tints are `color-mix()` references now, and the one remaining consumer of `presetPalettes` (`UsageBucketChartContent`'s series colors) carries its 13 hexes inline. `--color-accent-muted` is pinned as well (accent at 20%/25% alpha, like the status muted steps): Astryx's `useTheme()` re-applies the base theme's input tokens over generated ones, so an unpinned muted resolved to `neutralTheme`'s grey in the hook while the CSS showed the brand tint.

### The sweep, mechanically

`mapping.ts` — the shim's own verdict table — was the lookup: `token.marginMD` → `token('--spacing-5')`, `token.colorBgContainer` → `token('--color-background-surface')`, brand names → `--color-accent`/`--color-info`/…, `token.Layout?.headerBg` → `--header-bg` (`headerHeight` was never provided, so its `?? 60` fallbacks were constants — inlined). Value semantics changed from antd numbers to CSS strings, so the non-style positions were fixed by hand: negative margins via `calc()`, numeric props via `parseFloat()`, `BAIFlex gap` sites pass the rung name. `useTheme` is a real hook where `useToken` read a store, so the calls inside ChatMessageContent's render callbacks and after the TextHighlighters' early returns are hoisted. **No visual change intended** — every replacement resolves to the value the shim produced for the same document and mode.

Two things in the sweep are not token replacements:

- The About modal's built-in logo default was `/manifest/backend.ai-white-text.svg`, a file that does not exist; it now falls back to the shipped `manifest/backend.ai-webui-black.svg` / `-white.svg`. Both the login and the About defaults are relative paths (`manifest/…`), so they resolve the same way under a sub-path deployment.
- Four files carry prettier-only changes from the sweep's format pass: `bulkUserCSV.ts`, `resultTypes.test.ts`, `useAIAgent.ts`, `useWebUIConfig.ts`.

### Storybook

`ThemeShimProvider` leaves the decorators. The FR-3819 "Theme" toolbar presets keep their Astryx layer (`neutralTheme` vs the brand theme) — with every component on `useTheme()` that layer is the whole switch. `.storybook/theme.json` is a copy of the v2 document and `astryxBrandTheme.ts` builds from it through the shared BUI helpers, so the custom tokens reach stories too. Its `$schema` is `.storybook/theme.schema.json`, a BUI-local schema that describes only what `astryxBrandTheme.ts` reads (the default family's seeds, `headerBg`, `fontFamily`); BUI deliberately does not reference the WebUI's `resources/theme.schema.json`, so the package stays free of a dependency on the host repository.

### ⚠ Breaking for operator `theme.json`

- The loader reads **v2 only** — any document without `"schemaVersion": 2` (a v1 document included; the loader does not sniff v1 shapes, it reports the schema mismatch) logs a loud console error and falls back to Astryx's neutral theme (no Backend.AI colors: the code holds no brand values of its own); so does a fetch/parse failure. A document without `theme.families.default` — no `theme` block, no `families`, or a `families` without `default` — is accepted, with the same loud console error, since it renders brand-less too. Operators with a custom `theme.json` must migrate (release notes + migration guide required before release; field map below).
- Dark-side values render **as declared**. Under antd the dark seed was an input to `darkAlgorithm` (`#DC6B03` rendered as `#be5e06`); v2 pins it verbatim, so the shipped dark brand shifts to the document's values (accent/link `#DC6B03`, error `#DC4446`, success `#03A487`, info `#009BDD`, warning `#FAAD14`). An operator who wants the previously rendered darks writes those instead (`#be5e06`, `#be3d3f`, `#068e76`, `#0387bf`, `#d89614`).
- Roles stay derived in code (brand←accent, admin←info, secondary←success; `reverse` is a `<Theme mode>` inversion, not a role). The document stores brand seeds only.
- Dropped v1 keys with **no consumer anywhere** (verified): `colorText`, `colorBorder`, `colorBgContainer`, `colorBgLayout`, `colorFillSecondary`, `screen*` breakpoints, `components.Tag/Divider/Table`, `siderBg`/`lightSiderBg`, and the antd component-token passthrough beyond `Layout.headerBg`.
- `custom_primary_color` / `ThemeAccentColorPicker` removed (2026-08-19 decision); the shim's `usePrimaryColors` is gone with the shim.

### Branding editor

- Color pickers write whole seed tuples; the untouched scheme keeps the recipe's rendered default instead of inheriting the color just picked. Logo items write `branding.logo.*`; the font item writes `theme.fontFamily`. The dead **Text Color** item is removed.
- The JSON modal validates against the single `theme.schema.json` in Monaco (editor markers; the OK button refuses a draft with markers). The schema rejects unknown keys at every level (`additionalProperties: false`) and non-6-digit-hex seeds; `headerBg` alone takes any CSS color, since it is applied verbatim (the shipped `glass` family uses `rgba(…)`). The write boundary itself (`pickValidAppearanceConfig`) is a structural check only — `schemaVersion: 2` and a `default` family — so when the schema fetch fails a draft still cannot bypass the `schemaVersion: 2` rejection (the toast says the document does not match the theme schema), but unknown keys and non-hex seeds are not caught there; full schema enforcement at the write boundary is left for the app-config save (#9362). `antdThemeConfig.schema.json` and its generator script are deleted; a new `themeJson.test.ts` (ajv) keeps `theme.json`, the schema and the loader agreeing.
- A stale pre-v2 draft is reseeded from the **shipped** document (never from the applied one — in preview mode that is the draft itself).
- Apply-to-domain is **not** in this PR — the editor still previews only; saving to app config arrives at the top of the stack (#9362).

## v1 → v2 field map (for the migration guide)

| v1 (antd-shaped) | v2 |
|---|---|
| `light.token.colorPrimary` / `dark.token.colorPrimary` | `theme.families.<f>.seeds.accent` `[light, dark]` |
| `*.token.colorLink` | `theme.families.<f>.seeds.link` |
| `*.token.colorInfo` | `theme.families.<f>.seeds.info` |
| `*.token.colorError` / `colorSuccess` / `colorWarning` | `theme.families.<f>.seeds.error` / `success` / `warning` |
| `*.components.Layout.headerBg` | `theme.families.<f>.headerBg` |
| top-level `light`/`dark` (default family) | explicit `theme.families.default` entry |
| `families.<f>.label` | `branding.familyLabels.<f>` |
| `fontFamily` | `theme.fontFamily` |
| `sider.theme` | `theme.siderMode` |
| `logo`, `branding.companyName/brandName` | `branding.logo`, `branding.companyName/brandName` |
| `colorText`, `colorBorder`, `colorBg*`, `screen*`, `Tag/Divider/Table`, `siderBg` | dropped (no consumer) |

## Rebase onto main (2026-09-11)

`SSHKeyBlock.tsx` landed on main while this branch was in review and reads
`theme.useToken()`, so the shim retirement left it importing a deleted module —
a semantic conflict `tsc` catches but the merge does not. It is converted here,
in the layer that performs the deletion, so this PR's own CI stays green. The
committed `astryx-theme/built/*` artifacts and `searchIndex.json` were
regenerated rather than hand-merged.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, Vite warmup, StyleX, **Astryx theme build — prebuilt artifact `bai-r27-default-brand-h1ux0ib0` current (THEME_NAME_REV 27)**, z-index mirrors, Terminology; the one terminology finding is the pre-existing warn-only `ChangeGroupColor`)
- After the rebase onto `main` (`f134ba0c3`): `bash scripts/verify.sh` → `=== ALL PASS ===`; `node scripts/migration-gates/astryx-token-gate.mjs --strict` → no new findings.
- `vitest run` — react **132 files / 2276 tests**, backend.ai-ui **77 files / 2042 tests** (1 skipped), root **7 files / 178 tests**. `BAIFlex.test.tsx` now runs against the real Astryx resolver (its `vi.mock('../theme-shim')` mocked a module this PR deletes). New: `react/src/helper/themeJson.test.ts` (ajv: theme.json ↔ schema ↔ loader), `packages/backend.ai-ui/src/components/Table/BAITable.pxStringWidth.test.tsx` (px-string column widths), rewritten `customThemeConfig` / `backendAiTheme` suites; `antdColors.test.ts` is deleted with the palette generator.
- `build-storybook` → completed (decorators + brand theme on the shared BUI helpers, v2 `theme.json` copy)
- Live probe (dev server, light + dark, headless Chromium, measured on r26 — r27 changes only the artifact name and the `--color-warning-border-hover` token name, same seeds and same hash `h1ux0ib0`): the app boots on the **prebuilt** theme (`data-astryx-theme="bai-r2x-default-brand-h1ux0ib0"`) and resolves `--color-accent: light-dark(#FF7A00, #DC6B03)` (the document's dark value verbatim, not antd's `#be5e06`), `--color-link: light-dark(#FF7A00, #DC6B03)`, `--color-info: light-dark(#028DF2, #009BDD)`, `--color-error: light-dark(#FF4D4F, #DC4446)`, `--color-success: light-dark(#00BD9B, #03A487)`, `--color-warning: light-dark(#FAAD14, #FAAD14)`, `--header-bg: light-dark(#FF9729, #E88A28)`, `--color-accent-muted: light-dark(#FF7A0033, #DC6B033F)`, `--primary-5` as the `color-mix()` of the accent at 85%/65% over the surface, the `--color-text-*` alpha ramp and `--spacing-5: 20px`; no page/console errors beyond the pre-existing Geist-font CSP notice.
- Mechanics worth knowing when reviewing: `mapping.ts` (deleted) was the codemod's lookup table; every `token('<var>')` in the sweep is the var the shim already resolved that antd name to. The token *values* changed from antd numbers to CSS strings — the hand-fixed sites (calc() margins, parseFloat() props, BAIFlex gap rungs, the hoisted hooks) are listed in the sweep commit.

## Reviewer checklist

- [ ] 1. Start the dev server with a Claude `/color` and confirm the WebUI header band takes that color (`VITE_THEME_HEADER_COLOR` → `--header-bg`).
- [ ] 2. Remove `resources/theme.json` (or serve a v1 / non-v2 document) and confirm the app renders on Astryx's default theme — no Backend.AI colors — with a `[appearance] …` console error naming the reason.
- [ ] 3. Change `default.seeds.accent` in `resources/theme.json`, hard-reload, and confirm the first painted screen already carries the new color — Astryx's default theme never shows before `theme.json` loads.
- [ ] 4. Switch to dark mode and confirm the accent is the document's dark value `#DC6B03` (not antd's `#be5e06`).

## Stack

Restructured from single-PR #8860: this PR (document + shim retirement) → #8860 (serve from app config + bootstrap) → #9362 (userConfig family + Branding Apply).


ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3605-pr9361-convert-theme-json.sungchul.10-82-0-159.sslip.io)

yomybaby (changes requested):
Please check a field action. https://github.com/lablup/backend.ai-webui/actions/runs/33721727314/job/100549985759?pr=9361

agatha197 (changes requested):
The default header color was changed to red. Please revert default color.

agatha197:
Dev server on box `sujin`: stopped (was http://fr-3605-pr9361-convert-theme-json.sujin.10-82-0-159.sslip.io)

agatha197 (changes requested):
I just change `Light Mode Logo` but both `About Backend.AI Modal Logo (dark mode)`, `Login Logo (dark mode)` are changed too.

react/src/components/AdminDeploymentPresetSettingPageContent.tsx
```diff
@@ -916,7 +917,7 @@ const AdminDeploymentPresetSettingPageContent: React.FC<
                   <div
                     style={{
                       // Bottom gap comes from the component itself.
-                      marginTop: -token.margin,
+                      marginTop: -token('--spacing-4'),
```
agatha197: `-token('--spacing-4')` is `-'16px'` → `NaN`; React drops the style with a warning, so the -16px pull-up is lost (tsc accepts unary minus on a string, so verify.sh cannot catch it). Use `` `-${token('--spacing-4')}` `` or `calc(-1 * …)`. Same at line 958.

react/src/components/StorageStatusPanelCard.tsx
```diff
@@ -205,7 +205,7 @@ const StorageStatusPanelCard: React.FC<StorageStatusPanelProps> = ({
                     // invitation semantics are restated explicitly (the
                     // per-site colour decision BAIBadgeCount documents).
                     variant="error"
-                    offset={[-token.sizeXS, -token.sizeXS]}
+                    offset={[-token('--spacing-2'), -token('--spacing-2')]}
```
agatha197: `-token('--spacing-2')` is `NaN` here too; `BAIBadgeCount` writes `${offset[0]}px` (BAIBadgeCount.tsx:118), so the invitation pill gets `NaNpx` and loses its -8px offset. `-parseFloat(token('--spacing-2'))`.

react/src/pages/ModelStoreListPageV2.tsx
```diff
@@ -165,15 +168,15 @@ const ModelCardV2Card: React.FC<{
...
+                fontSize: token('--font-size-sm'),
                 display: 'inline-flex',
                 alignItems: 'center',
                 gap: 4,
               }}
             >
               <AuthorIcon
                 author={modelCard.metadata.author}
-                size={token.fontSizeSM}
+                size={parseFloat(token('--font-size-sm'))}
```
agatha197: `--font-size-sm` is `0.75rem` in the built theme (only `--font-size-lg/3xl/4xl` are pinned to px), so `parseFloat` yields 0.75 and `AuthorIcon` renders at 0.75px instead of 12. The same rem scale also reaches lucide `size=` verbatim at DeploymentRevisionHistoryTab.tsx:481/695 and DiagnosticResultList.tsx:81 — `size="1em"` like the sibling icons is the safe form there.

react/src/pages/SessionLauncherPage.tsx
```diff
@@ -1703,14 +1704,15 @@ const unifiedChipStyles = stylex.create({
...
+  const { token } = useTheme();
   // The description lives only in the backend slot-details response, not in
   // the local device_metadata.json, and is not scoped to a resource group.
   const { mergedResourceSlots } = useResourceSlotsDetails();
   const description = mergedResourceSlots[type]?.description ?? type;
   // One line of the description text, so the icon can be vertically centered
   // against the first line (not the whole wrapped block).
-  const lineHeightPx = token.fontSize * token.lineHeight;
+  const lineHeightPx =
+    parseFloat(token('--font-size-base')) * 1.5714285714285714;
```
agatha197: `--font-size-base` is `0.875rem`, so `lineHeightPx` becomes 1.375 instead of 22 and the icon box at line 1730 collapses. Either compute from a px value or hand the CSS string through (`calc(${token('--font-size-base')} * 1.5714)`).

react/src/pages/ChatPage.tsx
```diff
@@ -187,11 +187,11 @@ const ChatHistoryDrawer = ({
             },
             {
               key: 'actions',
-              width: token.sizeXXL,
+              width: token('--spacing-12'),
```
agatha197, line 190: `width` is now the string `'48px'`; BAITable only pins numeric widths (BAITable.tsx:776-779), so the action column flexes instead of staying 48px. `parseFloat(token('--spacing-12'))`.

react/src/astryx-theme/AstryxBrandTheme.tsx
```diff
@@ -34,9 +34,13 @@ const AstryxBrandTheme: React.FC<AstryxBrandThemeProps> = ({
   children,
 }) => {
   'use memo';
-  const { themeConfig, activeThemeFamily } = useCustomThemeConfig();
+  const { rawThemeConfig, activeThemeFamily } = useCustomThemeConfig();
```
agatha197, line 37: With the settle gate gone this is the only reader of the document, and `useRawCustomThemeConfig` snapshots `getCustomTheme()` during render but subscribes to `custom-theme-loaded` in a passive effect — theme.json is in flight during the first (concurrent, yielding) render, so a document that lands between the two is missed for the session, and this provider sits above the Suspense boundary so it never remounts. Before this PR a miss was invisible (the code carried the brand seeds); now it is a neutral grey app. A `useSyncExternalStore` read of the store — what the removed gate used — closes the window.

react/src/helper/customThemeConfig.ts
```diff
@@ -58,62 +74,107 @@ export type LogoConfig = {
...
+ * into and falls back to).
  */
-export function pickValidThemeFamilies(
+export const pickValidAppearanceConfig = (
   input: unknown,
-): Record<string, ThemeFamilyConfig> | undefined {
+  source: string,
+): BAIAppearanceConfig | undefined => {
   if (!_.isPlainObject(input)) {
+    warn(`${source} is not a JSON object; using the built-in defaults.`);
```
agatha197: "using the built-in defaults" (also line 154 "falls back to the built-in seeds", and the `DEFAULT_THEME_FAMILY` comment in useCustomThemeConfig.tsx:18): there are no built-in brand values any more — the operator gets Astryx's neutral theme with no Backend.AI colors. Say that, since checklist item 2 sends them to this message.

react/src/components/BrandingSettingItems/ThemeJsonConfigModal.tsx
```diff
@@ -167,7 +168,17 @@ const ThemeJsonConfigModal: React.FC<ThemeJsonConfigModalProps> = ({
                   message.error(t('theme.CannotApplyInvalidJsonConfig'));
                   return;
                 }
-                setDefaultTheme(parsedValue);
+                // The Monaco markers are advisory (the schema fetch can fail);
+                // this is the gate every draft passes before it is stored.
+                const validated = pickValidAppearanceConfig(
```
agatha197, line 173: `pickValidAppearanceConfig` checks only `schemaVersion` and the `default` family, not the schema, so the description's "validates against theme.schema.json at the write boundary" overstates it: unknown keys and non-hex seeds pass whenever the Monaco schema fetch fails. Either run ajv here (already a devDependency via themeJson.test.ts) or fix the description.

react/src/astryx-theme/built/index.ts
```diff
@@ -21,4 +21,4 @@ import './backendai-default-built.css';
       drift apart (built name ≠ derived default name), and
       `scripts/verify.sh` runs the CLI's `--check` for artifact staleness.
  */
-export { baiR22DefaultBrandHg6tffsTheme as builtBackendAiBrandTheme } from './bai-r22-default-brand-hg6tffs';
+export { baiR26DefaultBrandH1ux0ib0Theme as builtBackendAiBrandTheme } from './bai-r26-default-brand-h1ux0ib0';
```
agatha197, line 24: The description's Verification block predates the verbatim-dark commit: it cites `bai-r25` / REV 23 and probes `--color-accent: light-dark(#FF7A00, #be5e06)`, while the artifact exported here is r26 and the dev server resolves `light-dark(#FF7A00, #DC6B03)`. Please refresh it so checklist item 4 can be read against it.

react/src/components/BrandingSettingList.tsx
```diff
@@ -72,63 +68,66 @@ const BrandingSettingList: React.FC<BrandingSettingListProps> = () => {
...
-            resetColorThemeConfig('token.colorSuccess');
-          },
-        },
-        {
-          type: 'custom',
-          title: t('userSettings.theme.TextColor'),
-          description: t('userSettings.theme.TextColorDesc'),
-          children: <ThemeColorPicker tokenName="token.colorText" />,
+          children: (
+            <ThemeColorPicker seedPath="theme.families.default.seeds.success" />
```
agatha197, line 127: The Text Color picker is removed but the manual still lists it (packages/backend.ai-webui-docs/src/en/admin_menu/admin_menu.md:1977, ko :1658), and react/README.md:40 still sends operators to the antd theme editor for theme.json. Both need the v2 update alongside the migration guide this PR calls for.

agatha197 (changes requested):
Read the loader, recipe and Branding editor directly and swept all 246 rewritten files: every `token()` name resolves, hook order is clean (TextHighlighter/ChatMessageContent even fix old violations) and the replacements match the retired `mapping.ts`. The inline threads are what should change before merge.

react/src/components/AdminDeploymentPresetSettingPageContent.tsx
ironAiken2: Fixed in `96c4bb19c` — both pull-ups now use `calc(-1 * ${token('--spacing-4')})` over the token string (lines 920 and 958).

react/src/components/StorageStatusPanelCard.tsx
ironAiken2: Fixed in `96c4bb19c` — `-parseFloat(token('--spacing-2'))`, so `BAIBadgeCount` gets the numeric -8 it appends `px` to. Spacing tokens are px-valued, so this is exact.

react/src/pages/ModelStoreListPageV2.tsx
ironAiken2: Fixed in `96c4bb19c` — `AuthorIcon` only forwards `size` to lucide, which takes `string | number` and SVG accepts rem, so the prop is widened and `ModelStoreListPageV2` hands the `--font-size-sm` token through instead of `parseFloat` (0.75px). The lucide `size={token(...)}` reads in `DeploymentRevisionHistoryTab` and `DiagnosticResultList` are kept on purpose for the same reason: the rem string goes straight to the SVG `width`/`height`, which renders correctly, and it keeps the icon on the theme scale rather than the surrounding font size. Only the sites that consume a number (`parseFloat`, arithmetic, `BAITable` widths) were converted.

react/src/pages/SessionLauncherPage.tsx
ironAiken2: Fixed in `96c4bb19c` — the one-line icon box no longer `parseFloat`s a rem size against a hard-coded 1.5714; it is `calc(var(--text-body-size) * var(--text-body-leading))`, the same value read from the theme.

react/src/pages/ChatPage.tsx
ironAiken2: Fixed in `96c4bb19c` — `BAITable` now accepts a `<n>px` string width (what a spacing token reads as) and pins it like a number; `BAITable.pxStringWidth.test.tsx` covers both the px string and a non-px string, so the ChatPage action column stays 48px.

react/src/astryx-theme/AstryxBrandTheme.tsx
ironAiken2: Fixed in `9d37468bd` — `useRawCustomThemeConfig` reads the store through `useSyncExternalStore(subscribeCustomTheme, getCustomTheme)`; React re-reads the snapshot right after subscribing, so a document that settles between the first render and the subscription is no longer missed. The paint gate is not restored (no null return, no store flag).

Reproduced first on the dev server by delaying `/resources/theme.json`: 30–50 ms landed inside the window and rendered the neutral theme (accent `#262626`, runtime theme `…-h2vtk6a`); 0–20 ms landed before the render and 60 ms+ after the subscription, both fine. After the change the same sweep renders the brand theme at every delay.

The preview-mode `storage` listener also moved to a plain `useEffect` that returns its cleanup — the previous `useEffectEvent` wrapper discarded it.

react/src/helper/customThemeConfig.ts
ironAiken2: Fixed in `9d37468bd` — the three loader messages and the `DEFAULT_THEME_FAMILY` comment now say "rendering Astryx's neutral theme (no Backend.AI colors)" instead of "built-in defaults" / "built-in seeds".

react/src/components/BrandingSettingItems/ThemeJsonConfigModal.tsx
ironAiken2: Description fixed, code unchanged. The sentence now says Monaco validates against `theme.schema.json` (markers; OK refuses a draft with markers) and that the write boundary (`pickValidAppearanceConfig`) is a structural check only — `schemaVersion: 2` and a `default` family — so a failed schema fetch still cannot bypass the v1 rejection, but unknown keys and non-hex seeds are not caught there. Full schema enforcement at the write boundary is deferred to the app-config save in #9362, where the document first leaves the browser.

react/src/astryx-theme/built/index.ts
ironAiken2: Description refreshed — the Verification block now cites `bai-r26-default-brand-h1ux0ib0` / `THEME_NAME_REV` 26 (and "22 → 26" in the pipeline paragraph), the live probe values re-measured on the current dev server (`--color-accent: light-dark(#FF7A00, #DC6B03)`, `--color-info: light-dark(#028DF2, #009BDD)`, error/success/warning/link/accent-muted, `--primary-5` as the `color-mix()` expression), and the current vitest counts (react 122 files / 1947, backend.ai-ui 75 / 1958, root 5 / 112; `antdColors.test.ts` is gone with the palette generator, `BAITable.pxStringWidth.test.tsx` is new). Checklist item 4 can be read against it now.

react/src/components/BrandingSettingList.tsx
ironAiken2: Fixed in `14537d8d6` — the Text Color item is dropped from the Branding section of the manual (en and ko), and `react/README.md` now points at the v2 `{ theme, branding }` document, `resources/theme.schema.json` and the Branding page JSON Config modal instead of the antd theme editor.

ironAiken2:
Addressed the review-body feedback that had no inline thread:

- **CI (@yomybaby)** — the failing run was `Vitest (react)` + `react-lint` on the 2026-09-03 push. The PR had since become conflicting with `main` (`BulkCreateUserFromCSVModal.tsx`, FR-3854 vs. the token sweep), which is why no `pull_request` workflow ran on the later pushes; the stack is now rebased onto `main` (`fdf1e0736`) and every workflow is green on the current head `5237c379d` — Vitest (react / backend.ai-ui / root / agent-cli), Typecheck, docs-checks, PR Preview, terminology-content, Agent mappings. Locally: react 122 files / 1947 tests, backend.ai-ui 75 / 1958, root 5 / 112, and `bash scripts/verify.sh` passes.
- **Default header color turned red (@agatha197)** — that was the dev-server header override (`VITE_THEME_HEADER_COLOR`) leaking into what looked like the default; `52d078672` renames it for what it is and it only applies on a dev server that sets it. The shipped `default.headerBg` is unchanged (`#FF9729` / `#E88A28`).
- **Changing Light Mode Logo also changed the dark About/Login logos (@agatha197)** — landed separately as #9427 (FR-3852), which this branch includes: a login/About item now shows which logo it inherits instead of silently following the light one.

All ten inline threads are answered and resolved; the two description findings (validation wording, stale Verification block) are fixed in the PR body. #8860 and #9362 are rebased on top and pushed with it.

agatha197 (commented):
.

agatha197 (commented):
Cross-checked the 541 `token.X` -> `token('--y')` replacements against the deleted `mapping.ts` (0 mismatches), ran `astryx-token-gate --strict` (no new findings vs main), and the theme/i18n test suites pass. The v2 design — seed ownership fully in the document, no brand values left in code — is solid.

Two issues live in files this PR doesn't touch, so they can't be inline comments:

- **`packages/backend.ai-ui/README.md:17,56,60-61`** still describes `theme-shim` files (`tokenType.ts`, `vendor/antdColors.ts`, `themeShim.test.ts`) as live architecture — all deleted here. Please give it the same pass `react/README.md` got.
- **`packages/backend.ai-ui/src/components/BAIFlex.test.tsx:9`** — `vi.mock('../theme-shim', ...)` now mocks a module this PR deletes, so the mock is inert: exactly the failure mode its own comment warns about. Drop it and use the `defineTheme` + `<Theme>` pattern from `BAINameActionCell.test.tsx`.

Minor, also outside the diff: `MainLayout.tsx:77` / `WebUISider.tsx:65,82` comments still say `sider.theme` (the key is `theme.siderMode` now); `.storybook/theme.json` is a hand copy of `resources/theme.json` with nothing keeping them in sync — consider wiring it into `themeJson.test.ts`; the About-modal default logo path change (`/manifest/backend.ai-white-text.svg` -> `backend.ai-webui-black/white.svg`) looks like a broken-path fix but isn't mentioned in the PR body; the four prettier-only files (`bulkUserCSV.ts`, `resultTypes.test.ts`, `useAIAgent.ts`, `useWebUIConfig.ts`) are worth one line in the description.

packages/backend.ai-ui/src/form-engine/FormItemVisual.css
```diff
@@ -32,7 +32,7 @@
  for the JS side.
 */
 :where(:root) {
-  --bai-color-text-description: light-dark(
+  --color-text-description: light-dark(
```
agatha197: Dropping the `--bai-` prefix makes this file declare the Astryx-owned `--color-text-description` in `:root` (against the token rules) and duplicates `baiCustomTokens.ts:55` — the theme declares it now, so this block (and the stale `theme-shim/selfTokens.ts` reference above) can be removed.

packages/backend.ai-ui/.storybook/theme.json
```diff
@@ -1,64 +1,94 @@
...
-      "screenXXLMin": 1920,
-      "screenXXL": 1920,
-      "screenXXLMax": 2559,
-      "screenXXXLMin": 2560,
-      "screenXXXL": 2560
-    },
-    "components": {
-      "Tag": {
-        "borderRadiusSM": 1
+  "$schema": "./theme.schema.json",
```
agatha197, line 2: `./theme.schema.json` doesn't exist in `.storybook/` — point at `../../../resources/theme.schema.json` so editor validation actually runs.

packages/backend.ai-ui/.storybook/themeConfig.ts
```diff
@@ -1,79 +1,20 @@
...
-    theme: neutralTheme,
-    light: ASTRYX_NEUTRAL_SEEDS.light,
-    dark: ASTRYX_NEUTRAL_SEEDS.dark,
-  },
-  webui: {
-    theme: astryxBrandTheme,
-    light: webuiSeeds('light'),
-    dark: webuiSeeds('dark'),
-  },
+  astryx: { theme: neutralTheme },
```
agatha197: Bare `neutralTheme` has none of the BAI custom tokens (`--color-info`, `--color-error-bg`, `--color-text-tertiary`, `--preset-*-5`), so this preset drops colors in `BAINameActionCell` / `BAIAgentTable` / `useSemanticColorMap` — `defineTheme({ extends: neutralTheme, tokens: buildBaiCustomTokens({}) })` restores the vocabulary with Astryx fallbacks.

react/src/components/FolderExplorerModal.tsx
```diff
@@ -260,7 +261,7 @@ const FolderExplorerModal: React.FC<FolderExplorerProps> = ({
         vfolder_node ? (
           <FolderExplorerHeader
             titleStyle={{
-              zIndex: token.zIndexPopupBase + 2,
+              zIndex: 1000 + 2,
```
agatha197: `zIndexLadder.ts:15` documents this literal as a band the ladder doesn't own — add a `legacyPopupBase: 1000` rung to `BAI_Z_INDEX` and read `BAI_Z_INDEX.legacyPopupBase + 2` instead.

react/src/components/SessionMetricGraph.tsx
```diff
@@ -246,21 +249,21 @@ const SessionMetricGraphBody: React.FC<PrometheusMetricGraphProps> = ({
                 ).number
               }
               label="Avg Used"
-              stroke={token.red}
+              stroke={'#f5222d'}
```
agatha197: Raw hex inlined from `token.red` — hoist to a named constant (like `CHART_COLORS` elsewhere) or add `--preset-red-6` to `baiCustomTokens.ts` and read it via `token()`.

packages/backend.ai-ui/src/theme/baiCustomTokens.ts
```diff
@@ -0,0 +1,107 @@
...
+  '--header-bg': seeds.headerBg
+    ? [seeds.headerBg.light, seeds.headerBg.dark]
+    : 'var(--color-background-surface)',
+  '--color-error-bg': seeds.error
+    ? tint('--color-error', 10, 15)
+    : 'var(--color-error-muted)',
+  '--color-info-bg': seeds.info
+    ? tint('--color-info', 10, 15)
+    : 'var(--color-accent-muted)',
+  '--color-warning-hover': seeds.warning
```
agatha197: `colorWarningBorderHover` and `colorSuccessBorderHover` are the same antd concept (palette step 4), but only this one drops `-border-` — suggest `--color-warning-border-hover` to match line 94.

react/src/helper/customThemeConfig.ts
```diff
@@ -58,62 +74,122 @@ export type LogoConfig = {
...
+      return light;
+    }
+    return _.isString(value[1]) ? value[1] : light;
+  }
+  return undefined;
+};
 
-export const getCustomTheme = () => _customTheme;
+// eslint-disable-next-line no-console -- module-scope diagnostics; no logger exists outside React here
+const warn = (message: string) => console.error(`[appearance] ${message}`);
```
agatha197: A `warn` that calls `console.error` — rename to something like `logAppearanceError` (the PR body says "loud console error") or downgrade to `console.warn`.

react/src/helper/customThemeConfig.ts
```diff
@@ -58,62 +74,122 @@ export type LogoConfig = {
...
+        `${source} has no "schemaVersion": ${APPEARANCE_SCHEMA_VERSION} ` +
+          `(got ${JSON.stringify(doc.schemaVersion)}); rendering Astryx's ` +
+          'neutral theme (no Backend.AI colors).',
+      );
+    }
+    return undefined;
+  }
+  const families = (doc.theme as Record<string, unknown> | undefined)?.families;
+  if (
+    _.isPlainObject(families) &&
```
agatha197: A v2 doc with no `theme` block at all (`{ "schemaVersion": 2 }`) passes with no warning and renders brand-less — the exact "why are the colors gone?" case this function exists for; warn whenever `theme.families.default` is missing, not only when `families` is a plain object.

react/src/components/BrandingSettingItems/ThemeJsonConfigModal.tsx
```diff
@@ -167,7 +168,17 @@ const ThemeJsonConfigModal: React.FC<ThemeJsonConfigModalProps> = ({
...
+                  parsedValue,
+                  'the JSON editor',
+                );
+                if (!validated) {
```
agatha197, line 177: The real rejection reason (v1 document -> migration guide) only reaches the console; the admin sees a generic "invalid JSON" toast — have `pickValidAppearanceConfig` return the reason so the toast can carry it (fine to fold into #9362).

react/src/components/AnnouncementEditModal.tsx
```diff
@@ -247,12 +247,12 @@ const AnnouncementEditModal: React.FC<AnnouncementEditModalProps> = ({
...
-                border: `1px solid ${token.colorBorder}`,
-                borderRadius: token.borderRadius,
-                padding: token.paddingLG,
+                border: `1px solid ${token('--color-border-emphasized')}`,
+                borderRadius: token('--radius-inner'),
+                padding: token('--spacing-6'),
                 // Match the editor's outer height (its inner height + the
                 // toolbar bar and the editor wrapper's borders).
-                height: `calc(${EDITOR_HEIGHT} + ${token.controlHeightSM + 2}px)`,
+                height: `calc(${EDITOR_HEIGHT} + ${24 + 2}px)`,
```
agatha197: `24 + 2` is `controlHeightSM + 2` — use `--size-element-sm`, like the sweep already does with `--size-element-md` elsewhere.

react/src/components/SessionFormItems/SharedMemoryFormItems.tsx
```diff
@@ -135,15 +135,15 @@ const SharedMemoryFormItems: React.FC<SharedMemoryFormItemsProps> = ({
               justify="between"
               wrap="wrap"
               style={{
-                minHeight: token.controlHeightSM,
+                minHeight: 24,
```
agatha197: `24` is `controlHeightSM` — `--size-element-sm` instead of the literal.

react/src/components/MainLayout/MainLayout.tsx
```diff
@@ -142,7 +142,7 @@ function MainLayout() {
     };
   }, [navigate]);
 
-  const headerHeight = Number(token.Layout?.headerHeight) || 60;
+  const headerHeight = 60;
```
agatha197: Third copy of the header height (also `:442` and `WebUIHeader.tsx`) — one `HEADER_HEIGHT` constant, ideally feeding the `--headerHeight` CSS var `CSSTokenVariables` already emits.

react/src/components/Chat/ChatMessageContent.tsx
```diff
@@ -170,80 +170,88 @@ const ChatMessageContentBlock = memo<{ block?: string; isStreaming?: boolean }>(
...
+            }}
+          >
+            <table
+              {...props}
+              className="chat-markdown-table"
+              style={{
+                width: '100%',
+                borderCollapse: 'collapse',
+                fontSize: token('--font-size-sm'),
+                lineHeight: 1.5714285714285714,
```
agatha197: `1.5714285714285714` is `token.lineHeight` — `--text-body-leading` is already used further down this file.

react/src/components/ImageList.tsx
```diff
@@ -343,7 +341,7 @@ const ImageListInScope: React.FC<ImageListInScopeProps> = ({
         </BAIText>
       ),
       // Computed (`getImageFullName`) — not orderable on the server.
-      width: token.screenXS,
+      width: 480,
```
agatha197: `480` is antd's `screenXS`, duplicated in `CustomizedImageList.tsx` — hoist to a shared constant.

packages/backend.ai-ui/src/theme/index.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+
+ The antd-parity data the Astryx theme recipe (`react/src/astryx-theme`) is
+ built from: the vendored `@ant-design/colors` palette generator and the
```
agatha197: This header still cites "the vendored `@ant-design/colors` palette generator" — deleted in this PR; only the parity table and the custom-token builder remain.

react/src/helper/logoSource.ts
```diff
@@ -0,0 +1,65 @@
...
+  login: { light: 'loginLogoSrc', dark: 'loginLogoSrcDark' },
+  about: { light: 'aboutLogoSrc', dark: 'aboutLogoSrcDark' },
+} as const satisfies Record<
+  DetailLogoArea,
+  Record<ThemeScheme, keyof LogoConfig>
+>;
+
+export const DEFAULT_DETAIL_LOGO_SRC = {
+  login: {
+    light: 'manifest/backend.ai-text.svg',
```
agatha197, line 30: `login` paths are relative while `about` paths are absolute — carried over as-is, but they diverge under sub-path deployments; worth unifying.

e2e/utils/test-util.ts
```diff
@@ -1406,41 +1410,31 @@ export async function modifyThemeJson(
...
+        theme: {
+          fontFamily: "'Ubuntu', Roboto, sans-serif",
+          families: {
+            default: {
+              seeds: {
+                accent: ['#FF7A00', '#DC6B03'],
+                link: ['#FF7A00', '#DC6B03'],
+                info: ['#028DF2', '#009BDD'],
+                error: ['#FF4D4F', '#DC4446'],
+                success: ['#00BD9B', '#03A487'],
```
agatha197, line 1427: This fallback omits the `warning` seed that `resources/theme.json` now declares — under this PR's "code fills in nothing the document omits" contract, tests hitting the fallback get Astryx's default warning color.

packages/backend.ai-ui/.storybook/theme.json
ironAiken2: Fixed in `7e00c04d1` — `.storybook/theme.schema.json` now exists, so the `$schema` pointer resolves and editor validation runs. It is deliberately a BUI-local schema describing only what `astryxBrandTheme.ts` reads (the default family's seeds, `headerBg`, `fontFamily`; everything else passes through) rather than a pointer at `resources/theme.schema.json`: backend.ai-ui must not reference files of the host repository so the package stays splittable into its own repo. The same reason is why `.storybook/theme.json` is not wired into `themeJson.test.ts` — a copy the package owns, checked against the package's own reader contract. Noted in the `astryxBrandTheme.ts` header and the PR body's Storybook section.

packages/backend.ai-ui/src/theme/baiCustomTokens.ts
ironAiken2: Fixed in `7e00c04d1` — renamed to `--color-warning-border-hover` (builder + the three consumers: `BAITextHighlighter`, `TextHighlighter`, `SharedMemoryFormItems`). The rename changes the built CSS, so `THEME_NAME_REV` is bumped 26 → 27 and the prebuilt artifact is `bai-r27-default-brand-h1ux0ib0` (same seeds, same hash); the PR body's pipeline paragraph and Verification block are updated accordingly.

react/src/helper/customThemeConfig.ts
ironAiken2: Fixed in `7e00c04d1` — renamed to `logAppearanceError`; it keeps `console.error`, since a brand-less app is the loud case the PR body describes.

react/src/helper/customThemeConfig.ts
ironAiken2: Fixed in `7e00c04d1` — the check is now `theme.families.default` is missing → log, regardless of whether `theme` or `families` exist. `{ "schemaVersion": 2 }`, `{ schemaVersion: 2, theme: {} }` and a `families` without `default` all log the same `[appearance] … has no theme.families.default entry; the default family is rendering Astryx's neutral theme` error; `customThemeConfig.test.ts` covers the three shapes and asserts the shipped document logs nothing.

react/src/components/BrandingSettingItems/ThemeJsonConfigModal.tsx
ironAiken2: Fixed in `7e00c04d1`, with a different shape than suggested. Decision (with @sungchulhong): the loader no longer sniffs v1 documents at all — any document that is not `schemaVersion: 2` is one rejection, "does not match the theme schema (resources/theme.schema.json)", and that is also what the admin already sees: the modal's toast, `theme.CannotApplyInvalidJsonConfig`, reads "Theme settings do not match the schema. Please check the file." So the console and the toast now say the same thing, and there is no separate reason left to thread through the return value. The migration-guide pointer stays in the schema's `schemaVersion` description and the PR body. The v1 wording in the PR body (Breaking section, Branding editor) is updated to match.

react/src/components/AnnouncementEditModal.tsx
ironAiken2: Fixed in `7e00c04d1` — `calc(${EDITOR_HEIGHT} + ${token('--size-element-sm')} + 2px)`.

react/src/components/SessionFormItems/SharedMemoryFormItems.tsx
ironAiken2: Fixed in `7e00c04d1` — `minHeight: token('--size-element-sm')`.

react/src/components/MainLayout/MainLayout.tsx
ironAiken2: Fixed in `7e00c04d1` — without a new constant: the two inline sites (the Suspense fallback here and `WebUIHeader`'s band) now read `var(--webui-header-height)`, the variable `CSSTokenVariables` already emits, the way `AnnouncementBanner.css` / `BAISider.css` / `WebUIHeader.css` already do. The `'60px'` in `CSSTokenVariables` is the one remaining literal and the single source. (The `60px` fallbacks inside the CSS files' `var(--webui-header-height, 60px)` are pre-existing and left alone.)

ironAiken2:
Addressed the 2026-09-08 review (@agatha197) in `7e00c04d1`; the eleven inline threads are answered individually. The findings that had no thread:

- **`packages/backend.ai-ui/README.md`** — the peer-dependency row and the "no antd surface" section no longer describe `theme-shim` files (`tokenType.ts`, `vendor/antdColors.ts`, `themeShim.test.ts`) as live; they now say what remains under `src/theme/` (alignment tables + custom-token builder) and that no antd-family package is left, not even in `devDependencies`.
- **`BAIFlex.test.tsx`** — the inert `vi.mock('../theme-shim')` is gone. It does not need the `defineTheme` + `<Theme>` pattern either: `--spacing-*` are core Astryx tokens that `useTheme().token()` resolves from its JS `tokenDefaults` (unlike `--color-info`, a BUI custom token, in `BAINameActionCell.test.tsx`), so the suite runs against the real resolver with nothing mocked, and the gap-rung table still guards every rung.
- **`MainLayout.tsx` / `WebUISider.tsx` comments** now say `theme.siderMode`.
- **`.storybook/theme.json` ↔ `themeJson.test.ts` sync** — not wired, on purpose: backend.ai-ui must not reference the host repository's `resources/` (the package should stay splittable into its own repo), so the copy is validated against a BUI-local `.storybook/theme.schema.json` that describes exactly what `astryxBrandTheme.ts` reads. Same reasoning on the `.storybook/theme.json:2` thread.
- **About-modal default logo path** and the **four prettier-only files** are now recorded in the PR body ("The sweep, mechanically"); the old About default `/manifest/backend.ai-white-text.svg` pointed at a file that does not exist.

Also from the threads: `THEME_NAME_REV` 26 → 27 for the `--color-warning-border-hover` rename (artifact `bai-r27-default-brand-h1ux0ib0`, same seeds/hash), and the loader no longer sniffs v1 documents — every non-`schemaVersion: 2` document is reported as a schema mismatch, which is what the Branding modal's toast already says.

Verification: `bash scripts/verify.sh` → `=== ALL PASS ===`; vitest react 132 files / 2276 tests, backend.ai-ui 77 / 2042 (1 skipped; `BAIComplexSelect.popup.test.tsx` flaked once in the full run and passes in isolation, untouched here), root 7 / 178; `astryx-token-gate --strict` reports only the two pre-existing `--bai-dialog-z` findings.

Stack: #8860 (`feat/FR-1964-app-config-hooks`) and #9362 (`feat/FR-3834-theme-family-branding`) are rebased onto this head and pushed (clean rebases). `feat/FR-3877-announcement-app-config` above them is checked out in another worktree and was left as is; it needs a rebase before it is reviewed.

packages/backend.ai-ui/src/theme/index.ts
ironAiken2: Fixed in `7e00c04d1` — the header now names what is actually left: the measured alignment tables and the custom-token builder.

react/src/helper/logoSource.ts
ironAiken2: Fixed in `7e00c04d1` — the About defaults are relative (`manifest/backend.ai-webui-black.svg` / `-white.svg`) like the login ones, so both resolve the same way under a sub-path deployment. These are the last-rung built-in fallbacks (own logo → opposite-scheme sider logo → these), so nothing operator-configured is affected. The PR body now also records that the previous About default (`/manifest/backend.ai-white-text.svg`) pointed at a file that does not exist.

e2e/utils/test-util.ts
ironAiken2: Fixed in `7e00c04d1` — the fallback declares `warning: ['#FAAD14', '#FAAD14']`, matching `resources/theme.json`.

PR #9362, opened 2026-09-01
feat(FR-3834): move theme family to userConfig and save Branding edits to app config

### LINKED ISSUE #9356: Move theme family selection to userConfig and save Branding edits to app config
### Summary

Top PR of the app-config stack (base: FR-1964's app-config serving, which sits on FR-3605's v2 document): move theme family selection to userConfig and make the Branding editor save to app config. PR: #9362.

### Scope

- **Theme family → userConfig** — `userConfig.themeFamily` becomes the authoritative store, written from User Settings via a USER-scope raw read + upsert (`useUpdateMyUserAppConfig`). localStorage (`backendaiwebui.settings.themeFamily`) is demoted to the FOUC mirror the `index.html` bootstrap reads: synced on login by `ThemeFamilyUserConfigSync`, cleared on logout.
- **Branding Apply → app config** — Apply saves the whole v2 draft to the `appearance` subKey of `publicConfigByDomain` and reloads the page on success (the reloaded page is the feedback; no refresh API). Preview becomes a secondary button; the "contact support" banner is removed.

**(The `custom_primary_color` / `ThemeAccentColorPicker` removal moved down to FR-3605 — it is coupled to the v1→v2 internal-currency switch, not to the userConfig work.)**

### Acceptance criteria

- Theme family chosen in User Settings follows the user across browsers/devices (userConfig), with no FOUC on reload.
- Superadmin Branding Apply persists the domain document server-side; a plain reload shows the applied appearance.

JIRA Issue: FR-3834

ironAiken2:
Resolves #9356 (FR-3834)

## Summary

Top PR of the app-config stack: the write paths on top of the served v2 appearance document (#9361, #8860).

### Theme family selection → userConfig

- `userConfig.themeFamily` is the authoritative store, written from User Settings by `useUpdateMyUserAppConfig`: the setter re-reads the raw USER-scope fragment with `myAppConfigFragmentsByNames` (never the merged view, because a write replaces one scope's whole document), edits on top of it, and upserts through `myUpsertAppConfigFragments`. The scoped mutation (`scopedUpsertAppConfigFragments`) refuses `userConfig` at USER scope, so the own-user write has to go through the `my*` pair; the PUBLIC write for Branding still uses the scoped one. The localStorage mirror is written only after the server accepts; while the write is in flight the select shows the chosen family with a loading state, and a failed save toasts and falls back to the stored value.
- localStorage (`backendaiwebui.settings.themeFamily`) is demoted to a mirror of the server value: `DismissSplashOnMount` (MainLayout) suspends on `userConfig.themeFamily` and copies it into the mirror before the splash is dismissed, so the account's family is what the page reveals. The mirror survives logout — the signed-out screen keeps the last account's family — and the next login overrides it. A missing or unknown family resolves to the document's `default` family.

### Branding Apply → app config

- **Apply** asks for confirmation (the theme applies to every user in the domain), then saves the whole edited v2 draft to the domain slice's `appearance` subKey of `publicConfigByDomain` and reloads the page on success — the reloaded page renders the applied document, which is the feedback; the anonymous read path has no refresh API.
- The editor draft (what the colour pickers and the JSON modal show) now seeds from the saved domain document, falling back to `theme.json` only when the domain has none. Reset still restores the shipped values.
- Preview moves to a secondary button next to the JSON modal button; the "contact support to apply the exported JSON" banner is removed because Apply now exists. The JSON modal description is reworded accordingly.
- `useAppConfig` gains the read hooks (`useMyAppConfig`, `useDomainAppConfig`) and the two setters (`useUpdatePublicDomainAppConfig`, `useUpdateMyUserAppConfig`). Both setters use BUI's `useMutationWithPromise` (this file was the tree's only direct `commitMutation` caller), which now rejects an `Error` carrying the GraphQL error messages instead of the raw `PayloadError[]`, and still reject on a per-item `failed` entry. The two call sites resolve failures with `useErrorMessageResolver`, so the manager's own message (e.g. an allow-list refusal) reaches the user.

## Backend prerequisites

The app-config write gate is data, not code: every `(config_name, scope_type)` pair the UI writes must exist in the manager's `app_config_allow_list`, or the upsert fails with `backendai_generic_forbidden`. A dev fixture that covers everything this stack touches:

```json
{
  "app_config_definitions": [
    { "config_name": "userConfig" },
    { "config_name": "domainConfig" },
    { "config_name": "publicConfigByDomain" }
  ],
  "app_config_allow_list": [
    { "config_name": "userConfig",           "scope_type": "domain", "rank": 200 },
    { "config_name": "userConfig",           "scope_type": "user",   "rank": 300 },
    { "config_name": "domainConfig",         "scope_type": "domain", "rank": 200 },
    { "config_name": "publicConfigByDomain", "scope_type": "public", "rank": 100 }
  ]
}
```

The read side of Apply needs `general.apiDomainName` in the served `config.toml` (the webserver injects it from `webserver.conf`); without it the bootstrap never fetches the domain document and `theme.json` keeps winning, so a local dev server needs the key added to its own `config.toml`.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (re-run on the top of the stack after the rebase onto `d3ddfa1cb`).
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → no new findings.
- `pnpm exec vitest run` (react) → **109 files / 1717 tests passed**; root suite 111 tests.
- After the splash-sync refactor (`d207f73eb`): `pnpm exec vitest run` on MainLayout / useCustomThemeConfig / useLogout → 3 files / 53 tests passed.
- Review round (`53f69e482`): `bash scripts/verify.sh` → `=== ALL PASS ===`; targeted `vitest run` → react 3 files / 25 tests, BUI 3 files / 20 tests passed.
- Manually on a local manager with the fixture above: User Settings theme-family save round-trips through `myUpsertAppConfigFragments`; Branding Apply persists the domain slice and the reloaded page renders it.

## Stack

#9361 (v2 document) → #8860 (serve from app config + bootstrap) → this PR.


ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3834-pr9362-move-theme-family.sungchul.10-82-0-159.sslip.io)

agatha197 (commented):
Checked the i18n changes across all 21 locales (in sync), the five `graphql` tags against their `__generated__` outputs (all paired), and scope against the PR description (clean). The two findings on `MainLayout.tsx:480` and `useAppConfig.tsx:185` are blocking-level; the rest is cleanup.

Two things I couldn't anchor inline:

- **Migration**: an account that picked a theme family locally before this PR gets it silently reset on next login (server `themeFamily` is empty and overwrites the mirror) — is that intended?
- `useCustomThemeConfig.tsx:28` (untouched here) still says "a server-side `userConfig.themeFamily` store arrives with FR-1964; until then this key is the only store" — this PR makes that false.

react/src/components/MainLayout/MainLayout.tsx
```diff
@@ -465,11 +468,24 @@ export const CSSTokenVariables = () => {
...
+ *
+ * Also suspends on the signed-in user's `userConfig.themeFamily` and copies
+ * it into the localStorage mirror BEFORE the splash goes, so the account's
+ * family is what the page reveals (FR-3834). The mirror survives logout; the
+ * next login overrides it here.
  */
 const DismissSplashOnMount = () => {
   'use memo';
   useSuspendedBackendaiClient();
+  const userConfigFamily = useMyAppConfig<string>('themeFamily');
```
agatha197, line 480: This makes splash dismissal suspend on `myAppConfigs` with no error boundary nearer than the app root — on a pre-26.8.0 manager or an allow-list denial the query rejects, the splash never lifts, and the whole shell dies for a cosmetic read; move the mirror sync into a sibling `<ErrorBoundaryWithNullFallback><Suspense>` (the pattern already used 6x in this file).

react/src/hooks/useAppConfig.ts
```diff
@@ -0,0 +1,247 @@
...
+  const currentDomainName = useCurrentDomainValue();
+  const upsert =
+    useMutationWithPromise<useAppConfigUpsertMutation>(upsertMutation);
+
+  return async (
+    subKey: string | Array<string>,
+    nextValue: unknown,
+    domainName?: string,
+  ) => {
+    const targetDomainName = domainName ?? currentDomainName;
```
agatha197, line 185: This writes to `useCurrentDomainValue()` (the login account's domain) while the anonymous theme loader reads the `general.apiDomainName` slice from config.toml (`customThemeConfig.ts#resolveConfigDomainName`) — when they differ (e.g. superadmin on `default`, apiDomainName `mycompany`), Apply silently writes a slice nobody reads and reloads with no visible change; source the target domain the same way the reader does, or require it explicitly.

react/src/hooks/useAppConfig.ts
```diff
@@ -0,0 +1,247 @@
...
+ * so other domains' slices survive the whole-document write. Post-login admin
+ * surfaces only. The anonymous read path picks the new document up on the
+ * next load — the caller reloads for the own session (FR-1964).
+ */
+export const useUpdatePublicDomainAppConfig = () => {
+  'use memo';
+  const relayEnv = useRelayEnvironment();
+  const currentDomainName = useCurrentDomainValue();
+  const upsert =
+    useMutationWithPromise<useAppConfigUpsertMutation>(upsertMutation);
```
agatha197, line 178: When the mutation comes back with GraphQL `errors`, `useMutationWithPromise` rejects with the raw `PayloadError[]`, which `getErrorMessage` can't unwrap (falls through to "Unknown error") — worth verifying the manager's allow-list message actually reaches the toast (`failed[]` via `throwOnFailed` is fine).

react/src/pages/UserSettingsPage.tsx
```diff
@@ -233,11 +239,21 @@ const UserPreferencesPage = () => {
               onChange: (value: string | number | undefined) => {
                 if (typeof value === 'string') {
                   setThemeFamily(value);
+                  updateMyUserAppConfig('themeFamily', value).catch((error) => {
```
agatha197: Fire-and-forget: if this write fails the mirror keeps the new family, the server keeps the old one, and the next login silently reverts — roll `setThemeFamily` back in the catch; the identical onChange/onReset catch blocks could share one helper.

react/src/components/BrandingSettingList.tsx
```diff
@@ -27,10 +36,34 @@ const BrandingSettingList: React.FC<BrandingSettingListProps> = () => {
...
+      return;
+    }
+    try {
+      await updatePublicDomainAppConfig(DOMAIN_APPEARANCE_CONFIG_KEY, {
+        ...defaultTheme,
+        schemaVersion: APPEARANCE_SCHEMA_VERSION,
+      });
+      // The reloaded page renders the applied document — that IS the
+      // feedback; the anonymous read path has no refresh API (FR-1964).
+      window.location.reload();
```
agatha197: A domain-wide save with no confirmation and reload-as-the-only-feedback — a `modal.confirm` ("this applies to every user in this domain") before writing would be cheap insurance.

react/src/hooks/useAppConfig.tsx
```diff
@@ -0,0 +1,247 @@
...
+    ?.config as AppConfigDocument | undefined;
+  return (subKey === undefined ? doc : _.get(doc, subKey)) as T | undefined;
+};
+
+/**
+ * The domain-wide config (`domainConfig`) readable by every signed-in user
+ * of the domain (admin-writable only). Read only. Suspends; post-login
+ * surfaces inside `RelayEnvironmentProvider` only.
+ */
+export const useDomainAppConfig = <T = AppConfigDocument,>(
```
agatha197: `useDomainAppConfig`, `AppConfigName` (line 26), and the setter's `domainName` parameter have no consumers anywhere in the tree — consider adding them when a consumer lands.

react/src/hooks/useAppConfig.tsx
```diff
@@ -0,0 +1,247 @@
...
+  ) => {
+    const targetDomainName = domainName ?? currentDomainName;
+    const raw = await fetchQuery<useAppConfigPublicRawQuery>(
+      relayEnv,
+      publicRawQuery,
+      { configNames: ['publicConfigByDomain'] },
+      { fetchPolicy: 'network-only' },
+    ).toPromise();
+    const rawDoc =
+      (raw?.scopedAppConfigFragmentsByNames?.[0]?.config as
```
agatha197: The merged readers use `.find((c) => c.configName === ...)` but both raw reads use `?.[0]` (here and :233) — `[AppConfigFragment]!` elements are nullable, so `.find` is safer and consistent.

react/src/hooks/useAppConfig.ts
```diff
@@ -0,0 +1,247 @@
+/**
```
agatha197, line 1: No JSX here, so `.ts` matches the sibling hooks; also `useWebUIConfig.ts#fetchAppConfigOnce` already uses "app config" to mean config.toml — `useManagerAppConfig` would avoid the collision.

react/src/hooks/useDefaultTheme.ts
```diff
@@ -30,20 +31,19 @@ export const useDefaultTheme = () => {
...
-  // Note: useBAISettingUserState returns null (not undefined) when
-  // localStorage has no value.
+  // Seed the draft from the saved domain document, else the SHIPPED one —
+  // never from `rawThemeConfig`, which in preview mode IS the draft and would
+  // loop. A draft from before the v2 format (no schemaVersion) is reseeded
+  // rather than edited. Note: useBAISettingUserState returns null (not
+  // undefined) when localStorage has no value.
   const initializeDefaultTheme = useEffectEvent(() => {
-    const shipped = getStaticAppearanceConfig();
+    const seed = getDomainAppearanceConfig() ?? getStaticAppearanceConfig();
```
agatha197, line 41: Seed source changed here, but the header JSDoc (line 21) still says "Seeded from the shipped `theme.json`, not from the applied document" — update it to match.

agatha197 (changes requested):
Two blockers, details inline: a failed myAppConfigs query blocks login entirely (MainLayout.tsx:480), and Apply can silently write to a domain slice the theme loader never reads (useAppConfig.tsx:185).

react/src/components/MainLayout/MainLayout.tsx
ironAiken2: Kept the mirror sync where it is. This PR targets the LTS line, where the manager is >= 26.9.0, so the pre-26.9 `myAppConfigs` rejection is out of scope by decision. An allow-list denial is not an error path on this field — the schema documents it as an empty merge — so that case already falls through to the fallback family. Hardened the read to `data?.myAppConfigs?.find(...)` in `53f69e482`; a `data: null` response would still surface through Relay, which we accept for this line.

react/src/hooks/useAppConfig.ts
ironAiken2: By deployment contract one WebUI serving address serves exactly one domain, and every account belongs to one domain, so `general.apiDomainName` and the signed-in account's domain are always the same value — the write and the anonymous read address the same slice. The optional `domainName` argument stays for an explicit override. Keeping the code as is.

react/src/hooks/useAppConfig.ts
ironAiken2: Confirmed: a `data` + `errors` response rejected the raw `PayloadError[]`, which `getErrorMessage` cannot unwrap. Fixed in `53f69e482` — `useMutationWithPromise` now rejects an `Error` whose message joins the GraphQL error messages, so `getErrorMessage` (and the `error?.message` readers) show the manager's text. The `failed[]` path is unchanged.

react/src/pages/UserSettingsPage.tsx
ironAiken2: Fixed in `53f69e482` — the mirror is now written only after the server accepts. While the write is in flight the select shows the chosen family with a loading state (`SettingItem` select gained `loading`); on failure it toasts and falls back to the stored value. `onChange` / `onReset` share one `persistThemeFamily` helper.

react/src/components/BrandingSettingList.tsx
ironAiken2: Fixed in `53f69e482` — Apply now opens a `modal.confirm` ("This theme will be applied to every user in this domain.") and the ok button waits on the save; the two keys are added to all 21 locales.

react/src/hooks/useAppConfig.tsx
ironAiken2: Keeping `useDomainAppConfig`, `AppConfigName` and the `domainName` argument: they describe the three documents this app reads and the explicit-domain write, and their consumers land in the follow-up FRs on this stack. No behaviour depends on them.

react/src/hooks/useAppConfig.ts
ironAiken2: Renamed to `useAppConfig.ts` in `53f69e482`. Keeping the `useAppConfig` name: it mirrors the manager API (`myAppConfigs`, `AppConfigFragment`), which is the vocabulary these hooks wrap; the config.toml fetch in `useWebUIConfig.ts` is the legacy meaning.

react/src/hooks/useDefaultTheme.ts
ironAiken2: Updated in `53f69e482` — the header now reads "Seeded from the saved domain document, else the shipped `theme.json` — never from the applied (preview) document".

ironAiken2:
Addressed the review-body feedback from @agatha197:

- **Migration**: intended. `userConfig.themeFamily` is the authoritative store; an account that had picked a family only locally before this PR is reset once on its next login and picks again from User Settings. Accepted for this line.
- `useCustomThemeConfig.tsx:28`: comment updated in `53f69e482` — the key is now described as the FOUC mirror of the server store.

PR #9365, opened 2026-09-01
fix(FR-3106): neutralize CSV formula injection in the export helpers

### LINKED ISSUE #7842: Neutralize CSV formula injection in CSV export helpers

ironAiken2:
Resolves #7842 (FR-3106)

## What is wrong

Excel and Sheets evaluate a CSV cell whose first character is `=`, `+`, `-`, `@`, tab, or CR as a formula, and RFC 4180 quoting does not prevent it — the parser strips the quotes before interpreting the cell. None of the CSV builders neutralized these cells:

- `escapeCsvValue` in `react/src/helper/csv-util.ts`
- DiagnosticsPage's local `escCsv`
- the keypair row builder in `MyKeypairManagementModal.tsx`

User-controlled data (emails, names, backend export rows) flows into files explicitly intended for spreadsheets, so a crafted value like `=HYPERLINK(...)` executed on open (CSV formula injection).

## How this fixes it

- `escapeCsvValue` now prefixes a dangerous leading character with a single quote (`'`), which spreadsheets render as text. It is exported so all client-side CSV cell escaping goes through the one helper.
- Signed numbers (`-30`, `+1.5`) and the bare `-` placeholder are exempt from the `+`/`-` trigger so exported numeric columns survive unchanged; non-numeric `+`/`-` starts (`-1+2`) are still neutralized.
- DiagnosticsPage's local `escCsv` is deleted and its rows route through `escapeCsvValue`; same for the keypair row builder in `MyKeypairManagementModal.tsx` (which also gains null-safety its inline `v.replace` lacked).
- **Credential cells opt out (review follow-up).** Secret keys are `secrets.token_urlsafe(30)` on the manager, so about 1 in 64 starts with `-`, and the keypair CSVs are pasted into CLI configs rather than opened in a spreadsheet. `csvLiteral()` marks a cell that `escapeCsvValue` quotes but never neutralizes; the two keypair exports (`MyKeypairManagementModal`, `GeneratedKeypairListModal`) wrap only the secret key column. Access keys (`AKIA` prefix) and SSH public keys stay under the default, and any cell not marked is neutralized as before, so the opt-out cannot reach a user-controlled column by accident.
- Vitest cases cover the trigger set, the numeric exemptions, the `csvLiteral` opt-out (including a row where a `-`-leading secret key survives next to a neutralized `=HYPERLINK` cell), and the `JSONToCSVBody` integration path.

## Out of scope

- `POST /export/{nodeKey}/csv` rows arrive pre-serialized from the manager, so the server-side export path is not covered here. That is a backend follow-up.

## Verification

- `pnpm vitest run src/helper/csv-util.test.ts` — 43 passed
- `bash scripts/verify.sh` — `=== ALL PASS ===`

https://claude.ai/code/session_01BWJPUAzCTZvtFeKYVHe1i7
https://claude.ai/code/session_019KwHghmD57U6i2jcGbnx6D

### DIFF
--- a/react/src/components/MyKeypairManagementModal.tsx
+++ b/react/src/components/MyKeypairManagementModal.tsx
@@ -13,7 +13,7 @@ import { MyKeypairManagementModalRevokeMyKeypairMutation } from '../__generated_
 import { MyKeypairManagementModalSwitchMainKeyMutation } from '../__generated__/MyKeypairManagementModalSwitchMainKeyMutation.graphql';
 import { App } from '../app-shim';
 import { convertToOrderBy } from '../helper';
-import { downloadCSV } from '../helper/csv-util';
+import { downloadCSV, escapeCsvValue } from '../helper/csv-util';
 import { useBAIPaginationOptionState } from '../hooks/reactPaginationQueryOptions';
 import { useBAISettingUserState } from '../hooks/useBAISetting';
 import { theme } from '../theme-shim';
@@ -90,7 +90,7 @@ const downloadCredentialCSV = (credential: KeypairCredential) => {
     credential.secretKey,
     credential.sshPublicKey,
   ]
-    .map((v) => `"${v.replace(/"/g, '""')}"`)
+    .map(escapeCsvValue)
     .join(',');
 
   const csvContent = `${header}\n${row}\n`;
--- a/react/src/components/MyKeypairManagementModal.tsx
+++ b/react/src/components/MyKeypairManagementModal.tsx
@@ -13,7 +13,7 @@ import { MyKeypairManagementModalRevokeMyKeypairMutation } from '../__generated_
 import { MyKeypairManagementModalSwitchMainKeyMutation } from '../__generated__/MyKeypairManagementModalSwitchMainKeyMutation.graphql';
 import { App } from '../app-shim';
 import { convertToOrderBy } from '../helper';
-import { downloadCSV, escapeCsvValue } from '../helper/csv-util';
+import { csvLiteral, downloadCSV, escapeCsvValue } from '../helper/csv-util';
 import { useBAIPaginationOptionState } from '../hooks/reactPaginationQueryOptions';
 import { useBAISettingUserState } from '../hooks/useBAISetting';
 import { theme } from '../theme-shim';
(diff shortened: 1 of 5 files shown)


ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3106-pr9365-neutralize-csv-formula.sungchul.10-82-0-159.sslip.io)

react/src/components/MyKeypairManagementModal.tsx
```diff
@@ -90,7 +90,7 @@ const downloadCredentialCSV = (credential: KeypairCredential) => {
     credential.secretKey,
     credential.sshPublicKey,
   ]
-    .map((v) => `"${v.replace(/"/g, '""')}"`)
+    .map(escapeCsvValue)
```
agatha197, line 95: The manager generates secret keys as `secrets.token_urlsafe(30)` (`manager/models/keypair/row.py`), whose base64url alphabet includes `-`, so ~1.6% of them start with `-` and now export as `'-xxxx`. This CSV is copy-pasted into CLI configs / env, so the prefix silently breaks auth with no visible cause — same for the `secret key` column in `GeneratedKeypairListModal`'s `exportCSVWithFormattingRules` call. Credential columns should opt out of the formula neutralization.

agatha197 (changes requested):
The trigger set, the signed-number/bare-`-` exemptions, and the routing of all three client-side builders through the one helper look right. One case where the neutralization corrupts a value that is not read by a spreadsheet. (Separately, `/export/{nodeKey}/csv` rows still arrive pre-serialized from the manager, so that path stays unprotected — backend follow-up, not this PR.)

react/src/components/MyKeypairManagementModal.tsx
ironAiken2: Fixed in 61fc56c. `csvLiteral()` marks a cell that `escapeCsvValue` quotes but never neutralizes. `downloadCredentialCSV` wraps the secret key, and `GeneratedKeypairListModal` passes `{ 'secret key': csvLiteral }` as the format rule. Access keys (`AKIA` prefix) and SSH public keys never start with a trigger character, so they stay under the default, and any cell not marked is still neutralized. Tests: a `-`-leading secret key survives while an `=HYPERLINK` email cell in the same row is still neutralized.

agatha197 (approved):
LGTM

PR #9388, opened 2026-09-02
fix(FR-3680): keep a tooltip open while the pointer is on its surface

### LINKED ISSUE #9069: Help-icon tooltip closes when its body is clicked (model folder mount path field)
In the service/deployment launcher form, the **Model folder mount path (optional)** field has a "?" help icon whose tooltip shows the explanation text. While the tooltip is open, clicking inside the tooltip body immediately dismisses it — so the text cannot be selected or copied, and the tooltip vanishes on what feels like an accidental click.

**Expected:** clicking inside the tooltip content keeps it open (only moving the pointer away / clicking outside / Escape should close it).

**Observed:** any click on the tooltip body closes it (see attached screenshot).

Likely affects every "?" help-icon tooltip built the same way, not just this field.

—

JIRA Issue: FR-3680

nowgnuesLee:
Resolves #9069 (FR-3680)

## What

Every Astryx `Tooltip` in the app closed ~100 ms after the pointer left its trigger, whether or not the pointer had moved onto the tooltip's own surface. The reported symptom, "clicking inside the tooltip body dismisses it", is how users met it: the tip was already going away by the time they reached for the text. This fixes the hover bridge in `useTooltip` itself, through the existing pnpm patch of `@astryxdesign/core@0.5.2`, so every "?" help icon, `BAIQuestionIconWithTooltip`, `BAIFormItem tooltip`, and truncation tooltip is covered at once.

## Mechanism (measured, not inferred)

`useTooltip` keeps the tip open over its surface by cancelling the pending hide from the surface's React `onMouseEnter`. React derives that synthetic enter from the trigger's native `mouseout`, and the browser fires `mouseout` **before** `mouseleave`. The trigger's `mouseleave` is the native listener that arms the 100 ms hover-bridge timer, so the order in the live app is:

1. `mouseout(trigger → tip)` → React dispatches the tip's `onMouseEnter` → `cancelHide()` finds no timer.
2. `mouseleave(trigger)` → `scheduleHide()` arms the 100 ms timer.
3. `mouseover(tip)` → React ignores it (its `relatedTarget` is React-managed; the enter was already dispatched in step 1).
4. Timer fires → `layer.hide()`.

Traced with patched `setTimeout`/`clearTimeout` on `https://fr-3680.localhost:1355` (backend 10.82.0.241): the hide always came from the `scheduleHide` timer, and `cancelHide` never ran with a timer to clear. Proof of the mechanism: dispatching the same two events in the reverse order (synthetic `mouseleave` first, then `mouseout` with `relatedTarget` on the tip) made `cancelHide` clear the timer and the tip survive. Two hypotheses were ruled out on the way: the `<label htmlFor>` wrapping the form-item trigger is not involved (the popover is portalled out of the label, and focus stays on `body` after the click), and the dev-only `react-grab` inspector is not involved (same result with it disposed).

`HoverCard` in the same package already handles this correctly: its hide timer re-checks an `isHoveringContentRef` when it fires instead of relying on the cancel. The patch gives `useTooltip` the same shape (`isHoveringSurfaceRef`, set/cleared by the surface's enter/leave with the same touch guard, reset in `hideNow`). Astryx `latest` is 0.5.2, the installed version, so there is no upstream release to bump to.

## Before / after (real app, both modes)

| Step | Before | After |
|---|---|---|
| Pointer moves trigger → tooltip surface, wait 1 s | closed at ~100 ms | open |
| Click inside the tooltip text | closed | open, text selectable, focus stays on `body` |
| Pointer leaves the surface | closed | closed after the 100 ms bridge |
| Escape while on the surface | closed | closed |
| `pageerror` / unhandled rejections | 0 | 0 |

Measured on the deployments list header "리비전" help icon (dark, `data-theme="dark"`) and on the "배포 생성" modal's "배포 이름" `BAIFormItem tooltip` (light, `data-theme="light"`). Same mechanism and same result in both, so the mount-path field named in the issue (also a `BAIFormItem tooltip`) is covered.

## Screenshots (after)

The "before" state is the tooltip being absent after the click, so only the "after" state is pictured.

| Dark, deployments header "리비전" help icon, after a click inside the text | Light, "배포 생성" modal "배포 이름" `BAIFormItem tooltip`, after a click inside the text |
|---|---|
| after dark header | after light modal |

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (after `pnpm --filter backend.ai-ui build`; the theme-build step needs BUI's dist in a fresh worktree)
- `pnpm --filter backend.ai-ui exec vitest run` → 73 files pass, 1962 tests pass. One unrelated flake: `BAIComplexSelect.popup.test.tsx › clears the query and reports the clear upstream` (`user.type` dropped a keystroke, `"b"` vs `"br"`); it passes on rerun and touches no tooltip code.
- `pnpm --prefix ./react exec vitest run` → 110 files, 1730 tests pass
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → 2 pre-existing findings, both `var(--bai-dialog-z)` in `BAIDialog.css` / `BAIDrawerPortal.css`. This diff touches no stylesheet or `var()` usage.
- New regression test in `FormItemVisual.tooltip.test.tsx` replays the browser's exact event order with raw DOM events (RTL's `fireEvent.mouseLeave` fires a second `mouseout` afterwards, which is the reversed, working order). Confirmed it fails with the patch's fire-time check disabled (`Expected "block", Received "none"`) and passes with it.

## Residue

- The pnpm patch is the fix's home until Astryx ships it upstream; the patch file is the review surface (`react/patches/@astryxdesign__core@0.5.2.patch`, both `dist` and `src` copies as with the existing `useFocusTrap` hunk).
- A tooltip whose trigger is reached by touch is unchanged: the new ref is guarded by `touch.isTouchPointerRef`, mirroring HoverCard, so a tap on the surface cannot pin the tip open.
- `astryx component Tooltip` docs still say "Don't place interactive elements inside a tooltip"; this restores hoverable/selectable text (WCAG 1.4.13), not interactive content.

### DIFF
--- a/packages/backend.ai-ui/src/form-engine/FormItemVisual.tooltip.test.tsx
+++ b/packages/backend.ai-ui/src/form-engine/FormItemVisual.tooltip.test.tsx
@@ -18,17 +18,25 @@
  * the popup renders is Astryx's business and needs a real layout engine.
  */
 import BAIFormItemVisual from './FormItemVisual';
-import { render, screen } from '@testing-library/react';
-import { describe, expect, it } from 'vitest';
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const HELP = 'Computer memory is temporary.';
+
+const renderMemoryItem = () =>
+  render(
+    <BAIFormItemVisual label="Memory" tooltip={HELP}>
+      <input aria-label="memory" />
+    </BAIFormItemVisual>,
+  );
+
+const tooltipTrigger = () =>
+  document.querySelector<HTMLElement>('[data-bai-form-item-tooltip]');
 
 describe('BAIFormItemVisual — tooltip', () => {
   it('puts the tooltip text in an overlay layer, not in the label flow', () => {
-    render(
-      <BAIFormItemVisual label="Memory" tooltip="Computer memory is temporary.">
-        <input aria-label="memory" />
-      </BAIFormItemVisual>,
-    );
-    const text = screen.getByText('Computer memory is temporary.');
+    renderMemoryItem();
+    const text = screen.getByText(HELP);
     // jsdom implements neither the Popover API nor CSS anchor positioning, so
     // Astryx's layer sits in the tree and reads as "visible" here. What makes
     // it a TOOLTIP rather than inline prose is the pair of attributes the
@@ -41,12 +49,8 @@ describe('BAIFormItemVisual — tooltip', () => {
   });
 
   it('renders a focusable trigger after the label', () => {
-    render(
-      <BAIFormItemVisual label="Memory" tooltip="Computer memory is temporary.">
-        <input aria-label="memory" />
-      </BAIFormItemVisual>,
-    );
-    const trigger = document.querySelector('[data-bai-form-item-tooltip]');
+    renderMemoryItem();
+    const trigger = tooltipTrigger();
     expect(trigger).not.toBeNull();
     expect(trigger?.getAttribute('tabindex')).toBe('0');
     // The glyph, not the prose.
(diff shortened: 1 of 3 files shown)


nowgnuesLee:
We've decided to keep the current behavior, so I'm closing this issue.

PR #9400, opened 2026-09-02
docs: update deployment docs (2026-09-02)

lablup-octodog:
## What changed and why

Two fixes in this range made the Deployment page name the reason an action is unavailable instead of silently greying it out, and the manual documented only a subset of those blocking conditions:

- **fix(FR-3679): name the reason on the disabled access-token buttons (#9166)** — `DeploymentAccessTokensCard.tsx` now carries a reason inside the disabled state: `deployment.accessToken.DeploymentStopped`, `OnlyOwnerCanManage`, and `EndpointNotIssuedYet`. The same `mutationDisabled` value drives the per-row **Delete** action.
- **fix(FR-3722): derive row-action `disabled` from `disabledReason` (#9170)** — `DeploymentRevisionHistoryTab.tsx` extends the revision **Apply** button's blocking branches: current/applying revision, deployment in a stopped category, and an apply/rollback request already in flight. The detail drawer's Apply button carries the same conditions.

## Edits (en/ko/ja/th, prose only)

- **Access Tokens** (`deployment.md` ~line 717): the disabled-`Create Access Token` note now names all three blocking states with their on-screen tooltips, and notes that the owner and stopping/inactive conditions disable deleting a token as well.
- **Revision History** (~lines 610–615): the Apply-disabled conditions now include "while the deployment is stopped, and while an apply or rollback is already in progress"; the detail-drawer sentence was changed to reference the same conditions instead of restating a now-incomplete pair.

No screenshots: the plan assigns none to this chapter, and both edits describe states (stopped or non-owned deployment) the capture environment cannot depict.

## Deliberately skipped

Nothing chapter-relevant was skipped. Range-wide skips that touch Deployment-adjacent code: FR-3722's shared `BAINameActionCell` rework itself (internal component contract; its user-visible outcomes are exactly the edits above), and FR-3840's Runtime Parameter rename (its own PR already updated `deployment_presets.md` in all four languages — I confirmed the plan's claim stands and this chapter's files contain no stale term).

## Review findings

**No defects found.** Independently verified:

- All eight quoted tooltip strings (two per language) match `resources/i18n/{en,ko,ja,th}.json` verbatim.
- The delete-action claim is correct: `deleteDisabled` receives `mutationDisabled` (owner + stopped conditions only), not the endpoint condition — the docs' "those last two conditions" scoping is exact.
- "Stopping or no longer active" is accurate beyond the tooltip text: the `isDeploymentDestroying` prop is derived from `isDeploymentInStoppedCategory(deploymentStatus)` at the call site (`AdminDeployment.tsx`).
- The writer's disclosed deviation ("while an apply or rollback is already in progress" instead of the plan's "while a rollback is in progress") is justified: the table blocks the row whose request is in flight (`rollingBackRevisionId === record.id`) and the drawer blocks on `!!rollingBackRevisionId`; the chosen phrasing covers both without introducing "rollback" as an undefined standalone feature. Minor nuance for the reviewer: in the *table*, other rows technically stay clickable during an in-flight request — the generalized phrasing slightly widens the table-level claim, which I judge an acceptable simplification since the drawer condition is global and the request window is brief.
- Translation parity: all four languages received both edits with matching sentence structure; ja keeps the page's established **デプロイ** label for the Apply button, consistent with `deployment.Apply` in `ja.json`.
- No screenshots in the diff, no `TODO(screenshot)` markers, no edits outside `src/{en,ko,ja,th}/deployment/deployment.md`.

## Verification

- Wrapper-verified: diff touches only `packages/backend.ai-webui-docs/`, and `scripts/verify.sh` passes (`=== ALL PASS ===`).

> [!NOTE]
> This PR carries no `FR-` number by design (automated nightly docs sync), so it deviates from the repo's `prefix(FR-XXXX)` title convention and will not be linked by `project-status-sync`.

---


agatha197 (approved):
LGTM

Issue #9073, opened 2026-08-25
Modal forms no longer scroll to the first invalid field when OK is pressed


nowgnuesLee:
In modal forms (e.g. the **Create User** modal), pressing the OK/Create button while a required or otherwise invalid field sits outside the modal's visible scroll area used to auto-scroll the form to that field. That behaviour is gone: the click is rejected by validation, but the modal stays where it is, so the user sees no error and cannot tell which field blocked the submit.

**Expected:** on a failed submit, the modal scrolls the first invalid field into view (and focuses it), as it did before.

**Observed:** no scroll — the invalid field and its error message stay off-screen (see attached screenshot of a long modal form).

Likely affects every modal form, not just Create User.

—

Captured while working on branch: FR-3662

JIRA Issue: FR-3683

PR #9403, opened 2026-09-03
feat(FR-3683): turn scroll-to-first-error on for every form

agatha197:
Part of #9073 (FR-3683)

> Stack 2/2 — stacked on #9204, which makes `scrollToFirstError` work at all. Merge that one first.

## What changed

#9204 fixed the mechanism and converted the one modal the report named. This turns it on **app-wide**: one line on the `<FormConfigProvider>` in `DefaultProviders`, next to the `requiredMark` default it already sets there, so a form added next month is covered without remembering the prop. A form's own prop still wins, and the engine's default stays antd's off, so a `backend.ai-ui` consumer mounting no provider is unaffected.

Because the trigger is `form.submit()`, this reaches exactly the forms that submit — today `UserSettingModal` (converted in #9204), `ImportNotebookForm` and `DeploymentAddRevisionModal` — and nothing that validates in the background. **There is no opt-out list.** The remaining modals pick it up as they convert to `submit()` + `onFinish`; that conversion is the follow-up, split by area.

`DeploymentAddRevisionModal` is left as is: its `handleOk` only calls `submit()` after its own `validateFields()` passes, so the engine's scroll never fires there and its hand-rolled DOM-order walk keeps handling the failure case.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===`; `backend.ai-ui` `tsc` lane passes.
- `Form.scrollToFirstError.test.tsx` gains 2 cases: the provider turns it on for a form below it; a form's `scrollToFirstError={false}` overrides the provider.
- Full suites green with both PRs applied.

https://claude.ai/code/session_01Lefo3KxG9bKDgU2z72sxqU

### DIFF
--- a/packages/backend.ai-ui/src/form-engine/Form.scrollToFirstError.test.tsx
+++ b/packages/backend.ai-ui/src/form-engine/Form.scrollToFirstError.test.tsx
@@ -13,9 +13,11 @@
      wrapper's `data-bai-field-item`;
    - a `noStyle` field, which has no wrapper of its own, lands on the parent
      item that shows its error;
-   - the switch is off until `<Form scrollToFirstError>` says otherwise;
+   - the switch is off until `<Form scrollToFirstError>` or
+     `<FormConfigProvider scrollToFirstError>` says otherwise, form winning;
    - the store's own `scrollToField` works again on such controls.
  */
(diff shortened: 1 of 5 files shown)


agatha197:
not planned.

PR #9411, opened 2026-09-03
fix(FR-3727): suppress the press scale on disabled Astryx buttons

### LINKED ISSUE #9179: Astryx Button: a disabled button still plays the full press animation (scale 0.98 + darkening) on mouse-down
## Component

- **Name**: Astryx `Button` (reached through `BAIButton`, `packages/backend.ai-ui/src/components/BAIButton.tsx`)
- **Astryx primitive**: `Button` / `IconButton`
- **Scope**: every usage — the behaviour was measured on the shared Astryx `Button` styles themselves, not on one call site. `BAIButton` alone is referenced from 100 files.

## Where it is reachable

- `/deployment/:id` (menu: Serving → Deployments → a deployment) — the **Access Tokens** card's **Create Access Token** button, once the deployment is stopped.
- Any disabled `BAIButton` / row action reproduces it; the deployment page is simply the one this was found on.

## Steps to reproduce

1. Open the detail page of a **stopped** deployment.
1. Scroll to the **Access Tokens** card. **Create Access Token** renders disabled.
1. Press and **hold** the left mouse button on it.

## Observed

The disabled button plays the full press treatment while held: it scales down to

`0.98` and gains a 15% darkening overlay — pixel-identical to what an enabled

button does. Releasing restores it. No click handler runs, so nothing actually

happens.

Measured computed styles (Chromium, real mouse press, transition allowed to settle):

| Button | at rest | while pressed |
|---|---|---|

|—|—|—|
|---|---|---|

| enabled | `scale(1)`, no overlay | `scale(0.98)`, `rgba(0,0,0,0.15)` overlay |
|---|---|---|

| **disabled** (native `disabled`) | `scale(1)`, no overlay | **`scale(0.98)`, `rgba(0,0,0,0.15)` overlay** |
|---|---|---|

| **disabled** (`aria-disabled`, i.e. with a tooltip) | `scale(1)`, no overlay | **`scale(0.98)`, `rgba(0,0,0,0.15)` overlay** |
|---|---|---|

Both disabled variants behave identically to the enabled one.

Isolated check, no Astryx involved — a plain `<button disabled>` with

`button:active { … `} still matches `:hover` and `:active` in Chromium and is

still returned by `elementFromPoint`. So the native `disabled` attribute

suppresses **event dispatch** only; it does not stop these selectors from

matching. Only `pointer-events: none` does, and that is not usable here because

it would also kill the hover that a disabled button's explanatory tooltip

depends on.

## Expected

A disabled control gives no pressed feedback — no scale-down, no darkening —

regardless of whether it is natively `disabled` or `aria-disabled`.

Astryx already takes exactly this position for hover: the variant styles guard

it with `':hover:where(:not(:disabled,[aria-disabled="true"]))'`. The same

intent should hold for the pressed state.

## Environment

- **Mode**: light (not re-checked in dark; the overlay is a neutral alpha wash, so it is expected to behave the same)
- **Viewport**: ~1280px
- **Build**: branch `FR-3679` dev server; the native-`disabled` path was measured too and behaves the same, so this is **not** branch-specific and is present on `main`.

## Impact

Workaround exists (nothing breaks), but it is actively misleading: the press

feedback reads as "the click registered", so users retry and conclude the button

is broken rather than disabled. This is the second half of **FR-3679**, where it

was reported as "the button still appears to accept clicks". FR-3679's own PR

fixes only the missing-reason tooltip; this part is untouched by it.

## Evidence

- No screenshot — the effect only exists while the mouse is held down. The

computed-style table above was captured programmatically instead.

## Notes

Cause is already measured rather than suspected, so recording it here to save

the fix a re-run: in `Button.tsx` the variants' `':active'` key has **no**

`:where(:not(:disabled,[aria-disabled]))` guard although the neighbouring

`':hover'` key does, and `styles.pressable` is applied **after** `styles.disabled`

/ `styles.ariaDisabled`, so their `transform: none` suppression loses to

last-wins ordering.

JIRA Issue: FR-3727

nowgnuesLee:
Resolves #9179 (FR-3727)

## Mechanism

Astryx `Button` builds its class list with `stylex.props(...)`, which is last-wins per property. `styles.pressable` (`transform: scale(1)` / `:active → scale(0.98)`) was passed **after** `styles.disabled`, so the disabled style's `transform: none` never reached the DOM and a disabled button still shrank on mouse-down. `styles.ariaDisabled` (the path a disabled button with a tooltip takes) had no transform suppression at all.

The fix is a `pnpm patch` on `@astryxdesign/core@0.5.2` (`react/patches/@astryxdesign__core@0.5.2.patch`, extending the existing `useFocusTrap` patch):

- pass `styles.pressable` before `styles.disabled` / `styles.ariaDisabled`, so their `transform: none` wins;
- give `styles.ariaDisabled` the same `transform: { default: 'none', ':active': 'none' }`.

The `dist/Button/Button.js` edit reuses the classes StyleX already emits for `styles.disabled` (`x1c071of` = `transform:none`, `x1pdlv7q` = `:active{transform:none}`), so `dist/astryx.css` is unchanged. `src/Button/Button.tsx` carries the same change for reference.

The 15% darkening overlay from the original report is already gone in 0.5.2 — `interactionOverlayStyles` guards both `:hover` and `:active` with `:where(:not(:disabled,[aria-disabled="true"]))` — so only the scale remained.

## Measured (dev server, Chromium, real `mouse.down()` held 500 ms)

Disabled ghost `IconButton` on `/summary`. "Before" is the same element with the pre-patch class output (`x3oybdh xk4oym4`, no `x1c071of x1pdlv7q`) re-applied; "after" is what the patched build renders.

| Mode | State | transform (before) | transform (after) |
|---|---|---|---|
| light | at rest | `matrix(1,0,0,1,0,0)` | `none` |
| light | pressed | **`matrix(0.98,0,0,0.98,0,0)`** | `none` |
| dark (`data-theme="dark"` via the header toggle) | at rest | `matrix(1,0,0,1,0,0)` | `none` |
| dark | pressed | **`matrix(0.98,0,0,0.98,0,0)`** | `none` |

`backgroundImage` stayed `none` in every cell. Enabled control (`/serving`, primary button): rest `scale(1)` → pressed `scale(0.98)`, overlay `0.16 → 0.18`, i.e. the press feedback on enabled buttons is untouched. Zero `pageerror` in both modes.

SSR class check on the patched dist (`react-dom/server`), all three paths of `Button`:

| Props | `aria-disabled` | scale class | `transform:none` classes |
|---|---|---|---|
| `{}` | no | yes | no |
| `isDisabled` | no (native `disabled`) | no | yes |
| `isDisabled` + `tooltip` | **yes** | no | yes |

No screenshot: the effect exists only while the mouse is held, and a 0.98 scale on a 24 px icon button is not legible in a static PNG. The computed-style tables above are the evidence.

## Verification

- `bash scripts/verify.sh`: Relay / Lint / TypeScript / Vite warmup / StyleX / Astryx theme build / Astryx integration / z-index / Agent mappings / Terminology all PASS. **Format FAIL is pre-existing and not from this branch**: it flags an untracked `src/lib/backend.ai-client-node.js` in my working tree (not part of this PR).
- `pnpm --prefix ./react exec vitest run`: 111 files, 1764 tests passed.
- `pnpm --filter backend.ai-ui exec vitest run`: 1965 passed, 1 skipped, **1 failed** — `BAIComplexSelect.popup.test.tsx › clears the query and reports the clear upstream` (`user.type` yields `b` instead of `br`). Not touched by this PR, and it passes when the file is re-run alone (10/10), so it is a timing flake.
- `node scripts/migration-gates/astryx-token-gate.mjs --strict`: exit 1 on two **pre-existing** undeclared `var(--bai-dialog-z)` usages in `BAIDialog.css` / `BAIDrawerPortal.css` (from FR-3578). This PR adds no CSS.

## Residue

- This is a vendored patch; the same change should go upstream to Astryx (`Button.tsx`: order `pressable` before `disabled`, add `transform` to `ariaDisabled`). Until then, bumping `@astryxdesign/core` means re-applying the patch.
- Grouped buttons (`ButtonGroup`) never receive `styles.pressable`, so they were not affected and are not changed.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -204,7 +204,7 @@ overrides:
   lodash: ^4.18.0
 
 patchedDependencies:
-  '@astryxdesign/core@0.5.2': c9b0d04ba4d03f2fed56352870de082e2a58f9d92b509e247342e59c951918ed
+  '@astryxdesign/core@0.5.2': 7d72b4f11e736ec0ea6d4cfdf57e4bbe68baecee56c489f24fcc11db29684c16
   '@astryxdesign/lab@0.3.0-canary.12db2a1': 0e5e0cbe89a52670bf3b7b7324794cf671a257ce501858d58f346a993af51e08
   '@cloudscape-design/board-components@3.0.176': 06f6366f1016801dd543a4b968e088fbcbf34ee1a4473bc09a07e35206597a98
   ansi_up@6.0.6: 99856fc7fa4571c6ef4716cfd2dbf1151a025690c737775e248e8b7fe08a4459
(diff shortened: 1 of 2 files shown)


nowgnuesLee:
Closing this in favour of an upstream fix.

The cause is in Astryx core itself, not in this repo — reproduced with unmodified `@astryxdesign/core@0.5.2` in a standalone Vite + React app (https://github.com/nowgnuesLee/astryx-disabled-press-repro), and the same merge order is still on Astryx `main`. Filed as facebook/astryx#5961: https://github.com/facebook/astryx/issues/5961

Carrying a vendored patch for it would have to be re-applied on every `@astryxdesign/core` bump, so we will pick the fix up with the release that closes #5961 instead. FR-3727 / #9179 stay open until then.

PR #9417, opened 2026-09-03
fix(FR-3846): apply "Off" on the first auto-refresh interval pick

### LINKED ISSUE #9414: Auto-refresh interval menu: "Off" cannot be selected directly from the current interval
Selecting "Off" in the auto-refresh interval menu does not take effect when chosen directly. The user has to first pick a different interval (e.g. another number of seconds) and only then choose "Off" for it to actually stop the auto-refresh.

**Steps to reproduce**

1. Open the Sessions page (compute session list).
1. Open the auto-refresh interval menu on the fetch/refresh button (currently showing e.g. "15s").
1. Select "Off".

**Expected**: auto-refresh stops immediately.
**Observed**: nothing changes; auto-refresh keeps running. Selecting another interval first, then "Off", works as expected.

**Where**

- Page: Sessions (`/src/pages/ComputeSessionListPage.tsx`)
- Element pinned in `BAIFlex` (`packages/backend.ai-ui/src/components/BAIFlex.tsx`)
- Likely the auto-refresh interval menu on `BAIFetchKeyButton`.

----

Captured while working on branch: main (reported from a dev server running PR #9368 / FR-3813)

JIRA Issue: FR-3846

yomybaby:
Resolves #9414 (FR-3846)

Selecting **Off** in the auto-refresh interval dropdown did nothing when it was the user's *first* pick on a page. Picking any other interval first and *then* Off worked — the workaround the report describes.

**Dev server (dev VPN):** http://fr-3846.jongeun.10-82-0-159.sslip.io — Sessions page, open the refresh button's interval menu and pick **Off** as the very first choice.

## Mechanism

Not the Astryx dropdown. `AutoUpdateFetchKeyButton` (the host wrapper) decided *"has the user chosen an interval yet?"* with `hasStoredUserSetting`, a plain `localStorage` read outside React, while `useBAISettingUserState` reads **"never set"** and **"explicitly Off"** as the same `null`:

```
default in effect (nothing persisted)  →  atom value null
user picks "Off"                       →  writes null   →  atom value null
```

Jotai bails out of a derived value that is `Object.is`-equal, so the write notified nobody. No re-render → `hasStoredUserSetting` never re-evaluated → `defaultAutoUpdateDelay` (15 s on the Sessions page) kept winning, and auto-refresh kept running. Picking `30s` first moves the atom `null → 30000`, which *does* re-render — hence the workaround.

The write itself was never broken: `localStorage` held `"null"` the whole time. Only the render was stale.

**Fix:** track the pick in local state so the first Off is a real state change. The `localStorage` read stays as the cross-reload source of truth.

## Measured, live on a dev server

Fresh browser profile (nothing persisted), Sessions page, **Off clicked as the first pick**:

| | `localStorage` before | trigger label before | `localStorage` after | trigger label after |
|---|---|---|---|---|
| **Before fix** | `null` | `15s` | `"null"` | **`15s`** ← unchanged |
| **After fix** | `null` | `15s` | `"null"` | `""` (auto-refresh off) |

Both rows measured in light **and** dark (`data-theme="dark"` asserted on `<html>`); identical in both. Zero `pageerror` in every pass.

| Before fix — after clicking "Off" | After fix — after clicking "Off" |
|---|---|
| before | after |

The interval chip and its orange countdown border persist on the left; on the right the control falls back to a bare chevron, which is the off state.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (required `pnpm --filter backend.ai-ui build` first — this worktree had no BUI `dist`, the documented stale-dist trap)
- `pnpm --prefix ./react exec vitest run` → 112 files / 1768 tests passed
- `pnpm --filter backend.ai-ui exec vitest run` → 75 files / 1966 tests passed, 1 skipped
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → exits 1 on 2 **pre-existing** undeclared `var(--bai-dialog-z)` uses in `BAIDialog.css` / `BAIDrawerPortal.css`. Neither file is in this diff (2 files, no CSS).

New regression test `AutoUpdateFetchKeyButton.test.tsx` (4 cases) drives the real jotai atom and real `localStorage`, mocking only the presentational button. It fails on `main` with `expected 15000 to be null`.

## Residue

- The leg where the Astryx `DropdownMenu` emits `null` for the Off item is **not** unit-tested: jsdom has no Popover API, no BUI test currently exercises `DropdownMenu`, and adding the first one means polyfilling that API — out of scope here. It is covered by the live probe above and by the reporter's own workaround.
- Every other `settingId` consumer with a non-null `defaultAutoUpdateDelay` had the same bug and is fixed by the same change; consumers defaulting to `null` were never affected.

https://claude.ai/code/session_01F9aDnUBNjx7zE8aNvAGsmN

### DIFF
--- a/react/src/components/AutoUpdateFetchKeyButton.test.tsx
+++ b/react/src/components/AutoUpdateFetchKeyButton.test.tsx
@@ -1,0 +1,84 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+/**
+ * FR-3846: picking "Off" while the consumer's `defaultAutoUpdateDelay` is still
+ * in effect must turn auto-refresh off on the first try, not only after a
+ * detour through another interval.
+ */
+import AutoUpdateFetchKeyButton from './AutoUpdateFetchKeyButton';
+import '@testing-library/jest-dom';
+import { act, render } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Capture what the host wrapper actually hands the presentational button; the
+// dropdown's own rendering is not under test.
+let lastAutoUpdateDelay: number | null | undefined;
+let lastOnChangeAutoUpdateDelay: ((delayMs: number | null) => void) | undefined;
+
+vi.mock('backend.ai-ui', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('backend.ai-ui')>();
+  return {
+    ...actual,
+    BAIFetchKeyButton: (props: {
+      autoUpdateDelay?: number | null;
+      onChangeAutoUpdateDelay?: (delayMs: number | null) => void;
+    }) => {
+      lastAutoUpdateDelay = props.autoUpdateDelay;
+      lastOnChangeAutoUpdateDelay = props.onChangeAutoUpdateDelay;
+      return <div data-testid="fetch-key-button" />;
+    },
+  };
+});
+
+const SETTING_KEY =
+  'backendaiwebui.settings.user.fetchKeyAutoUpdateDelay.session-list';
+
+const renderButton = () =>
+  render(
+    <AutoUpdateFetchKeyButton
+      settingId="session-list"
+      defaultAutoUpdateDelay={15_000}
+      onChange={() => {}}
+    />,
+  );
+
+describe('AutoUpdateFetchKeyButton', () => {
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


yomybaby:
Dev server on box `jongeun`: http://fr-3846.jongeun.10-82-0-159.sslip.io

agatha197 (approved):
LGTM

PR #9419, opened 2026-09-03
refactor(FR-3848): give the assign modal's user select its own Suspense boundary

### LINKED ISSUE #9418: RBAC role detail > Role assignment: move the user query of the assign modal into its own Suspense-bounded select
Follow-up to FR-3725 (PR #9413).

#9413 stopped the **Add user** click from suspending the whole Role assignment tab by deferring the modal's `open` inside `AssignRoleModal` (`useDeferredValue(open, false)`). That keeps the tab rendered, but it relies on a deferred render plus a `store-only` first pass, which is indirect for what is really a missing Suspense boundary.

**Proposed change**

- Replace the deferred-open handling with a component split inside `react/src/components/AssignRoleModal.tsx`: the user options query (`AssignRoleModalQuery`), the search state and the `BAISelect` move into an `AssignRoleUserSelect` component that renders inside its own `Suspense` boundary.
- The boundary sits inside the `Form.Item` child, so the `userIds` field stays registered (and its `required` rule active) while the options load; the fallback is the same select in `loading` + `disabled` state.
- Drop the `maxTagCount` / `maxTagPlaceholder` props while moving the select: `BAISelect` accepts and ignores both.

**Expected**: same user-visible behaviour as after #9413 (tab stays rendered, modal opens at once, select shows loading), with the boundary expressed directly.

JIRA Issue: FR-3848

nowgnuesLee:
Resolves #9418 (FR-3848)

Follow-up to #9413 (FR-3725).

applied server: https://fr-3848-pr9419-give-assign-modal.seungwon.10-82-0-159.sslip.io/

## Why

#9413 stopped the **Add user** click from suspending the whole Role assignment tab by deferring the modal's `open` inside `AssignRoleModal` (`useDeferredValue(open, false)` plus a `store-only` first pass). That workaround existed because the modal's inline user select ran its own `useLazyLoadQuery` on mount.

Review on this PR asked to separate the select. Doing so surfaced that the inline select's server search had been dead since the move to `BAISelect` (Astryx `Selector` exposes no `onSearch`), so the modal always listed the first 50 users and filtered client-side.

### `BAIAdminUserV2Select` (new, `backend.ai-ui`)

A paginated user picker over `adminUsersV2`, the V2 sibling of `BAIUserSelect` (`user_nodes`). Built on `BAIComplexSelect` + `useLazyPaginatedQuery`, same shape as the other 18 `BAIComplexSelect` fragments:

- The option list is fetched **when the popup opens** (`store-only` while closed, `network-only` while open), so nothing suspends on mount and no Suspense boundary is needed. The trigger shows a loading state while the open-triggered fetch and searches are in flight.
- `limit` + `offset` pagination (one mode only, per `graphql-pagination.md`), 10 rows per page, `orderBy: EMAIL ASC`, scroll-driven `loadNext`.
- Server-side search: the typed text becomes `email: { iContains }`.
- `valuePropName="id" | "email"` (default `email`); under `id` the selected UUIDs are re-resolved to emails with `uuid: { in }` so a value chosen on page 1 keeps its label after paging.
- `filter?: UserV2Filter` and `excludeInactive` are combined through the new `combineFilters(filters, operator = 'AND')` helper (`helper/combineFilters.ts`, also adopted by `BAIProjectVfolderSelect`), so a caller's filter cannot clobber the search predicate.
- `onChange(value, option)` hands back the label pair(s), same contract as `BAIUserSelect`.
- Ships with `BAIAdminUserV2Select.doc.ts` and a story (`adminUsersV2` mocked via `RelayResolver`).

### `AssignRoleModal`

- The inline select, its `AssignRoleModalQuery`, and the deferred-open workaround are removed. The `userIds` field is a single `Form.Item` holding `BAIAdminUserV2Select multiple valuePropName="id"`; there is no Suspense boundary in the modal because the picker does not suspend on mount.
- The partial-failure table now shows emails: the label pair from `onChange` populates `userLabelsRef`, which the old `BAISelect` multiple mode never did (its second `onChange` argument was always `undefined`).
- `maxTagCount` / `maxTagPlaceholder` are dropped: `BAISelect` accepted and ignored both.

## Verification

- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → no new findings (the two pre-existing `--bai-dialog-z` hits only)
- `pnpm --filter backend.ai-ui test` → passing (`combineFilters` 6 cases, `BAIAdminUserV2Select` + `BAIComplexSelect` 14 cases)
- `pnpm --filter backend.ai-ui exec tsc --noEmit -p .` (covers the stories) → clean
- Verified live against `frontend.latest.staging.backend.ai`: Add user opens the modal at once with the tab intact; opening the picker fetches and lists users; typing `u1@` returns one server-filtered row; selecting shows the resolved email on the trigger.


react/src/components/AssignRoleModal.tsx
```diff
@@ -240,76 +286,40 @@ const AssignRoleModal: React.FC<AssignRoleModalProps> = ({
       {...baiModalProps}
     >
       <Form ref={formRef} layout="vertical">
-        <Form.Item
-          name="userIds"
-          label={t('credential.Users')}
-          rules={[{ required: true, message: t('rbac.PleaseSelectUsers') }]}
+        <Suspense
+          fallback={
+            <Form.Item label={t('credential.Users')}>
```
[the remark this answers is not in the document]
nowgnuesLee: Fixed in `c6a2e8dbc` — the fallback now renders the same `Form.Item` (`name="userIds"`, required rule) around the loading select, so the field stays registered and `validateFields()` still rejects an empty selection while the options load.

react/src/components/AssignRoleModal.tsx
```diff
@@ -40,6 +39,85 @@ interface FailedAssignment {
   message: string;
 }
 
+type AssignRoleUserSelectProps = Pick<
+  BAISelectProps<string[]>,
+  'value' | 'onChange' | 'placeholder' | 'status' | 'disabled'
+>;
+
+/** Owns the user options query; render it under a Suspense boundary. */
+const AssignRoleUserSelect: React.FC<AssignRoleUserSelectProps> = ({
```
agatha197, line 48: how about separate this select?

react/src/components/AssignRoleModal.tsx
```diff
@@ -40,6 +39,85 @@ interface FailedAssignment {
...
+        label: user?.basicInfo?.email || user?.id,
+        description: user?.basicInfo?.fullName,
+      }))}
+      optionRender={(option) => (
+        <div>
+          <div>{option.label}</div>
+          {option.data?.description && (
+            <div
+              style={{
+                fontSize: 12,
```
agatha197, line 107: please use token value.

agatha197 (changes requested):

react/src/components/AssignRoleModal.tsx
nowgnuesLee: Fixed in 05d8e99d1 — `fontSize: 12` → `fontSize: 'var(--font-size-sm)'` (the Astryx token that maps to exactly 12px, per `theme-shim/mapping.ts`).

react/src/components/AssignRoleModal.tsx
nowgnuesLee: Separated in da5a20eec + 1380984d6 — the select is now `BAIAdminUserV2Select` in `backend.ai-ui` (the `adminUsersV2` sibling of `BAIUserSelect`, on `BAIComplexSelect` with `limit`/`offset` pagination and `email iContains` server search), and the modal renders it with `multiple valuePropName="id"`. Doing this also revived the server search, which had been dead on the inline `BAISelect` (no `onSearch`), and the failure table now gets emails from the `onChange` label pair. PR body updated.

nowgnuesLee:
Dev server on box `seungwon`: http://fr-3848-pr9419-give-assign-modal.seungwon.10-82-0-159.sslip.io

packages/backend.ai-ui/src/components/fragments/BAIAdminUserV2Select.tsx
```diff
@@ -0,0 +1,297 @@
...
+  useState,
+  useTransition,
+} from 'react';
+import { graphql, useLazyLoadQuery } from 'react-relay';
+
+export type BAIAdminUserV2SelectFilter = NonNullable<
+  BAIAdminUserV2SelectPaginatedQuery['variables']['filter']
+>;
+
+export type AstryxAdminUserV2Node = NonNullable<
```
agatha197: we don't need to use 'astryx' prefix anymore.

agatha197 (dismissed):
Image
please match this to form error style.

nowgnuesLee:
> Image please match this to form error style.

Inside a form, the form handles error states on its own. This example is just showing how the component renders that error state, so no changes are needed here.

packages/backend.ai-ui/src/components/fragments/BAIAdminUserV2Select.tsx
nowgnuesLee: I fixed.

yomybaby (approved):
LGTM

PR #9427, opened 2026-09-04
fix(FR-3852): show which logo a login/About item inherits on the Branding page

### LINKED ISSUE #9425: Branding: login and About logos change with the sider logo and cannot be reset on their own
## Report

On the Branding page, changing **Light Mode Logo** also changes the **Login Logo (dark mode)** and the **About Backend.AI Modal Logo (dark mode)**; changing **Dark Mode Logo** changes their light-mode counterparts. The dependent items show the inherited image, but their **Reset** does nothing, so an operator has to click through all three settings to find where the value actually lives. (Reported on PR #9361 review; the behaviour predates FR-3605 — the same code is on `main`.)

## Cause

The shipped `theme.json` declares only `branding.logo.src` / `srcDark` (and the collapsed pair). The login and About logos have no explicit keys, so the components fall back **across modes**:

- dark mode: `loginLogoSrcDark` → `src` → default, `aboutLogoSrcDark` → `src` → default
- light mode: `loginLogoSrc` → `srcDark` → default, `aboutLogoSrc` → `srcDark` → default

(`react/src/components/LoginFormPanel.tsx`, `react/src/components/AboutBackendAIModal.tsx`). The Branding page's path input showed the inherited value as if it were the item's own, so the dependency was invisible; Reset on a login/About item restores a key that was never set, so nothing changes.

## Decision

The cross-mode fallback stays — an unset login/About logo deliberately follows the sider logo of the opposite mode — and is made **visible** instead of being pre-filled into the shipped document. The four keys are **not** added to `theme.json`, so a Light/Dark Mode Logo change still reaches an unset login/About logo, by design.

- On the Branding page a login/About item without its own value keeps its path input **empty**, shows no preview, and states which item the page borrows instead ("No value is set, so the Dark Mode Logo image is used."). With a value it shows the image preview.
- Reset restores from the shipped document, so on a login/About item it removes the key (the shipped document has none). Previously Reset read from `rawThemeConfig`, which in preview mode is the draft itself, so it was a no-op there.
- The login form and the About modal share one helper (`resolveDetailLogoSrc`) for the fallback chain they used to spell out inline.
- Clearing a login/About path returns the item to the fallback chain instead of storing an empty string.
- The About modal's built-in fallback asset (`/manifest/backend.ai-white-text.svg`, removed in #2683) is replaced with assets that ship.

## Acceptance

- A login/About item with no value of its own shows an empty path input, no preview, and a hint naming the source item (e.g. "Dark Mode Logo").
- Entering a path on a login/About item shows its preview and makes it independent of the sider logos; clearing the path removes the key and the hint reappears.
- Reset on each login/About item restores that item only and leaves no key behind.
- The login form, the About modal and the Branding preview render the same image for the same document.
- Every built-in fallback asset referenced by the resolver exists in `manifest/` (unit-tested).

## References

- PR #9361 (FR-3605) review comment by agatha197, 2026-09-04.

JIRA Issue: FR-3852

ironAiken2:
Resolves #9425 (FR-3852)

Base: `feat/FR-3605-theme-v2-document` (#9361), because the ticket targets the v2 `branding.logo.*` document. Not in the #9361 → #8860 → #9362 stack — gh-stack chains are linear, and this fix is independent of the two PRs above #9361.

## Summary

On the Branding page a login / About logo with no value of its own showed the sider logo it inherits as if it were its own path. So changing **Light Mode Logo** silently changed **Login Logo (dark mode)** and **About Backend.AI Modal Logo (dark mode)**, and **Reset** on those items did nothing visible. The cross-mode fallback is intentional — an unset detail logo borrows the sider logo of the opposite scheme — so this PR keeps it and makes it visible instead of pre-filling the shipped document.

## Changes

- `LogoPreviewer`: the path input edits only the item's own key (`getLogoThemeKey`) and stays empty when unset. With a value the item shows the image preview; without one it shows no preview, only a secondary-colored hint 2px under the input (the shape of a form item's explain row) naming the sider logo the page borrows (`getLogoFallbackKey`): *"No value is set, so the Dark Mode Logo image is used."* Committing an empty path removes the key instead of storing `''`.
- `useDefaultTheme.resetDefaultTheme`: restores from the **shipped** document (`getStaticAppearanceConfig`), the same source the draft is seeded from, instead of `rawThemeConfig` — in preview mode that is the draft itself, so per-item Reset was a no-op there. Reset on a login/About item now removes the key, since the shipped document has none.
- `react/src/helper/logoSource.ts` (new): `resolveDetailLogoSrc(logo, area, scheme)` — the fallback chain (own key → opposite-scheme sider logo → built-in asset) the two renderers `LoginFormPanel` and `AboutBackendAIModal` had each spelled out inline. An empty string counts as unset, matching the `||` chains it replaces.
- `AboutBackendAIModal`: the built-in fallback `/manifest/backend.ai-white-text.svg` (deleted in #2683) is replaced by `backend.ai-webui-black.svg` / `-white.svg`, the assets the shipped document already resolves to, so nothing changes visually.
- i18n: `userSettings.logo.FallbackFollowsItem` in all 21 locales.
- `logoSource.test.ts`: own / inherited / empty-string / default cases, plus a check that every built-in asset the resolver can fall back to exists in `manifest/`.

Not in this PR, by decision on the ticket: declaring `loginLogoSrc*` / `aboutLogoSrc*` in `resources/theme.json`. A Light/Dark Mode Logo change still reaches an unset login/About logo on purpose; the Branding page now says so.

## Test server

- Backend: `10.82.0.130`

## Verification

`bash scripts/verify.sh`:

```
--- Relay: PASS ---
--- Lint script coverage: PASS ---
--- Lint: PASS ---
--- Format: PASS ---
--- TypeScript: PASS ---
--- TypeScript (agent-cli): PASS ---
--- Vite warmup paths: PASS ---
--- StyleX cssInjectionTarget: PASS ---
--- Astryx theme build: PASS ---
--- Astryx integration (backend.ai-ui): PASS ---
--- z-index ladder mirrors: PASS ---
--- Agent mappings: PASS ---
--- Terminology: PASS ---
=== ALL PASS ===
```

`pnpm exec vitest run src/helper/logoSource.test.ts` (in `react/`): 5 passed. `make i18n` reports no drift beyond the two new keys.

## Screenshots

Branding → Detail Logo, fresh draft (no login/About keys in the shipped document): every item keeps its path input empty, shows no preview, and says which sider logo the page borrows.

detail logos inherited

| Own path entered (preview, no hint) | After Reset (key removed, hint only) |
|---|---|
| own value | cleared |

## Checklist

- [ ] 1. Open Settings → Branding → "Login Logo (dark mode)" and confirm the path input is empty, no preview is shown, and the hint reads "No value is set, so the Light Mode Logo image is used."
- [ ] 2. Enter a path in "Login Logo (dark mode)" and confirm the hint disappears and a preview of the entered image appears.
- [ ] 3. Press Reset on "Login Logo (dark mode)" and confirm the input empties, the preview disappears, the hint returns, and JSON Config shows no `loginLogoSrcDark` key.
- [ ] 4. Clear the path by hand and blur the field and confirm the same hint-only state as step 3.
- [ ] 5. Open the About Backend.AI modal in light and in dark mode and confirm the logo renders with no broken image.

https://claude.ai/code/session_01CjJkAm7xtvPwNnC7xDDuNo


ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3852-pr9427-show-which-logo.sungchul.10-82-0-159.sslip.io)

agatha197 (approved):
LGTM

Issue #8532, opened 2026-08-03
Finish folder selection and sub-path components


nowgnuesLee:
Part of the Folder UX enhancement epic.

- Improve the folder selection and sub-path components for real-world use.
- Use a path picker for the sub-path input.

—

Captured while working on branch: main

JIRA Issue: FR-3442

PR #9452, opened 2026-09-07
feat(FR-3863): pick the mount subpath with BAIVFolderPathPicker

nowgnuesLee:
Part of #8532 (FR-3442). Linked Jira: [FR-3863](https://lablup.atlassian.net/browse/FR-3863), a sub-task of FR-3442.

> Jira Subtask issues are not cloned to GitHub, so there is no `Resolves #N` line; the parent epic task FR-3442 (#8532) tracks this work and is not closed by this PR.

## What

`BAIVFolderMountConfigInput` (BUI, not yet used by any host page) rendered the per-row mount **subpath** as a free-typed text input. This PR replaces it with `BAIVFolderPathPicker`, which browses the row's own vfolder in `BAIDirectoryPickerModal` and can only emit a path that exists.

- `BAIVFolderMountConfigInput`: the subpath column is now a `BAIVFolderPathPicker` bound to `entry.vfolderId` (`size="sm"` so it lines up with the alias input). New rows still seed `subpath: ''` (the vfolder root), which the picker shows as `/`. The emitted `VFolderMountConfigValue` contract and the exported helpers (`getVFolderMountConfigStatuses`, `isVFolderMountConfigValid`, `inputToMountDestination`) are unchanged; the absolute / `..` subpath validation stays for programmatically supplied values.
- `BAIVFolderPathPicker`: `onChange` keeps its optional parameter (`undefined` is reserved for a future clear action; the mount row maps it back to the root), and it takes `size` and `width` (default `'100%'`, like `BAIComplexSelect`) instead of deriving width from `style`. The only other call site (`DeploymentAddRevisionModal`) is untouched.
- Storybook: the mock Relay environment + mock `BAIClient` file tree wiring that both file-browsing stories hand-rolled is extracted into `src/tests/MockVFolderFileProviders.tsx` (self-requeueing resolver that re-arms the picker's preloaded query on every open) and `src/tests/mockVFolderFileTree.ts`.
- Locale (21 files): `comp:BAIVFolderMountConfigInput.SubpathPlaceholder` removed (unused), `SubpathTooltip` reworded for the picker.

## Next

FR-3864 (stacked on this) wires `BAIVFolderMountConfigInput` into the session launcher's Data & Storage step, sending the picked subpath as `mount_options[vfolderId].subpath` (manager >= 26.4.4).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===`
[...]

PR #9453, opened 2026-09-07
refactor(FR-3866): make BAIVFolderMountConfigInput fetch the gated REST folder list and render its own select

nowgnuesLee:
Part of #8532 (FR-3442). Linked Jira: [FR-3866](https://lablup.atlassian.net/browse/FR-3866), a sub-task of FR-3442.

> Jira Subtask issues are not cloned to GitHub, so there is no `Resolves #N` line; the parent task FR-3442 (#8532) is not closed by this PR.

## Why

`BAIVFolderMountConfigInput` picked folders with `BAIVFolderSelect` (GraphQL `vfolder_nodes`, pinned to the project scope), which was a stopgap. The session launcher's `VFolderTable` lists folders from REST `GET /folders?owner_user_email=` and applies gates the GraphQL select cannot express: the admin "launch as another user" owner email, the `allowed_vfolder_hosts` `mount-in-session` gate, project accessibility, and the auto-mounted dotfile folders. FR-3864 (next in the stack) needs those gates to adopt the mount config input in the launcher. When the v2 GraphQL folder API lands, the swap happens inside this one component.

## What

- **`BAIVFolderMountConfigInput` fetches the REST folder list itself** (`useSuspenseTanQuery`, `GET /folders`, `?owner_user_email=` when `ownerEmail` is given) and renders `BAIComplexSelect` inline with labeled values (`{ value: dashed UUID, label: name }`), so the folder name travels with the selection and no name lookup is needed. `BAIVFolderSelect` itself is untouched for its other consumers.
- **The mount gates come from the parent as props**: `mountableHosts` (hosts carrying `mount-in-session`), `currentProjectId` (project accessibility), `autoMountedFolderNames` (excluded from the options, since the session mounts them anyway, and used for the alias-overlap check), and a display-only `filter` predicate. No BUI data hook for these; the host already has them.
- **Stale selections are pruned by the component**: after the list loads, entries this owner/project cannot mount are dropped from the value with a warning (`comp:BAIVFolderMountConfigInput.UnmountableFoldersRemoved`), mirroring `VFolderTable`'s selection validation. The only callback is `onChange`.
- **Create and refresh buttons beside the select** (same `ButtonGroup` as `VFolderTable`): refresh re-runs the component's own `GET /folders` query through the tanstack `refetch` and shows `isFetching`; the create button appears only when the host passes `onClickCreateFolder`, since the folder-creation modal lives in the host app. A `ref.refetch()` handle lets the host refresh the list after creating a folder.
- **No `Suspense` inside the component**: it suspends on the folder list, and the consumer's boundary catches it, so a form cannot be submitted before the list has arrived and been pruned.
- **Mount-value vocabulary exported from the same module**: `DEFAULT_ALIAS_BASE_PATH`, `mountDestinationToInput` (inverse of `inputToMountDestination`, round-trip tested), a distinct `overlappingWithAutoMount` alias error kind with its own row message, `resolveVFolderMounts` for display rows, `toMountCreationConfig` for the `mount_ids` / `mount_id_map` / `mount_options` payload, and `useVFolderMountConfigFormRule` for a ready-made Form rule. FR-3864 consumes all of these.
- `react/src/components/VFolderSelect.tsx`'s `VFolder` type is now an alias of BUI's `LegacyVFolder` (single definition of the REST shape).
- Stories run on a REST folder mock (`src/tests/MockVFolderFileProviders.tsx`): interactive, prefilled, overlapping paths, auto-mounted folders, no mountable host, form validation via `useVFolderMountConfigFormRule`, disabled. The select label uses its own key, `comp:BAIVFolderMountConfigInput.SelectFolder`.

## Follow-ups (not in this PR)

- A "create folder, then select it" flow will need a refetch path for the folder list.

## Verification

- `pnpm test` in `packages/backend.ai-ui` → 76 files passed, 1 skipped (the `BAIComplexSelect.popup.test.tsx` flake showed once under load and is green in isolation)

[FR-3866]: https://lablup.atlassian.net/browse/FR-3866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nowgnuesLee:
Dev server on box `seungwon`: http://fr-3864-pr9455-mount-folders-session.seungwon.10-82-0-159.sslip.io
serves stack #9452 → #9453 → #9455 (running: #9455)

yomybaby (approved):
LGTM

PR #2061, opened 2023-11-29
feature: provide model service validation

lizable:
This PR resolves #2058.
Related: [backend.ai#1754](https://github.com/lablup/backend.ai/pull/1754)

> ⚠️ NOTE: you need to change the core branch to related PR.

- [x] Check the result of validation changes when click validate button (`processing...` -> `finished`)
- [x] Check you can execute "real" service when validation is on processing
- [x] Check you can validate service only if the model-definition file is in model-usage vfolder (refer [here](https://webui.docs.backend.ai/en/latest/model_serving/model_serving.html#uploading-model-definition-file-to-model-type-folder))
- [x] Check you can see the container log clearly after clicking validate button.

**Checklist:** (if applicable)

- [x] Mention to the original issue
- [ ] Documentation
- [x] Minimum required manager version
- [ ] Specific setting for review (eg., KB link, endpoint or how to setup)
- [x] Minimum requirements to check during review
- [ ] Test case(s) to demonstrate the difference of before/after

### Screenshot(s)
Screenshot 2024-05-16 at 6 29 02 PM
Screenshot 2024-05-16 at 6 30 23 PM

PR #9470, opened 2026-09-07
fix(FR-3487): close the actionable CodeQL alerts from the migration PR

### LINKED ISSUE #8644: Triage the six pre-existing CodeQL alerts surfaced by the Astryx migration PR
CodeQL reports these against [PR #8626](https://github.com/lablup/backend.ai-webui/pull/8626), but they are **not introduced by it** — every file below already exists on `main`. CodeQL attributes them to the PR because the diff is 3,356 files and it warns as much itself:

> _Alerts not introduced by this pull request might have been detected because the code changes were too large._

The three alerts that PR genuinely introduced were fixed in it (`form-engine/namePath.ts` prototype pollution, and two migration scripts). These six are split out so the migration PR does not carry unrelated security work.

## Alerts

| Severity | Rule | Location |
|---|---|---|

|—|—|—|
|---|---|---|

| high | `js/insecure-randomness` | `e2e/rbac/rbac-role-detail.spec.ts:678` |
|---|---|---|

| high | `js/insecure-randomness` | `e2e/utils/test-util.ts:118` |
|---|---|---|

| high | `js/insecure-randomness` | `e2e/credential/my-keypair-management.spec.ts:137` |
|---|---|---|

| high | `js/clear-text-storage-of-sensitive-data` | `packages/backend.ai-client/src/client.ts:564` and `:572` |
|---|---|---|

| high | `js/incomplete-sanitization` | `packages/backend.ai-docs-toolkit/src/website-builder.ts:2451` |
|---|---|---|

| medium | `js/prototype-polluting-assignment` | `packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx:309`, `:312` |
|---|---|---|

| medium | `js/prototype-pollution-utility` | `packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.tsx:312` |
|---|---|---|

## Suggested triage

Not all of these are equally real, and the ticket should end with a decision per group rather than seven identical fixes:

- **`js/insecure-randomness` in `e2e/**`** — `Math.random()` used to build unique test fixture names. No security boundary; the honest outcome is probably a dismissal ("used in tests") rather than a change. Worth confirming none of the three feeds a credential.
- **`js/clear-text-storage-of-sensitive-data` in `backend.ai-client`** — this one deserves a real look. It is library code that runs in the browser, and the rule fires on writing a secret to storage.
- **`js/prototype-polluting-assignment` in `BAIGraphQLPropertyFilter`** — same class as the one just fixed in the form engine, and the fix shape is known: keep the plain assignment for ordinary keys and define an own property for `__proto__` / `constructor` / `prototype`. See `packages/backend.ai-ui/src/form-engine/namePath.ts` (`safeAssign` / `safeRead`) for the pattern.
- **`js/incomplete-sanitization` in `website-builder.ts`** — check whether the escape is partial in the same way the codemod's was (escaping one metacharacter and leaving the rest).

## Reference

Branch alerts: <https://github.com/lablup/backend.ai-webui/security/code-scanning?query=pr%3A8626+tool%3ACodeQL+is%3Aopen>

JIRA Issue: FR-3487

yomybaby:
Resolves #8644 (FR-3487)

Triage of the six pre-existing CodeQL alerts that the Astryx migration PR (#8626) surfaced but did not introduce. Three groups had a real code fix and are fixed here; two are false positives whose dismissal rationale is written out below (per the ticket, the dismissal itself needs a human with Security-tab write access — no alert was dismissed via the API).

## Fixed

**Alerts 127 / 128 / 129 — `js/insecure-randomness` (`e2e/**`)**
These are one taint flow, not three unrelated ones. `Math.random()` builds a fixture *run id*, the run id becomes `FIXTURE_USERNAME` / `ASSIGN_FIXTURE_USERNAME`, and that username is passed to `login()` in `e2e/utils/test-util.ts` — which CodeQL treats as a security context. That is why alert 128 points at `test-util.ts:118` (the shared `username` parameter) rather than at any `Math.random()` call: it is the same flow reported on the shared sink. The brief for this ticket assumed 128 was a stale location; it is not.

Fix: source the suffix from `randomUUID()` (`node:crypto`) in the two spec files that own the flow. Same shape, same 4-character length, so fixture names and the `e2e-*` cleanup sweep are unchanged — only the entropy source differs. Removing the source clears all three alerts. The `Math.random()` uses elsewhere in `e2e/` are untouched: none of them reaches a credential parameter, which is why CodeQL never flagged them.

**Alerts 99 / 100 / 101 — `js/prototype-polluting-assignment`, `js/prototype-pollution-utility` (`BAIGraphQLPropertyFilter.tsx`)**
`buildNestedFilter` already had a reserved-key / empty-segment guard (added in #7717), but CodeQL does not recognise a `keys.some(callback)` test as a barrier, so the `current[keys[i]] = {}` walk stayed a sink. Rather than duplicate the `safeAssign` / `Object.defineProperty` shape from `form-engine/namePath.ts`, the nested object is now built inside-out from computed object literals: `{ [key]: nested }` always defines an *own* property and can never reach the prototype chain, so the sink disappears entirely. The existing guard stays as input validation. Output is byte-identical for every path, including the single-segment case the removed `keys.length === 1` branch handled.

**Alert 134 — `js/incomplete-sanitization` (`website-builder.ts:2451`)**
The guard regex escaped only `.`. That is sufficient in practice — `basePath` is validated against `^[A-Za-z0-9][A-Za-z0-9._-]*$` a few lines above and `.` is the only regex metacharacter that charset admits — but the escaping was incomplete on its own terms (CodeQL's message: "does not escape backslash characters"). Replaced with a local `escapeRegExp` over the full metacharacter class. The emitted regex is unchanged for every input the charset guard accepts.

## Dismissal rationale (needs a human in the Security tab)

**Alert 157 — `js/clear-text-storage-of-sensitive-data` (`safe-storage.ts:8`)** — false positive, already mitigated. The sink is the generic `localStorage.setItem(key, value)` wrapper; the eight reported sources are all request bodies flowing through `Client._wrapWithPromise` into the `backendaiwebui.logs` entry. That path calls `redactRequestParameters(...)` immediately before the write (`client.ts:553-556`), which deep-copies the payload and masks every key containing a sensitive stem — `password`, `old_password`, `new_password`, `new_password2`, `secret`, `token`, `authorization`, … in both snake_case and camelCase. `login()` additionally passes an explicit `log:` override with the password already replaced by `********` (`client.ts:1093-1098`). The behaviour is covered by `packages/backend.ai-client/src/redact.test.ts`. CodeQL does not model `redactRequestParameters` as a sanitizer because it returns `unknown` and rebuilds the object through `Object.entries`, so the flow is still reported. Suggested dismissal reason: **False positive**.

**Alert 168 — `js/incomplete-sanitization` (`review-overlay/cli.test.ts:357`)** — newer than this issue (it came in with #9438 on 2026-09-06) but a one-token fix, so it is taken here rather than left for a follow-up. The test fabricates a GitHub quote-escaped pin link with `url.replace('#', '%23')`, a string pattern that replaces only the first `#`; the sibling `&` / `=v3` replacements already use `/…/g`. Now `replace(/#/g, '%23')`. Today's pin-set URL carries exactly one `#`, so the fixture is byte-identical — `pnpm exec vitest run vite-plugins/review-overlay/cli.test.ts`: 22 passed.

## Also in this PR — drop the unused `ansi_up` dependency

`ansi_up@6.0.6` was added in #2061 for `ServiceValidationView`, whose last consumer was removed in #7325. Nothing imports it any more (container logs render ANSI through `@melloware/react-logviewer`), but the dependency, its pnpm patch (`react/patches/ansi_up@6.0.6.patch`) and the `patchedDependencies` entry were left behind. Removed here; `pnpm-lock.yaml` regenerated (ansi_up entries only) and `pnpm install --frozen-lockfile` passes.

Context: CodeQL alerts **42 / 46** point at the *vendored* copy `src/lib/ansiup.js`, which #7717 already deleted. They still show as open only because of a stale instance from the removed `.github/workflows/codeql-analysis.yml` configuration (last analysed 2024-09-10); the current default-setup analysis already reports both as fixed. Nothing to fix in code — they need the old configuration's analyses deleted (or a dismissal) in the Security tab, alongside alert 157.

## Tests

- `pnpm --filter backend.ai-ui exec vitest run src/components/BAIGraphQLPropertyFilter.test.tsx` — 49 passed. Added a three-level nested path and a reserved key in a non-leading segment (`project.__proto__`, `project.__proto__.polluted`), which the previous cases did not cover.
- `pnpm exec tsx --test src/website-builder.test.ts` (in `packages/backend.ai-docs-toolkit`) — 89 passed. Added "treats a dotted basePath as a literal path segment": `/v26.4/` redirects, `/v26x4/` does not, so the `.` is proven escaped rather than wildcarding.
- e2e specs were not executed (they need a live cluster). The change is confined to how the fixture run id is generated.

## Verification

```
bash scripts/verify.sh
...
=== ALL PASS ===
```

No CSS, theme tokens or `var(--…)` were touched, so the Astryx token gate does not apply.

## Review notes

- `buildNestedFilter` is used by every filterable table via `BAIGraphQLPropertyFilter`. The thing to check is that the emitted filter object is unchanged: `buildNestedFilter('project.name', {eq:'x'})` → `{project:{name:{eq:'x'}}}`, `buildNestedFilter('name', {eq:'x'})` → `{name:{eq:'x'}}`. The round-trip fixtures per page already in the test file cover the real call sites.
- The e2e change is one line per file plus an import; confirm the fixture email/username still fit the `e2e-*` prefix the global-cleanup teardown sweeps.
- For the two dismissals, the evidence is `client.ts:553-556` (redaction before the write) and `client.ts:1093-1098` (masked login log) — worth a look before dismissing alert 157 in the Security tab.

- [ ] Documentation


ironAiken2 (approved):
LGTM

PR #9471, opened 2026-09-07
feat(FR-347): show associated projects on the container registry list

### LINKED ISSUE #3010: Admin can manage associated groups for Container Registries
Following the implementation of the API for associating and disassociating container registries with groups, a corresponding UI is needed to manage these relationships effectively.

**Required Features**

1. **Display Group Information in Container Registry List:**
   - Add a new column in the container registry list to display associated groups.
   - If there are multiple groups, display in the format "default group 외 x개".
1. **Modify Container Registry Creation/Editing Modal:**
   - Allow users to specify associated groups using a multi-select component.
   - Use a modified `ProjectSelect` component to display all selected groups.
   - Ensure initial values are displayed correctly by possibly creating a new `ProjectSelectNode` component that uses the `group_nodes` API.

**Impact**

- Users will be able to view and manage group associations directly from the UI, enhancing usability and control over container registry access.

**Testing Scenarios**

- Verify that the group information is displayed correctly in the container registry list.
- Ensure that users can select and modify group associations in the creation/editing modal.
- Test the new `ProjectSelectNode` component for correct functionality and display.

**Affected Files**

- `react/src/components/ContainerRegistryList.tsx`
- `react/src/components/ContainerRegistryEditorModal.tsx`

JIRA Issue: FR-347

yomybaby:
Resolves #3010 (FR-347)

The container registry create/edit modal has let admins pick the projects a registry is allowed in since FR-2171, but the **list** gave no way to see the result — an admin had to open every registry's modal one by one to learn which projects it was associated with. This PR adds the missing half.

## What changed

- `ContainerRegistryListQuery` now selects `is_global @since("24.09.0")` and `allowed_groups @since("25.3.0") { edges { node { id name } } }` on each registry node.
- A new **Allowed Projects** column sits between *Project* and *Username*, rendered with `BAITagList` (`variant="text"`, `maxInline={1}`): the first project name shows inline and the remainder collapse into a hover `+N` popup — the same dense-cell pattern `BAIAdminUserV2Table` uses.
- A registry with `is_global` renders `All projects` instead of the tag list, matching the modal's `is_global` semantics (when global, no association rows exist).

## Design decisions

- **No new i18n keys.** The column title reuses `registry.AllowedProjects` (the modal's own label) and the global case reuses `environment.AllProjects`, so the wording stays identical to what the modal already shows.
- **Minimal per-row selection** (`id`, `name` only, no pagination args) to keep the added per-row resolver cheap.
- **Version tolerance**: both fields are `@since`-gated, so older managers strip them and they arrive `undefined` — `record.is_global` falls through to the tag list, whose items are `_.compact(...)` of an empty edge list, so the cell renders the `-` empty state rather than crashing.
- The column joins the column-settings modal automatically (`useHiddenColumnKeysSetting` derives its key set from the `columns` array), so no extra registration was needed.
- The editor modal is untouched.

## Tests

- `bash scripts/verify.sh` (Relay compile + drift check, lint, format, TypeScript, Astryx/theme gates, terminology).
- No unit test added: the change is a declarative column definition with no branching logic beyond the `is_global` ternary, and `ContainerRegistryList` has no existing test harness.
- No e2e run (requires a live cluster).

## Verification

```
=== ALL PASS ===
```

## Review notes

- Open **Administration → Registries**: the table should show an *Allowed Projects* column after *Project*. A registry associated with several projects shows the first name plus a `+N` chip; hovering the chip lists the rest.
- Set a registry to *Set as Global Registry* in the edit modal and save — its cell should read *All projects*. Uncheck it and pick two projects — the cell should show `first-project +1`.
- Against a manager older than 25.3.0 the column should render `-` (fields stripped by `@since`) rather than erroring; the column can also be hidden from the gear/column-settings modal.

- [ ] Documentation
- [x] Minium required manager version — `is_global` 24.09.0, `allowed_groups` 25.3.0 (both `@since`-gated, degrade to the empty state below that)
- [x] Minimum requirements to check during review — an admin account and at least one container registry associated with 2+ projects

### DIFF
--- a/react/src/__generated__/ContainerRegistryListQuery.graphql.ts
+++ b/react/src/__generated__/ContainerRegistryListQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<932c8525c83cc9568f3aafc94900516f>>
+ * @generated SignedSource<<2d37ba9a0a1dbfeeb0fd30b73bff9234>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -22,7 +22,16 @@ export type ContainerRegistryListQuery$data = {
     readonly count: number | null | undefined;
     readonly edges: ReadonlyArray<{
       readonly node: {
+        readonly allowed_groups: {
+          readonly edges: ReadonlyArray<{
+            readonly node: {
+              readonly id: string;
+              readonly name: string | null | undefined;
+            } | null | undefined;
+          } | null | undefined>;
+        } | null | undefined;
         readonly id: string;
+        readonly is_global: boolean | null | undefined;
         readonly name: string | null | undefined;
         readonly password: string | null | undefined;
         readonly project: string | null | undefined;
@@ -168,10 +177,17 @@ v16 = {
   "alias": null,
   "args": null,
   "kind": "ScalarField",
-  "name": "count",
+  "name": "is_global",
   "storageKey": null
 },
 v17 = {
+  "alias": null,
+  "args": null,
+  "kind": "ScalarField",
+  "name": "count",
+  "storageKey": null
+},
+v18 = {
   "alias": null,
   "args": [
     {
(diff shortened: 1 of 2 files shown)


yomybaby:
Closing as superseded. Everything this PR set out to add is already on `main` via #9637 (FR-3911, merged 2026-09-16):

- the container registry list queries `is_global` and `allowed_groups` (with an explicit `first:` cap and `count`, so registries linked to more than the default page size are counted correctly — which also covers the open Copilot thread here)
- an **Allowed Projects** column shows "All projects" for global registries and otherwise up to 3 project badges plus the remaining count
- a hidden-by-default `is_global` column

One difference worth noting: this PR gated the fields with `@since` for older managers; the version on `main` does not. If managers older than 25.3.0 still need to be supported, that should be a separate follow-up.

PR #9476, opened 2026-09-07
fix(FR-3737): name why Add Rules / Add Revision / Add Access Token are disabled

### LINKED ISSUE #9203: Deployment detail page: Add Rules / Add Revision / banner Add Access Token render disabled with no reason
Follow-up to FR-3679, which fixed the missing-reason defect only on the Access Tokens card. Three more controls on the same Deployment detail page render disabled with no indication of **why**, so the user cannot learn what unblocks them.

## Affected controls

**1. Add Rules** — `react/src/components/DeploymentAutoScalingCard.tsx:306`

```
disabled={isEndpointDestroying || !isOwnedByCurrentUser}
```

The clearest case: these are the **same two conditions** FR-3679 turned into `mutationDisabledReason`, so the existing strings `deployment.accessToken.DeploymentStopped` and `deployment.accessToken.OnlyOwnerCanManage` already describe them. Only the card differed, which is why it was missed.

**2. Add Revision** — `react/src/components/DeploymentRevisionCard.tsx:97`

```
disabled={isAddRevisionDisabled}
```

fed from `react/src/pages/DeploymentDetailPage.tsx:436` as `isDeploymentDestroying || isProjectMismatch`.

**3. Add Access Token (private-deployment banner)** — `react/src/pages/DeploymentDetailPage.tsx:402`

```
disabled={isDeploymentDestroying}
```

Recorded as known residue in PR #9166's description. Note the current inconsistency: this banner button stays **enabled** for a non-owner while the Access Tokens card's button is disabled with the ownership reason — two controls that open the same modal disagree on their guard.

## Fix

Apply the pattern FR-3679 established: the reason string **is** the flag. `undefined` enables, a string disables and names why, and `disabled` is derived at the point of use (`disabled={!!reason`}), so "disabled with no reason" becomes unrepresentable. `BAIButton` forwards `title` to Astryx's `tooltip` (`BAIButton.tsx:179`), and Astryx swaps native `disabled` for `aria-disabled` when a tooltip is present so the reason stays keyboard-reachable while the control remains inert.

## Notes on scope

- **Add Revision crosses a component boundary.** The reason is known by the page but the button is drawn by the card, so `isAddRevisionDisabled?: boolean` must become `addRevisionDisabledReason?: string`. `react/src/pages/DeploymentDetailPage.test.tsx:124` asserts on the current prop and will need updating.
- `isProjectMismatch` is a **fourth** condition with no existing string — it needs one new i18n key across all 21 locales. The other conditions reuse FR-3679's keys.
- The keys currently live under the `deployment.accessToken.*` namespace. Reusing them for auto-scaling and revision controls warrants promoting them to a shared `deployment.*` namespace.

## Out of scope

The disabled-button press animation (scale 0.98 + darkening on mouse-down) is an Astryx-level defect tracked separately as FR-3727, and is not addressed here.

JIRA Issue: FR-3737

yomybaby:
Resolves #9203 (FR-3737)

FR-3679 made the reason string the disabled flag on the Deployment detail page's Access Tokens card, but three controls on the same page were left rendering disabled with no indication of what unblocks them. This applies the same idiom to all three: `undefined` enables, a string disables and names why, and `disabled` is derived at the point of use (`disabled={!!reason}`), so "disabled with no reason" cannot be expressed. `BAIButton` forwards `title` to Astryx's tooltip, which keeps the reason keyboard-reachable while the control stays inert.

**1. Add Rules** (`DeploymentAutoScalingCard`) — reuses the stopped / non-owner reasons, which are the exact two conditions it already guarded on.

**2. Add Revision** (`DeploymentRevisionCard`) — the page knows the reason and the card draws the button, so the boolean prop `isAddRevisionDisabled` becomes `addRevisionDisabledReason?: string`.

**3. Add Access Token in the private-deployment banner** (`DeploymentDetailPage`) — it now uses the same guard as the Access Tokens card's Create button, which it shares a modal with.

### Design decisions

- **Behaviour change on the banner button.** A non-owner previously saw the banner's Add Access Token button *enabled* while the Access Tokens card's equivalent was disabled with the ownership reason — two controls opening the same modal disagreeing on their guard, as the issue notes. The banner now derives the identical three-way guard (stopped → non-owner → endpoint-not-issued). In practice only the ownership branch is reachable today, because the banner itself only renders when the deployment is live and has an endpoint URL; the guard is kept complete so the two controls agree by construction rather than by coincidence.
- **i18n namespace.** `deployment.accessToken.DeploymentStopped` is promoted to `deployment.DeploymentStopped` — its text was already generic and it is now read by three cards. The translations move with it in all 21 locale files, so nothing needs re-translating. `deployment.accessToken.OnlyOwnerCanManage` stays where it is: its text names access tokens specifically, so the auto-scaling card gets a new generic `deployment.OnlyOwnerCanManage` instead. `deployment.AddRevisionDisabledProjectMismatch` is the one genuinely new condition (the issue's fourth case); it is phrased as the action that unblocks the button rather than as a restatement of the mismatch banner above it.
- Non-`en`/`ko` locales carry `__NOT_TRANSLATED__` for the two new keys, written by `make i18n` as usual. `th` is not in the scanner's `lngs` list, so it only receives the key move.

## Tests

- `pnpm exec vitest run src/pages/DeploymentDetailPage.test.tsx src/components/DeploymentAccessTokensCard.test.tsx` → 2 files, 10 tests passed.
- Updated the existing `DeploymentRevisionCard` probe in `DeploymentDetailPage.test.tsx` to assert the reason string rather than a boolean, so the project-mismatch case now checks *which* reason the page hands down.
- Added two page-level tests for the private-deployment banner (`FR-3737` describe block): a non-owner gets `aria-disabled` plus the ownership reason in the DOM; the owner gets an enabled button.
- Updated the `deployment.DeploymentStopped` assertion in `DeploymentAccessTokensCard.test.tsx` for the key move.

## Verification

## Review notes

- Open a deployment you do not own and hover **Add Rules** in the Auto Scaling card: it stays inert and now shows "Only the deployment owner can manage this deployment."
- Open a deployment from a different project than the currently selected one (the "belongs to a different project" banner shows) and hover **Add Revision**: "Switch to this deployment's project to add a revision."
- On a private deployment with no access tokens yet, viewed by a non-owner, the banner's **Add Access Token** button is now disabled with the same reason the Access Tokens card's Create button already showed — that is the one behaviour change to sanity-check.

- [ ] Documentation


nowgnuesLee (approved):
LGTM

PR #9479, opened 2026-09-07
feat(FR-1973): default the session launcher to the minimum required allocation

### LINKED ISSUE #5146: 기본 자원 프리셋 최소 요구사항 할당으로 변경
### 요약

기본 자원 프리셋을 최소 요구사항 할당으로 변경하는 작업입니다.

### 배경

현재 자원 프리셋이 하나라도 있으면 그것을 기본으로 사용하도록 설정되어 있습니다. 하지만 기본으로 들어 있는 small 프리셋에서 validation 에러가 발생하고 있습니다. 프리셋을 수동으로 변경할 수 있지만, 아예 최소 요구사항 할당으로 기본 프리셋을 변경하는 것이 필요합니다.

### 수용 기준

- 기본 자원 프리셋이 최소 요구사항 할당으로 설정되어야 합니다.
- validation 에러가 발생하지 않아야 합니다.

### 기타 정보

없음.

[View original message posted by Sujin Kim on 2026. 1. 21. ](https://teams.microsoft.com/l/message/19:2806b848642744138244b7358cd282d9@thread.skype/1768985481043)[오후](https://teams.microsoft.com/l/message/19:2806b848642744138244b7358cd282d9@thread.skype/1768985481043)[ 5:51:21](https://teams.microsoft.com/l/message/19:2806b848642744138244b7358cd282d9@thread.skype/1768985481043)

JIRA Issue: FR-1973

yomybaby:
Resolves #5146 (FR-1973)

## What changed

The session launcher's initial allocation preset was `_.sortBy(allocatablePresetNames)[0]` — the alphabetically first allocatable preset ('small' in a default install, but whatever happens to sort first in a site's own preset list). That silently over-allocates for a user who just wants to start a session, and the pick is arbitrary rather than a curated default.

The launcher now starts on **Minimum allocation** (`minimum-required`), which allocates exactly what the selected image declares as its minimum. Nothing else about the field changes: the user can still switch to any preset or to Custom.

## Design decisions

- **The old alphabetical pick stays as the fallback.** `ResourcePresetSelect` renders the `minimum-required` option only when `baiClient._config.allowCustomResourceAllocation` is on. On a deployment with it off, defaulting to `minimum-required` would leave the select showing a value it does not list, so those deployments keep the previous behaviour (first allocatable preset, else `null`).
- **`enabledAutomaticShmem` is set to `true`** alongside the preset, mirroring what `ResourcePresetSelect`'s `onChange` already does for the `minimum-required` case. The resource fields themselves are filled by the existing image-driven effect, not here.
- **Prefilled forms are untouched.** The effect still returns early when the form already holds `custom` / `minimum-required` or a preset that is allocatable in the current resource group, so session-template and re-launch prefill keep their preset. `FileBrowserButton*` already hardcoded `allocationPreset: 'minimum-required'` and is unaffected.
- The selection moved into an exported pure helper, `getAutoSelectedAllocationPreset`, so both config branches are unit-testable without rendering the form.

## Tests

- Added 4 cases to `react/src/components/SessionFormItems/ResourceAllocationFormItems.test.ts` covering `getAutoSelectedAllocationPreset`: custom-allocation on (with and without allocatable presets), custom-allocation off (alphabetical fallback), and the `null` cases.
- `pnpm exec vitest run src/components/SessionFormItems/ResourceAllocationFormItems.test.ts` — 18 passed.
- `pnpm exec vitest run src/components/SessionFormItems/ResourceAllocationFormItems.contract.test.tsx` — 2 passed (unchanged).

## Verification

`bash scripts/verify.sh`:

```
=== ALL PASS ===
```

## Review notes

- Open **Start session** on a deployment with `allowCustomResourceAllocation = true`: the Resource presets field should land on *Minimum allocation*, and CPU/RAM/accelerator should match the selected image's minimum (change the image and confirm the numbers follow).
- Set `allowCustomResourceAllocation = false` in `config.toml` and reload: the field should still land on the alphabetically first allocatable preset, exactly as before — *Minimum allocation* is not even offered there.
- Re-launch an existing session and open a session template: both should still arrive with their own preset selected, not with *Minimum allocation*.

- [ ] Documentation
- [x] Specific setting for review (eg., KB link, endpoint or how to setup) — `allowCustomResourceAllocation` in `config.toml` toggles the two branches

### DIFF
--- a/react/src/components/SessionFormItems/ResourceAllocationFormItems.test.ts
+++ b/react/src/components/SessionFormItems/ResourceAllocationFormItems.test.ts
@@ -10,6 +10,7 @@ import {
 import { Image } from '../ImageEnvironmentSelectFormItems';
 import {
   getAllocatablePresetNames,
+  getAutoSelectedAllocationPreset,
   getUnifiedSlotNameFromTag,
   isUnifiedAcceleratorSlot,
 } from './ResourceAllocationFormItems';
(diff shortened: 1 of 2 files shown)


yomybaby:
After introducing AppConfig, we plan to allow you to select the default option through domain-specific settings.

PR #9484, opened 2026-09-07
refactor(FR-3418): migrate BAIAdminResourceGroupSelect to adminResourceGroups

### LINKED ISSUE #8475: Migrate BAIAdminResourceGroupSelect off the deprecated resourceGroups connection
## Problem

The shared `BAIAdminResourceGroupSelect` component in the UI library queries the `resourceGroups` connection, which the schema marks:

```
@deprecated(reason: "Use admin_resource_groups instead. This API will be removed after v26.3.0. See BEP-1041 for migration guide.")
```

Its replacement, `adminResourceGroups`, is documented as "Added in 26.2.0. List resource groups (admin only)" and takes the identical argument set and returns the same `ResourceGroupConnection`, so the swap is mechanical. The repo already uses the successor elsewhere (the Fair Share list and its usage-bucket chart both query `adminResourceGroups`), so this component is the straggler.

Two call sites are affected today: the RBAC role form modal, and — newly — the resource-preset editor, which adopted this component when the preset editor stopped scoping resource groups by project (resource presets have no project dimension: the backend's `resource_presets` table relates only to a scaling group, and a null `scaling_group_name` means the preset is global).

The deprecated field still exists in the current schema, so nothing is broken today; this is about not accumulating call sites on a field that is slated for removal.

## Solution

Point `BAIAdminResourceGroupSelect` at `adminResourceGroups`, keeping the same connection/pagination/search behaviour, and move its `@since` gate from `26.1.0` to `26.2.0` to match the successor's introduction. Both existing call sites inherit the fix with no API change on the component itself.

Note the one behavioural consequence: the component's minimum manager version rises by one minor (26.1.0 → 26.2.0). On managers below that the client directive drops the field and the select lists nothing — the same graceful-degradation path it already has today, one version higher.

## Acceptance criteria

- [ ] `BAIAdminResourceGroupSelect` queries `adminResourceGroups` with the `@since` gate matching its introduction version
- [ ] Pagination, server-side search, and the selected-value behaviour are unchanged for both existing call sites
- [ ] No remaining query in the repo targets the deprecated `resourceGroups` connection
- [ ] Relay artifacts regenerated and committed; verification harness passes

## Blocked by

None — can start immediately. Independent of the FR-3407 epic, though the preset-editor call site arrives with FR-3415.

JIRA Issue: FR-3418

yomybaby:
Resolves #8475 (FR-3418)

`BAIAdminResourceGroupSelect` was the last source-level user of the deprecated
`resourceGroups` connection (`data/schema.graphql`: *"Use admin_resource_groups
instead. This API will be removed after v26.3.0. See BEP-1041 for migration
guide."*). This points its pagination fragment at the successor
`adminResourceGroups`, which has an identical argument list and returns the same
`ResourceGroupConnection`.

### Design decisions

- **Selected under its own name, no alias.** The field is selected as `adminResourceGroups(...)`
  (the `resourceGroups:` alias was dropped per review). The `@connection` key follows the field —
  `BAIAdminResourceGroupSelect_adminResourceGroups`, since Relay requires the key suffix to match
  the selected field or alias — and the component's two `data.resourceGroups` reads become
  `data.adminResourceGroups`. The fragment name and both call sites (`RoleFormModal`,
  `AgentSettingModal`) are untouched, so the diff is the component file and its regenerated artifacts.
- **`@since` raised 26.1.0 → 26.2.0**, matching the version in which
  `adminResourceGroups` was added.
- **Pagination stays cursor mode** (`first`/`after`) — `usePaginationFragment`
  requires it, and mixing in `limit`/`offset` would be rejected at runtime by the
  V2 connections (`.claude/rules/graphql-pagination.md`).

### Minimum manager version

This select's minimum manager version rises **26.1.0 → 26.2.0**. On a 26.1.x
manager the `@since` transformer strips the field from the document and the picker
renders an empty list — the same graceful-degradation path it already had, one
minor higher. Worth a release-note line.

## Tests

- `pnpm run relay` — regenerated the four affected artifacts (2 in `backend.ai-ui`,
  2 in `react`, from the two call-site queries).
- `bash scripts/verify.sh` — Relay, Lint, Format, TypeScript, Astryx gates, terminology.
- No new unit test: the change is the `graphql` tag plus the two matching `data.*` reads;
  the component's logic is unchanged.

## Verification

```
=== ALL PASS ===
```

## Review notes

- The real review surface is the `graphql` tag and the two `data.adminResourceGroups` reads in
  `packages/backend.ai-ui/src/components/fragments/BAIAdminResourceGroupSelect.tsx`;
  the `__generated__` churn is `pnpm relay` output (see `.claude/rules/review-ignored-paths.md`).
- Confirm the connection handle key in the generated artifacts is
  `"key": "BAIAdminResourceGroupSelect_adminResourceGroups"` (renamed with the field; the
  Relay store is in-memory, so nothing persisted depends on the old key) and `loadNext` keeps working.
- Manually: open **Admin → Roles → role form** (resource-group scope picker) and the
  **agent setting modal**, against a manager ≥ 26.2.0. The list, the type-to-search
  (`name contains` refetch) and scroll-to-load-more should behave exactly as before.
- `grep -rn 'resourceGroups(' react/src packages/backend.ai-ui/src` now returns no hits.

- [ ] Documentation
- [x] Minium required manager version — 26.1.0 → 26.2.0 for this component
- [x] Minimum requirements to check during review — manager ≥ 26.2.0

### DIFF
--- a/packages/backend.ai-ui/src/__generated__/BAIAdminResourceGroupSelectPaginationQuery.graphql.ts
+++ b/packages/backend.ai-ui/src/__generated__/BAIAdminResourceGroupSelectPaginationQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<f61dc7b60b11c952e23ba36b8b623a93>>
+ * @generated SignedSource<<ef152d2696065fbbf098dc4a464b53c0>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -112,7 +112,7 @@ return {
     "name": "BAIAdminResourceGroupSelectPaginationQuery",
     "selections": [
       {
-        "alias": "resourceGroups",
+        "alias": null,
         "args": (v1/*: any*/),
         "concreteType": "ResourceGroupConnection",
         "kind": "LinkedField",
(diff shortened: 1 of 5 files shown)


ironAiken2 (approved):
LGTM

PR #9494, opened 2026-09-07
fix(FR-3587): render the fallback image meta icon as a themed glyph

### LINKED ISSUE #8880: Session Launcher: the default </> image meta icon is too dark in dark mode
## Where

- **Page**: `/session/start` (menu: Sessions → Start) — Session Launcher, Environments step and the Review step
- **Region**: the image meta icon that precedes the image name in the environment select, its closed trigger, and the review step's Image row
- **Component** (if identified): `ImageMetaIcon` (`react/src/components/ImageMetaIcon.tsx`); it renders an `` whose source comes from the image metadata (`resources/image_metadata.json` → `resources/icons/*`)

## Observed

In dark mode the `</>` icon shown for images without a vendor-specific icon

(the default glyph) reads as very dark against the dark surface, noticeably

dimmer than the text and badges beside it.

## Expected

The icon should stay legible in dark mode — reviewer asked for it to be

"lighter", i.e. to carry a comparable contrast to the adjacent image name.

## Environment

- **Mode**: dark
- **Viewport**: Unknown
- **Browser/OS**: Unknown
- **Build**: branch `FR-3544` (PR #8783), 2026-08-19

## Reproduction

1. Switch the UI to dark mode.
1. Open Sessions → Start.
1. In the Environments step, pick an image that uses the default `</>` icon

(e.g. `aimet`), and look at the icon in the select, in the closed trigger,

and on the Review step's Image row.

## Evidence

- Raised in review of PR #8783: https://github.com/lablup/backend.ai-webui/pull/8783#pullrequestreview-3555553143

## Notes

Reported against the Session Launcher, but the same component renders the icon

on every image meta row, so other screens may show it too (usage breadth from

grep, not a reproduced claim). Possibly out of Astryx's theming reach — the

glyph is a raster asset, not a themed icon — but that is for the fix to settle.

JIRA Issue: FR-3587

yomybaby:
Resolves #8880 (FR-3587)

The default `</>` image meta icon is `resources/icons/default.png` — a 32x32
grayscale PNG of dark ink on transparency. A raster asset cannot follow the
theme, so on the Session Launcher's dark surfaces the fallback glyph is barely
visible.

`ImageMetaIcon` (host) and its BUI twin `BAIImageMetaIcon` now render an inline
lucide `CodeXml` glyph — the same `</>` shape — painted with the
`--color-text-primary` token when the image metadata declares no vendor icon.
The raster fallback is no longer used on that path, which takes it out of the
theming problem entirely. Images that resolve a real vendor icon still render
that asset untouched.

Design decisions:

- The fallback branch is selected by a new `hasImageIcon()` helper on both
  `useBackendAIImageMetaData` and `useBAIImageMetaData`, rather than comparing
  the resolved URL against the `default.png` filename.
- The glyph keeps the same `1em` box and `verticalAlign: middle`, and merges the
  caller's `style` exactly like the `` did, so the call sites that size it
  down (`ImageEnvironmentSelectFormItems` passes `15x15`) are unaffected.
- `--color-text-primary` (not `secondary`) was chosen so light mode looks
  essentially unchanged — the PNG was near-black ink — while dark mode inverts.
- `BAIImageMetaIcon` still renders `null` when the host provided no `imagePath`,
  so its documented "renders nothing outside `BAIMetaDataProvider`" contract is
  unchanged; only the fallback rendering moved.

## Tests

- New `packages/backend.ai-ui/src/components/BAIImageMetaIcon.test.tsx` (3 cases:
  declared vendor icon renders ``, undeclared icon renders the themed glyph,
  no `imagePath` renders nothing) — `pnpm --filter backend.ai-ui exec vitest run
  src/components/BAIImageMetaIcon.test.tsx` → 3 passed.
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → `undeclared: 2`,
  identical to the pre-existing baseline on `main` (no new findings).

## Verification

## Review notes

- Open the Session Launcher in dark mode and pick an image with no vendor icon
  in `resources/image_metadata.json` (any `imageInfo` entry without an `icon`
  key): the `</>` glyph should read at the same contrast as the image name next
  to it. Toggle to light mode — it should look as before.
- Check the other surfaces that render these icons for unchanged layout: the
  environment select dropdown (`ImageEnvironmentSelectFormItems`, sized 15x15),
  the session list image tags, and `BAIImageNodeSimpleTagV2`.
- Confirm a well-known image (e.g. a TensorFlow or PyTorch entry) still shows its
  coloured vendor logo, not the glyph.

- [ ] Documentation

### DIFF
--- a/packages/backend.ai-ui/src/components/BAIImageMetaIcon.doc.ts
+++ b/packages/backend.ai-ui/src/components/BAIImageMetaIcon.doc.ts
@@ -16,7 +16,7 @@ export const docs = {
   ],
   usage: {
     description:
-      'The framework icon for a Backend.AI container image. It reads image metadata through `useBAIImageMetaData()` and joins the host-supplied `imagePath` with the icon filename declared for that image (falling back to `default.png`), then renders a 1em square `<img>` aligned to the surrounding text. It renders nothing when the host application has not provided an `imagePath` — the package resolves no asset paths of its own — so a consumer must sit under `BAIMetaDataProvider` for an icon to appear.',
+      'The framework icon for a Backend.AI container image. It reads image metadata through `useBAIImageMetaData()` and joins the host-supplied `imagePath` with the icon filename declared for that image, then renders a 1em square `<img>` aligned to the surrounding text. An image whose metadata declares no icon renders a themed inline `</>` glyph instead of a raster fallback, so it stays legible in both light and dark themes. It renders nothing when the host application has not provided an `imagePath` — the package resolves no asset paths of its own — so a consumer must sit under `BAIMetaDataProvider` for an icon to appear.',
     bestPractices: [
       {
         guidance: true,
--- a/packages/backend.ai-ui/src/components/BAIImageMetaIcon.test.tsx
+++ b/packages/backend.ai-ui/src/components/BAIImageMetaIcon.test.tsx
@@ -1,0 +1,60 @@
+import BAIImageMetaIcon from './BAIImageMetaIcon';
+import { BAIMetaDataProvider } from './provider';
+import type { ImageMetaData } from './provider';
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+
+const imageMetaData: ImageMetaData = {
+  imageInfo: {
+    python: { name: 'Python', description: '', group: '', tags: [] },
+    tensorflow: {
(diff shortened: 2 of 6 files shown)


agatha197 (approved):
LGTM

PR #9498, opened 2026-09-07
fix(FR-3723): keep the edit-announcement body a skeleton until Monaco is ready

### LINKED ISSUE #9171: Edit announcement modal: form and editor render during Suspense instead of a skeleton-only body
While the **Edit announcement** modal is still loading (Suspense pending), it already renders the full form chrome — the content editor with its toolbar and an empty canvas, and an empty **Preview** pane — with a few skeleton bars floating inside the editor area. The result looks broken: empty widgets plus stray placeholder bars.

**Expected:** while suspended, the modal body renders only a skeleton (no editor, no toolbar, no preview pane); the real form appears once the data has resolved.

**Observed:** editor + preview render empty with skeleton bars inside (see attached screenshot).

—

JIRA Issue: FR-3723

yomybaby:
Resolves #9171 (FR-3723)

## What changed

The Edit-announcement modal already gated its body on the announcement query
(`isLoading ? <BAISkeleton rows={4}/> : <form>`), but there is a **second**,
nested loading phase: once the query resolves, the body immediately renders the
markdown toolbar and an empty Preview pane, while `BAICodeEditor` lazily imports
Monaco and paints its own skeleton bars *inside* the bordered editor box. That
intermediate frame — empty widgets plus stray placeholder bars — is what the
issue reports.

The body now waits for both phases:

- `MarkdownEditorField` gained an `onReady` callback, fired from the `onMount`
  it already had, so the modal knows when Monaco has actually mounted.
- The two-pane body is **mounted but hidden** (`display: none`) until the query
  has resolved *and* the editor has mounted, so Monaco boots behind the
  skeleton rather than in front of the user. Only `<BAISkeleton rows={4} />`
  is visible until then.
- Monaco measures its container on mount, and a hidden mount measures 0, so the
  editor is given `automaticLayout: true` to re-measure when the body is
  revealed.
- The footer's Delete and Publish buttons follow the same `isBodyReady` flag
  instead of `isLoading` alone — a modal whose body is still a skeleton should
  not offer either action.

## Design decisions

- **Kept modal-local.** Changing `BAICodeEditor`'s own fallback would ripple to
  its other call sites (ShellScriptEditModal, ContainerRegistryEditorModal,
  BAIJSONViewerModal), where the inner skeleton is the right default. Only this
  modal wants the whole body to wait.
- **Hidden-but-mounted rather than deferred mount.** The editor has to be in the
  tree for Monaco's lazy chunk to load and report readiness; unmounting it until
  "ready" would never resolve.

## Tests

- New `react/src/components/AnnouncementEditModal.test.tsx` (vitest) walks the
  three phases: query in flight → body absent; query resolved but editor not
  mounted → body present in the DOM yet not visible, Publish disabled; editor
  mounted → body visible, Publish enabled. Verified it **fails** on the previous
  behaviour (removing the `display` gate makes the phase-2 `not.toBeVisible()`
  assertion fail) and passes with the fix.
- `pnpm exec vitest run src/components/AnnouncementEditModal.test.tsx` → 1 passed.

## Verification

`bash scripts/verify.sh` →

```
=== ALL PASS ===
```

No CSS/token changes, so the Astryx token gate is unaffected.

## Review notes

- Open **Settings → Maintenance → Edit announcement** (or the pencil on the
  announcement banner) with the network throttled. The body should be a single
  skeleton block the whole time — no empty editor frame, no empty Preview pane,
  no skeleton bars inside a bordered box.
- When the editor appears, confirm it is laid out at full height/width (this is
  what `automaticLayout` buys us after the hidden mount) and that typing updates
  the Preview pane.
- Delete/Publish are disabled while the skeleton is showing and enable together
  with the body.

- [ ] Documentation


react/src/components/AnnouncementEditModal.tsx
```diff
@@ -212,10 +217,17 @@ const AnnouncementEditModal: React.FC<AnnouncementEditModalProps> = ({
...
+      {!isLoading && (
+        // Kept mounted but hidden while Monaco boots, so the editor loads
+        // behind the Skeleton instead of flashing an empty frame.
+        <BAIFlex
+          direction="row"
+          align="stretch"
+          gap="sm"
+          wrap="wrap"
+          style={{ display: isBodyReady ? undefined : 'none' }}
+        >
```
agatha197: Two gates for one state here. `!isBodyReady` hides, !isLoading unmounts. Do we need both?

agatha197 (changes requested):

react/src/components/AnnouncementEditModal.tsx
yomybaby: Good catch — the mount gate was redundant. Dropped it in `621b2a6`.

`BAICodeEditor` hands `value` straight to `@monaco-editor/react`, which is controlled, so mounting the body before the announcement query resolves is safe: the message arriving later just updates the model. Only the visibility gate (`isBodyReady`) remains, used for both the Skeleton and the body.

Side benefit: Monaco's lazy chunk now loads in parallel with the query instead of after it, so the modal reaches its ready state sooner. The visible sequence is unchanged (Skeleton until both phases finish), and the test now asserts phase 1 the same way as phase 2 — mounted, not visible, Publish disabled — instead of asserting the body is absent. Verified it still fails without the `display` gate.

yomybaby:
Follow-up: found and fixed a layout bug in this PR while checking it on the dev server — the preview pane was not visible.

**Cause.** The visibility gate was written as `display: isBodyReady ? undefined : 'none'`, but `BAIFlex` composes its style as `{ display: 'flex', ...style }` (packages/backend.ai-ui/src/components/BAIFlex.tsx:111,116). The `undefined` branch therefore deleted BAIFlex's own `display` rather than leaving it alone: the revealed body computed to `display: block`, so the editor and the preview stacked vertically and the preview landed ~700px down inside a 746px modal body — off screen.

**Fix** (`33d928e1b`): gate with `'flex'` instead of `undefined`, and pin it in the test — the revealed row is now asserted to still carry `display: flex`, which fails if the `undefined` form returns.

Both fixes verified: the modal shows the editor and preview side by side again on the dev server. Also rebased onto current `main` (`b8ee73ac2`).

Note on CI: `scripts/verify.sh` fails one step on this branch — `Search index` — and it is **not from this PR**. `main` currently ships `resources/i18n/en.json` with the FR-2898 keys (#7422) while `react/src/generated/searchIndex.json` was never regenerated, so the check is red on `main` itself. Left out of this PR deliberately; it needs its own `pnpm run search-index` commit.

agatha197:
@copilot fix ci failure.

agatha197:
Dev server on box `sujin`: stopped (was http://fr-3723-pr9498-keep-edit-announcement.sujin.10-82-0-159.sslip.io)

agatha197 (approved):
LGTM

PR #9500, opened 2026-09-07
fix(FR-3836): propagate EXT_HTTP_PROXY to the API-mode TCP gateway

### LINKED ISSUE #9364: Propagate EXT_HTTP_PROXY to the API-mode TCP gateway in the Electron local proxy
In the Electron local proxy, API mode hands the TCP gateway `aiclient._config` (`src/wsproxy/manager.js:320`) — a `ClientConfig`, which has no `ext_proxy_url` field — so the external-proxy branch in `src/wsproxy/lib/backend.ai-ws-appproxy.js` (`this._env.ext_proxy_url`) never fires and the `EXT_HTTP_PROXY` setting is silently ignored for API-mode app streams. SESSION mode is unaffected (its gateway receives the `cf` object, which carries `ext_proxy_url`).

History: before FR-3796 this path crashed outright (`this._env` was never assigned), so nothing regressed — the crash fix surfaced this second, pre-existing gap. Raised by human review on the FR-3796 stack (PR #9290) as the one blocking note; deferred here because the fix is a functional change (plumb the proxy URL through `ClientConfig`, or pass the full proxy configuration to the gateway) and verifying it needs an API-mode deployment behind a corporate HTTP proxy.

## Task

- Carry `ext_proxy_url` (manager.js `this.extHttpProxyURL`) into whatever the API-mode gateway receives, so the `HttpsProxyAgent` branch fires when `EXT_HTTP_PROXY` is set.
- Verify against an API-mode deployment behind an external HTTP proxy (TCP app stream, e.g. SSH or VSCode Desktop).
- Remove the `TODO(FR-…)` marker at the assignment site in `backend.ai-ws-appproxy.js`.

JIRA Issue: FR-3836

yomybaby:
Resolves #9364 (FR-3836)

In the Electron local proxy, API mode handed the TCP gateway `aiclient._config` — a `ClientConfig`, which has no `ext_proxy_url` field — so the external-proxy branch in `src/wsproxy/lib/backend.ai-ws-appproxy.js` never fired and `EXT_HTTP_PROXY` was silently ignored for API-mode app streams (SSH, VSCode Desktop, …). SESSION mode was unaffected: its gateway receives the `cf` object, which carries `ext_proxy_url`.

## What changed

- `src/wsproxy/manager.js` — the API-mode branch now passes `this.extHttpProxyURL` alongside `aiclient._config`.
- `src/wsproxy/gateway/tcpwsproxy.js` — stores the URL and forwards it to the app-proxy server.
- `src/wsproxy/lib/backend.ai-ws-appproxy.js` — the constructor takes the optional URL and resolves `extProxyURL ?? env.ext_proxy_url`; the `HttpsProxyAgent` branch reads that. The `TODO(FR-3836)` marker is removed.
- `src/wsproxy/lib/backend.ai-ws-appproxy.js` + `src/wsproxy/lib/console-appproxy.js` — `require('https-proxy-agent')` → `const { HttpsProxyAgent } = require(…)`.

### Design decisions

- **The URL travels as its own constructor argument, not on `ClientConfig`.** `aiclient._config` is the live client's config object, shared with every other request; adding `ext_proxy_url` to it would leak the setting into unrelated consumers. The explicit argument keeps the change additive, and the `??` fallback means SESSION mode (which arrives with `cf.ext_proxy_url`) is untouched.
- **The `https-proxy-agent` import fix is in scope, not drive-by.** The package is at v7, which exports `HttpsProxyAgent` as a **named** export. The existing default-style `require` bound the module namespace object, so `new HttpsProxyAgent(url)` throws `TypeError: … is not a constructor`. Verified locally against the installed 7.0.6:

  ```
  default-call FAILS: m is not a constructor
  named-call OK
  ```

  Without this, the branch this PR enables would crash on its first use. `console-appproxy.js` (the SESSION-mode twin, which already receives the URL today) had the identical latent break, so it is fixed in the same one-line shape.

## Tests

- Two new unit tests in `src/wsproxy/manager.test.ts` pin both hops of the wiring:
  - the API-mode `/add` route constructs the TCP gateway with `(aiclient._config, extHttpProxyURL)`;
  - `tcpwsproxy.start_proxy()` forwards `(env, extProxyURL)` to the app-proxy server.

  Both gateway modules require `backend.ai-client`, a build artifact the source-only checkout does not have (the suite's existing top-of-file comment explains this constraint), so each test seeds `require.cache` with a stub for the module one hop down and asserts on the arguments that arrive there. Both tests were confirmed to fail against the pre-fix code (the gateway received a single argument).
- `npx vitest run src/wsproxy/manager.test.ts` → **20 passed** (18 pre-existing + 2 new).

**Not covered by CI:** end-to-end verification needs an API-mode Backend.AI deployment reachable only through an external HTTP proxy, driving a TCP app stream (SSH or VSCode Desktop) with `EXT_HTTP_PROXY` set. The unit tests are the automated evidence; that check still needs a human on a proxied deployment.

## Verification

`bash scripts/verify.sh` →

```
=== ALL PASS ===
```

## Review notes

- The behavioural change is one argument at `manager.js:320`; everything downstream is pass-through. Confirm `aiclient._config` is **not** mutated — that was the alternative approach and it was deliberately rejected.
- To exercise it manually: start the Electron app with `EXT_HTTP_PROXY=http://<proxy>:3128`, log in to an **API-mode** endpoint, and open an SSH / VSCode Desktop app. The local proxy logs `- try using external http proxy: <url>` when the branch fires; before this PR that line never appeared in API mode.
- SESSION mode should be unchanged — the same log line, same behaviour — since `extProxyURL` is `undefined` there and the `?? env.ext_proxy_url` fallback takes over.

- [ ] Documentation
- [x] Specific setting for review (eg., KB link, endpoint or how to setup) — needs `EXT_HTTP_PROXY` set and an API-mode endpoint behind that proxy; see Review notes.
- [x] Minimum requirements to check during review — Electron build, API-mode login, a TCP app (SSH / VSCode Desktop).
- [x] Test case(s) to demonstrate the difference of before/after — `src/wsproxy/manager.test.ts`, `EXT_HTTP_PROXY propagation (FR-3836)`.

### DIFF
--- a/src/wsproxy/gateway/tcpwsproxy.js
+++ b/src/wsproxy/gateway/tcpwsproxy.js
@@ -1,11 +1,14 @@
 const Server = require('../lib/backend.ai-ws-appproxy.js');
 
 module.exports = proxy = class Proxy {
-  constructor(env) {
+  // `extProxyURL` is passed separately because API mode's `env` is a
+  // ClientConfig, which has no place to carry it (SESSION mode's `cf` does).
+  constructor(env, extProxyURL) {
     this.env = env;
+    this.extProxyURL = extProxyURL;
   }
   async start_proxy(kernelId, app, ip, port, envs = {}, args = {}) {
-    this.c = new Server(this.env);
+    this.c = new Server(this.env, this.extProxyURL);
     return this.c.start(kernelId, app, ip, port, envs, args);
   }
 
--- a/src/wsproxy/lib/backend.ai-ws-appproxy.js
+++ b/src/wsproxy/lib/backend.ai-ws-appproxy.js
@@ -4,7 +4,7 @@ const WebSocket = require('ws');
 const ai = require('backend.ai-client');
 const bind = require('./bindStream');
 const htmldeco = require('./htmldeco');
-const HttpsProxyAgent = require('https-proxy-agent');
+const { HttpsProxyAgent } = require('https-proxy-agent');
 /*
 const i18next = require('i18next');
 const i18next_backend = require('i18next-sync-fs-backend');
@@ -24,11 +24,12 @@ i18next
   */
 
 module.exports = proxy = class Proxy extends ai.backend.Client {
-  constructor(env) {
+  constructor(env, extProxyURL) {
     super(env);
-    // TODO(FR-3836): in API mode `env` is a ClientConfig, which carries no ext_proxy_url,
-    // so the external-proxy branch in _start never fires (SESSION mode gets it via cf).
     this._env = env;
+    // API mode passes the URL explicitly (its `env` is a ClientConfig with no
+    // ext_proxy_url); SESSION mode carries it on the `cf` object it passes as `env`.
+    this._extProxyURL = extProxyURL ?? env.ext_proxy_url;
     this._running = false;
     this._resolve = undefined;
     this._connectionCount = 0;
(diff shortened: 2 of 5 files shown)


ironAiken2 (approved):
LGTM

PR #9502, opened 2026-09-07
fix(FR-3874): feed the staged value back into renderInput controls

### LINKED ISSUE #9501: Property filter renderInput select does not show the staged value after the PowerSearch migration
## Symptom

In `BAIPropertyFilter` / `BAIGraphQLPropertyFilter`, a property that renders a custom picker through `renderInput` (project select on the admin session list, user select on RBAC and the user folder permission panel, storage host select on the model store and admin model card) loses its selection visually:

- Pick an option in the popover: the Apply button enables, but the select snaps back to its placeholder, so the user cannot see what they picked before applying.
- Click an existing token to edit it: the select opens empty instead of pre-populated with the current value.

## Cause

Since the Astryx `PowerSearch` rebuild the picked value is staged in PowerSearch state and committed by the popover's Apply button. `useRenderInputEditors` in `BAIPowerSearchAdapters.tsx` receives that staged `value` from PowerSearch but only forwards `onChange`, and every call site still pins `value={null`} (the antd-era convention, where a pick committed immediately and the control had to clear itself).

## Fix

Forward `value` (and `isDisabled`) into the `renderInput` render prop and feed it back into the control at the five call sites. Serialization is unchanged.

This is a subset of FR-3691, split out so the visible defect can ship without the `entitySource` redesign.

—

Captured while working on branch: feat/FR-3691-property-filter-entity-source

JIRA Issue: FR-3874

nowgnuesLee:
Resolves #9501 (FR-3874)

https://fr-3874-pr9502-feed-staged-value.seungwon.10-82-0-159.sslip.io

## Symptom

A filter property that renders its own picker through `renderInput` (project select on the admin session list, user select on RBAC and the user folder permission panel, storage host select on the model store and the admin model card) loses its selection visually:

- Pick an option in the edit popover: Apply enables, but the select snaps back to its placeholder, so the pick is invisible until Apply.
- Click an existing token to edit it: the select opens empty instead of pre-populated.

## Cause

Since the Astryx `PowerSearch` rebuild the picked value is staged in PowerSearch state and committed by the popover's Apply button. `useRenderInputEditors` (`BAIPowerSearchAdapters.tsx`) receives that staged `value` and `isDisabled` from PowerSearch but forwarded only `onChange`, and every call site pinned `value={null}` — the antd-era convention, where a pick committed immediately and the control had to clear itself.

## Change

- `FilterRenderInput` now receives `value: string | null` and `isDisabled?: boolean`; the cached editor passes them through. Serialization is untouched.
- The five `renderInput` call sites feed `value` / `isDisabled` back into their select.
- The stage/Apply contract prose lives once on `FilterRenderInput`; both filters' `renderInput` docs point at it.
- Stories updated to the current contract (`toLabeledValue` helper maps the staged string to `BAIComplexSelect`'s labeled value).
- New `BAIPowerSearchAdapters.test.tsx` covers the pass-through and the label recording.

This is the `value` forwarding part of FR-3691 (#9089), split out so the visible defect can ship on its own. The `entitySource` redesign in #9089 is on hold (it drops the selects' scroll pagination).

## Verification

```
=== ALL PASS ===   (bash scripts/verify.sh)

 Test Files  3 passed (3)
      Tests  107 passed (107)   (BAIPowerSearchAdapters, BAIPropertyFilter, BAIGraphQLPropertyFilter)
```

Live check against a local webserver (RBAC → 할당된 사용자): the picked email stays in the select with Apply enabled, the token reads the email while the URL carries the UUID, and re-opening the token shows the select pre-populated.

### DIFF
--- a/packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.stories.tsx
+++ b/packages/backend.ai-ui/src/components/BAIGraphQLPropertyFilter.stories.tsx
@@ -4,6 +4,18 @@ import type { Meta, StoryObj } from '@storybook/react-vite';
 import { useState } from 'react';
 import { action } from 'storybook/actions';
 
+const sampleStorageHostOptions = [
+  { label: 'local:volume1', value: 'local:volume1' },
+  { label: 'local:volume2', value: 'local:volume2' },
+  { label: 'nfs:data', value: 'nfs:data' },
+];
+
+/** `renderInput`'s staged string -> the labeled object `BAIComplexSelect` takes. */
+const toLabeledValue = <T extends { value: string }>(
+  options: ReadonlyArray<T>,
+  value: string | null,
+): T | null => options.find((option) => option.value === value) ?? null;
+
 const meta: Meta<typeof BAIGraphQLPropertyFilter> = {
   title: 'Filter/BAIGraphQLPropertyFilter',
   component: BAIGraphQLPropertyFilter,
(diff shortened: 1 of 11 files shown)


yomybaby (approved):
LGTM

PR #9506, opened 2026-09-07
fix(FR-3865): validate the app launcher's allowed client IPs

yomybaby:
Resolves FR-3865 (no webui GitHub clone; Jira only)

The app launcher's "Allowed client IPs" field carried a tooltip and a label but
no `rules`, so any string the user typed was trimmed and forwarded verbatim as
the `allowed_client_ips` proxy query parameter. A typo produced an access
restriction the user believed was in place but which the proxy could not honour.

### What changed

- **`AppLauncherModal.tsx`** — the `clientIps` `Form.Item` gets a validator built
  on the existing `isValidIPOrCidr()` helper, reusing the `credential.InvalidIP`
  message the user/credential forms already show (no new i18n keys). The rule is
  **inert while "Open to public" is unchecked**, so a stale chip left in the
  disabled field cannot block a launch that sends no restriction at all; a small
  effect revalidates the field when the checkbox toggles so the error appears and
  clears with it.
- **`AppLauncherModal.tsx`** — `handleAppLaunch` no longer swallows the
  `validateFields()` rejection. Previously `.catch(() => { return; })` returned
  from the *catch callback*, leaving `values` `undefined` and letting the launch
  continue with no IP restriction and no preferred port. An invalid entry now
  aborts the launch.
- **`useBackendAIAppLauncher.tsx`** — `normalizeAllowedClientIps` is extracted
  next to the code that builds the proxy query string and shared with the
  modal's validator, so what is validated is what is sent: blank entries (easy
  to produce: `tokenSeparators` is `[',', ' ']`, so a trailing comma makes an
  empty chip) are dropped before the join and cannot yield consecutive commas.

### Design decisions

- Reused `isValidIPOrCidr()` and `credential.InvalidIP` rather than adding a
  launcher-specific validator or message, matching `UserSettingModal.tsx`'s
  `allowed_client_ip` rule.
- Aborting the launch on validation failure is slightly beyond the literal bug —
  today an out-of-range `preferredPort` also degrades silently — but silently
  dropping a restriction the user explicitly asked for is the more dangerous
  behaviour, so the fix covers both.
- **Server-side validation of `allowed_client_ips` is tracked separately as
  BA-7719.** This change does not close the CLI / direct-API path.

## Tests

- New `react/src/hooks/useBackendAIAppLauncher.clientIps.test.ts` (3 cases):
  nullish input, trimming and blank-dropping in `normalizeAllowedClientIps`,
  and the no-consecutive-commas join. IP / CIDR acceptance is `isValidIPOrCidr`'s
  contract, not this PR's.
  `pnpm exec vitest run src/hooks/useBackendAIAppLauncher.clientIps.test.ts` →
  **1 file, 3 tests passed**.

## Verification

`bash scripts/verify.sh` (Relay, Lint, Format, TypeScript, Terminology):

```
=== ALL PASS ===
```

## Review notes

- Open a running session's app launcher with `openPortToPublic` enabled, check
  **Open to public**, type `not-an-ip` and launch: the field shows
  `Invalid IP: not-an-ip` and the launch is blocked. `10.0.0.1`,
  `192.168.0.0/24`, `2001:db8::/32` all pass.
- Leave a bad chip in the field and **uncheck** "Open to public": the error
  clears and the app launches normally (the rule must stay inert when the field
  is disabled — this is the main regression risk).
- Type `10.0.0.1,` (trailing comma) to create a blank chip and launch: the proxy
  URL's `allowed_client_ips` is `10.0.0.1`, not `10.0.0.1,`.

- [ ] Documentation
- [x] Minimum requirements to check during review — `config.toml` needs
      `openPortToPublic = true` for the field to render
- [x] Test case(s) to demonstrate the difference of before/after — see
      `useBackendAIAppLauncher.clientIps.test.ts` and the review notes above

### DIFF
--- a/react/src/components/ComputeSessionNodeItems/AppLauncherModal.tsx
+++ b/react/src/components/ComputeSessionNodeItems/AppLauncherModal.tsx
@@ -12,6 +12,7 @@ import {
   useSuspendedFilteredAppTemplate,
 } from '../../hooks/useAppTemplate';
 import {
+  findInvalidClientIps,
   TCP_APPS,
   useBackendAIAppLauncher,
 } from '../../hooks/useBackendAIAppLauncher';
@@ -46,7 +47,7 @@ import {
   useErrorMessageResolver,
 } from 'backend.ai-ui';
 import * as _ from 'lodash-es';
-import { useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { Trans, useTranslation } from 'react-i18next';
 import { graphql, useFragment } from 'react-relay';
 
@@ -89,6 +90,12 @@ const AppLauncherModal: React.FC<AppLauncherModalProps> = ({
   const { message } = App.useApp();
   const { getErrorMessage } = useErrorMessageResolver();
 
+  // The client-IP rule only applies while the field is enabled, so toggling
+  // it must raise or clear the error instead of leaving a stale one behind.
+  useEffect(() => {
+    formRef.current?.validateFields(['clientIps']).catch(() => undefined);
+  }, [openToPublic]);
+
   const session = useFragment(
     graphql`
       fragment AppLauncherModalFragment on ComputeSessionNode {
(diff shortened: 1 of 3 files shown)


nowgnuesLee (approved):
LGTM

PR #9507, opened 2026-09-07
fix(FR-3724): show a loading row in the select popup while the list is in flight

### LINKED ISSUE #9172: RBAC role create modal: scope target selector shows no loading state, only 'No results'
In **RBAC management → Create role**, choosing a **Scope** (e.g. User) makes the adjacent **Target** selector fetch the target list. While that fetch is in flight there is no loading indication at all: the selector looks ready, and opening it immediately shows only "No results" — so the user is misled into thinking there are no targets to choose from.

**Expected:** while the target list is loading, the selector shows a loading indicator (spinner / "Loading…" in the dropdown), and — to be decided — it may also be disabled so it cannot be opened until the data arrives.

**Observed:** no loading state; the dropdown opens empty with "No results" (see attached screenshot).

**Open question for the fix:** should the selector be non-interactive (disabled) during loading, or stay openable but show a loading row instead of "No results"? Please decide as part of the fix.

—

JIRA Issue: FR-3724

yomybaby:
Resolves #9172 (FR-3724)

In **RBAC management → Create role**, picking a **Scope** makes the adjacent **Target** selector fetch its list. Opening it during that fetch showed only "No results", which reads as "there are none" rather than "not loaded yet".

Two gaps produced that:

- **`BAIComplexSelect` never consulted `isLoading` for its empty state** — it rendered the NoResults text whenever `options.length === 0`. The popup now draws a spinner row with a new `general.Loading` string in that case. A non-empty list is untouched (no flicker on a background refetch), and an explicit `emptyContent` still wins over both.
- **The open-triggered refetch raised no pending flag.** The Relay-backed selects fetch the real list only once opened, under `fetchPolicy: deferredOpen ? 'network-only' : 'store-only'`. `deferredOpen` is a `useDeferredValue`, so the refetch is a deferred update: it suspends without a fallback and nothing — neither the trigger spinner nor the popup — knew a fetch was running. Each of the eight open-driven selects `ScopeIdSelect` can mount now folds the *opening* half of that transition into the `isLoading` expression it already builds.

## Design decisions

- **The issue's open question — disabled vs. openable during loading: openable, with a loading row.** The first mount is already covered: `ScopeIdSelect` wraps every branch in a `Suspense` whose fallback is a disabled, `isLoading` field (`RoleFormModal.tsx`). Making the *reopen* path disabled as well would fight that and take the popup away mid-interaction, so the popup stays open and explains itself instead.
- **Only the opening half of the open transition counts** (`!!controllableOpen && !deferredOpen`). `deferredOpen` lags on close too, and the naive `controllableOpen !== deferredOpen` would flash the trigger spinner on every close for a transition that fetches nothing.
- **Scope of the fragment change.** The `deferredOpen` pattern appears in ~29 files; this PR touches the eight the RBAC scope router can mount (User, Project, VFolder, Session, Model deployment, Container registry, Storage host, Keypair). The Domain and Resource-group branches do not use the pattern — they fetch eagerly and are covered by the Suspense fallback. The same one-line term applies to the rest and can follow.
- **`BAIUserSelect`'s PILOT-DECISION P26-7 is rewritten, not silently contradicted** — antd's `Skeleton.Input` placeholder stays dropped; what changes is that the empty popup is no longer unconditionally "No results".

## Tests

- Four new cases in `packages/backend.ai-ui/src/components/BAIComplexSelect.popup.test.tsx`: empty + not loading says "no results"; empty + loading says "Loading..."; **populated + loading keeps its rows and shows no loading row** (the flicker guard); `emptyContent` overrides both.
- `pnpm --filter backend.ai-ui exec vitest run src/components/BAIComplexSelect src/components/fragments` → `Test Files 3 passed | 1 skipped (4)`, `Tests 32 passed | 1 skipped (33)`.
- No e2e run (needs a live cluster).

## Verification

```
bash scripts/verify.sh
...
=== ALL PASS ===
```

No CSS or design-token changes, so the Astryx token gate is untouched.

## Review notes

- **Reproduce:** RBAC management → Create role → pick Scope = `User` → open **Target** before the list lands. Before: "No results". After: a spinner + "Loading...", replaced by the users once the query resolves. Throttle the network to make the window comfortable.
- **Check the flicker guard:** open and close any `BAI*Select` (e.g. the project selector in the header) repeatedly — the trigger spinner must not blink on close, and a select whose list is already populated must never show the loading row.
- **Blast radius:** `BAIComplexSelect` backs every `BAI*Select` in the app, but the change is confined to the `options.length === 0` branch; every populated popup renders exactly as before.

- [ ] Documentation

### DIFF
--- a/packages/backend.ai-ui/src/components/BAIComplexSelect.popup.test.tsx
+++ b/packages/backend.ai-ui/src/components/BAIComplexSelect.popup.test.tsx
@@ -288,3 +288,50 @@ describe('BAIComplexSelect popup — reopen', () => {
     expect(highlightedLabels()).toEqual([]);
   });
 });
+
+describe('BAIComplexSelect popup — empty state', () => {
+  it('says "no results" for an empty list that is not loading', async () => {
+    const user = userEvent.setup();
+    render(<BAIComplexSelect label="Targets" options={[]} />);
+
+    await user.click(trigger());
+    expect(listbox()).toHaveTextContent(/no results/i);
+    expect(listbox()).not.toHaveTextContent(/loading/i);
+  });
+
+  it('says "loading" instead while the list is still in flight (FR-3724)', async () => {
+    const user = userEvent.setup();
+    render(<BAIComplexSelect label="Targets" options={[]} isLoading />);
+
+    // The RBAC scope target select opens before its network-only refetch
+    // lands, and "No results" there reads as "there are none".
+    await user.click(trigger());
+    expect(listbox()).toHaveTextContent('Loading...');
+    expect(listbox()).not.toHaveTextContent(/no results/i);
+  });
+
+  it('leaves a populated list alone while loading — no flicker', async () => {
+    const user = userEvent.setup();
+    render(<BAIComplexSelect label="Targets" options={OPTIONS} isLoading />);
+
+    await user.click(trigger());
+    expect(optionRows()).toHaveLength(3);
+    expect(listbox()).not.toHaveTextContent(/loading/i);
+  });
+
+  it('still lets `emptyContent` win over both', async () => {
+    const user = userEvent.setup();
+    render(
+      <BAIComplexSelect
+        label="Targets"
+        options={[]}
+        isLoading
+        emptyContent={<span>pick a scope first</span>}
+      />,
+    );
+
+    await user.click(trigger());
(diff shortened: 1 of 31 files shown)


agatha197 (approved):
LGTM

PR #9508, opened 2026-09-07
docs(FR-2336): correct purge-users consequences in the admin manual

### LINKED ISSUE #6065: Update the documentation regarding the permanent deletion of users.
### Summary

Update the documentation regarding the permanent deletion of users.

### Context

This issue involves revising the existing documentation to include information about the process and implications of permanently deleting users.

### Acceptance criteria

- Documentation must clearly outline the steps for permanent user deletion.
- Include any potential consequences or considerations for users being deleted.

### Other information

N/A

JIRA Issue: FR-2336

yomybaby:
Resolves #6065 (FR-2336)

The user manual's **Purge inactive users** section already described the flow, but one of the two consequences it documented for this irreversible action was the inverse of what the manager actually does. This PR corrects it in all four languages (en / ko / ja / th). Docs-only — no UI copy changed.

### What was wrong

**1. "Delete shared virtual folders as well?" was documented backwards.**

That checkbox sets `purgeSharedVfolders` (`PurgeUsersModal.tsx:93` passes it straight through). The backend does the opposite of what the label reads:

- `data/schema.graphql:2497` — `purgeSharedVfolders: Boolean! = false`, described as *"If true, migrate shared virtual folders to the admin user before purging."*
- `manager/services/user/service.py::_purge_single_user` calls `migrate_shared_vfolders(deleted_user_uuid=…, target_user_uuid=<requesting admin>)` when the flag is set; `_migrate_shared_vfolders` re-owns those folders to the admin, appending a random suffix when a name collides with one of the admin's own folders.
- When the flag is **not** set there is no migration, so `delete_user_vfolders` removes them together with the rest of the user's folders.

So **checked = the shared folders survive under the admin**, **unchecked = they are deleted** — the manual said the reverse.

**2. The Users page status selector label was wrong.** The manual claimed it reads "Inactive (include keypair)". `AdminUserManagement.tsx:426-431` renders a `BAIRadioGroup` with `general.Active` / `general.Inactive`, i.e. plain **Active** / **Inactive**. The `credential.InactiveIncludeKeypair` string belongs to a different control (`UpdateUsersModal` / `UserSettingModal`).

### Design decisions

- The checkbox **label itself** is misleading — it says "delete" for an option whose effect is "keep". Changing UI copy is out of scope for this docs issue, so the manual now documents both what the label says and what it does, and a `:::warning` right below the option list flags the inversion explicitly. **A follow-up to fix the UI string (`credential.DeleteSharedVirtualFolders`) is worth filing**; the current wording will keep tripping admins up regardless of what the manual says.
- The `:::danger` consequences block was left as-is: it describes the default (unchecked) path, which is still accurate, and the option bullet immediately above it now carries the nuance.
- The pre-existing screenshot TODO for `purge_users_modal.png` is untouched — out of scope here.

## Tests

No unit-testable logic (documentation only).

- `pnpm --filter backend.ai-webui-docs run check:nav-titles` → `check-nav-titles: OK (112 nav entries checked)`
- `bash scripts/verify.sh`

## Verification

```
=== ALL PASS ===
```

(`scripts/verify.sh`: Relay, Lint, Format, TypeScript, Terminology — 0 blocking terminology findings; the single warn-only finding is pre-existing and unrelated.)

## Review notes

- Compare `packages/backend.ai-webui-docs/src/en/admin_menu/admin_menu.md` (the "Purge inactive users" section) against `data/schema.graphql`'s `BulkPurgeUsersV2Options.purgeSharedVfolders` description — they should now agree.
- Open **Users → Inactive** in the WebUI and confirm the status selector really reads **Active** / **Inactive** (not "Inactive (include keypair)").
- The ko / ja / th edits mirror the en ones one-for-one; the `:::warning` admonition is new in all four and sits between the option list and the modal screenshot.

- [x] Documentation


agatha197:
Blocked by **[BA-7733](https://lablup.atlassian.net/browse/BA-7733)** — closing for now.

## Why

This PR's main change documents the **Delete shared virtual folders as well?** option as behaving opposite to its label, and adds a warning box telling admins that checking it *keeps* the folders:

> The first option's label reads as if checking it deletes the shared folders, but the effect is the opposite: **checking it keeps them**, under your ownership, while leaving it unchecked deletes them along with the users' other data.

The described behavior is accurate — I traced it end to end in the manager — but it is a **backend regression, not intended behavior**, so the manual should not enshrine it.

`services/user/service.py:220-226` migrates the shared folders to the requesting admin when the flag is truthy, and the delete pass that follows only sweeps rows still owned by the purged user, so they survive:

```py
if action.purge_shared_vfolders.optional_value():
    await self._user_repository.migrate_shared_vfolders(...)
...
await self._user_repository.delete_user_vfolders(user_uuid=user_uuid, storage_manager=...)
```

The original code was negated (`if not props.purge_shared_vfolders:`), i.e. `True` meant delete — matching the field name and this checkbox's label. The `not` was dropped in core PR lablup/backend.ai#4103 (BA-1099, "Align action architecture code", 2025-04-10) with no mention of a behavior change. The Python client CLI help and the legacy REST DTO description still document the original semantics; only the v2 DTO description was rewritten to match the flipped code.

WebUI is not at fault here — `PurgeUsersModal.tsx:93` passes the checkbox through unchanged, which is correct against the field's documented-by-name contract. Note the sibling option already compensates for an honestly-named flag (`delegateEndpointOwnership = !deleteModelServices`); no such inversion is needed once the backend polarity is restored.

## What happens next

Once BA-7733 lands, the checkbox will match its label and both the warning box and the rewritten bullet become wrong. Reopening this PR after that would ship stale docs.

## Worth keeping from this PR

The **Inactive (include keypair)** correction is independent of the bug and still valid — the status selector really does just read **Active** / **Inactive**. Suggest re-landing that hunk on its own, and revisiting the purge-options wording after BA-7733.


[BA-7733]: https://lablup.atlassian.net/browse/BA-7733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #9522, opened 2026-09-07
refactor(FR-2655): use the typed v2 vfolder select in the admin model card form

### LINKED ISSUE #6892: Add a BAI vfolder selector component backed by adminVfoldersV2 for admin-scoped model-store folder selection
## Background

The existing `BAIVFolderSelect` component (packages/backend.ai-ui) is built on the Graphene-based `vfolder_nodes` query. Its `filter` prop accepts a plain string filter (e.g. `usage_mode == "model"`), which is untyped and cannot express structured filters like project type.

The Strawberry v2 API now exposes:

- `adminVfoldersV2(filter: VFolderFilter)` — system-wide, typed filter input (`usageMode`, `status`, `name`, etc.)
- `projectVfolders(projectId: UUID!, filter: VFolderFilter)` — project-scoped

`VFolderFilter` is a typed Strawberry input, which gives type-safe filtering without magic strings.

## Problem

Admin-level folder pickers (e.g. the Create Model Card modal on `/admin-serving?tab=model-store`) currently work around the limitation by passing string filters to `BAIVFolderSelect`. There is no clean way to filter by `usageMode: MODEL` using the typed v2 API.

## Proposed Solution

Create a new BAI selector component (e.g. `BAIAdminVFolderSelect`) that:

- Uses `adminVfoldersV2` (Strawberry v2) as its data source
- Accepts a typed `VFolderFilter` input instead of a raw filter string
- Follows the same prop/UX contract as `BAIVFolderSelect` (paginated, searchable, Relay fragment)
- Replaces the string-filter workaround in `AdminModelCardSettingModal` (introduced in FR-2651)

## References

- `packages/backend.ai-ui/src/components/fragments/BAIVFolderSelect.tsx` — existing component to model after
- `react/src/components/AdminModelCardSettingModal.tsx` — primary consumer
- Schema: `adminVfoldersV2`, `VFolderFilter`, `VFolderUsageModeFilter`
- Related: FR-2651 (current workaround using string filter), FR-2573 (general v2 migration)

—

Captured while working on branch: 04-22-fix_fr-2651_list_all_model-store_folders_when_creating_a_model_card

JIRA Issue: FR-2655

yomybaby:
Resolves #6892 (FR-2655)

The model-storage folder picker in the admin Model Card form was still the untyped `BAIVFolderSelect`, narrowed with the hand-written string filter `ownership_type == "group"` on the Graphene `vfolder_nodes` connection — the FR-2651 workaround. BUI already ships the typed replacement, `BAIProjectVfolderSelect`, backed by `projectVfolders(projectId:, filter: VFolderFilter)`; it had no consumer yet. This PR swaps the call site over.

`projectVfolders(projectId:)` resolves to the same row set the old `scope_id: project:<id>` + `ownership_type == "group"` pair produced (`ProjectVFolderOperationScope.to_condition()` is `VFolderRow.group == project_id`), so the admin sees exactly the same folders — the filter is now the schema's own input type instead of a string the compiler cannot check.

**Design decisions**

- **`projectVfolders`, not `adminVfoldersV2`.** The issue proposed `adminVfoldersV2`, but `VFolderFilter` exposes no ownership-type or project field (`name` / `host` / `status` / `usageMode` / `cloneable` / `createdAt` only), so the system-wide connection cannot be narrowed to the model-store project's folders. `projectVfolders` is the only typed field that expresses the existing behaviour. The page is superadmin-only either way (`AdminDeploymentPage` is routed `access: 'superadmin'`), so neither field's authorization gate is the deciding factor.
- **No `usageMode: MODEL` filter.** Adding one would be a behaviour change relative to today's picker, which does not filter usage mode — a model-store folder created as `GENERAL` would silently disappear from the list. Left alone deliberately.
- **Id shape.** `BAIProjectVfolderSelect` emits the folder's *local* id where the old select emitted the Relay *global* id, so the two conversions around it move with it: the create mutation now sends `convertToUUID(values.vfolderId)` for its `vfolderId: UUID!`, and a just-created folder is written into the form as `toLocalId(result.id)` rather than being re-encoded back to a global id.
- The `!isModelStoreProjectResolved` render guard became `!modelStoreProjectContext` so TypeScript narrows the now-required `projectId` prop; the boolean is still what disables the OK button.

## Tests

- New `react/src/components/AdminModelCardSettingModal.test.tsx` (3 cases): the picker is scoped to the model-store project, the create mutation receives the dashed UUID, and a just-created folder lands in the form as its local id. Both id assertions fail against the previous code (`toLocalId` on a raw UUID throws).
- `pnpm exec vitest run src/components/AdminModelCardSettingModal.test.tsx` from `react/` — 3 passed.
- Not run: e2e (needs a live cluster).

## Verification

`bash scripts/verify.sh` →

```
=== ALL PASS ===
```

## Review notes

- Open **Deployments → Model Cards** as a superadmin and hit **Create**: the folder dropdown should list the same project folders as before, and searching should still filter by name.
- Create a folder from the `+` button next to the picker — it should be selected automatically when the dialog closes, and saving should succeed (this is the id-shape path).
- Edit mode is untouched: it renders a `FolderLink`, not the picker.

- [ ] Documentation
- [x] Minium required manager version — `projectVfolders` / `vfolderV2` are "Added in 26.4.2"
- [x] Minimum requirements to check during review — a superadmin account and a MODEL_STORE project with at least one project folder

### DIFF
--- a/react/src/components/AdminModelCardSettingModal.test.tsx
+++ b/react/src/components/AdminModelCardSettingModal.test.tsx
@@ -1,0 +1,151 @@
+/**
+ @license
+ Copyright (c) 2015-2026 Lablup Inc. All rights reserved.
+ */
+import AdminModelCardSettingModal from './AdminModelCardSettingModal';
+import '@testing-library/jest-dom';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { RelayEnvironmentProvider } from 'react-relay';
+import { createMockEnvironment } from 'relay-test-utils';
+import type { RelayMockEnvironment } from 'relay-test-utils/lib/RelayModernMockEnvironment';
+
+/**
+ * The model-storage folder picker is project-scoped and emits the folder's
+ * LOCAL id, while `CreateModelCardV2Input.vfolderId` is a `UUID!` — a
+ * regression back to a global id would only surface against a live manager.
+ */
+
+vi.mock('react-i18next', async () => {
+  const React = await import('react');
+  return {
+    useTranslation: () => ({
+      t: (key: string) => key,
+      i18n: { language: 'en', changeLanguage: () => new Promise(() => {}) },
+      ready: true,
+    }),
+    Trans: (props: any) => React.createElement('span', null, props.i18nKey),
+    initReactI18next: { type: '3rdParty', init: () => {} },
+  };
+});
+
+vi.mock('../hooks', async (importOriginal) => {
+  const originalModule = await importOriginal<typeof import('../hooks')>();
+  return {
+    ...originalModule,
+    useCurrentDomainValue: () => 'default',
+  };
+});
+
+const MODEL_STORE_PROJECT = {
+  id: 'project-0000-1111-2222-333333333333',
+  name: 'model-store',
+};
+// Undashed, as a Relay global id decodes to on some manager builds.
+const PICKED_VFOLDER_LOCAL_ID = '11111111111122223333444444444444';
+const PICKED_VFOLDER_UUID = '11111111-1111-2222-3333-444444444444';
+const CREATED_VFOLDER_UUID = 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee';
+const CREATED_VFOLDER_GLOBAL_ID = btoa(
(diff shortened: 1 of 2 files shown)


ironAiken2 (approved):
LGTM

PR #9523, opened 2026-09-07
fix(FR-2371): resolve deployment project name from projectV2

### LINKED ISSUE #6144: Endpoint GQL type returns project as ID, cannot resolve project name
## Problem

The Endpoint GraphQL type exposes the `project` field as a plain String (UUID), making it impossible to display the human-readable project name in the UI without an additional query.

## Root Cause

The `Endpoint` type in `schema.graphql` defines `project: String` instead of referencing the full project object. Modern types like `SessionV2` correctly expose `project: ProjectV2` as a full object reference.

## Current Workaround

`SwitchToProjectButton` component makes a separate `group_node(id: $projectId)` query to fetch the project name — an extra network round-trip per endpoint.

## Suggested Fix Direction

**Backend**: Add `project_name: String` field (or `project_object: Group`) to the Endpoint type resolver, similar to how `SessionV2` exposes the project relationship.

**Frontend**: Once available, update `EndpointList` and `EndpointDetailPage` to use the new field instead of additional queries.

## Related Files

- `data/schema.graphql` — Endpoint type definition (line ~4389)
- `react/src/components/EndpointList.tsx` — displays raw project ID
- `react/src/pages/EndpointDetailPage.tsx` — uses SwitchToProjectButton workaround
- `react/src/components/SwitchToProjectButton.tsx` — workaround pattern

## Related Issues

- FR-2368 (same pattern for ComputeSessionNode)
- FR-2370 (related scope filtering issue)

JIRA Issue: FR-2371

yomybaby:
Resolves #6144 (FR-2371)

`DeploymentDetailPage` renders a project-mismatch banner with a "switch to this project" shortcut. The page query only selected `metadata.projectId`, so `SwitchToProjectButton` had to resolve the human-readable name itself with a separate `group_node(id:)` query — the workaround the issue describes.

The modern deployment type already carries the project as a full object: `ModelDeploymentMetadata.projectV2: ProjectV2` (added in 26.4.3), whose `basicInfo.name` is the name the button needs. This PR selects it in the page query and hands it to the button.

**What changed**

- `DeploymentDetailPage.tsx` — the `metadata { … }` selection now also takes `projectV2 @since(version: "26.4.3") { basicInfo { name } }`, and passes `projectName` to `SwitchToProjectButton`.
- `SwitchToProjectButton.tsx` — new optional `projectName` prop. When the caller supplies it, the component renders and switches with **no query at all**; when it is absent, the existing `group_node` lookup runs as before, behind the same `Suspense`.

**Design decisions**

- The fallback is kept rather than deleted: `projectV2` is `@since 26.4.3`, so the runtime GraphQL transformer strips the field on older managers and `projectName` arrives `null`. The query-backed path therefore lives in a separate child component, so the fallback's hook is never conditional.
- The issue's original "Related Files" (`EndpointList.tsx`, `EndpointDetailPage.tsx`) no longer exist on `main`; `DeploymentDetailPage` is the successor page and the only `SwitchToProjectButton` call site. The legacy Graphene `Endpoint.project: String` field is untouched — nothing in the WebUI reads it any more.

## Tests

- New `react/src/components/SwitchToProjectButton.test.tsx`: asserts that a supplied `projectName` produces the label and the `switchProject({ projectId, projectName })` call with **zero** pending Relay operations, and that omitting it still resolves the name through `group_node`.
- `pnpm exec vitest run src/components/SwitchToProjectButton.test.tsx src/pages/DeploymentDetailPage.test.tsx` → 2 files, 4 tests passed.
- `pnpm run relay` re-run; `__generated__/DeploymentDetailPageQuery.graphql.ts` committed.

## Verification

`bash scripts/verify.sh`

```
=== ALL PASS ===
```

## Review notes

- Open a deployment that belongs to a project other than the currently selected one — the warning banner's button should read "Switch to \<project name\>" immediately, with no second network request in the Network tab (previously a `SwitchToProjectButtonQuery` fired).
- Clicking it must still switch project the same way (URL `:projectName` segment rewritten on project scope, atom updated on global admin scope).
- Against a manager older than 26.4.3, `projectV2` is stripped from the document and the button falls back to the `group_node` query — the button briefly shows its loading state, then behaves exactly as before.

- [ ] Documentation
- [x] Minium required manager version — `projectV2` needs 26.4.3; older managers keep the `group_node` fallback
- [x] Minimum requirements to check during review — a deployment whose project differs from the selected one, so the mismatch banner renders

### DIFF
--- a/react/src/__generated__/DeploymentDetailPageQuery.graphql.ts
+++ b/react/src/__generated__/DeploymentDetailPageQuery.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<7b1f8b5de41b47cef366c1651696ac34>>
+ * @generated SignedSource<<730c854422157388e983a02aeda5a4f2>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -34,6 +34,11 @@ export type DeploymentDetailPageQuery$data = {
     readonly metadata: {
       readonly name: string;
       readonly projectId: string;
+      readonly projectV2: {
+        readonly basicInfo: {
+          readonly name: string;
+        };
+      } | null | undefined;
       readonly status: DeploymentStatus;
     };
     readonly networkAccess: {
@@ -100,6 +105,18 @@ v5 = {
 v6 = {
   "alias": null,
   "args": null,
+  "concreteType": "ProjectBasicInfo",
+  "kind": "LinkedField",
+  "name": "basicInfo",
+  "plural": false,
+  "selections": [
+    (v3/*: any*/)
+  ],
+  "storageKey": null
+},
+v7 = {
+  "alias": null,
+  "args": null,
   "concreteType": "ModelDeploymentNetworkAccess",
   "kind": "LinkedField",
   "name": "networkAccess",
@@ -122,7 +139,7 @@ v6 = {
   ],
   "storageKey": null
 },
-v7 = {
+v8 = {
   "alias": null,
   "args": null,
   "concreteType": "ReplicaState",
@@ -140,7 +157,7 @@ v7 = {
   ],
   "storageKey": null
 },
-v8 = [
+v9 = [
   {
     "alias": null,
     "args": null,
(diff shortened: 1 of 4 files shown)


nowgnuesLee (approved):
LGTM

PR #9527, opened 2026-09-08
feat(FR-3877): move the system announcement to the domain app config

### LINKED ISSUE #9526: App config: move the system announcement to the domain app config
## Goal

Move the WebUI system announcement from the legacy manager announcement API to the domain app config, in line with the other app-config migrations under FR-1203 (appearance document: FR-1964, FR-3834).

## Current state

- Admins edit the announcement on the Maintenance settings page (`MaintenanceSettingList` → `AnnouncementEditModal`), which writes through `baiClient.service.update_announcement(enabled, message)`.
- `AnnouncementBanner` reads the same legacy endpoint and shows the banner to every user on the cluster.

## Scope

- Store the announcement (enabled flag, message, and any display options) as a slice of the domain app config instead of the legacy announcement endpoint.
- Read it through the app-config hooks introduced by FR-1964 so the banner follows the current domain.
- Update `AnnouncementEditModal` to save into app config (same save path as the Branding "apply to domain" flow in FR-3834).
- Keep the legacy endpoint as a fallback for managers that do not serve app config, or document the cutover.

## Out of scope

- New announcement features (scheduling, per-user dismissal) — separate issues.

JIRA Issue: FR-3877

ironAiken2:
Resolves #9526 (FR-3877)

## Summary

Top of the FR-1203 app-config stack (#9361 → #8860 → #9362 → this). The system announcement moves from the manager's legacy announcement endpoint (`/manager/announcement`, one markdown string in etcd, first line lifted out as the title) to the **domain app config**:

```jsonc
// domainConfig (DOMAIN scope, merged into `myAppConfigs(['domainConfig'])`)
{
  "announcement": {
    "enabled": true,
    "title": "Maintenance tonight",          // plain text, shown in every banner state
    "body": "Details in **markdown** …",     // optional, revealed on expand
    "updatedAt": "2026-09-08T05:12:00.000Z"  // keys per-session dismissal
  }
}
```

### Hooks (`react/src/hooks/useAppConfig.tsx`)

- **`useUpdateDomainAppConfig()`** — the missing DOMAIN-scope setter next to `useUpdatePublicDomainAppConfig` / `useUpdateMyUserAppConfig`. `(subKey, nextValue)`: takes the domain uuid from the client (`useCurrentDomain().id`, stored at login — see below), re-reads the raw DOMAIN-scope `domainConfig` fragment, replaces one sub-key (`undefined` removes it) through `scopedUpsertAppConfigFragments`, then refetches the merged `myAppConfigs(['domainConfig'])` view so every `useDomainAppConfig` reader updates — no `window.location.reload()`, no per-save `domainV2` round-trip.
- Reads go through the existing `useDomainAppConfig<T>(subKey)` (first consumer).

### Banner / editor

- `AnnouncementBanner` reads `useDomainAppConfig<DomainAnnouncement>('announcement')`. Title in every state; the body renders behind the expand toggle; a title past the 120-code-point cutoff collapses like before. Dismissal is keyed by `updatedAt`, so a re-published announcement resurfaces the banner.
- `AnnouncementEditModal` gains a **Title** field (a `BAIFormItem` inside `Form`, `required` + `whitespace`, so its error renders like every other form field and an untouched empty title stays error-free; Publish / Save as Draft run `validateFields()` first) and a **Publish** split button whose menu offers **Save as Draft** (`enabled: false`, which the legacy endpoint could not persist — it stored by presence only); a saved draft is marked by a Draft token in the modal title. **Delete** stands alone on the left and removes the sub-key. `body` is trimmed on save. The content suspends on the Relay read behind a skeleton-bodied modal fallback, so the Maintenance page opens it without a page-level Suspense.
- Deleted: `announcementSummary.ts` (first-line extraction, markdown-to-prose) and `useSuspenseGetAnnouncement.tsx` (TanStack). `helper/announcement.ts` keeps the title cutoff and the visibility / collapsibility predicates, with tests. `service.get_announcement` / `update_announcement` stay in `backend.ai-client` (library API) but have no caller in the app.
- i18n: `summary.AnnouncementTitle`, `summary.AnnouncementTitleRequired` (all 21 locales); `summary.AnnouncementMessageRequired` removed (no consumer).

### Login reads `myUserV2` (`react/src/helper/loginSessionAuth.ts`)

`connectViaGQL` used to chain `keypair` → `user` → `group.list` and only ever stored the domain *name*, so the DOMAIN-scope setter above had to resolve the uuid on every save. It now reads `myUserV2` once — `id`, `basicInfo`, `organization { domainName role }`, `domain { entityId basicInfo { name } }`, `projects(filter: { isActive: true })` — and stores the uuid as `_config.domainId` (new `useCurrentDomain()` hook returning `{ name, id }`; `LoginConfigState.domain_id`). The read is a module-level Relay `graphql` tag used through `fetchQuery(RelayEnvironment, …)`: `RelayEnvironment`'s fetch resolves `globalThis.backendaiclient` at request time, which `connectViaGQL` installs first, so the response type is the generated `loginSessionAuthMyUserQuery` rather than a hand-written one.

- `UserV2.projects` caps an unpaginated read at the manager's default page size (10), so the login walks the pages with `limit: 100` / `offset`, bounded by the connection's `count` (a single pagination mode; no cursor to stall on).
- V2 `id`s are Relay global ids; `toLocalId` decodes them for `user_uuid` and `groupIds`.
- Dropped what nothing read: the stored `resource_policy` and `need_password_change`, and the fetched-but-unstored `username` / `is_active`. `groupIds` now holds only the user's projects (the legacy map held every group in the domain); its sole reader, `current_group_id()`, only ever looks up the user's current project.
- `useAppConfigDomainIdQuery` is gone with its `__generated__` artifact.

`myUserV2` is 26.2.0+, below the 26.9.0 this PR already requires, so no gate.

### No version fallback

`myAppConfigs` / `scopedUpsertAppConfigFragments` need manager 26.9.0+. This ships in the LTS that guarantees it (decision: FR-3877 discussion), so there is no legacy-endpoint fallback — on an older manager the banner query fails inside `ErrorBoundaryWithNullFallback` and the Maintenance page's Edit modal errors on load.

## Verification

- `bash scripts/verify.sh` → all checks pass at commit time (Lint, Format, TypeScript ×2, Vite warmup, StyleX, Astryx theme build, Astryx integration, z-index, Agent mappings, Terminology). The Relay check flagged only the two new `__generated__` artifacts as uncommitted before this commit; they are committed.
- `vitest run src/helper/announcement.test.ts` — 8 tests.
- `vitest run src/helper/loginSessionAuth.test.ts` — 11 tests on a `createMockEnvironment()` stand-in for `RelayEnvironment` (single read, offset walk bounded by `count`, role → `is_admin`/`is_superadmin` mapping, domain fallback, remembered project, unauthenticated logout, endpoint history); `STokenLoginBoundary.test.tsx` and `useWebUIConfig.test.ts` still pass.
- `pnpm relay` — `useAppConfigDomainRawQuery`, `loginSessionAuthMyUserQuery` generated; `useAppConfigDomainIdQuery` removed.
- **Not yet verified against a live manager** (the 10.82.0.x test network was unreachable from the authoring box). Assumptions to confirm on the first run: `domainConfig` is allow-listed at DOMAIN scope for the read (`myAppConfigs`); `myUserV2.domain.entityId` is the uuid `AppConfigScopeRef.scopeId` expects; and `ProjectV2.id` / `UserV2.id` come back as Relay global ids (`base64("ProjectV2:<uuid>")`), which is what the manager source declares (`NodeID` on `PydanticNodeMixin`). A dev server for this branch is up for that: `https://fr-3877.localhost:1355`.

- Review round (`7d5d2c7ff`): `bash scripts/verify.sh` → `=== ALL PASS ===`; `vitest run src/helper/loginSessionAuth.test.ts src/helper/announcement.test.ts` → 2 files / 19 tests passed.
- Stack rebased onto `main` (`f8d974dd3`) and the title moved to the form engine (`826d213b5`): `bash scripts/verify.sh` → `=== ALL PASS ===`.

## Reviewer checklist

- [ ] 1. As superadmin, Maintenance → Edit Announcement: set a title and a markdown body, Publish. The banner appears at the top with the title and an expand toggle that reveals the body.
- [ ] 2. Dismiss the banner, then re-publish with any change → the banner comes back (dismissal is keyed by `updatedAt`).
- [ ] 3. Save as Draft (Publish menu) → banner gone, the modal reopens with a Draft token and the saved title/body.
- [ ] 4. Delete → banner gone, the modal opens empty with Delete disabled.
- [ ] 5. Sign in as a non-admin user of the same domain → the banner shows without the Edit button.
- [ ] 6. Sign in as a user who belongs to more than 10 active projects → the project selector lists all of them, and the remembered project survives a reload (login now pages `myUserV2.projects`).

https://claude.ai/code/session_01B13bqRK3V12jiCvia8fm2w


agatha197 (changes requested):
Image
1. The title validation error has a block unlike form item. Let's remove the background.
2. Instead of enabled button, let's add a draft publish (replace better representation ;)) to the left of the publish button with primary outlined button. And leave delete button to the left only.

react/src/helper/loginSessionAuth.ts
```diff
@@ -50,93 +54,96 @@ export async function checkLoginSession(apiEndpoint: string): Promise<boolean> {
...
+      { first: PROJECT_PAGE_SIZE, after },
+      { fetchPolicy: 'network-only' },
+    ).toPromise();
+    if (!response?.myUserV2) {
+      await client.logout();
+      throw new Error('User information is missing.');
+    }
+    me = response.myUserV2;
+    for (const { node } of me.projects?.edges ?? []) {
+      projects.push({ id: toLocalId(node.id), name: node.basicInfo.name });
```
agatha197, line 120: `toLocalId` (`atob`-based) throws on anything that isn't a base64 global id, and this runs inside `connectViaGQL` — one unexpected id format fails the whole login (same for `me.id` at line 131). Wrap it so a bad id degrades to the raw value instead of throwing.

react/src/helper/loginSessionAuth.ts
```diff
@@ -50,93 +54,96 @@ export async function checkLoginSession(apiEndpoint: string): Promise<boolean> {
...
+    }
+    after = me.projects?.pageInfo.hasNextPage
+      ? (me.projects.pageInfo.endCursor ?? null)
+      : null;
+  } while (after);
```
agatha197: Nothing bounds this loop — a manager that keeps answering `hasNextPage: true` (or repeats the same `endCursor`) hangs login forever with no error. Add a page cap, or bail out when the cursor doesn't advance.

react/src/helper/loginSessionAuth.ts
```diff
@@ -50,93 +54,96 @@ export async function checkLoginSession(apiEndpoint: string): Promise<boolean> {
...
+  // `UserV2.projects` caps an unpaginated read at the manager's default page
+  // size, so the login walks the cursor (the legacy `group.list` had no cap).
+  const PROJECT_PAGE_SIZE = 100;
+  const projects: Array<{ id: string; name: string }> = [];
+  let me: NonNullable<loginSessionAuthMyUserQuery['response']['myUserV2']>;
+  let after: string | null = null;
+  do {
+    const response = await fetchQuery<loginSessionAuthMyUserQuery>(
+      RelayEnvironment,
+      graphql`
```
agatha197: The sibling hooks in `useAppConfig.tsx` declare their queries as module-level constants — hoist this tag out of the loop body to match.

react/src/helper/loginSessionAuth.ts
```diff
@@ -50,93 +54,96 @@ export async function checkLoginSession(apiEndpoint: string): Promise<boolean> {
...
+        query loginSessionAuthMyUserQuery($first: Int!, $after: String) {
+          myUserV2 {
```
agatha197: Every page after the first re-fetches `basicInfo` / `organization` / `domain` along with the next project page — a lighter follow-up query (or a larger `first`) would avoid the repeat.

react/src/components/AnnouncementEditModal.tsx
```diff
@@ -203,18 +205,31 @@ const AnnouncementEditModal: React.FC<AnnouncementEditModalProps> = ({
...
-        <BAISkeleton rows={4} />
-      ) : (
+      <BAIFlex direction="column" align="stretch" gap="sm">
+        <TextInput
+          label={t('summary.AnnouncementTitle')}
+          isRequired
+          width="100%"
+          value={title}
+          onChange={setTitleDraft}
+          status={
```
agatha197: With no announcement yet, the red "Please enter a title." shows the moment the modal opens, before the user has typed anything — gate the error `status` behind a touched/blurred flag.

react/src/components/AnnouncementEditModal.tsx
```diff
@@ -35,119 +40,122 @@ import {
...
-      },
-    );
+  const handleSubmit = async () => {
+    if (isTitleMissing) return;
+    setIsPublishing(true);
+    try {
+      const next: DomainAnnouncement = {
+        enabled,
+        title: title.trim(),
+        body,
```
agatha197: `title` is trimmed but `body` is stored verbatim, so a whitespace-only body relies on the banner's own trim to stay non-collapsible — `body: body.trim() || undefined` keeps the stored document clean.

react/src/components/AnnouncementEditModal.tsx
```diff
@@ -225,18 +240,9 @@ const AnnouncementEditModal: React.FC<AnnouncementEditModalProps> = ({
             <Text weight="semibold">{t('summary.AnnouncementMessage')}</Text>
```
agatha197, line 289: The model now pairs `title` with `body`, but this label is still `summary.AnnouncementMessage` ("Message") — renaming to Body/Content would match; it touches all 21 locales, so a separate decision is fine.

agatha197 (changes requested):
Requesting changes for the two login-path issues below (`loginSessionAuth.ts:119` and `:124`) — both are unverified assumptions sitting on the app's only entry path, and both are cheap to guard. The remaining comments are minor cleanups.

Everything else checks out: logic and Relay patterns follow the sibling hooks, i18n is in sync across all 21 locales, regenerated Relay artifacts show no drift, and the `projects` connection uses a single pagination mode (`first`+`after`).

react/src/helper/loginSessionAuth.ts
ironAiken2: Keeping it as is. These ids feed `groupIds` / `current_group_id()` / `user_uuid`, which every project-scoped request uses after login, so an id the manager returns in an unexpected format should fail login loudly here rather than let the session proceed with ids nothing downstream can use.

react/src/helper/loginSessionAuth.ts
ironAiken2: Fixed in `7d5d2c7ff` — the walk now uses `limit`/`offset` bounded by the connection's `count`; `offset` grows every pass, so a manager that repeats a cursor or never ends one cannot spin it. `pageInfo` is gone from the query.

react/src/helper/loginSessionAuth.ts
ironAiken2: Fixed in `7d5d2c7ff` — the query is a module constant (`myUserQuery`) like the sibling hooks.

react/src/helper/loginSessionAuth.ts
ironAiken2: Keeping the single query: the repeat only happens for accounts with more than 100 active projects and re-reads a handful of scalar fields per page. A second, lighter query would add a Relay artifact and a code path for that case; not worth it here.

react/src/components/AnnouncementEditModal.tsx
ironAiken2: Fixed in `7d5d2c7ff` — the error shows only once the title has been edited (`titleDraft !== undefined`), and the input uses `statusVariant="detached"` so the message sits below the field like the form items instead of the bordered block.

react/src/components/AnnouncementEditModal.tsx
ironAiken2: Fixed in `7d5d2c7ff` — `body: body.trim() || undefined`, so a whitespace-only body is dropped from the stored document.

react/src/components/AnnouncementEditModal.tsx
ironAiken2: Keeping `summary.AnnouncementMessage` for now; renaming touches all 21 locales, so I'd rather take it as the separate decision you suggested.

ironAiken2:
Addressed the review-body feedback from @agatha197 in `7d5d2c7ff`:

1. The title validation message now uses `TextInput`'s detached status placement, so it reads like the form items (no bordered block), and it appears only after the title has been edited.
2. The Enabled checkbox is gone. Publish is a split button (same shape as the session launcher's Launch) whose menu offers **Save as Draft** — same document, `enabled: false` — and Delete stands alone on the left. A saved draft is marked with a Draft token in the modal title.

ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3877-pr9527-move-system-announcement.sungchul.10-82-0-159.sslip.io)

react/src/components/AnnouncementEditModal.tsx
ironAiken2: Follow-up in `826d213b5`: the title is now a `BAIFormItem` inside `Form`, so the required error renders as plain text under the field like every other form (the detached Astryx status box was still a tinted block), and the form's validateTrigger keeps an untouched empty title error-free. Publish / Save as Draft run `validateFields()` first.

agatha197 (changes requested):
Image
Could you change “quote” into something easier to understand? At the moment, the color is so faint that it is hard to distinguish.

Collapsed version - centered
Image
Expanded version - move to the top
Image
Please change it so that it always come to the center.

ironAiken2:
Addressed the review from @agatha197 in `ccdfd431f`:

1. **Quote legibility** — a blockquote inside the announcement body now draws its rule in the band's info colour and its text in the primary text colour (`.webui-announcement-body .astryx-blockquote`, applied to both the banner and the editor preview), so it no longer sinks into the info tint.
2. **Centering** — Banner only centres its icon and actions while it has no description, which is why they jumped to the top once expanded; the banner header now keeps `align-items: center` in both states.

Measured on the dev server: icon and Edit sit at the header's vertical centre collapsed and expanded (offset ≤ 0.01px); the quote rule is `rgb(2, 141, 242)` with primary text.

| Collapsed | Expanded |
|---|---|
| collapsed | expanded |

agatha197:
Dev server on box `sujin`: stopped (was http://fr-3877-pr9527-move-system-announcement.sujin.10-82-0-159.sslip.io)

PR #9574, opened 2026-09-10
fix(FR-3897): render the vfolder permission code as an inline code tag

### LINKED ISSUE #9573: Folder invitation card renders permission code "R" as plain text instead of code
In the folder invitation card (the "invited folders" / accept-invitation UI), the permission row shows the human-readable label followed by the raw permission code, e.g. `읽기 전용 R` ("Read Only R"). The trailing `R` is expected to be rendered as an inline code element (monospace / code-styled tag) so it reads as a permission code, but it currently renders as a plain text character appended to the label.

**Observed**

- Folder: `tests`
- Row `권한` (Permission) displays `읽기 전용 R` with `R` in the same plain font as the label.

**Expected**

- The permission code `R` should be rendered with code formatting (inline code / tag), visually distinct from the label text.

Screenshot captured locally by the reporter.

—

JIRA Issue: FR-3897

nowgnuesLee:
Resolves #9573 (FR-3897)

## Problem

The folder invitation card (and the legacy/V2 folder permission cells) show the permission label followed by its code, e.g. `읽기 전용 R`. The `R` / `RW` letters were rendered with Astryx `Text type="code"`, which only switches the font family — no background, border, or padding. A single monospace letter next to a proportional label is indistinguishable from plain text, so the code looked like a stray character appended to the label.

## Fix

Render the letters with `BAIText code` instead. That is the BUI wrapper the other ~30 inline-code call sites already use, and its `.bai-text-code` style reproduces the bordered, muted inline tag antd's `Typography.Text code` produced. Applied to both `VFolderPermissionCell` (invitation card, legacy rows) and `VFolderPermissionCellV2` (folder list) so the two cells stay identical. The migration-era comment blocks in both files were dropped per `comment-density.md`.

## Files

- `react/src/components/VFolderPermissionCell.tsx`
- `react/src/components/VFolderPermissionCellV2.tsx`

### Folder invitation modal (`VFolderPermissionCell`)
| Before | After |
|--------|-------|
| invitation-modal-before | invitation-modal-after |

### Folder list, Mount Permission column (`VFolderPermissionCellV2`)
| Before | After |
|--------|-------|
| folder-list-before | folder-list-after |

The folder list rows are live data. The invitation modal was captured with the invitation-list response stubbed through a Playwright route (two pending invitations, `ro` and `rw`): on the shared test manager every non-admin account is refused vfolder REST calls by RBAC, so no account can actually invite the admin. The modal renders the stubbed rows through the same code path as a live response.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Lint, Format, TypeScript, terminology self-test). Astryx token gate not run: no CSS or `var(--…)` touched.

### DIFF
--- a/react/src/components/VFolderPermissionCell.tsx
+++ b/react/src/components/VFolderPermissionCell.tsx
@@ -4,7 +4,6 @@
  */
 import { VFolderPermissionCellFragment$key } from '../__generated__/VFolderPermissionCellFragment.graphql';
 import { HStack } from '@astryxdesign/core/Stack';
-import { Text } from '@astryxdesign/core/Text';
 import { BAIText } from 'backend.ai-ui';
 import * as _ from 'lodash-es';
 import React, { useMemo } from 'react';
@@ -62,7 +61,7 @@ const VFolderPermissionCell: React.FC<VFolderPermissionCellProps> = ({
 
   return (
     <HStack gap={2} {...props}>
-      <Text>{permissionInfo?.label}</Text>
+      <BAIText>{permissionInfo?.label}</BAIText>
       <HStack>
         {_.map(permissionInfo?.icon, (tag) => (
           <BAIText key={tag} code>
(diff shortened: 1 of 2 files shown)


agatha197 (approved):
LGTM

PR #9577, opened 2026-09-10
docs: update header docs (2026-09-10)

lablup-octodog:
## What changed

Backend.AI WebUI gained a **global search palette**: a **Search** (magnifier) button in the top bar plus a `Ctrl-K` / `Cmd-K` shortcut, gated behind the experimental **Command palette** user setting (off by default). The `header` chapter documented every other top-bar control but had no mention of search, so this PR adds a new `## Search` section (with a stable `<a id="search"></a>` anchor, placed before Notification to match the button's actual position in `WebUIHeader.tsx`) in all four locales, and a new `global_search_palette.png` captured per locale.

## Driving commits

- `82389a0f8` feat(FR-3558): add the global search palette — index-backed search core, UI, and actions (#8812)
- `efb7ae12c` feat(FR-3558): gate the global search palette behind an experimental setting (#8875)
- Context only: `fb61a4daa` (#8811) generates the build-time search index the palette reads; no user-visible surface of its own.

## Deliberately skipped (this chapter)

- No refresh of `header.png`: the Search button only renders while the experimental setting is on, so the default top-bar strip that image shows remains accurate.
- No ranking/scoring details beyond what a user observes, per the plan.
- The **User Settings** side of this feature (the Command palette checkbox itself) belongs to the `user_settings` chapter, reviewed separately; this section refers to it generically without a cross-chapter anchor, as planned.

## Review findings (independent, read-only review)

**No blocking issues.** Detail of what was checked:

- **Claims vs. code** — all verified: the gate and default-off (`GlobalSearchPaletteButton.tsx` early return; `UserSettingsPage` default `false`), the `mod+k` hotkey registered with `allowInInputs: true` (shortcut works from a focused form field), the "Found in …" line shown only when the match came from an item's description rather than a direct title match (`rank.ts` `matchedIn`, direct hits suppress the found-in row), the Administration / Project administration scope tags (`buildHits.ts` `scopeTextOf`), and the empty-query Recent list vs. typed-query "No results" (`emptySearchText`) distinction.
- **Writer's plan deviations** — all three are correct against source and improve accuracy: the i18n keys live under `webui.search.*` (not the plan's `webui.menu.search.*`); "No results" and the empty-query Recent state are separate states, documented separately; **Change UI language** actually lives in the `Appearance` action group (`actions.ts`), not "Panels & help" as the plan said — the prose follows the code.
- **Translation parity** — identical section structure, anchor, admonition, image reference, and bullet lists in en/ko/ja/th. Every quoted UI string in every locale matches the live `resources/i18n/{lang}.json` label exactly (including the ja "実験的特徴" label, which is the live UI string).
- **Screenshots** — `global_search_palette.png` present in all four `src/{lang}/images/` directories, 43–60 KB, 1280×960 each. I opened the en and ko variants: both show the palette in the correct locale with page hits, an Administration scope tag, and a "Found in Session" / "세션에서 찾음" line — exactly what the prose claims. No `TODO(screenshot)` markers.
- **Minor note (non-blocking)** — the `Create` action group can also contain an **Open FastTrack** entry, but only when a FastTrack endpoint is configured (`gate: ctx.config.fasttrackEndpoint` in `actions.ts`); the docs list the three unconditional Create actions. Defensible omission of deployment-specific UI; flagging so the reviewer can decide whether a "(when configured)" mention is wanted.
- **Scope** — the diff touches only `packages/backend.ai-webui-docs/src/{en,ko,ja,th}/header/` and the four new images; no scope creep.

Verified by the wrapper: the diff touches only `packages/backend.ai-webui-docs/`, and `bash scripts/verify.sh` ends with `=== ALL PASS ===`.

> [!NOTE]
> This PR is generated by the nightly docs-sync job and carries no `FR-` number by design; it deviates from the repo's `prefix(FR-XXXX)` title convention and will not be linked by `project-status-sync`.

---


nowgnuesLee (approved):
LGTM

PR #9605, opened 2026-09-11
feat(FR-1564): decide overwrite per duplicate item on folder upload

### LINKED ISSUE #4407: Improve UI for Handling Duplicate Files in Neo FileExplorer
Currently, the Neo FileExplorer provides a bulk option for handling duplicate files, which applies to all files at once. This approach lacks granularity and does not allow users to make individual decisions on how to handle each duplicate file.

To enhance user experience, we propose implementing a modal that will allow users to decide on a case-by-case basis whether to overwrite or cancel for each duplicate file in the list. This improvement aims to provide greater control and flexibility in managing duplicate files, ultimately leading to a more efficient file management process.

**Proposed Solution:**

- Develop a modal interface that appears when duplicate files are detected.
- The modal should display a list of duplicate files with options to "Overwrite" or "Cancel" for each file.
- Ensure the modal is user-friendly and clearly communicates the implications of each choice.

This enhancement will significantly improve the usability of the Neo FileExplorer when dealing with duplicate files.

JIRA Issue: FR-1564

ironAiken2:
Resolves #4407 (FR-1564)

## Summary

The folder explorer asked one bulk question before an upload — "the file or folder with the same name already exists, do you want to overwrite?" — where OK meant overwrite everything and Cancel meant upload nothing. There was no way to keep one file and replace another.

It was also only ever looking at `fileList[0]`. A collision on the second or later file of a multi-file pick was never reported, and that file was overwritten with no prompt at all.

This replaces the single confirm with a per-item decision:

- `useUploadVFolderFiles` groups the pick into the **top-level entries** it will create in the target directory. A folder pick collapses to its root directory name, because that is the only segment that can collide with something in the current directory; a plain pick is one entry per file.
- Every entry is compared against the directory listing, not just the first one.
- Collisions open the new `OverwriteConfirmModal`: one row per colliding entry, the existing item's size and modified time next to the incoming one's, and a checkbox per row. Rows start checked, so confirming without touching anything reproduces the old "overwrite all" outcome.
- Unchecked rows are dropped from the pick. Entries whose name was free upload either way and are reported as a count under the table.
- Cancel still uploads nothing.

The hook moved from the two upload triggers up into `BAIFileExplorer`. It cannot live in `DragAndDrop`: that overlay unmounts on drop, which would take the pending decision with it at the moment the modal is about to open. `DragAndDrop` and `ExplorerActionControls` now just hand their pick to the explorer.

## Screenshots

Re-uploading two files that already exist, alongside one new name. Selected rows are the ones that will be overwritten, so deselecting a row visibly drops it out of the highlight.

| All collisions selected (default) | One collision deselected |
|---|---|
| all selected | one deselected |

## Scope note

Overwriting a file is not reversible, so `destructive-confirmation.md` is worth naming here. This stays on the anchored/modal tier rather than `BAIDeleteConfirmModal` with `requireConfirmInput`: the deliberate pause the rule is after comes from the user having to read a list and choose per row, and a typed confirmation string on every upload of an existing file would make routine re-uploads unusable. The rule's irreversible tier lists deletion and purge flows, and this path never used it.

`CreateFileModal` keeps the single `modal.confirm` for its one-file duplicate check — it creates one named file, so there is nothing to itemize.

## New i18n keys

`comp:FileExplorer.ExistingItem`, `.UploadingItem`, `.SelectItemsToOverwrite`, `.OtherItemsWillBeUploaded`, added to all 21 locales. `DuplicatedFiles` ("Overwrite Confirmation") is reused as the modal title; `DuplicatedFilesDesc` still serves `CreateFileModal`.

## Verification

```
bash scripts/verify.sh
=== ALL PASS ===

undeclared: 2  — both pre-existing (BAIDialog.css, BAIDrawerPortal.css); no new findings

pnpm --filter backend.ai-ui exec vitest run src/components/baiClient/FileExplorer/useUploadVFolderFiles.test.ts
Test Files 1 passed | Tests 7 passed
```

The unit tests cover: no collision uploads straight away; a collision on a later file is detected (the `fileList[0]` bug); unchecked entries are dropped and the rest kept; nothing uploads when all collisions are deselected and no entry is new; cancel uploads nothing; a folder pick is one entry named after its root; an unlistable target directory still uploads as before.

The screenshots above come from driving the real flow against a live cluster: two files uploaded, the same two re-picked with a third new name, one collision deselected, confirmed — after which the new file had arrived and both existing files were still in place.

`e2e/vfolder/file-upload-duplicate.spec.ts` is updated to the new confirmation (the two existing specs asserted the old copy and clicked OK), plus a third spec for the mixed pick. It was not run end to end here: this box's dev session lands already authenticated, so the suite's `loginAsUser` helper finds no login form. Every locator the specs use was exercised live in the probe run that produced the screenshots.

## Review checklist

- [ ] Upload a file whose name already exists and confirm the modal lists it with both sizes and times
- [ ] Upload several files where only a later one collides and confirm that collision is reported
- [ ] Uncheck one row, confirm, and check that only the remaining items reached the folder
- [ ] Cancel the modal and confirm nothing was uploaded
- [ ] Drag-drop a colliding file and confirm the modal opens after the overlay closes

https://claude.ai/code/session_01L3UBa3HnWm5XkSj4jVnUQA


ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-1564-pr9605-decide-overwrite-per.sungchul.10-82-0-159.sslip.io)

yomybaby (approved):
LGTM

PR #9625, opened 2026-09-14
docs: update sessions_all docs (2026-09-14)

lablup-octodog:
## What changed in the product

Backend.AI Manager 26.8.0 adds a preemption surface to compute sessions. The scheduler can now park a session in `RESERVED`, `PREEMPTED`, or `RESCHEDULING`, and the WebUI translates the scheduler's raw `status_info` codes into plain-language reason text in the session status tag's tooltip. Separately, the audit-log status vocabulary gained `DENIED`, which renders as a warning badge instead of being dropped.

The **Compute Sessions** chapter documented neither, and its audit-log status list was a closed enumeration that had become incomplete.

## Driving commits for this chapter

- `b103a1e04` — `feat(FR-3889): render preemption session statuses and reason copy` (#9555)
- `6166c75e4` — `chore(FR-3888): sync data/schema.graphql with the live manager (26.8.x preemption surface)` (#9554) — only the hand-written `BAIAuditLogStatusTag.tsx` / `BAIAuditLogNodes.tsx` part; the other 54 files in that commit are Relay `__generated__` output.

## What this PR changes

`src/{en,ko,ja,th}/sessions_all/sessions_all.md`:

1. New `### Preemption statuses` subsection under **Session detail panel**, placed before `### Session scheduling history` (anchor `preemption-statuses`). Describes the three statuses, states that all three still occupy resources and therefore list under **Running**, and tabulates the four scheduler reason codes against their user-facing descriptions.
2. **Audit log columns → Status** now lists `DENIED`, with a gloss that it is a deliberate policy rejection rather than an error.

- **Adding `DENIED` to the "Audit log filters → Status" list** (a planned edit). The writer deviated from the plan here, and the deviation is **correct** — see findings below.
- **Screenshots.** Both planned entries were marked "no screenshot": a preempted or reserved session cannot be staged on the capture backend, and `session_audit_log.png` only shows `DENIED` if the backend happens to hold a policy-denied entry. No image in this chapter went stale, so no `TODO(screenshot)` marker was owed or left. (The nightly screenshot environment was unreachable this run, but this chapter required no capture regardless.)

## Review findings

Independent review against the source. **No blocking issues.**

**Verified as correct:**

- `RESERVED` / `PREEMPTED` / `RESCHEDULING` all exist in `SessionV2Status` (`data/schema.graphql`) and are gated at 26.8.0 by `this._features['session-preemption-statuses']` in `packages/backend.ai-client/src/client.ts`, whose comment confirms the enum values ship in the 26.8.0 release.
- The **Running, not Finished** claim holds on two independent paths: `react/src/helper/sessionStatusBuckets.ts` lists all three in `RUNNING_STATUSES` (admin / project-admin / agent-drawer surfaces), and `ComputeSessionListPage.tsx`'s `NOT_FINISHED_STATUS_FILTER` (`status != "TERMINATED" & status != "CANCELLED"`) puts them under **Running** on the personal Sessions page.
- All four reason rows match `STATUS_INFO_DESCRIPTION_KEY` in `SessionStatusTag.tsx`, and each locale's descriptions are quoted **verbatim** from `resources/i18n/{en,ko,ja,th}.json` (`session.StatusReason*`) — checked string by string, all four languages.
- "hovering the status tag shows the reason instead of the raw internal value" matches the shipped rendering: on a 26.3.0+ Manager, `SessionDetailContent` passes `showInfo={!supportsSessionSchedulingHistory}` = `false`, so the tag takes the scheduling-history branch where the friendly text is the tooltip and the raw code is not printed.
- `DENIED` is real: `AuditLogStatus` in `BAIAuditLogStatusTag.tsx` includes it and maps it to `warning` ("Denied by policy — deliberate rejection, not a system error"); `toAuditLogStatus` in `BAIAuditLogNodes.tsx` accepts it. The session **Audit Log** tab reaches that badge via `SessionDetailContent.tsx` → `ScopedAuditLog` → `BAIAuditLogNodes`.
- **Bucket labels match the live UI**: the bolded **Running** / **Finished** (and 실행 중 / 종료됨, 実行中 / 終了セッション, กำลังทำงาน / เสร็จสิ้น) are exactly `session.Running` / `session.Finished` in each locale file, and match how the rest of the chapter already writes them.
- **Translation parity is clean.** All four files get the same edits; the heading-level sequence is byte-identical across en/ko/ja/th (38 headings, 25 anchors each), and the new anchor + table structure is present in all four.
- **No scope creep.** The diff is four files, all `sessions_all`.

**The writer's deviation from the plan is right, and the plan was wrong:**

The plan asked for `DENIED` to be added to the **Audit log filters → Status** list too. `react/src/components/ScopedAuditLog.tsx` hard-codes that filter's `options` as `SUCCESS | ERROR | RUNNING | UNKNOWN` with `strictSelection: true`; FR-3888 widened only the badge union and `toAuditLogStatus`, not the filter options. Documenting `DENIED` as filterable would have sent users looking for a choice that does not exist. Leaving line 558 alone is correct. **Product follow-up worth an FR:** a status that can appear in the table but cannot be filtered for is a gap on the product side, not a documentation one.

**Minor, non-blocking — worth a reviewer decision:**

1. The closing `:::note` says that on an older server "the session status tag keeps showing the raw status information value". That is true only below Manager **26.3.0**. Between 26.3.0 (where `session-scheduling-history` turns on) and 26.8.0, `SessionStatusTag`'s scheduling-history branch renders `showTooltip && statusInfoDescriptionKey ? <Tooltip> : badge` — with no mapping key it renders **no tooltip at all**, rather than the raw value. Separately, `STATUS_INFO_DESCRIPTION_KEY` itself is not version-gated (only the status buckets are), so a mapped code from any Manager would still get the friendly text. Suggested softening: "these statuses appear only on Manager 26.8.0 or later" and drop the specific claim about what an older tag falls back to.
2. **Terminology:** the preemption vocabulary is new to the manual in all four languages (ko `선점`, ja `プリエンプション`, th `การเบียดออก`) and has no `concepts[]` entry in `packages/backend.ai-webui-docs/terminology.json`. `verify.sh`'s terminology check passes (no `avoid[]` row is hit), but per `TERMINOLOGY.md`'s new-term gate this should get a curated entry with a `decidingFR` so the next chapter to mention preemption does not coin a different word — especially for th.

## Verification

- The diff touches only `packages/backend.ai-webui-docs/` (this chapter's four language files) — verified by the pipeline wrapper.
- `bash scripts/verify.sh` passes — verified by the pipeline wrapper.

## Convention note

This PR carries no `FR-` number by design: it is produced by the nightly documentation sync, not by a Jira work item. It therefore deviates from the repo's `prefix(FR-XXXX): title` PR-title convention and will not be picked up or linked by the `project-status-sync` workflow.

---


agatha197 (changes requested):
@copilot remove version related info.

agatha197 (approved):

PR #9629, opened 2026-09-14
feat(FR-3921): serve the Portless dev-server list as JSON at /servers

### LINKED ISSUE #9628: Portless proxy: serve the live dev-server list as JSON at /servers for the Jira helper and devbox gateway
## Summary

Portless (the dev-server proxy used by `pnpm run dev`) has no machine-readable way to list the dev servers it currently routes. `portless list` is text-only with no `--json`, and the only place the route list appears over HTTP is the 404 HTML page (a 150 KB page with embedded fonts). Tooling that wants to know "which dev servers are up on this box" — the Jira helper browser extension's Dev server row and the devbox gateway's `/servers/` path — has to scrape that HTML.

## Proposal

Patch Portless (via `pnpm patch`, kept under `react/patches/`) so the proxy answers `GET /servers` on any host that does not match a registered app with a JSON array of live routes:

```json
[{"hostname": "fr-1234-pr5678-word.localhost", "port": 4947, "pid": 26560, "url": "https://fr-1234-pr5678-word.localhost:1355"}]
```

- Served from both the HTTP/1 and HTTP/2 request handlers; query strings ignored; dead-pid routes are already filtered by Portless.
- Requests to a registered app host are untouched, so an app's own `/servers` route is not shadowed.
- Non-matching hosts on other paths keep the existing 404 HTML.

## Consumers this unblocks

- Jira helper extension: read the dev-server list with one `fetch` instead of parsing the 404 page.
- Devbox gateway: forward `<box>.<domain>/servers` straight to the box's Portless instead of scraping HTML on the gateway side.

## Follow-up

Contribute the same endpoint upstream to Portless so the pnpm patch can be dropped on the next version bump.

—

JIRA Issue: FR-3921

nowgnuesLee:
Resolves #9628 (FR-3921)

## Problem

`portless list` is text-only (no `--json`), and the only place the Portless proxy exposes its live route list over HTTP is the 404 HTML page it renders for an unknown host. Tooling that wants to know which dev servers are running on a box — the Jira helper extension's Dev-server row and the devbox gateway's `/servers/` path — has had to scrape that 150 KB page.

## Change

A pnpm patch on `portless@0.15.5` (`react/patches/portless@0.15.5.patch`, registered in `pnpm-workspace.yaml`) makes the proxy answer `GET /servers` (also `/servers/`, query string ignored) on any host that matches no registered app:

```json
[{"hostname":"fr-3737-pr9476-name-why-add.localhost","port":4947,"pid":26560,"url":"https://fr-3737-pr9476-name-why-add.localhost:1355"}]
```

- Added to both the HTTP/1 request handler and the HTTP/2 compat handler; the route list is the same live, dead-pid-filtered `getRoutes()` the 404 page already uses.
- Requests to a registered app host are not touched, so an app's own `/servers` route is not shadowed.
- Other paths on an unknown host keep the existing 404 HTML.
- The `pnpm-lock.yaml` diff is only the `patch_hash` registration.

Portless 0.15.6 (current upstream) has no equivalent, so the same change should go upstream and this patch dropped on the next bump.

## How to use

```bash
curl -sk https://localhost:1355/servers | jq
```

## Verification

Live on the 1355 daemon started from this checkout:

- `https://localhost:1355/servers` with no apps → `[]`, 200, `application/json`
- with a registered app → the array above, identical over HTTP/2, HTTP/1.1 and with `?x=1`
- `https://<app>.localhost:1355/servers` → the app's own response (not shadowed)
- `https://localhost:1355/other` → 404 `text/html` as before
- `pnpm install --frozen-lockfile --offline` passes

Caveats: the daemon runs from whichever checkout first ran `pnpm run dev`, so a checkout without this patch installed serves no `/servers`; `pnpm install` there fixes it.

## Checklist

- [ ] Minimum requirements for docs
- [x] Minimum requirements for i18n (not applicable)


yomybaby (approved):
LGTM

PR #9639, opened 2026-09-14
feat(FR-3914): add missing filters, orderings and columns to the Deployments page

### LINKED ISSUE #9615: Add missing filters, orderings and columns to the Deployments page tabs
Page: /admin/deployments. Source: report-deployments.md. Depends on the gating fix issue for the shared feature flags.

**Deployments tab**:

- createdUserId filter (user select) gated >=26.4.3 (`model-deployment-extended-filter`).
- replicas nested filter (replica status / health status / traffic status) gated >=26.8.0 (`deployment-replica-nested-filter`, flag exists).
- DESTROYED_AT ordering + a destroyedAt column (hidden by default), gated >=26.4.3.

**Deployment Presets tab**: add a `rank` column with RANK ordering (already in the sorter allow-list, column missing).

**Model Store tab**: render the project name (project.basicInfo.name) instead of the raw project UUID; add updatedAt column (hidden by default).

JIRA Issue: FR-3914

yomybaby:
Resolves #9615 (FR-3914)

> Stacked on top of #9638 (FR-3913), which adds the feature flags this PR gates against. Review/merge that one first.

## Summary

Everything the five `/admin/deployments` tabs' root fields support but the UI never exposed, each behind the flag its schema addition belongs to.

### Deployments tab

- **`createdUserId` filter** — driven by `BAIUserSelect` the same way the RBAC page's assigned-user filter is (`valuePropName="id"`, forwarding the option label so the condition chip reads as an email rather than a UUID). Labelled *Owner*, matching the existing `owner` column. Gated on `model-deployment-extended-filter` (26.4.3).
- **`replicas` nested filter** — replica lifecycle status, health status and traffic status, as strict-selection enum properties. Gated on `deployment-replica-nested-filter` (26.8.0).
  The keys nest **three** levels: `replicas.some.status` → `{replicas:{some:{status:{equals:…}}}}`. `buildNestedFilter` already builds arbitrary depth from the dotted key and `extractNestedConditions` already recurses back out of it, so **no change to `BAIGraphQLPropertyFilter` was needed** — verified live, see the table.
  Option labels reuse the existing `replicaStatus.*` keys, so no new enum strings.

### Deployment Presets tab

- **`rank` column** (with the `RANK` sorter). `'rank'` was already in `availablePresetSorterKeys`, but no column carried that `dataIndex`, so the ordering was **unreachable** — this is what makes it reachable. Column header carries the same "lower first" tooltip the Runtime Parameters table uses.

### Model Store tab

- The **project column renders `project.basicInfo.name`** instead of the raw UUID. `ModelCardV2.project` is `@since(26.4.3)`, so the query spreads it under that directive and the column falls back to the copyable UUID when it resolves to null — i.e. exactly the old behaviour on an older manager.
- **`updatedAt` column**, hidden by default.

### Not shipped — `DESTROYED_AT` ordering and a `destroyedAt` column

The issue asks for both. The supergraph has `DeploymentFilter.destroyedAt` (input) and `DeploymentOrderField.DESTROYED_AT` (enum), **but no `destroyedAt` output field anywhere** — not on `ModelDeployment`, not on `ModelDeploymentMetadata` (`grep -n destroyedAt data/schema.graphql` returns exactly one line, the filter input). There is nothing to render in the column, and a `BAITable` sorter has to hang off a column's `dataIndex`, so the ordering has no reachable trigger either — it would be the very `rank` situation this PR fixes on the presets tab.

The `destroyedAt` **filter** is already exposed (gated, since 26.4.3) and keeps working. Adding the column and the sorter needs a backend field first.

## Live verification

Manager **26.8.0rc1** (`http://10.82.0.130:8090`), super-admin. Argument shapes replayed against the live `/func/admin/gql` through the page's session; the new controls also driven through the real UI.

| item | kind | server ok? | result changed? | note |
|---|---|---|---|---|
| **`createdUserId` — end-to-end through the UI** | filter | ✅ | 2 → 1 row | picked *admin@lablup.com* in the user select → URL became `{"createdUserId":{"equals":"f38dea23-…"},"status":{"notIn":["STOPPED"]}}`, table showed only that owner's deployment |
| `createdUserId equals` (2 real ids) | filter | ✅ | 35 → 3 and 35 → 32 | two different owners, disjoint as expected |
| `createdUserId equals` (all-zero uuid) | filter | ✅ | 35 → 0 | no-match case |
| `replicas.some.status` = `TERMINATED` | filter | ✅ | 35 → **3** | |
| `replicas.some.status` = `FAILED_TO_START` | filter | ✅ | 35 → **2** | |
| `replicas.some.status` = `RUNNING` / `PROVISIONING` / `TERMINATING` | filter | ✅ | 35 → 0 | no such replicas on this manager |
| `replicas.some.healthStatus` = `NOT_CHECKED` | filter | ✅ | 35 → **5** | |
| `replicas.some.healthStatus` = `HEALTHY` / `UNHEALTHY` / `DEGRADED` | filter | ✅ | 35 → 0 | |
| `replicas.some.trafficStatus` = `INACTIVE` | filter | ✅ | 35 → **5** | |
| `replicas.some.trafficStatus` = `ACTIVE` | filter | ✅ | 35 → 0 | |
| 3-level nesting combined with `AND` | filter | ✅ | 35 → 3 | `AND: [{replicas:{some:{status:{equals:TERMINATED}}}}, {openToPublic:false}]` |
| `RANK` ASC / DESC on deployment presets | order | ✅ | count 5 | |
| **`rank` column renders** | column | ✅ | — | headers `["Name","Runtime","Rank","Image","Replica Count","Created At","Modified At"]`, first row rank `100` |
| model cards: `project { basicInfo { name } }` + `updatedAt` selected | column | ✅ | — | query accepted, no errors; **the store has 0 model cards on this manager**, so the rendered name could not be observed on a real row (the UUID fallback path is what the old code did) |
| all 5 tabs after the change | page | ✅ | — | no error banner, no GraphQL errors |

**0 GraphQL errors across the sweep.**

### One thing the reviewer should know

`deployment-replica-nested-filter` is pinned to `26.8.0`, and the test manager is **26.8.0rc1**, which sorts *before* the final tag — so `baiClient.supports('deployment-replica-nested-filter')` is **`false`** there and the three replica properties are correctly hidden in the UI. That is the gate working, but it also means they cannot be exercised through the UI on this manager; the results above are the same argument shapes sent straight to that manager's GraphQL endpoint, which accepts them and filters correctly.

The flag threshold is pre-existing (FR-3332) and shared with `ChatPage` / `Chat/DeploymentSelect`, so I did not change it here. The repo does have precedent for pinning to an rc tag exactly for staging managers (`26.4.4rc7`, `26.4.4rc9` in `client.ts`); if you want the replica filters visible on the 26.8.0rc* test servers, that is a one-line change to `'26.8.0rc1'` — say the word and I will add it (here or in a follow-up), noting it also switches the two Chat call sites on.

## Verification

```
bash scripts/verify.sh
=== ALL PASS ===
```

`pnpm relay` output and `make i18n`-propagated locale keys are committed. No CSS / token changes.

**Checklist:**

- [ ] Documentation
- [x] Minimum required manager version — new capabilities gate themselves; nothing regresses below 26.4.0
- [x] Specific setting for review — dev server below, super-admin, `/admin/deployments`
- [x] Minimum requirements to check during review — Owner filter on the Deployments tab, Rank column on Deployment Presets, Project column on Model Store
- [x] Test case(s) to demonstrate the difference of before/after — live table above

🔗 **Dev server**: http://fr-3914-pr9639-add-missing-filters.jongeun.10-82-0-159.sslip.io


react/src/components/AdminDeploymentPresetTable.tsx
```diff
@@ -151,6 +155,23 @@ const AdminDeploymentPresetTable: React.FC<AdminDeploymentPresetTableProps> = ({
...
       },
+      {
+        key: 'rank',
+        title: (
+          <BAIFlex gap="xxs" align="center">
+            {t('adminDeploymentPreset.Rank')}
+            <Tooltip content={t('adminDeploymentPreset.RankTooltip')}>
+              <Text color="placeholder" style={{ cursor: 'help' }}>
+                <CircleHelp size="1em" />
+              </Text>
```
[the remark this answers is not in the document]
yomybaby: Fixed in `ce4ef6e` — replaced the hand-rolled `Tooltip` + non-focusable `Text` trigger with `BAIQuestionIconWithTooltip`, which is what the analogous rank header in `BAIRuntimeVariantPresetTable.tsx` already uses. It renders the same `CircleHelp` glyph on the focusable reset button from `BAIIconWithTooltip`, with an `aria-label` derived from the tooltip content.

The `Text`, `Tooltip` and `CircleHelp` imports were only used by that block, so they are gone from the file. `bash scripts/verify.sh` → `=== ALL PASS ===`.

yomybaby:
Dev server on box `jongeun`: http://fr-3914-pr9639-add-missing-filters.jongeun.10-82-0-159.sslip.io
serves stack #9638 → #9639 (running: #9639)

agatha197:
@copilot translate all added keys. There're some __NOT_TRANSLATED__ and missing keys.

resources/i18n/en.json
```diff
@@ -43,7 +43,7 @@
     "PresetUpdated": "Deployment preset has been updated.",
     "PresetValues": "Preset Values",
     "Rank": "Rank",
-    "RankTooltip": "Display ordering among presets of the same runtime. Lower values appear first.",
+    "RankTooltip": "Display ordering among presets of the same runtime variant. Lower values are shown first.",
```
agatha197: "runtime variant" is not a term this UI uses — the column right beside it is `adminDeploymentPreset.Runtime` = "Runtime", and `SelectRuntimeVariant` reads "Select runtime". Please keep "the same runtime" here, or register "runtime variant" in `packages/backend.ai-webui-docs/terminology.json` first (FR-3840 put "runtime variant preset" on the avoid list for the same reason).

resources/i18n/ko.json
```diff
@@ -43,7 +43,7 @@
     "PresetUpdated": "배포 프리셋이 업데이트되었습니다.",
     "PresetValues": "프리셋 값",
     "Rank": "순위",
-    "RankTooltip": "같은 런타임의 프리셋 간 표시 순서입니다. 값이 낮을수록 먼저 표시됩니다.",
+    "RankTooltip": "같은 런타임 배리언트의 프리셋 중 표시 순서입니다. 값이 작을수록 먼저 표시됩니다.",
```
agatha197: "런타임 배리언트" would be the first "배리언트" in any locale file; the existing Korean for this concept is "런타임" (`PortInheritTooltip`) or "인퍼런스 런타임 종류" (`EnvironmentVariableNotForRuntime`). Please keep "같은 런타임의".

agatha197 (changes requested):
Checked the new filters, the `rank` column/sorter and the model-card columns against the schema and the flags — those look right. One terminology issue in the tooltip string, which this PR newly surfaces in the column header.

resources/i18n/en.json
yomybaby: Fixed in `ecfe0eae7` — restored main's text: `adminDeploymentPreset.RankTooltip` now reads "Display ordering among presets of the same runtime. Lower values appear first." No 'runtime variant' left in this key. The other 20 locales were never changed by this PR and already say 'same runtime' (ja 同じランタイム, de derselben Runtime, …), so only en/ko were touched.

resources/i18n/ko.json
yomybaby: Fixed in `ecfe0eae7` — restored main's text: ko `RankTooltip` now reads "같은 런타임의 프리셋 간 표시 순서입니다. 값이 낮을수록 먼저 표시됩니다." '배리언트' is gone. Only en/ko were touched; the other locales were unchanged by this PR and already use their 'same runtime' wording.

yomybaby:
Rebased the stack (#9638 → #9639) onto current `main`; #9639 is mergeable again.

- Conflict: `react/src/__generated__/AdminModelCardQuery.graphql.ts` (Relay output) — took one side and regenerated with `pnpm run relay`, no hand-merge.
- Also fixed formatting drift the stack carried that made `verify.sh`'s Format step fail: `style(FR-3913)` `10e7a1205` (prettier on `BAIModelDeploymentNodes.tsx`) and `style(FR-3914)` `e1d0502d0` (prettier on `AdminModelCard.tsx` + key sort of the translated `modelService.Replica*Status` keys in 19 locales, no text change).
- `bash scripts/verify.sh`: `=== ALL PASS ===` on both branches.

PR #9393, opened 2026-09-02
fix(FR-3841): show LLM sampling parameter labels in English in every locale

ironAiken2:
Resolves #9392 (FR-3841)

## What changed

The Chat page's **Parameters** popover labels — Max Tokens, Temperature, Frequency Penalty, Presence Penalty — are universal LLM vocabulary that users recognise by the English name, so they are shown in English in every locale instead of translated. Top K / Top P were already English everywhere.

- `resources/i18n/*.json` (20 non-en catalogs): `chatui.chat.parameter.label.{MaxTokens,Temperature,FrequencyPenalty,PresencePenalty}` set back to the en value. The per-parameter descriptions (`chatui.chat.parameter.{MaxTokens,…}`) and the popover title stay translated — they are prose, not parameter names.
- `packages/backend.ai-webui-docs/src/ko/chat/chat.md`: the one prose line naming the fields ("최대 토큰, 온도") now uses the labels the UI shows; en/ja/th prose already did.

## Context

ko, ja, zh-CN and zh-TW carried translations from the start; FR-3839 (#9389) listed these keys under "values left in English" and translated the remaining 16 locales. This reverses that part on purpose — the keys were English by design, not by omission — and FR-3839 carries a comment recording the carve-out. PR #9309 (nightly docs refresh of `parameter_settings.png`) was closed because its ko/ja captures show the translated labels; the nightly docs job will re-capture once this lands.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0191uajZ5hZ1uo6mQyqNvRMU

PR #9389, opened 2026-09-02
fix(FR-3839): translate the catalog values left in English and add the missing non-en keys

ironAiken2:
Resolves #9387 (FR-3839)

The review of PR #9383 found `comp:FileExplorer.RenameFile` holding the English "Rename" in ko/ja/th. A full audit of both catalog sets against `en.json` (3,566 webui keys and 619 BUI keys, 20 non-en catalogs each) turned that one key into a class: values never translated, keys never added to the non-en catalogs, and a few English-source defects. This PR clears the class and adds a guard against the part of it that comes back silently.

## Keys missing from the non-en catalogs (they fell back to English at runtime)

- **BUI**: `comp:BAIColorPicker.{Clear,HexValue,NoColor,SelectColor}` and `comp:BAIPropertyFilter.{Apply,SearchLabel,operator.*}` — 14 keys, missing from **all 20** catalogs. The operator labels reuse each catalog's existing `comp:BAIGraphQLPropertyFilter.operator.*` translations.
- **webui**: `modelService.Command` (reuses `deployment.Command`) and `modelService.PreStartActionsTooltip` in 19 catalogs; `general.Value` in th.

## Values left in English, now translated

- **webui**: `aiAgent.*` (21 keys, 19 langs), `sTokenLoginBoundary.*` (15, 19 langs), `eduapi.*` (3), `chatui.chat.parameter.label.*` (4), `diagnostics.Csp*Passed` (3), `deployment.Revision*` / `import.HuggingFaceRevision`, `data.UserQuotaScopeId` (reuses `data.QuotaScopeId`), `session.launcher.ApplicationMemory`, `time.Day`/`Min`/`Ms`, and a per-language tail: the `rbac.types.*` object types for el/id/it/mn/ms/th/tr/vi, the ko session-type labels (`Interactive`/`Batch`/`Inference` → 대화형/배치/추론 per `terminology.json`), and the mn/th/fi/ru/el/pl leftovers listed in the issue.
[...]

PR #9650, opened 2026-09-14
fix(FR-3918): fix Created At / Updated At sorting on the RBAC roles list and widen the scope type filter

### LINKED ISSUE #9619: Fix Created At / Updated At sorting on the RBAC roles list and widen the scope type filter
Page: /rbac. Source: report-projects-rbac-reservoir-scheduler.md section 2.

- Bug: the createdAt / updatedAt columns emit camelCase order strings while `availableRoleSorterKeys` holds snake_case, so nuqs drops the value and sorting is a no-op. Align the keys (sortKey or allow-list).
- The mappedScope.scopeType filter offers 10 of the 46 RBACElementType values (list borrowed from the create-role form). Offer the full enum for filtering.

JIRA Issue: FR-3918

yomybaby:
Resolves #9619 (FR-3918)

### 1. Created At / Updated At sorting was a no-op (bug)

`BAITable` builds the order string from `sortKey ?? String(dataIndex)`. The two
date columns in `RoleNodes.tsx` declare `dataIndex: 'createdAt'` / `'updatedAt'`
and no `sortKey`, so the table emitted `createdAt` / `-updatedAt` — while
`availableRoleSorterKeys` held the snake_case `created_at` / `updated_at`.
`RBACManagementPage` stores the value through
`parseAsStringLiteral(availableRoleSorterValues)`, and nuqs returns `null` for a
non-member, so the value round-tripped to `null`: `orderBy` was dropped from the
query and the header showed no sort arrow. Only `name` ever worked.

**Fix — align the allow-list on the camelCase spellings** rather than adding
`sortKey: 'created_at'` / `'updated_at'` to the columns.

*Why this side, on URL compatibility:* camelCase is the only spelling that has
ever been written to the URL. The table wrote `?order=createdAt`; nothing ever
wrote `?order=created_at`, because the only producer is the table itself. So
already-shared links carry `createdAt` / `-updatedAt` and start working with
this change, whereas the `sortKey` fix would leave them parsing to `null`
forever. `convertToOrderBy` snake-cases + uppercases either spelling, so the
document sent to the manager is identical (`CREATED_AT` / `UPDATED_AT`).

### 2. The scope type filter offered 10 of 45 values

`mappedScope.scopeType`'s options came from `RBAC_ELEMENT_TYPES` — a whitelist
picked for the **role-creation form**, where every entry needs a UI-side
scope-id selector. As a *filter* that constraint does not apply: the server
accepts any `RBACElementType`, so filtering roles by e.g. `IMAGE` or `ARTIFACT`
scope was impossible for no reason.

New `react/src/helper/rbacElementTypes.ts` exports `ALL_RBAC_ELEMENT_TYPES`
(all 45 values, SDL order), and `rbacElementTypes.test.ts` keeps it honest: it
parses `data/schema.graphql` and asserts the constant equals the enum, and that
every value has an English label.

**i18n — no humanized fallbacks were needed.** 30 of the 45 already had a
`rbac.types.*` label; the 15 that did not get one in `en` + `ko`
(`IDLE_CHECKER_ASSIGNMENT`, `APP_CONFIG_DEFINITION`, `APP_CONFIG_ALLOW_LIST`,
`APP_CONFIG_FRAGMENT`, `MODEL_CARD`, `KERNEL_HISTORY`, `PROJECT_ADMIN_PAGE`,
`DOMAIN_ADMIN_PAGE`, `DEPLOYMENT_TOKEN`, `DEPLOYMENT_POLICY`,
`DEPLOYMENT_REVISION`, `IMAGE_ALIAS`, `VFOLDER_DATA`, `SESSION_APP_SERVICE`,
`USER_EMAIL`). The existing `{ defaultValue: type }` fallback stays as the
safety net for a value added to the schema later.

No gating change: the sorting fix is on fields available since ≤26.4.0, and the
scope-type property keeps its existing `supports('role-mapped-scope-filter')`
(26.8.0) gate.

## Live verification

Manager **26.8.0rc1** at `http://10.82.0.130:8090`, superadmin, `/rbac`,
83 active roles. Queries issued through the app's own GraphQL client;
`ok` = the response carried no `errors`.

| item | kind | server ok? | result changed? | note |
|---|---|---|---|---|
| `createdAt` header click (asc) | order (fixed) | yes | yes | URL becomes `?order=createdAt`, arrow appears, first rows `role_monitor, role_project_2de2b969_member, role_project_8e32dd28_member` — previously no arrow and no re-query |
| `createdAt` header click (desc) | order (fixed) | yes | yes | `?order=-createdAt`, first rows `preset_user, user-edd3a14d, user-dbb5b5f9` |
| `updatedAt` header click (asc/desc) | order (fixed) | yes | yes | `?order=updatedAt` / `?order=-updatedAt`, arrow state follows |
| `orderBy: CREATED_AT ASC` / `DESC` | order (server) | yes | yes | direct query — both orders differ, 83 rows |
| `orderBy: UPDATED_AT ASC` / `DESC` | order (server) | yes | yes | direct query — both orders differ, 83 rows |
| `orderBy: NAME ASC` | order (existing) | yes | yes | regression check |
| no `orderBy` | order (existing) | yes | n/a | regression check |
| `mappedScope.scopeType.equals` — **all 45 enum values** | filter (widened) | **yes, 45/45** | yes | `DOMAIN` 7 rows, `PROJECT` 31, `USER` 20; the other 42 return 0 rows on this deployment, none returns an error |
| Scope Type option list renders all 45 | filter UI | n/a | n/a | see the gating note below |

**Gating note.** `role-mapped-scope-filter` requires manager **26.8.0**, and the
test manager reports `26.8.0rc1`, which is *below* 26.8.0 in semver —
`baiClient.supports('role-mapped-scope-filter')` is `false` there, so the Scope
Type / Scope ID properties do not render on this backend. Verified two ways:

1. **Unit test** — `react/src/helper/rbacElementTypes.test.ts` (2 tests, both
   pass): the constant matches the SDL enum, and every value has an English
   label.
2. **Forcing the flag in the dev build** — the gate was temporarily replaced
   with `true`, the page reloaded, and the Scope Type value dropdown listed all
   45 options with readable labels (`Domain … Artifact Revision`, including the
   15 newly labelled ones). The override was then reverted; `git diff` against
   the committed tree is clean, and with the gate restored the property list is
   back to `Role Name / Source / Assigned User`.

Screenshot: `/home/ubuntu/.claude/jobs/a027968f/tmp/shot-FR-3918.png`

## Verification harness

`pnpm exec vitest run src/helper/rbacElementTypes.test.ts` → 2 passed

🔗 **Dev server**: http://fr-3918-pr9650-fix-created-updated.jongeun.10-82-0-159.sslip.io


yomybaby:
Addressing the suppressed finding in @copilot's review body (`resources/i18n/en.json:2366` — "add the new \`rbac.types\` entries to every supported locale or explicitly change the translation contract").

**Not applying** — this is the repository's translation contract working as designed, and the same call PR #9636's ready gate made on the identical finding.

Three specifics:

1. **`make i18n` cannot produce these keys.** `RBACManagementPage.tsx` looks them up as `t(\`rbac.types.${type}\`)` — a template literal. `i18next-scanner` extracts static keys only, so `make i18n` never sees `rbac.types.*` and will not write the `__NOT_TRANSLATED__` placeholder `i18n.config.js` returns for non-`en` locales. Absent is the only state these keys can have in the other catalogs; there is no generated placeholder to fill.

2. **The fallback is an English label, not a raw identifier.** `fallbackLng: 'en'` (`react/src/components/DefaultProviders.tsx:173`), so a `ja` user resolving `rbac.types.MODEL_CARD` gets `"Model Card"` from `en.json`. The `{ defaultValue: type }` in the call is the second-level net for a value added to the schema later, not the first. This is also the pre-existing state, not something this PR introduces: the other 20 catalogs already carry 31 of the 46 `rbac.types` keys and have always fallen back for the rest.

3. **`en` and `ko` are the locales this project hand-translates**; the rest are filled in by dedicated translation PRs (#9389, #9391, #9393) rather than by the feature PR that introduces a key. Machine-writing 15 labels × 19 locales here would be unreviewable scope creep in a sorting fix.

`react/src/helper/rbacElementTypes.test.ts` already enforces the part that is enforceable — every one of the 46 enum values has an English label — and `bash scripts/verify.sh` passes (`=== ALL PASS ===`).

yomybaby:
Dev server on box `jongeun`: http://fr-3918-pr9650-fix-created-updated.jongeun.10-82-0-159.sslip.io

agatha197 (changes requested):
@copilot translate other languages.

agatha197:
Dev server on box `sujin`: stopped (was http://fr-3918-pr9650-fix-created-updated.sujin.10-82-0-159.sslip.io)

agatha197 (approved):
LGTM

PR #9661, opened 2026-09-14
fix(FR-3931): wrap the property filter editor row instead of clipping it

### LINKED ISSUE #9660: Property filter editor popover clips the field selector and value input
## Symptom

Opening the filter editor popover of a GraphQL property filter (`BAIGraphQLPropertyFilter` / `BAIPropertyFilter`, both built on Astryx `PowerSearch`) with a narrow search box shows the field selector shrunk to a few characters (e.g. `M…`) and the value input (date picker) cut off at the right edge of the popover. Seen on the Resource Policy list with a datetime / number property selected.

## Cause

Astryx `PowerSearch` sizes the editor popover to the anchor width with a 400px floor, and its field / operator / value row only wraps when the popover is **narrower** than 400px. At exactly the floor the row stays on one line: the operator selector does not shrink and `DateInput` has a 180px minimum, so the field selector takes the leftover ~20px and the value editor overflows the `overflow: hidden` container.

## Expected

The three editor controls are fully readable at any search-box width — either the row wraps or the popover is wide enough for its content.

## Scope

Fix on the BUI side (`packages/backend.ai-ui`), since `@astryxdesign/core` is an external dependency.

JIRA Issue: FR-3931

agatha197:
Resolves #9660 (FR-3931)

## Summary

The filter editor popover of both property filters (`BAIPropertyFilter`, `BAIGraphQLPropertyFilter`) clipped its controls whenever the search box was narrow: the field selector shrank to a couple of characters (`C.`) and the date input ran past the popover's right edge.

A filter token wider than the search box also ran past the box's border, cut off without an ellipsis.

Astryx `PowerSearch` sizes the popover to the anchor width with a 400px floor and wraps the field / operator / value row only when the popover is *narrower* than 400px. At the floor the row stays on one line, the operator selector does not shrink (`flexShrink: 0`) and `DateInput` has a 180px minimum, so the field selector takes whatever is left and the value editor overflows the `overflow: hidden` container.

## Changes

- `BAIPowerSearchAdapters.tsx`: `BAIPowerSearchEditor` wraps Astryx's default `PowerSearchFilterEditor` in a `.bai-power-search-editor` class hook, and `baiPowerSearchComponents` registers it as the `components.Editor` for every operator value type (typed as the full `Record`, so a new Astryx type fails `tsc` instead of silently missing the hook).
- `BAIPowerSearchAdapters.css`: lets the editor row wrap on content (`@layer components` beats `astryx.css`, whose rules live in `@layer astryx-base`).
- Both filters pass `components={baiPowerSearchComponents}` to `PowerSearch`.
- Both filters also carry a `.bai-power-search` root class. Astryx `Tokenizer` wraps each token in a non-shrinking span sized to its content (so `Token`'s own `max-width: 100%` never bites), and `Token`'s `all: unset` click target inlines the label, which disables its `text-overflow: ellipsis`. The stylesheet caps the wrapper at the box width and makes the label a block so it truncates with an ellipsis.
- Tests: the wrapper renders the hook around the editor, every operator value type is registered, and both filters render the root class hook.

Rows that already fit on one line render exactly as before (checked with a 600px anchor). `@astryxdesign/core` is an external dependency, so the fix lives in BUI rather than upstream.

## Screenshots

Storybook `Filter/BAIPropertyFilter → NumberAndDatetime`, search box forced to 240px, editor opened on **Created At**.

**Editor popover**

| Before | After |
|--------|-------|
| before | after |

**Tokens in the search box** (same story, editor closed)

| Before | After |
|--------|-------|
| tokens before | tokens after |

## Verification

- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → no new findings
- `pnpm exec vitest run` on `BAIPowerSearchAdapters`, `BAIGraphQLPropertyFilter`, `BAIPropertyFilter` → 110 passed

https://claude.ai/code/session_01TnmniN15v43A9qRFAE3xF4


nowgnuesLee (approved):
LGTM

PR #9664, opened 2026-09-14
feat(FR-3676): explain how to pass a JSON string in the start command hint

### LINKED ISSUE #9064: Explain how to pass a JSON string in the launch command's help text
Users frequently ask how to enter a JSON string as input for the launch (run) command, and this generates a steady stream of customer inquiries.

Add guidance on how to pass a JSON string — quoting/escaping rules and a worked example — directly where the user needs it: the launch command field's help text, tooltip, placeholder, or a similar inline hint, so the answer is discoverable without contacting support.

—

Captured while working on branch: feat/FR-3668-substep-3a-fullscreen-modal

JIRA Issue: FR-3676

nowgnuesLee:
Resolves #9064 (FR-3676)

## Summary

Users kept contacting support to ask how to pass a JSON string (for example vLLM's `--override-generation-config`) through the model service start command. This PR puts the answer in the form field where they type the command.

- Adds a new `modelService.CommandJsonArgumentHelper` hint, translated into all 21 locales: *"To pass a JSON string as a single argument, wrap the whole JSON in single quotes and keep the double quotes inside as-is (e.g. `--override-generation-config '{"temperature": 0.7}'`)."*
- Shows the hint as a second line under the existing execution-mode helper on the shared Service Configuration **Command** field (`ServiceConfigurationFormItems`). That field is used in the deployment revision modal and in the admin deployment preset's model config.
- Shows the same hint under the admin deployment preset's **Startup Command** field (`AdminDeploymentPresetSettingPageContent`), which already reused the shell-syntax helper.
- One hint covers every execution mode, because single-quoted JSON stays one argument on all three paths: Shell (`bash -c`), Exec (`shlex.split`), and the WebUI tokenizer used for pre-26.8.0 managers. A new `tokenizeShellCommand` test covers the tokenizer path for the documented example.
- `react/src/generated/searchIndex.json` was regenerated to pick up the new i18n key.

## Scope note

The issue says "the launch command field". This PR reads that as the model service start command, whose Korean label is "실행 명령어". The session launcher's batch-mode **Startup Command** is left unchanged. If that field should get the hint as well, it is a one-line follow-up.

## Verification

- `bash scripts/verify.sh`: `=== ALL PASS ===`
- `pnpm exec vitest run src/helper/parseCliCommand.test.ts`: 44 passed
- `pnpm exec vitest run src/form-engine/formEngineAcceptance.test.tsx`: 31 passed
- Not checked in a browser. This run was headless and no dev server was running.

🔗 **Dev server**: http://fr-3676-pr9664.seungwon.10-82-0-159.sslip.io


nowgnuesLee:
## Copilot review triage (pass 1)

Copilot reviewed this draft at "Lite" level. It opened no inline threads. Its 5 findings are only in the review body, under "Suppressed comments", so here is an answer to each. No code changed in this pass.

### Needs a human decision (3)

| # | File:line | Copilot says | Why I left it open |
|---|-----------|--------------|--------------------|
| 1 | `ServiceConfigurationFormItems.tsx:185` | The hint is wrong for JSON that contains a single quote (e.g. `{"prompt":"don't stop"}`). Every path (the tokenizer, `bash -c`, `shlex.split`) splits a single-quoted token at an embedded `'`. Copilot suggests limiting the hint to JSON without single quotes or documenting an escaped fallback. | The technical point is correct. Fixing it changes user-visible copy in 21 locales, and the hint is meant to stay a single line. Options: (a) keep it as is, since the common vLLM case has no `'`; (b) add "if the JSON has no single quotes"; (c) document the `'\''` escape. A person should choose. |
| 2 | `AdminDeploymentPresetSettingPageContent.tsx:1187` | The batch-mode session launcher's `StartUpCommand` field (`SessionLauncherPage.tsx`) still has no hint, so batch users hit the same problem. | The PR body's "Scope note" leaves that field out on purpose and calls it a one-line follow-up. Whether to widen the scope is a product decision. |
| 3 | `ServiceConfigurationFormItems.tsx:175` | On pre-26.8 managers the command goes out as a tokenized argv `startCommand` with no shell. Copilot says the `StartCommandHelperShell` hint wrongly suggests shell operators work there and wants an argv-specific helper. | My read is that the hint still holds: `/bin/bash -c "cmd1; cmd2"` tokenizes to `["/bin/bash","-c","cmd1; cmd2"]`, so bash does run. That behaviour predates this PR (FR-3166) and this PR does not change it. I am leaving it for a person to confirm rather than dismissing it. |

### Not applied (2)

| # | File:line | Copilot says | Reply |
|---|-----------|--------------|-------|
| 4 | `ServiceConfigurationFormItems.tsx:177` | With `align="start"` on a column `BAIFlex`, the long hint shrink-wraps and overflows on narrow screens. | In a column flex, a non-stretched item gets a `fit-content` width, which is capped by the container. The text wraps at the container edge and does not overflow. `align="start"` is there because `BAIFlex` defaults to `center`, which would center-align the hint lines. `react/src` already uses `direction="column" align="start"` in 26 other places. |
| 5 | `AdminDeploymentPresetSettingPageContent.tsx:1184` | Same `align="start"` concern. | Same answer as #4. |

This PR stays a draft until #1–#3 are decided.

nowgnuesLee:
### Response to the Copilot review (5 suppressed comments)

No code changes. Each finding is answered below.

| # | Where | Finding | Outcome |
|---|---|---|---|
| 1 | `AdminDeploymentPresetSettingPageContent.tsx:1184` | `align="start"` stops the hint from wrapping | **Won't apply.** `BAIFlex` maps `align` to `alignItems`. A `flex-start` item in a column container sizes to `fit-content`, and `fit-content` never exceeds the container's width, so long localized text still wraps. |
| 2 | `AdminDeploymentPresetSettingPageContent.tsx:1187` | The batch-mode Startup Command in the session launcher has no hint | **Out of scope, kept on purpose.** See the PR body's Scope note: this PR covers the model service start command. |
| 3 | `ServiceConfigurationFormItems.tsx:177` | Same as #1 | **Won't apply.** Same reason as #1. |
| 4 | `ServiceConfigurationFormItems.tsx:185` | Single-quoting fails when a JSON value contains `'` | **Accepted limitation, no copy change.** Confirmed: `'{"prompt": "don't stop"}'` splits in the tokenizer and fails in `shlex`/`bash`. `'\''` or double quotes with `\"` escapes work on all three paths. The motivating case (`--override-generation-config` with numeric sampling values) has no single quotes, so the hint stays short instead of covering the rare case. |
| 5 | `ServiceConfigurationFormItems.tsx:175` | The legacy branch shows shell syntax although `startCommand` runs without a shell | **Won't apply.** `StartCommandHelperShell` reads `Shell syntax: /bin/bash -c "cmd1; cmd2"`. It tells the user to invoke bash explicitly, which tokenizes to `['/bin/bash', '-c', 'cmd1; cmd2']` and does run through a shell. This hint was added in FR-3166 and this PR does not change it. |

yomybaby (approved):
LGTM

PR #9665, opened 2026-09-14
fix(FR-3927): verify a vfolder download reached the browser before reporting success

### LINKED ISSUE #9652: Folder explorer: file download does nothing, for both a single row and a multi-row selection
Reported by Jongeun Lee while reviewing the FR-1564 dev server (PR [#9605](https://github.com/lablup/backend.ai-webui/pull/9605)).

## Symptom

Downloading from the folder explorer does nothing — no file is saved. Both download paths are affected:

- the per-row **Download** action on a single file, and
- the toolbar archive download for a multi-row selection.

## Reproduce

1. Open the reporter's [test folder](https://fr-1564-pr9605-decide-overwrite-per.sungchul.10-82-0-159.sslip.io/project/a%ED%95%9C%EA%B5%AD%EC%96%B4%EA%B0%80%EB%8A%A5_cde/data?folder=6ac20b2c-3237-43eb-b06c-a5c3841a0d07).
1. Click the download icon on one file row — nothing is downloaded.
1. Select several rows and use the toolbar download — nothing is downloaded.

## Where

- Per-row: `FileNameCell.tsx` → `downloadFileMutation` → `baiClient.vfolder.request_download_token()` → `initiateDownload()`
- Multi-select: `ExplorerActionControls.tsx` → `downloadArchiveMutation` → `baiClient.vfolder.request_download_archive()` → `initiateDownload()`
- `packages/backend.ai-ui/src/helper/index.ts` → `initiateDownload()` (creates a hidden `<a download>` and clicks it)

## Investigation pointers

- `initiateDownload()` resolves as soon as the anchor is clicked, so the "Download started" toast fires whether or not the browser ever begins a transfer. The toast is not evidence the download worked — check the Network tab and the storage proxy response instead.
- The `download` attribute is ignored for cross-origin URLs, and the storage-proxy download URL is a different origin from the WebUI. Verify the proxy URL returned by `request_download_token` / `request_download_archive` is reachable from the browser, and what status/headers (`Content-Disposition`, CORS) it returns.
- Confirm whether the same steps fail on `main` and against a non-dev deployment, to separate a code regression from an environment/proxy-reachability problem on the dev host.

## Notes

- Not caused by PR #9605 — that PR changes no download code path.

JIRA Issue: FR-3927

ironAiken2:
Resolves #9652 (FR-3927)

## Summary

The folder explorer's two download paths told the user "Download started" whether or not the browser ever began a transfer — `initiateDownload` handed the storage-proxy URL to a hidden `<a download>` and resolved as soon as the anchor was clicked. A proxy origin the browser cannot reach was therefore indistinguishable from a working download, which is exactly how the reported symptom ("nothing happens") arises.

This PR makes the download path verify itself before it reports success.

## What the investigation found

Against the FR-1564 dev deployment, checked live from a real Chromium:

| Check | Result |
|---|---|
| `POST /func/folders/<id>/request-download` | `200`, returns `http://10.82.0.130:6021/download` |
| `GET` on that URL | `200`, `Content-Disposition: attachment; filename="notes.txt"` |
| `POST .../request-download-archive` → `GET /download-archive` | `200`, `application/zip`, correct `Content-Disposition` |
| Per-row download, page over plain HTTP | file saved |

So **the storage proxy and the download code were both working** — there is no code regression in the download logic on `main`. Two things can still leave the browser silent. The issue's repro link is `https://` and the dev gateway does answer TLS, while the manager and the storage proxy at `10.82.0.130` speak plain HTTP only — so from an `https:` page the browser may refuse the `http://…:6021/download` URL as mixed content, depending on its security policy. The other is the storage-proxy origin simply not being reachable from the reporter's machine. In both cases the UI had no way to say so.

That silence is the defect this PR fixes. The WebUI cannot route around an unreachable proxy, but it must not claim success for it.

## Changes

- **`initiateDownload` probes the URL before clicking the anchor** (`packages/backend.ai-ui/src/helper/index.ts`). The storage proxy has allowed cross-origin GETs on `/download` and `/download-archive` since 22.03.0, and its download token is a stateless JWT valid until `exp`, so the probe costs one request whose body is cancelled at the status line and the token still works for the real transfer. (`HEAD` is not an option — the proxy answers `405`.) The browser already XHRs the same origin for uploads, so no deployment can allow uploads and block this.
- **A failure is now an error, not a toast that lies.** `DownloadFailedError` carries `reason` (`unreachable` / `rejected` / `popup-blocked`), the proxy `origin` and the HTTP `status`; both mutations render it through one shared `useDownloadErrorMessage`, so the per-row and the multi-select paths read identically. Three new i18n keys across all 21 BUI locales.
- **No client-side guess about what the browser will allow.** A pre-check that rejected an `http:` download URL on an `https:` page was tried and dropped: whether such a request goes through is the browser's decision (its mixed-content and Local Network Access policies), not the client's. A download the browser does block surfaces as `unreachable`, naming the proxy origin.
- **`window.open` returning `null` on iOS Safari no longer hangs.** The old promise never settled in that branch (`newWindow && resolve()`); it now reports `popup-blocked`. The flag is real — `manifest/app.js` sets `globalThis.iOSSafari`.
- **Two smaller defects in the same path.** The per-row toast and the saved filename used the `currentPath`-prefixed path (`"./notes.txt"`) where the bare name was meant; the toolbar archive button, unlike the Delete button beside it, carried no `aria-label`, leaving an icon-only button with no accessible name.
- **The probe gives up after 15 seconds.** A proxy that never answers (firewalled, or the wrong port) used to hold the button's pending state for the browser's own connect timeout; the probe now aborts at `DOWNLOAD_PROBE_TIMEOUT_MS` and reports `unreachable`.
- **A failed download no longer closes the folder explorer.** Both download actions awaited `mutateAsync` inside Astryx `Button`'s `clickAction`, which runs in a React transition without a catch — so the rejection `onError` had already turned into a toast went on to the page's error boundary and unmounted the modal. The actions now swallow the rejection; the toast is the report.
- **Tests** for all four `initiateDownload` outcomes and the 15-second give-up in `packages/backend.ai-ui/src/helper/index.test.ts`.

## Verification

- `pnpm --filter backend.ai-ui test` → 82 files passed (1 skipped), 2090 tests passed, 1 skipped
- Live, driven through the real UI against the dev deployment:
  - per-row download → `notes.txt` saved, toast `File "notes.txt" download has started.`
  - multi-select archive → `vfolder-fr1564-probe-…-20260914T091821Z.zip` saved
  - storage-proxy origin blocked at the network layer → **no download**, and the toast now reads `Could not reach the storage proxy at http://10.82.0.130:6021. The download did not start.`
  - storage-proxy request held open (never answered) → the same toast after 16.9 s (15 s probe + token request), explorer modal still open
  - storage-proxy connection refused → the same toast after 0.9 s, explorer modal still open, no error-boundary fallback

## Reviewer checklist

- [ ] 1. Download a single file from the folder explorer and confirm the file is saved and the toast names the bare filename.
- [ ] 2. Select several files, use the toolbar download, and confirm a `.zip` is saved.
- [ ] 3. Block the storage-proxy origin in devtools, download again, and confirm an error naming that origin appears instead of "Download started".
- [ ] 4. Point the storage proxy at a port that drops packets, download, and confirm the error toast appears within about 15 seconds.
- [ ] 5. Trigger any download error and confirm the folder explorer modal stays open.


ironAiken2:
Dev server on box `help`: http://fr-3927-pr9665-verify-vfolder-download.help.10-82-0-159.sslip.io

ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3927-pr9665-verify-vfolder-download.sungchul.10-82-0-159.sslip.io)

packages/backend.ai-ui/src/helper/index.ts
```diff
@@ -491,36 +491,105 @@ export const useSemanticColorMap = (): Record<SemanticColor, string> => {
...
+ * valid until `exp` — so this costs one extra request whose body is dropped as
+ * soon as the status line arrives, and the token still works for the real
+ * transfer. It does not support `HEAD` (405).
+ */
+const verifyDownloadURL = async (downloadURL: string): Promise<void> => {
+  let response: Response;
+  try {
+    response = await fetch(downloadURL);
+  } catch (error) {
+    throw new DownloadFailedError(downloadURL, 'unreachable', {
```
agatha197, line 551: A mixed-content block lands here too: on an HTTPS page, `fetch` of an `http://` proxy URL is refused by the browser with the same `TypeError`, so the user sees "Could not reach the storage proxy at http://…" and goes to check a proxy that is fine. Please split that case out — when `location.protocol === 'https:'` and the download origin is `http:`, throw a distinct reason (e.g. `insecure`) with a message that says the browser blocked an HTTP download from an HTTPS page. It also bears on the root cause: the issue's repro link is `https://` and that gateway does answer TLS, so the reporter may well have hit exactly this.

agatha197 (changes requested):
Checked the probe's premises against the storage-proxy source (CORS on /download and /download-archive, GET-only routing so HEAD is 405, stateless JWT, zipstream archives) and they hold; tests and locales look good. One change requested below.

packages/backend.ai-ui/src/helper/index.ts
ironAiken2: Done in 226cf114d. `initiateDownload` now compares the page and download protocols before probing: an `http:` URL from an `https:` page rejects with the new reason `insecure`, no fetch is attempted, and the toast reads "The browser blocked the download because this page is served over HTTPS but the storage proxy at {{origin}} uses plain HTTP." (new key `comp:FileExplorer.DownloadInsecure` in all 21 locales). Tests cover the blocked case and the `https:` page → `https:` proxy path that must still probe.

On the root cause: you are right that the gateway answers TLS, so I dropped the "transcription slip" claim from the PR body. The manager and the storage proxy at 10.82.0.130 speak HTTP only, so I could not reproduce the blocked path live from an `https:` page without also disabling the block itself; the unit test is the verification for that branch.

ironAiken2: Reverted in 305e86a7d after discussion with the PR owner: the client should not decide up front what the browser will allow. Whether an `http:` download URL goes through from an `https:` page is the browser's call (its mixed-content and Local Network Access policies), so the `insecure` pre-check, its message and the `DownloadInsecure` key are gone. A download the browser does block still fails the probe and surfaces as `unreachable` with the proxy origin. Reopening this thread so the decision is visible to you rather than buried under a resolved mark.

packages/backend.ai-ui/src/components/baiClient/FileExplorer/hooks.ts
```diff
@@ -169,6 +171,35 @@ export const useSearchVFolderFiles = (vfolder: string, fetchKey?: string) => {
   };
 };
 
+/**
+ * Turns a failed download into the message shown to the user. Both download
+ * paths — the per-row action and the toolbar archive — share it so a proxy the
+ * browser cannot reach reads the same either way.
+ */
+export const useDownloadErrorMessage = () => {
```
agatha197, line 179: `useDownloadErrorMessage` has no test, and nothing references the three new locale keys either. This hook is what turns a failure into the text the user actually reads, so the lying-toast regression this PR exists to fix could come back through it while the `initiateDownload` tests all stay green. Please add a test that each `reason` maps to its key and that a non-`DownloadFailedError` still falls back to `message` / `title`.

agatha197 (changes requested):
Agreed on dropping the mixed-content pre-check, and there is a stronger reason than the one recorded: browsers treat http://localhost, 127.0.0.1 and [::1] as potentially trustworthy and do not block them, so the pre-check would have refused a working download against a loopback storage proxy. The initiateDownload suite holds up well under mutation testing (six deliberate breaks, all caught, including the 15s abort boundary). One coverage gap left.

packages/backend.ai-ui/src/components/baiClient/FileExplorer/hooks.ts
ironAiken2: Added in 399ec0609: `useDownloadErrorMessage.test.ts` renders the hook against BUI's real i18n instance (no `react-i18next` mock), so each `reason` is asserted against the English copy the user actually reads — `unreachable` → `DownloadUnreachable` with the proxy origin, `rejected` → `DownloadRejected` with the HTTP status, `popup-blocked` → `DownloadPopupBlocked`. Because the real bundle is used, a missing locale key would come back as the raw key string and fail the assertion, which also covers the "nothing references the three keys" gap. A non-`DownloadFailedError` falls back to `message`, then `title`, and an error carrying neither returns `undefined`.

agatha197 (approved):
LGTM

PR #9680, opened 2026-09-14
docs: update vfolder docs (2026-09-14)

lablup-octodog:
> [!NOTE]
> **Updated 2026-09-15 — screenshot captured.** `vfolder_upload_overwrite_modal.png` now exists in `src/en|ko|ja|th/images/`: the **Overwrite Confirmation** dialog with one colliding row (`train.py`, **Name** / **Existing** / **Uploading** columns, checked by default) and the "1 other item(s) will be uploaded as well" line. The `TODO(screenshot)` markers are removed in all four locales; the upload was cancelled after capture, so nothing was written to the folder.

## What changed in the product

FR-1564 replaced the all-or-nothing "The file or folder with the same name already exists. Do you want to overwrite?" confirm on folder-explorer uploads with a per-item decision. `useUploadVFolderFiles` now partitions a pick into colliding and free top-level entries, and the new `OverwriteConfirmModal` renders a checkbox table where the user chooses, row by row, which existing items to overwrite. Unchecked rows are dropped from the upload entirely; entries whose names are free upload either way and are only counted.

The `vfolder` chapter never documented the old dialog at all, so this lands as a **new `### Overwrite existing files` H3** rather than an edit of existing prose. It sits at the end of the *File operations* area, immediately before *Edit text files* — one position later than the plan suggested (the plan placed it before the `:::warning` about `upload-file` permission). The writer's reason is sound: the warning stays adjacent to the **Upload** paragraph it qualifies.

## Driving commit for this chapter

- `5eb421ddb` — `feat(FR-1564): decide overwrite per duplicate item on folder upload (#9605)`

The other three commits in the range (`c7196d90f` FR-3910 #9644, `485681c6d` FR-3917 #9643, `4f5bda04c` FR-3920 #9654) drive the `admin_menu` chapter and are handled in its own PR.

- **"The Upload button is disabled only when every row is unchecked and there are no other items to upload."** True in the code (`okButtonProps.disabled: _.isEmpty(overwritingNames) && newEntryCount === 0`), but a micro-affordance note that adds no task capability — style guide "What NOT to Document" §2/§5.
- **"The decision is remembered for the directory you were in when you started the upload."** Also true (`uploadPath` is captured at request time in `requestUpload`), but rare-edge-case narration. Reviewer may reasonably disagree and ask for it back; both behaviours are verified-correct, they were judged not to be manual material.
- **`comp:FileExplorer.DuplicatedFilesDesc`** — still in every BUI locale file but no longer rendered after FR-1564. Dead translation data; the manual never quoted it, so nothing to remove here.

## Review findings

Verified independently against the source, not against the writer's summary.

**Claims checked against code — all supported:**

| Claim in the new section | Evidence |
|---|---|
| Dialog opens before any transfer starts | `requestUpload` calls `onUpload` only via `resolvePendingUpload`; a collision sets `pendingUpload` instead |
| Both the **Upload** button and drag-and-drop reach it | `BAIFileExplorer.tsx` wires `onUpload={requestUpload}` on *both* `DragAndDrop` and `ExplorerActionControls` |
| A folder is matched by its top-level folder name only | `resolveUploadEntry` takes `relativePath.slice(0, separatorIndex)` — only the first path segment can collide |
| One row per conflicting item; **Name** / **Existing** / **Uploading** columns with size + modification time | `OverwriteConfirmModal` columns + `ItemInfo` (`convertToDecimalUnit` + `dayjs(...).format('lll')`) |
| Every row selected by default | `useState(() => _.map(duplicatedEntries, 'name'))` |
| Clearing a row keeps the existing item and skips that upload | `resolvePendingUpload` `_.reject`s the skipped names from `fileList` |
| Non-colliding items upload regardless, counted below the table | `newEntryCount` from `_.partition`, rendered as `OtherItemsWillBeUploaded` under the `BAITable` |

No unsupported or invented behaviour. In particular the section correctly does **not** claim a rename/keep-both option or a per-file choice inside a folder — neither exists.

**Translation parity:** complete. All four files get the same H3 in the same structural position (after *File operations*, before *Edit text files*), with identical heading counts and ordering across `en`/`ko`/`ja`/`th`. Every UI label in the prose matches the live BUI i18n store shipped in the same commit — `comp:FileExplorer.DuplicatedFiles` (Overwrite Confirmation / 덮어쓰기 확인 / 確認の上書き / การยืนยันการเขียนทับ), `Name`, `ExistingItem` (기존 항목 / 既存 / ที่มีอยู่), `UploadingItem` (업로드할 항목 / アップロード / ที่จะอัปโหลด) and `general.button.Upload` (업로드 / アップロード / อัพโหลด). No terminology drift: nothing in the new prose collides with `terminology.json` (`storage folder`, `vfolder`, `Data`).

**Screenshots:** none present, none recaptured — see the admonition above. No stale image elsewhere in the chapter is invalidated by this change.

**Scope creep:** none. The diff is four files, +44/−0, all `packages/backend.ai-webui-docs/src/*/vfolder/vfolder.md`.

**Minor nits (non-blocking, reviewer's call):**

- The new text says "with the **Upload** button", but the actual affordances are a dropdown with **Upload Files** and **Upload Folder** (`comp:FileExplorer.UploadFiles` / `UploadFolder`; the folder one drives a `webkitdirectory` input). This matches the chapter's existing wording one paragraph earlier, so it is consistent rather than wrong — but the folder-vs-file distinction is what makes the "matched by top-level folder name" sentence meaningful.
- The top-level-name grouping is verified for the `webkitdirectory` picker, which populates `webkitRelativePath`. For a folder *dropped* onto the overlay, that attribute is browser-dependent and neither the diff nor `e2e/vfolder/file-upload-duplicate.spec.ts` (file-based cases only) exercises it. The sentence is not wrong for the documented path; just note it is unverified for dropped directories.

**Verified by the wrapper:** the diff touches only `packages/backend.ai-webui-docs/`, and `scripts/verify.sh` passes.

## Convention note

This PR carries no `FR-` number **by design** — it is produced by the nightly documentation sync, which follows a commit range on `main` rather than a Jira work item. It therefore deviates from the repo's `prefix(FR-XXXX): title` PR-title convention and will not be picked up or linked by the `project-status-sync` workflow. No `Resolves #NNNN (FR-NNNN)` line is present for the same reason.

---


ironAiken2:
As the overwrite function is a secondary feature supporting the upload function rather than a core feature, there is no need to describe it in detail. I will close this PR without merging it.

PR #9724, opened 2026-09-16
fix(FR-3969): show the picked mount permission at once and load until the refetch lands

### LINKED ISSUE #9723: Mount permission selector in folder details keeps the old value until the refetch lands
## Problem

In the Folder Explorer modal details panel (`VFolderNodeDescriptionV2`), changing the **Mount Permission** `Selector` does not reflect the choice immediately. The trigger keeps showing the previous value (e.g. **Read only**) while the REST update (`vfolder.update_folder`) and the follow-up `vfolderV2` refetch are in flight, and only flips once the Relay store is updated by the refetch. There is no loading indicator on the control during that window.

## Expected

- The selector shows the newly chosen value as soon as the user picks it.
- The selector shows its loading state (`isLoading`) until the update and the store refresh complete.
- On failure the selector reverts to the server value and the error message is shown (existing behaviour).

## Where

- `react/src/components/VFolderNodeDescriptionV2.tsx` — the `permission` item of the metadata list, `onChange` → `updateMutation.mutate` → `fetchQuery` refetch.

## Notes

Reported from the PR #9665 review server: Data page › Folder Explorer › "읽기 및 쓰기" selector (bai-review `c_7vraofz`).

JIRA Issue: FR-3969

yomybaby:
Resolves #9723 (FR-3969)

## Summary

The **Mount Permission** `Selector` in the folder details panel (`VFolderNodeDescriptionV2`) kept showing the server value after the user picked a new one. The pick only became visible once the REST update finished and the follow-up `vfolderV2` refetch updated the Relay store, with no loading indication in between, so the change looked ignored.

- Keep the picked value in local state and render it in the `Selector` right away.
- Show the `Selector`'s `isLoading` state until the REST update **and** the store refresh settle.
- Drop the pending value on error so the control falls back to the server value (the error toast is unchanged).
- Ignore a re-pick while one is in flight, and a pick equal to the current value.

## Verification

`bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, lint, format, tsc, search index; token gate reports only the two pre-existing `--bai-dialog-z` findings).

## How to check

Data page → open a folder → details panel → change **Mount Permission**. The selector should switch to the picked value immediately with a spinner, then settle once the update lands.


yomybaby:
Dev server on box `jongeun`: http://fr-3969-pr9724-show-picked-mount.jongeun.10-82-0-159.sslip.io

PR #9739, opened 2026-09-17
refactor(FR-3980): gate GraphQL compat by manager version in components

### LINKED ISSUE #9738: Gate GraphQL compatibility by manager version in components and drop pre-LTS compat code
## Background

GraphQL backward-compatibility checks are currently registered in `baiClient` as named feature flags (e.g. `client-ip`), checked via `supports('...')`. This does not scale: every new query, or even a single new field, needs its own entry in `baiClient`. That makes things more confusing, not less. A flag name alone also hides the minimum manager version, so anyone reading the component has to search again to learn which manager version it needs.

## Decision

- For GraphQL, the component that uses a field or query checks the manager version itself.
- Put the version in the component so it can be read at a glance, e.g.:

```
const supportsClientIp = isManagerVersionCompatibleWith('26.4.x');
```

- Do not add new GraphQL-only feature flags to `baiClient`.

## Scope

1. Remove the backward-compatibility code for managers older than LTS (26.4) from `baiClient` and the React side (`@since` gates, `supports()` flags, and fallback paths for older managers).
1. For GraphQL checks that are still needed (newer than 26.4), move them from `baiClient` feature flags to `isManagerVersionCompatibleWith(...)` checks in the component that uses them.
1. Remove `baiClient` feature-flag entries that no longer have callers.

## Timeline

Open the PR now. Review and merge after the upcoming release.

—

JIRA Issue: FR-3980

nowgnuesLee:
Resolves #9738 (FR-3980)

## Summary

GraphQL backward compatibility is now gated by the manager version in the component that uses the field, and the compatibility code for managers older than the 26.4 LTS is removed.

- `baiClient._updateSupportList()` is empty. Every named flag is gone from `packages/backend.ai-client/src/client.ts`; `supports()` stays as API surface but nothing in the app calls it.
- Flags introduced at or before manager 26.4.0 are treated as always on. Their call sites lose the fallback branch (for example the legacy mounted-folder links, the single-agent launcher mode, `ForceTOTPChecker`, the pre-25.6 IO charts, the legacy session status detail modal path, the `MyKeypairInfoModalLegacy` modal, and route/menu gates for reservoir, scheduler and RBAC).
- Flags introduced after 26.4.0 (26.4.2 through 26.9.0, including the `26.4.4rc7` / `26.4.4rc9` pins) became inline checks such as `baiClient.isManagerVersionCompatibleWith('26.9.0')`, with the fallback kept.
- `@since(version: "X")` directives with X ≤ 26.4.0 are stripped from the graphql tags (75 sites); newer `@since` gates stay because they already name the version at the field. The `@deprecatedSince("24.12.0")` image `name` fields and their `namespace ?? name` fallbacks are removed, as is the `Query` → `Queries` fragment shim for managers older than 25.14.0 in `graphql-transformer.ts`.
- REST gates in `backend.ai-client` `vfolder.ts` (`vfolder-id-based`, `background-file-delete`) now always use the id-based / async paths.
- The global search palette context exposes `isManagerVersionCompatibleWith` instead of `supports`.
- E2E: `skipUnlessClientFeature` is replaced by `skipUnlessManagerVersion`, flag overrides now patch `isManagerVersionCompatibleWith`, and specs that only exercised removed pre-LTS fallbacks are deleted.

## Notes for review

- 26.4.x patch-level gates (26.4.2 / 26.4.3 / 26.4.4 / rc pins) were treated as newer than the LTS baseline 26.4.0 and kept as version checks, which is the conservative reading of "older than LTS (26.4)".
- Left out of scope: the API v3/v4 field shims in `backend.ai-client` resources (REST, not feature flags), the now-dead `force2FA` config plumbing, and i18n keys that may now be unused (`session.launcher.DescSelectAgentBefore2516`, `statistics.IOReadDesc`, `statistics.IOWriteDesc`, `autoScalingRule.MetricSourceInferenceFramework`). They can follow in a cleanup PR.
- `baiClient.supports()` now always returns `false`; an external plugin that still calls it silently loses the gated behaviour. `scripts/release-risk-report.mjs` still scans for `_features[...]` additions and will report none; both belong in the follow-up cleanup.
- There is no minimum-manager-version guard at login, so a pre-26.4 manager now receives queries it cannot answer instead of the removed fallbacks.
- Per the issue timeline, review and merge after the upcoming release.

## Verification

`VERIFY_TESTS=1 bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Search index, Lint, Format, TypeScript, Vitest react / backend.ai-ui / agent-cli / root). The implementation review and the security/performance review found no CRITICAL or HIGH issues; stale `supports()` / `@since` references they found in a skill example and a query comment were removed. E2E specs were updated but not run (no backend cluster in this environment).

🔗 **Dev server**: http://fr-3980-pr9739.seungwon.10-82-0-159.sslip.io


nowgnuesLee:
### Ready gate: not passed (Copilot did not review)

- Copilot was requested on head `a2d7803` but returned: *"unable to review this pull request because the user who requested the review has reached their quota limit."*
- No review threads exist, so nothing was fixed, resolved, or escalated.
- `bash scripts/verify.sh` → `=== ALL PASS ===` on the same head.
- Dev server URL added to the PR body. No Teams thread was recorded, so nothing was posted there.

This PR stays a draft. Request Copilot again once the quota resets, or decide to skip the bot review.

PR #9455, opened 2026-09-07
feat(FR-3864): mount folders in the session launcher with BAIVFolderMountConfigInput

nowgnuesLee:
Part of #8532 (FR-3442). Linked Jira: [FR-3864](https://lablup.atlassian.net/browse/FR-3864), a sub-task of FR-3442.

> Jira Subtask issues are not cloned to GitHub, so there is no `Resolves #N` line; the parent task FR-3442 (#8532) is not closed by this PR.

## What

The session launcher's **Data & Storage** step now uses `BAIVFolderMountConfigInput` instead of the host-side `VFolderTableFormItem` / `VFolderTable`, and the picked mount **subpath** is sent to the backend.

- **New form field `vfolderMounts: VFolderMountConfigValue[]`** replaces `mount_ids` / `mount_id_map` / `vfoldersNameMap` / `autoMountedFolderNames`. The deprecated `mounts` is still cleared with a warning by `SessionLauncherFormIncompatibleValueChecker` (now reading the raw store through `Form.useWatch` with `preserve`).
- **`SessionLauncherStorageStep`** (new host component) passes the mount gates into the BUI component: `mountableHosts` from the existing `useMergedAllowedStorageHostPermission`, `autoMountedFolderNames` from a small host hook `useSuspendedAutoMountedFolderNames` (extracted from `VFolderMountFormItem`, which now calls it too), the admin owner-email override, and a ready-only filter. The form rule is BUI's `useVFolderMountConfigFormRule` (alias format / overlap / overlap with an auto-mounted folder / subpath). No `Suspense` of its own: the route-level boundary around the launcher catches the folder-list fetch, so nothing can be submitted before the list has loaded and stale entries were pruned.
- **Create a folder from the step**: the step hosts `FolderCreateModalV2` (opened by the mount config input's create button), then refetches the folder list through the component's `ref.refetch()` and appends the new folder to `vfolderMounts`, as `VFolderTable` did.
- **Payload** (`useStartSession`): `toMountCreationConfig` builds `mount_ids` (dashed UUIDs), `mount_id_map` (resolved container paths), and `mount_options[id].subpath` only for entries with a picked subpath. The manager accepts `mount_options` since 26.4.4 (`_mount_entries_from_creation_config`). The pre-25.12 name-keyed `mounts` / `mount_map` branch is removed per the LTS 26.4 baseline. `mount_id_map` / `mount_options` added to `packages/backend.ai-client` types.
[...]

PR #9745, opened 2026-09-17
style(FR-3984): polish the design of BAIVFolderMountConfigInput

### LINKED ISSUE #9743: Polish the design of BAIVFolderMountConfigInput
Follow-up to FR-3864 (the session launcher now mounts folders with `BAIVFolderMountConfigInput`). The component works but its design is unfinished.

## Scope

- Show a folder identicon (same generator as `VFolderNodeIdenticon`) before the folder name in the folder select options and in the mount rows. Move the identicon into `backend.ai-ui` if needed.
- The folder select popup only shows the name and the storage host. Lay out more identifying information in a tidy row: permission (RO/RW), usage mode, ownership type (user/project), host.
- Redesign the per-folder mount row. A mount is "source (folder + subpath) mounted at a container path (alias)", but the row currently puts the alias first and the subpath second, so the direction is hard to read. Put the source first, the destination second, and make the direction visible.
- The path alias and subpath inputs must each have a clear button.

## Out of scope

- Mount payload and validation rules (unchanged).

JIRA Issue: FR-3984

yomybaby:
Resolves #9743 (FR-3984)

Stacked on #9455 (FR-3864).

## Summary

Design polish for `BAIVFolderMountConfigInput`. The emitted value, the validation rules and the mount payload are unchanged.

- **Folder identicon.** New `BAIVFolderIdenticon` in `backend.ai-ui` (same dicebear `shapes` generator, seeded with the `VirtualFolderNode` global id so the glyph matches the folder lists). It is drawn before the folder name in the select options and in every mount row. The host `VFolderNodeIdenticon` / `VFolderNodeIdenticonV2` now render it, so `@dicebear/core` / `@dicebear/shapes` become BUI dependencies (they were already app dependencies).
- **Folder select popup.** Each option has a leading checkbox and the identicon as a two-line-tall avatar centered with it, the name, a second line with the owner (the user's email, or the project name), usage mode, host and creation date, and the R/W/D permission as welded Astryx `Token`s in the data page's colours at the trailing edge. `GET /folders` leaves `user_email` and `group_name` empty, so the owner falls back to `creator` and to a new `currentProjectName` prop the host passes. `BAIComplexSelect` gains `selectionMark` (`'check'` default, `'checkbox'` at the row's start) and an option `icon` slot (handed to `SelectorOption`); its option rows now stretch, so `extra` lands at the trailing edge — no other select passes `extra` today.
- **Mount rows read source → destination.** A mount is "this folder's subpath, mounted at this container path", but the row showed the alias first and the subpath second. The row is now `[identicon folder /][subpath]` → `[mount path]`: the folder name is the fixed head of the source path and the subpath picker its editable tail, sharing one outline, with an arrow to the mount path input. The columns are labelled "Path in folder (source)" and "Container path (destination)". Rows sit `--spacing-3` apart. The `/` separator takes the field-icon tone and the identicon in the folder head is 17px. Both controls use the default (md) height; only the folder name shrinks in the folder head, so the `/` stays visible for long names. In a narrow container (≤ 520px, container query) each mount stacks with the arrow pointing down.
- **Mount path is a plain input.** No fixed-looking `/home/work/` prefix: the input starts empty and its placeholder shows the resolved default `${aliasBasePath}${name}`; the resolved path appears under the input only when a relative alias is typed. `ResolvedVFolderMount.isDefaultAlias` compares the resolved path with the default rather than testing for an empty input, so typing the default path by hand still greys it in the preview.
- **Card.** The header and rows sit in a small `BAICard`, separating the mount configuration from the folder select above it.
- **Launcher preview.** The mount table uses the same two columns: "Path in folder" renders identicon + folder name (linked to the folder explorer), then `/` + subpath as text only when a subpath is set; "Container path" shows the resolved path. The preview titles carry no "(source)" / "(destination)" suffix. The separate Subpath column and `session.launcher.FolderSubpath` go away; `MountSourcePath` / `MountDestinationPath` are added to the 21 host locales.
- **Clear buttons.** The alias `TextInput` uses `hasClear`. `BAIVFolderPathPicker` gains `allowClear` and `placeholder`; `BAIComplexSelect` gains `allowClear` (emits `[]` / `null`), used by the folder select. Both ride on a new `hasClear` / `onClear` on Astryx `ComplexSelector`, added by **`react/patches/@astryxdesign__core@0.5.4.patch`**: `ComplexSelector` has no clear affordance (still none in 0.6.2, while `Selector` has one), and a consumer-side overlay collides with the loading `Spinner` that `ComplexSelector` draws in the same slot. The patch renders the clear button between the spinner and the chevron exactly as `Selector` does, reusing `@astryx.selector.clearLabel`, and stops the click before the container's popover-open handler. Upstream: facebook/astryx#6361, PR facebook/astryx#6362. In the mount row an unset subpath reads "Select a path", and clearing returns to it.
- i18n: `SourcePath`, `DestinationPath`, `OwnerUser`, `OwnerProject` (21 BUI locales); `Name` and `PathAndAlias` are removed.

Three row layouts were prototyped (single-line columns; fused path inputs; one card per folder); the fused-path one was chosen in review, without a fixed `/home/work/` prefix on the destination and without permission / host in the row.

## Verification

- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → no new findings
- BUI vitest (`fragments`, `baiClient`, `BAIComplexSelect`) → 9 files, 80 tests passed, 1 skipped (4 new for `isDefaultAlias`)
- Patched `ComplexSelector` checked in Storybook: trigger row renders `[button][spinner][clear][chevron]`, clearing the subpath and the whole folder selection works, no console errors
- Live launcher (admin + user accounts, two projects) walked through 10 scenarios by an independent agent; findings outside this PR are reported separately (project-type folders are missing from the unscoped `GET /folders` list the base stack relies on)
- Storybook `Fragments/BAIVFolderMountConfigInput` checked in Chromium: prefilled rows, overlapping / invalid alias errors, the option popup (checkbox, avatar and row vertically centered on the same line), the stacked layout at 380px, a truncated folder name (name ellipsized, `/` kept), focus and disabled states of the fused field
- Pre-ready review: Copilot credits were exhausted, so an independent reviewer pass read the final diff. Fixed from it: an `aria-hidden` header holding focusable tooltip buttons, the `allowClear` guard keyed on the wrong prop, `allowClear` changing the select's width contract, the fused head not following focus / disabled, a diverging `R` rule in the permission tag, missing `isDefaultAlias` tests, stale doc text.


yomybaby:
Dev server on box `jongeun`: http://fr-3984-pr9745-polish-design-baivfoldermountconfig.jongeun.10-82-0-159.sslip.io

agatha197 (approved):
LGTM

PR #9752, opened 2026-09-18
fix(FR-3562): disable API sign-in where the endpoint is a webserver

### LINKED ISSUE #8819: Do not offer API sign-in mode where it cannot work (webserver-served deployments)
When the WebUI is served by a **webserver**, `allowChangeSigninMode = true` still offers **API** on the login screen — but API mode cannot work against a webserver address. The user picks a mode that is guaranteed to fail.

## Why API mode cannot work through the webserver

Verified against the backend.ai source:

1. `/func/*` is session-only. **`get_api_session` (`src/ai/backend/web/auth.py`) raises 401 unless the request carries an authenticated web session holding a `keypair` token, then signs upstream with** that session's\*\* keypair — the caller's own keypair is never used. Confirmed empirically: a valid session cookie plus a deliberately garbage `Authorization` header still succeeds.
1. **Signatures are host-bound and the proxy does not rewrite them.** The manager will verify against the original URL when `X-Forwarded-URL` is present (`src/ai/backend/manager/api/rest/middleware/auth.py`), but **nothing in backend.ai sets that header** — the manager only reads it, for an external reverse proxy. A signature computed for the webserver host cannot validate at the manager.
1. **`POST /admin/gql` is not a route on the webserver.** The last route is the SPA catch-all `GET /{path:.*$`} (`src/ai/backend/web/server.py`), so a POST returns `405` with `Allow: GET,OPTIONS`.

API mode therefore requires a directly reachable **manager** endpoint. A team member reached the same conclusion in the session-max-count thread, describing API mode as "not going through the webserver ... because the endpoint has to be 8091" — though that remark was explicitly hedged as a guess, so it corroborates the finding rather than establishing it.

## What the user sees today

The failure is already reported at a reasonable level — `connectUsingAPI` catches and shows `error.CannotConnectToServer` ("Cannot connect to the server. Please check the endpoint address and network connection."). The CORS message only appears in the devtools console, so **no CORS wording belongs in the UI**.

The defect is narrower than the message: an option is presented that cannot succeed in this deployment.

## Proposal

Do not offer a mode that cannot work. When the WebUI is served by a webserver, hide the Session/API switch (or disable the API option). If a reason is shown at all, keep it in product language — the endpoint address is what the user can act on, not the transport error.

Worth confirming first: whether any shipped deployment legitimately serves the WebUI from a webserver while expecting users to point API mode at a separate manager address. If so, the switch should stay and the fix is limited to guidance.

JIRA Issue: FR-3562

agatha197:
Resolves #8819 (FR-3562)

## Summary

When the endpoint is a Backend.AI **Webserver**, the login screen still offered **API** sign-in, which cannot work there: the webserver proxies only session-authenticated `/func/*` (re-signing upstream with the session's own keypair) and has no `POST /admin/gql` route at all. The user could pick a mode guaranteed to fail.

`react/src/helper/webServerEndpoint.ts` asks the endpoint itself: a webserver renders the WebUI's `config.toml`, a manager answers 404 for it. Nothing is inferred from the port or from the page's own origin. `LoginView` runs that probe when the endpoint settles, and on a webserver it **disables** the Session/API switch, carrying the reason as Astryx's `disabledMessage`, and derives the connection mode as `SESSION` — the mode, not just the switch, because the login path, the silent re-login and the session check all act on it. Derived rather than stored: a config refresh re-applies the configured mode at any time, so a value written once would be reverted while the switch stayed disabled. Anything the config cannot be — a 404, an unreachable host, a blocked cross-origin read, or a static host answering its SPA fallback — means "not a webserver", so API mode stays on offer rather than being removed where it works.

The switch is disabled rather than hidden because it sits **above** the endpoint field it depends on: hiding it moved the top of the form whenever the address below changed, and it moved twice per edit, since a changed address drops the previous verdict and the switch reappeared mid-typing. The reason is the whole control's `disabledMessage` because Astryx has no per-segment reason, and a disabled control swallows the hover an external tooltip would need.

The address also gets 500 ms to settle before it is asked. The field reports every keystroke, so a typed-out address was firing one request per character.

### Measured against a live pair

| probe | webserver `:8090` | manager `:8091` |
|---|---|---|
| `GET /config.toml` | 200, readable cross-origin | 404, not readable |
| `POST /admin/gql` | 405 `Allow: GET,OPTIONS` | 401 |

The 405 is the defect itself: the route API mode calls does not exist on a webserver. The manager's 404 carries no CORS headers so the fetch rejects, and that rejection is the manager answer — the failure direction is the safe one.

Live check on a dev server with `allowChangeSigninMode = true`:

| endpoint | Session/API switch |
|---|---|
| webserver `:8090` | visible, disabled, tooltip shown |
| manager `:8091` | visible, enabled |
| back to webserver `:8090` | visible, disabled again |

### Why not compare origins, and why not `isElectron`

An earlier revision compared the page origin with `apiEndpoint`. That only recognises a browser sitting on the webserver and misses the desktop app or a dev server pointed at one, which is most of the bug. `isElectron` misses a different subset: with `[server].webServerURL` set the desktop app loads the page from the webserver and still reports `isElectron === true`. Neither answers the question the endpoint can be asked directly.

### Follow-up worth a backend ticket

The webserver's own description of `[service] allow-change-signin-mode` (backend.ai `web/config/unified.py`) says it switches sign-in *methods* (password / SSO), while the WebUI has always read it as the Session/API switch (`config.toml.sample`). This PR fixes the WebUI side; aligning the backend description, or dropping the key from the webserver template where it is inert, is a backend change not made here.

## Tests

`react/src/helper/webServerEndpoint.test.ts` covers the request the probe makes, the trailing-slash case, a manager's 404, a 200 that is an SPA fallback rather than the config, a body the config section sits inside, an unreadable answer, and the endpoints it refuses to fetch at all.

### Known gap

The probe settles 500 ms after the address stops changing, so pasting a webserver address and pressing Login within that window still submits in API mode and fails the way it does today. Closing that would mean blocking the submit on a network round trip, which is a worse trade for the common path.

## Verification

`VERIFY_TESTS=1 bash scripts/verify.sh` in a clean worktree: every lane passes except `GlobalSearchPalette.test.tsx`, which times out in its `beforeAll` hook on an unmodified `main` on the same machine too (59/60 react test files pass). `react/src/helper/webServerEndpoint.test.ts` covers the request it makes, the trailing-slash case, the 404 answer, an unreadable answer, and the endpoints it refuses to fetch at all.

Internal code review (Copilot quota exhausted) ran on the earlier revision; its findings are folded into this one.

🔗 **Dev server**: http://fr-3562-pr9752-do-offer-api.sujin.10-82-0-159.sslip.io

## Walkthrough

- [Walkthrough](https://github.com/lablup/backend.ai-webui/pull/9752#issuecomment-5725225899)

Not shown in the walkthrough:

- `isWebServerEndpoint()` in `react/src/helper/webServerEndpoint.ts` and its unit tests — the probe itself has no element of its own.
- The connection mode being pinned to SESSION alongside the locked switch: on this dev server the configured mode is already SESSION, so nothing on screen changes.
- The 500 ms settle before the address is asked what it is, which turns a request per keystroke into one per typed address.
- The desktop-app case (`[server].webServerURL`), where the page comes from a webserver — it needs an Electron build, not a browser.


agatha197:
Dev server on box `sujin`: http://fr-3562-pr9752-do-offer-api.sujin.10-82-0-159.sslip.io

agatha197:
📍 **Walkthrough · 1 stops · aafbe70** — [Open the walkthrough](http://fr-3562-pr9752-do-offer-api.sujin.10-82-0-159.sslip.io/#bai=v3.c_gaqi2ow.vZPNattYFMdf5XJWLcgfsp041mIgi1kUOkxgoJvShSzdOKKOZOQbJ0MwOK0bhO3QtMS2ktpGpZomgSzMWKUuiD6Q7vE7DPfGE4e025mVpKNz_ud3vg6hAVpegTpoYFoNzWY7KWc7xf6s0UeFx-QX8tCY-5lR_Q-M_1siaSzvMebYDwBAgRpokAEFmF4BDW69xOcBAw02t56AAma1ApqqgLEDGvCPo-TrHMdzsjg_49cRjiOCwdFi0iMYvMPjE4KfYjzu4fCU8KsYgxbhNzEOIj6LklmgEGzP8Y1P7nT4ZY_w6QfeOVtG8uuISNVOiKM2fvP5XzHBSZB8jQl_H4p_OAqksd_jQYjhCe96vBumCfoeBm0cnq7U8LxF8PVRErVweCxkeccnyew23vMJTjwcepJsc-uJsCazKz6OV4CLvi-Q-NFY1JRMW9geYT8m2A0XvavFwMPOlyUAKGC8BA3wspXMvSVBMm0RNZveyKXVfDatFnPaRraUxXEkstwVKbxEkX9fSaRBJFIs-r4k45ceXkyxE0pq_4Z_vhFIv-m2XqHuvTyLfhsvzgRxN5Q96Hv8VST-8EFwq3j9b7M2zYZuG9SUmsMIB9FSSJY69sktZls8Vf5xRPj0LPk-l-OZBNgdyWnwbihfvR7_fCNhZdgolhGnPn_bW9IuPUWR345WDavv6KCBrm-XaTFrGPpaoZxXDbWwUSjn1Hx-rWzqpfXSdn59o7Cey4uNdUErFddyCjhVU2zkDwPlJz7BoL0491f9BQVsui9mMwmS-TSZBT8u0qdY9uxeiLgT0GDXMa1ti5qgwEvLFknrtLJLbUbNlOHYzHWqYvSOSUF7fgg1nYlTcalusEzdNTI7tFqjbmaflv-gboO6v9pmzbFslmZ1UKBq2RS0ogLMAS2_3lR-omA4uzXHpjarZ546Fct-ZtH9NKsf3IXniqWlQDbXfKFAw9Ili1G1xEoeArME916duinTdWqms2-nlufeFClXjlTe_lOnQn7fY9Bsvmj-Aw)
Served by `fr-3562-pr9752-do-offer-api`.

1. **로그인 › Session / API 전환 스위치** — modified segmented-control  
   로그인 화면이 적힌 접속 주소에 먼저 물어보고, 세션 로그인만 받는 주소면 이 스위치를 잠근 뒤 이유를 알려줍니다. 예전에는 주소와 상관없이 늘 고를 수 있었고, API를 골라 로그인하면 끝에 가서야 실패했습니다.  
   `늘 고를 수 있던 전환 스위치` → `잠기고 이유를 알려주는 스위치`  
   check: 지금 주소가 10.82.130.172:8090이라 스위치가 잠겨 있어야 하고, 마우스를 올리면 Manager 주소가 필요하다는 안내가 떠야 합니다. Advanced를 열어 주소 끝의 8090을 8091로 바꾸면 잠시 뒤 다시 눌리고, 8090으로 되돌리면 다시 잠깁니다.  
   code: [react/src/helper/webServerEndpoint.ts:7](https://github.com/lablup/backend.ai-webui/pull/9752/files#diff-93733e1fdb46dc70106317d1a98e90baf4c63cbd3b4516b6c644365a3574c6f0R7-R36), [react/src/components/LoginView.tsx:279](https://github.com/lablup/backend.ai-webui/pull/9752/files#diff-78871ce01cdb8854299d1b8f091b884984d947675083529d0a033c5301ab29d1R279-R302)

react/src/helper/loginConfig.ts
```diff
@@ -187,12 +182,31 @@ export function getDefaultLoginConfig(): LoginConfigState {
...
+export function isServedByWebServer(apiEndpoint: unknown): boolean {
+  const pageOrigin = globalThis.location?.origin;
+  if (typeof apiEndpoint !== 'string' || !apiEndpoint || !pageOrigin)
+    return false;
+  try {
+    return new URL(apiEndpoint).origin === pageOrigin;
+  } catch {
+    return false;
+  }
+}
```
yomybaby: Instead of this, can't we just use a variable to determine whether `globalThis.isElectron` is present?

yomybaby (changes requested):

react/src/helper/loginConfig.ts
agatha197: `isElectron` answers a different question, and it goes wrong in both directions here:

- **Electron web-shell mode *is* webserver-served.** With `[server].webServerURL` set, `electron-app/main.js:414-421` loads the page from the webserver itself, and `isElectron()` falls back to a user-agent sniff (`manifest/app.js:18-22`), so `globalThis.isElectron` is `true` while the page really is served by the webserver. Gating on `!isElectron` would leave the FR-3562 bug exactly as it is in that mode.
- **Not every browser page is webserver-served.** A static bundle — or the Vite dev server — pointed straight at a manager has `isElectron === false` and API mode works there. `!isElectron` would hide the switch on a deployment where it is valid, the same regression class as the forced-`SESSION` line dropped in 77a3bc9.

The origin test gets both right: in web-shell mode the page origin and `apiEndpoint` are both the webserver, and on a static bundle they differ.

What invited the question was my own comment, which claimed "Electron's bundle origin never matches an http one" — true only outside web-shell mode. Rewritten in 7e0ec6b to name the two cases the predicate exists to separate. Leaving this thread open for you.

agatha197: You were right to push on this, and my previous reply was wrong to defend the origin comparison. I tested it against the live pair instead of reasoning about it, and the origin check turned out to miss most of the bug.

The question that matters is whether the **endpoint** is a webserver, not where the page came from. Those differ whenever the page is served somewhere else:

| case | origin check | `!isElectron` | should be |
|---|---|---|---|
| browser on the webserver | hide | hide | hide |
| desktop app → webserver | **show** | **show** | hide |
| dev server → webserver | **show** | hide | hide |
| desktop app → manager | show | show | show |

The dev server on this PR was itself the third row: endpoint `10.82.130.172:8090`, switch still offered, sign-in still impossible. So the fix was not fixing the case I was demonstrating it on.

`ed3c8bce` drops the origin comparison and asks the endpoint directly. A webserver renders the WebUI's `config.toml`; a manager answers 404 for it:

| probe | webserver `:8090` | manager `:8091` |
|---|---|---|
| `GET /config.toml` | 200, readable cross-origin | 404, not readable |
| `POST /admin/gql` | 405 `Allow: GET,OPTIONS` | 401 |

Live on the dev server, with `allowChangeSigninMode = true`: endpoint `:8090` hides the switch, `:8091` shows it, and switching back hides it again.

Leaving this thread to you — the mechanism you questioned is gone, but it was replaced rather than changed to `isElectron`, so it is worth a look.

agatha197: Following up now that the mechanism you questioned is gone entirely.

`isElectron` answers "which app am I running in", but the thing that decides this is "what is the address I am about to sign in against". The two are independent, and the decisive counter-example sits inside the browser, measured on both:

| screen | `globalThis.isElectron` | API sign-in | switch should be |
|---|---|---|---|
| WebUI served by a webserver | `false` | impossible | locked |
| dev server, endpoint is a manager | `false` | works | operable |

Same value, opposite answers, so no expression over `isElectron` alone separates them. The Electron side splits the same way: the desktop app pointed at a webserver cannot sign in with a keypair, pointed at a manager it can, and it reports `true` either way.

That the manager side really is usable is measurable too — the call API mode makes, `POST /admin/gql`, answers 401 on a manager (route present, auth required) and 405 `Allow: GET,OPTIONS` on a webserver (no such route).

So the code no longer asks where the page came from at all. It asks the endpoint for `config.toml`, which a webserver renders and a manager 404s, and nothing in it looks at the port or the origin. That request is not new either: the app already fetches exactly it in Electron before login, so it is a path that is known to work there.

On the UX you and I landed on separately: the switch is no longer hidden, only disabled, with the reason as Astryx's own `disabledMessage`. Hiding it moved the top of the form whenever the address below changed, and it moved twice per edit.

Issue #567, opened 2020-06-09
Change session store method for console-server


inureyes:
* Current session management is based on API key, popular before 19.09. 
 * Change the session information handling to  compatible with the console server, and is improved with a session handling method without additional burden.

PR #9753, opened 2026-09-18
fix(FR-3561): keep an API-mode sign-in for the tab across page reloads

### LINKED ISSUE #8818: API mode login does not persist across a page refresh
API (access key / secret key) login does not survive a page refresh. After signing in with API mode the app works normally, but reloading the page returns to the login screen.

## Why it happens

Three independent reasons, all verified in the current code:

1. **API mode is stateless by design.** It signs every request with the keypair and never creates a server session — there is nothing on the server to restore. (Confirmed by the team's own SESSION vs API comparison in the session-max-count thread: `client.login()` not called, `check_login()` not called, server session: none.)
1. **The keypair is never persisted.** `backendaiwebui.login.api_key` / `backendaiwebui.login.secret_key` are **removed** in two places (`react/src/components/LoginView.tsx` `clearSavedLoginInfo`, `react/src/helper/loginSessionAuth.ts`) and written **nowhere** in the repository. The credentials live only in memory, so a reload drops them.
1. **There is no restore path for non-SESSION modes.** `checkLogin` (`react/src/components/LoginView.tsx:1001`) only attempts `client.check_login()` when `connectionMode === 'SESSION'`; for any other mode it returns `false` immediately.

Secondary: `connectionMode` itself is reset from config on boot (`LoginView.tsx:224`), so in a deployment with `allowChangeSigninMode = true` a refresh also drops the user back to the configured default mode.

## Reproduce

1. Set `allowChangeSigninMode = true` in `config.toml`.
1. On the login screen pick **API**, enter a reachable **manager** endpoint plus an access key / secret key, and sign in — the app loads and GraphQL calls succeed.
1. Refresh the page → back to the login screen.

## What to decide

Persisting an access key / secret key in browser storage is a security decision, not an obvious fix. Options, in the order they should be considered:

- State plainly in the UI that API-mode sign-in does not persist across reloads (cheapest, no new risk).
- Offer opt-in persistence, scoped like the existing "Remember ID" (which deliberately stores only an identifier, never a secret).
- Give API mode a real restore path.

## Notes

Unrelated to the login redesign in FR-3560 — `loginSessionAuth.ts` and `packages/backend.ai-client/` have no diff on that branch, and the `LoginView.tsx` diff does not touch `clearSavedLoginInfo` or `checkLogin`.

JIRA Issue: FR-3561

agatha197:
Resolves #8818 (FR-3561)

## Summary

API (access key / secret key) sign-in did not survive a page reload: the mode is stateless by design, the keypair lived only in memory, and there was no restore path for non-SESSION modes. The keypair is now kept **for the tab** and the reload's silent login uses it.

- New `react/src/helper/apiKeypairSession.ts`: `saveApiKeypair` / `loadApiKeypair` / `clearApiKeypair` over one `sessionStorage` entry, and `resolveInitialConnectionMode`, which starts the login screen in API mode when a keypair is kept and the Session/API switch is offered.
- `LoginView`: a successful API connect saves the keypair; `connectUsingAPI` falls back to it when the form is empty (the silent login a reload triggers); a kept keypair the manager rejects is dropped and the login panel opens instead of looping; `checkLogin` answers "signed in" for API mode when a keypair is kept, so the auto-logout branch behaves like SESSION mode; `logoutSession` clears it. The `showError` flag `connectUsingAPI` already received is now honoured, so a silent failure no longer pops the "cannot connect" toast.

### Why sessionStorage, and what CodeQL flags

The Lit shell kept the keypair in `localStorage` until #567/#568 (2020-06) removed it together with the stored session password, and nothing has persisted it since. `sessionStorage` is the middle ground: it survives a reload of the same tab, is gone when the tab closes, and `useLogout`'s `clearLoginStorage()` already calls `sessionStorage.clear()`, so sign-out drops it without a new code path.

**CodeQL reports `js/clear-text-storage-of-sensitive-data` (high) on the `setItem` call, and it is right about what the code does** — the secret key is stored in clear text. Two things decide whether that is acceptable here, and both belong in the review, not in a silent dismissal:

- **Against XSS the marginal risk is close to zero.** In API mode the client is already published as `globalThis.backendaiclient` and holds the keypair at `_config.secretKey` for the life of the page, so a same-origin script can read the secret key today, with or without this change. Both exposures last exactly as long as the tab.
- **The one genuinely new exposure is disk.** Browsers persist `sessionStorage` for crash recovery and session restore (Chrome's Session Storage LevelDB, Firefox's `recovery.jsonlz4`), so the keypair can reach the profile directory, which the in-memory client never did. An earlier revision of this description claimed nothing is written to disk; that was wrong.

Also worth weighing: a Backend.AI keypair is long-lived and is not revoked by signing out, unlike a web session, so an exfiltrated secret key outlives the incident until an admin rotates it.

Obfuscating or encrypting the value in the page would silence the alert without changing any of the above — any key the page can use, injected script can use too — so it is deliberately not done.

Not addressed here: FR-3562 (#9752) hides the API option where the webserver serves the page, so a kept keypair can only exist where API mode is actually offered.

## Verification

`VERIFY_TESTS=1 bash scripts/verify.sh` in a clean worktree: `=== ALL PASS ===`. `react/src/helper/apiKeypairSession.test.ts` covers the round-trip, malformed entries, the localStorage boundary, and the initial-mode rule.


agatha197:
Closing without merging: keeping the keypair is not worth what it costs.

- CodeQL reports `js/clear-text-storage-of-sensitive-data` (high) on the `sessionStorage` write, and it is not a false positive — the secret key really is stored in clear text. It would have been this repository's first accepted-risk dismissal of that rule; the three existing dismissals are all cases where nothing sensitive was actually stored.
- A Backend.AI keypair is long-lived and is **not** revoked by signing out, unlike a web session, so an exfiltrated secret key outlives the incident until an admin rotates it.
- Browsers persist `sessionStorage` for crash recovery and session restore, so the keypair can reach the profile directory — something the in-memory client never did.
- Persistence was removed deliberately in #567 / #568 (2020-06), together with the stored session password. This would have reversed that decision for a convenience.

Encrypting or obfuscating the value in the page would silence the alert without changing any of the above, since any key the page can use injected script can use too, so that was not pursued either.

Net effect: behaviour is unchanged — an API-mode sign-in still does not survive a reload. #9752 (FR-3562) separately stops offering API sign-in where it cannot work at all, which removes the most common way to hit this.

PR #9757, opened 2026-09-18
fix(FR-3989): keep personal projects out of the user edit project selector

### LINKED ISSUE #9755: Admin user edit modal: project selector shows a raw UUID and its item count does not match the checked projects
## Problem

Admin → Users → edit user → **Project** selector.

- The closed field displays a raw project UUID alongside project names (e.g. `01a0ae35-…, coredev, model-store`).
- The number of items shown in the field does not match the options actually checked in the dropdown: the field lists three entries while only two options (`coredev`, `model-store`) are checked.

The selected value looks wrong and the extra UUID entry cannot be matched to any option in the list.

## Expected

The field shows only the selected projects, by name, and the count matches the checked options.

See the attached screenshot.

—

JIRA Issue: FR-3989

nowgnuesLee:
Resolves #9755 (FR-3989)

## Summary

Admin → Users → edit user → **Project** showed a bare UUID next to the project names, and the number of tags in the field did not match the options checked in the dropdown.

The field is filled from `UserV2.projects` — *every* membership — while `ProjectSelect` only offers the domain's active `GENERAL` / `MODEL_STORE` projects. Manager 26.9.0 added the `PERSONAL` project type ("project holding one user's own resources, created and removed with that user"), so every user now carries one membership the selector can never list. With no matching option, the select rendered the raw value: the UUID.

The personal project is a real membership, so instead of hiding it this PR shows it by name and locks it — per the reporter: the personal project must not be removable from this form.

## Changes

- **`UserSettingModal`** reads `basicInfo { name type }` for the user's projects, keeps every membership (personal included) in the form value, and passes them to `ProjectSelect` as `fallbackProjects`. In the edit form `PERSONAL` is a locked type (creation keeps locking `MODEL_STORE` as before). `groupIds` is sent as the form holds it.
- **`ProjectSelect`**
  - `fallbackProjects`: labels for selected projects the option list cannot contain. Only entries that are **currently selected** become options, so the assignable set is unchanged.
  - Options are `disabled` by locked **type**, fallback options included; the personal project appears under a new **Personal** group label.
  - A locked personal option shows a lock icon whose tooltip reads "The personal project assigned to the user cannot be removed." (same `BAIIconWithTooltip` pattern as the project-admin badge).
  - `'use memo'`, and the no-accessible-projects error state counts fallback options too.
- **`BAISelect`** (multiple mode): a selected `disabled` option is now locked for the clear-all path as well — the clear button and Delete/Backspace keep it, and the clear button is hidden when only locked values are selected. Astryx `MultiSelector` already blocked toggling such an option but not clearing it; this also covers the `MODEL_STORE` lock on user creation.
- **i18n**: `projectSelect.Personal` and `projectSelect.PersonalProjectCannotBeRemoved` in every locale.

## Why no re-attach on submit

`UpdateUserV2Input.groupIds` is documented as replacing the membership set, which suggested the personal project had to be sent back on every save. It does not: the manager's `replace_user_projects` (26.9.0a1+, BA-7660) excludes `PERSONAL` projects from both the target and the joined set, so omitting the id never detaches the user and sending it is ignored. The earlier re-attach logic in both the single and the bulk edit (`UpdateUsersModal`) is therefore removed, and the bulk path is back to one shared input.

## Note

Follow-up worth filing: `BAIAdminUserV2Table`'s `projects` column also lists the personal project — no longer a bare UUID problem there, but one surface over from this form.

## Verification

`VERIFY_TESTS=1 bash scripts/verify.sh` → `=== ALL PASS ===`

Tests:
- `react/src/components/ProjectSelect.test.tsx` — a selected project missing from the option list is labeled, an unselected fallback is ignored, no duplication; a locked personal option is disabled and carries the tooltip, an unlocked personal or a locked model-store option does not.
- `packages/backend.ai-ui/src/components/BAISelect.lockedOption.test.tsx` — clearing keeps a locked value, no clear button when only locked values are selected, clicking a locked option does nothing.

Not yet clicked through against a manager 26.9.0+ deployment; the manager behavior above is read from the backend source.

🔗 **Dev server**: http://fr-3989-pr9757.seungwon.10-82-0-159.sslip.io


nowgnuesLee:

Ran the `pr-ready-gate` steps on this PR overnight, unattended. The gate did **not** pass, and the PR stays a draft.

| Step | Outcome |
|---|---|
| `bash scripts/verify.sh` | `=== ALL PASS ===` (14s; Astryx token gate reports only the 2 pre-existing `--bai-dialog-z` findings in `BAIDialog.css` / `BAIDrawerPortal.css`, neither touched here) |
| Copilot review — pass 1 | Refused: *"Copilot was unable to review this pull request because the user who requested the review has reached their quota limit."* |
| Copilot review — pass 2 (re-requested) | Same refusal |
| Findings triaged | none — there is no review to triage |
| Unresolved review threads | 0 (none were ever opened) |
| Dev server URL | published to the PR body below |

Two passes is the cap, so I stopped rather than retrying further. **Needs a human call:** either wait for the Copilot quota to reset and re-request (`gh pr edit 9757 --add-reviewer @copilot`), or flip to ready and let a human be the first reader.

Nothing in the branch changed during this run — the head commit is still `22cfefb`.

nowgnuesLee:
Dev server on box `seungwon`: http://fr-3989-pr9757-keep-personal-projects.seungwon.10-82-0-159.sslip.io

yomybaby (approved):
LGTM

PR #9751, opened 2026-09-18
feat(FR-3987): adopt the vfolder mount-policy model from backend #14679

agatha197:
Resolves #9749 (FR-3987)

## Summary

lablup/backend.ai#14679 (BA-7917, 26.9.0) separates the vfolder mount level from RBAC access: `vfolders.permission` becomes `default_mount_permission` (`none` / `ro` / `rw`, `wd` stored as `rw`), per-user levels live in the new `vfolder_user_mount_policies` table, and `vfolder_permissions` / `vfolder_invitations` are dropped. Two facts drive the UI changes: **every personal folder's default becomes `none`** (migration `d4a7c2e9f018`, `PersonalVFolderCreator`) while its owner mounts it read-write regardless and an invitee mounts at their own policy row; and only REST `GET /folders` `permission` is resolved for the caller — the GraphQL fields are the folder default. The legacy invitation / share endpoints the WebUI uses keep working (deprecated), so nothing breaks outright. This PR is the compatibility half; the feature half (per-user mount policy UI, exposing `none` as a settable default) is FR-3988.

- **Schema**: nothing to do any more. This branch originally hand-applied the #14679 additions (`NONE` on `VFolderMountPermission`, `vfolderMountPolicies`, `setVFolderMountPolicy` / `unsetVFolderMountPolicy` and their input / payload types) to `data/schema.graphql`; `main`'s own supergraph sync has since landed all of them, so the rebase onto `main` dropped that commit's content — `data/schema.graphql` is now `main`'s verbatim and only the Relay artifacts it regenerates remain.
- **V2 Mount Permission cell** (`VFolderPermissionCellV2`, project-admin Data page and the V2 list): `accessControl.permission` is the folder default. `NONE` on a personal folder shows **Read & Write to its owner** and `-` to anyone else (their level is their policy row, which V2 does not expose); `NONE` on a project folder shows **Not mountable** (`data.NotMountable`, 21 locales); an enum member this build does not know shows `-` instead of Read & Write. Before, `NONE` rendered as Read & Write for everyone.
- **V2 detail selector** (`VFolderNodeDescriptionV2`): `NONE` selects nothing instead of "Read only". A personal folder therefore shows no selection until FR-3988 exposes `none`; picking `ro` / `rw` sets the folder default as before.
[...]

PR #9760, opened 2026-09-18
feat(FR-3991): mint a walkthrough on demand for any PR by number

### LINKED ISSUE #9758: walkthrough skill: mint a walkthrough on demand for any PR by number, booting the dev server when none is live
## Background

The `walkthrough` skill (`.claude/skills/walkthrough/`, FR-3951) is written for the PR the session just implemented: `scripts/mint.mjs` resolves the PR and the Portless app from the **current branch** and reads that app's boot record under `~/.local/state/fw/dev-servers/`. `--app` / `--pr` exist, but they still assume a worktree on that branch and a dev server that `dev-server` has already advertised, and the stop manifest has to be authored by the session from the diff. There is no path that takes a PR number and produces a walkthrough.

Observed on PR #9751 (FR-3987): the walkthrough could only be added after the user asked for a dev server by hand, and a second session or a later date has no command to attach one to an arbitrary PR.

## Scope

- Accept a PR number (`/walkthrough <pr>` and "PR #N에 walkthrough 달아줘"). Resolve the branch, reuse or create a worktree for it, and reuse the live boot record when one exists.
- When no dev server is live for that branch, boot and advertise one through the `dev-server` skill (endpoint from the PR body, then the shell / `.env` value, as `dev-server` §2c does) instead of failing preflight.
- Read the PR's diff, write the stop manifest, mint, `upsert` the comment and `describe` the PR body — the same tail as today.
- Refuse with one line when the backend is unreachable or the app shell dies after login; never post a partial walkthrough.
- Wording rule for `ck`: state the condition that produces the new behaviour (value → label) even when the connected backend cannot produce it yet, plus one clause saying what this server cannot show. "Should look the same as before" is not a check.

## Related

- FR-3951 — the skill as landed (#9698).
- The AGENTS.md workflow change that makes dev-server + walkthrough an ordered post-PR step (linked issue).
- lablup/claude-mp: `pr-ready-gate` offering a walkthrough as its last step (linked).

JIRA Issue: FR-3991

agatha197:
Resolves #9758 (FR-3991)

## Summary

The `walkthrough` skill was written for the PR the session just implemented: `mint.mjs` took the PR and the Portless app from the **current branch** and read that app's boot record. Given only a PR number there was no path, so a PR opened by another session, or on another day, could not get a walkthrough (that is how #9751 went out without one). ADR 0004 (decision 13) already names on-demand minting as the skill's trigger; this widens it to a PR by number.

- **`mint.mjs --pr <n>`** looks the PR up in `--repo` (default `lablup/backend.ai-webui`, the constant `advertise.sh` / `comment.sh` share — never the cwd's remote), refuses a PR that is not open, and picks the **live** boot record that serves it: never stopped, pid alive, same repo, the record on the PR's own branch before a stack layer above it (whose `served[]` lists every lower layer), then the newest boot. `--app` narrows that to one record, which still has to be live and serve the PR. No app-name prediction on this path: the record is the only source, and with none it exits 3 naming the branch to boot with the `dev-server` skill.
- **The server must serve the sha.** The served commit comes from the server itself (`/__review/state.head`), else the record's worktree. Equal is fine; a commit that **contains** the PR head (the top of a stack, unpushed commits) is accepted with a note, so one server serves a whole stack; a commit that does not, or a checkout that cannot be read, is refused.
- **`--dry-run` needs no manifest**, so the on-demand path can start with the one question it needs; the two Portless probes run concurrently.
- **`scripts/resolve.mjs`** holds the record lookup and the `served[]` shape (`prFromRecord` moved there), with `node:test` coverage.
- **`SKILL.md`**: §1a, the on-demand steps (dry run → boot from a checkout at the head when nothing serves the PR, worktrees under `.claude/worktrees/` where `EnterWorktree` and `fw:cleanup-worktrees` keep them → diff → mint → post); §6 is the one statement of the resolution rule; §5 gains the rule that `ck` states the condition that produces the new behaviour (value → what shows) even when the connected backend cannot show it yet. The description no longer claims the skill never touches the dev-server comment — booting a server for a PR goes through `dev-server`, which does.

Booting stays with the `dev-server` skill (agent-driven, credentials never touch the script). Recording liveness at the source — `dev.mjs` writing `stoppedAt` on exit, one `advertise.sh serving <pr>` query for every reader — is FR-3993. FR-3992 (stacked above) makes the workflow call this.

## Verification

- `node --test …/manifest.test.mjs …/resolve.test.mjs` → 18 pass; `bash …/test-comment.sh` → 38 passed
- Dry runs from a checkout that is **not** on the PR's branch, no manifest: `--pr 9751` (live server) resolves the record, the PR head and confirms the served commit; `--pr 9653` (no server) exits 3 with the boot hint; `--pr 9698` (merged) exits 3; `--pr 9751 --app <other>` exits 3; `--pr abc` exits 2.


ironAiken2 (approved):
LGTM

PR #9761, opened 2026-09-18
docs(FR-3992): make dev-server + walkthrough an ordered post-PR step

### LINKED ISSUE #9759: AGENTS.md: make dev-server + walkthrough an ordered post-PR step of the implementation workflow
## Background

`AGENTS.md` (line 58, landed with FR-3951) says: **after `dev-server` has advertised the dev server for a PR you implemented, run the `walkthrough` skill as the last step.** Nothing in the implementation workflow tells a session to boot a dev server after opening the PR — the only other mentions of the dev server in `AGENTS.md` are the environment notes, and `pr-ready-gate` only reads the boot record. So the precondition is met only when the user asks for a dev server themselves, and the walkthrough is never added otherwise.

Observed on PR #9751 (FR-3987, 2026-09-18): the PR was opened without a walkthrough; it was minted only after the user asked for a dev server. A second cause on that day was a stale local `main` (the session started before #9698 landed, so neither the instruction nor the skill was loaded until `/reload-plugins`).

## Scope

- Rewrite the workflow bullet as an ordered post-PR step: after `gh pr create` / `gh stack submit --auto`, run the `dev-server` skill for the branch (endpoint from the PR body, then `.env`), then the `walkthrough` skill, and put the `[Walkthrough](…) · N stops` line in the final message.
- Name the two skips explicitly, each with a one-line reason in the final message: the backend is unreachable or kills the app shell; the PR has no change a person can recognize on screen (schema, tests, docs, i18n only).
- Say that a walkthrough minted against a backend that does not yet ship the feature still gets condition-based `ck` text and a "Not shown in the walkthrough" entry, rather than being skipped.
- One sentence that the skill list is fixed at session start: a session that started before a skill landed on `main` needs `/reload-plugins` (or a new session) before the workflow can call it.

## Related

- FR-3951 (#9698) — where the current bullet came from.
- The on-demand `/walkthrough <pr>` path (linked issue), which this step calls.

JIRA Issue: FR-3992

agatha197:
Resolves #9759 (FR-3992)

## Summary

`AGENTS.md` told a session to run the `walkthrough` skill *after `dev-server` has advertised the PR's server*, and nothing in the workflow booted one — so the precondition was met only when the user asked for a dev server by hand, and the walkthrough never ran on its own (#9751 went out without one; a stale local `main` at session start was the other half of that day).

The bullet now reads as an ordered post-PR step:

- right after `gh stack submit --auto` / `gh pr create`, boot the branch's dev server with `dev-server` — or reuse the live one its boot record names, one per branch — then run `walkthrough`, without waiting to be asked; endpoint and credentials follow `dev-server` §2c–§2d as written there (the earlier draft restated that rule wrongly);
- the two skips, each named in the final message with its reason: backend unreachable or app shell dies after login; nothing a person can recognize on screen — schema, tests, docs, generated files, i18n key plumbing (a relabel is on screen and gets a stop);
- a backend that does not yet ship the feature is **not** a skip — mint with condition-based checks (`walkthrough` §5) and list what that server cannot show;
- a stack submitted in one sitting gets a walkthrough per layer with on-screen changes, minted with `/walkthrough <pr>` (`walkthrough` §1a, FR-3991 below) against the top layer's server, which contains the lower heads — no second boot; the same command serves a PR from another session;
- stop the server when the PR is done (`advertise.sh stop`, `dev-server` §5) — a server nobody stopped is a ~1 GB process and a boot record every reader has to distrust (FR-3993);
- the skill list is fixed at session start, so a session that predates a skill landing on `main` needs `/reload-plugins` (observed 2026-09-18).

Docs only; `CLAUDE.md` is a symlink to `AGENTS.md`, so both read the same.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (the tree is unchanged apart from `AGENTS.md`)


ironAiken2 (approved):
LGTM

PR #9777, opened 2026-09-18
style(FR-4002): outline Token and tint Badge through the Astryx theme

### LINKED ISSUE #9776: Restyle Token (outlined) and Badge (light fill + outline) through the Astryx theme
Restyle Astryx `Token` and `Badge` app-wide through the theme.

## Scope

- `Token`: no fill, a 1px outline in the token's own colour (all 11 colours, `default` included).
- `Badge`: a light fill in the badge's colour plus a 1px outline in the same colour, dark text. The solid semantic variants (`info`, `success`, `warning`, `error`) move to the same light-fill treatment.
- Both components zero their border and set their fill through StyleX atoms that outrank `@layer astryx-theme`, so the change re-points the colour variables they read inside the component scope and draws the outline as an inset `box-shadow`.
- Mirror the overrides into `packages/backend.ai-ui/.storybook/astryxBrandTheme.ts` (KEEP IN SYNC) and rebuild the committed theme artifacts.

## Out of scope

- Per-call-site colour choices.

JIRA Issue: FR-4002

yomybaby:
Resolves #9776 (FR-4002)

## Summary

App-wide restyle of Astryx `Token` and `Badge`, done in the theme so every call site follows.

- **Token — outlined.** No fill, a 1px outline in the token's own colour (all 11 colours, `default` included).
- **Badge — tinted.** A light fill, a 1px outline and dark text, all in one hue. The palette variants were already a light fill with dark text and only gain the outline. The solid semantic variants move onto the closest palette hue: `info` → blue, `success` → teal, `warning` → yellow, `error` → red, `neutral` → gray.
- **`BAIDoubleTag` weld.** Now that badges carry an outline, neighbours overlap by the outline width so the two strokes read as one seam (what antd's bordered tags did), instead of leaving a 1px gap.
- The Storybook theme (`packages/backend.ai-ui/.storybook/astryxBrandTheme.ts`) mirrors both overrides (KEEP IN SYNC), and the committed theme artifacts are rebuilt.

## How, and why not plain `border` / `backgroundColor`

Both components zero their border with StyleX atoms, which outrank `@layer astryx-theme`, so a `border` declared in the theme silently no-ops — the outline is an inset `box-shadow` (neither component declares one). Token's fill is an atom too, so it is removed by re-pointing the variable it reads (`--color-background-<colour>`, `--color-neutral` for `default`) inside the token's scope. The semantic Badge fills are different: the base theme declares their `background-color` in the theme layer itself, so they are overridden directly, while the text colour is an atom and is reached through `--color-on-*`.

## Worth a look

- `info` badges were the brand accent (orange) and become blue. That matches the banner's info surface and antd's `processing` tag, but it is a visible hue change.
- A few palette hues have an outline token equal to their fill (`pink`), so their outline is invisible. That is the theme's token values, not this change.
- Affected surfaces: every `Badge` (`BAITag`, `BAIDoubleTag`, status tags, tables), every `Token` (closable `BAITag`, `BAIComplexSelect` badges mode, chat, Astryx multi-select chips).

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (includes the theme artifact drift check)
- `react` vitest `src/astryx-theme` → 29 passed; BUI vitest `BAITag`, `BAIDoubleTag`, `BAIBadge` → 49 passed
- Storybook `Tag/BAITag` (colour mapping) and `Tag/BAIDoubleTag` checked in Chromium: computed fill / text / outline per variant, and the welded seam


yomybaby:
Dev server on box `jongeun`: http://fr-4002-pr9777-outline-token-tint.jongeun.10-82-0-159.sslip.io

PR #9781, opened 2026-09-18
refactor(FR-4002): split BUI Tag components into Badge and Token by value semantics

yomybaby:
Resolves #9776 (FR-4002)

Stacked on #9777.

## Summary

#9777 restyled Astryx `Badge` (tinted fill) and `Token` (outline). This PR makes the two primitives mean different things and renames every chip component to match. The rule is recorded in ADR 0007.

- **Badge**: a value the system changes on its own over time. This covers lifecycle and health status, in-progress markers (Installing, Deploying, Applying), system pointers (Current, Latest), live tickers, and counts.
- **Token**: a value that changes only when a user edits it, or a category label. This covers names, types, permissions, versions, tags, on/off settings, and recorded outcomes (login result, audit status, error records).
- `StatusDot` / `BAIBadge` are unchanged.

### BUI renames (no deprecated aliases)

| Before | After |
|---|---|
| `BAITag` | deleted; call sites use Astryx `Badge` / `Token` with the `astryxTagVariant` helpers |
| `BAIDoubleTag` | `BAIDoubleToken` + `BAIDoubleBadge` |
| `BAITagList` / `BAIBadgeList` / `BooleanTag` | `BAITokenList` / `BAITokenRow` / `BAIBooleanToken` |
| `BAIDeploymentStatusTag` / `BAIAuditLogStatusTag` / `BAIArtifactStatusTag` | `…StatusBadge` |
| `BAIArtifactTypeTag` / `BAISessionTypeTag(V2)` | `…TypeToken` |
| `BAIDeploymentTagChips` | `BAIDeploymentTagTokens` (uses Token `onClick` instead of a `role="button"` Badge) |
| `PRIMARY_TAG_VARIANT` | `PRIMARY_TOKEN_COLOR` |

`BAIImageNodeSimpleTag*` keep their names, because "Tag" there is the image tag (a domain noun, see ADR 0005). Their inner chips are now Tokens.

### Host renames

| Before | After |
|---|---|
| `SessionStatusTag` / `ReplicaStatusTag` / `AgentStatusTag` | `…StatusBadge` |
| `VFolderPermissionTag` | `VFolderPermissionToken` |
| `ImageTagBadges` / `AliasedImageDoubleTags` | `ImageTagTokens` / `AliasedImageTagTokens` |
| `PortTag` | `PortToken` |
| `ValidationStatusTag` | deleted (it had no importers) |

Direct `Badge` call sites across `react/src` are reclassified by the same rule.

### Bug fix

`BAIRouteNodes` passed hex colours from `useSemanticColorMap` into `BAITag`, so every route status and health rendered gray. It now uses `badgeVariantForStatus('route')`, with a regression test.

### Docs and rules

- `docs/adr/0007-badge-for-live-values-and-token-for-settled-values.md` and its `docs/ARCHITECTURE.md` row.
- `.claude/rules/badge-vs-token.md`.
- A STATUS SEMANTICS line in the ASTRYX agent block (`AGENTS.md`, `react/AGENTS.md`). It overrides the generated "Badge = counts only" line.
- ADR 0005 and the storybook instructions now use the new names.

### Follow-ups (out of scope)

- User manual wording ("status tag", "tag chip").
- Legacy e2e locators still target `.ant-tag`; these were already broken before this PR.
- `isActive` StatusDots → `BAIBooleanToken`.
- A colour map for artifact revision status.

## Verification

```
bash scripts/verify.sh → === ALL PASS ===
```

The Astryx token gate reports only the two pre-existing `BAIDialog`/`BAIDrawerPortal` findings, so this PR adds none.

## Visual checks

- Image list Tags column: search highlighting inside Tokens.
- Selected state of `BAITabCountBadge` under the new outline.
- Agent detail status and elapsed-time double badges.
- Route table colours are no longer gray.


yomybaby:
Dev server on box `jongeun`: http://fr-4002-badge-token.jongeun.10-82-0-159.sslip.io

PR #9782, opened 2026-09-18
fix(FR-3986): find project admin roles by scope_admin permission instead of role name

### LINKED ISSUE #9748: Projects page: find project admin roles by scope_admin permission instead of role name
## Problem

The Projects page lets an admin set a project's administrators. Today it finds the role that grants project admin permission by **role name**. After the RBAC migration that lookup no longer holds, so the admin assignment needs to be updated.

## Expected

- Work out how the project admin role is represented after the migration, and branch the lookup accordingly.
- Use the `adminRoles` filter to collect **every** role in the project's scope whose permission, matched through `nestedPermissionFilter` by entity type, is `scope_admin`.
- Grant the selected users to **all** of those roles, not to a single role matched by name.

—

JIRA Issue: FR-3986

nowgnuesLee:
Resolves #9748 (FR-3986)

## Summary

The Projects page's "Set project admin" action found the role that grants project admin by **role name**. It asked `adminRoles` for the project scope's SYSTEM role pair (`project-<id>-member` / `project-<id>-admin`) and picked the one whose name ends in `admin`.

The single-scope RBAC migration (BA-7796, the change ADR 0006 and `useCurrentUserProjectRoles` already gate on) made that lookup wrong. Project admin is now a `scope_admin` **permission**, and any number of roles in the scope may carry it.

- The `adminRoles` filter asks for the `scope_admin` entity type through `RoleFilter.permissions: { some: PermissionFilter }` (backend #14873, BA-8077). **Every** returned role counts. The first cut used `RoleFilter.permission` (`PermissionNestedFilter`) from backend #14678; #14847 (BA-8061) removed that field before it shipped, so every deployed manager rejected the query with `Field "permission" is not defined by type "RoleFilter"`.
- **Two branches on `rbac-single-scope-role`.** The action stays hidden below 26.8.0 (`role-mapped-scope-filter`, unchanged). From 26.9.0 the filter asks for the `scope_admin` permission; below that (`RoleFilter.permissions` does not exist there, so the request would be rejected outright) it keeps the previous lookup: the project scope's SYSTEM role pair (`project-<id>-member` / `project-<id>-admin`), of which `selectProjectAdminRoles` keeps the one whose name ends in `admin`. 26.8.0 deployments must keep working.
- `mappedScope.scopeType` is matched with `iEquals: 'project'`. ADR 0006 records that 26.9 returns scope types as lowercase strings, and `RoleMappedScopeNestedFilter.scopeType` is a plain `StringFilter`, so the old `equals: 'PROJECT'` would match nothing.
- Assign and revoke act on **all** of the project's admin roles instead of a single one.
- A user holding several of those roles shows up as one table row, and revoking the row drops every one of them. The row key moved from the assignment id to the user id accordingly.
- Both mutations use `Promise.allSettled`.
  - `BulkAssignRolePayload.failed` is always empty on 26.9 and every refusal raises. With N roles, "one granted, one refused" is the normal shape of a failure, and `Promise.all` would refetch the table while the remaining calls were still in flight.
  - `projectId`, which also writes the user's allowed-project list, is sent on one call only, so concurrent calls do not race on the same list.
  - Per-user failure notifications are keyed by role, so one role's message no longer overwrites another's.
- `data/schema.graphql` is the supergraph at backend main `05f5255fbb` (the #14873 merge commit); the generated Relay artifacts follow it.
- The `adminRoles` page size went from 10 to 50, because the permission filter can legitimately match more than the old SYSTEM pair.
- The modal title's RBAC shortcut opens the first admin role by name. The RBAC page shows one role at a time, and the roles are sorted by name so the choice is stable.
- The Projects page's pre-open check ("does this project have an admin role?") uses the same helpers as the modal, so the check and the modal agree on what counts as an admin role.

ADR 0006 is followed rather than extended, so no new ADR.

## Decisions to review

- **Custom roles are included.** There is no `source: SYSTEM` restriction, because the issue asks for *every* role in the scope carrying `scope_admin`. A custom role carrying `scope_admin` is therefore granted and revoked together with the system one. The table does not show role names, so a revoke is not visibly multi-role.
- **No permission-bit filter.** ADR 0006's glossary defines project admin as **`READ`** on `scope_admin`, and `useCurrentUserProjectRoles` enforces that bit. This filter matches a role carrying *any* bit on `scope_admin`, because the issue specifies matching by entity type. Adding `permission: { equals: 'READ' }` would align the two definitions as a one-line follow-up.
- **The role list is capped at 50** with no signal. `count` is fetched but not read, so a scope with more matching roles would silently under-report.
- **Revoke has no confirmation step.** `.claude/rules/destructive-confirmation.md` rule 6 names "revoking a role assignment" as a candidate for `BAIDeleteConfirmModal` with `reversible`. The missing confirmation predates this PR, but one click now drops N assignments. It is out of scope here because it needs new i18n keys.
- `Role.users` has been `@deprecated` since 26.9 in favour of `usersV2`. It is not touched here and is worth a follow-up issue.

## Test plan

- `VERIFY_TESTS=1 bash scripts/verify.sh` → `=== ALL PASS ===`.
- New unit tests in `react/src/components/ProjectAdminSettingModal.test.ts` (9 tests) check three things:
  - the filter shape: the `scope_admin` permission, the lowercase scope type, and no `source` restriction on 26.9.0+; the SYSTEM pair below that; and that the two branches never mix;
  - that every returned role is kept, in name order, on 26.9.0+, and only the `-admin` one below;
  - the per-user grouping: one user across two roles is one row with two `roleIds`.
- Manual check on a 26.9 deployment carrying backend #14873: open a project's "Set project admin", then add and remove a user. A 26.9 manager without #14873 (or a gateway still serving the older supergraph) answers `Field "permissions" is not defined by type "RoleFilter"`.
- Repeat on a 26.8 deployment and confirm the request carries the SYSTEM pair filter and the `-admin` role is found. Note that a manager reporting a pre-release version such as `26.9.0a3` compares below `26.9.0` and takes this branch too.

## Files

- `react/src/components/ProjectAdminSettingModal.tsx`
  - adds the `buildProjectAdminRoleFilter`, `selectProjectAdminRoles` and `groupProjectAdminAssignmentsByUser` helpers, the first two switching on `rbac-single-scope-role`;
  - makes assign and revoke act on every admin role.
- `react/src/pages/ProjectPage.tsx`: the pre-open existence check uses the same helpers.
- `react/src/components/ProjectAdminSettingModal.test.ts`: new.
- `data/schema.graphql` + `**/__generated__/**`: re-synced to backend main `05f5255fbb` (#14873).

🔗 **Dev server**: http://fr-3986-pr9782.seungwon.10-82-0-159.sslip.io


nowgnuesLee:
## Ready gate — blocked, needs a human call

Ran the `pr-ready-gate` flow headless overnight. **The PR is still a draft and I did not flip it.**

### Copilot could not review

Requested twice (15:25 UTC, 15:27 UTC). Both passes came back with the same refusal, not a review:

Review threads on this PR: **0**. So there was nothing to triage, reply to, or resolve — the gate's whole premise (a first reader before the humans) has not been met. Someone with remaining Copilot quota needs to re-request `@copilot`, or the gate has to be waived deliberately.

### What did pass

- `bash scripts/verify.sh` → `=== ALL PASS ===` (18s; the two Astryx token-gate findings are pre-existing on `main`, not from this branch).
- Working tree clean, `HEAD` == `origin/HEAD` for this branch — nothing unpushed.
- Dev server published to the PR body: <http://fr-3986-pr9782.seungwon.10-82-0-159.sslip.io>
- I read the diff myself in Copilot's place and found nothing objectively wrong to fix, so **no new commits were pushed**. One behavioural delta worth knowing about, not a defect: on a *partial* assign failure the form is no longer reset, so the users that were attempted stay in the field. That looks deliberate and better; flag it if not.

### Still needs your call (carried over from the implementation — all of these are in the PR description too)

| # | Decision | Why I did not decide it |
|---|---|---|
| 1 | `source: SYSTEM` dropped on the 26.9 branch | A **custom** role carrying `scope_admin` is now granted and revoked alongside the system one, and the table shows no role names, so a multi-role revoke is invisible. It is what "every role carrying `scope_admin`" implies, but it is a real behavioural widening. |
| 2 | No permission-bit filter | ADR 0006 defines project admin as **`READ`** on `scope_admin` and `useCurrentUserProjectRoles` enforces that bit; this filter matches *any* bit on the entity type. `permission: { equals: 'READ' }` would align the two — one line. |
| 3 | Role list capped at 50, `count` fetched but unread | A scope with more matching roles silently under-reports. Needs a decision on paging vs. a warning. |
| 4 | Revoke has no confirmation step | `.claude/rules/destructive-confirmation.md` rule 6 names "revoking a role assignment" as a `BAIDeleteConfirmModal` (`reversible`) candidate. Pre-existing, but one click now drops N assignments. Left out of scope — needs new i18n keys. |
| 5 | Not exercised against a live manager | Headless run with no backend session, so the two 26.9 literals (`scope_admin`, lowercase scope type) are argued from `data/schema.graphql` and ADR 0006, not from a live query. Worth a manual pass on a 26.9 deployment **and** on a 26.8 one (the legacy branch must still resolve the single SYSTEM admin role). |
| 6 | `Role.users` is `@deprecated` since 26.9 in favour of `usersV2` | Not touched here; worth a follow-up issue. |

Nothing is blocked on me. Next step is a human's: either re-run Copilot with available quota, or answer the six above and flip to ready.

nowgnuesLee:
Dev server on box `seungwon`: http://fr-3986-pr9782-find-project-admin.seungwon.10-82-0-159.sslip.io

ironAiken2:
Dev server on box `sungchul`: stopped (was http://fr-3986-pr9782-find-project-admin.sungchul.10-82-0-159.sslip.io)

ironAiken2 (approved):
LGTM

PR #9784, opened 2026-09-18
fix(FR-3998): name the IP restriction when login finds no keypair

nowgnuesLee:
Linked Jira: FR-3998 https://lablup.atlassian.net/browse/FR-3998

## Summary

A login from outside an account's `allowed_client_ip` range authenticates fine and then fails on the post-login `keypair` query — the manager scopes keypairs by client IP, so the query returns nothing. The login panel showed "Login failed. Check login information." plus "Keypair information is missing.", which is indistinguishable from a real keypair problem: neither the user nor the admin could tell the connection was blocked by the IP allow-list.

The block itself is correct and unchanged; only what the UI says about it changes.

## What changed

- `react/src/helper/loginSessionAuth.ts` — `connectViaGQL` throws a typed `KeypairUnavailableError` instead of a bare `Error`. It carries an `isKeypairUnavailable` marker and an exported `isKeypairUnavailableError()` type predicate so callers duck-type it across module boundaries (HMR, mocked imports), the same way `TokenLoginFailedError` is already classified. The cleanup `client.logout()` before the throw is best-effort: if it rejects (API-key mode has no webserver session, so `POST /server/logout` can 401) the typed error must still be what propagates.
- `react/src/components/LoginView.tsx` — all three connect paths act on it. `handleLoginError` is the important one: `connectUsingSession` runs the post-auth GQL bootstrap inside the same `try` as `client.login()`, so the reported scenario (no session cookie, user types credentials) lands there, not in `handleGQLError`. `handleGQLError` (already-logged-in fast path) and `connectUsingAPI` (API key / secret key mode) are covered too.
- `react/src/components/STokenLoginBoundary.tsx` — new `keypair-unavailable` error kind, classified ahead of the token-exchange branches so it covers both the `tokenLogin` path and the existing-session fast path that calls `connectViaGQL` directly. The default error card picks up `sTokenLoginBoundary.ErrorKeypairUnavailableTitle` / `...Description` through its existing kind-to-key mapping, and suppresses the raw `causeDetail` line for this kind — it is the untranslated internal constant the localized description exists to replace. It stays in the "Copy error details" payload.
- `resources/i18n/*.json` — four new keys in all 21 locales, each translated against that locale's existing `credential.AllowedClientIP` wording.

## Copy

> Cannot access your keypair.
>
> Your sign-in succeeded, but no keypair is available for your account from this network address. Access is usually limited to the addresses listed in Allowed client IPs. Contact your administrator to allow your current IP address.

An empty keypair result does not prove the IP rule is the cause — a deleted or inactive keypair, or a manager-side resolver error, reaches the same branch. So the message names the allow-list as the usual reason and points at the administrator rather than asserting it. If a future backend returns an explicit `client-ip-not-allowed` problem type on this path, the copy can be tightened to a definite statement.

## Tests

- `react/src/helper/loginSessionAuth.test.ts` (new) — `connectViaGQL` logs out and throws `KeypairUnavailableError` on an empty keypair, still throws it when the cleanup logout rejects, and the predicate recognizes a structurally identical copy while rejecting unrelated errors.
- `react/src/components/STokenLoginBoundary.test.tsx` — two cases asserting `kind: 'keypair-unavailable'` on the token-exchange path and on the existing-session fast path, and no `backend-ai-connected` dispatch in either.

## Verification

`VERIFY_TESTS=1 bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, Search index, Lint, Format, TypeScript, Astryx gates, Vitest react / backend.ai-ui / agent-cli / root). The Astryx token gate reports its two pre-existing `--bai-dialog-z` findings, unchanged by this branch.

## Known gaps

- Not reproduced against a live cluster: the failure needs an account whose `allowed_client_ip` excludes the tester's address. The error paths themselves are covered by the unit tests above.
- There is no `LoginView` test file in the repository, and this PR does not add one — the `handleLoginError` routing gap above was found by review, not by a test, and a render harness for that component is more work than this fix warrants. Worth a follow-up.

🔗 **Dev server**: http://fr-3998-pr9784.seungwon.10-82-0-159.sslip.io


nowgnuesLee:

Ran the `fw:pr-ready-gate` procedure (steps 1–7) unattended overnight. It did **not** pass, and the PR stays a draft.

**Copilot could not review.** Two passes were requested (15:53Z and 15:56Z); both came back with the same body and no inline threads:

This is a per-account quota on the requester (@seungwon2), not a problem with the diff. Nothing to triage — buckets A / B / C are all empty because no findings were produced.

**What did pass:**

- `bash scripts/verify.sh` → `=== ALL PASS ===` on `7b42fd7` (the Astryx token gate reports only its two pre-existing `--bai-dialog-z` findings).
- Dev server URL published to the PR body: <http://fr-3998-pr9784.seungwon.10-82-0-159.sslip.io>. No Teams thread was recorded for this PR at boot, so nothing was posted there.

**What is left for a human:**

1. Re-request Copilot once the quota resets (or request it from another account), then run steps 3–5 on the review it produces.
2. Flip the PR to ready only after that. Per the gate's own rule, ready-for-review means a first reader has already been through it — right now nobody has.

Review threads: 0 unresolved, 0 total, not truncated — but that is the absence of a review, not a clean one.

PR #9785, opened 2026-09-18
fix(FR-4000): stay put after creating a folder in the sub path picker

### LINKED ISSUE #9773: Sub path picker modal: stay in the current folder after creating a folder instead of opening it
## Problem

In the modal for selecting a folder's sub path, creating a new folder immediately navigates into that new folder. Because a just-created folder is empty, the user lands on an empty listing right after pressing Create, which reads like something went wrong rather than a deliberate move.

## Proposal

Remove the auto-navigation after folder creation. Keep the user in the current folder, show the newly created folder in the list, and let them open it themselves if they want to.

—

JIRA Issue: FR-4000

nowgnuesLee:
Resolves #9773 (FR-4000)

## Summary

The sub path picker modal (`BAIVFolderPathPicker` → `BAIDirectoryPickerModal` → `BAIFileExplorer` in `directoryPicker` mode) navigated into a folder the instant it was created. Because a fresh folder is empty, pressing **Create** dropped the user on an empty listing, which reads like the create failed rather than like a deliberate move.

The picker now stays put: the newly created folder simply appears in the current listing, and the user opens it if they want to.

## Changes

- `BAIFileExplorer.tsx` — no longer passes an `onFolderCreated` handler in `directoryPicker` mode (the handler called `navigateDown(folderName)`).
- `ExplorerActionControls.tsx` — the `onFolderCreated` prop goes away; nothing else used it.
- `CreateDirectoryModal.tsx` — `onRequestClose` drops its now-unused `createdFolderName` argument.

No refresh wiring was needed: the existing `onRequestClose(true)` path already calls `refetch()`, so the new folder shows up in place.

## Test plan

- New regression test `packages/backend.ai-ui/src/components/baiClient/FileExplorer/BAIFileExplorer.createFolder.test.tsx`: renders the explorer in `directoryPicker` mode, creates a folder, and asserts the listing gains the new folder while the existing entry stays visible and `onChangeCurrentPath` is never called again. Verified that it fails against the pre-change code (the old behaviour navigated away, so the sibling entry disappeared).
- `VERIFY_TESTS=1 bash scripts/verify.sh` → BUI / agent-cli / root suites pass. One react-suite failure, `GlobalSearchPalette.test.tsx`, is a `beforeAll` lazy-import warm-up timeout under parallel load and is unrelated to this change — it passes when run on its own.

## Manual check

Open a session launcher folder mount → sub path picker, press **Create Folder**, name it, confirm. The picker should stay in the folder you were browsing, with the new folder listed.

🔗 **Dev server**: http://fr-4000-pr9785.seungwon.10-82-0-159.sslip.io


nowgnuesLee:
## Ready gate — blocked on Copilot quota (unattended run)

Ran the `fw:pr-ready-gate` steps 1–7 overnight. Copilot could **not** review this PR:

| Pass | Review | Result |
|---|---|---|
| 1 | [5250037907](https://github.com/lablup/backend.ai-webui/pull/9785#pullrequestreview-5250037907) | `Copilot was unable to review this pull request because the user who requested the review has reached their quota limit.` |
| 2 | [5250051825](https://github.com/lablup/backend.ai-webui/pull/9785#pullrequestreview-5250051825) | same quota message |

Two passes is the skill's cap, so no third request was made.

**State**

- Review threads: 0 (nothing to triage, reply to, or resolve)
- `bash scripts/verify.sh` → `=== ALL PASS ===` (Astryx token gate reports 2 findings, both pre-existing in `BAIDialog.css` / `BAIDrawerPortal.css` — untouched by this PR)
- Dev server URL published to the PR body

**Needs a human**

The PR is deliberately left as a **draft**. The gate's premise — that a reviewer is the second reader, not the first — is not satisfied, because no first read happened. Someone needs to either re-run the gate once the Copilot quota resets, or decide to send it to human review without a bot pass.

PR #9786, opened 2026-09-18
feat(FR-4004): make the dev review overlay client host-agnostic so the Jira Helper extension can run it on any site

### LINKED ISSUE #9783: Make the dev review overlay client host-agnostic so the Jira Helper Chrome extension can run it on any site
The review pin overlay (`react/vite-plugins/review-overlay/client`) only runs where the webui dev server or the nightly static build injects it. Reviewers want the same pins on any website through the Lablup Jira Helper Chrome extension (`lablup/lablup-jira-cli-tool`): make a pin anywhere, open someone else's pin link anywhere. ADR 0002 keeps one codec, so the extension will vendor this client byte-identical instead of reimplementing it.

This issue makes that possible on the webui side:

1. Move the boot into an exported `bootOverlay(host)` behind an `OverlayHost` seam (server state, boot hash, pick chord, auto-navigation, palette, host marker) whose default host is today's behaviour.
1. Fix the few spots that do not compile under `exactOptionalPropertyTypes` / without `DOM.Iterable`.
1. Record an ADR naming the extension as a second host of the same client and the sync/drift-check mechanism.

No wire-format, block-grammar or label change.

JIRA Issue: FR-4004

yomybaby:
Resolves #9783 (FR-4004)

## Summary

The dev review overlay client (`react/vite-plugins/review-overlay/client/`) only ran where this repository's dev server or the nightly static build injects it. Reviewers want the same pins on any website through the Lablup Jira Helper Chrome extension (lablup/lablup-jira-cli-tool): make a pin anywhere, open someone else's pin link anywhere. This PR is the Web UI half. It lands **ADR 0008** and follows ADR 0002 and ADR 0004.

- **Host seam.** The boot moved verbatim from `client/main.ts` into `client/boot.ts`, which has no import-time side effect and exports `bootOverlay(options)`. `OverlayHostOptions` lets a host answer what only a dev server could: `state`, `bootHash`, `pageChords`, `autoNavigate`, `palette`, `marker`, `expectReactGrab`. `main.ts` stays the served/bundled entry and boots the default host, **which is today's behaviour** — no wire-format, block-grammar, label, cap or id change.
- **ADR 0002 stays in force, and this is the part a human should look at.** The extension does not reimplement the codec: it vendors `client/*.ts` byte-identical (sync script stamping the upstream commit and a sha256 per file, a drift gate in its CI, and this directory's vitest suite running there as the conformance suite). ADR 0008 records that as a second *host* of the one codec, not a second reader. I read that as extending 0002 rather than reversing it; if you read it as a reversal, this is the flag `.claude/rules/adr.md` rule 4 asks for.
- **What the extension host turns off, and why it is an option at all:** on a page the overlay does not own, swallowing ⌘C/Ctrl+C and guided mode's bare `n`/`p`/`c`… (`pageChords`), letting a crafted fragment `location.assign` a reader to another same-origin path (`autoNavigate`), and trusting the page's `--color-*` tokens (`palette: 'own'`) are all unacceptable. A host that turns page chords off is never shown them in the chrome either.
- **Compile contract.** The client now compiles unmodified under the extension's flags (`exactOptionalPropertyTypes`, no `DOM.Iterable`, `moduleResolution: node`); `tsconfig.host.json` + `host-conformance.ts` keep it that way as a new `verify.sh` lane (4 s, root TypeScript, no dependency change).
- **Three resolve/draw fixes found by running the client on github.com from the extension**, all real here too: a pin could resolve to a `display: none` look-alike that comes first in the DOM; a held element the page hid was never looked for again; and `place()` drew marker, box and card at (0,0) when the element's rect went empty during a re-render. Each new test fails on the previous commit. In jsdom, which lays nothing out, every new preference stands down, so the existing suite sees the old behaviour. So the default host does change in one direction: an element with no box is now given up and looked for again instead of being drawn on at (0,0) — a `display: contents` wrapper, whose children render but which has no box of its own, lands in that case too. Beside them the dock's away rows drop their "where" onto a line of its own, which is the one purely visual change in this PR; ADR 0008's Consequences name both.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (17 lanes incl. the new `TypeScript (review overlay host seam)`; the token gate reports the same 2 pre-existing findings in files this PR does not touch).
- `cd react && pnpm vitest run vite-plugins/review-overlay` → 29 files, **750 tests** passed (716 before; no existing assertion was removed or weakened — `git diff origin/main...HEAD -- '*.test.ts'` deletes no `expect`, and the only edit to an existing test file is the `pageChords` line every `createSetDock` call now passes).
- Default host checked by diffing the moved boot against `origin/main` side by side and by replaying the plugin's own `clientModule()` mapping: `/__review/overlay.js → main.ts → ./boot.js`, and `.claude/skills/walkthrough/scripts/mint.mjs`'s imports (`anchor`, `codec`, `id`, `stop-guard`, `guided`) are unchanged and still served.
- The extension host was exercised in a real Chromium with the extension loaded — 22 scenarios incl. github.com and developer.mozilla.org, strict CSP + Trusted Types, SPA `pushState`, native stand-down. That run is what found the three fixes above.
- **The default host was then exercised in a real Chromium against this branch's dev server** (URL in the dev-server comment below), on the signed-in app rather than the login screen. What was checked, in one pass: `/__review/overlay.js` and `/__review/boot.js` served as JavaScript and `/__review/state` fetched by the client; exactly one `[data-bai-review-overlay]` host, marker value empty; Ctrl+C still handing the pick to react-grab (this build loads it, so the fallback chord stays disarmed, as before); pick → composer → note → ⌘⏎ writing a block through `execCommand` on the plain-http gateway origin, carrying the ⚛️ stack, `pr=9786`, and the `Open on dev server` label; that link opened in a fresh tab drawing the pin on the same element (not at 0,0) with its card and note, and scrubbing `bai=v3` out of the fragment; the dock advertising `Ctrl⇧H`, and the chord hiding and re-showing the cards. No page errors, and nothing behaved differently from `main`.
- **Not exercised:** guided mode, on either host.

## Follow-ups (not in this PR)

- `ui.ts` / `guided.ts` bind their Escape handlers on `document` in the bubble phase; a host that shields overlay keystrokes from the page has to let Escape through by hand. Binding them on the shadow root as well would remove that coupling.
- After merge, the extension re-syncs its vendored copy against the merge commit on `main`.

🔗 **Dev server**: http://fr-4004-pr9786-make-dev-review.jongeun.10-82-0-159.sslip.io


yomybaby:
Dev server on box `jongeun`: http://fr-4004-pr9786-make-dev-review.jongeun.10-82-0-159.sslip.io

yomybaby:
## Ready gate — reviewer substituted

Copilot was requested on this draft twice (before and after the commit below) and both times answered with the same body, not a review:

So the gate's first reader is **not** Copilot on this PR. In its place the final diff was read adversarially against three things: whether the default host still behaves as it did on `main`, whether the three resolve/draw fixes are correct, and whether ADR 0008 says what the diff does. That substitute reader is an agent, not a human — the gate's point is that a human is the *second* reader, and here the first one was weaker than usual. Worth knowing before you read.

### Default host, in a browser

Before the review, the default host was exercised in a real Chromium against this branch's dev server (URL in the body). Everything the PR claims about it held: both `/__review/*.js` entries served, `/__review/state` fetched, one `[data-bai-review-overlay]` host with an empty marker, Ctrl+C handing the pick to react-grab, pick → composer → note → ⌘⏎ producing a `#bai=v3` block with the ⚛️ stack and `pr=9786`, that link drawing the pin on the same element in a fresh tab and scrubbing the fragment, and `Ctrl⇧H` advertised and working. No page errors. Nothing regressed.

### What the review found

Two claims were wrong, both in prose, neither in code:

1. **"Default host unchanged" was too wide.** The three resolve/draw fixes are not seam plumbing — they change what the default host does with an element that has no box (given up and looked for again, instead of drawn on at 0,0; a `display: contents` wrapper is one of those elements). And the dock's away rows now wrap their "where" onto its own line, which is a purely visual change with nothing to do with the host seam. Fixed in 38ab38581: ADR 0008's Consequences now names both, the PR body says the same, and the stale `display: contents` mention in `place()`'s comment is gone. `bash scripts/verify.sh` → `=== ALL PASS ===`, overlay suite 750/750.
2. **The Verification section claimed a `location.pathname` assertion had been removed.** It had not: `git diff origin/main...HEAD -- '*.test.ts'` deletes no `expect` at all, and the only edit to an existing test file is the `pageChords` line every `createSetDock` call now passes. The body now says that instead.

Nothing else was changed. The rendered-over-hidden preference in `resolve.ts`, the re-resolve of a hidden `held` in `pin.ts`, and the empty-rect give-up in `place()` were each walked through by hand against their candidate orderings and found correct, and all three stand down under jsdom, which is why the old suite still sees the old behaviour.

### Left open for a human — please do not treat this as settled

**Does ADR 0008 extend ADR 0002, or reverse it?** ADR 0002 says the format has *one reader*. The extension vendors `client/*.ts` byte-identical behind a sync script, a sha256 drift gate and the same vitest suite, so there is one codec and one *copy* of it rather than a second implementation — which reads as extending 0002. But "one reader" could just as well be read as "one place the code lives", and a vendored copy is a second place. That is a judgement about a decision in force, so `.claude/rules/adr.md` rule 4 puts it in front of a person, not an agent. The PR description raises it too; it stays open here deliberately.

Also still not exercised: guided mode, on either host.

agatha197 (approved):
LGTM

PR #9793, opened 2026-09-20
fix(FR-4005): reset the folder explorer's side panel per session

### LINKED ISSUE #9787: Folder explorer: the selected side-panel tab persists after closing the explorer and opening another folder
## Problem

The folder explorer's side panel has two tabs, **Metadata** and **Audit log**. The selected tab is remembered across explorer sessions: after selecting Audit log, closing the explorer, and opening a different folder, the explorer still shows the Audit log tab instead of starting on Metadata.

## Expected

Opening the explorer resets the side panel to its default tab (Metadata), so a newly opened folder always starts on the same tab.

—

JIRA Issue: FR-4005

nowgnuesLee:
Resolves #9787 (FR-4005)

## Problem

The folder explorer's side panel (Metadata / Audit log) remembered its tab across explorer sessions: select **Audit log**, close the explorer, open a different folder — the panel still showed Audit log.

The explorer is mounted for the app's lifetime (`FolderExplorerOpener` renders it always and only toggles `open`), so `activeTab` — and the loaded audit-log query ref with it — outlived the folder it was chosen for.

## Change

The tab and the audit-log query loader move out of `FolderExplorerModalV2` into a new local `VFolderInfoPanelSession`, which publishes the rendered panel through context so the layout branches can slot it in:

- It mounts **inside the modal body**, and `BAIModal` renders no children while closed (`BAIModal.tsx:441`) — so every explorer session starts on Metadata.
- It is **keyed by folder**, so changing `?folder=` under an already-open explorer is a new session too.
- It sits **above the `xl` layout fork**, so crossing the breakpoint still leaves the panel alone (the two branches are different element types, so anything below the fork remounts on resize).

Side effect, deliberate: the previous folder's audit-log query ref is released with the panel, so the panel can no longer show folder A's audit log under folder B's header — the `auditLogQueryRef == null` guard used to keep it alive. It also stops the ref being retained for the app's lifetime.

An earlier attempt reset the state from a `useEffect` on close; the React Compiler lint rule (`react-hooks/set-state-in-effect`) rejects that, and the structural fix is the better answer anyway.

## Test

`react/src/components/FolderExplorerModalV2.test.tsx` gains an FR-4005 case: open, select Audit log (asserted via `aria-current` on the tab), close, open another folder, assert the metadata panel is back. Verified to fail against `main`'s component and pass with the change. The existing render helper now takes a session `{ open, vfolderID }` and can rerender, mirroring what the opener does.

## Verification

`VERIFY_TESTS=1 bash scripts/verify.sh` → `=== ALL PASS ===` (Vitest react / backend.ai-ui / agent-cli / root)

🔗 **Dev server**: http://fr-4005-pr9793.seungwon.10-82-0-159.sslip.io


nowgnuesLee:

Ran the `fw:pr-ready-gate` steps unattended (steps 1–7; the ready flip was out of scope for this run).

| Step | Result |
|---|---|
| Preconditions | draft ✓, tree clean ✓, head pushed (`05ddf51`) ✓, body final ✓ |
| `bash scripts/verify.sh` | `=== ALL PASS ===` (18s) |
| Copilot pass 1 | **refused** — "unable to review this pull request because the user who requested the review has reached their quota limit" |
| Copilot pass 2 (re-requested) | **refused, same reason** — so it is the account quota, not a transient failure |
| Review surface | 0 inline threads, 0 review comments — there is nothing to triage, fix, or resolve |
| Dev server URL | published to the PR body |

**Nothing needs a decision from the diff** — there are no findings, because no review ran.
The one thing that needs a human: the Copilot review quota has to reset (or the review be
requested from another account) before this PR has had a first reader. Per the gate's own
rule, a PR whose Copilot pass did not happen does not go to humans as "already reviewed",
so **it stays a draft**.

Dev server: <http://fr-4005-pr9793.seungwon.10-82-0-159.sslip.io>

PR #9795, opened 2026-09-20
fix(FR-4006): clear every filter chip with one clear-all press

### LINKED ISSUE #9788: GraphQL property filter: the clear-all button removes only one filter chip per click
## Problem

The GraphQL property filter input (e.g. Data → Folders) shows a clear-all ✕ at the right edge of the field, next to the per-chip ✕ on each filter. Pressing the clear-all button removes a single filter chip instead of clearing them all, so with several filters applied the user has to press it once per chip.

## Expected

The button at the right edge of the input clears every applied filter at once; the per-chip ✕ keeps removing just that chip.

See the attached screenshot.

—

JIRA Issue: FR-4006

nowgnuesLee:
Resolves #9788 (FR-4006)

## Problem

The clear-all `✕` at the right edge of the property filter (Data → Folders, and every other page that mounts `BAIPropertyFilter` / `BAIGraphQLPropertyFilter`) removed **one** chip per press. With three filters applied the user had to press it three times.

## Cause

Upstream, in `@astryxdesign/core` 0.5.4. `Tokenizer.handleClearAll` reports the clear as `onChange([], { item: lastToken, type: 'remove' })` — an empty item list plus, by convention, the last token as the removed one. `PowerSearch.handleTokenizerChange` ignores the item list and removes only the filter at `item`'s index, so each press dropped exactly the last chip.

## Fix

Extends the existing `react/patches/@astryxdesign__core@0.5.4.patch` with a hunk on `dist/PowerSearch/PowerSearch.js`: when a `remove` change arrives with an emptied token list **and more than one filter was applied**, `PowerSearch` clears them all instead of one. The `filters.length > 1` guard keeps every single-chip path (a chip's own `✕`, backspace on an empty input, a clear-all with one filter) on the original code, so the patch only touches the reported case. Filters marked `isReadOnly` are kept, matching `canInteract`, which already denies them a per-chip remove button — no call site in this repository sets that flag today.

Patching is the same route the repository already takes for the `ComplexSelector` `hasClear` gap in this dependency; no BUI call site or public prop changes. The `index` argument `PowerSearch` reports alongside a bulk clear has no meaningful value; the hunk passes `0` as a placeholder and both consumers ignore it.

## Tests

A regression test in each filter component — two filters applied (asserted, so the test cannot pass unpatched), one click on "Clear all", one `onChange` carrying the empty value. `BAIPropertyFilter` additionally asserts that a chip's own remove button still removes only that chip.

The e2e suite already covered this (`e2e/environment/environment.spec.ts` scenarios 2.7 and 2.9), but its `resetAllFilters` helper retries the click until the button disappears, so the bug passed as a slow clear rather than a failure. Those scenarios now pass on the first click; they needed no change.

## Verification

`VERIFY_TESTS=1 bash scripts/verify.sh` → `=== ALL PASS ===` (Relay, search index, lint, format, tsc, Astryx gates, Vitest react + backend.ai-ui + agent-cli + root).

🔗 **Dev server**: http://fr-4006-pr9795.seungwon.10-82-0-159.sslip.io


nowgnuesLee:

Ran the `pr-ready-gate` steps 1–7 unattended. Copilot **could not review this PR**: both passes came back with

(requested at 16:01Z and again at 16:04Z after a 2-minute wait — same answer both times, so it is not transient.)

| Bucket | Count | Outcome |
|---|---|---|
| A — fixed | 0 | no findings to act on |
| B — needs judgement | 1 | **this**: the gate has no review to gate on |
| C — won't apply | 0 | — |

Copilot passes: 2 requested, 0 delivered. Review threads on the PR: 0 (none were created).

**Needs a human call in the morning:** re-request Copilot once the quota window resets, or decide to send this straight to a human reviewer. The PR stays a **draft** either way — nothing was flipped.

Preconditions were otherwise green: working tree clean, head pushed (`01e99c2`), `bash scripts/verify.sh` → `=== ALL PASS ===` (Astryx token gate reports its 2 pre-existing `--bai-dialog-z` findings, unchanged from `main`).

Dev server for review is in the PR body above.

nowgnuesLee:
Closing: this is an Astryx bug. Tracked in the Astryx repo here: https://github.com/facebook/astryx/issues/6409

Issue #9789, opened 2026-09-20
Session Scheduling History modal: the whole modal scrolls horizontally instead of the table


nowgnuesLee:
## Problem

In the Session Scheduling History modal the table is wider than the modal, and the horizontal scroll is applied to the **entire modal** instead of the table body. The header, the search box and the pagination row scroll out of view together with the columns.

The table already passes `scroll=` x: 'max-content' `` (`packages/backend.ai-ui/src/components/fragments/BAISchedulingHistoryNodes.tsx:165`), so the prop is not producing a table-scoped scroll container here. The expanded-row step table shows the same overflow.

## Expected

Only the table body scrolls horizontally; the modal itself, its title, search input and pagination stay fixed.

See the attached screenshot.

—

Captured while working on branch: main

JIRA Issue: FR-4007

PR #9825, opened 2026-09-21
fix(FR-4007): contain a wide table's horizontal scroll in flex and grid parents

agatha197:
Refs #9789 (FR-4007)

## What this is

General `BAITable` hardening: it contains a wide table's horizontal scroll inside
Astryx's own scroll wrapper when the table sits in a **flex-row, grid, or
unstretched-column** parent. Without it, such a parent sizes `BAITable`'s
outermost div to the table rather than to the available width, the scroll wrapper
never scrolls, and the overflow escapes to whatever scroll container is above —
the page, or a modal's shared title/filter/table/pagination region.

> [!IMPORTANT]
> **This is not the fix for #9789.** That report describes the Session Scheduling
> History modal, whose table sits in `BAIFlex direction="column" align="stretch"`
> — the one flex arrangement that already clamps the item. Measured live on this
> build, that chain contains its own overflow with **and** without this PR's rule,
> byte-identically, at every viewport width (table A below). The defect this PR
> fixes is real and reproducible in the same component for every other flex
> arrangement (table B), which is why the issue is referenced rather than closed.
> #9789 needs to go back to its reporter.

## The defect

Astryx's `Table` already renders the scroll container (`TableScrollWrapper`,
`overflow-x: auto`); since FR-3500/FR-3681 `BAITable`'s `scroll.x` only decides
whether the `<table>` may be wider than it. But that wrapper's width is
`calc(100% + …)` of the box around it, so it only scrolls while **that** box is
capped at the available width. `BAITable` left its own outer wrapper
unconstrained, so a flex or grid parent could size it to the table instead; the
scroll wrapper then inherited the full table width, never scrolled, and the
excess bled up to the nearest scroll container — in a modal, Astryx's
`LayoutContent`, the single region the filter row, the table and the pagination
bar share.

Two routes, neither of which `tsc`, ESLint or a runtime warning can see:

- a **row** flex parent floors the item at its content's min-content width (a
  flex item's automatic minimum size in the main axis). The exemption that would
  zero it — "an item whose overflow is not visible gets an automatic minimum size
  of 0" — does not apply, because the `overflow-x: auto` is on Astryx's wrapper
  two levels down, not on this box.
- a **column** flex parent that does not stretch its items (`BAIFlex`'s `align`
  defaults to `center`) sizes the item to max-content in the cross axis, with no
  clamp to the parent.

The reset belongs on `BAITable` rather than on any one caller: any flex or grid
ancestor around any `scroll.x` table reproduces it. `min-width: 0` is exactly the reset Astryx
applies to its own stack items (`stackItem.stylex.ts`); `max-width: 100%` closes
the unstretched-column route.

```css
.bai-table-astryx-root {
  min-width: 0;
  max-width: 100%;
}
```

### Why this is not CSS-only

The reset cannot live on `.bai-table-astryx-dim-layer`: that layer is a plain
block child, its `min-width: auto` already computes to `0px`, and a build with
the rule there measures no change at all. The element that is actually the flex
item is `BAITable`'s outer wrapper, which had no class of its own. Hence the
one-line TSX change that names it `bai-table-astryx-root`; the caller's
`className` is still merged onto the same element, so no call site changes.

## Live measurement (manager 26.9.0a3, Chromium)

The Session Scheduling History modal is used here as the **measurement harness**
— a real `scroll.x` `BAITable` with real data — not because this PR changes it.
Session `844d3dd2…` (9 history rows), dev server on this branch. **"rule off"** re-measures the same live page with
`.bai-table-astryx-root { min-width: auto; max-width: none }` injected — the
computed state of `main`, which has no such rule.

`LayoutContent` is the modal's shared scroll region (`overflow-x: auto`) that the
title, search box, table and pagination all sit in. **The bug is present when its
`scrollWidth` exceeds its `clientWidth`.**

### A — the modal's real chain (`BAIFlex direction="column" align="stretch"`)

| viewport | rule | `LayoutContent` sw / cw | `TableScrollWrapper` sw / cw | table |
|---|---|---|---|---|
| 1280 | on  | 1280 / 1280 | 1280 / 1280 | 1280 |
| 1280 | off | 1280 / 1280 | 1280 / 1280 | 1280 |
| 1000 | on  | 1000 / 1000 | **1175 / 1000** | 1175 |
| 1000 | off | 1000 / 1000 | **1175 / 1000** | 1175 |
| 800  | on  | 800 / 800   | **1175 / 800**  | 1175 |
| 800  | off | 800 / 800   | **1175 / 800**  | 1175 |
| 640  | on  | 640 / 640   | **1175 / 640**  | 1175 |
| 640  | off | 640 / 640   | **1175 / 640**  | 1175 |

Repeated with a row expanded (the `BAISubStepNodes` step table rendered) — same
numbers — and with the Message column forced long so the table measures 1443px:
at 1000px and 700px, `LayoutContent` still 1000/1000 and 700/700, the wrapper
scrolling 1443. All eight columns the table offers are shown by default, so
there is no wider column configuration to try; the modal is `variant="fullscreen"`
and exposes no window controls, so there is no narrower dialog state either.

Viewport screenshots for rule-on and rule-off are **byte-identical** at every
width (`md5` match), which is the same no-op this PR's Storybook sweep found for
block and stretched-column parents.

live modal, real chain — the table scrolls inside itself; title, search and pagination stay put

### B — the same live page with the parent switched to a flex row

Same modal, same data, `flex-direction: row` set on the table's parent from
DevTools. This is what a caller that lays a table out beside a sibling writes.

| viewport | rule | `LayoutContent` sw / cw | `TableScrollWrapper` sw / cw |
|---|---|---|---|
| 1000 | **off** | **1481 / 1000 — whole modal scrolls** | 1175 / 1175 (never scrolls) |
| 1000 | on      | 1000 / 1000 | **1175 / 694** |

And an unstretched column parent (`align-items: center`, `BAIFlex`'s default):

| viewport | rule | `LayoutContent` sw / cw | `TableScrollWrapper` sw / cw |
|---|---|---|---|
| 1000 | **off** | **1088 / 1000 — whole modal scrolls** | 1175 / 1175 |
| 1000 | on      | 1000 / 1000 | **1175 / 1000** |

Rule off — the header row runs off the right edge and the search box has been
pushed away with it; the table's own wrapper is not scrolling:

rule off, row parent — the whole modal scrolls

Rule on — the modal is back to its own width and the table scrolls inside itself:

rule on, row parent — the table scrolls inside itself

## Why #9789 stays open

#9789 describes the Session Scheduling History modal. On the current tree that
modal's parent is `BAIFlex direction="column" align="stretch"`, the one flex
arrangement that already clamps the item, so the symptom cannot arise there and
this PR changes nothing about it (table A). What this PR fixes is the identical
defect in the same component for every other flex arrangement (table B).

Possible readings, for the reporter to settle:

- the modal on the reporter's build differed from `main` today (a different
  Astryx version, a local edit, or a browser other than Chromium);
- the report is about a different table than the one measured here;
- the arrangement changed under the modal between the report and now — nothing in
  `git log` since 2026-09-10 touches `SessionSchedulingHistoryModal.tsx` or the
  scheduling-history table, so this looks unlikely.

This PR therefore lands as general `BAITable` hardening and references #9789
without closing it. If the reporter can still reproduce the modal symptom, the
repro steps (build, browser, session) would pin down what differs from the
measurements below.

## Blast radius

`.bai-table-astryx-root` is on every `BAITable` in the app, so this was measured
rather than argued.

- **Block context (the common case): provably inert.** `min-width: auto` already
  resolves to `0px` for a non-flex item, and a block box's width never exceeds
  `100%`. The `HorizontalScroll` story (560px block container) is byte-identical
  before and after.
- **Before/after sweep in Storybook, 39 table stories compared** (all 10
  `BAITable` stories, `BAIUserNodes` ×6, `BAINameActionCell` ×14, …): **zero
  differences** in root width, table width, or scroll-wrapper client/scroll width.
  The remaining 44 stories in the wider sweep (`BAIAgentTable`,
  `BAIArtifactTable`, `BAIProjectTable`) all already render root width ==
  container width with the table wider inside the wrapper, so the reset can only
  ever be inert for them too.
- **`max-width: 100%` cannot shrink a deliberately-wider table**: no call site in
  `react/src` or `packages/backend.ai-ui/src` sets an explicit width on
  `BAITable`'s root via `style`.
- **No `min-height: 0`.** `scroll.y` has no `true` form (`BAITable.tsx` maps it
  straight through `toCssLength`), so the `max-height` cap in `BAITable.css` is
  always an explicit length and never waits on a flex parent to release the block
  axis. The vertical case is therefore not the mirror image of this one, and the
  declaration would only add a regression risk (a height-capped column parent
  squashing rows under `overflow: visible`).

### Not changed

`BAISubStepNodes` (the expanded-row step table) needs nothing: it already owns a
local `.bai-substep-scroll` (`overflow-x: auto`) and is excluded from
`BAITable`'s outer `scroll.x` rules by `:not(table table)`. Measured expanded at
1280 / 1000 / 800 / 640 px on the live modal — it stays inside the outer scroll
wrapper.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (61s)
- `VERIFY_TESTS=1 bash scripts/verify.sh` → Vitest backend.ai-ui / agent-cli /
  root PASS. The react lane reports one failure,
  `GlobalSearchPalette.test.tsx` ("Hook timed out in 10000ms", 1120 other tests
  passing) — pre-existing and unrelated: it fails identically with `main`'s
  `packages/backend.ai-ui` checked out over this branch's, and this PR touches no
  file it imports.
- `node scripts/migration-gates/astryx-token-gate.mjs --strict` → `undeclared: 2`,
  the two pre-existing `--bai-dialog-z` findings in `BAIDialog.css` /
  `BAIDrawerPortal.css`. **No growth vs main.**
- Two focused cases in `BAITable.scrollX.test.tsx`: the root wrapper carries
  `bai-table-astryx-root` and contains the dim layer (with and without
  `scroll.x`), and a caller `className` still lands on the same element. jsdom has
  no layout, so these pin the hook the CSS reaches for, not the pixels — the pixel
  evidence is the live measurement above.


agatha197:
Dev server on box `sujin`: http://fr-4007-pr9825-keep-wide-table.sujin.10-82-0-159.sslip.io

PR #9805, opened 2026-09-21
docs: update admin_menu docs (2026-09-21)

lablup-octodog:
### What changed in the product

Three commits in `99c0ea2..ad23eda` changed what a super-admin sees and can do on the pages this chapter documents:

- **feat(FR-3908): add missing filters, orderings and columns to the super-admin Users page (#9631)** — Users list gains a property filter (base set of 10 properties, plus 10 more behind the `user-v2-extended-filter` manager capability), a default-hidden **Projects** column, a CSV-export caveat for unsupported filters; Credentials tab gains **Full Name** / **Last Used** columns and a 9-property filter.
- **feat(FR-3911): add missing filters, orderings and columns to the Environments page lists (#9637)** — Images tab gains six columns (**Image Status**, **Type**, **Local**, **Size**, **Aliases**, **Supported Accelerators**; five of them default-hidden) and a 15-property filter with the live-only-until-status-filtered behavior; Registries tab gains **Allowed Projects** and a default-hidden **Global** column; Resource Presets tab gains a Name/Resource Group filter and sorting.
- **fix(FR-3902): explain what the resource group Public toggle controls (#9602)** — the bare **Public** toggle got a real description (`resourceGroup.PublicDesc`); the chapter's old one-liner ("visible to all users") described roughly the opposite of what the toggle does.

The chapter documented none of this. After review (next section) the prose change is down to the Images column list and the **Public** description, in all four languages, plus 5 planned screenshot refreshes and 7 recapture-only images (FR-4002 Token/Badge restyle, FR-3182 image meta row) from the wrapper's image batch.

### Review follow-up

Review on the ko preview asked to drop the filter / column-settings prose: narrowing a list with the property filter and turning a column on from the settings gear are things a reader works out alone. c940a150a removes those nine paragraphs from en/ko/ja/th — the Users filter, **Projects** column and CSV-caveat paragraphs, the Credentials filter + **Full Name** / **Last Used** paragraph, the Images hidden-columns paragraph, the Images filter paragraph and its `:::note`, the Registries **Allowed Projects** / **Global** paragraph, and the Resource Presets filter paragraph. That also settles minor note 1 below, in the direction `DOCUMENTATION-STYLE-GUIDE.md` "What NOT to Document" §3 points. What stays: the six new Images column entries, the **Public** rewrite, and the screenshots.

### Review findings (independent review)

**No blocking issues found.** Spot-checked every factual claim against the source:

[...]

PR #9827, opened 2026-09-21
fix(FR-4020): drop the shared gh-pages concurrency group so a burst of PRs keeps every preview

### LINKED ISSUE #9826: PR preview publish job drops sibling PRs' previews through its gh-pages concurrency group
## Symptom

On 2026-09-21 the nightly docs bot opened 13 `docs/auto-*` PRs (#9805–#9817) within two minutes. Only 5 of them got a GitHub Pages preview; the other 8 (including #9805) ended with the `publish` job of the **PR Preview** workflow **cancelled** after one second, with no log, and `https://lablup.github.io/backend.ai-webui/pr/9805/docs/next/en/` returned 404.

## Cause

`publish` in `.github/workflows/pr-preview.yml` declares a job-level concurrency group shared across every PR:

```yaml
concurrency:
  group: pr-preview-gh-pages
  cancel-in-progress: false
```

GitHub Actions keeps **one running and one pending** job per concurrency group. `cancel-in-progress: false` protects the running job, but a newly queued job still replaces the pending one, which is cancelled. While #9806's publish was running (04:06:56–04:07:26), #9805's publish entered the pending slot at 04:07:05 and was cancelled one second later when #9807's publish arrived. The same displacement repeated down the chain.

The comment above the block intended a serial queue ("gh-pages takes one push at a time"), but that is not what the setting does.

## Fix

Every PR must always get its preview, so serialization through concurrency is the wrong tool here. Remove the job-level concurrency group and let the existing push-retry loop in the deploy step absorb push conflicts between sibling previews as well as the cleanup job: raise the attempt count and add jitter to the back-off so a burst of a dozen PRs all land.

JIRA Issue: FR-4020

ironAiken2:
Resolves #9826 (FR-4020)

## Summary

The `publish` job of the PR Preview workflow serialized gh-pages pushes with a concurrency group shared across every PR (`pr-preview-gh-pages`, `cancel-in-progress: false`). GitHub keeps one running and one pending job per group and cancels the older pending job whenever a newer one arrives, so when the nightly docs bot opened 13 PRs (#9805–#9817) within two minutes only 5 previews were published; the other 8 `publish` jobs were cancelled before their first step and their preview URLs 404ed.

Every PR must always get its preview, so this drops the job-level concurrency group and lets the existing push-retry loop absorb conflicts between sibling previews as well as the cleanup job. The per-PR workflow-level group (`pr-preview-<n>`, `cancel-in-progress: true`) is unchanged.

## Changes

- `.github/workflows/pr-preview.yml`: remove `concurrency` from `publish`; raise the deploy push retry from 5 fixed attempts to 12 with a jittered back-off.
- `.github/workflows/pr-preview-cleanup.yml`: reword the comment on its own concurrency group, which referred to the deploy's group.

## Verification

- Both workflows parse (js-yaml) and pass `prettier --check`; neither file is covered by `scripts/verify.sh` or lint-staged.
- This PR touches `pr-preview.yml`, so its own PR Preview run executes the modified `publish` job end to end.
- The burst case is exercised by the next nightly docs run: every `docs/auto-*` PR should get a `pr/<n>/docs` directory on gh-pages.

## Checklist

- [ ] 1. Open this PR's PR Preview run and confirm `publish` succeeds without a concurrency wait.
- [ ] 2. After the next nightly docs run, list `pr/` on gh-pages and confirm every `docs/auto-*` PR has a directory.


agatha197 (approved):
LGTM

PR #9833, opened 2026-09-21
fix(FR-4012): keep the session scheduling subscription inside the request line

### LINKED ISSUE #9796: Session scheduling-events subscription fails with a 400 from the strawberry endpoint (RESPONSE_VALIDATION_FAILED)
## Problem

`BAIComputeSessionNodeNotificationItemSubscription` — the session notification subscription on `schedulingEventsBySession` — fails immediately. The gateway sends the operation to the strawberry endpoint as a **GET** request and the manager answers `400 Bad Request` with a bare problem document, so the client reports `RESPONSE_VALIDATION_FAILED`.

Operation:

```
subscription BAIComputeSessionNodeNotificationItemSubscription($session_id: ID!) {
  schedulingEventsBySession(sessionId: $session_id) {
    reason
    session {
      status
      ...BAIComputeSessionNodeNotificationItemFragment
      ...SessionNodesFragment
      ...SessionDetailContentFragment
      id
    }
  }
}
```

Variables: `{ "session_id": "1aaf7b85-9cd1-4047-8cbf-a0bba26de185" `}

Error:

```json
{
  "errors": [
    {
      "message": "Unexpected empty \"data\" and \"errors\" fields in result: {\"type\":\"https://api.backend.ai/probs/generic-bad-request\",\"title\":\"Bad request.\",\"error_code\":\"backendai_generic_bad-request\"}",
      "path": ["schedulingEventsBySession"],
      "extensions": {
        "request": { "method": "GET", "url": "http://<manager-host>/admin/gql/strawberry?query=subscription+BAIComputeSessionNodeNotificationItemSubscription...&operationName=BAIComputeSessionNodeNotificationItemSubscription" },
        "response": { "status": 400, "statusText": "Bad Request", "body": { "type": "https://api.backend.ai/probs/generic-bad-request", "title": "Bad request.", "error_code": "backendai_generic_bad-request" } },
        "code": "RESPONSE_VALIDATION_FAILED",
        "serviceName": "strawberry"
      }
    }
  ]
}
```

The full operation text (with every spread fragment) and the untruncated error, including the complete request URL, are in the attached log.

## Expected

The subscription is established and delivers scheduling events for the session, instead of failing on the first request.

—

JIRA Issue: FR-4012

agatha197:
Resolves #9796 (FR-4012)

## Summary

`BAIComputeSessionNodeNotificationItemSubscription` spread `SessionNodesFragment` and `SessionDetailContentFragment`, so relay-compiler emitted every fragment reachable from them — 34 fragments, **7,754 bytes** of operation text.

A federated gateway opens a subgraph subscription over SSE, which carries the whole operation in the **GET query string**. URL-encoded, that document makes a **11,863-byte** request line. The manager's HTTP server caps a request line at **8,190 bytes** and answers a longer one with a bare problem document:

```json
{"type":"https://api.backend.ai/probs/generic-bad-request","title":"Bad request.","error_code":"backendai_generic_bad-request"}
```

which the gateway surfaces to the client as `RESPONSE_VALIDATION_FAILED`. The subscription therefore died instead of streaming scheduling events.

There is no "keep the fragments, trim a little" option — each spread is over budget on its own:

| spread | fragments | request line |
|---|---|---|
| all three (before) | 34 | 11,863 B — over |
| `SessionNodesFragment` alone | 33 | ~11,641 B — over |
| `SessionDetailContentFragment` alone | 29 | ~9,344 B — over |
| `BAIComputeSessionNodeNotificationItemFragment` alone | 11 | ~3,943 B — fits, but see below |
| field selection (this PR) | 1 (image tag) | **2,189 B** |

## Fix

Select only the fields a scheduling event actually changes. Relay's store is normalized by `id`, so writing those fields onto the session record refreshes every component that reads them — masking governs reads, not writes. The three fragments are untouched and still read by their own components (`SessionNodes`, `SessionDetailContent`, the notification item), each getting its ref from its own query; `useSubscription` returns `void`, so the spreads could never have handed data to a component. Their only effect was the store write.

The notification-fragment-only option fits the budget but is strictly worse: it is 2.3× larger, drags the AppLauncher / SFTP / ContainerLog / Terminate closures' static fields onto every event, and still misses `terminated_at`, `occupied_slots`, `agent_ids` and `idle_checks`.

### Where it changed, and what each step cost

The whole change is the `session { … }` selection set inside the subscription. Nothing else in the operation moved.

| step | what changed in the selection | operation text | request line |
|---|---|---|---|
| before | `...BAIComputeSessionNodeNotificationItemFragment` + `...SessionNodesFragment` + `...SessionDetailContentFragment` | 7,754 B | 11,863 B — over |
| 1. spreads → fields | the three spreads replaced by the 15 session fields a scheduling event changes, plus 4 volatile kernel fields | 660 B | 1,327 B |
| 2. + kernel identity | `row_id`, `cluster_role`, `cluster_idx`, `cluster_hostname` added so a kernel record the store sees for the first time is complete for `ConnectedKernelListFragment` | 757 B | 1,728 B |
| 3. − version gates | `@since` dropped from `queue_position` (25.13.0) and `idle_checks` (24.12.0) | 703 B | 1,650 B |
| 4. + kernel image | `image { ...BAIImageNodeSimpleTagFragment }` added — `SessionNodes` and `SessionDetailContent` read it off the same unargumented `kernel_nodes` connection | **994 B** | **2,189 B** |

Net: **7.8× smaller** operation text, **5.4× shorter** request line, 6,001 bytes of headroom under the 8,190-byte cap.

Step 3 is safe because `schedulingEventsBySession` is itself *Added in 25.15.0* (`data/schema.graphql`), which is above both fields' `@since` floors — any manager that has this subscription at all has `queue_position` and `idle_checks`. The gates were unreachable here. They stay in place on `SessionStatusBadge` / `SessionDetailContent`, which are reached on older managers.

### What is no longer refreshed

Of the 67 session fields the spreads carried, 34 are kept and 33 dropped. All but two of the dropped ones are fixed at session creation (`type`, `user_id`, `project_id`, `domain_name`, `scaling_group`, `access_key`, `cluster_mode` / `cluster_size`, `startup_command`, `resource_opts`, `requested_slots`, `owner`, `vfolder_mounts` / `vfolder_nodes`, kernel `image`) or change only through a mutation that patches the store itself (`name`, `priority`, `tag`). The two exceptions:

- **`dependees` / `dependents` statuses** — other sessions, with their own data path.
- **kernel `live_stat`** — the session list's `AutoUpdateFetchKeyButton` refetches it on its own interval; a scheduling event is a status transition, not a stats tick.

A scheduling event can also be the first time the store sees a kernel record, so the kernel selection carries the identity fields `ConnectedKernelListFragment` renders by (`row_id`, `cluster_role`, `cluster_idx`, `cluster_hostname`) plus the `image` spread `SessionNodes` and `SessionDetailContent` read off the same unargumented `kernel_nodes` connection — without it a kernel the subscription writes first lands with no `image`, and the Environments column renders `-` until the next refetch. Not only the volatile ones.

## Verification

`bash scripts/verify.sh` — all lanes pass. One unrelated pre-existing failure, reproduced on clean `origin/main` (`3857e7d576`): `Format`, on untracked `e2e/_cap_*.spec.ts` scratch files in the shared checkout. (`Vitest (react)`'s `GlobalSearchPalette.test.tsx` `beforeAll` hook timeout also reproduces on clean main; 583 tests pass, 0 test failures.)

**Live check against the manager at `10.82.130.172`**: subscribed to real sessions through their `PENDING → CREATING → RUNNING → TERMINATED` transitions with the compiled document. Every event arrives with all selected fields resolved, kernel identity fields included. The ungated `queue_position` / `idle_checks` document validates against that manager with no error. Probe sessions were terminated afterwards.

**Regression guard**: `react/src/subscriptionRequestSize.test.ts` asserts the budget for every compiled subscription artifact in `react/` and `backend.ai-ui`. It fails on the pre-fix artifact (12,063 B) and passes on the new one — nothing else in the toolchain catches this, since an oversized document compiles and type-checks fine and only fails on deployments whose gateway uses the GET/SSE subgraph hop.

## Note on reproduction

The dev manager (26.9.0a3) does **not** use the GET subgraph hop — the full document works there, which is why this is deployment-dependent and was reported from elsewhere. The size math is the same on every deployment, so the document was a latent landmine regardless; the guard keeps it from growing back.

🔗 **Dev server**: http://fr-4012-pr9833-keep-session-scheduling.sujin.10-82-0-159.sslip.io


agatha197:
Dev server on box `sujin`: stopped (was http://fr-4012-pr9833-keep-session-scheduling.sujin.10-82-0-159.sslip.io)

yomybaby (approved):
LGTM

Issue #9889, opened 2026-09-22
Session launcher folder list: read the caller's mount level from GraphQL instead of REST /folders


agatha197:
## Background

Review feedback on FR-3987 (PR #9751): the session launcher's mountable-folder list is fetched over REST (`GET /folders`, `useSuspendedLegacyVFolders`), and the reviewer asked for it to move to GraphQL.

Two problems with the REST call, both raised in review:

- `GET /folders?owner_user_email=B` scopes the **search** to B but computes each row's `permission` for the **caller**, so the level shown is not the level the named owner would mount at. The reviewer's position is that a user should not be able to enumerate another user's folders at all under RBAC.
- It is the last REST dependency in the launcher's folder step, which FR-3866 introduced.

## Why this is possible now

`vfolder_nodes.permissions` (legacy GraphQL, `VFolderPermissionValueField`) carries the **caller's effective mount level**, not the folder default. Verified against the 26.9.0 dev manager on 2026-09-22:

- folder whose resolved level is `rw` -> `mount_ro`, `mount_rw`, `mount_wd`
- folder whose resolved level is `ro` -> `mount_ro` only
- folder whose resolved level is `none` -> no `mount_*` bit at all

So the launcher can derive `ro` / `rw` / `none` from GraphQL without the backend adding a field. (The v2 `VFolder` type cannot do this: `accessControl.permission` is the folder default and `VFolder.permissions` is the RBAC bit set — READ/UPDATE/CREATE/SOFT_DELETE/HARD_DELETE — with no mount levels.)

## Scope

- Replace the REST fetch in `useSuspendedLegacyVFolders` (backend.ai-ui) with a Relay query on `vfolder_nodes`.
- Derive the caller's mount level from `permissions` and keep the `none` exclusion FR-3987 added.
- `useSuspendedAutoMountedFolders` and `BAIVFolderMountConfigInput` are the two consumers.

## Open decision

`vfolder_nodes` has no owner parameter, so the on-behalf case (`owner_user_email`) cannot be reproduced. Dropping it removes a working superadmin behaviour — launching a session for another user currently lists that user's folders. It was discussed in the FR-3987 thread: the suggestion was to replace it with an "acted as" superadmin feature, and the reviewer excluded it from the #9751 review. This issue should not land until that is settled.

JIRA Issue: FR-4045

PR #9900, opened 2026-09-22
feat(FR-4045): read the launcher's folder list and mount level from GraphQL

agatha197:
Resolves #9889 (FR-4045)

Stacked on #9751 (FR-3987) — base is `feat/FR-3987-vfolder-mount-policy-schema`, so review that one first.

## Summary

Review feedback on #9751: the session launcher's mountable-folder list is fetched over REST (`GET /folders`), and it should use GraphQL. This moves it.

**Why this is possible without a backend change.** `vfolder_nodes.permissions` is resolved **per caller** — the manager runs `resolve_mount_policy` (owner → the caller's policy row → the folder default, backend.ai#14679) and emits the result as mount verbs. So a folder the caller cannot mount carries no `mount_*` verb at all. Checked against the dev manager (10.82.130.172, 26.9.0):

| caller's resolved level | `permissions` carries |
|---|---|
| `rw` | `mount_ro`, `mount_rw`, `mount_wd` |
| `ro` | `mount_ro` |
| `none` | no `mount_*` verb |

The v2 `VFolder` type cannot do this: `accessControl.permission` is the folder **default**, and `VFolder.permissions` is the RBAC bit set (`READ`/`UPDATE`/`CREATE`/`SOFT_DELETE`/`HARD_DELETE`) with no mount levels. That is why the launcher stayed on REST when #9751 landed.

## Changes

- **`useSuspendedLegacyVFolders`** (`backend.ai-ui`) runs a Relay query on `vfolder_nodes` instead of `GET /folders`, and derives `ro` / `rw` / `none` from the mount verbs (`mountLevelFromPermissions`, `wd` folds into `rw`). The FR-3987 `none` exclusion is unchanged in meaning — it still reads a value that is the caller's own.
- Deleted rows are filtered server side (the same `status != …` expression the Data page uses). The query asks for one page of 500; `GET /folders` was unbounded, so this is a new ceiling worth a second opinion.
- **`is_owner` and `type` leave `LegacyVFolder`** — `vfolder_nodes` has neither. Nothing read them except `VFolderTable`'s Type-column sorter, which now sorts on the `ownership_type` its own renderer already used.
- Test fixtures answer `vfolder_nodes` from the same rows, so the existing prune tests keep asserting the same behaviour; `mountLevelFromPermissions` gets its own unit test.

## The open decision — do not merge before this is settled

**`ownerEmail` is gone**, from the hook, from `BAIVFolderMountConfigInput` and from both launcher steps. `vfolder_nodes` takes no owner parameter, so a session launched for another user now lists **the caller's** folders, not that user's.

This is a deliberate removal of working behaviour, and it is the reason this PR is a draft. From the FR-3987 thread:

- the current REST behaviour is itself inconsistent — `GET /folders?owner_user_email=B` scopes the *search* to B but computes each row's `permission` for the **caller**, so the level shown is not the level B would mount at;
- the position raised in review is that a user should not be able to enumerate another user's folders under RBAC at all;
- the suggestion was to replace this with an "acted as" superadmin feature, and the on-behalf case was excluded from the #9751 review.

I removed the prop rather than accepting and ignoring it, so every call site fails to compile instead of silently listing the wrong person's folders. `ownerEmailFromOwner` is left in place (now only exercised by its own test) because reinstating it is a one-liner if the decision goes the other way.

## Verification

- `bash scripts/verify.sh` → `=== ALL PASS ===` (with `VERIFY_TESTS=1`)
- Behaviour checked live on the #9751 dev server: with four project folders whose default is `none` and which have no policy row, the launcher's folder select omits all four and the remaining folders keep their R / R W tags.


agatha197:
Dev server on box `sujin`: stopped (was http://fr-4045-pr9900-read-launcher-s.sujin.10-82-0-159.sslip.io)

agatha197:
Folded into #9751 — the change is now a commit on that branch (`feat(FR-3987): read the launcher's folder list and mount level from GraphQL`) rather than a stacked PR, so the GraphQL switch and the `none` rendering it depends on land together. The open decision about `ownerEmail` / on-behalf listing moves to #9751's description. Closing.

